PR #10792, opened 2023-04-06
applications: nrf_desktop: Add devinfo config channel option to dfu

MarekPieta:
PR introduces device information configuration channel option to the dfu module. The option allows to fetch information about device's Vendor ID, Product ID and generation. The generation string allows to distinguish configurations that use the same board and bootloader, but are not interoperable.

PR also introduces needed follow-up changes to the HID configurator Python script.

Jira: NCSDK-19213

### DIFF
--- a/applications/nrf_desktop/doc/dfu.rst
+++ b/applications/nrf_desktop/doc/dfu.rst
@@ -48,6 +48,21 @@ If the buffer is small, the host must perform the DFU progress synchronization m
    For this reason, make sure that you use configuration with two image partitions.
    For more information on configuring the memory layout in the application, see the :ref:`nrf_desktop_flash_memory_layout` documentation.
 
+Device identification information
+=================================
+
+The DFU module provides the following information about the device through the :ref:`nrf_desktop_config_channel`:
+
+* Vendor ID (:ref:`CONFIG_DESKTOP_CONFIG_CHANNEL_DFU_VID <config_desktop_app_options>`)
+* Product ID (:ref:`CONFIG_DESKTOP_CONFIG_CHANNEL_DFU_PID <config_desktop_app_options>`)
+* Generation (:ref:`CONFIG_DESKTOP_CONFIG_CHANNEL_DFU_GENERATION <config_desktop_app_options>`)
+
+These values are fetched using the :ref:`devinfo <dfu_devinfo>` configuration channel option.
+
+.. note::
+   By default, the reported Vendor ID, Product ID, and generation are aligned with the values defined globally for the nRF Desktop application.
+   The default values of Kconfig options used by the DFU module are based on respectively :ref:`CONFIG_DESKTOP_DEVICE_VID <config_desktop_app_options>`, :ref:`CONFIG_DESKTOP_DEVICE_PID <config_desktop_app_options>` and :ref:`CONFIG_DESKTOP_DEVICE_GENERATION <config_desktop_app_options>`.
+
 Implementation details
 **********************
 
@@ -73,6 +88,7 @@ The module provides a simple protocol that allows the update tool on the host to
 The following :ref:`nrf_desktop_config_channel` options are available to perform the firmware update:
 
 * :ref:`fwinfo <dfu_fwinfo>` - Passes the information about the currently executed image from the device to the host.
+* :ref:`devinfo <dfu_devinfo>` - Passes the identification information about the device from the device to the host.
 * :ref:`reboot <dfu_reboot>` - Reboots the device.
 * :ref:`start <dfu_start>` - Starts the new update image transmission.
 * :ref:`data <dfu_data>` - Passes a chunk of the update image data from the host to the device.
@@ -86,6 +102,16 @@ fwinfo
 
    * Version and length of the image.
    * Partition ID of the image, used to specify the image placement on the flash.
+
+.. _dfu_devinfo:
+
+devinfo
+   Perform the fetch operation on this option to get the following information about the device:
+
+   * Vendor ID and Product ID
+   * Generation
+
+The device generation allows to distinguish configurations that use the same board and bootloader, but are not interoperable.
 
 .. _dfu_reboot:
 
--- a/applications/nrf_desktop/src/modules/Kconfig.config_channel
+++ b/applications/nrf_desktop/src/modules/Kconfig.config_channel
@@ -78,6 +78,37 @@ config DESKTOP_CONFIG_CHANNEL_DFU_SYNC_BUFFER_SIZE
 	  The host must perform progress synchronization at least
 	  every synchronization buffer bytes count.
 
+config DESKTOP_CONFIG_CHANNEL_DFU_VID
+	hex "Configuration channel Vendor ID"
+	range 0 0XFFFF
+	default DESKTOP_DEVICE_VID
+	help
+	  Configuration channel allows devices to report their Vendor ID. The
+	  Vendor ID is reported together with Product ID and device generation.
+
+	  The information is useful if nRF Desktop peripheral is connected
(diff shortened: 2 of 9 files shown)


scripts/hid_configurator/README.rst
```diff
@@ -268,6 +269,23 @@ To obtain information about the firmware running on the device, run the followin
...
+
+.. parsed-literal::
+    :class: highlight
+
+    python3 configurator_cli.py DEVICE devinfo
+
+.. note::
+  Only devices with :ref:`nrf_desktop_dfu` support the ``devinfo`` command.
+
+The command can be used to obtain Vendor ID and Product ID of devices connected via an nRF Desktop dongle.
```
mkapala-nordic, line 285: This can only be done when connected via dongle?

scripts/hid_configurator/modules/dfu.py
```diff
@@ -106,6 +106,24 @@ def __str__(self):
                                                  self.ver_rev, self.ver_build_nr)
 
 
+class DevInfo:
+    def __init__(self, fetched_data):
+        fmt = '<HH'
+        assert struct.calcsize(fmt) < len(fetched_data)
+        vals = struct.unpack(fmt, fetched_data[:struct.calcsize(fmt)])
+        self.vid, self.pid = vals
```
mkapala-nordic, line 114: Can't we unpack it directly to destination variables?

mkapala-nordic (approved):
Two minor comments/questions, besides it looks good.

applications/nrf_desktop/src/modules/Kconfig.config_channel
```diff
@@ -78,6 +78,38 @@ config DESKTOP_CONFIG_CHANNEL_DFU_SYNC_BUFFER_SIZE
 	  The host must perform progress synchronization at least
 	  every synchronization buffer bytes count.
 
+config DESKTOP_CONFIG_CHANNEL_DFU_GENERATION
```
kapi-no, line 81: BTW, are we planning to support the generation string for the MCUmgr DFU transport?

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -504,6 +506,33 @@ static void handle_dfu_bootloader_variant(uint8_t *data, size_t *size)
...
+		     CONFIG_CHANNEL_FETCHED_DATA_MAX_SIZE,
+		     "CONFIG_DESKTOP_CONFIG_CHANNEL_DFU_GENERATION is too long");
+
+	sys_put_le16(vid, &data[pos]);
+	pos += sizeof(vid);
+
+	sys_put_le16(pid, &data[pos]);
+	pos += sizeof(pid);
+
+	strcpy(&data[pos], generation);
```
kapi-no, line 530: You can consider using the TLV (Type, Length, Value) approach for each parameter. It is easier to maintain backward compatibility this way, as you may add a new type to the devinfo response packet that older versions of response packet parsers will ignore.

The lighter version would be to add a length field for the generation parameter as it will be possible to add another parameter after the Generation field. Currently, the generation value is considered to be the remaining bytes of the packet message and there is no way of adding new field in the future without breaking the compatibility.

scripts/hid_configurator/README.rst
MarekPieta: No, but there are other ways to get VID and PID for devices connected directly over USB or BLE (connection through dongle is the reason why we introduced the values to config channel)

applications/nrf_desktop/src/modules/Kconfig.config_channel
MarekPieta: Currently there is no plan to do it (MCUmgr DFU transport is not supported in the LVFS too)

applications/nrf_desktop/src/modules/Kconfig.config_channel
```diff
@@ -78,6 +78,38 @@ config DESKTOP_CONFIG_CHANNEL_DFU_SYNC_BUFFER_SIZE
...
+	string "Device generation"
+	default "default"
+	help
+	  The Kconfig option is used to specify device generation. The option
+	  allows to distinguish configurations that use the same board and
+	  bootloader, but are not interoperable.
```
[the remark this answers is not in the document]
MarekPieta: That's nice idea. Added. Thanks

applications/nrf_desktop/src/modules/dfu.c
MarekPieta: The configuration channel response can contain only up to the 16 bytes (in our case it's 4 bytes for VID+PID and remaining bytes can be used for the generation string). Using TLV could be more generic approach but it would introduce an additional overhead. Also we already have some options that return multiple values one after another in predefined format (e.g. `fwinfo`). The values returned for given option should not change, if needed we can always introduce a new configuration channel module's option to handle new data.

Please also keep in mind that our Python scripts rely on the fact that incoming data has predefined length in some cases to they may not be interoperable with the proposed expandable message format.

kapi-no: Right, 16 bytes seem too low for this type of format. Thanks for the detailed clarification

kapi-no (approved):

mkapala-nordic (approved):

scripts/hid_configurator/README.rst
```diff
@@ -268,6 +269,23 @@ To obtain information about the firmware running on the device, run the followin
 .. note::
   Only devices with :ref:`nrf_desktop_dfu` support the ``fwinfo`` command.
 
+Getting identification information about the device
+===================================================
+
+To obtain information about the device's Vendor ID, Product ID and generation, run the following command:
```
divipillai, line 275:
```suggestion
To obtain information about the device's Vendor ID, Product ID, and generation, run the following command:
```

divipillai (commented):
Nits.

applications/nrf_desktop/doc/dfu.rst
```diff
@@ -87,6 +103,16 @@ fwinfo
    * Version and length of the image.
    * Partition ID of the image, used to specify the image placement on the flash.
 
+.. _dfu_devinfo:
+
+devinfo
+   Perform the fetch operation on this option to get the following information about the device:
+
+   * Vendor ID and Product ID.
+   * Generation.
```
annwoj, line 112:
```suggestion
   * Vendor ID and Product ID
   * Generation
```

scripts/hid_configurator/README.rst
```diff
@@ -268,6 +269,23 @@ To obtain information about the firmware running on the device, run the followin
...
+
+To obtain information about the device's Vendor ID, Product ID, and generation, run the following command:
+
+.. parsed-literal::
+    :class: highlight
+
+    python3 configurator_cli.py DEVICE devinfo
+
+.. note::
+  Only devices with :ref:`nrf_desktop_dfu` support the ``devinfo`` command.
```
annwoj, line 283:
```suggestion
  Only devices with the :ref:`nrf_desktop_dfu` support the ``devinfo`` command.
```

scripts/hid_configurator/README.rst
```diff
@@ -268,6 +269,23 @@ To obtain information about the firmware running on the device, run the followin
...
+
+The command can be used to obtain Vendor ID and Product ID of devices connected via an nRF Desktop dongle.
```
annwoj, line 285:
```suggestion
The command can be used to obtain Vendor ID and Product ID of devices connected through an nRF Desktop dongle.
```

annwoj (commented):
Adding a few small suggestions

divipillai (approved):

annwoj (approved):

PR #10799, opened 2023-04-11
Bluetooth: controller: Allow CONFIG_BT_MAX_CONN=1

jori-nordic:
BT_MAX_CONN defines the number of host contexts. It seemed like a good idea at the time to piggyback on it for setting the number controller central/peripheral contexts, but that is not correct:

There is a valid use-case for an application that wants to be central or peripheral, but not simultaneously. In that case, enforcing this 1-1 relationship would waste memory in the host.

The upstream controller has no such restriction, so that also clears up build system headaches.

### DIFF
--- a/subsys/bluetooth/controller/hci_driver.c
+++ b/subsys/bluetooth/controller/hci_driver.c
@@ -35,23 +35,8 @@
 #include "zephyr/logging/log.h"
 LOG_MODULE_REGISTER(bt_sdc_hci_driver);
 
-#if defined(CONFIG_BT_CONN)
-/* It should not be possible to set CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT larger than
- * CONFIG_BT_MAX_CONN. Kconfig should make sure of that, this assert is to
- * verify that assumption.
- */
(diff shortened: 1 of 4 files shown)


subsys/bluetooth/controller/hci_driver.c
```diff
@@ -36,22 +36,14 @@
...
- */
-BUILD_ASSERT(CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT <= CONFIG_BT_MAX_CONN);
-
-#define SDC_CENTRAL_COUNT (CONFIG_BT_MAX_CONN - CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT)
 
-#else
-
-#define SDC_CENTRAL_COUNT 0
+BUILD_ASSERT(CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT <= CONFIG_BT_MAX_CONN);
+BUILD_ASSERT(CONFIG_BT_CTLR_SDC_CENTRAL_COUNT <= CONFIG_BT_MAX_CONN);
```
rugeGerritsen, line 41: I guess we do not need those build asserts now as it is already validated in KConfig?

rugeGerritsen (approved):

aescolar:
Nice, thanks :)

KAGA164 (approved):

alwa-nordic (approved):

PR #10803, opened 2023-04-12
applications: nrf_desktop: dfu: config channel and mcumgr coexistence

kapi-no:
Added support for coexistence between the config channel and MCUmgr DFU methods inside the nRF Desktop application.

Ref: NCSDK-20061


applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -27,10 +29,11 @@ LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_CONFIG_CHANNEL_DFU_LOG_LEVEL);
...
 /* DFU state values must match with values used by the host. */
-#define DFU_STATE_INACTIVE 0x00
-#define DFU_STATE_ACTIVE   0x01
-#define DFU_STATE_STORING  0x02
-#define DFU_STATE_CLEANING 0x03
+#define DFU_STATE_INACTIVE   0x00
+#define DFU_STATE_ACTIVE     0x01
+#define DFU_STATE_STORING    0x02
+#define DFU_STATE_CLEANING   0x03
+#define DFU_STATE_OTHER_HOST 0x04
```
MarekPieta, line 36: `DFU_STATE_BLE_SMP` ?

MarekPieta: Or maybe `ACTIVE_CONFIG_CHANNEL` and `ACTIVE_SMP` ?

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -188,6 +220,10 @@ static void background_erase_handler(struct k_work *work)
 
 	__ASSERT_NO_MSG(!is_flash_area_clean);
 
+	if (IS_ENABLED(CONFIG_DESKTOP_CONFIG_CHANNEL_DFU_MCUMGR_SUPPORT)) {
+		k_work_reschedule(&multi_dfu_driver_timeout, MULTI_DFU_DRIVER_TIMEOUT);
```
MarekPieta, line 224: Can we have active timeout at this point ?

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -27,10 +29,11 @@ LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_CONFIG_CHANNEL_DFU_LOG_LEVEL);
 
 
 /* DFU state values must match with values used by the host. */
-#define DFU_STATE_INACTIVE 0x00
-#define DFU_STATE_ACTIVE   0x01
-#define DFU_STATE_STORING  0x02
-#define DFU_STATE_CLEANING 0x03
+#define DFU_STATE_INACTIVE   0x00
+#define DFU_STATE_ACTIVE     0x01
```
MarekPieta, line 33: `DFU_STATE_ACTIVE` -> `DFU_STATE_ACTIVE_CONFIG_CHANNEL`

applications/nrf_desktop/src/modules/dfu.c
MarekPieta, line 36: `DFU_STATE_OTHER_HOST` -> `DFU_STATE_ACTIVE_OTHER`

MarekPieta (changes requested):
I also wonder if we shouldn't split functionalities here (use 2 application modules for DFU transports and a separate utility to handle locking; the image slot erase operation could be part of the utility too - module that claimed secondary slot could request image cleanup and be notified once the operation is complete).
The main rationale is the fact that DFU module is becoming complicated. Smaller modules with clean API and responsibilities split are far less error-prone and easier to maintain I think.

applications/nrf_desktop/src/util/dfu_lock.c
```diff
@@ -0,0 +1,85 @@
...
+LOG_MODULE_REGISTER(dfu_lock, CONFIG_DESKTOP_DFU_LOCK_LOG_LEVEL);
+
+#include "dfu_lock.h"
+
+static const struct dfu_lock_owner *previous_owner;
+static const struct dfu_lock_owner *current_owner;
+
+static atomic_t dfu_lock = ATOMIC_INIT(false);
+
+bool dfu_lock_try(const struct dfu_lock_owner *new_owner)
```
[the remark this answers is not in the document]
MarekPieta: I wonder if flash accesses shouldn't actually go through the locking utility (so that we would actually enforce proper API usage)? That would make solution far less error-prone.
We could either gate read/write/erase or maybe at least opening and closing the flash area.

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,138 @@
...
+
+		return false;
+	}
+
+	if (is_module_state_event(aeh)) {
+		const struct module_state_event *event =
+			cast_module_state_event(aeh);
+
+		if (check_state(event, MODULE_ID(main), MODULE_STATE_READY)) {
+#if CONFIG_BOOTLOADER_MCUBOOT
```
MarekPieta, line 115: Is it actually possible to use the module without the MCUboot bootloader?

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -106,6 +109,18 @@ const static char * const opt_descr[] = {
 	[DFU_OPT_DEVINFO] = "devinfo"
 };
 
+static void dfu_lock_owner_changed(const struct dfu_lock_owner *new_owner)
+{
+	LOG_DBG("Flash marked dirty due to different DFU process");
+
+	schedule_erase_on_dfu_start = true;
```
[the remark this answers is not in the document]
MarekPieta: Whole invalidation of DFU state could be done here. That would prevent reporting outdated data during `handle_dfu_sync`.

applications/nrf_desktop/src/util/dfu_lock.c
```diff
@@ -0,0 +1,85 @@
...
+	previous_owner = current_owner;
+	current_owner = NULL;
+
+	is_ownership_released = atomic_cas(&dfu_lock, true, false);
+	__ASSERT_NO_MSG(is_ownership_released);
+
+	LOG_DBG("DFU lock released by %s", owner->name);
+}
+
+bool dfu_lock_owner_check(const struct dfu_lock_owner *owner)
```
MarekPieta, line 63: Not sure if actually needed, module could remember that it successfully locked flash and that the module is currently the owner (if we guard opening FLASH area, then module could just check that it obtained the FLASH area).

MarekPieta (changes requested):

applications/nrf_desktop/src/util/dfu_lock.c
MarekPieta, line 21: Discussed offline. Please change it do `dfu_lock_claim`

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -545,15 +639,27 @@ static void handle_dfu_devinfo(uint8_t *data, size_t *size)
...
+	bool is_lock_claimed;
 
-	if (IS_ENABLED(CONFIG_CAF_POWER_MANAGER_EVENTS)) {
-		power_manager_restrict(MODULE_IDX(MODULE), POWER_MANAGER_LEVEL_SUSPENDED);
+	*size = sizeof(bool);
+
+	is_lock_claimed = true;
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK)) {
+		is_lock_claimed = dfu_lock_claim(&config_channel_owner);
+		/* No need to release the lock at this point as the system will reboot. */
```
MarekPieta, line 649: You can actually release the lock after calling `sys_reboot` (so that you would not rely on the reboot implementation). Handling such case properly would not hurt (may be better than commenting here actually)

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -754,6 +860,13 @@ static bool app_event_handler(const struct app_event_header *aeh)
 			k_work_init_delayable(&background_erase, background_erase_handler);
 			k_work_init_delayable(&background_store, background_store_handler);
 
+			if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK)) {
+				bool is_lock_claimed;
+
+				is_lock_claimed = dfu_lock_claim(&config_channel_owner);
```
MarekPieta, line 866: Maybe we could combine claiming lock with background erase scheduling for the firs time (both operations must be done together)

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -337,6 +374,15 @@ static void handle_dfu_data(const uint8_t *data, size_t size)
 		return;
 	}
 
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK)) {
```
MarekPieta, line 377: You could make it `if(IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) && !dfu_lock_claim() ...{ return}`

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -227,6 +253,12 @@ static void background_erase_handler(struct k_work *work)
 			LOG_ERR("Cannot erase page (%d)", err);
 			flash_area_close(flash_area);
 			flash_area = NULL;
+
+			if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) &&
+			    !k_work_delayable_is_pending(&dfu_timeout)) {
```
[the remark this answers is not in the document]
MarekPieta: That may also break DFU logic since we rely on fact that FLASH area is open to determine that DFU operation is already ongoing

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,147 @@
...
+#include "dfu_lock.h"
+
+#define MODULE dfu_mcumgr
+#include <caf/events/module_state_event.h>
+#include <caf/events/ble_smp_event.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_DFU_MCUMGR_LOG_LEVEL);
+
+#define DFU_TIMEOUT	K_SECONDS(5)
```
MarekPieta, line 20: Maybe we could make it configurable ?

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,147 @@
...
+static int32_t mcumgr_img_mgmt_cb(uint32_t event,
+				  int32_t rc,
+				  bool *abort_more,
+				  void *data,
+				  size_t data_size)
+{
+	LOG_INF("MCUmgr Image Management Event with the %d ID",
+		__builtin_ctz(MGMT_EVT_GET_ID(event)));
+
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK)) {
```
MarekPieta, line 59: `if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) && !dfu_lock_claim) {return MGMT_ERR_EACCESSDENIED;}`

MarekPieta: Same for reboot handler

applications/nrf_desktop/src/modules/dfu.c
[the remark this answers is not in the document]
kapi-no: Yes, it can. Consider the following scenario:

Config channel DFU source stops sending DFU packages at some point. Once the DFU is still not timed out, the DFU source will restart DFU from scratch setting the image offset to zero. The background erase operation will be scheduled in this corner case before the DFU timeout.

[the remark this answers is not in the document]
MarekPieta: Wouldn't the following check in `handle_dfu_start` function prevent the subsequent DFU from starting?
```
        if (flash_area) {                                                                            
                LOG_WRN("DFU already in progress");                                                  
                return;                                                                              
        }
```

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -545,15 +660,27 @@ static void handle_dfu_devinfo(uint8_t *data, size_t *size)
 static void handle_reboot_request(uint8_t *data, size_t *size)
 {
 	LOG_INF("System reboot requested");
+	bool is_lock_claimed;
 
-	if (IS_ENABLED(CONFIG_CAF_POWER_MANAGER_EVENTS)) {
-		power_manager_restrict(MODULE_IDX(MODULE), POWER_MANAGER_LEVEL_SUSPENDED);
+	*size = sizeof(bool);
+
+	is_lock_claimed = true;
```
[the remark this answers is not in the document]
MarekPieta: Please keep in mind that the reboot request may be sent by host at any time. The work used to handle operation may be submitted e.g. during an ongoing background erase or DFU transfer operation (the DFU timeout may even happen between the reboot work is submitted here and the reboot work handler is actually executed - I guess that releasing the DFU lock in such case is not expected).

applications/nrf_desktop/src/util/Kconfig
```diff
@@ -21,4 +21,18 @@ config DESKTOP_ADV_PROV_UUID16_ALL
 	  Adds all UUID16 to the advertising payload if used Bluetooth local
 	  identity has no bond.
 
+config DESKTOP_DFU_LOCK
+	bool "DFU lock module"
```
MarekPieta, line 25: I wonder if we should not switch from implying the option in DFU transports to setting the option to `y` by default here if both transports are enabled (thanks to that we would keep the DFU lock module disabled when not needed without need of explicitly disabling it in most of the configurations). In such case, maybe we could  actually accept that the module would know about the clients (both DFU lock and clients are part of nRF Desktop application).

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,138 @@
...
+	if (atomic_cas(&mcumgr_event_active, false, true)) {
+		APP_EVENT_SUBMIT(new_ble_smp_transfer_event());
+	}
+
+	return MGMT_ERR_EOK;
+}
+
+static struct mgmt_callback img_mgmt_callback = {
+	.callback = mcumgr_img_mgmt_cb,
+	.event_id = MGMT_EVT_OP_IMG_MGMT_ALL,
```
MarekPieta, line 76: Should we react on all of the operations or only on `MGMT_EVT_OP_IMG_MGMT_DFU_CHUNK`? Our returned value seems ignored when we are notified about operations other than `MGMT_EVT_OP_IMG_MGMT_DFU_CHUNK`.

MarekPieta: Also I actually noticed that the mgumgr img mgmt may require further improvements - e.g. we cannot prevent from triggering img mgmt erase commend as an application (`img_mgmt_erase` is handled without application's approval)
I think that most of such commands require user action (so potential consequences would be limited), but we should still keep this in mind

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dfu_mcumgr.c
kapi-no: > Should we react on all of the operations or only on `MGMT_EVT_OP_IMG_MGMT_DFU_CHUNK`? Our returned value seems ignored when we are notified about operations other than `MGMT_EVT_OP_IMG_MGMT_DFU_CHUNK`.

Yes, but returning an error does not do any harm. We need to react to other events to reschedule DFU timeout on MCUmgr activity

applications/nrf_desktop/src/modules/dfu.c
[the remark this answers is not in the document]
kapi-no: You are right, but we need to guard against the pending reboot handler

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,138 @@
...
+
+		return false;
+	}
+
+	if (is_module_state_event(aeh)) {
+		const struct module_state_event *event =
+			cast_module_state_event(aeh);
+
+		if (check_state(event, MODULE_ID(main), MODULE_STATE_READY)) {
+			int err = boot_write_img_confirmed();
```
MarekPieta, line 116: btw. Should we confirm the image also in case MCUboot+XIP without revert is used? I think that confirmation is not needed in that case (see DFU application module for reference)

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,143 @@
...
+
+	k_work_reschedule(&dfu_timeout, DFU_TIMEOUT);
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) && dfu_lock_claim(&mcumgr_owner)) {
+		(void)k_work_cancel_delayable(&dfu_timeout);
+		return MGMT_ERR_EACCESSDENIED;
+	}
+
+	if (atomic_cas(&mcumgr_event_active, false, true)) {
+		APP_EVENT_SUBMIT(new_ble_smp_transfer_event());
+	}
```
MarekPieta, line 71: Maybe the event could be submitted only if the mcumgr uses BLE transport? Your implementation does not enforce transport (unlinke CAF's implementation)

applications/nrf_desktop/src/modules/dfu_mcumgr.c
MarekPieta: Ok, accepted

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,143 @@
...
+
+		return false;
+	}
+
+	if (is_module_state_event(aeh)) {
+		const struct module_state_event *event =
+			cast_module_state_event(aeh);
+
+		if (check_state(event, MODULE_ID(main), MODULE_STATE_READY)) {
+			if (!IS_ENABLED(CONFIG_BOOT_BUILD_DIRECT_XIP_VARIANT)) {
```
MarekPieta, line 119: Can we make a new Kconfig option to allow user to change behavior here (for consistency with DFU module implementation) ?

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,143 @@
...
+static atomic_t mcumgr_event_active = ATOMIC_INIT(false);
+
+/* Declare the function here as it is not available in the MCUmgr API. */
+void img_mgmt_reset_upload(void);
+
+static void dfu_lock_owner_changed(const struct dfu_lock_owner *new_owner)
+{
+	LOG_DBG("MCUmgr progress reset due to different DFU process operations");
+
+	img_mgmt_reset_upload();
```
MarekPieta, line 34: That may actually not be safe (but I think we need to leave it as is since we currently have no better way of doing it...). Please open a follow-up ticket to improve it later on.

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -171,6 +186,16 @@ static void terminate_dfu(void)
...
 	}
+
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) &&
+	    !k_work_delayable_is_pending(&background_erase) &&
+	    !k_work_delayable_is_pending(&reboot_request)) {
+		int err;
+
+		err = dfu_lock_release(&config_channel_owner);
+		__ASSERT_NO_MSG(!err);
+		ARG_UNUSED(err);
```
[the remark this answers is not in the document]
MarekPieta: Then we would be sure that lock would not be released prematurely

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -337,6 +400,10 @@ static void handle_dfu_data(const uint8_t *data, size_t size)
 		return;
 	}
 
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) && dfu_lock_claim(&config_channel_owner)) {
+		return;
+	}
```
MarekPieta, line 405: Maybe we could assert on claiming lock success? If flash area is open we should be sure that config channel DFU has already started and therefore we must be the DFU lock owner.

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -171,6 +186,16 @@ static void terminate_dfu(void)
 	if (IS_ENABLED(CONFIG_CAF_POWER_MANAGER_EVENTS)) {
 		power_manager_restrict(MODULE_IDX(MODULE), POWER_MANAGER_LEVEL_MAX);
 	}
+
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) &&
+	    !k_work_delayable_is_pending(&background_erase) &&
```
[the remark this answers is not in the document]
kapi-no: No, it cannot (leftover from previous approach)

applications/nrf_desktop/src/modules/dfu.c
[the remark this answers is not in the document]
kapi-no: > checks for all of the pending works and assertion on the returned value).

It is not possible to do that because you need to have different pending work checks based on the current execution context (in the background erase you cannot check if background erase is pending)

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -754,6 +890,14 @@ static bool app_event_handler(const struct app_event_header *aeh)
...
 
+			if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK)) {
+				int ret;
+
+				ret = dfu_lock_claim(&config_channel_owner);
+				__ASSERT_NO_MSG(!ret);
+				ARG_UNUSED(ret);
+			}
+
 			k_work_reschedule(&background_erase, K_NO_WAIT);
```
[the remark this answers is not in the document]
kapi-no: We reschedule the background erase operation without claiming the lock in the background erase handler as we know that the lock is claimed at that point

kapi-no:
Last force push: rebase on the upmerge changes without any diff change in the scope of this PR

applications/nrf_desktop/src/modules/dfu.c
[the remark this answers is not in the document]
MarekPieta: That could be left as is (I just wanted to claim lock together with the first scheduling operation)

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,144 @@
...
+	LOG_INF("MCUmgr Image Management Event with the %d ID",
+		u32_count_trailing_zeros(MGMT_EVT_GET_ID(event)));
+
+	k_work_reschedule(&dfu_timeout, DFU_TIMEOUT);
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) && dfu_lock_claim(&mcumgr_owner)) {
+		(void)k_work_cancel_delayable(&dfu_timeout);
+		return MGMT_ERR_EACCESSDENIED;
+	}
+
+	if (IS_ENABLED(CONFIG_MCUMGR_TRANSPORT_BT) &&
```
MarekPieta, line 69: If this is not enabled I would also desist from enabling the event and subscribing for it.

applications/nrf_desktop/src/modules/Kconfig.dfu_mcumgr
```diff
@@ -0,0 +1,42 @@
...
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menuconfig DESKTOP_DFU_MCUMGR_ENABLE
+	bool "Enable DFU using MCUmgr"
+	select CAF_BLE_SMP_TRANSFER_EVENTS if MCUMGR_TRANSPORT_BT
+	depends on !CAF_BLE_SMP
+	depends on MCUMGR_MGMT_NOTIFICATION_HOOKS
```
MarekPieta, line 11: Consider using `select` instead of `depends on` to simplify configuration of the module (allow user to enable it with a single Kconfig).

applications/nrf_desktop/src/modules/dfu_mcumgr.c
kapi-no: created a task for it

applications/nrf_desktop/src/modules/Kconfig.dfu_mcumgr
```diff
@@ -0,0 +1,42 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menuconfig DESKTOP_DFU_MCUMGR_ENABLE
+	bool "Enable DFU using MCUmgr"
+	select CAF_BLE_SMP_TRANSFER_EVENTS if MCUMGR_TRANSPORT_BT
+	select MCUMGR
```
MarekPieta, line 10: We also need to select the missing dependencies here, for example these:
```
select NET_BUF
select ZCBOR
select CRC
select IMG_MANAGER
select STREAM_FLASH
select FLASH_MAP
select FLASH
```

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,150 @@
...
+				  size_t data_size)
+{
+	LOG_INF("MCUmgr Image Management Event with the %d ID",
+		u32_count_trailing_zeros(MGMT_EVT_GET_ID(event)));
+
+	k_work_reschedule(&dfu_timeout, DFU_TIMEOUT);
+	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) && dfu_lock_claim(&mcumgr_owner)) {
+		(void)k_work_cancel_delayable(&dfu_timeout);
+		return MGMT_ERR_EACCESSDENIED;
+	}
```
MarekPieta, line 70: To make this operation safe, I think we need to somehow ensure that the taken lock is kept until mcumgr finishes storing the DFU chunk (otherwise mcumgr may be storing while mcumgr dfu times out). Maybe we could use `CONFIG_MCUMGR_SMP_COMMAND_STATUS_HOOKS` for that purpose (we could wait for the hook after each DFU chunk and only after the hook was called we would release the dfu_lock to ensure we write to FLASH only in locked state).

MarekPieta: Since this would require quite significant amount of work, we may consider doing it in a follow-up PR.

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -416,6 +486,12 @@ static void handle_dfu_start(const uint8_t *data, const size_t size)
 				csum, img_csum,
 				offset, cur_offset);
 
+			if (!k_work_delayable_is_pending(&dfu_timeout) &&
+			    !k_work_delayable_is_pending(&background_erase) &&
```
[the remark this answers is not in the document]
kapi-no: you are right

applications/nrf_desktop/src/modules/dfu.c
```diff
@@ -442,6 +518,12 @@ static void handle_dfu_start(const uint8_t *data, const size_t size)
 		LOG_ERR("Cannot open flash area (%d)", err);
 
 		flash_area = NULL;
+
+		if (!k_work_delayable_is_pending(&dfu_timeout) &&
+		    !k_work_delayable_is_pending(&background_erase) &&
```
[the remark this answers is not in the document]
kapi-no: yes, as in your first comment

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -0,0 +1,150 @@
...
+
+static struct k_work_delayable dfu_timeout;
+
+static atomic_t mcumgr_event_active = ATOMIC_INIT(false);
+
+/* nRF Desktop MCUmgr DFU module cannot be enabled together with the CAF BLE SMP module. */
+BUILD_ASSERT(!IS_ENABLED(CONFIG_CAF_BLE_SMP));
+
+/* Declare the function here as it is not available in the MCUmgr API. */
+void img_mgmt_reset_upload(void);
```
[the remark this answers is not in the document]
kapi-no: sure

applications/nrf_desktop/src/modules/dfu_mcumgr.c
kapi-no: Created a task for it based on your feedback

alstrzebonski (approved):

mkapala-nordic (approved):

MarekPieta (approved):

kapi-no:
Closing PR as the change is already on main (due to some GitHub bug).

PR #10809, opened 2023-04-12
samples: add Anjay-LwM2M-Client to the nRF samples

JZimnol:
Add support for demo application of [`Anjay-zephyr-client`](https://github.com/AVSystem/Anjay-zephyr-client) and add it as sample for `nRF` boards. This sample supports `thingy91_nrf9160_ns`, `nrf9160dk_nrf9160_ns` (also with external flash) and `nrf52840dk_nrf52840`.

The `README.rst` file contains all the required instructions on how to build and flash application. This is a part of `AVSystem integration`.


SeppoTakalo (changes requested):
Hi @JZimnol I have started an separate email discussion with AVSystem regarding this PR.

The summary is that as of now, Nordic will not accept this PR in. The PR adds 3rd party dependency and code with difference license into our SDK.

We instead recommend that Anjay would be developed in a model called [Zephyr Workspace application](https://docs.zephyrproject.org/3.2.0/develop/application/index.html#zephyr-workspace-app) so that it can point to nRF Connect SDK releases in its west.yaml file, but would not be distributed by Nordic.

PR #10813, opened 2023-04-13
lib: azure_iot_hub: Increase default username buffer size

junqingzou:
Increase the default username buffer size from 128 to 160, which is value based on experiment.

Also correct minor issue in doc.

### DIFF
--- a/doc/nrf/libraries/networking/azure_iot_hub.rst
+++ b/doc/nrf/libraries/networking/azure_iot_hub.rst
@@ -99,8 +99,7 @@ To connect to Azure IoT Hub without using DPS, complete the following minimum re
    If the ``device_id.size`` buffer size is zero, the compile-time option :kconfig:option:`CONFIG_AZURE_IOT_HUB_DEVICE_ID`` is used.
 #. Make sure that the device is already registered with your Azure IoT Hub, or follow the instructions in `Registering the device with Azure IoT Hub`_.
 #. Set the :kconfig:option:`CONFIG_MQTT_HELPER_SEC_TAG` option to the security tag used in :ref:`azure_iot_hub_flash_certs`.
-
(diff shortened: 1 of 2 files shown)


simensrostad (approved):

wiba-nordic (approved):

jtguggedal (approved):

doc/nrf/libraries/networking/azure_iot_hub.rst
```diff
@@ -99,8 +99,7 @@ To connect to Azure IoT Hub without using DPS, complete the following minimum re
    If the ``device_id.size`` buffer size is zero, the compile-time option :kconfig:option:`CONFIG_AZURE_IOT_HUB_DEVICE_ID`` is used.
 #. Make sure that the device is already registered with your Azure IoT Hub, or follow the instructions in `Registering the device with Azure IoT Hub`_.
 #. Set the :kconfig:option:`CONFIG_MQTT_HELPER_SEC_TAG` option to the security tag used in :ref:`azure_iot_hub_flash_certs`.
```
nordicjm, line 101: Seems to be missed from update, and why are these being updates anyway? Commit title says a username buffer Kconfig is being updated, this has no relation

nordicjm (changes requested):

doc/nrf/libraries/networking/azure_iot_hub.rst
junqingzou: This is the primary sec_tag, comes with no _SECONDARY_

nordicjm: This needs splitting up into 2 commits, the changes here have no relation to a username buffer size

nordicjm (changes requested):

nordicjm (approved):

PR #10814, opened 2023-04-13
bluetooth: update manifest

olivier-le-sage:


hermabe (approved):

subsys/bluetooth/controller/Kconfig
```diff
@@ -246,6 +246,13 @@ config BT_SDC_ADDITIONAL_MEMORY
 	help
 	  This is for testing and internal use only.
 
+config BT_CTLR_FAL_SIZE
+	int "Configures the maximum number of addresses in the Filter Accept List"
+	range 0 255
+	default 8
+	help
```
laxiLang, line 253: no dependency for this config?

olivier-le-sage: none that I can think of ? whitelisting is used by all SDC variants

olivier-le-sage: I can't think of a BLE use-case involving neither advertising nor scanning

laxiLang (approved):

PR #10834, opened 2023-04-13
Bluetooth: Mesh: Fix in scheduler message validity test

akredalen:
Fixed bug where ctx was passed as NULL to scheduler client api without any defined publish parameters, triggering a segmentation fault in model_ackd_timeout_get.

Note: I hope that this will fix the bug reported by @balaji-nordic, however, I am not able to reproduce the bug locally to verify.


alxelax (changes requested):
Please check with Valgrind. If you need assistance running it, please let me know.

balaji-nordic (approved):
The fix works. The test runs and passes when optimizations are disabled (it is failing in main when optimizations are disabled).

alxelax (approved):

ludvigsj (approved):

PR #10838, opened 2023-04-14
lib: pdn: only execute init hook if initialization is successful

lemrey:
Only perfom the setup steps in libmodem init hook if libmodem's initialization was successful. The setup steps consist of AT commands that would otherwise fail and print errors.

@jtguggedal `memfault` init hook doesn't seem to check the initialization result either.
Shall I add a commit to fix that, or is there a reason why it is not checking it?

### DIFF
--- a/lib/pdn/pdn.c
+++ b/lib/pdn/pdn.c
@@ -176,6 +176,11 @@ static void on_modem_init(int ret, void *ctx)
 	int err;
 	(void) err;
 
+	if (ret != 0) {
+		/* Return if modem initialization failed */
+		return;
+	}
+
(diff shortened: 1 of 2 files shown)


stig-bjorlykke (approved):

nordic-mare (approved):

jtguggedal:
> `memfault` init hook doesn't seem to check the initialization result either.
> Shall I add a commit to fix that, or is there a reason why it is not checking it?

@lemrey That's an oversight, good catch. If you would add a commit fixing it, that would be great.

rlubos (approved):

PR #10840, opened 2023-04-14
application: serial_lte_modem: Allow auto-connect for NIDD

junqingzou:
For the use case of FOTA over NIDD with lwm2m_carrier, 
.Enable auto-connect NIDD after FOTA image is applied.
.Enable auto-connect NIDD on all kinds of reset.
.Add `AT#XCARRIER="auto-connect","read|write"[,<auto-connect-flag>]` to control YES/NO of auto-connect.

Dependence: libmodem v2.4.2

UPDATE: the solution has been expanded to support both U-bind and N-bind for LwM2M_Carrier.

### DIFF
--- a/applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
@@ -76,6 +76,7 @@ Syntax
 
 The ``<cmd>`` command is a string, and can be used as follows:
 
+* ``AT#XCARRIER="auto_connect","read|write"[,<auto-connect-flag>]``
 * ``AT#XCARRIER="app_data"[,<data>]``
 * ``AT#XCARRIER="battery_level",<battery_level>``
 * ``AT#XCARRIER="battery_status",<battery_status>``
@@ -106,6 +107,19 @@ The response syntax depends on the commands used.
 
 Examples
 ~~~~~~~~
+
+::
+
+   AT#XCARRIER="auto_connect","read"
+   #XCARRIER: auto_connect 1
+   OK
+
+   AT#XCARRIER="auto_connect","write",0
+   OK
+
+   AT#XCARRIER="auto_connect","read"
+   #XCARRIER: auto_connect 0
+   OK
 
 ::
 
--- a/applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
+++ b/applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
@@ -10,12 +10,15 @@
 #include <zephyr/logging/log.h>
 #include <modem/nrf_modem_lib.h>
 #include "slm_at_host.h"
+#include "slm_settings.h"
 #include "slm_util.h"
 
 LOG_MODULE_REGISTER(slm_carrier, CONFIG_SLM_LOG_LEVEL);
 
 /**@brief LwM2M Carrier operations. */
 enum slm_carrier_operation {
+	/* LTE auto-connect operation for LwM2M Carrier */
+	CARRIER_OP_AUTO_CONNECT,
 	/* Carrier AppData Operation */
 	CARRIER_OP_APPDATA_SEND,
 	/* Carrier Device Operation */
@@ -54,6 +57,7 @@ struct carrier_op_list {
 static int m_mem_free;
 
 /** forward declaration of cmd handlers **/
+static int do_carrier_auto_connect(void);
 static int do_carrier_appdata_send(void);
 static int do_carrier_device_battery_level(void);
 static int do_carrier_device_battery_status(void);
@@ -76,6 +80,7 @@ static int do_carrier_request_link_up(void);
 
 /**@brief SLM AT Command list type. */
 static struct carrier_op_list op_list[CARRIER_OP_MAX] = {
+	{CARRIER_OP_AUTO_CONNECT, "auto_connect", do_carrier_auto_connect},
 	{CARRIER_OP_APPDATA_SEND, "app_data", do_carrier_appdata_send},
 	{CARRIER_OP_DEVICE_BATTERY_LEVEL, "battery_level", do_carrier_device_battery_level},
 	{CARRIER_OP_DEVICE_BATTERY_STATUS, "battery_status", do_carrier_device_battery_status},
@@ -99,10 +104,16 @@ static struct carrier_op_list op_list[CARRIER_OP_MAX] = {
 
 #define SLM_CARRIER_OP_STR_MAX (sizeof("battery_status") + 1)
 
-/* Global variables and functions defined in different files. */
+struct k_work_delayable reconnect_work;
+
+/* Global variables defined in different files. */
+extern int32_t auto_connect;
 extern struct k_work_q slm_work_q;
 extern struct at_param_list at_param_list;
 extern uint8_t data_buf[SLM_MAX_MESSAGE_SIZE];
+
+/* Global functions defined in different files. */
+int lte_auto_connect(void);
 
 static void print_err(const lwm2m_carrier_event_t *evt)
 {
@@ -178,15 +189,26 @@ static void on_event_app_data(const lwm2m_carrier_event_t *event)
 	data_send(data_buf, app_data->buffer_len);
 }
 
+static void reconnect_wk(struct k_work *work)
+{
+	(void)lte_auto_connect();
+}
+
 int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 {
 	int err = 0;
+	static bool fota_started;
 
 	switch (event->type) {
 	case LWM2M_CARRIER_EVENT_LTE_LINK_UP:
 		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_LINK_UP");
-		/* AT+CFUN=1 to be issued. */
-		err = -1;
+		if (fota_started) {
+			fota_started = false;
+			k_work_reschedule(&reconnect_work, K_MSEC(100));
+		} else {
+			/* AT+CFUN=1 to be issued. */
+			err = -1;
+		}
 		break;
 	case LWM2M_CARRIER_EVENT_LTE_LINK_DOWN:
 		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_LINK_DOWN");
@@ -209,6 +231,7 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 		break;
 	case LWM2M_CARRIER_EVENT_FOTA_START:
 		LOG_DBG("LWM2M_CARRIER_EVENT_FOTA_START");
+		fota_started = true;
 		break;
 	case LWM2M_CARRIER_EVENT_FOTA_SUCCESS:
 		LOG_DBG("LWM2M_CARRIER_EVENT_FOTA_SUCCESS");
@@ -759,6 +782,37 @@ static int do_carrier_request_link_up(void)
 	return lwm2m_carrier_request(LWM2M_CARRIER_REQUEST_LINK_UP);
 }
 
+/* AT#XCARRIER="auto_connect","read|write"[,<auto-connect-flag>] */
+static int do_carrier_auto_connect(void)
+{
+	int ret = 0;
+	int flag;
+	char operation[6];
+	size_t size = sizeof(operation);
+
+	ret = util_string_get(&at_param_list, 2, operation, &size);
+	if (ret) {
+		return ret;
+	}
+
+	if (slm_util_cmd_casecmp(operation, "READ")) {
+		rsp_send("\r\n#XCARRIER: auto_connect %d\r\n", auto_connect);
+	} else if (slm_util_cmd_casecmp(operation, "WRITE")) {
+		ret = at_params_int_get(&at_param_list, 3, &flag);
+		if (ret) {
+			return ret;
+		}
+		if (flag == 0 || flag == 1) {
+			auto_connect = flag;
+			(void)slm_settings_auto_connect_save();
+		} else {
+			ret = -EINVAL;
+		}
+	}
+
+	return ret;
+}
+
 /**@brief API to handle Carrier AT command
  */
 int handle_at_carrier(enum at_cmd_type cmd_type)
@@ -787,6 +841,8 @@ int handle_at_carrier(enum at_cmd_type cmd_type)
 
 int slm_at_carrier_init(void)
 {
+	k_work_init_delayable(&reconnect_work, reconnect_wk);
+
 	return 0;
 }
 
--- a/applications/serial_lte_modem/src/main.c
+++ b/applications/serial_lte_modem/src/main.c
@@ -39,6 +39,9 @@ static struct k_work_delayable indicate_work;
 /* global variable used across different files */
 struct k_work_q slm_work_q;
 
+/* global variable defined in different files */
+extern int32_t auto_connect;
+
 /* global functions defined in different files */
 int poweron_uart(void);
 
@@ -46,6 +49,18 @@ int poweron_uart(void);
 static void indicate_wk(struct k_work *work);
 
 BUILD_ASSERT(CONFIG_SLM_WAKEUP_PIN >= 0, "Wake up pin not configured");
+
+NRF_MODEM_LIB_ON_INIT(lwm2m_init_hook, on_modem_lib_init, NULL);
+
+static void on_modem_lib_init(int ret, void *ctx)
+{
+	ARG_UNUSED(ctx);
+
+	/** ret: Zero on success, a positive value @em nrf_modem_dfu when executing
+	 *  Modem firmware updates, and negative errno on other failures.
+	 */
+	LOG_INF("lib_modem init: %d", ret);
+}
 
 #if defined(CONFIG_NRF_MODEM_LIB_ON_FAULT_APPLICATION_SPECIFIC)
 static void on_modem_failure_shutdown(struct k_work *item);
@@ -333,6 +348,39 @@ static void handle_mcuboot_swap_ret(void)
 	}
 }
 
+int lte_auto_connect(void)
+{
+	int err = 0;
+
+#if defined(CONFIG_SLM_CARRIER)
+	int stat;
+
+	if (!auto_connect) {
+		return 0;
+	}
+	err = nrf_modem_at_scanf("AT+CEREG?", "+CEREG: %d", &stat);
+	if (err != 1 || (stat == 1 || stat == 5)) {
+		return 0;
+	}
+
+	LOG_INF("auto connect");
+#if defined(CONFIG_LWM2M_CARRIER_SERVER_BINDING_N)
+	err = nrf_modem_at_printf("AT%%XSYSTEMMODE=0,1,0,0");
+	if (err) {
+		LOG_ERR("Failed to configure RAT: %d", err);
+		return err;
+	}
+#endif /* CONFIG_LWM2M_CARRIER_SERVER_BINDING_N */
+	err = nrf_modem_at_printf("AT+CFUN=1");
+	if (err) {
+		LOG_ERR("Failed to turn on radio: %d", err);
+		return err;
+	}
+#endif /* CONFIG_SLM_CARRIER */
+
+	return err;
+}
+
 int start_execute(void)
 {
 	int err;
@@ -351,6 +399,7 @@ int start_execute(void)
 	k_work_queue_start(&slm_work_q, slm_wq_stack_area,
 			   K_THREAD_STACK_SIZEOF(slm_wq_stack_area),
 			   SLM_WQ_PRIORITY, NULL);
+	(void)lte_auto_connect();
 
 	return 0;
 }
--- a/applications/serial_lte_modem/src/slm_settings.c
+++ b/applications/serial_lte_modem/src/slm_settings.c
@@ -23,6 +23,8 @@ enum fota_stage fota_stage = FOTA_STAGE_INIT;		/* FOTA: stage of FOTA process */
 enum fota_status fota_status;		/* FOTA: OK/Error status */
 int32_t fota_info;		/* FOTA: failure cause in case of error or download percentage*/
 
+int32_t auto_connect = 1;	/* Auto connect network at startup or after FOTA update */
+
 bool uart_configured;		/* UART: first-time configured */
 struct uart_config slm_uart;	/* UART: config */
 
@@ -47,6 +49,11 @@ static int settings_set(const char *name, size_t len, settings_read_cb read_cb, 
 		if (len != sizeof(fota_info))
 			return -EINVAL;
 		if (read_cb(cb_arg, &fota_info, len) > 0)
+			return 0;
+	} else if (!strcmp(name, "auto_connect")) {
+		if (len != sizeof(auto_connect))
+			return -EINVAL;
+		if (read_cb(cb_arg, &auto_connect, len) > 0)
 			return 0;
 	} else if (!strcmp(name, "uart_configured")) {
 		if (len != sizeof(uart_configured))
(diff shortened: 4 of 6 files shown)


junqingzou:
Hi @mgolu, is auto-connect to NIDD network acceptable to your customers?
We need it to complete FOTA over NIDD sequence in time.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -8,11 +8,14 @@
 #include <time.h>
 #include <lwm2m_carrier.h>
 #include <zephyr/logging/log.h>
+#include <modem/nrf_modem_lib.h>
 #include "slm_at_host.h"
 #include "slm_util.h"
 
 LOG_MODULE_REGISTER(slm_carrier, CONFIG_SLM_LOG_LEVEL);
 
+#define REBOOTLESS_FOTA 0
```
mgolu, line 17: Should this be changed to a Kconfig item for customers to be able to build with NIDD support?

mgolu:
> Hi @mgolu, is auto-connect to NIDD network acceptable to your customers? We need it to complete FOTA over NIDD sequence in time.

Overall this will work, since the auto-connect to NIDD is configurable via the new XCARRIER command.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
junqingzou: This is TEMP code for rebootless MFW FOTA debug

junqingzou:
> > Hi @mgolu, is auto-connect to NIDD network acceptable to your customers? We need it to complete FOTA over NIDD sequence in time.
> 
> Overall this will work, since the auto-connect to NIDD is configurable via the new XCARRIER command.

The point is by default auto-connect is eanbled in the setting page, which is behavior change of SLM with NIDD.

applications/serial_lte_modem/src/main.c
```diff
@@ -366,6 +367,46 @@ void handle_mcuboot_swap_ret(void)
...
+	}
+
+	LOG_INF("NIDD auto connect");
+	err = nrf_modem_at_printf("AT%%XSYSTEMMODE=0,1,0,0");
+	if (err) {
+		LOG_ERR("Failed to configure RAT: %d", err);
+		return err;
+	}
+	err = nrf_modem_at_printf("AT+CGDCONT=0,\"Non-ip\"");
+	if (err) {
```
junqingzou, line 391: This step is to be changed, as some NIDD network like Jio in India does requires APN.
So PDP Context configuration will have to be left to PDN init after modem init.

wiba-nordic:
This seems like something that deserves a changelog entry, so remember to add one after this is no longer work in progress.

VTPeltoketo:
acronym NIDD, which words it comes from?

junqingzou:
> acronym NIDD, which words it comes from?

Non-IP Data Delivery (3GPP R13~)

applications/serial_lte_modem/overlay-carrier-nidd.conf
```diff
@@ -0,0 +1,61 @@
...
+
+# Download client for DFU
+CONFIG_DOWNLOAD_CLIENT_MAX_FILENAME_SIZE=230
+
+# Modem info
+CONFIG_MODEM_INFO_BUFFER_SIZE=512
+
+# Disable AT host library as it's incompatible with the carrier library's
+# requirement to initialize the modem library
+CONFIG_AT_HOST_LIBRARY=n
```
kacperradoszewski, line 27: No longer needed, forgot to remove this from the overlay

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -83,6 +83,7 @@ The ``<cmd>`` command is a string, and can be used as follows:
 * ``AT#XCARRIER="error","add|remove",<error>``
 * ``AT#XCARRIER="link_down"``
 * ``AT#XCARRIER="link_up"``
+* ``AT#XCARRIER="auto_connect"[,<auto-connect-flag>]``
```
wiba-nordic, line 86: This should be in alphabetical order like the other commands.

wiba-nordic (changes requested):
I believe this needs a changelog entry for the added AT command.

applications/serial_lte_modem/src/main.c
```diff
@@ -357,6 +358,75 @@ void handle_mcuboot_swap_ret(void)
...
+	int err = 0;
+
+#if defined(CONFIG_LWM2M_CARRIER_SERVER_BINDING_N)
+	int cfun;
+
+	if (!fota_reconnect) {
+		return 0;
+	}
+	err = nrf_modem_at_scanf("AT+CFUN?", "+CFUN: %d", &cfun);
+	if (err != 1 || (cfun == 1 || cfun == 5)) {
```
VTPeltoketo, line 372: Do we support cfun=5?

applications/serial_lte_modem/src/main.c
```diff
@@ -357,6 +358,75 @@ void handle_mcuboot_swap_ret(void)
...
+		return 0;
+	}
+
+	LOG_INF("NIDD auto connect");
```
VTPeltoketo, line 376: These are not done if modem is up?

VTPeltoketo (changes requested):

applications/serial_lte_modem/src/main.c
junqingzou: 5 is "roaming" status in the AT READ result.

applications/serial_lte_modem/src/main.c
junqingzou: Yes if already connected to LTE

applications/serial_lte_modem/src/main.c
```diff
@@ -357,6 +358,75 @@ void handle_mcuboot_swap_ret(void)
...
+	if (!fota_reconnect) {
+		return 0;
+	}
+	err = nrf_modem_at_scanf("AT+CFUN?", "+CFUN: %d", &cfun);
+	if (err != 1 || (cfun == 1 || cfun == 5)) {
+		return 0;
+	}
+
+	LOG_INF("IP auto connect");
+	err = nrf_modem_at_printf("AT+CGDCONT=0,\"IPV4V6\",\"kddiint.lpwa.kddi.com\"");
```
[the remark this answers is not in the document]
junqingzou: Temp code for a modem UICC issue in JIRA, needs IP-based FOTA test with Leshan, will be removed in the future as auot-connect used for NIDD only

applications/serial_lte_modem/src/main.c
```diff
@@ -375,6 +445,7 @@ int start_execute(void)
 	k_work_queue_start(&slm_work_q, slm_wq_stack_area,
 			   K_THREAD_STACK_SIZEOF(slm_wq_stack_area),
 			   SLM_WQ_PRIORITY, NULL);
+	(void)nidd_auto_connect();
```
VTPeltoketo, line 448: There have to be #ifdefs around this function call

applications/serial_lte_modem/src/main.c
```diff
@@ -357,6 +358,75 @@ void handle_mcuboot_swap_ret(void)
...
+		LOG_ERR("Failed to configure RAT: %d", err);
+		return err;
+	}
+	/* NOTE - PDP Context for NIDD needs to be configured per network */
+	err = nrf_modem_at_printf("AT+CGDCONT=0,\"Non-ip\"");
+	if (err) {
+		LOG_ERR("Failed to define NIDD PDN: %d", err);
+		return err;
+	}
+	err = nrf_modem_at_printf("AT%%REDMOB=1");
```
VTPeltoketo, line 388: Two times %-character in the AT-command

applications/serial_lte_modem/src/main.c
```diff
@@ -357,6 +358,75 @@ void handle_mcuboot_swap_ret(void)
...
+		return 0;
+	}
+
+	LOG_INF("IP auto connect");
+	err = nrf_modem_at_printf("AT+CGDCONT=0,\"IPV4V6\",\"kddiint.lpwa.kddi.com\"");
+	if (err) {
+		LOG_ERR("Failed to define PDN: %d", err);
+		return err;
+	}
+	err = nrf_modem_at_printf("AT%%REDMOB=1");
```
VTPeltoketo, line 415: Two times %-character in the AT-command

applications/serial_lte_modem/src/main.c
```diff
@@ -357,6 +358,75 @@ void handle_mcuboot_swap_ret(void)
...
+	if (!fota_reconnect) {
+		return 0;
+	}
+	err = nrf_modem_at_scanf("AT+CFUN?", "+CFUN: %d", &cfun);
+	if (err != 1 || (cfun == 1 || cfun == 5)) {
+		return 0;
+	}
+
+	LOG_INF("NIDD auto connect");
+	err = nrf_modem_at_printf("AT%%XSYSTEMMODE=0,1,0,0");
```
VTPeltoketo, line 377: Two times %-character in the AT-command

applications/serial_lte_modem/src/main.c
```diff
@@ -357,6 +358,75 @@ void handle_mcuboot_swap_ret(void)
...
+	int cfun;
+
+	if (!fota_reconnect) {
+		return 0;
+	}
+	err = nrf_modem_at_scanf("AT+CFUN?", "+CFUN: %d", &cfun);
+	if (err != 1 || (cfun == 1 || cfun == 5)) {
+		return 0;
+	}
+
```
VTPeltoketo, line 408: No SYSTEMMODE settings at all in #else branch. It can be what so ever

applications/serial_lte_modem/src/slm_settings.c
```diff
@@ -144,6 +155,7 @@ void slm_setting_fota_init(void)
 	fota_stage = FOTA_STAGE_INIT;
 	fota_status = FOTA_STATUS_OK;
 	fota_info = 0;
+	fota_reconnect = 1;
```
VTPeltoketo, line 158: Should this be really default on, even if the overlay has not been used?

applications/serial_lte_modem/src/main.c
stig-bjorlykke: @junqingzou You are probably mixing [AT+CFUN](https://infocenter.nordicsemi.com/topic/ref_at_commands/REF/at_commands/mob_termination_ctrl_status/cfun_read.html) and [AT+CEREG](https://infocenter.nordicsemi.com/topic/ref_at_commands/REF/at_commands/nw_service/cereg_read.html).

applications/serial_lte_modem/src/main.c
stig-bjorlykke: Double `%` is needed because `nrf_modem_at_printf()` takes a string fmt argument.

applications/serial_lte_modem/src/main.c
junqingzou: What a shame! 🤦‍♂️

applications/serial_lte_modem/src/main.c
junqingzou: #ifdef is inside this function.

applications/serial_lte_modem/src/slm_settings.c
junqingzou: Yes, this helps to make LwM2M Registration happen automatically in case of NIDD.

kacperradoszewski (commented):
How about extending this functionality to IP context as well? I don't see why this should be restricted to NIDD only, because the binding does not really affect the FOTA sequence and it would be quite convenient to reconnect automatically over UDP as well.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -178,56 +202,77 @@ static void on_event_app_data(const lwm2m_carrier_event_t *event)
...
 		break;
 	case LWM2M_CARRIER_EVENT_FOTA_SUCCESS:
-		LOG_DBG("LWM2M_CARRIER_EVENT_FOTA_SUCCESS");
+		LOG_INF("LWM2M_CARRIER_EVENT_FOTA_SUCCESS");
 		break;
 	case LWM2M_CARRIER_EVENT_REBOOT:
-		LOG_DBG("LWM2M_CARRIER_EVENT_REBOOT");
+		LOG_INF("LWM2M_CARRIER_EVENT_REBOOT");
+#if defined(CONFIG_LWM2M_CARRIER_SERVER_BINDING_N)
+		/* Return 0 to allow the reboot during FOTA scenario, only for NIDD connection. */
```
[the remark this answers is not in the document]
junqingzou: Yes this was lef-over for previous rebootful FOTA testing.

junqingzou:
> How about extending this functionality to IP context as well? I don't see why this should be restricted to NIDD only, because the binding does not really affect the FOTA sequence and it would be quite convenient to reconnect automatically over UDP as well.

Ther reason that we don't want to support IP network is due to:
(1) PDP Context activation normals needs network-specific configuration and use authentication
(2) Don't want to change behavior when SLM is built withour LwM2M_Carrier

junqingzou:
A modem firware fix is verified for a known issue after applying this PR.
The PR carries a workaround in lwm2m_carrier scope due to possible late or no modem fix.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -166,6 +166,10 @@ nRF9160: Serial LTE modem
   * UART transmit has been refactored to utilize buffering.
     Multiple responses can now be received in a single transmission.
 
+* Added:
+
+  * A ``auto_connect`` operation in carrier command ``#XCARRIER``. The operation controls automatic registration of UE to NIDD network.
```
wiba-nordic, line 171:
```suggestion
  * An ``auto_connect`` operation in the ``#XCARRIER`` carrier command.
    The operation controls automatic registration of UE to NIDD network.
```

wiba-nordic (changes requested):

kacperradoszewski:
> > How about extending this functionality to IP context as well? I don't see why this should be restricted to NIDD only, because the binding does not really affect the FOTA sequence and it would be quite convenient to reconnect automatically over UDP as well.
> 
> Ther reason that we don't want to support IP network is due to: (1) PDP Context activation normals needs network-specific configuration and use authentication (2) Don't want to change behavior when SLM is built withour LwM2M_Carrier

If there is need to configure the PDP context, the user can always disable auto-connect and use AT commands to do so as they would do until now. I don't see why SLM would have specific handling for Non-IP binding, which in the LwM2M Carrier library's case is definitely the less common choice. If it's more convenient to set up the context in the code for NIDD, I would simply restrict that section of code to be conditional on CONFIG_LWM2M_CARRIER_SERVER_BINDING_N, but I do find it slightly odd for the SLM to have what seems like application-specific implementation. Setting the functional mode of the modem to 1 (CFUN=1) can apply regardless of the binding, and that can be made dependent on CONFIG_SLM_CARRIER, so this wouldn't affect the behaviour when SLM is built without carrierlib.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -16,6 +16,8 @@ LOG_MODULE_REGISTER(slm_carrier, CONFIG_SLM_LOG_LEVEL);
 
 /**@brief LwM2M Carrier operations. */
 enum slm_carrier_operation {
+	/* Carrier Connect Operation */
+	CARRIER_OP_REQUEST_AUTO_CONNECT,
```
kacperradoszewski, line 20: The `REQUEST` prefix corresponds to the `lwm2m_carrier_request()` API in lwm2m_carrier.h. This operation is not a Carrier library API, so I'd name it simply `CARRIER_OP_AUTO_CONNECT`, and then improve the comment here since there is no documentation for it as opposed to the library's APIs, and it is not intuitive to think that this will only work for NIDD as is now.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -99,11 +103,30 @@ static struct carrier_op_list op_list[CARRIER_OP_MAX] = {
...
+NRF_MODEM_LIB_ON_INIT(lwm2m_init_hook, on_modem_lib_init, NULL);
+
+static void on_modem_lib_init(int ret, void *ctx)
+{
+	ARG_UNUSED(ctx);
+
+	/** ret: Zero on success, a positive value @em nrf_modem_dfu when executing
+	 *  Modem firmware updates, and negative errno on other failures.
+	 */
+	LOG_INF("lib_modem init: %d", ret);
```
kacperradoszewski, line 127: Perhaps this hook could be made generic for the SLM? Don't think it has much to do with the Carrier library per se, the library has its own hook in `lwm2m_carrier.c`, and the modem initialization in SLM takes place in main.

applications/serial_lte_modem/src/main.c
```diff
@@ -333,6 +336,43 @@ static void handle_mcuboot_swap_ret(void)
...
+		LOG_ERR("Failed to configure RAT: %d", err);
+		return err;
+	}
+	/* NOTE - PDP Context for NIDD needs to be configured per network */
+	err = nrf_modem_at_printf("AT+CGDCONT=0,\"Non-ip\"");
+	if (err) {
+		LOG_ERR("Failed to define NIDD PDN: %d", err);
+		return err;
+	}
+	err = nrf_modem_at_printf("AT+CFUN=1");
```
kacperradoszewski, line 366: As per my suggestion, I think CFUN=1 does not need to be restricted to NIDD binding, but maybe just CONFIG_SLM_CARRIER.

junqingzou:
> > > How about extending this functionality to IP context as well? I don't see why this should be restricted to NIDD only, because the binding does not really affect the FOTA sequence and it would be quite convenient to reconnect automatically over UDP as well.
> > 
> > 
> > Ther reason that we don't want to support IP network is due to: (1) PDP Context activation normals needs network-specific configuration and use authentication (2) Don't want to change behavior when SLM is built withour LwM2M_Carrier
> 
> If there is need to configure the PDP context, the user can always disable auto-connect and use AT commands to do so as they would do until now. I don't see why SLM would have specific handling for Non-IP binding, which in the LwM2M Carrier library's case is definitely the less common choice. If it's more convenient to set up the context in the code for NIDD, I would simply restrict that section of code to be conditional on CONFIG_LWM2M_CARRIER_SERVER_BINDING_N, but I do find it slightly odd for the SLM to have what seems like application-specific implementation. Setting the functional mode of the modem to 1 (CFUN=1) can apply regardless of the binding, and that can be made dependent on CONFIG_SLM_CARRIER, so this wouldn't affect the behaviour when SLM is built without carrierlib.

I get your point. Something special with LwM2M_Carrier integration is that you enabled the PDN library in the overlay. This means that customer can add extra CONFIG defined by the library to configure LTE connection, upon libmodem is initialized by SLM main(). Then I'll use CONFIG_SLM_CARRIER to enclose the auto connection logic, which will contain CFUN=1 only.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
junqingzou: yes, good idea as SLM need the log for fault handling as well.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
junqingzou: It's a leftover or copy/paste CARRIER_OP_REQUEST_LINK_UP :)

applications/serial_lte_modem/src/slm_settings.c
```diff
@@ -23,6 +23,8 @@ enum fota_stage fota_stage = FOTA_STAGE_INIT;		/* FOTA: stage of FOTA process */
 enum fota_status fota_status;		/* FOTA: OK/Error status */
 int32_t fota_info;		/* FOTA: failure cause in case of error or download percentage*/
 
+int32_t auto_connect = 1;	/* Auto connect network at startup or after FOTA update */
```
kacperradoszewski, line 26: Not sure whether this should be enabled by default, but I have no preference myself

kacperradoszewski (approved):
Nice and useful improvement, thanks 👍

applications/serial_lte_modem/src/slm_settings.c
junqingzou: I think it'd better be YES for LwM2M_Carrier and NO for other cases

junqingzou: It's safe to be YES here as the real connect logic is enclosed by #if defined(CONFIG_SLM_CARRIER).
There will be no impact on normal SLM build.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -172,6 +172,8 @@ nRF9160: Serial LTE modem
 
   * ``#XMODEMRESET`` AT command to reset the modem while keeping the application running.
     It is expected to be used during modem firmware update, which now only requires a reset of the modem.
+  * An ``auto_connect`` operation in the ``#XCARRIER`` carrier command.
+    The operation controls automatic registration of UE to NIDD network.
```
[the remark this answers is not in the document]
junqingzou: Yes, changed to "LTE network"

applications/serial_lte_modem/src/slm_settings.c
kacperradoszewski: Yeah I suppose I was referring to the Carrier build specifically, as this will change the default behaviour between versions, but yeah no strong opinion from my side, I will probably always have auto-connect enabled :)

nordic-hani (approved):

applications/serial_lte_modem/src/main.c
```diff
@@ -333,6 +348,39 @@ static void handle_mcuboot_swap_ret(void)
...
+	if (!auto_connect) {
+		return 0;
+	}
+	err = nrf_modem_at_scanf("AT+CEREG?", "+CEREG: %d", &stat);
+	if (err != 1 || (stat == 1 || stat == 5)) {
+		return 0;
+	}
+
+	LOG_INF("auto connect");
+#if defined(CONFIG_LWM2M_CARRIER_SERVER_BINDING_N)
```
VTPeltoketo, line 367: Why the SYSTEMMODE setting has to be under CONFIG_LWM2M_CARRIER_SERVER_BINDING_N?

junqingzou: > Why the SYSTEMMODE setting has to be under CONFIG_LWM2M_CARRIER_SERVER_BINDING_N?

Because modem only support NB-IoT for NIDD connection.

junqingzou:
== DNM == 
this PR contains workaround code for modem failure, which will be fixed in libmodem instead.
Will test once the solution in libmodem is released to Main branch.

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -107,6 +108,16 @@ The response syntax depends on the commands used.
 Examples
 ~~~~~~~~
 
+::
+
+   AT#XCARRIER="auto_connect"
```
VTPeltoketo, line 113: Why auto_connect does not have "read" and "write" actions when all other writable parameters have those?

junqingzou: No special reason.
 AT#XCARRIER="auto_connect"  ......... READ
 AT#XCARRIER="auto_connect",<value>  ..... WRITE then READ

VTPeltoketo: I suggest that the format is constant over all parameters

VTPeltoketo (approved):

PR #10849, opened 2023-04-14
matter: samples: alignment in configurations tags

jnsgsr:
- no duplicates for plain debug configuration
- required more complex quarantine file to limit builds in CI
- new quarantine file prepared to be enabled in CI

test-sdk-nrf: PR-515

### DIFF
--- a/samples/matter/light_bulb/sample.yaml
+++ b/samples/matter/light_bulb/sample.yaml
@@ -3,13 +3,6 @@ sample:
   name: Matter Light Bulb
 tests:
   # Excluded in quarantine.yaml to limit resources usage in integration builds
-  sample.matter.light_bulb.debug:
-    build_only: true
-    integration_platforms:
-      - nrf52840dk_nrf52840
-      - nrf5340dk_nrf5340_cpuapp
-      - nrf7002dk_nrf5340_cpuapp
-    platform_allow: nrf52840dk_nrf52840 nrf5340dk_nrf5340_cpuapp nrf7002dk_nrf5340_cpuapp
   sample.matter.light_bulb.release:
     build_only: true
     extra_args: CONF_FILE=prj_release.conf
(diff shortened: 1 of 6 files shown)


samples/matter/light_bulb/sample.yaml
```diff
@@ -39,12 +32,12 @@ tests:
       - nrf21540dk_nrf52840
     platform_allow: nrf52840dk_nrf52840 nrf21540dk_nrf52840 nrf5340dk_nrf5340_cpuapp nrf7002dk_nrf5340_cpuapp
   # ---------------
-  sample.matter.light_bulb.debug.nrf5340dk:
+  sample.matter.light_bulb.debug:
     build_only: true
     integration_platforms:
       - nrf5340dk_nrf5340_cpuapp
     platform_allow: nrf52840dk_nrf52840 nrf5340dk_nrf5340_cpuapp nrf7002dk_nrf5340_cpuapp
-  sample.matter.light_bulb.debug.nrf7002_ek:
```
doublemis1, line 47: why debug is removed while the device is build as debug configuration

doublemis1: if no `.debug` in sample name means that device is in debug configuration, then it need to be align across all sample configuration definitions

jnsgsr: I think with those changes it will be consistent across all samples.yaml, that "debug" in name will be only if there are no additional parameters changed from default set.

samples/matter/template/sample.yaml
```diff
@@ -24,13 +18,9 @@ tests:
       - nrf21540dk_nrf52840
     platform_allow: nrf52840dk_nrf52840 nrf21540dk_nrf52840 nrf5340dk_nrf5340_cpuapp nrf7002dk_nrf5340_cpuapp
   # ---------------
-  sample.matter.template.debug.nrf5340dk:
+  sample.matter.template.debug:
```
kkasperczyk-no, line 21: Why don't we have nrf52840dk_nrf52840 in the intergration_platforms section here and in every other sample?

jnsgsr: Integration with nrf52840dk_nrf52840 is covered by sample.matter.lock.smp_dfu_ffs it is also taken for system tests later, maybe we need to include more configurations with nrf52840dk_nrf52840 but better to start discussion on this with this document  https://projecttools.nordicsemi.no/confluence/display/KRKNWK/Configurations+Selection+for+Integration+Builds

LuDuda (approved):

PR #10851, opened 2023-04-14
samples: aes_gcm: Move accelerator selection to boards/*.conf

SebastianBoe:
Move selection of the appropriate HW accelerator for the platform from the generic prj.conf to the board-specific boards/*.conf files.

Applied to all boards in sample.yaml.

Not specifying the accelerator in prj.conf makes the sample more portable.


frkv (approved):
LGTM (test errors seems unrelated)

SebastianBoe:
Some work that is coming today will conflict with this PR so we drop this PR. The upcoming PR is more complete than this one.

Closing after I get a green CI run as I am interested in the results.

SebastianBoe:
This is deprecated by another PR

PR #10858, opened 2023-04-17
samples: nrf9160: nrf_cloud_rest_fota: Disable SPI for mcuboot

nordicjm:
Disables SPI and SPI NOR being included in the mcuboot build, which gives a build error due to the lack of multithreading. Neither are required for this, so explicitly disable them for the mcuboot build.

### DIFF
--- a/samples/nrf9160/nrf_cloud_rest_fota/child_image/mcuboot.conf
+++ b/samples/nrf9160/nrf_cloud_rest_fota/child_image/mcuboot.conf
@@ -1,0 +1,9 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Disable SPI and SPI NOR as the bootloader needs neither
+CONFIG_SPI=n
+CONFIG_SPI_NOR=n


jayteemo (approved):

plskeggs:
Isn't SPI needed when the full modem FOTA  overlay is used, due to needing external flash?  In that case, do we enable multi threading for mcuboot (responding on my phone)?

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

nordicjm:
> Isn't SPI needed when the full modem FOTA overlay is used, due to needing external flash? In that case, do we enable multi threading for mcuboot?

I had a look at the PM config and it did not have the option for using external flash as the secondary slot enabled, I don't have an nrf9160dk to test this on however.

nordicjm:
From the generated partitions.yml:
```
fmfu_storage:
  address: 0x0
  align: 0x1000
  device: DT_CHOSEN(nordic_pm_ext_flash)
  end_address: 0x400000
  placement:
    before:
    - end
  region: external_flash
  size: 0x400000
mcuboot_secondary:
  address: 0x90000
  end_address: 0xf8000
  placement:
    after:
    - mcuboot_primary
    align:
      start: 0x8000
  region: flash_primary
  share_size:
  - mcuboot_primary
  size: 0x68000
```
So this looks fine to me, secondary slot is in internal flash, no SPI access should be needed for mcuboot (fmfu is an application thing, correct)?

plskeggs:
> > Isn't SPI needed when the full modem FOTA overlay is used, due to needing external flash? In that case, do we enable multi threading for mcuboot?
> 
> I had a look at the PM config and it did not have the option for using external flash as the secondary slot enabled, I don't have an nrf9160dk to test this on however.

This is a problem.  Do not change nrf9160dk samples if you do not have a way to test your change.

We have an overlay file in this project you are modifying which enables using external flash for the secondary MCUboot partition -- it is called `overlay_full_modem_fota.conf`.  You must specify this overlay file when building the sample in order to see the change in the PM config.

See: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/nrf9160/nrf_cloud_rest_fota/README.html#building-and-running

nordicjm:
@SeppoTakalo Can you explain this system, does it use mcuboot to update or does it have nothing to do with the mcuboot image itself and only uses mcuboot slots but is actually fully application driven? And if does use mcuboot, why does it appear that using the `overlay_full_modem_fota.conf` overlay doesn't seem to enable any (that I can see) modem DFU Kconfigs in the mcuboot image?

SeppoTakalo:
Full modem update is completely application driven process.
It has nothing to do with MCUboot, so I think this change is correct. MCUBoot can start up without SPI and then application may use SPI, if it chooses to.

nordicjm:
Another commit is needed (will do it in a separate PR) because the Kconfig for this specifically mentions it is dealt with by mcuboot which is incorrect

nordicjm:
@plskeggs can you re-review?

plskeggs (approved):

plskeggs:
@nordicjm -- this needs to get merged asap -- looks like you need to update the changelist first

nordicjm:
Changelog entry not needed as nothing is changed behaviour wise from previous releases

PR #10865, opened 2023-04-17
samples: crypto: Move crypto accelerator selection into board configs

mswarowsky:
Moved the selection of which crypto accelerator is used into the board files.


SebastianBoe (approved):
I'm not going to check the configs, I'll trust no mistakes were made...

vili-nordic (approved):

PR #10866, opened 2023-04-17
samples: nrf9160: https_client: remove SOCK_NATIVE_TLS

richardmccrae-nordicsemi:
Remove logical ORing of SOCK_NATIVE_TLS from socket() call when CONFIG_SAMPLE_TFM_MBEDTLS=y

This caused the connect() call to fail with 22 (EINVAL) since the merging of sdk-nrf/commit/6e303d8c

### DIFF
--- a/samples/nrf9160/https_client/src/main.c
+++ b/samples/nrf9160/https_client/src/main.c
@@ -227,11 +227,7 @@ void main(void)
 		  INET6_ADDRSTRLEN);
 	printk("Resolved %s (%s)\n", peer_addr, net_family2str(res->ai_family));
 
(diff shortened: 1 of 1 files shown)


rlubos:
What's the point of having native TLS overlay then if the sample will not use native TLS socket anyway?

I'd rather look for the rootcause of the failure, or remove the overlay whatsoever. Perhaps try renabling the configs at the overlay level, that were disabled in the linked commit (RSA)?
https://github.com/nrfconnect/sdk-nrfxlib/pull/949/files#diff-ffcd4016aa46528f0b7beb62756627aeb3b9d5f9ea6f4db892f25bb95609f3d0

PR #10867, opened 2023-04-17
Bluetooth: controller: Allow CONFIG_BT_MAX_CONN=1

jori-nordic:
https://github.com/nrfconnect/sdk-nrf/pull/10799 was a breaking change. This attempts to keep the original defaults, while still allowing the special case where support for both roles is compiled in but with only one host conn context.

This PR reverts the previous attempt and applies a different workaround.

### DIFF
--- a/subsys/bluetooth/Kconfig
+++ b/subsys/bluetooth/Kconfig
@@ -13,7 +13,6 @@ if BT
 # When both connection roles are enabled there has to be one for each role.
 config BT_MAX_CONN
 	int
-	range 2 20 if BT_LL_SOFTDEVICE && BT_CENTRAL && BT_PERIPHERAL
 	range 1 20 if BT_LL_SOFTDEVICE
 	default 2 if BT_LL_SOFTDEVICE && BT_CENTRAL && BT_PERIPHERAL
 
--- a/subsys/bluetooth/Kconfig
+++ b/subsys/bluetooth/Kconfig
@@ -8,6 +8,15 @@ menu "Bluetooth Low Energy"
 
 if BT
 
+# BT_MAX_CONN is declared in Zephyr, here we add the range and default for
+# BT_LL_SOFTDEVICE which is tested with 20 connections.
+# When both connection roles are enabled there has to be one for each role.
+config BT_MAX_CONN
+	int
+	range 2 20 if BT_LL_SOFTDEVICE && BT_CENTRAL && BT_PERIPHERAL
+	range 1 20 if BT_LL_SOFTDEVICE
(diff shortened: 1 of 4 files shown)


jori-nordic:
failing zigbee job is a known failure

Tschet1 (approved):

alwa-nordic:
Could we have three independent variables `CONFIG_BT_MAX_CONN`, `CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT`, and (new?) `CONFIG_BT_CTLR_SDC_CENTRAL_COUNT`. It seems simpler than a special case for `CONFIG_BT_MAX_CONN=1`.

alwa-nordic:
This looks like a removal of a feature. Before this change it would be possible to have `CONFIG_BT_MAX_CONN = CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT = CONFIG_BT_CTLR_SDC_CENTRAL_COUNT` for when you know the maximum number of connections but not which role they will have. Is it not better to adjust the kconfig defaults instead?

jori-nordic:
> This looks like a removal of a feature. Before this change it would be possible to have `CONFIG_BT_MAX_CONN = CONFIG_BT_CTLR_SDC_PERIPHERAL_COUNT = CONFIG_BT_CTLR_SDC_CENTRAL_COUNT` for when you know the maximum number of connections but not which role they will have. Is it not better to adjust the kconfig defaults instead?

Yeah I thought it'd be a nice feature, but:
- no one asked for it (yet)
- it broke a lot of apps, as it changed the defaults
- kconfig is not very flexible (e.g. can't do math without custom functions)
- I only need the `BT_MAX_CONN=1` w/ both `BT_CENTRAL=y` and `BT_PERIPHERAL=y` special case so it's build-compatible with the upstream controller (and bsim tests)

alwa-nordic:
> * I only need the `BT_MAX_CONN=1` w/ both `BT_CENTRAL=y` and `BT_PERIPHERAL=y` special case so it's build-compatible with the upstream controller (and bsim tests)

Then we only need to remove the cmake warning, right? Other than that, it will work as expected, I think?

jori-nordic:
> > * I only need the `BT_MAX_CONN=1` w/ both `BT_CENTRAL=y` and `BT_PERIPHERAL=y` special case so it's build-compatible with the upstream controller (and bsim tests)
> 
> Then we only need to remove the cmake warning, right? Other than that, it will work as expected, I think?

It does not. Please re-read the pull request message, and see that there are two commits, one reverting a breaking change, and another one allowing this special case.

PR #10880, opened 2023-04-18
quarantine_windows: Add EBC tests

rawoNCS:
Add EBC tests to quarantine.
Rename file from quarantine_windows to quarantine_windows_mac as it is common for both platforms.

Signed-off-by: Rafal Wozniakowski <rafal.wozniakowski@nordicsemi.no>

### DIFF
--- a/scripts/quarantine_windows_mac.yaml
+++ b/scripts/quarantine_windows_mac.yaml
@@ -7,20 +7,7 @@
 #  platforms:
 #    - None
 
-- scenarios:
-    - applications.asset_tracker_v2.aws
-    - applications.asset_tracker_v2.azure
-    - applications.asset_tracker_v2.cloud.cloud_codec.json_common.azure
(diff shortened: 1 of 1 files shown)


gopiotr (changes requested):
Since last Zephyr upmerge there is possiblity to use in quarantine regexes and "platforms under quarantine". So, in this PR you can also change the way for excluding tests which are dedicated for native_posix and QEMU. For example like this (without listing all testscenarios):
``` yaml
- platforms:
    - native_posix
    - qemu_cortex_m3
    - qemu_x86
  comment: "Cannot build samples for Native POSIX and QEMU on Windows OS"
```
More information about this you can find here:
https://docs.zephyrproject.org/latest/develop/test/twister.html#quarantine

gopiotr (approved):

miha-nordic (approved):

PR #10442, opened 2023-02-24
net: lib: download_client: Non blocking connect 

SeppoTakalo:
### Refactor API for non-blocking support
    
* Don't block on download_client_connect() or download_client_disconnect()
* Refactor download_client_connect() to download_client_set_host()
  so the name indicates that it does not yet initiate anything.
* Add function download_client_get() that combines set_host,
  start and disconnect.
* Do all name queries and socket connections in the background
  thread.
* Add new event `DOWNLOAD_CLIENT_EVT_CLOSED` to indicate that background thread has closed the connection and new download can be started.
    
Now there are two options to use this API:
1. set_host() and call start() for multiple files,
   and finally call disconnect().
2. Call get().
    
On both cases, init() should only be called once.
* Both cycles can be repeated.
* Both use cases can be stopped/aborted by disconnect().

### Refactor the handler loop 

* Add mutexes to protect internal state
* Refactor the handling loop so there is no "goto". This should lead to more maintainable code. I had hard time understanding the request-loop and its many corner cases. It should be clearer now.
* Various minor fixes

### Remove pause and resume
    
Pause and resume are non thread safe operations and
there are no use for them as we can just disconnect
and resume with offset.

### Various fixes

* Return -EALREADY if connect called twice
* download_client_disconnect() should be called before connecting again.
* Fix unranged HTTP GET
* Fix CoAP download ending
   For GET requests where we don't know the full size, we must
   use the last_block marker and end the transfer at that point.

PR #10882, opened 2023-04-18
net: lib: download_client: Add support for sec_tag list

stig-bjorlykke:
Add support for giving a list of sec tags to the download client. This will improve the download logic when having multiple sec tags, and will reduce the number of download attempts on CA mismatch.

### DIFF
--- a/applications/serial_lte_modem/src/dfu/slm_at_dfu.c
+++ b/applications/serial_lte_modem/src/dfu/slm_at_dfu.c
@@ -64,6 +64,7 @@ static uint32_t crc_32;
 static char hostname[URI_HOST_MAX];
 static char filepath[SLM_MAX_URL];
 static struct download_client dlc;
+static int sec_tag_list[1];
 static int socket_retries_left;
 static bool first_fragment;
 static uint16_t mtu;
@@ -308,13 +309,18 @@ static  int dfu_download_start(const char *host, const char *file, int sec_tag,
 	int err = -1;
 
 	struct download_client_cfg config = {
-		.sec_tag = sec_tag,
 		.frag_size_override = fragment_size,
 		.set_tls_hostname = (sec_tag != -1),
 	};
 
 	if (host == NULL || file == NULL) {
 		return -EINVAL;
+	}
+
+	if (sec_tag != -1) {
+		sec_tag_list[0] = sec_tag;
+		config.sec_tag_list = sec_tag_list;
+		config.sec_tag_count = 1;
 	}
 
 	socket_retries_left = CONFIG_SLM_DFU_SOCKET_RETRIES;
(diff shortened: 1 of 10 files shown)


applications/serial_lte_modem/src/dfu/slm_at_dfu.c
```diff
@@ -307,8 +307,10 @@ static  int dfu_download_start(const char *host, const char *file, int sec_tag,
...
 	int err = -1;
 
+	int sec_tag_list[] = { sec_tag };
 	struct download_client_cfg config = {
-		.sec_tag = sec_tag,
+		.sec_tag_list = sec_tag_list,
+		.sec_tag_count = (sec_tag != -1) ? 1 : 0,
 		.frag_size_override = fragment_size,
 		.set_tls_hostname = (sec_tag != -1),
 	};
```
SeppoTakalo, line 316: Wov.. there seem to be a stack / object lifetime issue.

`struct download_client_cfg` is allocated from stack, but the Download client may refer to it from its internal thread.

While you are here, please move the configuration structure outside from stack allocation.

SeppoTakalo: Sorry, my bad.. the configuration structure seem to be copied. But the `sec_tag_list` must then have a lifetime as long as the download client is running. So at least the list needs to be global static.

lib/bin/lwm2m_carrier/os/lwm2m_os.c
```diff
@@ -507,8 +507,10 @@ static lwm2m_os_download_callback_t lwm2m_os_lib_callback;
 
 int lwm2m_os_download_connect(const char *host, const struct lwm2m_os_download_cfg *cfg)
 {
+	int sec_tag_list[] = { cfg->sec_tag };
```
SeppoTakalo, line 510: Same lifetime issue here.

SeppoTakalo (changes requested):
Like commented of few places.

sec_tag_list cannot come from stack. It has a lifetime longer than call to `download_client_connect()`

applications/serial_lte_modem/src/dfu/slm_at_dfu.c
stig-bjorlykke: Lifetime issue should be fixed.

SeppoTakalo (approved):

SeppoTakalo (approved):

umapraseeda (approved):

junqingzou (approved):

MarkusLassila (approved):

plskeggs (approved):

SeppoTakalo:
We most probably end up in conflict when this PR https://github.com/nrfconnect/sdk-nrf/pull/10442 is going to be merged. Or other way around.

Lets see which one goes first. Both seem slow because there is no trust in CI now.

SeppoTakalo:
> I'm waiting for someone to fix failures unrelated to this changes. When all builds are green it will be merged. Any conflicts should be easy to fix.

See the internal discussion Cellular IOT Testing channel. This is going to take some time, as it falls into infrastructure issues. I think we need to push maintainers to ignore those CI jobs.

rlubos (approved):

PR #10885, opened 2023-04-18
samples: radio_test, dtm: Clean up erratas

dchat-nordic:
Clean up erratas used in radio_test and dtm.
Use mdk functions for checking application of erratas.

### DIFF
--- a/samples/bluetooth/direct_test_mode/src/dtm.c
+++ b/samples/bluetooth/direct_test_mode/src/dtm.c
@@ -67,7 +67,7 @@
 						 _irq_handler)
 /* Note that the timer instance 2 is used in the FEM driver. */
 
-#ifdef NRF52840_XXAA
+#if NRF52_ERRATA_172_PRESENT
 /* Timer used for the workaround for errata 172 on affected nRF5 devices. */
 #define ANOMALY_172_TIMER_INSTANCE     3
 #define ANOMALY_172_TIMER_IRQ          NRFX_CONCAT_3(TIMER,		    \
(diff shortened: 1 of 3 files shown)


peknis (approved):

samples/peripheral/radio_test/src/radio_test.c
```diff
@@ -422,15 +422,18 @@ static void errata_117(nrf_radio_mode_t mode)
...
+static void errata_117(nrf_radio_mode_t mode)
+{
+	ARG_UNUSED(mode);
+}
+#endif /* NRF53_ERRATA_117_PRESENT */
 
 static void radio_mode_set(NRF_RADIO_Type *reg, nrf_radio_mode_t mode)
 {
-#if NRF53_ERRATA_117_ENABLE_WORKAROUND
 	if (nrf53_errata_117()) {
```
KAGA164, line 434: Why does this errata have different schema of handling it than others? In other erratas you check inside the handling function.

dchat-nordic: Corrected

samples/bluetooth/direct_test_mode/src/dtm.c
```diff
@@ -1132,7 +1110,58 @@ static void anomaly_172_strict_mode_set(bool enable)
...
+		anomaly_172_strict_mode_set(false);
+		nrfx_timer_disable(&dtm_inst.anomaly_timer);
+		dtm_inst.anomaly_172_wa_enabled = false;
+	}
+}
+#else
+static void errata_172_handle(bool enable)
+{
+	ARG_UNUSED(enable);
+}
```
maje-emb, line 1134: Is it necessary to create an empty function?
In the `errata_172_handle` function there is a condition `if (!nrf52_errata_172())`, which seems to be sufficient.

dchat-nordic: It is necessary, because there are references to anomaly_timer, which is not present in for example nRF5340 compiled code. This would cause build errors and warnings.

maje-emb (approved):

KAGA164 (approved):

PR #10891, opened 2023-04-18
lib: nrf_modem_lib: rewrite UART trace backend to use zephyr driver

eivindj-nordic:
* Use zephyr driver for the UART trace backend. This enables the
backend to be configured through the device tree, and the application
can choose which UART instance to use.
* On fault, modem reinitialization depends on the trace thread to complete
the processing of traces. In this case we want to prevent processing
traces forever. Therefore we abort if the trace backend write returns
zero when the modem is not initialized.
* Add overlays for UART tracing to samples and applications, to simplify
enabling modem traces with the UART trace backend.


applications/serial_lte_modem/prj.conf
```diff
@@ -109,3 +109,8 @@ CONFIG_SLM_LOCATION=n
 # Serial DFU support
 CONFIG_SLM_NRF52_DFU=n
 CONFIG_SLM_NRF52_DFU_LEGACY=n
+
+#CONFIG_NRF_MODEM_LIB_TRACE=y
+CONFIG_LOG_BACKEND_RTT=y
+CONFIG_LOG_BACKEND_UART=n
+#CONFIG_NRF_MODEM_LIB_TRACE_BACKEND_UART=y
```
mgolu, line 116: I'm not sure that these 5 lines should be added. The `CONFIG_LOG_BACKEND` settings are already on line 18 and 19, and the `LIB_TRACE` related ones are in the overlay.

applications/serial_lte_modem/overlay-modem-trace-uart.conf
mgolu, line 1: With this overlay, lines 30 and 31 in `prj.conf` should be removed. Those lines are:

`# Enable below for modem trace`
`#CONFIG_NRF_MODEM_LIB_TRACE=y`

applications/serial_lte_modem/prj.conf
eivindj-nordic: These were added when testing and should not be here :)

MarkusLassila:
@eivindj-nordic: First of all, thanks for making this change. For SLM, This will allow the RTT-logs to be collected simultaneously to  modem traces. Previously that was not possible as modem traces were heavily corrupted with any amount of traffic.

As we discussed, enabling the traces is bit clumsy in nRF Connect for VS Code and I (relatively new employee) could not get that to work out-of-the-box. The problem was that while adding the Kconfig is straightforward, setting the device tree overlay file required Extra CMake argument: -DDTC_OVERLAY_FILE="modem-trace-uart.overlay;boards/nrf9160dk_nrf9160_ns.overlay".

I would have needed documentation on how to set this or better UI support from nRF Connect for VS Code.

applications/serial_lte_modem/modem-trace-uart.overlay
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/* Enable uart1 for tracing. */
+&uart1 {
+	status = "okay";
+	current-speed = < 1000000 >;
```
MarkusLassila, line 10: Should this have hw-flow-control; as well? RTT-logging backend is used by default and that does not work with this without hw-flow-control.

lib/nrf_modem_lib/trace_backends/uart/uart.c
```diff
@@ -1,141 +1,159 @@
...
+	err = uart_config_get(uart_dev, &uart_cfg);
+	if (err) {
+		LOG_ERR("UART_config_get failed: %d", err);
+		return err;
+	}
 
-	IRQ_CONNECT(DT_IRQN(UART1_NL), DT_IRQ(UART1_NL, priority), nrfx_isr,
-		    &nrfx_uarte_1_irq_handler, UNUSED_FLAGS);
+	if (uart_cfg.baudrate < TRACE_UART_BAUD_RECOMMENDED) {
+		LOG_WRN("Trace UART baudrate is lower than recommended (%d<%d)",
```
MarkusLassila, line 75: I think that lower baudrates do not work with Fidoless-tracing using trace collector. Should this be mentioned in the warning?

lib/nrf_modem_lib/trace_backends/uart/uart.c
```diff
@@ -1,141 +1,159 @@
...
 		size_t idx = len - remaining_bytes;
 
-		if (k_sem_take(&tx_sem, K_MSEC(UART_TX_WAIT_TIME_MS)) != 0) {
-			LOG_WRN("UARTE TX not available!");
-			break;
+		for (int i = 0; i < UART_TX_RETRIES; i++) {
+			err = uart_tx(uart_dev, &buf[idx], transfer_len,
+			      UART_TX_WAIT_TIME_MS * USEC_PER_MSEC);
+			if (err) {
+				LOG_ERR("uart error: %d", err);
```
MarkusLassila, line 122: Should we release the semaphore here? Depending on the error, the event_handler will not be called.

eivindj-nordic: Definitely!

eivindj-nordic:
Updated to keep the existing UART backends for now, for backward compability and avoid breaking changes. The new backend is selected by `CONFIG_NRF_MODEM_LIB_TRACE_BACKEND_UART_NEW`. The already existing backends will be removed at a later time.

lib/nrf_modem_lib/trace_backends/uart/uart.c
```diff
@@ -1,142 +1,172 @@
...
-#endif /* CONFIG_PM_DEVICE */
+	for (int i = 0; i < UART_TX_RETRIES; i++) {
+		err = uart_tx(uart_dev, data, len,
+			UART_TX_WAIT_TIME_MS * USEC_PER_MSEC);
+		if (err) {
+			LOG_ERR("uart error: %d", err);
+			return err;
+		}
+
+		err = k_sem_take(&tx_sem, K_FOREVER);
```
tokangas, line 113: With infinite timeout, I don't think `k_sem_take()` returns unless `k_sem_reset()` is called or the semaphore is successfully taken. I don't see `k_sem_reset()` being called, so the return value check is probably not needed?

lib/nrf_modem_lib/trace_backends/uart/uart.c
```diff
@@ -1,142 +1,172 @@
...
-	nrfx_err_t err;
+	int err;
+	int ret;
 
 	/* Split RAM buffer into smaller chunks to be transferred using DMA. */
 	uint8_t *buf = (uint8_t *)data;
 	const size_t MAX_BUF_LEN = (1 << UARTE1_EASYDMA_MAXCNT_SIZE) - 1;
 	size_t remaining_bytes = len;
 
+	err = k_sem_take(&tx_sem, K_FOREVER);
```
tokangas, line 137: Same here, return value check is probably not needed.

lib/nrf_modem_lib/trace_backends/uart/Kconfig
```diff
@@ -8,18 +8,36 @@
...
+	  Enable synchronous processing for UARTE1 trace backend.
+
 config NRF_MODEM_LIB_TRACE_BACKEND_UART_SYNC
 	bool "UARTE1 SYNC"
+	select NRFX_UARTE1
+	select DEPRECATED
 	help
 	  Enable synchronous processing for UARTE1 trace backend.
 
+config NRF_MODEM_LIB_TRACE_BACKEND_UART_NEW
```
tokangas, line 32: Could the new implementation be called something else than "new"? How about for example `NRF_MODEM_LIB_TRACE_BACKEND_UART_ZEPHYR`, that would be a bit more descriptive? And if the old versions are deprecated, could this one be enabled by default?

lib/nrf_modem_lib/trace_backends/uart/uart.c
```diff
@@ -1,142 +1,172 @@
...
+	switch (evt->type) {
+	case UART_TX_DONE:
+		tx_succeeded = true;
 		k_sem_give(&tx_sem);
-	}
-
-	if (event->type == NRFX_UARTE_EVT_TX_DONE) {
+		break;
+	case UART_TX_ABORTED:
+		tx_succeeded = false;
```
tokangas, line 43: Looking at the documentation, `UART_TX_ABORTED` doesn't necessarily mean that nothing was sent:

`When flow control is enabled, there is a possibility that TX transfer won’t finish in the allotted time. Some data may have been transferred, information about it can be found in event data.`

In this case, the current implementation would try to send the whole chunk again, so part of the data would be sent twice.

lib/nrf_modem_lib/trace_backends/uart/uart.c
```diff
@@ -1,142 +1,172 @@
...
+		return -EAGAIN;
+	}
+
 	while (remaining_bytes) {
 		size_t transfer_len = MIN(remaining_bytes, MAX_BUF_LEN);
 		size_t idx = len - remaining_bytes;
 
-		if (k_sem_take(&tx_sem, K_MSEC(UART_TX_WAIT_TIME_MS)) != 0) {
-			LOG_WRN("UARTE TX not available!");
+		ret = uart_send(&buf[idx], transfer_len);
```
tokangas, line 147: I'm not that familiar with how the modem tracing is implemented, but I think we should probably try to avoid corrupting data when sending it to the UART. If all traces can not be output, the modem will start dropping traces. The good thing about this is, that there's trace which indicates how much traces have been lost. If corruption/dropping of the traces happens at this level, we have no idea how much has been lost.

Because of this, I think `uart_send()` should always return the number of bytes it was able to send and that number should be given to `trace_processed_callback()`. Currently, when flow control activates, `uart_send()` returns zero although part of the data may have been sent. Also, `trace_processed_callback()` would not get called.

lib/nrf_modem_lib/trace_backends/uart/Kconfig
eivindj-nordic: I can rename it to `_ZEPHYR`.
Regarding what is the default, the idea to keep the existing one as the default for now to avoid breaking changes. So that existing documentation and guides, that does not contain information about Device tree changes etc., will still work, but the user will get a deprecation warning.

lib/nrf_modem_lib/trace_backends/uart/uart.c
eivindj-nordic: Thank you for pointing that out. I'll update this :)

lib/nrf_modem_lib/trace_backends/uart/Kconfig
tokangas: Does `choice` always have a value? Currently, none of the options is marked as default. If `choice` always has a value, probably `ASYNC` gets implicitly selected because it's the first one. I think it would be good to mark it explicitly as the default, if that's the intended behavior.

eivindj-nordic: That is correct, I'll mark it as default.

lib/nrf_modem_lib/trace_backends/uart/uart.c
MarkusLassila: Setting a different baud rate would likely work with FIDO-tracing, so I am a bit unsure what approach should be taken here.

Can someone offer an opinion on this?

lib/nrf_modem_lib/trace_backends/uart/uart.c
```diff
@@ -1,142 +1,161 @@
...
-	nrfx_err_t err;
+	int err;
+	int ret;
 
 	/* Split RAM buffer into smaller chunks to be transferred using DMA. */
 	uint8_t *buf = (uint8_t *)data;
 	const size_t MAX_BUF_LEN = (1 << UARTE1_EASYDMA_MAXCNT_SIZE) - 1;
 	size_t remaining_bytes = len;
 
+	k_sem_take(&tx_sem, K_FOREVER);
```
MarkusLassila, line 130: Can this method be called from multiple threads? If it can, then there is a race condition between thread 1, which has taken this semaphore and is waiting in line 110 and thread 2, which is waiting to take this semaphore. If thread 2 takes the semaphore, then tx_bytes will be changed before thread 1, can take this.

eivindj-nordic: It should only be called by the trace thread. However, it can in theory be called from elsewhere, so I can rewrite this to avoid the potential issue.

lib/nrf_modem_lib/trace_backends/uart/uart.c
eivindj-nordic: That is unrelated to the trace backends and modem tracing in NCS.

eivindj-nordic:
Introducing [snippets](https://docs.zephyrproject.org/latest/build/snippets/index.html) to avoid Device tree and Kconfig overlays in all nRF91 samples for modem tracing with UART. This makes the PR dependent on the ongoing Zephyr upmerge (#10876), that is bringing snippets from upstream. A small update to sdk-zephyr is also needed to make snippets visible in the ${NRF_DIR}/snippets directory.

lib/nrf_modem_lib/trace_backends/uart/uart.c
MarkusLassila: Check. Had used FIDO tracing exactly once, and had only recollection that the baudrate could be controlled.

MarkusLassila (approved):
Looks good from SLM perspective.

samples/nrf9160/https_client/modem-trace-uart.overlay
```diff
@@ -0,0 +1,18 @@
...
+	status = "okay";
+	current-speed = < 1000000 >;
+	hw-flow-control;
+};
+
+/ {
+	chosen {
+		nordic,modem-trace-uart = &uart1;
+	};
+};
```
SeppoTakalo, line 18: Why is this file copied, instead of instructed to use the same snippet?

doc/nrf/libraries/modem/nrf_modem_lib.rst
```diff
@@ -121,6 +121,80 @@ If the difference in the values of the :kconfig:option:`CONFIG_NRF_MODEM_LIB_TRA
 
 To enable logging of the modem trace bitrate, enable the :kconfig:option:`CONFIG_NRF_MODEM_LIB_TRACE_BITRATE_LOG` Kconfig.
 
+.. _modem_trace_backend_uart_nrf9160dk:
+
+Sending traces over UART on the nRF9160DK
+=========================================
+
+When sending modem traces over UART1 on the nRF9160DK, configuration must be added for the UART device in the devicetree.
+This is done by adding the following code snippet to the board devicetree or overlay file.
```
SeppoTakalo, line 130: Is this `.rst` file inherited from modem library, or is there some other reason why the `west build -S` snippet trick is not mentioned here.

Can we add it here as well? Someone might end up reading this page, without knowledge of the added snippet.

SeppoTakalo (approved):

SeppoTakalo (approved):

doc/nrf/snippets/nrf91_modem_trace_uart.rst
```diff
@@ -0,0 +1,28 @@
+.. _nrf91_modem_trace_uart_snippet:
+
+nRF91 modem tracing with UART backend
+#####################################
+
+The snippet is tailored for tracing on the nRF91-DKs but can work with other applications as well.
```
MirkoCovizzi, line 6: This sentence is a bit confusing, do you mean applications or boards?

eivindj-nordic: boards*

simensrostad (approved):
ATv2 changes look OK. Nothing more to add other than what has been discussed in earlier comments.
More docs in apps/samples/Vscode extension on how to enable traces is definitely important when the old backend is eventually removed.

lib/nrf_modem_lib/trace_backends/uart/Kconfig
```diff
@@ -8,18 +8,37 @@
...
+choice NRF_MODEM_LIB_TRACE_BACKEND_UART_VERSION
+	prompt "UART trace backend version"
+	default NRF_MODEM_LIB_TRACE_BACKEND_UART_ASYNC
+
+config NRF_MODEM_LIB_TRACE_BACKEND_UART_ASYNC
+	bool "UARTE1 ASYNC"
+	select NRFX_UARTE1
+	select DEPRECATED
+	help
+	  Enable synchronous processing for UARTE1 trace backend.
```
didrikr, line 24: Is the help text or name here wrong?
Does it enable synchronous or asynchronous processing of the traces?

eivindj-nordic:
> More docs in apps/samples/Vscode extension on how to enable traces is definitely important when the old backend is eventually removed.

Agreed, more documentation is in the works by @divipillai.

eivindj-nordic:
> Is this planned for 2.4.0? There is an [upstream bug in the snippet implementation](https://github.com/zephyrproject-rtos/zephyr/issues/56531) which cause problems for the asset_tracker_v2 for the nRF9160 DK. When using the snippet, the board overlay files (_nrf9160dk_nrf9160_ns.overlay_ and _nrf9160dk_nrf9160_ns_0_14_0.overlay_) are not taken into account. This causes a compile error when using the board revision >=0.14.0: `west build -b nrf9160dk_nrf9160_ns@1.1.0`.

I think this will go in after NCS 2.4.0, unless the bug is fixed upstream before that and we can add the bugfix downstream.

divipillai:
> > More docs in apps/samples/Vscode extension on how to enable traces is definitely important when the old backend is eventually removed.
> 
> Agreed, more documentation is in the works by @divipillai.

As per discussion with @eivindj-nordic, the functionality about enabling modem tracing with the UART backend using the snippet is applicable for the following applications and samples:

• Asset Tracker v2 Application
• Serial LTE modem application
• All samples in the nRF9160 folder except slm_shell, modem_trace_flash, and modem_trace_backend

The documentation has been added to these application and sample documentation. 
With regards to VS code extension support, for now, the snippets can be added using the CMake arguments.
I have also added details about this in the documentation.

Please take a look and share if you have any suggestions on this.

plskeggs (approved):

gregersrygg (commented):
I really like many of the approaches here with the snippet and the chosen node to select which uart to use for modem tracing. So overall it's looking good, but I think it would be nice to have some unit test coverage.

lib/nrf_modem_lib/trace_backends/uart/Kconfig
```diff
@@ -8,18 +8,37 @@
...
-	# Modem tracing over UART use the UARTE1 as a dedicated peripheral.
-	# This enables the UARTE1 peripheral and includes nrfx UARTE driver.
-	select NRFX_UARTE1
+	bool "UART"
 
 if NRF_MODEM_LIB_TRACE_BACKEND_UART
 
+choice NRF_MODEM_LIB_TRACE_BACKEND_UART_VERSION
+	prompt "UART trace backend version"
+	default NRF_MODEM_LIB_TRACE_BACKEND_UART_ASYNC
```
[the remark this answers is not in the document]
eivindj-nordic: The reason for using the deprecated backend for default is that there is a lot of documentation and guides around that needs to be updated for the new trace backend (building with snippet instead of setting a Kconfig). So the plan is to keep the deprecated one as default for now, so that these guides will still work, but give a warning to the user. Then we will change the default and remove the old one in a later NCS version.

tests/lib/nrf_modem_lib/trace_backends/uart/src/main.c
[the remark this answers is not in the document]
eivindj-nordic: Yeah, I've had some issues adding them with CMock as the kernel and device drivers etc. aren't straight forward to mock. But I might look at the FFF test framework to see if that works better.

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -368,6 +378,14 @@ This section provides detailed lists of changes by :ref:`script <scripts>`.
...
       This change slows down the licenses detector, because it is no longer executed simultaneously on all files.
 
+Snippets
+========
+
+* Added:
+
+  * Support for nRF built-in snippets :ref:`nRF built-in snippets <nrf_snippets>`.
+  * Snippet for nRF91 modem tracing with the UART backend.
+
```
divipillai, line 388:
```suggestion
```
The content is already covered above.

doc/nrf/working_with_nrf/nrf91/nrf9160.rst
```diff
@@ -263,6 +263,43 @@ Building and programming on the command line
...
 
+.. _nrf91_modem_trace_uart_snippet:
+
+nRF91 modem tracing with UART backend using snippets
+====================================================
+
+The |NCS| provides built-in snippets that can be used with our sample and applications.
+See the zephyr's :ref:`snippets` documentation to know more about snippets in general.
+
+The snippet is tailored for tracing on the nRF91 Series devices but can work with other boards as well.
```
divipillai, line 274:
```suggestion
The :ref:`snippets` is tailored for tracing on the nRF91 Series devices but can work with other boards as well.
```

doc/nrf/working_with_nrf/nrf91/nrf9160.rst
```diff
@@ -263,6 +263,43 @@ Building and programming on the command line
...
       The device resets and runs the programmed sample or application.
 
+.. _nrf91_modem_trace_uart_snippet:
+
+nRF91 modem tracing with UART backend using snippets
+====================================================
+
+The |NCS| provides built-in snippets that can be used with our sample and applications.
+See the zephyr's :ref:`snippets` documentation to know more about snippets in general.
+
```
divipillai, line 273:
```suggestion
```
The link to zephyr documentation is provided in line no:274.

doc/nrf/working_with_nrf/nrf91/nrf9160.rst
```diff
@@ -263,6 +263,40 @@ Building and programming on the command line
...
 
+.. _nrf91_modem_trace_uart_snippet:
+
+nRF91 modem tracing with UART backend using snippets
+====================================================
+
+The :ref:`snippets` is tailored for tracing on the nRF91 Series devices but can work with other boards as well.
+It enables the :kconfig:option:`CONFIG_NRF_MODEM_LIB_TRACE` Kconfig option and chooses the Zephyr UART driver for the backend, with the necessary Kconfig options.
+Snippets also enables the UART1 peripheral with a baud rate of 1 Mbd and hardware flow control enabled.
+If this configuration does not match your requirements, you can add a snippet or Kconfig and device tree overlays to your application with the desired setup.
```
divipillai, line 274:
```suggestion
If this configuration does not match your requirements, you can add a snippet or Kconfig and devicetree overlays to your application with the desired setup.
```

doc/nrf/libraries/modem/nrf_modem_lib.rst
```diff
@@ -159,6 +159,69 @@ See the :ref:`external flash <nrf9160_external_flash>` documentation for more de
...
 
+.. _modem_trace_backend_uart_nrf9160dk:
+
+.. modem_lib_sending_traces_UART_start
+
+Sending traces over UART on the nRF9160 DK
+==========================================
+
+To send modem traces over UART on the nRF9160 DK, configuration must be added for the UART device in the devicetree and Kconfig.
+This is done by adding the :ref:`nrf91_modem_trace_uart_snippet` snippet when building and programming.
```
divipillai, line 170:
```suggestion
This is done by adding the :ref:`modem trace UART snippet <nrf91_modem_trace_uart_snippet>` when building and programming.
```

doc/nrf/working_with_nrf/nrf91/nrf9160.rst
```diff
@@ -263,6 +263,40 @@ Building and programming on the command line
 
       The device resets and runs the programmed sample or application.
 
+.. _nrf91_modem_trace_uart_snippet:
+
+nRF91 modem tracing with UART backend using snippets
+====================================================
+
+The :ref:`snippets` is tailored for tracing on the nRF91 Series devices but can work with other boards as well.
```
divipillai, line 271:
```suggestion
:ref:`snippets` are tailored for tracing on the nRF91 Series devices but can work with other boards as well.
```

divipillai (commented):
Some minor suggestions.

divipillai (approved):

lib/nrf_modem_lib/nrf_modem_lib_trace.c
```diff
@@ -204,12 +204,19 @@ static int trace_fragment_write(struct nrf_modem_trace_data *frag)
...
 			LOG_WRN("trace_backend wrote 0 bytes.");
+			/* On fault, modem reinitialization depends on the trace thread to complete
+			 * the processing of traces. In this case we want to prevent processing
+			 * traces forever. Therefore we abort if the trace backend write returns
+			 * zero when the modem is not initialized.
+			 */
+			if (!nrf_modem_is_initialized()) {
+				return -ESHUTDOWN;
+			}
 		}
```
lemrey, line 220: Could you tell in which situations the `write` function of the backend can return `0`? I think it should return either a positive value (on success) or a negative one (on error); `uart_send` in this file should never return `0`.

eivindj-nordic: Yes, ideally it should not return 0. I think I added this while developing the uart backend, when I saw that, if the backend were to return 0, it would block the modem shutdown/reinit. So, this is more a safeguard to be sure that this does not happen.

lib/nrf_modem_lib/trace_backends/uart/Kconfig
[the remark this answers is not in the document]
gregersrygg: Did you see the question about defaulting to the new uart backend when there is a chosen node for `nordic,modem-trace-uart`?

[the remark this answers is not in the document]
eivindj-nordic: No, that went past me :/ Added it now, thanks!

eivindj-nordic:
Hi all, 
I'm planning to merge this by the end of the week, so please let me know if you have any comments as soon as possible :)

tokangas (approved):

doc/nrf/working_with_nrf/nrf91/nrf9160.rst
```diff
@@ -263,6 +263,40 @@ Building and programming on the command line
...
       The device resets and runs the programmed sample or application.
 
+.. _nrf91_modem_trace_uart_snippet:
+
+nRF91 modem tracing with UART backend using snippets
+====================================================
+
+:ref:`snippets` are tailored for tracing on the nRF91 Series devices but can work with other boards as well.
+It enables the :kconfig:option:`CONFIG_NRF_MODEM_LIB_TRACE` Kconfig option and chooses the Zephyr UART driver for the backend, with the necessary Kconfig options.
+Snippets also enables the UART1 peripheral with a baud rate of 1 Mbd and hardware flow control enabled.
```
MirkoCovizzi, line 273:
```suggestion
Snippets also enable the UART1 peripheral with a baud rate of 1 Mbd and hardware flow control enabled.
```

lemrey (approved):
Good job

MarkusLassila (approved):
Works great

PR #10893, opened 2023-04-18
Crypto samples change logging backend to UART

mswarowsky:
There is no reason to use RTT for the crypto samples, so switch to RTT to make the config simpler


SebastianBoe (approved):

joerchan (approved):

samples/crypto/persistent_key_usage/prj.conf
```diff
@@ -4,14 +4,9 @@ CONFIG_SYS_CLOCK_TICKS_PER_SEC=1000
 CONFIG_MAIN_STACK_SIZE=4096
 CONFIG_HEAP_MEM_POOL_SIZE=4096
 
-# Enable loging using RTT and UART
+# Enable loging
```
frkv, line 7: Typo: loging => logging (multiple files)

samples/crypto/aes_cbc/prj.conf
```diff
@@ -4,14 +4,9 @@ CONFIG_SYS_CLOCK_TICKS_PER_SEC=1000
 CONFIG_MAIN_STACK_SIZE=4096
 CONFIG_HEAP_MEM_POOL_SIZE=4096
 
-# Enable loging using RTT and UART
+# Enable loging
```
torsteingrindvik, line 7: logging*

samples/crypto/persistent_key_usage/prj.conf
mswarowsky: Wonder how often that typo already got copy pasted, as it is in all samples :D, gonna fix it

frkv (changes requested):
Documentation for all crypto samples states output to RTT. If this change is to happen, then documentatin must reflect this

frkv (approved):
LGTM

mia-ko (changes requested):
Doc changes look ok to me :)
Two questions:
1. Does this also apply to crypto tests? I see they are still mentioning RTT.
2. Should we have an entry in the changelog for this?

melwee01:
Is this logging integrated with the logging levels we are putting in with https://github.com/nrfconnect/sdk-nrf/pull/10266/files?

mia-ko (approved):

mswarowsky:
> Is this logging integrated with the logging levels we are putting in with https://github.com/nrfconnect/sdk-nrf/pull/10266/files?

No, that doesn't seem to be related

PR #10895, opened 2023-04-19
doc: BT Direct Test Mode sample: fix and clean up

peknis:
Cleaning up the document. Added the
Configuration chapter with a sample-specific
Kconfig option introduced. Also added working
links to this option.


samples/bluetooth/direct_test_mode/README.rst
```diff
@@ -40,15 +40,15 @@ Test procedures are defined in the document `Bluetooth® Low Energy RF PHY Test
...
 
-DTM sample
-==========
+Implementation
+==============
 
 The DTM sample includes two parts:
 
-* The DTM, Direct Test Mode library, which manages the nRF radio and controls the standard DTM procedures.
+* The DTM, Direct Test Mode library that manages the nRF radio and controls the standard DTM procedures.
```
wiba-nordic, line 48:
```suggestion
* The Direct Test Mode library that manages the nRF radio and controls the standard DTM procedures.
```

samples/bluetooth/direct_test_mode/README.rst
```diff
@@ -121,16 +121,16 @@ The behavior of the commands vary depending on the hardware configuration and Kc
 
 * DTM with front-end module support:
 
-   * The official ``0x09`` DTM command when the :kconfig:option:`CONFIG_DTM_POWER_CONTROL_AUTOMATIC` Kconfig option is enabled, which is a default state.
+   * The official ``0x09`` DTM command when the :ref:`CONFIG_DTM_POWER_CONTROL_AUTOMATIC <CONFIG_DTM_POWER_CONTROL_AUTOMATIC>` Kconfig option is enabled, which is a default state.
      This command takes into account the SoC output power and the front-end module gain to set total output power.
 
      .. note::
-        The returned output power using this command is valid only for channel 0.
+        The returned output power using this command is valid only for channel ``0``.
```
wiba-nordic, line 128: I don't think this change is correct, since it's not talking about channel as a value for for anything.

samples/bluetooth/direct_test_mode/README.rst
```diff
@@ -121,16 +121,16 @@ The behavior of the commands vary depending on the hardware configuration and Kc
 
 * DTM with front-end module support:
 
-   * The official ``0x09`` DTM command when the :kconfig:option:`CONFIG_DTM_POWER_CONTROL_AUTOMATIC` Kconfig option is enabled, which is a default state.
+   * The official ``0x09`` DTM command when the :ref:`CONFIG_DTM_POWER_CONTROL_AUTOMATIC <CONFIG_DTM_POWER_CONTROL_AUTOMATIC>` Kconfig option is enabled, which is a default state.
```
wiba-nordic, line 124: This sentence seems to be lacking what the command does. Or, in other words, there should be something between "DTM command" and "when". Could also probably be written the same as the Line 133 as they seem to be contrasting options.

Also, "the default" instead of "a default".

samples/bluetooth/direct_test_mode/README.rst
```diff
@@ -314,12 +314,28 @@ Debugging
...
+
+Configuration options
+=====================
+
+The following configuration parameters are associated with this sample:
+
+.. _CONFIG_DTM_POWER_CONTROL_AUTOMATIC:
+
+CONFIG_DTM_POWER_CONTROL_AUTOMATIC - Automatic power control
+   Set the SoC output power and the front-end module gain to achieve the Tx output power requested by user.
```
wiba-nordic, line 335:
```suggestion
   Set the SoC output power and the front-end module gain to achieve the TX output power requested by the user.
```

wiba-nordic (changes requested):
There are also several instances of using "Tx" when "TX" is the correct capitalization

samples/bluetooth/direct_test_mode/README.rst
```diff
@@ -121,7 +121,7 @@ The behavior of the commands vary depending on the hardware configuration and Kc
 
 * DTM with front-end module support:
 
-   * The official ``0x09`` DTM command when the :kconfig:option:`CONFIG_DTM_POWER_CONTROL_AUTOMATIC` Kconfig option is enabled, which is a default state.
+   * The official ``0x09`` DTM command when the :ref:`CONFIG_DTM_POWER_CONTROL_AUTOMATIC <CONFIG_DTM_POWER_CONTROL_AUTOMATIC>` Kconfig option is enabled, which is a default state.
```
grochu, line 124: I would expand it to something like:
```suggestion
   * When the :ref:`CONFIG_DTM_POWER_CONTROL_AUTOMATIC <CONFIG_DTM_POWER_CONTROL_AUTOMATIC>` Kconfig option is enabled, which is a default state, the official ``0x09`` DTM command sets the final output power to the desired, or the closest available value by configuring the SoC output power and the front-end module gain.
```
But let's wait for Kamil so he can confirm.

grochu (approved):
Looks good, thanks for cleaning up these docs.

samples/bluetooth/direct_test_mode/README.rst
KAGA164: Looks good to me

samples/bluetooth/direct_test_mode/README.rst
```diff
@@ -314,12 +313,28 @@ Debugging
...
+Configuration
+*************
+|config|
+
+Configuration options
+=====================
+
+The following configuration parameters are associated with this sample:
+
+.. _CONFIG_DTM_POWER_CONTROL_AUTOMATIC:
```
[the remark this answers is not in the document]
peknis: This was the only one linked to in this sample doc. We can add others later, if we want to link to them.

KAGA164 (approved):

PR #10897, opened 2023-04-19
debug: Add ETB tracing library

maxd-nordic:
Add library to capture instruction traces and store them to the Embedded Trace Buffer. This also enables adding the buffer's contents to Memfault coredumps. Since ETB tracing is using lots of power, there is also a feature to turn off tracing while sleeping.

This is mostly @jtguggedal 's work.


jtguggedal:
Superseded by #10938 10938

PR #10910, opened 2023-04-21
mcuboot: cmake: Add DFU zip variant for MCUboot Direct-XIP multi-image

mkapala-nordic:
If there is MCUboot Direct-XIP used, there are two images generated, one for primary slot and one for secondary slot.
This change adds another variant of DFU zip which includes both images.

Jira: NCSDK-20558

Example manifest:
```
{
    "format-version": 0,
    "time": 1682426865,
    "files": [
        {
            "type": "application",
            "board": "nrf52840dk_nrf52840",
            "soc": "nRF52840_QIAA",
            "load_address": 49664,
            "version_MCUBOOT+XIP": "0.0.0+0",
            "slot": "0",
            "size": 194808,
            "file": "app_update.bin",
            "modtime": 1682426865
        },
        {
            "type": "application",
            "board": "nrf52840dk_nrf52840",
            "soc": "nRF52840_QIAA",
            "load_address": 545280,
            "version_MCUBOOT+XIP": "0.0.0+0",
            "slot": "1",
            "size": 194806,
            "file": "mcuboot_secondary_app_update.bin",
            "modtime": 1682426865
        }
    ],
    "name": "peripheral_lbs",
    "firmware": {
        "zephyr": {
            "revision": "74c7bca86e6a7d590333681eef90e159d79bcfe8"
        },
        "nrf": {
            "revision": "54d0addc0afcb8464aaccda4ba7adcb921451800-dirty"
        }
    }
}
```


modules/mcuboot/CMakeLists.txt
```diff
@@ -489,13 +489,27 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
...
       # No network core update
-      set(generate_bin_files
-        ${PROJECT_BINARY_DIR}/${app_core_binary_name}
-        )
-      set(generate_script_params
-        "load_address=$<TARGET_PROPERTY:partition_manager,PM_APP_ADDRESS>"
-        "version_MCUBOOT=${CONFIG_MCUBOOT_IMAGE_VERSION}"
-        )
+      if(CONFIG_BOOT_BUILD_DIRECT_XIP_VARIANT)
+        set(secondary_app_core_binary_name mcuboot_secondary_app_update.bin)
```
MarekPieta, line 493: Should we enclose `mcuboot_secondary_app_update.bin` in double quotes?

modules/mcuboot/CMakeLists.txt
```diff
@@ -489,13 +489,27 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
         )
     else()
```
MarekPieta, line 490: Please fix commit message
`mcuboot: cmake: Add DFU zip variant for MCUboot Direct-XIP`
This prevents confusing it with nRF53 multi-image update

MarekPieta (approved):

modules/mcuboot/CMakeLists.txt
mkapala-nordic: As far as i know it is not needed, here i aligned how it was done for example here: `set(app_core_binary_name app_update.bin)`

kapi-no (approved):

kapi-no (commented):
Consider performing the necessary documentation adjustments within this PR as the change is small (release notes + conceptual documentation). Currently, I found the following mentions of the direct-xip feature in the NCS documentation:

https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/mcuboot/readme-ncs.html
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/app_dev/bootloaders_and_dfu/index.html

In the context of this change, I think it would be good to mention the new way of generating dfu_application.zip in the following section:

https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/app_dev/build_and_config_system/index.html#mcuboot-output-build-files

It would be also great to mention the `mcuboot_secondary_app_update` as it seems to be missing from the documentation.

We may need to revise the direct-xip documentation at the high level and add a dedicated section regarding it or a short guide on how the user can work with it.

mkapala-nordic:
Added some documentation as per @kapi-no comment + rebased

doc/nrf/app_dev/build_and_config_system/index.rst
```diff
@@ -165,7 +165,7 @@ MCUboot output build files
...
 +==================================================================+====================================================================================================================================================================================================================================================================+
-| :file:`dfu_application.zip`                                      | Contains both the MCUboot-compatible update image for one or more cores and a manifest describing its contents (:file:`app_update.bin` and a :file:`manifest.json` file).                                                                                          |
+| :file:`dfu_application.zip`                                      | Contains both the MCUboot-compatible update image for one or more cores (or slots when MCUboot is in the Direct-XIP mode) and a manifest describing its contents (:file:`app_update.bin` and a :file:`manifest.json` file).                                        |
```
MarekPieta, line 168: Maybe we could inform that there is `manifest.json` and related `*.bin` files. You do not mention all of the files, so maybe such generic information would be better.

mkapala-nordic:
Applied @MarekPieta's comments

MarekPieta (approved):

philips77:
When sending *bin* from a client we need to know which file to send. 
We get the slot of the active firmware by requesting *image list* and checking *slot* property of the image with *active = true*. Then we need to send the fw corresponding to the other slot.
I would ask for a *slot* parameter to be added to json. The primary image could have `"slot" :0` and secondary `"slot": 1`.
Then we don't have to derive that from order or `load_address`.

mkapala-nordic:
Added slot metadata into manifest.json, example manifest in PR description.
@philips77 please take a look

kapi-no (approved):

MarekPieta (approved):

tejlmand (approved):

doc/nrf/app_dev/build_and_config_system/index.rst
```diff
@@ -165,7 +165,7 @@ MCUboot output build files
...
 +==================================================================+====================================================================================================================================================================================================================================================================+
-| :file:`dfu_application.zip`                                      | Contains both the MCUboot-compatible update image for one or more cores and a manifest describing its contents (:file:`app_update.bin` and a :file:`manifest.json` file).                                                                                          |
+| :file:`dfu_application.zip`                                      | Contains both the MCUboot-compatible update image for one or more cores (and slots when MCUboot is in the Direct-XIP mode) and a manifest describing its contents (all related :file:`*.bin` files and a :file:`manifest.json` file).                              |
```
annwoj, line 168: @mkapala-nordic I think might have to stick to the version discussed offline

Contains the following:
* The MCUboot-compatible update image for one or more cores when MCUboot is not in the Direct-XIP mode and a manifest describing its contents (all related :file:`*.bin` files and a :file:`manifest.json` file).
* The MCUboot-compatible update image for the primary and secondary slots when MCUboot is in the Direct-XIP mode and manifest describing its contents (all related :file:`*.bin` files and a :file:`manifest.json` file).

Unless we could rephrase it to something like this if it doesn't affect the meaning:

Contains a manifest describing all related :file:`*.bin` files and a :file:`manifest.json` file as well as the MCUboot-compatible update image for the following:
* One or more cores when MCUboot is not in the Direct-XIP mode
* The primary and secondary slots when MCUboot is in the Direct-XIP mode

mkapala-nordic: thanks, let's stick with the first version

annwoj (approved):

PR #10916, opened 2023-04-21
mcuboot: cmake: Fix signing in case of Direct-XIP image mismatch check

mkapala-nordic:
Add --rom-fixed option to the imgtool exec in order for the update using MCUmgr work correctly if
CONFIG_MCUMGR_GRP_IMG_REJECT_DIRECT_XIP_MISMATCHED_SLOT is currently enabled or was previously enabled (in case if the previous app image used it and in the current one it is no longer used).

Jira: NCSDK-20580

### DIFF
--- a/doc/nrf/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases/release-notes-changelog.rst
@@ -588,6 +588,9 @@ The following list summarizes both the main changes inherited from upstream MCUb
   * Support for the downgrade prevention feature using hardware security counters (:kconfig:option:`CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION`).
   * Generation of a new variant of the :file:`dfu_application.zip` when the :kconfig:option:`CONFIG_BOOT_BUILD_DIRECT_XIP_VARIANT` Kconfig option is enabled.
     Mentioned archive now contains images for both slots, primary and secondary.
+  * Encoding of the image start address into the header when the :kconfig:option:`CONFIG_BOOT_BUILD_DIRECT_XIP_VARIANT` Kconfig option is enabled.
+    The encoding is done using the ``--rom-fixed`` argument of the :file:`imgtool.py` script.
+    If the currently running application also has the :kconfig:option:`CONFIG_MCUMGR_GRP_IMG_REJECT_DIRECT_XIP_MISMATCHED_SLOT` Kconfig option enabled, the MCUmgr will reject application image updates signed without the start address.
 
 Zephyr
 ======
--- a/modules/mcuboot/CMakeLists.txt
+++ b/modules/mcuboot/CMakeLists.txt
@@ -52,9 +52,10 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
     # SIGNED_BIN_FILE_OUT - (optional) Signed bin output image
     # SIGNED_HEX_TEST_FILE_OUT - (optional) Signed hex output test image including IMAGE_MAGIC
     # IMAGE_DEPENDENCIES - (optional) Dependencies between signed image and other images
+    # ROM_FIXED - (optional) Start address of image
     # DEPENDS - (optional) One or more dependencies for signing the hex image
     set(oneValueArgs SIGNED_BIN_FILE_IN SIGNED_HEX_FILE_NAME_PREFIX SLOT_SIZE SIGNED_HEX_FILE_OUT)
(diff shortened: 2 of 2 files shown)


MarekPieta (approved):

mkapala-nordic:
Applied @MarekPieta's comments + added release note as per @kapi-no's suggestion + rebased

MarekPieta (approved):

kapi-no (approved):

mkapala-nordic:
Hi @tejlmand, could you please take a look as a code owner at this PR?

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -547,6 +547,8 @@ The code for integrating MCUboot into |NCS| is located in the :file:`ncs/nrf/mod
 The following list summarizes both the main changes inherited from upstream MCUboot and the main changes applied to the |NCS| specific additions:
 
 * Added support for the downgrade prevention feature using hardware security counters (:kconfig:option:`CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION`).
+* Added encoding of the image start address into the header (using imgtool's ``--rom-fixed`` argument) when the :kconfig:option:`CONFIG_BOOT_BUILD_DIRECT_XIP_VARIANT` Kconfig option is enabled.
+  If the currently running application also has the :kconfig:option:`CONFIG_MCUMGR_GRP_IMG_REJECT_DIRECT_XIP_MISMATCHED_SLOT` Kconfig option enabled, the MCUmgr will reject application image updates signed without the start address.
```
annwoj, line 551:
```suggestion
* Added:

  * Support for the downgrade prevention feature using hardware security counters (:kconfig:option:`CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION`).
  * Encoding of the image start address into the header when the :kconfig:option:`CONFIG_BOOT_BUILD_DIRECT_XIP_VARIANT` Kconfig option is enabled. 
    The encoding is done using the ``--rom-fixed`` argument of the :file:`imgtool.py` script.
    If the currently running application also has the :kconfig:option:`CONFIG_MCUMGR_GRP_IMG_REJECT_DIRECT_XIP_MISMATCHED_SLOT` Kconfig option enabled, the MCUmgr will reject application image updates signed without the start address.
```

Reshuffled a bit, please let me know if the meaning of the 2nd bullet point is affected.

mkapala-nordic: Looks fine, thanks

annwoj (approved):

modules/mcuboot/CMakeLists.txt
```diff
@@ -127,6 +132,7 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
       # IMAGE_MAGIC into the image trailer.
       ${sign_cmd}
       ${sign_dependencies}
+      ${rom_fixed_option}
```
tejlmand, line 135: minor nit: it can be risky to use a variable inside a fuction where it has not been ensured whether such variable has been defined or not.

Reason for that is that in case `rom_fixed_option` already exists in the caller's scope, then that value will actually be used, which often come as a surprise and lead to hard-to-debug issues.

See other comment for proposed handling.
```suggestion
      ${SIGN_ARG_ROM_FIXED}
```

modules/mcuboot/CMakeLists.txt
```diff
@@ -71,6 +72,10 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
       set(${SIGN_ARG_SIGNED_HEX_TEST_FILE_OUT} ${test_update_hex} PARENT_SCOPE)
     endif()
 
+    if(DEFINED SIGN_ARG_ROM_FIXED)
+      set(rom_fixed_option --rom-fixed ${SIGN_ARG_ROM_FIXED})
```
tejlmand, line 76: as mentioned in other comment, this only sets the var if `SIGN_ARG_ROM_FIXED` is defined, but in cases where it is not defined then the value of `rom_fixed_option` is picked up from the parent scope.

One option is to ensure the local scope var is always unset, but a another approach can simply be to prepend the arg var with the required prefix, like:
```suggestion
      set(SIGN_ARG_ROM_FIXED --rom-fixed ${SIGN_ARG_ROM_FIXED})
```

tejlmand (changes requested):

mkapala-nordic:
Applied @tejlmand's comments

tejlmand (approved):

PR #10918, opened 2023-04-24
samples: matter: Fix NUS switching.

ArekBalysNordic:
There should be a mechanism to disable NUS connection and eventing during the commissioning to Matter process.

Added all lock states to NUS service.
It is needed to send the new state after changing to BLE NUS BT characteristic to allow reading it by the controller.

### DIFF
--- a/samples/matter/common/src/bt_nus_service.cpp
+++ b/samples/matter/common/src/bt_nus_service.cpp
@@ -60,30 +60,42 @@ bool NUSService::Init(uint8_t priority, uint16_t minInterval, uint16_t maxInterv
 
 	mAdvertisingRequest.onStarted = [](int rc) {
 		if (rc == 0) {
-			LOG_INF("NUS BLE advertising started");
+			GetNUSService().mIsStarted = true;
+			LOG_DBG("NUS BLE advertising started");
 		} else {
 			LOG_ERR("Failed to start NUS BLE advertising: %d", rc);
 		}
 	};
-	mAdvertisingRequest.onStopped = []() { LOG_DBG("NUS BLE advertising stopped"); };
+	mAdvertisingRequest.onStopped = []() {
+		bt_conn_auth_info_cb_unregister(&sConnAuthInfoCallbacks);
+		bt_conn_auth_cb_register(nullptr);
+		bt_nus_init(nullptr);
 
-	settings_load();
+		GetNUSService().mIsStarted = false;
+		LOG_DBG("NUS BLE advertising stopped");
+	};
+
 #if defined(CONFIG_BT_FIXED_PASSKEY)
-	VerifyOrReturnError(bt_passkey_set(CONFIG_CHIP_NUS_FIXED_PASSKEY) == 0, false);
+	if (bt_passkey_set(CONFIG_CHIP_NUS_FIXED_PASSKEY) != 0)
+		return false;
 #endif
-	VerifyOrReturnError(bt_conn_auth_cb_register(&sConnAuthCallbacks) == 0, false);
-	VerifyOrReturnError(bt_conn_auth_info_cb_register(&sConnAuthInfoCallbacks) == 0, false);
-	VerifyOrReturnError(bt_nus_init(&sNusCallbacks) == 0, false);
 
 	return true;
 }
 
-void NUSService::StartServer()
+bool NUSService::StartServer()
 {
 	if (mIsStarted) {
 		LOG_WRN("NUS service was already started");
-		return;
+		return false;
 	}
+
+	if (bt_conn_auth_cb_register(&sConnAuthCallbacks) != 0)
+		return false;
+	if (bt_conn_auth_info_cb_register(&sConnAuthInfoCallbacks) != 0)
+		return false;
+	if (bt_nus_init(&sNusCallbacks) != 0)
+		return false;
 
 	PlatformMgr().LockChipStack();
 	CHIP_ERROR ret = BLEAdvertisingArbiter::InsertRequest(mAdvertisingRequest);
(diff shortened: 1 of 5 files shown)


samples/matter/common/src/bt_nus_service.cpp
```diff
@@ -60,20 +60,25 @@ bool NUSService::Init(uint8_t priority, uint16_t minInterval, uint16_t maxInterv
 
 	mAdvertisingRequest.onStarted = [](int rc) {
 		if (rc == 0) {
-			LOG_INF("NUS BLE advertising started");
+			GetNUSService().mIsStarted = true;
```
kkasperczyk-no, line 63: I think `GetNUSService()` is not needed, as the field should be accessible directly from this method.

samples/matter/common/src/bt_nus_service.cpp
```diff
@@ -85,6 +90,10 @@ void NUSService::StartServer()
 		return;
 	}
 
+	VerifyOrReturn(bt_conn_auth_cb_register(&sConnAuthCallbacks) == 0);
```
kkasperczyk-no, line 93: Maybe we should change the type returned by `StartServer` to bool or int? So far there wasn't much that could fail in this method, but now we've got some checks moved from `Init` that actually may fail.

samples/matter/lock/README.rst
```diff
@@ -234,6 +234,10 @@ The PIN code is different depending on the build type:
 
 See `Testing door lock using Bluetooth LE with Nordic UART Service`_ for more information about how to test this feature.
 
+.. note::
+   In this sample, there is only one Bluetooth connection available simultaneously.
```
kkasperczyk-no, line 238:
```suggestion
   In this sample, there is only one Bluetooth LE connection available simultaneously.
```

samples/matter/lock/README.rst
```diff
@@ -581,6 +589,9 @@ To test the :ref:`matter_lock_sample_ble_nus` feature, complete the following st
 
 The Bluetooth LE connection between a phone and the DK will be suspended when the commissioning to the Matter network is in progress or there is an active session of SMP DFU.
 
+To read the current door lock state from the device, read RX Bluetooth LE characteristics.
```
kkasperczyk-no, line 592:
```suggestion
To read the current door lock state from the device, read Bluetooth LE RX characteristic.
```

samples/matter/common/src/bt_nus_service.cpp
ArekBalysNordic: But then I cannot assign in the simple way that lambda to onStarted method.

LuDuda (approved):

samples/matter/lock/README.rst
```diff
@@ -234,6 +234,10 @@ The PIN code is different depending on the build type:
 
 See `Testing door lock using Bluetooth LE with Nordic UART Service`_ for more information about how to test this feature.
 
+.. note::
+   This sample supports one Bluetooth LE connection at a time.
+   Matter commissioning, DFU, and NUS over Bluetooth LE have to be run separately.
```
greg-fer, line 239:
```suggestion
.. note::
   This sample supports one Bluetooth LE connection at a time.
   Matter commissioning, DFU, and NUS over Bluetooth LE must be run separately.
```
All good, but I would place this in line 223, with appropriate empty lines before and after.

samples/matter/lock/README.rst
```diff
@@ -581,6 +588,9 @@ To test the :ref:`matter_lock_sample_ble_nus` feature, complete the following st
 
 The Bluetooth LE connection between a phone and the DK will be suspended when the commissioning to the Matter network is in progress or there is an active session of SMP DFU.
 
+To read the current door lock state from the device, read Bluetooth LE RX characteristic.
```
greg-fer, line 591:
```suggestion
To read the current door lock state from the device, read the Bluetooth LE RX characteristic.
```

samples/matter/lock/README.rst
```diff
@@ -581,6 +588,9 @@ To test the :ref:`matter_lock_sample_ble_nus` feature, complete the following st
...
+The new lock state is updated after changing the state from any of the following source: NUS, buttons, Matter stack.
```
greg-fer, line 592:
```suggestion
The new lock state is updated after changing the state from any of the following sources: NUS, buttons, Matter stack.
```

samples/matter/lock/src/app_task.cpp
```diff
@@ -636,22 +638,28 @@ void AppTask::LockStateChanged(BoltLockManager::State state, BoltLockManager::Op
...
+#ifdef CONFIG_CHIP_NUS
+		GetNUSService().SendData("locking", sizeof("locking"));
+#endif
 		break;
 	case BoltLockManager::State::kLockingCompleted:
 		LOG_INF("Lock action completed");
 		sLockLED.Set(true);
 #ifdef CONFIG_CHIP_NUS
-		GetNUSService().SendData("Locked", sizeof("Locked"));
+		GetNUSService().SendData("locked", sizeof("locked"));
```
markaj-nordic, line 649: BTW. NUS API would be more concise if the user didn't have to provide the size of a command. Not sure if we can assume null terminated string each time though.

samples/matter/common/src/bt_nus_service.cpp
```diff
@@ -60,41 +60,50 @@ bool NUSService::Init(uint8_t priority, uint16_t minInterval, uint16_t maxInterv
...
+			GetNUSService().mIsStarted = true;
+			LOG_DBG("NUS BLE advertising started");
 		} else {
 			LOG_ERR("Failed to start NUS BLE advertising: %d", rc);
 		}
 	};
-	mAdvertisingRequest.onStopped = []() { LOG_DBG("NUS BLE advertising stopped"); };
+	mAdvertisingRequest.onStopped = []() {
+		bt_conn_auth_info_cb_unregister(&sConnAuthInfoCallbacks);
+		bt_conn_auth_cb_register(NULL);
```
markaj-nordic, line 71: nit: NULL -> nullptr?

samples/matter/common/src/bt_nus_service.cpp
```diff
@@ -60,41 +60,50 @@ bool NUSService::Init(uint8_t priority, uint16_t minInterval, uint16_t maxInterv
...
-void NUSService::StartServer()
+bool NUSService::StartServer()
 {
 	if (mIsStarted) {
 		LOG_WRN("NUS service was already started");
-		return;
+		return false;
 	}
 
+	VerifyOrReturnError(bt_conn_auth_cb_register(&sConnAuthCallbacks) == 0, false);
```
markaj-nordic, line 93: nit: is there any difference between `VerifyOrReturnError` and `VerifyOrReturnValue`? Isn't the first one dedicated to CHIP_ERROR type?

markaj-nordic (approved):
My general thought is that we should not mix the NCS application specific code with Matter SDK code, i.e. NUS Service shouldn't know anything about the Matter. Looks like here we use chip namespaces only to synchronize the chip stack and leverage the VerifyOrReturn...() utilities. To remove these dependencies probably the BLE Arbiter would need to be slightly refactored to provide thread safety, but this is obviously not the scope of this PR.

greg-fer (approved):

ArekBalysNordic:
@markaj-nordic I addressed your comments and removed most of the chip-related code from NUS. I need to leave BLEArbiter there until it is not separated from the Matter stack. 

After discussion, we decided to disable Bluetooth bonding for the NUS feature. PR has been rebased.

PR #10919, opened 2023-04-24
manifest: Update Matter revision to pull Rotating Device UID

ArekBalysNordic:
The fix allows user to use Kconfigs to set Rotating Device UID.


LuDuda (commented):
Just a question.. For NCS examples, don't we also need to calll:

```
SetDeviceInstanceInfoProvider(&DeviceInstanceInfoProviderMgrImpl());
```

In some places?

ArekBalysNordic:
> Just a question.. For NCS examples, don't we also need to calll:
> 
> ```
> SetDeviceInstanceInfoProvider(&DeviceInstanceInfoProviderMgrImpl());
> ```
> 
> In some places?

Right, I will add it.

LuDuda (approved):

markaj-nordic (approved):

PR #10923, opened 2023-04-24
samples: matter: Fixed build failure for DFU over BT SMP.

kkasperczyk-no:
The DFU over BT SMP cannot be built without enablign Matter OTA Requestor, which is incorrect dependency.

* Moved ota_image_processor_base_impl.h include under Matter OTA Requestor ifdef
* Updated Matter revision to pull Kconfig dependencies fix

### DIFF
--- a/samples/matter/common/src/ota_util.cpp
+++ b/samples/matter/common/src/ota_util.cpp
@@ -5,7 +5,6 @@
  */
 
 #include "ota_util.h"
-#include "ota_image_processor_base_impl.h"
 
 #if CONFIG_CHIP_OTA_REQUESTOR
 #include <app/clusters/ota-requestor/BDXDownloader.h>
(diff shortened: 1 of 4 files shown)


markaj-nordic (approved):
Thanks for the fix, I think we should also only add these source files when building with CONFIG_CHIP_OTA_REQUESTOR:
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/lock/CMakeLists.txt#L69

kkasperczyk-no:
> Thanks for the fix, I think we should also only add these source files when building with CONFIG_CHIP_OTA_REQUESTOR: https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/lock/CMakeLists.txt#L69

Right, I would even say `THREAD_WIFI_SWITCHING` symbol should depend on `CONFIG_CHIP_OTA_REQUESTOR`. I updated PR with this change.

LuDuda (approved):

PR #10924, opened 2023-04-24
mpsl: cx: Do not support CX with FEM on 53

weeTike:
Generic Coex shall not be used when FEM is enabled on nRF5340. This can lead to undefined behavior.


martintv (approved):

martintv:
> Any suggestions on other places we should document this are welcome.

here maybe? https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/app_dev/wifi_coex/

doc/nrf/app_dev/wifi_coex/index.rst
```diff
@@ -24,6 +24,8 @@ Requirements
 
 If your application uses SR protocols and requires coexistence with Wi-Fi, you must enable :ref:`nrfxlib:mpsl_cx` in the :ref:`nrfxlib:mpsl` (MPSL) library.
 
+On the nRF5340, Wi-Fi Coexistence shall not be enabled in conjunction with FEM, as this can lead to undefined behavior.
```
wiba-nordic, line 27:
```suggestion
Do not enable Wi-Fi coexistence on the nRF5340 DK in conjunction with FEM, as this can lead to undefined behavior.
```

wiba-nordic (changes requested):

subsys/mpsl/cx/Kconfig
```diff
@@ -21,13 +21,16 @@ config MPSL_CX_NRF700X_SUPPORT
 config MPSL_CX
 	bool "Radio Coexistence interface support"
 	depends on MPSL_CX_ANY_SUPPORT
+	depends on !MPSL_FEM && !SOC_SERIES_NRF53X
```
jciupis, line 24: Is this correct? I'm assuming the intention was to express that
```
config MPSL_CX
    depends on !(MPSL_FEM && SOC_SERIES_NRF53X) # we don't want CX && FEM && NRF53 at the same time
```

which without parentheses should be:
```
config MPSL_CX
    depends on !MPSL_FEM || !SOC_SERIES_NRF53X
```
In my understanding in its current shape, this change removes support for CX on nRF53 _and_ it removes support for CX + FEM for all families.

jciupis (changes requested):
Perhaps I'm missing something, but the implementation seems incorrect to me.

subsys/mpsl/cx/Kconfig
weeTike: You are right, good catch, thank you.

wiba-nordic (approved):

jciupis (approved):

martintv (approved):

PR #10927, opened 2023-04-24
applications: asset_tracker_v2: replace battery voltage with fuel gauge

maxd-nordic:
This patch uses the fuel gauge function introduced in PR 9526 to get a battery percentage instead of a voltage.


applications/asset_tracker_v2/src/events/sensor_module_event.h
```diff
@@ -44,6 +44,11 @@ enum sensor_module_event_type {
 	 */
 	SENSOR_EVT_ENVIRONMENTAL_DATA_READY,
 
+	/** Battery fuel gauge data has been sampled.
+	 *  Payload is of type @ref sensor_module_data (batt).
```
jtguggedal, line 48: `batt` -> `bat`

jtguggedal (approved):

coderbyheart:
Could you provide an explanation in the commit, why percentage is better and voltage should be ignored?

umapraseeda:
Do we need a changelog entry for this? If not, the label could be removed.

jtguggedal:
> Do we need a changelog entry for this? If not, the label could be removed.

I think we need it

applications/asset_tracker_v2/src/modules/sensor_module.c
```diff
@@ -242,6 +246,27 @@ static void apply_config(struct sensor_msg_data *msg)
 #endif /* CONFIG_EXTERNAL_SENSORS */
 }
 
+static void battery_data_get(void)
+{
+#if defined(CONFIG_ADP536X)
+	int rc;
```
simensrostad, line 252:
```suggestion
	int err;
```

applications/asset_tracker_v2/src/modules/sensor_module.c
```diff
@@ -350,13 +387,17 @@ static void on_state_running(struct sensor_msg_data *msg)
...
+		    msg->module.app.count)) {
+			environmental_data_get();
 		}
 
-		environmental_data_get();
+		if (battery_data_requested(
+		    msg->module.app.data_list,
+		    msg->module.app.count)) {
+			battery_data_get();
+			}
```
simensrostad, line 400: Formatting shananigans?

simensrostad (commented):
Should AWS, Azure, and LwM2M codecs be updated as well?

maxd-nordic:
> Should AWS, Azure, and LwM2M codecs be updated as well?

As far as I could see only nRF Cloud mentioned the value as being a voltage, so I think it should work fine on all of them with the patch. Of course the range changes.

simensrostad (approved):

PR #10646, opened 2023-03-21
lib: dk_buttons_and_leds: fix for nRF9160DK IO expander

eivindj-nordic:
The driver for the PCAL6408A IO expander on the nRF9160DK does not support disabling the interrupts while they are being processed. Move disabling of the interrupts to the buttons_scan_fn function. Also, the IO expander driver only supports edge driven interrupts, so this is added to the configuration.

PR #10930, opened 2023-04-24
lib: dk_buttons_and_leds: reduce power consumption in idle mode

eivindj-nordic:
Edge triggered interrupts, using GPIOTE IN events has higher idle current consumption compared to active high triggered interrupts using GPIOTE port events. Hence, #10646 increased the idle power consumption when the DK Buttons and LEDs library was enabled. This PR fixes the issue by having different interrupt configuration with the internal GPIO0 peripheral and the external GPIO expander.

### DIFF
--- a/lib/dk_buttons_and_leds/dk_buttons_and_leds.c
+++ b/lib/dk_buttons_and_leds/dk_buttons_and_leds.c
@@ -18,6 +18,12 @@ LOG_MODULE_REGISTER(dk_buttons_and_leds, CONFIG_DK_LIBRARY_LOG_LEVEL);
 #define BUTTONS_NODE DT_PATH(buttons)
 #define LEDS_NODE DT_PATH(leds)
 
+#define GPIO0_DEV DEVICE_DT_GET(DT_NODELABEL(gpio0))
+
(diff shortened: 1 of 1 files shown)


moksanen:
I did a quick current measurement test run for this PR, and PSM idle seems to be back down at ~3uA, which is the expected level.
If the change is otherwise ok, please proceed with approvals and merging.

eivindj-nordic:
> I did a quick current measurement test run for this PR, and PSM idle seems to be back down at ~3uA, which is the expected level.
If the change is otherwise ok, please proceed with approvals and merging.

Thank you for checking! @MirkoCovizzi @jhn-nordic, please have a look :)

moksanen:
Upmerge was merged, so I suppose this needs to be rebased. Then, I hope this can be merged soon.

eivindj-nordic:
Rebased.

ktaborowski (approved):
looks good to me - reduces idle power consumption in nrf52840 dk sample

moksanen:
I guess another rebase is needed to get twister tests pass, since a test failing bug was introduced during upmerge.

PR #10935, opened 2023-04-25
lib: lte_link_control: Proprietary PSM in modem sleep notification

trantanen:
Added support for proprietary PSM (supported when mfw >= 2.0.0) in modem sleep notification.
Added "proprietary PSM" information in modem shell too for modem sleep notification.

JIRA: MOSH-442


trantanen:
These changes have been only tested with unit test as I don't have a SIM card which would connect to a network that doesn't provide PSM functionality. LTE link controller changes have been tested with my new set of lte_lc_api tests that I will submit later as part of adding some more unit tests in there.

jtguggedal (commented):
Looks good. 
Could you also add a test for the new sleep type in `nrf/tests/lib/lte_lc/src/main.c:test_parse_xmodemsleep`?

jtguggedal (approved):

tokangas (approved):

PR #10940, opened 2023-04-25
manifest: Update to nrf_cc3xx 0.9.17 and nrf_oberon 3.0.13

joerchan:
Update crypto runtime libraries:
nrf_cc3xx 0.9.17
nrf_oberon 3.0.13

test-sdk-nrf: testing-cryptocell-oberon-release-04-25

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -127,7 +127,7 @@ manifest:
     - name: mbedtls
       path: modules/crypto/mbedtls
       repo-path: sdk-mbedtls
-      revision: 3c2bc903cb396af6033115c8186f0807ef81bdaf
+      revision: 9ca9f885bafb3a5d4896600f478147ad69757d82
     - name: nrfxlib
       repo-path: sdk-nrfxlib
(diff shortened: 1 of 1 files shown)


melwee01:
To summarize today: Doc build is failing because of formatting issues in the Mbed TLS bignum.h file. This comes from nrfxlib PR 960, but it isn't a file we have ultimate control over. The error showed here, not in that PR, because that repository doesn't build docs by default. 

We can get more information on build errors like these by doing a local build. The errors in PR builds are based on the output file and don't specify the error-causing text. Since the build stops, we can't look at the output file and find the correct line. The output files are much larger than the source files, so the line numbers don't match up.

In this case, we can make the change ourselves to get the build going again. But the source in Mbed TLS will need to be changed if we don't want to replicate this change every time. Quick fix now, future fix TBD.

frkv (approved):
LGTM pending resolving tests and adding correct SHA

PR #10943, opened 2023-04-25
Bluetooth: Mesh: Add motion threshold to sensor server/client

Balaklaka:
Sensor server configuration update to improve preformance and added motion threshold to sensor server and client.


samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -23,15 +23,15 @@ static void sensor_cli_data_cb(struct bt_mesh_sensor_cli *cli,
 		if (value->val1) {
 			printk("Presence detected\n");
 		} else {
-			printk("End of presence\n");
+			printk("No presence detected\n");
```
PavelVPV, line 26: Note that this change will brake current system tests for the sensor client sample.

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -267,6 +267,11 @@ Bluetooth mesh samples
 * Removed the :file:`hci_rpmsg.conf` file from all samples that support nRF5340 DK or Thingy:53.
   This configuration is moved to the :file:`ncs/nrf/subsys/bluetooth/mesh/hci_rpmsg_child_image_overlay.conf` file.
 
+* :ref:`bluetooth_mesh_sensor_server` and :ref:`bluetooth_mesh_sensor_client` samples:
+
+  * Added support for motion threshold as setting for the presence detection.
```
mia-ko, line 272:
```suggestion
  * Added support for motion threshold as a setting for the presence detection.
```

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -267,6 +267,11 @@ Bluetooth mesh samples
...
+  * Increased Kconfig options :kconfig:option:`CONFIG_BT_MESH_SUBNET_COUNT`, :kconfig:option:`CONFIG_BT_MESH_APP_KEY_COUNT`, :kconfig:option:`CONFIG_BT_MESH_CRPL` and :kconfig:option:`CONFIG_BT_MESH_MSG_CACHE_SIZE`.
```
mia-ko, line 273:
```suggestion
  * Increased the following Kconfig options:
    
    * :kconfig:option:`CONFIG_BT_MESH_SUBNET_COUNT`
    * :kconfig:option:`CONFIG_BT_MESH_APP_KEY_COUNT`
    * :kconfig:option:`CONFIG_BT_MESH_CRPL`
    * :kconfig:option:`CONFIG_BT_MESH_MSG_CACHE_SIZE`.
```

samples/bluetooth/mesh/sensor_client/README.rst
```diff
@@ -39,6 +39,7 @@ The following Bluetooth mesh sensor types are used in this sample:
 * :c:var:`bt_mesh_sensor_rel_runtime_in_a_dev_op_temp_range` - Periodically requested by the client.
 * :c:var:`bt_mesh_sensor_presence_detected` - Published when a button is pressed on the server.
 * :c:var:`bt_mesh_sensor_time_since_presence_detected` - Periodically requested by the client and published by the server according to its publishing period.
+* :c:var:`bt_mesh_sensor_motion_threshold` - Used as a setting for the :c:var:`bt_mesh_sensor_presence_detected` sensor type to set the time (0-10 sek) before the presence is detected.
```
mia-ko, line 42:
```suggestion
* :c:var:`bt_mesh_sensor_motion_threshold` - Used as a setting for the :c:var:`bt_mesh_sensor_presence_detected` sensor type to set the time (0-10 seconds) before the presence is detected.
```

samples/bluetooth/mesh/sensor_server/README.rst
```diff
@@ -45,6 +45,7 @@ The following Bluetooth mesh sensor types are used in this sample:
 * :c:var:`bt_mesh_sensor_rel_runtime_in_a_dev_op_temp_range` - Periodically requested by the client.
 * :c:var:`bt_mesh_sensor_presence_detected` - Published when a button is pressed on the server.
 * :c:var:`bt_mesh_sensor_time_since_presence_detected` - Periodically requested by the client and published by the server according to its publishing period (see :ref:`bluetooth_mesh_sensor_server_conf_models`).
+* :c:var:`bt_mesh_sensor_motion_threshold` - Used as a setting for the :c:var:`bt_mesh_sensor_presence_detected` sensor type to set the time (0-10 sek) before the presence is detected.
```
mia-ko, line 48:
```suggestion
* :c:var:`bt_mesh_sensor_motion_threshold` - Used as a setting for the :c:var:`bt_mesh_sensor_presence_detected` sensor type to set the time (0-10 seconds) before the presence is detected.
```

samples/bluetooth/mesh/sensor_server/README.rst
```diff
@@ -99,7 +100,7 @@ Buttons:
    All buttons have the same functionality during the provisioning procedure.
 
 Button 1:
-   Simulates presence detected (after the provisioning procedure is finished).
+   Simulates presence detected (after the provisioning procedure is finished). Time before presence is detected is dependent of the motion threshold.
```
mia-ko, line 103:
```suggestion
   Simulates presence detected (after the provisioning procedure is finished). The time before the presence is detected depends on the motion threshold.
```

mia-ko (changes requested):

mia-ko (approved):

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -266,26 +357,32 @@ static void button_handler_cb(uint32_t pressed, uint32_t changed)
 		return;
 	}
 
-	if (pressed & BIT(0)) {
-		int err;
+	if (pressed & changed & BIT(0)) {
+		k_work_reschedule(&presence_detected_work, K_MSEC(100*pres_mot_thres.val1));
```
PavelVPV, line 361:
```suggestion
		k_work_reschedule(&presence_detected_work, K_MSEC(100 * pres_mot_thres.val1));
```

samples/bluetooth/mesh/sensor_server/README.rst
```diff
@@ -99,7 +100,7 @@ Buttons:
    All buttons have the same functionality during the provisioning procedure.
 
 Button 1:
-   Simulates presence detected (after the provisioning procedure is finished).
+   Simulates presence detected (after the provisioning procedure is finished). The time before the presence is detected depends on the motion threshold.
```
PavelVPV, line 103: Should we add that (as I understand) the button needs to be pressed the required time to send presence detection?

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -23,15 +23,15 @@ static void sensor_cli_data_cb(struct bt_mesh_sensor_cli *cli,
...
 		}
 	} else if (sensor->id ==
 		   bt_mesh_sensor_time_since_presence_detected.id) {
 		if (value->val1) {
 			printk("%s second(s) since last presence detected\n",
 			       bt_mesh_sensor_ch_str(value));
 		} else {
-			printk("No presence detected, or under 1 second since presence detected\n");
+			printk("Presence detected, or under 1 second since presence detected\n");
```
PavelVPV, line 34: I think the previous message was more correct, because `prev_pres` is 0 at the start when no presence is detected yet.

Balaklaka: prev_pres is set at start, so it is not 0 at start. So it will start reporting last presence detected right away.
The other thing is that it is strange to say "No presence detected,.." when there is actual a presence there. As long it is registered a presence this value can be sent, as it is possible to have a longer presence now. (Long as the user hold the button there is a presence.)

PavelVPV: Ok, but is it correct to send that presence was detected at the start?

Balaklaka: Before it is started is no way to say if presence or not exist. So it could be argued both ways. In my view this is the most correct way at least at startup. It says something about time since last presence is detected or it could have been a presence before startup.

PavelVPV (approved):

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -23,15 +23,15 @@ static void sensor_cli_data_cb(struct bt_mesh_sensor_cli *cli,
 		if (value->val1) {
 			printk("Presence detected\n");
 		} else {
-			printk("End of presence\n");
+			printk("No presence detected\n");
 		}
 	} else if (sensor->id ==
 		   bt_mesh_sensor_time_since_presence_detected.id) {
 		if (value->val1) {
```
omkar3141, line 30: Could you please also add handling for special value 0xFFFF here (ref: GSS v5, section 3.219.2)? I think this was missed originally.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -266,26 +357,32 @@ static void button_handler_cb(uint32_t pressed, uint32_t changed)
...
+	}
 
-		/* This sensor value must be boolean -
-		 * .val1 can only be '0' or '1'
-		 */
-		struct sensor_value val = {
-			.val1 = 1,
-		};
+	if ((!pressed) & changed & BIT(0)) {
+		if (prev_pres) {
```
omkar3141, line 365: I don't think this will work to scale the button pressing time by motion threshold. I think you will have to sample uptime when k_work is scheduled and then compare it with current and check how many seconds have elapsed.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -207,8 +208,96 @@ static struct bt_mesh_sensor rel_chip_temp_runtime = {
...
 };
 
+static void presence_motion_threshold_get(struct bt_mesh_sensor_srv *srv,
+	struct bt_mesh_sensor *sensor,
+	const struct bt_mesh_sensor_setting *setting,
+	struct bt_mesh_msg_ctx *ctx,
+	struct sensor_value *rsp)
+{
+	rsp[0] = pres_mot_thres;
+	printk("Presence motion threshold: %u\n", rsp[0].val1);
```
omkar3141, line 218:
```suggestion
	printk("Presence motion threshold: %u (%d ms)\n", rsp[0].val1, 100 * rsp[0].val1);
```

omkar3141 (changes requested):

samples/bluetooth/mesh/sensor_client/prj.conf
```diff
@@ -15,7 +15,9 @@ CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=4096
 CONFIG_FLASH=y
 CONFIG_FLASH_MAP=y
 CONFIG_NVS=y
+CONFIG_NVS_LOOKUP_CACHE=y
```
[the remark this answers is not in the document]
PavelVPV: I'm uncertain at the current moment if we need to make them default. There are more option that would be good to make default for mesh as well, for example, partial erase. But for now I'd add these 2 options to this page: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/protocols/bt_mesh/configuring.html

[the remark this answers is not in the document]
omkar3141: Ok.

samples/bluetooth/mesh/sensor_client/src/model_handler.c
Balaklaka: Added

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -240,24 +329,26 @@ static struct bt_mesh_sensor *const sensors[] = {
...
+		.val1 = 1,
 	};
 
 	err = bt_mesh_sensor_srv_pub(&sensor_srv, NULL, &presence_sensor, &val);
 
 	if (err) {
 		printk("Error publishing end of presence (%d)\n", err);
 	}
+
+	prev_pres = 0;
```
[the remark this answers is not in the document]
Balaklaka: When detection is done it will report 0 as it should do. So it need to clear the prev_detection.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
omkar3141: I think, the logic to compute time since presence detected is still incorrect. Let's imagine following sequence of actions:
1. Send Sensor Get to fetch 'Time since presence detected'. Received response will be 0 (or 0xFFFF).
2. Wait for 60 seconds or more.
2. Press and hold button to detect the presence. 
3. When detection delayed work triggers (i.e. successful presence detection), the "[prev_detect = 0](https://github.com/nrfconnect/sdk-nrf/pull/10943/files#diff-7fd712d674be1168023ebf7ca743f337b1cf32f4aca217e754b1abd972175c96R351)" here will cause old timestamp to be erased. 
4. Again, quickly send Sensor Get to fetch 'Time since presence detected'. Now, you should something less than 60 seconds, however, the "[prev_detect = 0](https://github.com/nrfconnect/sdk-nrf/pull/10943/files#diff-7fd712d674be1168023ebf7ca743f337b1cf32f4aca217e754b1abd972175c96R351)" will cause reported time to be from the beginning of the system boot (>60 seconds), which will be obviously incorrect.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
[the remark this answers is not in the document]
omkar3141: Umm, why do you think so? When detection is done, it should keep reporting the time since last known presence. That is what the property is.  

> 3.9.6 Time Since Presence Detected property 
This property represents the time that has elapsed since the sensor last detected presence. This value can, for example, be used to turn off lights in the absence of any activity in a space.

omkar3141 (changes requested):
I think reporting of 'Time Since Presence Detected' is incorrect.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
Balaklaka: This will work like this:
1. Send Sensor Get to fetch "Time since presence detected". Received response will be x seconds since last detection.
2. Press and hold button to detected the presence. Detection presence.
3. Wait for 60 seconds.
4. Send Sensor Get to fetch "Time since presence detected". Received response will be 0 seconds since last detection as presence is active.
5. Wait for 60 seconds.
6. Release button to represent no presence presence. No presence.
7. Send Sensor Get to fetch "Time since presence detected". Received response will be x seconds since last detection.

The question is last present running from when presence is started detected, or when presence is no longer active. I would assume when the "Time since present detected" is from when presence is no longer present. If not then the time for something moving will be added to the time, and that is a little bit strange to add.
I may give the effect that if someplace has a movement over time and "Time since presence detected" is ex. 5 min it will turn of the light, instead of after 5 min without movement.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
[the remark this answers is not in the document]
Balaklaka: This will work like this:

1. Send Sensor Get to fetch "Time since presence detected". Received response will be x seconds since last detection.
2. Press and hold button to detected the presence. Detection presence.
3. Wait for 60 seconds.
4. Send Sensor Get to fetch "Time since presence detected". Received response will be 0 seconds since last detection as presence is active.
5. Wait for 60 seconds.
6. Release button to represent no presence presence. No presence.
7. Send Sensor Get to fetch "Time since presence detected". Received response will be x seconds since last detection.

The question is last present running from when presence is started detected, or when presence is no longer active. I would assume when the "Time since present detected" is from when presence is no longer present. If not then the time for something moving will be added to the time, and that is a little bit strange to add.
I may give the effect that if someplace has a movement over time and "Time since presence detected" is ex. 5 min it will turn of the light, instead of after 5 min without movement.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
Balaklaka: It will be 0 as long as presence is active, last presence is 0 as long as it is in presence detected. Time should be set when detection is no longer detected. See scenario bellow.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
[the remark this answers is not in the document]
omkar3141: Ok, this looks fine then.

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -165,6 +176,17 @@ static void button_handler_cb(uint32_t pressed, uint32_t changed)
 			printk("Error getting sensor descriptor (%d)\n", err);
 		}
 	}
+	if (pressed & BIT(3)) {
```
ludvigsj, line 179: Not really related to your PR, but this will fire if I hold down button 3 while pressing a different button, since this cb fires because of the unrelated press and because it's held down `pressed & BIT(3)` is `true`. All of these should be `if(changed & pressed & BIT(n))`.

ludvigsj (changes requested):

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -240,24 +328,26 @@ static struct bt_mesh_sensor *const sensors[] = {
...
+		.val1 = 1,
 	};
 
 	err = bt_mesh_sensor_srv_pub(&sensor_srv, NULL, &presence_sensor, &val);
 
 	if (err) {
 		printk("Error publishing end of presence (%d)\n", err);
 	}
+
+	pres_detect = 1;
```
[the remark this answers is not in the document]
ludvigsj: Oh, sorry, I see this is handled in the getter. False alarm.

samples/bluetooth/mesh/sensor_client/src/model_handler.c
omkar3141: I assumed that's a feature. One can do parallel button presses.

omkar3141 (approved):

samples/bluetooth/mesh/sensor_client/src/model_handler.c
ludvigsj: No, you misunderstand. The way it works now, every button which is down will trigger the routine once _any_ of them changes. If three buttons are held down, all four messages will be sent when I press or relase the final one.

ludvigsj (approved):

PR #10944, opened 2023-04-25
drivers: wifi: Improve UDP Rx throughput

krish2718:
Using co-operative threads improves UDP RX throughput by 10%, this is needed to pass a certification case.


rlubos (approved):

krish2718:
This needs more experiments and the default is good enough, one can always override this in the samples/applications.

PR #10947, opened 2023-04-26
lib: lte_lc: Add support for CE level and no IMEI modem events

jtguggedal:
Add support to enable modem domain events of both warning and informational level.
Add parsing of the modem domain events for PRACH CE-level and "no IMEI".


jtguggedal:
Put into draft state because `AT%MDMEV=2` is not supported by mfw <= v1.3.4, so need to solve that first.

lib/lte_link_control/lte_lc_helpers.h
```diff
@@ -141,14 +141,19 @@
...
-#define AT_MDMEV_ENABLE				"AT%%MDMEV=1"
+#define AT_MDMEV_ENABLE				"AT%%MDMEV=2"
 #define AT_MDMEV_DISABLE			"AT%%MDMEV=0"
 #define AT_MDMEV_RESPONSE_PREFIX		"%MDMEV: "
 #define AT_MDMEV_OVERHEATED			"ME OVERHEATED\r\n"
 #define AT_MDMEV_BATTERY_LOW			"ME BATTERY LOW\r\n"
 #define AT_MDMEV_SEARCH_STATUS_1		"SEARCH STATUS 1\r\n"
 #define AT_MDMEV_SEARCH_STATUS_2		"SEARCH STATUS 2\r\n"
 #define AT_MDMEV_RESET_LOOP			"RESET LOOP\r\n"
+#define AT_MDMEV_NO_IMEI			"NO IMEI\r\n"
```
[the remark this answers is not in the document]
jtguggedal: There was a reason to include them at some point. I wonder if it was to make sure that the entire response was actually parsed, and not just partially. Maybe it's not valid anymore. Can give it a go in a separate PR.

lib/lte_link_control/lte_lc_helpers.h
```diff
@@ -141,14 +141,19 @@
 #define AT_CONEVAL_DL_PATHLOSS_INDEX		17
 
 /* MDMEV command parameters */
-#define AT_MDMEV_ENABLE				"AT%%MDMEV=1"
+#define AT_MDMEV_ENABLE				"AT%%MDMEV=2"
```
jtguggedal, line 144: @trantanen @tokangas Any preference on how to handle this, where mfw <= 1.3.x doesn't support type 2? Should we just take the simplest route and try `2` first, if that fails, try `1` in all cases?

tokangas: @jtguggedal I think it's fine to always try `2` first and then fall back to `1` in case the first attempt fails.

jtguggedal: Alright, I went for that

lib/lte_link_control/lte_lc.c
```diff
@@ -1612,7 +1612,17 @@ int lte_lc_conn_eval_params_get(struct lte_lc_conn_eval_params *params)
 
 int lte_lc_modem_events_enable(void)
 {
-	return nrf_modem_at_printf(AT_MDMEV_ENABLE) ? -EFAULT : 0;
+	/* First try to enable both warning and informational type events, which is only supported
+	 * by mfw > v1.3.x.
```
trantanen, line 1616: We have used "modem firmware versions" rather than "mfw" in many places.
And I think it would be better to say `>=` than `>` especially because > 1.3.x is a bit harder to understand as people could start to look 1.4.x etc. while it's only really supported in 2.0.0.

```suggestion
	 * by modem firmware versions >= 2.0.0.
```

simensrostad (approved):

tokangas (approved):

lib/lte_link_control/lte_lc.c
tokangas: I agree, it's clearer to say that it's supported by `>= 2.0.0`. Or another possibility is to remove the version number completely, I'm not sure it's mandatory to say that in code comments. It could be just `newer modem firmware versions`, IMHO.

include/modem/lte_lc.h
```diff
@@ -504,6 +504,21 @@ enum lte_lc_modem_evt {
...
+	LTE_LC_MODEM_EVT_CE_LEVEL_0,
+
+	/** Selected CE-level in RACH procedure is 1, see 3GPP TS 36.331 for details. */
+	LTE_LC_MODEM_EVT_CE_LEVEL_1,
+
+	/** Selected CE-level in RACH procedure is 2, see 3GPP TS 36.331 for details. */
+	LTE_LC_MODEM_EVT_CE_LEVEL_2,
+
+	/** Selected CE-level in RACH procedure is 3, see 3GPP TS 36.331 for details. */
+	LTE_LC_MODEM_EVT_CE_LEVEL_3,
```
tokangas, line 521: Should we say that these are only supported by modem firmware version >= 2.0.0?

lib/lte_link_control/lte_lc.c
trantanen: But I don't see what harm exact version number brings when we know it clearly.

trantanen: Fixed

include/modem/lte_lc.h
trantanen: Added

trantanen (approved):

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -508,6 +508,7 @@ Modem libraries
       Modem firmware version v1.3.4 or newer is required to receive these events.
     * The Kconfig option :kconfig:option:`CONFIG_LTE_AUTO_INIT_AND_CONNECT` is now deprecated.
       The application calls the :c:func:`lte_lc_init_and_connect` function instead.
+    * New events added to enumeration :c:enum:`lte_lc_modem_evt` for RACH CE-levels and no IMEI case.
```
peknis, line 511: Would **missing IMEI** be wrong (instead of no IMEI)?
And I'm pretty sure CE level would not need the hyphen.

peknis (changes requested):

doc/nrf/releases/release-notes-changelog.rst
trantanen: Fixed

peknis (approved):

PR #10956, opened 2023-04-26
applications: nrf5340_audio: Get SDU size properly

rick1082:
Use API to get SDU size in stream_recv_cb() would
consume too much time and cause stream receving
not work properly. SDU size should get from the
existed structure after stream is configured.

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/le_audio_cis_headset.c
+++ b/applications/nrf5340_audio/src/bluetooth/le_audio_cis_headset.c
@@ -445,7 +445,7 @@ static void stream_recv_cb(struct bt_bap_stream *stream, const struct bt_iso_rec
 		bad_frame = true;
 	}
 
-	uint32_t octets_per_frame = bt_codec_cfg_get_octets_per_frame(stream->codec);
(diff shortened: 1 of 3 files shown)


alexsven (approved):

applications/nrf5340_audio/src/bluetooth/le_audio_cis_headset.c
```diff
@@ -445,7 +445,7 @@ static void stream_recv_cb(struct bt_bap_stream *stream, const struct bt_iso_rec
 		bad_frame = true;
 	}
 
-	uint32_t octets_per_frame = bt_codec_cfg_get_octets_per_frame(stream->codec);
+	uint32_t octets_per_frame = stream->qos->sdu;
```
alexsven, line 448: I think we should do similar to what we do in BIS, store the configuration at the start and then check against that all the way. That way we are also guarded against the stream suddenly changing without us knowing it. One more thing is that sdu might be more than one frame, so octets_per_frame might not be the correct name.

alexsven (changes requested):

PR #10959, opened 2023-04-26
scripts: hid_configurator: Add MCUboot direct-xip mode support

MarekPieta:
Change introduces support for MCUboot bootloader built in direct-xip mode.

Jira: NCSDK-20586

### DIFF
--- a/scripts/hid_configurator/modules/dfu.py
+++ b/scripts/hid_configurator/modules/dfu.py
@@ -201,7 +201,7 @@ def b0_get_dfu_image_bootloader_var():
     return 'B0'
 
 
-def mcuboot_is_dfu_file_correct(dfu_bin):
+def mcuboot_common_is_dfu_file_correct(dfu_bin):
     res, _, _ = imgtool.image.Image.verify(dfu_bin, None)
 
     if res != imgtool.image.VerifyResult.OK:
@@ -211,11 +211,7 @@ def mcuboot_is_dfu_file_correct(dfu_bin):
     return True
 
 
-def mcuboot_get_dfu_image_name(dfu_slot_id):
-    return 'app_update.bin'
-
-
-def mcuboot_get_dfu_image_version(dfu_bin):
+def mcuboot_common_get_dfu_image_version(dfu_bin):
     res, ver, _ = imgtool.image.Image.verify(dfu_bin, None)
 
     if res != imgtool.image.VerifyResult.OK:
@@ -225,8 +221,25 @@ def mcuboot_get_dfu_image_version(dfu_bin):
     return ver
 
 
+def mcuboot_get_dfu_image_name(dfu_slot_id):
+    return 'app_update.bin'
+
+
 def mcuboot_get_dfu_image_bootloader_var():
     return 'MCUBOOT'
+
+
+def mcuboot_xip_get_dfu_image_name(dfu_slot_id):
+    assert dfu_slot_id in (0, 1)
+
+    if dfu_slot_id == 0:
+        return 'app_update.bin'
+    else:
+        return 'mcuboot_secondary_app_update.bin'
+
+
+def mcuboot_xip_get_dfu_image_bootloader_var():
+    return 'MCUBOOT+XIP'
 
 
 B0_API = {
@@ -237,14 +250,22 @@ B0_API = {
 }
 
 MCUBOOT_API = {
-    'get_dfu_image_version' : mcuboot_get_dfu_image_version,
+    'get_dfu_image_version' : mcuboot_common_get_dfu_image_version,
     'get_dfu_image_bootloader_var' : mcuboot_get_dfu_image_bootloader_var,
     'get_dfu_image_name' : mcuboot_get_dfu_image_name,
-    'is_dfu_file_correct' : mcuboot_is_dfu_file_correct,
+    'is_dfu_file_correct' : mcuboot_common_is_dfu_file_correct,
+}
+
+MCUBOOT_XIP_API = {
+    'get_dfu_image_version' : mcuboot_common_get_dfu_image_version,
+    'get_dfu_image_bootloader_var' : mcuboot_xip_get_dfu_image_bootloader_var,
+    'get_dfu_image_name' : mcuboot_xip_get_dfu_image_name,
+    'is_dfu_file_correct' : mcuboot_common_is_dfu_file_correct,
 }
 
 BOOTLOADER_APIS = {
     'MCUBOOT' : MCUBOOT_API,
+    'MCUBOOT+XIP' : MCUBOOT_XIP_API,
     'B0' : B0_API,
 }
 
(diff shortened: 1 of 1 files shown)


scripts/hid_configurator/modules/dfu.py
```diff
@@ -229,6 +229,19 @@ def mcuboot_get_dfu_image_bootloader_var():
     return 'MCUBOOT'
```
mkapala-nordic, line 229: consider grouping firstly `mcuboot_common_*` functions together and later `mcuboot_*` so they won't intertwine with each other

scripts/hid_configurator/modules/dfu.py
```diff
@@ -292,16 +313,22 @@ def __init__(self, dfu_package, dev_fwinfo, dev_board_name, dev_bootloader_varia
...
+            VERSION_PREFIX = "version_"
+
+            version_keys = tuple(filter(lambda x: x.startswith(VERSION_PREFIX), f.keys()))
+            if len(version_keys) != 1:
+                return None
+
+            temp_bootloader_name = version_keys[0][len(VERSION_PREFIX):]
+
+            if (bootloader_name is not None) and (temp_bootloader_name != bootloader_name):
+                return None
```
kapi-no, line 329: same here

scripts/hid_configurator/modules/dfu.py
```diff
@@ -292,16 +313,22 @@ def __init__(self, dfu_package, dev_fwinfo, dev_board_name, dev_bootloader_varia
...
+
+        for f in manifest_json["files"]:
+            VERSION_PREFIX = "version_"
+
+            version_keys = tuple(filter(lambda x: x.startswith(VERSION_PREFIX), f.keys()))
+            if len(version_keys) != 1:
+                return None
+
+            temp_bootloader_name = version_keys[0][len(VERSION_PREFIX):]
+
```
kapi-no, line 327: consider removing the blank line as you use directly use this variable in the if statement

kapi-no (approved):

mkapala-nordic (approved):

scripts/hid_configurator/modules/dfu.py
```diff
@@ -225,10 +221,27 @@ def mcuboot_get_dfu_image_version(dfu_bin):
...
 def mcuboot_get_dfu_image_bootloader_var():
     return 'MCUBOOT'
 
 
+def mcuboot_xip_get_dfu_image_name(dfu_slot_id):
+    assert dfu_slot_id in (0, 1)
+
+    if dfu_slot_id == 0:
+        return 'app_update.bin'
```
nordicjm, line 236: @tejlmand this looks problematic for sysbuild

MarekPieta: Hi @nordicjm
Do you mean that the `dfu_application.zip` content is going to change ?

nordicjm: yes, filenames will change and they might be put in folders

MarekPieta: Modifying content of `dfu_application.zip` would actually break backwards compatibility of existing DFU mechanisms. Can we somehow ensure backwards compatibility of the zip file content?

nordicjm: DFU should not be assuming anything of the .bin files, that's why there is a manifest in the zip. If something assumes a specific bin file, that script is wrong.

MarekPieta: Can we rely on the order of elements in `files` to determine the used slot then? E.g. `B0` bootloader's update files have no `slot` property in the manifest.

nordicjm: Is there a sample application in tree that builds an application that uses this?

MarekPieta: `applications/nrf_desktop` uses B0 bootloader in many configurations

tejlmand: > @tejlmand this looks problematic for sysbuild

I believe this `dfu.py` is only used by [hid_configurator](https://github.com/nrfconnect/sdk-nrf/tree/main/scripts/hid_configurator) and as such is outside scope of the build system.

That said, I agree that depending on hard coded names / fixed knowledge about zip file content is a bad idea.
Instead such information should be obtained through manifest files, and if such info is missing there I would say it should be fixed by extending the manifest fields.

But that's just my personal opinion based on experience.

nordicjm: So the output look like this:
```
{
    "format-version": 0,
    "time": 1682684519,
    "files": [
        {
            "version_B0": "1",
            "type": "application",
            "board": "nrf52840dk_nrf52840",
            "soc": "nRF52840_QIAA",
            "load_address": 24576,
            "size": 351304,
            "file": "signed_by_b0_s0_image.bin",
            "modtime": 1682684518
        },
        { 
            "version_B0": "1",
            "type": "application",
            "board": "nrf52840dk_nrf52840",
            "soc": "nRF52840_QIAA",
            "load_address": 532480,
            "size": 351304,
            "file": "signed_by_b0_s1_image.bin",
            "modtime": 1682684519
        }
    ],
    "name": "nrf_desktop"
}
```
I think we're going to need to change and improve that to include e.g. a slot number, and image number, because relying on a specific filename is not sustainable going forwards.

MarekPieta: I understand that using hardcoded file names may not be perfect. On the other hand, there used to be no other information in the manifest file previously, so that customers may also already rely on these names or file order (e.g. our `nordic-hid` plugin in the  `LVFS fwupd` external DFU tool already relies on file order in the manifest file).

MarekPieta: Also, please keep in mind that even if we introduce some new fields to the `manifest,json`, the script shall be backwards compatible (it must be capable of using old firmware packages). We could improve script to rely on the new values if available, but if these values are not present, the script would need to have some fallback mechanism (relying on filenames or element order in `files` array).

nordicjm: Filenames will all be `<CONFIG_KERNEL_BIN_NAME>.signed.bin` where `CONFIG_KERNEL_BIN_NAME` is `zephyr` by default, and can be changed on a per-image basis.

MarekPieta: I agree that scripts would need to be aligned at this point (keeping backwards compatibility in mind).

MarekPieta: Hi @nordicjm
Could we desist from using filenames in a separate PR (and merge this one)? I would open o follow-up Jira ticket for this one to make sure we do not forget about it.
Lack of the change block us from using `MCUboot+XIP` in our Fast Pair configurations. The filename based approach was already used for `B0` and `MCUboot` with swap (so I would say we do not introduce new issues here). Changing approach in the script could require some bigger changes (we may also need to agree on new manifest content for B0 bootloader and update it since currently it lacks needed information).

nordicjm: The comments here are not blocking this PR and outside the scope of this PR, no problem with merging this PR just wanted to bring attention to this for sysbuild as I was not aware of it and assume neither was Torsten.

PR #10968, opened 2023-04-27
samples: matter: Aligned RESET_ON_FATAL_ERROR for app and net cores.

kkasperczyk-no:
Set the CONFIG_RESET_ON_FATAL_ERROR value for hci_rpmsg and multiprotocol_rpmsg images in a consistent way with value set on the application core.


doublemis1:
I think that we don't want to define that on release configuration,
I thought that we want to have debug configuration which has halt on assert while for release we would like to restart device on crash.
However for debug configuration there was missing config for network core.

Damian-Nordic (approved):

LuDuda (approved):

doublemis1 (approved):

PR #10981, opened 2023-04-28
drivers: wifi: Add support to dump RPU stats

krish2718:
Add a nRF700x utility shell command to dump all or individual RPU module statistics, quite handy in debug.


rado17 (approved):

rado17 (approved):

sachinthegreen (commented):
Please hold off on this till a debug shell is created (happening as part of another PR soon). We can have this as part of that shell.

srinivas1201 (approved):

srinivas1201 (approved):

krish2718:
> Please hold off on this till a debug shell is created (happening as part of another PR soon). We can have this as part of that shell.

This is needed for WFA cert, so, I would suggest to merge this and we can easily move around once debug shell is in.

sachinthegreen (approved):

PR #10982, opened 2023-04-28
tests: lib: edge_impulse: Add missing dependencies

rakons:
This commit adds missing dependencies for edge impulse testing.

Temporary - DNM. Testing if this works. I cannot recreate the issue on local machine.

### DIFF
--- a/tests/lib/edge_impulse/CMakeLists.txt
+++ b/tests/lib/edge_impulse/CMakeLists.txt
@@ -35,3 +35,5 @@ target_sources(app PRIVATE src/main.cpp)
 # Test uses ei_test_params.h file from edge_impuse_zip directory to verify if
 # ei_wrapper properly forwards the data between application and EI library.
 target_include_directories(app PRIVATE src/edge_impulse_zip/)
+
+add_dependencies(app edge_impulse edge_impulse_project offsets)


tests/lib/edge_impulse/CMakeLists.txt
```diff
@@ -35,3 +35,5 @@ target_sources(app PRIVATE src/main.cpp)
 # Test uses ei_test_params.h file from edge_impuse_zip directory to verify if
 # ei_wrapper properly forwards the data between application and EI library.
 target_include_directories(app PRIVATE src/edge_impulse_zip/)
+
+add_dependencies(app edge_impulse edge_impulse_project offsets)
```
MarekPieta, line 39: Shouldn't it be `add_dependencies(edge_impulse_project offsets)`? So that `edge_impulse_project` depends on `offsets`. It seems that actually `edge_impulse_project` build fails, because of failing to include the `offsets.h` header

MarekPieta: Maybe we could even use `add_dependencies(edge_impulse_project zephyr_generated_headers)` (so that we get dependency on both offsets and version). Since that dependency seems needed also for zip files downloaded from EI studio (the EI SDK includes `version.h`), I would consider adding it under `ExternalProject_Add(edge_impulse_project` in the `nrf/lib/edge_impulse/CMakeLists.txt` file.

MarekPieta (changes requested):

tests/lib/edge_impulse/CMakeLists.txt
rakons: This is not a good idea. The problem is, when compiling test, under /tests/lib/edge_impulse/src/edge_impulse_zip/edge-impulse-sdk/cmake/zephyr/CMakeLists.txt we have a line:
```
target_sources(app PRIVATE ${EI_SDK_FOLDER}/ei_run_classifier_mock.cpp)
```

And this file gives us trouble when compiling.

rakons:
It seems that the issue is solved by https://github.com/nrfconnect/sdk-nrf/pull/11043 .
Currently the solution presented here has been proven not to work - it failed once on local build.

PR #10983, opened 2023-04-28
samples: nrf9160: modem_shell: fix LwM2M initialization

tokangas:
There was a race condition with nRF modem library and LwM2M initialization, sometimes causing the reading of IMEI to fail. `cloud_lwm2m_init()` is now called using `NRF_MODEM_LIB_ON_INIT()` to make sure modem library has been initialized before trying to read the IMEI.

### DIFF
--- a/samples/nrf9160/modem_shell/src/cloud/cloud_lwm2m.c
+++ b/samples/nrf9160/modem_shell/src/cloud/cloud_lwm2m.c
@@ -6,6 +6,7 @@
 
 #include <zephyr/kernel.h>
 #include <zephyr/device.h>
+#include <modem/nrf_modem_lib.h>
 #include <modem/modem_info.h>
 #include <lwm2m_object.h>
 #include <net/lwm2m_client_utils.h>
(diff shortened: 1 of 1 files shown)


jhirsi (approved):

samples/nrf9160/modem_shell/src/cloud/cloud_lwm2m.c
```diff
@@ -163,14 +175,12 @@ static int cloud_lwm2m_init(const struct device *unused)
 	ret = modem_info_init();
 	if (ret < 0) {
 		printk("LwM2M: Unable to init modem_info (%d)", ret);
-		return ret;
```
trantanen, line 166: How does the code work if modem_info_init() fails and we request IMEI then?

samples/nrf9160/modem_shell/src/cloud/cloud_lwm2m.c
```diff
@@ -163,14 +175,12 @@ static int cloud_lwm2m_init(const struct device *unused)
...
 	}
 
 	/* Query IMEI. */
 	ret = modem_info_string_get(MODEM_INFO_IMEI, imei_buf, sizeof(imei_buf));
 	if (ret < 0) {
 		printk("LwM2M: Unable to get IMEI");
-		return ret;
```
trantanen, line 173: How does the code work if we don't get IMEI and we use it in endpoint_name?

trantanen (commented):
Asking because I don't know how lwm2m should be constructed.

samples/nrf9160/modem_shell/src/cloud/cloud_lwm2m.c
tokangas: And same here.

trantanen (approved):

tokangas:
Need to wait for upmerge to be finished because of a merge conflict.

PR #10986, opened 2023-04-28
cmake: multi_image: Preload variant images with CONF_FILE_BUILD_TYPE

MarekPieta:
If 'CACHED_CONF_FILE' is specified instead of 'CONF_FILE', the build system does not determine build type automatically. In that case, the 'CONF_FILE_BUILD_TYPE' shall be passed explicitly to prevent Partition Manager build issues.

Jira: NCSDK-21357

### DIFF
--- a/cmake/multi_image.cmake
+++ b/cmake/multi_image.cmake
@@ -53,6 +53,12 @@ else()
       list(APPEND application_vars ${var_name})
     endif()
 
+    # If 'CACHED_CONF_FILE' is specified instead of 'CONF_FILE', the build system does not determine
+    # build type automatically. In that case, the 'CONF_FILE_BUILD_TYPE' shall be passed explicitly.
+    if("${var_name}" MATCHES "CONF_FILE_BUILD_TYPE")
+      list(APPEND application_vars ${var_name})
+    endif()
+
     # '-DDTC_OVERLAY_FILE' is given helptext by the build system. Therefore
     # we need this special handling for passing the value to the variant image.
     if("${var_name}" MATCHES "DTC_OVERLAY_FILE")


cmake/multi_image.cmake
```diff
@@ -53,6 +53,12 @@ else()
       list(APPEND application_vars ${var_name})
     endif()
 
+    # If 'CACHED_CONF_FILE' is specified instead of 'CONF_FILE', the build system does not determine
+    # build type automatically. In that case, the 'CONF_FILE_BUILD_TYPE' shall be passed explicitly.
```
tejlmand, line 57: This seems like improper fix.

Already this makes me specious: 
https://github.com/nrfconnect/sdk-nrf/blob/800bd44488952f6f8a85b6b2a6224afc4f8e3430/cmake/multi_image.cmake#L49-L54

as we should not pass down internal CMake cache vars like this, instead this should probably be passed down using `CONF_FILE`.

However, this proposal indicates that upstream itself has a problem wrt. `CONF_FILE_BUILD_TYPE`, as this value is not being set on CMake re-runs, which could result in unexpected behavior.

tejlmand (changes requested):
blocking, as this is not a proper fix for this.

Instead there seems to be an issue which must be handled upstream.

cmake/multi_image.cmake
MarekPieta: It seems that we cannot access the `CONF_FILE` at this point.
If I understand correctly, passing `CACHED_CONF_FILE` as `CONF_FILE` here may result in passing multiple Kconfig configuration files which would also prevent the build system from properly detecting the build type. @tejlmand do you have an idea how can we fix this in a proper way?

tejlmand: > If I understand correctly, passing CACHED_CONF_FILE as CONF_FILE here may result in passing multiple Kconfig configuration files which would also prevent the build system from properly detecting the build type.

correct, so the solution here is actually non-trivial.

tejlmand: seems this PR https://github.com/nrfconnect/sdk-nrf/pull/6890/ introduced caching of build type.
A little concerned we will keep having unwanted, but undiscovered, behavior wrt. build type until this is better handled upstream.

Approving to allow us to move forward with NCS and because I have not discovered unexpected side-effects, but this should be fixed upstream and then we could cleanup a lot of this special handling.

tejlmand (approved):
Approving to allow us to move forward with NCS and because I have not discovered unexpected side-effects, but this should be fixed upstream and then we could cleanup a lot of this special handling.

kapi-no (approved):

MarekPieta:
Thanks a lot for the review @tejlmand

alstrzebonski (approved):

PR #10991, opened 2023-05-01
TF-M: Print memory usage of TF-M executable

joerchan:


frkv (approved):
LGTM

modules/tfm/tfm/boards/CMakeLists.txt
```diff
@@ -179,3 +179,8 @@ if(LOG_MEMORY_PROTECTION)
   src/log_memory_protection.c
   )
 endif()
+
+target_link_options(tfm_s
+  PRIVATE
+    $<$<C_COMPILER_ID:GNU>:-Wl,--print-memory-usage>
+)
```
SebastianBoe, line 186: Strange.

Why isn't this working:

ncs/modules/tee/tf-m/trusted-firmware-m/toolchain_GNUARM.cmake
61:    set(MEMORY_USAGE_FLAG LINKER:--print-memory-usage)

?

SebastianBoe (changes requested):
We should find out why upstream's attempt at doing the same thing is not working.

joerchan:
> We should find out why upstream's attempt at doing the same thing is not working.

I didn't know about that one, CONFIG_TFM_MEMORY_USAGE_QUIET is enabled, so that should make sense IMO.
I will try to change this config value instead of adding it like this.

Thanks.

joerchan:
Close in favour of: https://github.com/zephyrproject-rtos/zephyr/pull/57487

PR #11011, opened 2023-05-03
tfm_(psa|regression)_test: Add tags to each test

Mangl10:


tests/tfm/tfm_psa_test/testcase.yaml
```diff
@@ -13,28 +13,38 @@ common:
       - "TOTAL FAILED *: 0"
 tests:
   tfm.psa_test_protected_storage_lvl1:
+    tag: lvl1
```
torsteingrindvik, line 16: I think it's `tags` not `tag`

Mangl10: Yes, missed that.

torsteingrindvik (approved):

joerchan (changes requested):
If this is a generic tag in global scope taking lvl1/2 for this seems not very future proof. I suggges tfm_lvl1/2

gmarull:
please user forks, branches in upstream repo are not allowed.

PR #11020, opened 2023-05-04
samples: use a separate random per peer

Tschet1:
If multiple devices initiate a ranging to the same peer at the same time, this could result in the attempts interfering with each other.

This can for example happen if an advertiser is scanned by multiple devices at the same time. In that case, all scanners receive the scan response from the advertiser and start a ranging at the same time.

Added a per-peer random based on its address, such that this should not happen. The random is passed as input to the ranging library. This makes sure that only the pair of devices that have the same random can range with each other. This should limit the interference from other devices trying to range to the same device at the same time.


doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -264,6 +264,10 @@ Bluetooth samples
 
   * Fixed the unit of the :kconfig:option:`CONFIG_BT_NUS_UART_RX_WAIT_TIME` Kconfig option to comply with the UART API.
 
+* :ref:`nrf_dm` sample:
+
+  * Improved scalability of the sample when it is being used with more devices.
```
wiba-nordic, line 269:
```suggestion
  * Improved the scalability of the sample when it is being used with more devices.
```

wiba-nordic (changes requested):

samples/bluetooth/nrf_dm/src/main.c
```diff
@@ -104,7 +109,18 @@ static bool data_cb(struct bt_data *data, void *user_data)
...
+			 * A scan response that is received by this device can be received by
+			 * multiple other devices which can all start a ranging at the same time
+			 * as a consequence. To prevent this, we need to make sure that we set a
+			 * per-peer random as the random seed. This helps the ranging library to
+			 * avoid interference from other devices trying to range at the same time.
+			 *
+			 * This means that the initiator and the reflector need to set the same value
+			 * for the random seed.
+			 */
+			req.rng_seed =
```
eriksandgren, line 122: This could still cause multiple devices to initiate a ranging right? The difference is that it shall only be successful with one of them. 

Did you consider / would it be possible to send the addr of the scanner in the scan response, and then just initiate the ranging if this addr matches your own?

wiba-nordic (approved):

samples/bluetooth/nrf_dm/src/main.c
eriksandgren: Ok, thanks for explaining.

Tschet1:
> How did you test these changes?
> 
> I guess we should at least make sure that ranging multiple peers still work

tested this with 4 devices on my desk

eriksandgren (approved):

alwa-nordic (approved):

maje-emb (approved):

samples/bluetooth/nrf_dm/src/main.c
```diff
@@ -116,6 +132,25 @@ static bool data_cb(struct bt_data *data, void *user_data)
...
+	size_t count = CONFIG_BT_ID_MAX;
+
+	bt_id_get(addrs, &count);
+
+	__ASSERT(count == 1, "The sample assumes a single ID addr");
+
+	return scanner_addr_to_random_share(&addrs[0].a);
+}
+
+static uint32_t scanner_addr_to_random_share(bt_addr_t *p_scanner_addr)
```
KAGA164, line 147: Can it be const?
```suggestion
static uint32_t scanner_addr_to_random_share(const bt_addr_t *p_scanner_addr)
```

samples/bluetooth/nrf_dm/src/main.c
```diff
@@ -116,6 +132,25 @@ static bool data_cb(struct bt_data *data, void *user_data)
...
+{
+	return (p_scanner_addr->val[0] | p_scanner_addr->val[1] << 8 |
```
KAGA164, line 149: Just a question, I am not an expert here. Is it some kind of known hashing algorithm used here? I worried whether it gets a unique result for different BLE addresses

Tschet1: It does not need to be unique. But there is indeed an issue with the code here. We should not use OR.

Tschet1: actually. sorry. I think what we have is correct.
We don't need to guarantee uniqueness. We only need to make sure to create a uint32 from the scanner address.

KAGA164 (approved):

PR #11022, opened 2023-05-04
samples: mqtt: Fix modem initialization and enable RSA for Wi-Fi

jtguggedal:
The test server certificates are signed using RSA, so RSA must
be enabled in mbedTLS when building for nRF7002 DK.

Modem initialization at boot time was deprecated, so the application
was started without modem initialized, causing a reboot loop.

Fixed by adding modem initialization in the network thread.
In addition, readout of IMEI and credential provisioning had
to be moved to be certain that it happens after modem is on.


jtguggedal:
No changelog needed, this is fixing regressions

dilner (approved):

rlubos (approved):

PR #11027, opened 2023-05-04
applications: asset_tracker_v2: Notify if fuel gauge is not supported

jtguggedal:
Introduce the event `SENSOR_EVT_FUEL_GAUGE_NOT_SUPPORTED` that is sent when a board does not have a fuel gauge.

Fixes NCSDK-21405


applications/asset_tracker_v2/src/modules/sensor_module.c
```diff
@@ -263,8 +264,16 @@ static void battery_data_get(void)
 	sensor_module_event->data.bat.timestamp = k_uptime_get();
 	sensor_module_event->data.bat.battery_level = percentage;
 	sensor_module_event->type = SENSOR_EVT_FUEL_GAUGE_READY;
-	APP_EVENT_SUBMIT(sensor_module_event);
+#else
+	sensor_module_event = new_sensor_module_event();
+
+	if (!sensor_module_event) {
+		SEND_ERROR(sensor, SENSOR_EVT_ERROR, -ENOMEM);
```
trantanen, line 271: Is there a reason why this is here but not in the `#if defined(CONFIG_ADP536X)` branch?

applications/asset_tracker_v2/src/events/sensor_module_event.c
```diff
@@ -20,10 +20,10 @@ static char *get_evt_type_str(enum sensor_module_event_type type)
 		return "SENSOR_EVT_MOVEMENT_IMPACT_DETECTED";
 	case SENSOR_EVT_ENVIRONMENTAL_DATA_READY:
 		return "SENSOR_EVT_ENVIRONMENTAL_DATA_READY";
-	case SENSOR_EVT_FUEL_GAUGE_READY:
-		return "SENSOR_EVT_FUEL_GAUGE_READY";
```
trantanen, line 24: I guess this shouldn't be deleted or am I missing something?
Did you intend to move it like you did in `sensor_module_event_type`?

trantanen (changes requested):

applications/asset_tracker_v2/src/modules/sensor_module.c
tokangas: Looking at the `SEND_ERROR()` macro implementation, the first thing it does is to try to allocate a new event by calling `new_sensor_module_event()`. 😃 It does assert that the allocation succeeds, though. So, if asserts are enabled, at least the device would crash with a clearly visible reason.

I'm not sure if the check is needed here, but at least it would be nice to be consistent.

jtguggedal: > Is there a reason why this is here but not in the `#if defined(CONFIG_ADP536X)` branch?

No, that's probably just an overlook. There should be a check, I'll add it.

> Looking at the `SEND_ERROR()` macro implementation, the first thing it does is to try to allocate a new event by calling `new_sensor_module_event()`. 😃 It does assert that the allocation succeeds, though. So, if asserts are enabled, at least the device would crash with a clearly visible reason.
> 
> I'm not sure if the check is needed here, but at least it would be nice to be consistent.

I did it this way to avoid asserting to also catch it when asserts are available, but need to reconsider that.
I'll just switch back to assert and we'll need to revisit this topic later.

applications/asset_tracker_v2/src/modules/sensor_module.c
```diff
@@ -263,8 +264,16 @@ static void battery_data_get(void)
 	sensor_module_event->data.bat.timestamp = k_uptime_get();
 	sensor_module_event->data.bat.battery_level = percentage;
 	sensor_module_event->type = SENSOR_EVT_FUEL_GAUGE_READY;
-	APP_EVENT_SUBMIT(sensor_module_event);
+#else
+	sensor_module_event = new_sensor_module_event();
```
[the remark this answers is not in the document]
jtguggedal: Yeah, that's what I wanted at first, but decided to keep it consistent with environmental data.
Definitely worth considering changing in the future.

applications/asset_tracker_v2/doc/asset_tracker_v2_description.rst
```diff
@@ -191,6 +191,7 @@ After programming the application and all the prerequisites to your development
    <inf> app_event_manager: APP_EVT_DATA_GET - Requested data types (MOD_DYN, BAT, ENV, LOCATION)
    <inf> app_event_manager: LOCATION_MODULE_EVT_ACTIVE
    <inf> app_event_manager: SENSOR_EVT_ENVIRONMENTAL_NOT_SUPPORTED
+   <inf> app_event_manager: SENSOR_EVT_FUEL_GAUGE_NOT_SUPPORTED
    <inf> app_event_manager: MODEM_EVT_MODEM_DYNAMIC_DATA_NOT_READY
    <inf> app_event_manager: SENSOR_EVT_FUEL_GAUGE_READY
```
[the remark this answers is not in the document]
jtguggedal: Nope, fixing

applications/asset_tracker_v2/src/events/sensor_module_event.c
jtguggedal: That was the intention, fixed

jtguggedal:
@trantanen @tokangas I have addressed your comments

simensrostad (approved):

tokangas (approved):

trantanen (approved):

trantanen:
I've rebased this as blockers from main should be solved now.

divipillai (approved):

PR #11030, opened 2023-05-04
applications: serial_lte_modem: Coverity fixes

jarlamsa:
Some minor coverity fixes:
Skipping initialization while still in scope when using goto.
Always true compare.

### DIFF
--- a/applications/serial_lte_modem/src/http_c/slm_at_httpc.c
+++ b/applications/serial_lte_modem/src/http_c/slm_at_httpc.c
@@ -197,16 +197,21 @@ static int payload_cb(int sock, struct http_request *req, void *user_data)
 		k_sem_take(&http_req_sem, K_FOREVER);
 	}
 
-	if (httpc.total_sent >= 0) {
-		httpc.state = HTTPC_REQ_DONE;
-		rsp_send("\r\n#XHTTPCREQ: 0\r\n");
-	}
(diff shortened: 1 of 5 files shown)


applications/serial_lte_modem/src/http_c/slm_at_httpc.c
```diff
@@ -197,16 +197,25 @@ static int payload_cb(int sock, struct http_request *req, void *user_data)
...
+	rsp_send("\r\n#XHTTPCREQ: 0\r\n");
+
 	return httpc.total_sent;
 }
 
 static int do_http_connect(void)
 {
 	int ret;
+	struct sockaddr sa = {
+		.sa_family = AF_UNSPEC
```
MarkusLassila, line 210: How about removing this initialization when you are touching the area? I don't believe that .sa_family is actually utilized before it is overwritten.

MarkusLassila (approved):

VTPeltoketo (approved):

applications/serial_lte_modem/src/http_c/slm_at_httpc.c
jarlamsa: True, `util_resolve_host` just seems to overwrite it. I'll remove those and update the PR after the CI is healthy again.

PR #11037, opened 2023-05-04
manifest: Update zephyr for app dependency on offsets header fix

joerchan:
Update zephyr version:
Fix app dependency on offsets header file generated by the build system.


joerchan:
Closing this as sdk-nrf has another fix, we can wait for the upstream PR to be resolved and brought to NCS in a regular upmerge.

PR #11040, opened 2023-05-04
tests: bluetooth: tester: Set tester default log level to ERROR

m-alperen-sener:
Having all modules' logs enabled by default causes tester to behave unstable. Reducing the default log level from 4(DBG) to 1(ERROR).


PavelVPV (approved):

tests/bluetooth/tester/boards/nrf52840dk_nrf52840.conf
```diff
@@ -17,6 +17,6 @@ CONFIG_SEGGER_RTT_BUFFER_SIZE_UP=4096
 CONFIG_LOG_BACKEND_SHOW_COLOR=n
 CONFIG_LOG_PROCESS_THREAD_STACK_SIZE=1024
 
-CONFIG_LOG_DEFAULT_LEVEL=4
+CONFIG_LOG_DEFAULT_LEVEL=1
```
Balaklaka, line 20: Why do we not want info and warnings?

m-alperen-sener: We don't need them for conformance.

Balaklaka (approved):

PR #11041, opened 2023-05-04
Bl storage fixes

SebastianBoe:
2 commits:

[bl_storage: Refactor to use nrfx_nvmc_uicr_word_read](https://github.com/nrfconnect/sdk-nrf/commit/b46618d8cda651022c933350c4cd70e014108309) 

Use nrfx_nvmc_uicr_word_read to read from UICR/OTP instead of applying
the DSB errata 7 workaround manually.

This improves readability and code re-use.

nrfx_nvmc_uicr_word_read also has a slightly more optimal
implementation as it omits the DSB instruction for unaffected
platforms.

[bl_storage: Use nrfx_nvmc_uicr_word_read when accessing OTP](https://github.com/nrfconnect/sdk-nrf/commit/369116691aa21eed8f5d525ed30168eebcbd5aca) 

Errata 7 for nrf91 requires a DSB instruction after accessing
UICR/OTP.

It's unclear if we can be affected by the errata here but we apply the
workaround anyway just to be safe.

### DIFF
--- a/include/bl_storage.h
+++ b/include/bl_storage.h
@@ -201,14 +201,15 @@ enum lcs {
  * time. Writes to @p dst are done a byte at a time.
  *
  * @param[out] dst destination buffer.
- * @param[in] src source buffer. Must be 4-byte-aligned.
+ * @param[in] src source buffer in OTP. Must be 4-byte-aligned.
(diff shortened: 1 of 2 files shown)


hakonfam (approved):
Nice!

joerchan (approved):

PR #11043, opened 2023-05-04
tests: edge_impulse: Fix missing dependency on zephyr generated headers

joerchan:
Fix missing dependency on zephyr generated headers for the edge_impulse_project.
This target is built as an external project, and cannot depend on zephyr_generated_headers where it is declared, because zephyr_generated_headers is a custom command instead of an interface library.

### DIFF
--- a/lib/edge_impulse/CMakeLists.txt
+++ b/lib/edge_impulse/CMakeLists.txt
@@ -60,6 +60,7 @@ ExternalProject_Add(edge_impulse_project
     BUILD_ALWAYS True
     USES_TERMINAL_BUILD True
 )
+add_dependencies(edge_impulse_project zephyr_generated_headers)
 
 # This targets remove the `edge_impulse_project-download` stamp file, which
 # causes the Edge impulse library to be fetched on each build invocation.


tests/lib/edge_impulse/CMakeLists.txt
```diff
@@ -35,3 +35,5 @@ target_sources(app PRIVATE src/main.cpp)
 # Test uses ei_test_params.h file from edge_impuse_zip directory to verify if
 # ei_wrapper properly forwards the data between application and EI library.
 target_include_directories(app PRIVATE src/edge_impulse_zip/)
+
+add_dependencies(edge_impulse_project zephyr_generated_headers)
```
MarekPieta, line 39: Adding `zephyr_generated_headers` under `DEPENDS` of `ExternalProject_Add(edge_impulse_project` actually seems not to work. It looks like dependency added one line under the ExternalProject_Add invocation in the following way works fine:
`add_dependencies(edge_impulse_project zephyr_generated_headers)`.
That dependency seems needed also for zip files downloaded from EI studio (the EI SDK includes version.h). Can we move this line to `lib/edge_impulse/CMakeLists.txt` ?

MarekPieta (changes requested):

tests/lib/edge_impulse/CMakeLists.txt
joerchan:
```
CMake Error at /home/joakim/repos/ncs/nrf/lib/edge_impulse/CMakeLists.txt:63 (add_dependencies):
  The dependency target "zephyr_generated_headers" of target
  "edge_impulse_project" does not exist.
```
I couldn't make that work.
So this is a temporary solution, I think https://github.com/zephyrproject-rtos/zephyr/pull/57554 is the better solution.
This can be a workaround until we get that fix instead.

MarekPieta: Strange. For me it works. Are you sure you check with correct Zephyr ?

joerchan: Correct, I still had the linked PR where that did not exist.
Using it on external project directly does not work because the target does not exists where it is invoked.

```
Make Error at /home/joakim/.local/lib/python3.8/site-packages/cmake/data/share/cmake-3.26/Modules/ExternalProject.cmake:3496 (get_property):
  get_property could not find TARGET zephyr_generated_headers.  Perhaps it
  has not yet been created
```

MarekPieta (approved):

SebastianBoe (approved):

PR #11045, opened 2023-05-04
asset_tracker_v2: tests: Make json_common use unity instead of ztest

balaji-nordic:
The json_common unit test is now made to use the unity test framework
instead of ztest. This is done to align it with the other unit tests
in the asset_tracker_v2 folder that also use unity instead of ztest.

Replaced all zassert functions with the corresponding TEST_ASSERT 
macros in unity. 

The message parameter of the assert functions were removed as they
were superfluous. If the test fails, it will indicate which assert failed.

Fixed a segmentation fault that occurred because we were 
invoking `cJSON_Delete()` on random pointers.  This is now fixed by  
memsetting the dummy variable to 0 in `tearDown()`, which clears all pointers.
The `cJSON_Delete()` will not try to free the pointer if it is NULL.

### DIFF
--- a/applications/asset_tracker_v2/tests/json_common/src/main.c
+++ b/applications/asset_tracker_v2/tests/json_common/src/main.c
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 
-#include <zephyr/ztest.h>
+#include <unity.h>
 #include <zephyr/kernel.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


applications/asset_tracker_v2/tests/json_common/src/main.c
```diff
@@ -1166,170 +1184,30 @@ static void test_floating_point_encoding_configuration(void)
...
+				decoded_values.accelerometer_inactivity_timeout);
 
 	cJSON_Delete(decoded_root_obj);
 }
 
-/* Setup and teardown functions. Used to allocate root and array objects used in test and to
- * cleanup allocated memory afterwards.
- */
-
-static void test_setup_object(void)
```
balaji-nordic, line 1196: All setup and tear down functions are merged and moved to the top.

applications/asset_tracker_v2/tests/json_common/src/main.c
```diff
@@ -1166,170 +1184,30 @@ static void test_floating_point_encoding_configuration(void)
...
-	cJSON_FreeString(dummy.buffer);
-	cJSON_Delete(dummy.array_obj);
-}
-
-void test_main(void)
+int main(void)
 {
 	cJSON_Init();
-
-	ztest_test_suite(json_common,
```
balaji-nordic, line 1224: This whole block is no longer needed as unity framework automatically picks up the test cases by searching this file for `test_` functions.

jtguggedal (approved):

simensrostad (approved):

PR #11054, opened 2023-05-05
doc: Extend supported board list in Features of nRF70 Series

richabp:
Extended supported board list in the
Features of nRF70 Series guide.

SHEL-1587

### DIFF
--- a/doc/nrf/working_with_nrf/nrf70/features.rst
+++ b/doc/nrf/working_with_nrf/nrf70/features.rst
@@ -7,10 +7,7 @@ Features of nRF70 Series
     :local:
     :depth: 2
 
-Introduction
-************
-
-The nRF70 Series of wireless companion ICs add Wi-Fi capabilities to another System-on-Chip (SoC), Microprocessor Unit (MPU), or Microcontroller (MCU) host device.
+The nRF70 Series of wireless companion ICs add Wi-Fi® capabilities to another System-on-Chip (SoC), Microprocessor Unit (MPU), or Microcontroller (MCU) host device.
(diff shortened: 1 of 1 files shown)


richabp:
@mjd90 @bama-nordic : As suggested by @sachinthegreen, do we need to add nRF7001 DK in the list?

krish2718 (approved):

mjd90 (commented):
Can we add an additional column in the table, e.g. "Companion module", between DK and PCA columns. The first row (7002 DK) would be N/A, the next 3 would be "nRF7002 EK" and the 4th would be "nRF7002 EB". We don't need the PCA number for these.

Once available, we should add links to the EK user guide (in the existing column).

doc/nrf/working_with_nrf/nrf70/features.rst
```diff
@@ -29,25 +26,54 @@ The following figure illustrates the partitioning of the Wi-Fi stack between the
...
+     - nRF7002 EK
+     - PCA10090
+     - ``nrf9160dk_nrf9160_ns``
+     - | `Product Specification <nRF9160 Product Specification_>`_
+       | :ref:`Getting started <ug_nrf9160_gs>`
+       | `User Guide <nRF9160 DK Hardware_>`_
+   * - Thingy:53
+     - nRF7002 EB
+     - PCA20053
+     -
```
bama-nordic, line 71: Please add Thingy53 board target - `thingy53_nrf5340_cpuapp`

bama-nordic (approved):

mjd90 (approved):
LGTM

PR #11055, opened 2023-05-05
tfm: Add support for boards using nrf9120 SoC

joerchan:


joerchan:
Draft: This PR still requires changes for full nrf9161 support.

samples/nrf9160/modem_shell/sample.yaml
```diff
@@ -5,70 +5,70 @@ tests:
     build_only: true
     integration_platforms:
       - nrf9160dk_nrf9160_ns
-    platform_allow: nrf9160dk_nrf9160_ns
+    platform_allow: nrf9160dk_nrf9160_ns nrf9161dk_nrf9161_ns
```
trantanen, line 8: I'm asking because I don't know and we should have a conscious decision about these:

1. Do we need to add `nrf9161dk_nrf9161_ns` also to `integration_platforms`?
2. What about the rest of the samples under nrf9160? So we need similar compilation for all of our samples for nrf9161dk_nrf9161_ns?

joerchan:
Addressing uncovered license and yaml issues here: https://github.com/nrfconnect/sdk-nrf/pull/11106
This PR needs to be merged first.

samples/nrf9160/modem_shell/sample.yaml
eivindj-nordic: 1. AFAIK, we currently do not have any test setup in CI for nrf9161 DK. The plan is to add it to integration_platforms once we have test nodes with it. 
2. This is the samples and applications that it has been decided to support for now. More samples are to be supported later.

trantanen: Ok

trantanen (approved):
modem_shell looks good for me.

VTPeltoketo (approved):
SLM tests passed

joerchan:
@eivindj-nordic Rebased, "modules: memfault: add nRF9161 support" commit had merge-conflicts, can you verify that I resolved it correctly?

eivindj-nordic:
> @eivindj-nordic Rebased, "modules: memfault: add nRF9161 support" commit had merge-conflicts, can you verify that I resolved it correctly?

Thanks, looks correct to me!

lib/modem_antenna/Kconfig
```diff
@@ -7,7 +7,7 @@
 menuconfig MODEM_ANTENNA
 	bool "Modem antenna configuration"
 	depends on NRF_MODEM_LIB
-	default y if (BOARD_NRF9160DK_NRF9160_NS || BOARD_THINGY91_NRF9160_NS)
+	default y if SOC_SERIES_NRF91X
```
tokangas, line 10: The library is intended to be enabled by default for nRF916x DKs and Thingy:91s. `SOC_SERIES_NRF91X` is enabled also for customer boards, where depending on their design the library may not be needed. Maybe it would be better to just add `BOARD_NRF9161DK_NRF9161_NS` flagging instead of using `SOC_SERIES_NRF91X`.

samples/nrf9160/modem_shell/sample.yaml
```diff
@@ -162,7 +162,7 @@ tests:
...
     extra_args: >
       OVERLAY_CONFIG="overlay-modem-trace.conf;overlay-memfault.conf"
       DTC_OVERLAY_FILE="nrf9160dk_ext_flash.overlay"
```
tokangas, line 168: nRF9161 DK build should probably use `nrf9161dk_ext_flash.overlay` instead.

lib/modem_antenna/Kconfig
eivindj-nordic: My thought was that this is most likely used by other boards as well, and it was better to use the SoC series to avoid a long list of boards in the end. But I am fine with keeping it too :)

samples/nrf9160/modem_shell/sample.yaml
eivindj-nordic: I'm not sure what is the best way to allow different overlays based on the platform in this case, so I'll drop this for now and add it later for this test.

lib/modem_antenna/Kconfig
```diff
@@ -25,16 +27,16 @@ endchoice
...
 	help
 	  Defines the AT%XMAGPIO command to be sent to the modem. Leave
 	  empty if this command should not be sent.
 
 config MODEM_ANTENNA_AT_COEX0
 	string "AT%XCOEX0 command"
-	default "AT\%XCOEX0=1,1,1565,1586" if (BOARD_NRF9160DK_NRF9160_NS || BOARD_THINGY91_NRF9160_NS) && MODEM_ANTENNA_GNSS_ONBOARD
-	default "AT\%XCOEX0" if (BOARD_NRF9160DK_NRF9160_NS || BOARD_THINGY91_NRF9160_NS) && MODEM_ANTENNA_GNSS_EXTERNAL
+	default "AT\%XCOEX0=1,1,1565,1586" if (SOC_SERIES_NRF91X && MODEM_ANTENNA_GNSS_ONBOARD)
+	default "AT\%XCOEX0" if (SOC_SERIES_NRF91X && MODEM_ANTENNA_GNSS_EXTERNAL)
```
tokangas, line 39: I think we could just remove `SOC_SERIES_NRF91X` from these conditions. I probably put the DK/Thingy flagging here myself originally, but I don't think it's really needed. When the library is enabled, the default COEX0 configuration can be selected based on the type of the antenna.

tokangas (approved):

mia-ko (commented):
Do we need a changelog entry for these changes?

mia-ko (approved):

rlubos (approved):

rlubos:
All jobs passed on rerun

PR #11065, opened 2023-05-05
Configuring Tx power back off value for Edge band frequencies

srinivas1201:


srkanordic (approved):

sachinthegreen (approved):

rado17 (approved):

drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -114,6 +114,66 @@ config NRF700X_ANT_GAIN_5G_BAND3
 	default 0
 	range 0 6
 
+config NRF700X_BAND_2G_LOWER_EDGE_BACKOFF
```
krish2718, line 117: I don't think these should be part of Kconfig, as these are HW related they have to be in the DTS.

krish2718 (changes requested):

drivers/wifi/nrf700x/Kconfig.nrf700x
sachinthegreen: Agree. Created a Jira https://projecttools.nordicsemi.no/jira/browse/SHEL-1613 for this. We can target this in a subsequent PR.

krish2718 (approved):

PR #11068, opened 2023-05-05
manifest:  Add option to build and use PSA core from Oberon

joerchan:
test-sdk-nrf: testing-psa-core-09-05


modules/trusted-firmware-m/Kconfig.psa.defconfig
```diff
@@ -115,10 +115,10 @@ config PSA_WANT_ECC_BRAINPOOL_P_R1_256
 	default y
 config PSA_WANT_ECC_BRAINPOOL_P_R1_384
 	bool
-	default y
```
Vge0rge, line 118: You can remove the default line completely, right? Why do you need to keep it?

modules/trusted-firmware-m/Kconfig.psa.defconfig
```diff
@@ -67,7 +67,7 @@ config PSA_WANT_ALG_RIPEMD160
 	default n
 config PSA_WANT_ALG_MD5
 	bool
-	default y
+	default n
```
frkv, line 70: Note: TCP module used legacy MD5 APIs when CONFIG_NET_TCP_ISN_RFC6528 is enabled. 

We currently convert PSA_WANT_ALG_MD5 to MBEDTLS_MD5_C in the file to convert PSA_WANT_ALG_XXXX to builtin equivalents. Please see: https://github.com/nrfconnect/sdk-nrfxlib/blob/main/nrf_security/configs/nrf-config.h#L47

I recommend changing this to `default y if !NET_TCP_ISN_RFC6528`

joerchan: It would be impossible to satisfy this condition with CONFIG_PSA_CORE_OBERON.
We could have:
```
config PSA_WANT_ALG_MD5
      bool
      depends on PSA_CORE_BUILTIN
      default y if NET_TCP_ISN_RFC6528
```

joerchan: Given that it appears CONFIG_NET_TCP_ISN_RFC6528  is using legacy APIs and do not support PSA I don't think we need to enable this.

modules/trusted-firmware-m/Kconfig.psa.defconfig
joerchan: Technically, yes. But this file sets the default configurations to either y or n, so having it here for completeness, also in case they original default changes this would not be affected by that.

modules/trusted-firmware-m/Kconfig.psa.defconfig
frkv: We have multiple ways of solving this if it becomes a problem in the future. The proper solution is to not use MD5 for this in Zephyr. We should focus our attention on that.

frkv (approved):
LGTM pending green test and correct SHA

Vge0rge (approved):

west.yml
```diff
@@ -59,7 +59,7 @@ manifest:
     # https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/modules.html
     - name: zephyr
       repo-path: sdk-zephyr
-      revision: c2859f736bfc959b91cca472ba32563afccea0bc
+      revision: pull/1157/head
```
rlubos, line 62: Zephyr SHA: `6ae8ab054126778fcc0054d5243580e72197359c`

west.yml
```diff
@@ -131,7 +131,7 @@ manifest:
     - name: nrfxlib
       repo-path: sdk-nrfxlib
       path: nrfxlib
-      revision: 37e3dd8997b3db49527c7cdcea92021ee3c0e97c
+      revision: pull/972/head
```
rlubos, line 134: nrfxlib SHA: `4812d76369271546120386dff7fa38f717f1d787`

rlubos (approved):

rlubos:
There was a little havoc in the CI, but jobs passed after (multiple) reruns.

PR #11080, opened 2023-05-08
Fix scan filter get manufacturer

alwa-nordic:


doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -450,6 +450,11 @@ Bluetooth libraries and services
     * Salt size in the Fast Pair not discoverable advertising from 1 byte to 2 bytes, to align with the Fast Pair specification update.
     * The :kconfig:option:`CONFIG_BT_FAST_PAIR_CRYPTO_OBERON` Kconfig option is now the default Fast Pair cryptographic backend.
 
+* :ref:`nrf_bt_scan_readme` library:
+
+  * Fixed:
+
+    * The :c:func:`bt_scan_filter_status_get` was missing ``manufacturer_data.enabled``.
```
wiba-nordic, line 457:
```suggestion
    * The :c:func:`bt_scan_filter_status_get` function was missing the ``manufacturer_data.enabled`` status check.
```
Not sure if "status check" is correct. Do use a more appropriate term if it is wrong.

wiba-nordic (changes requested):

doc/nrf/releases/release-notes-changelog.rst
alwa-nordic: Is a long-form explanation is like this good? I'm not sure how to best word the "output argument" part.

```suggestion
    * The :c:func:`bt_scan_filter_status_get` was never setting the :c:member:`bt_filter_status.manufacturer_data.enabled` field on its output argument. This field is now correctly set to reflect the current status of the filter when the function is called.
```

wiba-nordic:
```suggestion
    * The output arguments of the :c:func:`bt_scan_filter_status_get` function.
      The :c:member:`bt_filter_status.manufacturer_data.enabled` field is now correctly set to reflect the status of the filter when the function is called.
```
It's preferable to word  bullet point so that the heading (fixed/added/removed) fits like it was part of the sentence. But long-form explanation is more than fine.

wiba-nordic (approved):
Line 458 doesn't have to be empty, but it can be. Both are OK.

KAGA164 (approved):

carlescufi (approved):

PR #11082, opened 2023-05-08
wifi: scan: Fix scan results display

D-Triveni:
As LOG_PRINTK is enabled, printk is buffering the
scan results. So, disabled the config

Fixes SHEL-1450


samples/wifi/scan/prj.conf
```diff
@@ -31,6 +31,7 @@ CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
 # Logging
 CONFIG_LOG=y
 CONFIG_PRINTK=y
+CONFIG_LOG_PRINTK=n
```
krish2718, line 34: Can you please add a comment as to why we need this?

udaynordic (approved):

bama-nordic (approved):

PR #11095, opened 2023-05-09
lib: nrf_modem: trace: add shell command

gregersrygg:
Move modem_trace command from modem_shell sample to nrf_modem_lib. Can be used in any application or sample that have CONFIG_SHELL and CONFIG_NRF_MODEM_LIB_TRACE enabled.


doc/nrf/libraries/modem/nrf_modem_lib.rst
```diff
@@ -329,6 +329,76 @@ Other configurations set in the devicetree, such as the current speed, are overw
...
+
+Examples
+--------
+
+* Modem trace everything (LTE, IP and GNSS):
+
+  .. code-block:: console
+
+     modem_trace start full
+     <test using gnss-, lte-, or ip-commands>
```
tokangas, line 349: References to `modem_shell` commands should probably be removed from here.

tokangas (commented):
These are the first shell commands for modemlib, but more could be added later, I guess. So, just thinking if we should instead add command `modem` with a sub-command `trace`? This way all modemlib related commands could be under the same `modem` command.

lib/nrf_modem_lib/shell/modem_trace_memfault.c
```diff
@@ -61,10 +66,13 @@ static sMemfaultCdrSourceImpl s_my_custom_data_recording_source = {
 	.mark_cdr_read_cb = mark_cdr_read_cb,
 };
 
-void modem_trace_memfault_send(uint32_t trace_duration_ms)
+void modem_trace_memfault_send(const struct shell *sh, size_t size, uint32_t trace_duration_ms)
 {
 	static bool memfault_cdr_source_registered;
-	size_t size = nrf_modem_lib_trace_data_size();
+
+	if (shl == NULL) {
```
trantanen, line 73: I guess this NULL check is kind of unnecessary and shl could be set always but doesn't harm either.

modules/memfault-firmware-sdk/Kconfig
```diff
@@ -187,6 +187,7 @@ config MEMFAULT_NCS_INTERNAL_FLASH_BACKED_COREDUMP
 
 config MEMFAULT_CDR_ENABLE
 	bool "Custom Data Recording"
+	default y if NRF_MODEM_LIB_TRACE_SHELL
```
trantanen, line 190: Is it a good practice to refer to a Kconfig of a client of a module given the client is elsewhere in the code tree?
I guess we could handle this within nrf_modem_lib Kconfig file?

doc/nrf/libraries/modem/nrf_modem_lib.rst
gregersrygg: Good catch 👍

modules/memfault-firmware-sdk/Kconfig
trantanen: Or could it be that libraries are not able to override a Kconfig. Not sure...

What about using `select MEMFAULT_CDR_ENABLE` (could be `imply` or `depends on` too) in nrf_modem_lib Kconfig file?

> This kconfig is one we made in sdk-nrf for our wrapper around the memfault lib, so I don't see why that is problematic.

Maybe it's just me that sees these references from a library Kconfig to its client as suboptimal from architectural point of view. You can do that as Kconfig allows references anywhere but I just feel it's a bit odd. In my opinion it should be the client that defines the libraries it's using.

eivindj-nordic (commented):
I'm wondering if this should be part of the modem lib integration or split into a separate library. If we start to add shell commands to the modem library integration, I guess there is a lot of other shell commands that can be added as well, and it might become quite extensive. I'm not sure what is best, just wanted to discuss it.

modules/memfault-firmware-sdk/Kconfig
gregersrygg: I've tried the different alternatives, but `select MEMFAULT_CDR_ENABLE if MEMFAULT` seemed to work the best.
- `imply` will cause a compilation error without more `#ifdef`s in the code, and `MEMFAULT_CDR_ENABLE` would also still end up as `y` even with `CONFIG_MEMFAULT_CDR_ENABLE=n` in a config file.
- `depends on` can't have an `if`-expression

gregersrygg:
> I'm wondering if this should be part of the modem lib integration or split into a separate library. If we start to add shell commands to the modem library integration, I guess there is a lot of other shell commands that can be added as well, and it might become quite extensive. I'm not sure what is best, just wanted to discuss it.

@eivindj-nordic Good question. This is mostly your domain. Do you want to discuss it in your team and let me know where you decide to have it?

lib/nrf_modem_lib/shell/Kconfig
```diff
@@ -0,0 +1,5 @@
+config NRF_MODEM_LIB_TRACE_SHELL
+	bool "Shell commands for modem traces"
+	default y
```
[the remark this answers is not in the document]
gregersrygg: It's only being enabled by default if the dependencies are being met. So `...if SHELL && NRF_MODEM_LIB_TRACE` doesn't have any effect.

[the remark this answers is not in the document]
eivindj-nordic: Ok, I was thinking this would give an error because of unsatisfied dependencies, but I see now that it is only shown if this is enabled by the user when the dependencies are not set.

eivindj-nordic:
> I'm wondering if this should be part of the modem lib integration or split into a separate library.

I'll create a separate chat for this discussion :)

lemrey:
I think that it's fine to have shell commands for functionality that is offered "natively" by `nrf_modem_lib`, for example enabling/disabling traces, changing trace level, turning modem on/off etc. If the functionality is not provided by `nrf_modem_lib` (ncs) or the `nrf_modem` library (nrfxlib), then it should be put somewhere else. In this case, I think the `memfault` integration does not belong to `nrf_modem_lib`, it should perhaps be put in `memfault`.

lib/nrf_modem_lib/shell/modem_trace_shell.c
```diff
@@ -58,57 +58,66 @@ int modem_traces_stop(void)
...
-	mosh_print("Clearing modem traces");
+	shell_print(sh, "Clearing modem traces");
 
 	err = nrf_modem_lib_trace_clear();
-	if (err) {
-		mosh_error("Failed to clear modem traces: %d", err);
+	if (err == -ENOTSUP) {
+		shell_error(sh, "The current modem trace backend does not support this operation");
+	} else if (err) {
+		shell_error(sh, "Failed to clear modem traces: %d", err);
```
balaji-nordic, line 73: Add else block as requested  here https://github.com/nrfconnect/sdk-nrf/pull/11582#discussion_r1240340232

lib/nrf_modem_lib/shell/modem_trace_memfault.c
lemrey, line 1: The `memfault` functionality doesn't belong here, we should move it somewhere else.

lemrey (changes requested):

PR #11096, opened 2023-05-09
lib: pdn: fix PDN context initialization

kacperradoszewski:
When creating a new PDN context, assign the event handler regardless of
whether the input callback is NULL or not. The callback pointer is
uninitialized having been allocated from the heap, which could lead to
illegal access upon receiving events. Additionally, free the PDN context,
should the initialization fail.

### DIFF
--- a/doc/nrf/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases/release-notes-changelog.rst
@@ -491,6 +491,11 @@ Modem libraries
 
   * Updated the library to use ePCO mode if the Kconfig option :kconfig:option:`CONFIG_PDN_LEGACY_PCO` is not enabled.
 
+  * Fixed:
+
+    * A bug in the initialization of a new PDN context without a PDN event handler.
+    * A memory leak in the :c:func:`pdn_ctx_create` function.
+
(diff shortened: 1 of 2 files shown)


lemrey (approved):

lemrey:
We'll need an entry in the changelog if this bug was present in the previous NCS version.

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -489,7 +489,14 @@ Modem libraries
 
 * :ref:`pdn_readme` library:
 
-  * Updated the library to use ePCO mode if the Kconfig option :kconfig:option:`CONFIG_PDN_LEGACY_PCO` is not enabled.
+  * Updated:
+
+    * Updated the library to use ePCO mode if the Kconfig option :kconfig:option:`CONFIG_PDN_LEGACY_PCO` is not enabled.
```
divipillai, line 494:
```suggestion
  * Updated the library to use ePCO mode if the Kconfig option :kconfig:option:`CONFIG_PDN_LEGACY_PCO` is not enabled.
```
Since we just have a single entry.

divipillai (approved):

PR #11102, opened 2023-05-09
driver: wifi: Enforce minimum value for NRF700X_MAX_TX_TOKENS

balaji-nordic:
The minimum value of `NRF700X_MAX_TX_TOKENS` has to be atleast `WIFI_NRF_FMAC_AC_MAX` (5). If not, this will lead to the setting of `num_tx_tokens_per_ac` to 0 by `wifi_nrf_fmac_init()` which in-turn will lead to divide by zero errors in `clear_spare_desc_q_map()` and `set_spare_desc_q_map()`. This is now fixed by enforcing a minimal value for the KConfig symbol `NRF700X_MAX_TX_TOKENS`.
Also added ASSERT statements to catch this condition if it ever arises.

Found as violation of CERT, INT33-C and MITRE, CWE-369 by sonarcloud.

### DIFF
--- a/drivers/wifi/nrf700x/Kconfig.nrf700x
+++ b/drivers/wifi/nrf700x/Kconfig.nrf700x
@@ -126,6 +126,7 @@ config NRF700X_MAX_TX_AGGREGATION
 
 config NRF700X_MAX_TX_TOKENS
 	int "Maximum number of TX tokens"
+	range 5 12 if !NRF700X_RADIO_TEST
 	default 10
 
 config NRF700X_TX_MAX_DATA_SIZE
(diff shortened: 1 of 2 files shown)


drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -126,6 +126,7 @@ config NRF700X_MAX_TX_AGGREGATION
 
 config NRF700X_MAX_TX_TOKENS
 	int "Maximum number of TX tokens"
+	range 5 255 if !NRF700X_RADIO_TEST
```
krish2718, line 129: The range should only be till 12, we will not support more than that.

krish2718 (approved):

drivers/wifi/nrf700x/Kconfig.nrf700x
balaji-nordic: sure. will fix

PR #11118, opened 2023-05-09
samples: nrf9160: nrf_cloud_rest_fota: disable mcuboot ext flash

jayteemo:
Disable using external flash for the MCUboot secondary partition for the default build configuration.
This feature is not supported by the sample.
NCSIDB-981


samples/nrf9160/nrf_cloud_rest_fota/prj.conf
```diff
@@ -25,6 +25,7 @@ CONFIG_MCUBOOT_IMG_MANAGER=y
 CONFIG_IMG_ERASE_PROGRESSIVELY=y
 CONFIG_SECURE_BOOT=y
 CONFIG_BUILD_S1_VARIANT=y
+CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=n
```
glarsennordic, line 28: Does `CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=y` need to be added to the FMFU Kconfig overlay?

glarsennordic (approved):

samples/nrf9160/nrf_cloud_rest_fota/prj.conf
glarsennordic: I guess that isn't done with MCUBoot so the answer would be no, wouldn't it?

plskeggs (approved):

hellesvik-nordic (approved):

PR #11121, opened 2023-05-10
Serial LTE Modem - Coverity fixes

jarlamsa:
Fixed some coverity issues for Serial LTE Modem.
Two of the fixes are trivial issues, one fix also had some refactoring included.

### DIFF
--- a/applications/serial_lte_modem/src/slm_at_host.c
+++ b/applications/serial_lte_modem/src/slm_at_host.c
@@ -31,6 +31,8 @@ LOG_MODULE_REGISTER(slm_at_host, CONFIG_SLM_LOG_LEVEL);
 #define FATAL_STR	"FATAL ERROR\r\n"
 #define SLM_SYNC_STR	"Ready\r\n"
 #define CRLF_STR	"\r\n"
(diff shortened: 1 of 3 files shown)


applications/serial_lte_modem/src/slm_at_host.c
```diff
@@ -778,71 +780,74 @@ static int cmd_grammar_check(const uint8_t *cmd, uint16_t length)
...
+	}
+
+	for (int i = len1 - len2; i >= 0; i--) {
+		if (strncmp(str1 + i, str2, len2) == 0) {
+			return (char *)str1 + i;
+		}
+	}
+
+	return NULL;
+}
```
VTPeltoketo, line 806: Why default library strstr -function does not work here?

jarlamsa: This is `str**r**str`, AKA find the last occurrence of a string in a string, which is not on the standard C library.

VTPeltoketo (approved):

rlubos (approved):

SeppoTakalo (approved):

PR #11135, opened 2023-05-11
bootloader: cmake: keys: Change from source dir to app config dir

nordicjm:
Changes the base directory for signing keys to be based on the application configuration directory instead of the source directory.


tejlmand (changes requested):
a big reason for this is to advocate better security practices, especially for applications where board configurations are divided into dedicated configuration folders, then we should highlight that in the commit messages.

something like:
>  bootloader: cmake: keys: Change from source dir to app config dir
>
> Changes the base directory for signing keys to be based on the
> application configuration directory instead of the source directory.
>
> For simple samples where APPLICATION_CONFIG_DIR is identical to APPLICATION_SOURCE_DIR, this change has no impact,
> but for applications where each board is having a dedication configuration folder, such as the nrf_desktop,
> then this change introduces a much better security practice because it now indicates that each board should have a unique private key.
> The current approach of sharing the private key across boards is not considered a good security practice.

frkv:
There is a chapter in documentation explaining how to [add a custom signing key](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/app_dev/bootloaders_and_dfu/bootloader_adding.html#adding-a-custom-signature-key-file) that should be updated as a result of this change.

This talks about relative paths of private keys like this "The relative path to the file from the application directory."

MarekPieta (approved):

greg-fer (approved):

PR #11152, opened 2023-05-12
lib: nrf_modem: do CFUN=0 in nrf_modem_lib_shutdown if not done already

eivindj-nordic:
The modem is required to be in minimal functional mode (CFUN=0) before it is shut down. This commit checks if this is the case when calling nrf_modem_lib_shutdown. If not, throw a warning and set the functional mode.

### DIFF
--- a/doc/nrf/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases/release-notes-changelog.rst
@@ -505,6 +505,8 @@ Modem libraries
       Use the :c:func:`nrf_modem_lib_bootloader_init` function to initialize the Modem library in bootloader mode.
     * The Kconfig option :kconfig:option:`CONFIG_NRF_MODEM_LIB_SYS_INIT` is now deprecated.
       The application initializes the modem library using the :c:func:`nrf_modem_lib_init` function instead.
+    * The :c:func:`nrf_modem_lib_shutdown` function now checks that the modem is in minimal functional mode (``CFUN=0``) before shutting down the modem.
+      If not, a warning is given to the application, and minimal functional mode is set before calling the :c:func:`nrf_modem_shutdown` function.
 
   * Removed:
 
--- a/include/modem/nrf_modem_lib.h
+++ b/include/modem/nrf_modem_lib.h
@@ -61,6 +61,8 @@ int nrf_modem_lib_bootloader_init(void);
 /**
  * @brief Shutdown the Modem library and turn off the modem.
  *
+ * @note The modem must be put in minimal function mode before being shut down.
+ *
  * @return int Zero on success, non-zero otherwise.
  */
 int nrf_modem_lib_shutdown(void);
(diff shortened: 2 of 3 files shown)


lib/nrf_modem_lib/nrf_modem_lib.c
```diff
@@ -151,12 +154,23 @@ int nrf_modem_lib_bootloader_init(void)
...
 	LOG_DBG("Shutting down modem library");
 	STRUCT_SECTION_FOREACH(nrf_modem_lib_shutdown_cb, e) {
 		LOG_DBG("Modem shutdown callback: %p", e->callback);
 		e->callback(e->context);
 	}
 
+	/* The application must set CFUN=0 before calling nrf_modem_shutdown. */
+	ret = nrf_modem_at_scanf(AT_CFUN_READ, "+CFUN: %hu", &mode);
+	if (ret == 1 && mode != AT_CFUN0_VAL) {
+		LOG_WRN("Application should set minimal functional mode (CFUN=0) before
```
stig-bjorlykke, line 169: Do we really need this warning?
This function already check CFUN, why should we need to duplicate this in the application?

eivindj-nordic: Perhaps not. The reasoning for it is that the application sets the functional mode to CFUN=1 etc. so it should be responsible for power off (CFUN=0) as well. But we add it here to prevent issues if the application has not done it already. One alternative we looked at was to return an error, but we want the application to be able to use the NRF_MODEM_LIB_SHUTDOWN hook for doing CFUN=0.

eivindj-nordic: @lemrey Any thoughts?

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -503,6 +503,8 @@ Modem libraries
       Use the :c:func:`nrf_modem_lib_bootloader_init` function to initialize the Modem library in bootloader mode.
     * The Kconfig option :kconfig:option:`CONFIG_NRF_MODEM_LIB_SYS_INIT` is now deprecated.
       The application initializes the modem library using the :c:func:`nrf_modem_lib_init` function instead.
+    * The :c:func:`nrf_modem_lib_shutdown` function now checks that the modem is in minimal functional mode (CFUN=0) before shutting down the modem.
+      If not, a warning is given to the application and minimal functional mode is set before calling :c:func:`nrf_modem_shutdown`.
```
divipillai, line 507:
```suggestion
    * The :c:func:`nrf_modem_lib_shutdown` function now checks that the modem is in minimal functional mode (``CFUN=0``) before shutting down the modem.
      If not, a warning is given to the application, and minimal functional mode is set before calling the :c:func:`nrf_modem_shutdown` function.
```

lib/nrf_modem_lib/nrf_modem_lib.c
lemrey: We print a warning because the application should be responsible of controlling the modem's functional mode. We don't want the application to rely on `nrf_modem_lib_shutdown` to set the functional mode. Before shutting down the modem, the application should CFUN to zero; if it doesn't it's a bug in the application code.

include/modem/nrf_modem_lib.h
```diff
@@ -61,6 +61,8 @@ int nrf_modem_lib_bootloader_init(void);
 /**
  * @brief Shutdown the Modem library and turn off the modem.
  *
+ * @note The modem should be in minimal functional mode (CFUN=0).
```
lemrey, line 64: nit: `The modem must be put in minimal function mode before being shut down.`

lemrey (approved):

divipillai (approved):

stig-bjorlykke (approved):

PR #11154, opened 2023-05-12
samples: nrf9160: https_client: handle PDN_EVENT_NETWORK_DETACH

MechanicV1:
added support to handle PDN_EVENT_NETWORK_DETACH.

### DIFF
--- a/samples/nrf9160/https_client/src/main.c
+++ b/samples/nrf9160/https_client/src/main.c
@@ -100,6 +100,9 @@ void pdn_event_handler(uint8_t cid, enum pdn_event event, int reason)
 	case PDN_EVENT_DEACTIVATED:
 		printk("PDP context %d deactivated\n", cid);
 		break;
+	case PDN_EVENT_NETWORK_DETACH:
+		printk("PDP context %d network detached\n", cid);
+		break;
 #if !IS_ENABLED(CONFIG_PDN_DEFAULT_FAM_IPV4)
 	case PDN_EVENT_IPV6_UP:
 		printk("PDP context %d IPv6 up\n", cid);


rlubos:
@MechanicV1 Seems that you need to update your `user.name` git config as it does not match the signoff entry (see compliance results).

Also please follow https://projecttools.nordicsemi.no/confluence/display/NCS/Contributors#Contributors-Becomingacontributor, this should not be identified as an external contribution.

samples/nrf9160/https_client/src/main.c
```diff
@@ -100,6 +100,9 @@ void pdn_event_handler(uint8_t cid, enum pdn_event event, int reason)
 	case PDN_EVENT_DEACTIVATED:
 		printk("PDP context %d deactivated\n", cid);
 		break;
+	case PDN_EVENT_NETWORK_DETACH:
+		printk("PDP context %d network detached\n", cid);
```
lemrey, line 104: @stig-bjorlykke should be log the `cid` here?

stig-bjorlykke: The `cid` is not included in the AT notification, but it is included in the event sent to all pdn contexts.
This is ok, to indicate the context in use for the https_client.

lemrey (approved):

lemrey:
This is not an external contribution, so I removed the `external` label.
It also does not need really a changelog entry, it is required by CI and it just adds a log line, so I removed the `changelog-entry-required` label too.

lemrey:
I have added the milestone for the next release; this PR is needed to make sure the automated tests for the `https_client` sample are green.

eivindj-nordic (approved):

PR #11155, opened 2023-05-12
samples: nrf9160: modem_shell: Increase workqueue stack size

trantanen:
Increase MOSH_COMMON_WORKQUEUE_STACK_SIZE to 9126 to accommodate commands sent from cloud.
Normal commands are run in shell thread but commands sent from cloud are run in modem shell's own common thread.
This means we need to increase the stack for the common thread a lot just because of the commands sent from the cloud as we cannot run them in the same thread as normal shell commands.

Jira: MOSH-444


jhirsi (approved):

tokangas (approved):

greg-fer:
Please verify if a changelog entry is needed. If not, remove the label.

PR #11169, opened 2023-05-12
net: lib: nrf_cloud: add nrf_cloud_obj struct for encoding/decoding cloud data

jayteemo:
Add nrf_cloud_obj structure for encoding/decoding cloud data.
Add and use function to create a P-GPS request message using
nrf_cloud_obj.  Remove old encoding function.
Replace nrf_cloud_location_request_msg_json_encode() with
similar function, nrf_cloud_obj_location_request_create(),
which uses nrf_cloud_obj.

IRIS-6060


tony-le-24 (approved):
LGTM!

include/net/nrf_cloud.h
```diff
@@ -743,13 +746,14 @@ int nrf_cloud_sensor_data_stream(const struct nrf_cloud_sensor_data *param);
 /**
  * @brief Send data to nRF Cloud.
  *
- * This API is used to send pre-encoded data to nRF Cloud.
+ * This API is used to data to nRF Cloud.
```
peknis, line 749:
```suggestion
 * This API is used for sending data to nRF Cloud.
```

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -266,7 +266,17 @@ Modem libraries
...
+
+  * Added:
+
+    * :c:struct:`nrf_cloud_obj` structure and functions for encoding and decoding nRF Cloud data.
+    * :c:func:`nrf_cloud_obj_pgps_request_create` function that creates a P-GPS request for nRF Cloud.
+
+  * Removed:
+
+    * Unused internal codec function ``nrf_cloud_format_single_cell_pos_req_json()``.
+    * :c:func:`nrf_cloud_location_request_msg_json_encode` function and replaced with :c:func:`nrf_cloud_obj_location_request_create`.
```
peknis, line 279: I believe the removed one needs to be without the :c:func: tag and in double backticks, as it will not be there anymore and is not linkable.

include/net/nrf_cloud_codec.h
```diff
@@ -20,37 +29,517 @@ extern "C" {
...
+ * @retval -EINVAL Invalid parameter.
+ * @retval -ENOTEMPTY Object already initialized.
+ * @retval -ENOMEM Out of memory.
+ * @retval -ENOTSUP Action not supported for the object's type.
+ * @retval 0 Success; object initialized.
+ */
+int nrf_cloud_obj_init(struct nrf_cloud_obj *const obj);
+
+/**
+ * @brief Reset the state of an object; does not free memory
```
peknis, line 217:
```suggestion
 * @brief Reset the state of an object; does not free memory.
```

include/net/nrf_cloud_codec.h
```diff
@@ -20,37 +29,517 @@ extern "C" {
...
+ * @retval -EIO Error adding object.
+ * @retval -ENOTSUP Action not supported for the object's type.
+ * @retval 0 Success; item added.
+ */
+int nrf_cloud_obj_bulk_add(struct nrf_cloud_obj *const bulk, struct nrf_cloud_obj *const obj);
+
+/**
+ * @brief Add a timestamp to an object.
+ *
+ * @param[out] obj Object to contain timestamp
```
peknis, line 262:
```suggestion
 * @param[out] obj Object to contain timestamp.
```

peknis (changes requested):

tests/subsys/net/lib/nrf_cloud/cloud/src/main.c
```diff
@@ -1478,6 +1478,28 @@ ZTEST(nrf_cloud_test, test_cloud_send_invalid_topic_type)
 	zassert_equal(-ENODATA, ret, "return should be -ENODATA with invalid topic type");
 }
 
+/* Verify that nrf_cloud_send will fail if object encoding fails */
+ZTEST(nrf_cloud_test, test_cloud_send_obj_encode_fail)
```
glarsennordic, line 1482: I think there needs to be a commit between this commit and the next one that includes unit tests for each of the newly added functions -- That way we don't wind up with an increasing backlog of non-verified functions

Just some basic pass/fail tests of each of the modifier functions would suffice, I think

I'm not sure if a test suite for nrf_cloud_codec.c already exists, but if not, it would be, IMO, worthwhile to create one now and just populate it with tests for the new functions for the time being

include/net/nrf_cloud_codec.h
```diff
@@ -505,6 +508,27 @@ int nrf_cloud_obj_modem_pvt_add(struct nrf_cloud_obj *const obj,
...
+ * @retval -EINVAL Invalid parameter.
+ * @retval -EBADF Invalid object type.
+ * @retval -ENOTEMPTY Object already initialized.
+ * @retval -ENOMEM Out of memory.
+ * @retval -ENOTSUP Action not supported for the object's type.
+ * @retval 0 Success; GNSS message created.
+ */
+int nrf_cloud_obj_pgps_request_create(struct nrf_cloud_obj *const obj,
+				      const struct gps_pgps_request * const request);
+#endif
```
glarsennordic, line 530: See private discussion regarding this, but I suggest adding a dummy definition in the negative case

include/net/nrf_cloud_codec.h
```diff
@@ -20,6 +26,485 @@ extern "C" {
...
+/** @brief Define an nRF Cloud JSON object.
+ *
+ * This macro defines a codec object with the type of NRF_CLOUD_OBJ_TYPE_JSON.
+ *
+ * @param _name	Name of the object.
+ */
+#define NRF_CLOUD_OBJ_JSON_DEFINE(_name) \
+	struct nrf_cloud_obj _name = { .type = NRF_CLOUD_OBJ_TYPE_JSON, .json = NULL, \
+				       .enc_src = NRF_CLOUD_ENC_SRC_NONE, \
+				       .encoded_data = { .ptr = NULL, .len = 0 } }
```
glarsennordic, line 74: I think the one thing I would consider changing is having this as a macro -- Could these default values be moved into the various `_init` functions? Then users would just declare `struct nrf_cloud_obj name` and call init on the empty/invalid struct.

Maybe this makes global/static declarations of nrf_cloud_obj tough, though. But do we intend to support that?

Or maybe I am being thick-headed and there is some other reason this is a cleaner method. Or maybe my preference against using macros for this sort of thing is just me being weird. What is your stance on the matter?

Pinging @plskeggs as well for his stance

glarsennordic (changes requested):
Requesting changes only due to the possible bug with `json_decode`, the rest are just my suggestions

include/net/nrf_cloud_codec.h
jayteemo: https://github.com/jayteemo/fw-nrfconnect-nrf/pull/7#discussion_r1177115225

tests/subsys/net/lib/nrf_cloud/cloud/src/main.c
jayteemo: yes, tests will need to be written for `nrf_cloud_codec.c`, a separate ticket exists: IRIS-4788

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -0,0 +1,778 @@
...
+	/* Only try to parse JSON if the data is null-terminated */
+	if (memchr(input->ptr, '\0', input->len + 1) != NULL) {
+		json = cJSON_Parse(input->ptr);
+		if (json) {
+			obj->json = json;
+			obj->type = NRF_CLOUD_OBJ_TYPE_JSON;
+		}
+	}
+
+	return ret;
```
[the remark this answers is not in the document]
jayteemo: good catch, i obviously didn't test this after updating it for https://github.com/jayteemo/fw-nrfconnect-nrf/pull/7#discussion_r1177046758

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -0,0 +1,778 @@
...
+	{
+		char *str = NULL;
+		bool match = true;
+
+		if (!obj->json) {
+			return -ENOENT;
+		}
+
+		if (app_id) {
+			(void)get_string_from_obj(obj->json, NRF_CLOUD_JSON_APPID_KEY, &str);
```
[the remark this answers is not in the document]
jayteemo: seems fine to use `nrf_cloud_obj_str_get` instead

include/net/nrf_cloud_codec.h
glarsennordic: Oh my bad, we've had this conversation before

tests/subsys/net/lib/nrf_cloud/cloud/src/main.c
glarsennordic: My interpretation of htat ticket is that it's for retroactively creating unit tests for the functions that already exist in `nrf_cloud_codec`, as opposed to the functions added in this PR. I think we want to move toward implementing unit tests for functions as we add them, right?

jayteemo: yes that is the goal. 
it will probably take me an entire sprint to implement tests for the cloud codec since i haven't written tests in a very long time.
i'll defer to @plskeggs

glarsennordic (approved):
LGTM -- unresolved comments from last review still apply, but are only suggestions

include/net/nrf_cloud_codec.h
glarsennordic: Feel free to resolve, sorry about my forgetfulness

I'm half-tempted to suggest pointing out that its a convenience function in the macro description, but I'm not sure its worth it. Your call

peknis (approved):

include/net/nrf_cloud_codec.h
```diff
@@ -20,37 +29,517 @@ extern "C" {
...
+ *
+ * This macro defines a codec object with the type of NRF_CLOUD_OBJ_TYPE__UNDEFINED
+ * and initializes the encoded data with the provided parameters.
+ *
+ * @param _name	Name of the object.
+ * @param _data	Pointer to pre-encoded data.
+ * @param _len	Size of the pre-encoded data.
+ */
+#define NRF_CLOUD_OBJ_PRE_ENC_DEFINE(_name, _data, _len) \
+	struct nrf_cloud_obj _name = { .type = NRF_CLOUD_OBJ_TYPE__UNDEFINED, \
```
plskeggs, line 96: In order to use this with CoAP, we need to know what the format of the data is (JSON vs. CBOR), so we can set the Content-Type properly.

include/net/nrf_cloud_codec.h
```diff
@@ -20,37 +29,517 @@ extern "C" {
...
+ * @param[out] str String associated with the provided key.
+ *
+ * @retval -EINVAL Invalid parameter.
+ * @retval -ENODEV Object does not contain the provided key.
+ * @retval -ENOENT Object is not initialized.
+ * @retval -ENOTSUP Action not supported for the object's type.
+ * @retval -ENOMSG Value associated with the key is not a string.
+ * @retval 0 Success; string found.
+ */
+int nrf_cloud_obj_str_get(const struct nrf_cloud_obj *const obj, const char *const key,
```
plskeggs, line 161: What about an integer, float, or double value?

plskeggs (approved):
Epic amount of work.

include/net/nrf_cloud_codec.h
jayteemo: added `nrf_cloud_obj_num_get`

include/net/nrf_cloud_codec.h
jayteemo: this can be addressed when COAP support is implemented.

PR #11175, opened 2023-05-15
applications: nrf5340_audio: Suppress false error message of image state

pirun:
image state command will try to read image state before upload app. This commit introduces false error message while opening ram_disk partition which does not exist in application.

https://github.com/nrfconnect/sdk-zephyr/commit/df0fa6d9651ff935eafedaff383f6d098bac18e5

### DIFF
--- a/applications/nrf5340_audio/dfu/conf/Kconfig.dfu
+++ b/applications/nrf5340_audio/dfu/conf/Kconfig.dfu
@@ -70,6 +70,10 @@ config MCUMGR_GRP_IMG
 	bool
 	default y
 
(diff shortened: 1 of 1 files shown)


rick1082 (approved):

pirun:
Application is unaware of what mgmt module return, since it is called by SMP server.


> I have discussed and thought about this one. If you can check that whenever an ERR is printed in this module, we get a return with an error code and that this is caught in the calling code, we may be able to merge this.

This is what I'm trying to do on Zephyr PR. Unfortunately it is unacceptable.

PR #11178, opened 2023-05-15
net: fota_download: Don't stop on network errors

SeppoTakalo:
Fix a typo in switch-case that was not including break.

### DIFF
--- a/subsys/net/lib/download_client/src/download_client.c
+++ b/subsys/net/lib/download_client/src/download_client.c
@@ -493,7 +493,6 @@ static int reconnect(struct download_client *dl)
 		err = close(dl->fd);
 		if (err) {
 			LOG_DBG("disconnect failed, %d", err);
-			return err;
 		}
 		dl->fd = -1;
 	}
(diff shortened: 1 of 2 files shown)


subsys/net/lib/download_client/src/download_client.c
```diff
@@ -493,7 +493,6 @@ static int reconnect(struct download_client *dl)
 		err = close(dl->fd);
 		if (err) {
 			LOG_DBG("disconnect failed, %d", err);
-			return err;
```
tokangas, line 496: Just wondering if this change is intentional and necessary? I guess the original idea has been that if `close()` fails, reconnect is not tried again.

SeppoTakalo: It is intentional.
When `close()` fails, there is nothing else to do. But it does not mean that reconnect should not be tried again.

tokangas (approved):

rlubos (approved):

PR #11185, opened 2023-05-15
samples: bluetooth: mesh: use default bt rx stack size

alxelax:
Bt rx thread stack shouldn't be less than upstream's one. Some use case scenarios cause thread stack overflow.


rugeGerritsen (approved):

alxelax:
Rebased branch. For some reason, CI added a bunch of tests not related to BLE Mesh including Zigbee despite PR doesn't touch anything related to it. Zigbee tests are failed.

PR #11190, opened 2023-05-15
wifi: Fix versions

krish2718:
This needs more work, but its a start. We could automate based on NCS version ( `VERSION` file) and also add a net_mgmt command to fetch the version by applications rather than a global C function as API.

### DIFF
--- a/drivers/wifi/nrf700x/zephyr/inc/zephyr_fmac_main.h
+++ b/drivers/wifi/nrf700x/zephyr/inc/zephyr_fmac_main.h
@@ -29,6 +29,8 @@
 #include <fmac_api.h>
 #include <host_rpu_umac_if.h>
 
+#define NRF700X_DRIVER_VERSION "1.2.4.0"
+
 #ifndef CONFIG_NRF700X_RADIO_TEST
 /* Use same timeout as WPA supplicant, this is high mainly to handle
  * connected scan.
@@ -108,4 +110,5 @@ struct wifi_nrf_drv_priv_zep {
 };
 
 void wifi_nrf_scan_timeout_work(struct k_work *work);
+const char *wifi_nrf_get_drv_version(void);
 #endif /* __ZEPHYR_FMAC_MAIN_H__ */
--- a/drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
+++ b/drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
@@ -278,9 +278,8 @@ enum wifi_nrf_status wifi_nrf_fmac_dev_add_zep(struct wifi_nrf_drv_priv_zep *drv
 	sleep_type = SLEEP_DISABLE;
 #endif /* CONFIG_NRF700X_RADIO_TEST */
 #endif /* CONFIG_NRF_WIFI_LOW_POWER */
-	unsigned int umac_ver = 0;
-	unsigned int lmac_ver = 0;
 	struct nrf_wifi_tx_pwr_ctrl_params tx_pwr_ctrl_params;
+	unsigned int fw_ver = 0;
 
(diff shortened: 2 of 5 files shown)


drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
```diff
@@ -88,6 +88,11 @@ static const unsigned int rx3_buf_sz = 1000;
 
 struct wifi_nrf_drv_priv_zep rpu_drv_priv_zep;
 
+void wifi_nrf_get_drv_version(void)
```
sachinthegreen, line 91: Return type has to be char *. Also lets move this API to  drivers/wifi/nrf700x/osal/fw_if/umac_if/src/fmac_api.c and call it wifi_nrf_fmac_drv_ver_get

drivers/wifi/nrf700x/zephyr/inc/zephyr_fmac_main.h
```diff
@@ -29,6 +29,8 @@
 #include <fmac_api.h>
 #include <host_rpu_umac_if.h>
 
+#define NRF700X_DRIVER_VERSION "1.2.4.0"
```
sachinthegreen, line 32: Lets move this to drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fmac_api.h

drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
```diff
@@ -313,19 +317,18 @@ enum wifi_nrf_status wifi_nrf_fmac_dev_add_zep(struct wifi_nrf_drv_priv_zep *drv
 	}
 
 	status = wifi_nrf_fmac_ver_get(rpu_ctx,
```
sachinthegreen, line 319: Lets rename this to wifi_nrf_fmac_fw_ver_get

drivers/wifi/nrf700x/zephyr/src/zephyr_wifi_util.c
```diff
@@ -529,17 +529,20 @@ static int nrf_wifi_util_show_host_rpu_ps_ctrl_state(const struct shell *shell,
...
 	enum wifi_nrf_status status = WIFI_NRF_STATUS_FAIL;
 	struct wifi_nrf_fmac_dev_ctx *fmac_dev_ctx = NULL;
-	unsigned int umac_ver, lmac_ver;
+	unsigned int fw_ver;
 
 	fmac_dev_ctx = ctx->rpu_ctx;
 
-	status = wifi_nrf_fmac_ver_get(fmac_dev_ctx, &umac_ver, &lmac_ver);
+	shell_fprintf(shell, SHELL_INFO, "Driver version: %s\n",
+				  NRF700X_DRIVER_VERSION);
```
sachinthegreen, line 543: Better to use wifi_nrf_fmac_drv_ver_get

sachinthegreen (changes requested):

drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
krish2718: I will fix the return type, but the version has to be separate per-each OS, so, it should reside in OS specific not agnostic.

drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
krish2718: Better to leave the FMAC API generic as we can always extend the args .

sachinthegreen (approved):

bama-nordic (approved):

rado17 (approved):

PR #11194, opened 2023-05-16
doc: matter: Updated low power configuration guide

kkasperczyk-no:
Added information about SLEEPY_ACTIVE_THRESHOLD configuration to the low power configuration guide.


doc/nrf/protocols/matter/getting_started/low_power_configuration.rst
```diff
@@ -50,6 +50,14 @@ To configure the SED wake-up intervals, set both of the following Kconfig option
 * :kconfig:option:`CONFIG_CHIP_SED_IDLE_INTERVAL` to a value in milliseconds that determines how often the device wakes up to poll for data in the idle state (for example, ``1000``).
 * :kconfig:option:`CONFIG_CHIP_SED_ACTIVE_INTERVAL` to a value in milliseconds that determines how often the device wakes up to poll for data in the active state (for example, ``200``).
 
+You can also configure the SED active threshold that makes the device to stay active longer after the network activity, instead of going to idle state immediately.
```
wiba-nordic, line 53:
```suggestion
You can also configure the SED active threshold to make the device stay active after the network activity, instead of going to idle state immediately.
```

wiba-nordic (changes requested):
And the same changes to the SSED section, obviously.

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -104,6 +104,7 @@ Matter
   * The :ref:`ug_matter` protocol page with a table that lists compatibility versions for the |NCS|, the Matter SDK, and the Matter specification.
   * The :ref:`ug_matter_tools` page with installation instructions for the ZAP tool, moved from the :ref:`ug_matter_creating_accessory` page.
   * The :ref:`ug_matter_tools` page with information about CHIP Tool, CHIP Certificate Tool, and the Spake2+ Python tool.
+  * The :ref:`ug_matter_device_low_power_configuration` page with information about ``SLEEPY_ACTIVE_THRESHOLD`` parameter configuration.
```
wiba-nordic, line 107: Just noticed, but shouldn't the parameter match the actual one?

kkasperczyk-no: I'm not quite sure if that's what you mean, but the `SLEEPY_ACTIVE_THRESHOLD` is a parameter name from the Matter specification, that has some default value and we want to allow changing it using the `CONFIG_CHIP_SED_ACTIVE_THRESHOLD` Kconfig.

wiba-nordic: Ah, that makes sense. Should `SLEEPY_ACTIVE_THRESHOLD` then be mentioned in the documentation?

kkasperczyk-no: We haven't done similar explanations for multiple other terminologies straight from the Matter specification, so I don't want to diverge from it without consulting with Grzegorz. I think that my mistake is that I used this terminology altogether with a style reminding parameter name in the code, so it became confusing. Would it less confusing for you if I rephrase it to the "normal" style?
```suggestion
  * The :ref:`ug_matter_device_low_power_configuration` page with information about sleepy active threshold parameter configuration.
```

wiba-nordic: I think that's a good solution.

wiba-nordic (approved):

LuDuda (approved):

PR #11196, opened 2023-05-16
samples: bluetooth: peripheral_mds: fix heap memory size configuration

maje-emb:
This commit configures the heap size.
The heap size is 4096B when BT_RPMSG is enabled.

Ref: NCSDK-21623

### DIFF
--- a/samples/bluetooth/peripheral_mds/boards/nrf5340dk_nrf5340_cpuapp.conf
+++ b/samples/bluetooth/peripheral_mds/boards/nrf5340dk_nrf5340_cpuapp.conf
@@ -1,0 +1,8 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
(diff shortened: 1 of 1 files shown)


samples/bluetooth/peripheral_mds/boards/nrf5340dk_nrf5340_cpuapp.conf
```diff
@@ -0,0 +1,8 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Heap memory is required for the memfault_demo_cli.c
```
KAGA164, line 7: It is actually no true because 4k of memory is needed by OpenAMP so maybe it should be 4k plus something

maje-emb: Good point.

KAGA164 (approved):

dchat-nordic (approved):

grochu (approved):

PR #11114, opened 2023-05-09
Fast Pair: publish Model IDs and update the documentation

alstrzebonski:
Updated documentation to publish Fast Pair demos' provisioning data.
Added default Fast Pair provisioning data that is used in the Fast Pair sample when no other provisioning data is specified.
Updated the documentation to align with the new way of displaying notifications for Fast Pair debug Model IDs.

Jira: NCSDK-20845
Jira: NCSDK-21329

PR #11201, opened 2023-05-16
doc: fast_pair: Improve provisioning data doc

alstrzebonski:
Adds fast_pair_device.txt to standardize the way of presenting Fast Pair devices in documentation.
This is a follow-up to [#11114 (comment)](https://github.com/nrfconnect/sdk-nrf/pull/11114#discussion_r1192461168) from https://github.com/nrfconnect/sdk-nrf/pull/11114.

Jira: NCSDK-21647

### DIFF
--- a/applications/nrf_desktop/README.rst
+++ b/applications/nrf_desktop/README.rst
@@ -868,19 +868,23 @@ Selecting a build type from command line
 
    * NCS keyboard - The Fast Pair Provider meant to be used with keyboards:
 
-      * Device name: NCS keyboard
-      * Model ID: ``0x52FF02``
-      * Anti-spoofing key (base64, uncompressed): ``8E8ulwhSIp/skZeg27xmWv2SxRxTOagypHrf2OdrhGY=``
-      * Type: Input device
-      * Additional features: Data only connection, Support personalized name, Ringing device unsupported
+      .. |name| replace:: NCS keyboard
+      .. |model ID| replace:: ``0x52FF02``
+      .. |anti-spoofing key| replace:: ``8E8ulwhSIp/skZeg27xmWv2SxRxTOagypHrf2OdrhGY=``
+      .. |type| replace:: Input device
+      .. |additional features| replace:: Data only connection, Support personalized name, Ringing device unsupported
+
+      .. include:: /includes/fast_pair_device.txt
 
    * NCS gaming mouse - Fast Pair Provider meant to be used with gaming mice:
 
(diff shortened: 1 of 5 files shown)


kapi-no (commented):
We seem to have a few sentences of introduction before describing the model specifics:

```
Before you can use your device as a Fast Pair Provider, the device model must be registered with Google.
This is required to obtain Model ID and Anti-Spoofing Private Key.
You can register your own device or use the debug Model ID and Anti-Spoofing Public/Private Key pair obtained by Nordic for development purposes.
```
(Copy paste from Fast Pair sample doc)

Consider moving it to the includes section as well

alstrzebonski:
Hi @divipillai, the documentation here does not build. It's because in nRF Desktop we have a multiple definition of "name" and also other symbols. I tried little bit different approach (commit 2 of this PR) but the result is the same. We need to define "name" twice because we need to include `fast_pair_device.txt` twice with different parameters (as we have two Fast Pair devices in nRF Desktop). I guess we could, as a workaround, have separate `.rst` files for each device and include them (similar solution to what was introduced in commit 2 but with `.rst` files instead of `.txt` files). Would that be possible? I'm afraid it would break the convention as we only include `.txt` files and also it may simply not work. Maybe you have some other idea on how to do it?

kapi-no:
> > We seem to have a few sentences of introduction before describing the model specifics:
> > ```
> > Before you can use your device as a Fast Pair Provider, the device model must be registered with Google.
> > This is required to obtain Model ID and Anti-Spoofing Private Key.
> > You can register your own device or use the debug Model ID and Anti-Spoofing Public/Private Key pair obtained by Nordic for development purposes.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (Copy paste from Fast Pair sample doc)
> > Consider moving it to the includes directory as well and reusing it in the nRF Desktop context
> 
> @kapi-no, I thought about it and I'm not convinced. The messages in sample and nRF Desktop are slightly different because they are introduced in different context. I don't think it would be reusable and I would rather keep it as it is if you're okay with that.

If there is no simple way to change the documentation for nRF Desktop and Fast Pair sample to have the same Fast Pair model guidelines, then I am fine with leaving things how they are now

kapi-no (approved):

alstrzebonski:
I'm closing this PR as there is no way to make the improvement work.

PR #11205, opened 2023-05-16
applications: asset_tracker_v2: Fix race for credential provisioning

jtguggedal:
The Memfault SDK provisions credentials on devices. For nRF91, this needs to happen after modem initialization and before functional mode is set to normal.

Previously there was a race because both the Memfault integration code and the modem module used NRF_MODEM_LIB_ON_INIT hooks to trigger their actions.

Now this is changed so that the modem moule does both operations sequentially. It's not optimal to let the modem module handle the Memfault credentials this way, but the alternatives caused a lot of added code and events between the modem module and debug module, which did not seem worth the trade-off.


applications/asset_tracker_v2/src/modules/modem_module.c
```diff
@@ -702,6 +706,13 @@ static int setup(void)
 {
 	int err;
 
+#if IS_ENABLED(CONFIG_MEMFAULT) && IS_ENABLED(CONFIG_MEMFAULT_NCS_PROVISION_CERTIFICATES)
```
gregersrygg, line 709: The comment in overlay-memfault.conf says the app controls it, but is that elsewhere than here? `CONFIG_MEMFAULT_NCS_PROVISION_CERTIFICATES` is turned off in the overlay, so I don't think this code will run. Do the user need to run extra steps to provision the certificates another way?

de-nordic:
@jtguggedal Pleas fix CI issues and request reviews, if want this to remain candidate for 2.4.0.

applications/asset_tracker_v2/src/modules/modem_module.c
jtguggedal: You're right, the last condition should be `!IS_ENABLED(CONFIG_MEMFAULT_NCS_PROVISION_CERTIFICATES)`. Fixed now

jhn-nordic (approved):

stig-bjorlykke (approved):

gregersrygg (approved):
Can confirm it's working now 👍

PR #11210, opened 2023-05-16
drivers: wifi: Fix sending action frames

krish2718:
For action frames from WPA supplicant "noack" is set to 0 but wait_time is also zero, so, driver prematurely timeout and fails the action frame send.

### DIFF
--- a/drivers/wifi/nrf700x/zephyr/src/zephyr_wpa_supp_if.c
+++ b/drivers/wifi/nrf700x/zephyr/src/zephyr_wpa_supp_if.c
@@ -20,6 +20,8 @@ LOG_MODULE_DECLARE(wifi_nrf, CONFIG_WIFI_LOG_LEVEL);
 
 K_SEM_DEFINE(wait_for_event_sem, 0, 1);
 
+#define ACTION_FRAME_RESP_TIMEOUT_MS 5000
+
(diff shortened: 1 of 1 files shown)


drivers/wifi/nrf700x/zephyr/src/zephyr_wpa_supp_if.c
```diff
@@ -1246,6 +1248,9 @@ int wifi_nrf_nl80211_send_mlme(void *if_priv, const u8 *data,
 	 * to 0 always.
 	 */
 	if (wait_time || !noack) {
+		if (!noack && !wait_time)
+			wait_time = ACTION_FRAME_RESP_TIMEOUT_MS;
```
rlubos, line 1252:
```suggestion
		if (!noack && !wait_time) {
			wait_time = ACTION_FRAME_RESP_TIMEOUT_MS;
		}
```
Not sure why didn't checkpatch catch this...

krish2718: In Linux world braces for single if is prohibited, so, may be its allowing both styles?

rlubos (approved):

rado17 (approved):

PR #11222, opened 2023-05-17
scripts: Moved wget installation from requirements buld to fixed.

kkasperczyk-no:
The wget package was incorrectly added to the requirements-build instead of the requirements-fixed, so in the end the package is not properly installed and customers are struggling with its lack.


de-nordic:
@kkasperczyk-no Please re-trigger CI to make the PR go green.

PR #11224, opened 2023-05-17
matter: applications/samples: Fix using deprecated variable

nordicjm:
Fixes an issue with matter projects whereby it relied upon a user
    having sourced the zephyr environmental file, which has long been
    deprecated.


samples/matter/template/CMakeLists.txt
```diff
@@ -6,12 +6,12 @@
 
 cmake_minimum_required(VERSION 3.20.0)
 
-get_filename_component(MATTER_MODULE_ROOT $ENV{ZEPHYR_BASE}/../modules/lib/matter REALPATH)
+set(MATTER_MODULE_ROOT "\\\${ZEPHYR_BASE}/../modules/lib/matter")
```
kkasperczyk-no, line 9: I'm wondering, if after your change we use `ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR` to point the Matter project location in a tree, do we even need to set `MATTER_MODULE_ROOT`, as it seems to contain the same information?

Damian-Nordic (approved):

samples/matter/template/CMakeLists.txt
nordicjm: It is possible, updated.

PR #11230, opened 2023-05-17
drivers: wifi: Fix output of TX stats

krish2718:
Displaying pending queue len globally but displaying outstanding tokens per-AC does't make sense, move everything to per-AC.

Also, display hardcoded peer (this will change when we AP mode support is in).


de-nordic:
@krish2718 Please rebase and push to fix CI to green
@sachinthegreen @rado17 @rlubos This has been set with 2.4.0 as a target, please review or nack if non-applicable.

sachinthegreen (approved):

rado17 (approved):

PR #11244, opened 2023-05-19
application: serial_lte_modem: Support DTLS connection ID

junqingzou:
Support RFC9146 DTLS v1.2 Connection Identifier.
- Requires lib_modem from NCS v2.4.0 and beyond.
- Requires mfw-1.3.5 and beyond.

Supported in DTLS client proxy #XUDPCLI only.


junqingzou:
Under testing with mfw_1.3.5 pre-release

applications/serial_lte_modem/src/slm_at_udp_proxy.c
```diff
@@ -481,7 +529,7 @@ int handle_at_udp_server(enum at_cmd_type cmd_type)
 }
 
 /**@brief handle AT#XUDPCLI commands
- *  AT#XUDPCLI=<op>[,<url>,<port>[,<sec_tag>]
+ *  AT#XUDPCLI=<op>[,<url>,<port>[,<sec_tag>,<connection_id>]
```
SeppoTakalo, line 532:
```suggestion
 *  AT#XUDPCLI=<op>[,<url>,<port>[,<sec_tag>[,<connection_id>]
```

Connection ID parameter seem optional.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
```diff
@@ -210,6 +230,34 @@ static int do_udp_client_connect(const char *url, uint16_t port)
...
+			 * NRF_SO_SEC_DTLS_CID_STATUS_BIDIRECTIONAL 3
+			 */
+			LOG_INF("DTLS connection ID status: %d", status);
+			if (status != NRF_SO_SEC_DTLS_CID_STATUS_DISABLED) {
+				ret = nrf_setsockopt(proxy.sock, NRF_SOL_SECURE,
+						     NRF_SO_SEC_DTLS_CONN_SAVE,
+						     NULL, 0);
+				if (ret) {
+					LOG_WRN("Save DTLS connection failed: %d", ret);
+				}
```
SeppoTakalo, line 256: Why do we blindly save the session if connection ID is enabled?
I would propose that we don't use that feature as it might run out of memory.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
```diff
@@ -191,6 +192,25 @@ static int do_udp_client_connect(const char *url, uint16_t port)
 			ret = -errno;
 			goto cli_exit;
 		}
+
+		/* Best effort to enable/disable DTLS conn_id and load it */
+		uint32_t dtls_cid = (proxy.conn_id != 0) ?
+			NRF_SO_SEC_DTLS_CID_ENABLED : NRF_SO_SEC_DTLS_CID_DISABLED;
+
+		ret = nrf_setsockopt(proxy.sock, NRF_SOL_SECURE, NRF_SO_SEC_DTLS_CID,
+				     &dtls_cid, sizeof(dtls_cid));
```
SeppoTakalo, line 201:
```suggestion
		if (proxy.conn_id) {
				ret = nrf_setsockopt(proxy.sock, NRF_SOL_SECURE, NRF_SO_SEC_DTLS_CID,
						     &proxy.conn_id, sizeof(proxy.conn_id));
		}				  
```
I propose that we only set the socket option if it is other than zero.
Also require that AT parameter is 0, 1 or 2, so it is already same as what our socket API can take.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
```diff
@@ -191,6 +192,25 @@ static int do_udp_client_connect(const char *url, uint16_t port)
...
+		} else if (proxy.conn_id != 0) {
+			ret = nrf_setsockopt(proxy.sock, NRF_SOL_SECURE,
+					     NRF_SO_SEC_DTLS_CONN_LOAD,
+					     NULL, 0);
+			if (ret) {
+				LOG_WRN("Load DTLS connection failed: %d", ret);
+			} else {
+				LOG_INF("DTLS connection resumed");
+			}
+		}
```
SeppoTakalo, line 213: My proposal would be not to use to storing/loading, at least not in the first phase.

Either drop it, or have it as a separate option / AT command.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
```diff
@@ -516,6 +564,9 @@ int handle_at_udp_client(enum at_cmd_type cmd_type)
 			proxy.sec_tag  = INVALID_SEC_TAG;
 			if (at_params_valid_count_get(&at_param_list) > 4) {
 				at_params_unsigned_int_get(&at_param_list, 4, &proxy.sec_tag);
+				if (at_params_valid_count_get(&at_param_list) > 5) {
+					at_params_int_get(&at_param_list, 5, &proxy.conn_id);
```
SeppoTakalo, line 568: We should verify here that `conn_id` is either 0, 1 or 2, so we can feed it directly to Socket API.

SeppoTakalo (changes requested):
Left my review remarks and proposals.

junqingzou:
> Left my review remarks and proposals.

Sorry to be late in replying due to local expo and tech tour

applications/serial_lte_modem/src/slm_at_udp_proxy.c
junqingzou: I think there will be only ONE session saved on modem side, as only SAVE/LOAD defined and no DELETE, so there should be no incremental memeory usage. SAVE after confirmation of CID active would only update the session data as I assumed.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
junqingzou: You're right, it's optional.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
junqingzou: It is a bit awkward to set TLS_DTLS_CID_SUPPORTED, but will be allowed.
Then we'll need to differentiate TLS_DTLS_CID_SUPPORTED and TLS_DTLS_CID_ENABLED.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
junqingzou: I don't think modem will automatically SAVE/LOAD CID so if APP does not do that, CID might not take effect for DTLS re-handshake.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
junqingzou: Yes we can do that

SeppoTakalo:
Can you verify that there is no misunderstanding of connection ID SAVE/LOAD functionality, because I believe it works differently than what this code is using.
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrfxlib/nrf_modem/doc/sockets.html

My understanding of the documentation is that:
* When you have a socket created, you can enable the CID functionality.
* It might be enabled after you have connected. Depends on the server.
* If you call CONN_SAVE, the existing DTLS session will be serialized into RAM and active connection is dropped to save some RAM from modem.
* Socket is now unusable, until CONN_LOAD is called.
* If you call `close()` on the socket, the saved CID will be removed. Nothing is permanently stored.

junqingzou:
> Can you verify that there is no misunderstanding of connection ID SAVE/LOAD functionality, because I believe it works differently than what this code is using. https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrfxlib/nrf_modem/doc/sockets.html
> 
> My understanding of the documentation is that:
> 
> * When you have a socket created, you can enable the CID functionality.
> * It might be enabled after you have connected. Depends on the server.
// It's enabled in Client Hello so during connect() for DTLS handshake I think (as seen in Wireshark)

> * If you call CONN_SAVE, the existing DTLS session will be serialized into RAM and active connection is dropped to save some RAM from modem.
> * Socket is now unusable, until CONN_LOAD is called.
// This was misunderstood by me. In my test, server does not responded with STATUS_ENABLED so I was not able to test out the socket unusable.

> * If you call `close()` on the socket, the saved CID will be removed. Nothing is permanently stored.

I get your point. I referred to lib_modem socket test code, but it does not really reflect the correct sequence of LOAD and SAVE.

applications/serial_lte_modem/src/slm_at_udp_proxy.c
MarkusLassila: Same CID should not be re-used in DTLS re-handshake.

https://datatracker.ietf.org/doc/html/rfc9146:

In DTLS 1.2, CIDs are exchanged at the beginning of the DTLS session only. There is no dedicated "CID update" message that allows new CIDs to be established mid-session, because DTLS 1.2 in general does not allow TLS 1.3-style post-handshake messages that do not themselves begin other handshakes. **When a DTLS session is resumed or renegotiated, the "connection_id" extension is negotiated afresh.**

Also:
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrfxlib/nrf_modem/doc/sockets.html

NRF_SO_SEC_DTLS_CONN_SAVE
Save DTLS connection. This option is write-only. This option require a DTLS v1.2 connection with renegotiation disabled.

MarkusLassila (changes requested):

SeppoTakalo:
@junqingzou Can you drop the SAVE/LOAD functionality.

Connection Identifier itself does not require it, it should be considered as a separate thing.
To enable CID, you only need to set one socket option. For example https://github.com/nrfconnect/sdk-nrf/pull/11493

junqingzou:
I can actually close this PR and keep the code as local test object.
If someone knows how to test with a known good DTLS server please teach me.
With openssl X.509-based DTLS is OK but server ignores extension 54 in Client_Hello.

PR #11254, opened 2023-05-19
Mfp display

rado17:
Incorrect information of MFP capabilities was getting displayed. This fixes the information that driver passes up to the application.

### DIFF
--- a/drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fw/host_rpu_umac_if.h
+++ b/drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fw/host_rpu_umac_if.h
@@ -2878,6 +2878,9 @@ struct nrf_wifi_umac_event_new_scan_results {
 #define NRF_WIFI_802_11AC (1 << 4)
 #define NRF_WIFI_802_11AX (1 << 5)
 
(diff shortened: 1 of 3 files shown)


drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fw/host_rpu_umac_if.h
```diff
@@ -2889,7 +2892,7 @@ struct umac_display_results {
 	unsigned short capability;
 	struct nrf_wifi_signal signal;
 	unsigned char twt_support;
-	unsigned char reserved2;
+	unsigned char pmf_flag;
```
sachinthegreen, line 2895: Can we stick to one terminology...if we are calling it mfp everywhere then it is better to call this too as mfp_flag.

sachinthegreen (changes requested):

sachinthegreen (approved):

krish2718 (approved):

drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fw/host_rpu_umac_if.h
rado17: Done.

PR #11261, opened 2023-05-22
manifest: Update SDC revision to not include headers when build from src

ppryga-nordic:
Update SDC revision to avoid include nrfxlib/softdevice_controller header files when SDC is build from sources as submodule.

Signed-off-by: Piotr Pryga [piotr.pryga@nordicsemi.no](mailto:piotr.pryga@nordicsemi.no)

### DIFF
--- a/subsys/bluetooth/Kconfig
+++ b/subsys/bluetooth/Kconfig
@@ -16,6 +16,12 @@ config BT_MAX_CONN
 	range 1 20 if BT_LL_SOFTDEVICE
 	default 2 if BT_LL_SOFTDEVICE && BT_CENTRAL && BT_PERIPHERAL
 
+config BT_LL_SOFTDEVICE_HEADERS_INCLUDE
+	bool
+	default y if !BT_CTLR
+	help
+	  Include SoftDevice header files provided with the library.
+
 if BT_CTLR
 rsource "controller/Kconfig"
 endif
--- a/subsys/bluetooth/controller/Kconfig
+++ b/subsys/bluetooth/controller/Kconfig
@@ -47,6 +47,7 @@ choice BT_LL_SOFTDEVICE_BUILD_TYPE
 
 config BT_LL_SOFTDEVICE_BUILD_TYPE_LIB
 	bool "Use library"
+	select BT_LL_SOFTDEVICE_HEADERS_INCLUDE
 
 endchoice
 
--- a/west.yml
+++ b/west.yml
@@ -131,7 +131,7 @@ manifest:
     - name: nrfxlib
       repo-path: sdk-nrfxlib
       path: nrfxlib
-      revision: 7a073076f8df142918e053f7143589d00009e955
+      revision: e9735b27c7b341f6517c1d0c24a1467562f6042f
     - name: trusted-firmware-m
       repo-path: sdk-trusted-firmware-m
       path: modules/tee/tf-m/trusted-firmware-m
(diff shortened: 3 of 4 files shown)


subsys/bluetooth/controller/Kconfig
```diff
@@ -42,11 +42,18 @@ endchoice
 if BT_LL_SOFTDEVICE
 menu "SoftDevice Controller"
 
+config BT_LL_SOFTDEVICE_HEADERS_INCLUDE
+	bool
+	default y if !BT_LL_SOFTDEVICE
```
rugeGerritsen, line 47: The headers should also be included when building as lib
```suggestion
	default y if !BT_LL_SOFTDEVICE || BT_LL_SOFTDEVICE_BUILD_TYPE_LIB
```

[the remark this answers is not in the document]
ppryga-nordic: moved up, thank you for spotting that!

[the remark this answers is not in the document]
rugeGerritsen: We need to move it up one more level so that it can be reused when `BT_CTRL` is not set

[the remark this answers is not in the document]
ppryga-nordic: sorry I should have been spot that! Thank you

rugeGerritsen (approved):

subsys/bluetooth/Kconfig
```diff
@@ -16,6 +16,12 @@ config BT_MAX_CONN
 	range 1 20 if BT_LL_SOFTDEVICE
 	default 2 if BT_LL_SOFTDEVICE && BT_CENTRAL && BT_PERIPHERAL
 
+config BT_LL_SOFTDEVICE_HEADERS_INCLUDE
```
KAGA164, line 19: I have two questions here.
What is a use case for having this option?
Why do we want to use headers by default when SD is not used?

rugeGerritsen: The headers are used to construct vendor specific commands or parse vendor specific events when the host is not located on the same device as the controller. Example: An application runs on the 53 app core and wants to enable QoS reporting. Then it needs the HCI definitions defined by the SDC

KAGA164: Is it ok to have this option enabled when different controller is used? For example BT_LL_SW_SPLIT?

rugeGerritsen: That probably does not make sense. We could probably change this to !BT_CTLR

subsys/bluetooth/Kconfig
```diff
@@ -16,6 +16,12 @@ config BT_MAX_CONN
...
+	bool
+	default y if !BT_LL_SOFTDEVICE && !BT_CTLR
```
rugeGerritsen, line 21: You can remove !BT_LL_SOFTDEVICE as BT_LL_SOFTDEVICE depends on BT_CTLR

ppryga-nordic: done

rugeGerritsen (approved):

KAGA164 (approved):

west.yml
```diff
@@ -131,7 +131,7 @@ manifest:
     - name: nrfxlib
       repo-path: sdk-nrfxlib
       path: nrfxlib
-      revision: 7a073076f8df142918e053f7143589d00009e955
+      revision: pull/990/head
```
rlubos, line 134: `e9735b27c7b341f6517c1d0c24a1467562f6042f`

PR #11272, opened 2023-05-22
manifest: Bugfix for external flash with single slot MCUBoot

sigvartmh:
This is a bugfix where you can't update the network core using only a single secondary slot in external flash. This is due to the fact that the network core update data is inaccessible to the network core in external flash and needs to be copied into the ram of the application core.

Ref: NCSDK-21379

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -128,7 +128,7 @@ manifest:
           compare-by-default: true
     - name: mcuboot
       repo-path: sdk-mcuboot
-      revision: c28768eb2aa7b68e1420b7e260f6139b5b019ebd
+      revision: pull/257/head
       path: bootloader/mcuboot
     - name: qcbor
       url: https://github.com/laurencelundblade/QCBOR.git


de-nordic:
@sigvartmh In 2.4? Missing tagging.

sigvartmh:
Wasn't sure if I should have added the tag as I assumed maybe you would do another manifest update for MCUBoot.

tejlmand (approved):

nordicjm (approved):

shanthanordic:
removed milestone as agreed in release sync

rakons:
nrfconnect/sdk-mcuboot#257 is already merged. Can we update this PR with the SHA?

rakons:
NOTE: This was already in the sdk, but was reverted as it breaks some CI:
Revert PR: https://github.com/nrfconnect/sdk-mcuboot/pull/272

PR #11276, opened 2023-05-22
cmake: Enable link time optimization

balaji-nordic:
Link time optimization is now enabled when the compiler is GNU and the
KConfig option CONFIG_SIZE_OPTIMIZATIONS is enabled.

Application | Compiled for | FLASH usage with LTO | FLASH usage Before LTO | Improvement
-- | -- | -- | -- | --
nrf5340_audio | nrf5340_audio_dk_nrf5340_cpuapp | 578988 | 578988 | 0
atv2 | nrf9160dk_nrf9160_ns | 254012 | 263072 | 9060
atv2 | nrf9160dk_nrf9160_ns -- -DSHIELD=nrf7002ek_nrf7002 -DOVERLAY_CONFIG=overlay-nrf7002ek-wifi-scan-only.conf -D DTC_OVERLAY_FILE="nrf9160dk_with_nrf7002ek.overlay" | 365180 | 376864 | 11684
atv2 | nrf9160dk_nrf9160_ns -- -DSHIELD=nrf7002ek_nrf7002 -DOVERLAY_CONFIG=overlay-nrf7002ek-wifi-scan-only.conf;overlay-debug.conf -D DTC_OVERLAY_FILE="nrf9160dk_with_nrf7002ek.overlay" | OVERFLOW by 5584 bytes | OVERFLOW by 17 Kbytes |  
nrf_desktop | nrf52833dk_nrf52820 | 227680 | 227620 | -60
zigbee_weather_station | thingy53_nrf5340_cpuapp | 275260 | 275324 | 64
machine_learning | thingy53_nrf5340_cpuapp | 491996 | 491996 | 0
serial_lte_modem | nrf9160dk_nrf9160_ns | 245072 | 244524 | -548

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,11 @@ set(NRF_DIR ${CMAKE_CURRENT_LIST_DIR} CACHE PATH "NCS root directory")
 # is cached.
 set(CONF_FILE_BUILD_TYPE ${CONF_FILE_BUILD_TYPE} CACHE INTERNAL "The build type")
 
+# Enable Link Time Optimization
+if(CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CONFIG_SIZE_OPTIMIZATIONS)
+	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -flto")
+endif()
+
 include(cmake/extensions.cmake)
 include(cmake/version.cmake)
 include(cmake/multi_image.cmake)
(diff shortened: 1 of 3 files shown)


balaji-nordic:
Closing because this only gives significant improvement for atv2 application. So I will do the changes only for that application instead of the entire ncs tree.

dberlin:
Random driveby from a former GCC optimization maintainer  - One thing to keep in mind - without -fwhole-program (which is not the same as -flto), it does not know that it has the complete binary at link time.  It assumes it could be relinked further, etc.  (The linker flags/etc to discard unused functions don't turn on whole-program).   You should see some small additional size benefit from adding -fwhole-program as well.  It will depend on how much was being discarded anyway.

It also enables it to make stronger assumptions about aliasing across function calls, and do things like whole-program devirtualization in the case of C++ (since it can now assume it has the complete set of classes, and nothing will be further overridden).

For applications, it will help more, since people rarely pay strong attention to visibility, etc.

It is "dangerous" in the non-embedded world becuase of dynamic linkers, etc, but in the embedded world, it should be safe.

PR #11279, opened 2023-05-23
nrf_802154: nrf_802154_serialization_error never returns

ankuns:
Functions nrf_802154_serialization_error and
nrf_802154_sl_fault_handler shall never return.
However `__ASSERT()` might expand to an empty instruction CONFIG_ASSERT=n and is not enough to stop execution.

Similar changes to 802154_rpmsg app and the shim ieee802154_nrf5.c in upstream zephyr https://github.com/zephyrproject-rtos/zephyr/pull/58173 - they are expected to be available on regular upmerge.

### DIFF
--- a/samples/nrf5340/multiprotocol_rpmsg/src/main.c
+++ b/samples/nrf5340/multiprotocol_rpmsg/src/main.c
@@ -265,9 +265,11 @@ void nrf_802154_serialization_error(const nrf_802154_ser_err_data_t *err)
 {
 	(void)err;
 	__ASSERT(false, "802.15.4 serialization error");
+	k_oops();
 }
 
 void nrf_802154_sl_fault_handler(uint32_t id, int32_t line, const char *err)
 {
 	__ASSERT(false, "module_id: %u, line: %d, %s", id, line, err);
+	k_oops();
 }


ankuns:
FYI: @jciupis @ahasztag @mariuszpos @czeslawmakarski

samples/nrf5340/multiprotocol_rpmsg/src/main.c
```diff
@@ -265,9 +265,11 @@ void nrf_802154_serialization_error(const nrf_802154_ser_err_data_t *err)
 {
 	(void)err;
 	__ASSERT(false, "802.15.4 serialization error");
+	k_panic();
```
hubertmis, line 268: > A kernel error is a software triggered fatal error invoked by `k_panic()`. This should be used to indicate that the Zephyr kernel is in an unrecoverable state.

I don't think that serialization error should be treated as a kernel unrecoverable state. It is possible that other threads run just fine while serialization crashes.

Maybe `k_oops()` instead? Serialization seems closer to application logic than to kernel state

> A kernel oops is a software triggered fatal error invoked by `k_oops()`. This should be used to indicate an unrecoverable condition in application logic.

hubertmis: Also from fatal_types.h:
```
        /** Moderate severity software error */
        K_ERR_KERNEL_OOPS,

        /** High severity software error */
        K_ERR_KERNEL_PANIC,
```

ankuns: Changed to `k_oops` as you suggest, as well as in upstream PR https://github.com/zephyrproject-rtos/zephyr/pull/58173

PR #11318, opened 2023-05-24
samples: Bluetooth: Mesh: fix id collision between ble and ble mesh gatt services

alxelax:
BLE and BLE MESH GATT services used the same id. That caused the situation when mobile app wasn't able to connect to mesh proxy if ble gatt is active.

### DIFF
--- a/samples/bluetooth/mesh/ble_peripheral_lbs_coex/src/lb_service_handler.c
+++ b/samples/bluetooth/mesh/ble_peripheral_lbs_coex/src/lb_service_handler.c
@@ -40,6 +40,8 @@ static const struct bt_data sd[] = {
 	BT_DATA_BYTES(BT_DATA_UUID128_ALL, BT_UUID_LBS_VAL),
 };
 
+static struct bt_le_ext_adv *adv;
+static struct bt_le_ext_adv_start_param ext_adv_param;
 static bool ble_button_state;
 
 static void connected(struct bt_conn *conn, uint8_t err)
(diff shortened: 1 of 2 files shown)


samples/bluetooth/mesh/ble_peripheral_lbs_coex/src/lb_service_handler.c
```diff
@@ -40,6 +40,8 @@ static const struct bt_data sd[] = {
 	BT_DATA_BYTES(BT_DATA_UUID128_ALL, BT_UUID_LBS_VAL),
 };
 
+static struct bt_le_ext_adv *adv;
+static struct bt_le_ext_adv_start_param ext_adv_param = { .num_events = 0, .timeout = 0 };
```
PavelVPV, line 44: Minor: no need to initialize to zero static structs.

PavelVPV (approved):

alxelax:
Had to create shorter commit name

Balaklaka (approved):

samples/bluetooth/mesh/ble_peripheral_lbs_coex/src/lb_service_handler.c
```diff
@@ -89,14 +91,47 @@ static void button_changed(uint32_t button_state, uint32_t has_changed)
...
 	}
 
-	printk("Advertising successfully started\n");
+	err = bt_le_ext_adv_set_data(adv, ad, ARRAY_SIZE(ad), sd, ARRAY_SIZE(sd));
+	if (err) {
+		printk("Setting LBS service adv data failed (err %d)\n", err);
+		return;
+	}
+
+	err = bt_le_ext_adv_start(adv, &ext_adv_param);
```
akredalen, line 129: Could BT_LE_EXT_ADV_START_DEFAULT be passed here instead and then remove line 44?

akredalen (approved):
LGTM. Just a small suggestion.

samples/bluetooth/mesh/ble_peripheral_lbs_coex/src/lb_service_handler.c
alxelax: I'll add it in a small separate PR to not postpone the fix for the release.

ludvigsj (approved):

PavelVPV:
One small comment. The following code should be added to connection callbacks: https://github.com/nrfconnect/sdk-nrf/blob/8c420a58e74c1b5f138d87d5435b213ee8dea95f/samples/bluetooth/mesh/common/smp_bt.c#L54-L65
This doesn't change anything in the sample but important for users to know that all connections will be caught by these callbacks but not all are related to this advertisement.

PR #11319, opened 2023-05-24
applications: Fixed ICD subscriptions handling for weather station

kkasperczyk-no:
Support for ICD subscriptions doesn't work for weather station, because it was only enabled in Kconfig and there is not necessary code implemented to handle it in application.


LuDuda (approved):

kkasperczyk-no:
> Tested with ecosystem and the max interval value is not increased to 60 seconds

@doublemis1 how did you verify it? I don't have PPK to verify, but I can see the following logs, so I assume it should work:

> [00:01:39.338,012] <dbg> chip: [IM]Received Subscribe request
[00:01:39.338,989] <inf> chip: [DMG]Final negotiated min/max parameters: Min = 0s, Max = 60s

and 

> [00:01:41.951,354] <inf> chip: [DMG]Refresh Subscribe Sync Timer with min 0 seconds and max 60 seconds

doublemis1 (approved):

PR #11326, opened 2023-05-25
net: lib: lwm2m_client_utils: Minor bugfixes

jarlamsa:
Fix 2 coverity issues from location assistance.
Unnecessary header and unused return value.


SeppoTakalo (approved):
OK.

Just a side note: It is not a coverity fix, if it fixes a real problem. Like in this case. I would much prefer to see from the commit message that what was actually fixed. It also tells whether the developer understood what was the issue, or is he/she just blindy changing the code to silence the warnings from the tool.

rlubos (approved):

PR #11334, opened 2023-05-25
applications: asset_tracker_v2: Remove PSM handling from the application

simensrostad:
Remove PSM handling from the application and depend on the
Link Controller library to request PSM through the
`CONFIG_LTE_PSM_REQ` option.

This is to prevent conflict when having PSM be requested in
multiple places at the same time and to have a single point of handling.

Without this PR, the current version of ATv2 does not request PSM. Maintaining low power is one of the main design pillars of the application so this should be considered as a critical bug.


jhn-nordic (approved):

tokangas (commented):
`configure_low_power()` is also called from here: https://github.com/nrfconnect/sdk-nrf/blob/5c370a4b806cc879323c2c3f89ee855a41190fdf/applications/asset_tracker_v2/src/modules/modem_module.c#L792

So, either the function has to stay, or `lte_lc_psm_req()` needs to be called directly from `on_all_states()`.

jtguggedal (approved):

umapraseeda (approved):

applications/asset_tracker_v2/src/modules/modem_module.c
```diff
@@ -815,10 +789,10 @@ static void on_all_states(struct modem_msg_data *msg)
 	if (IS_EVENT(msg, cloud, CLOUD_EVT_USER_ASSOCIATED)) {
 
 		/* Re-enable low power features after cloud has been associated. */
-		int err = configure_low_power();
+		int err = lte_lc_psm_req(true);
```
tokangas, line 792: Should it be possible to disable PSM by setting `CONFIG_LTE_PSM_REQ=n`? If yes, then this would still enable PSM after user association, so you could check the Kconfig option here.

simensrostad: Good catch, fixed

PR #11349, opened 2023-05-25
asset_tracker_v2: Enable link time optimization

balaji-nordic:
Enable link time optimization for the application code. This is intended
for reducing the flash usage.

The flash usage for the asset tracker application reduces by 1576 bytes
when built for nrf9160dk_nrf9160_ns platform (without any other
additional options). The RAM usages increases by 48 bytes.

I tried enabling this globally across NCS in this PR https://github.com/nrfconnect/sdk-nrf/pull/11276 . The improvement there was significant (9k) but only for asset tracker v2 (see the table in that PR).  The other applications had negligible or even negative improvement. So I decided to only enable this for the application code for asset tracker v2 application.

### DIFF
--- a/applications/asset_tracker_v2/CMakeLists.txt
+++ b/applications/asset_tracker_v2/CMakeLists.txt
@@ -39,3 +39,7 @@ if (CONFIG_BOARD_QEMU_X86 OR CONFIG_BOARD_NATIVE_POSIX)
         target_compile_options(app PRIVATE
 	        -DCONFIG_LTE_NEIGHBOR_CELLS_MAX=10)
 endif()
+
+if(CONFIG_ASSET_TRACKER_V2_LTO)
+        target_compile_options(app PRIVATE "-flto")
+endif()
--- a/applications/asset_tracker_v2/Kconfig
+++ b/applications/asset_tracker_v2/Kconfig
@@ -19,6 +19,14 @@ config ASSET_TRACKER_V2_APP_VERSION_MAX_LEN
 	int "Maximum length of the application firmware version"
 	default 150
 
+config ASSET_TRACKER_V2_LTO
+	bool "Enable link time optimization"
+	default y if SIZE_OPTIMIZATIONS
+	help
+	  Compile the Asset Tracker application code with link time optimization enabled. This
+	  option is only applied for the application code and not the libraries and external modules
+	  that are linked in.
+
 rsource "src/modules/Kconfig.modules_common"
 rsource "src/modules/Kconfig.cloud_module"
 rsource "src/cloud/Kconfig.lwm2m_integration"
--- a/doc/nrf/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases/release-notes-changelog.rst
@@ -134,7 +134,10 @@ This section provides detailed lists of changes by :ref:`application <applicatio
 nRF9160: Asset Tracker v2
 -------------------------
 
-|no_changes_yet_note|
+* Updated:
+
+  * Enabled link time optimization to reduce the flash size of the application.
+    You can disable this using the Kconfig option :kconfig:option:`CONFIG_ASSET_TRACKER_V2_LTO`.
 
 nRF9160: Serial LTE modem
 -------------------------


applications/asset_tracker_v2/CMakeLists.txt
```diff
@@ -39,3 +39,8 @@ if (CONFIG_BOARD_QEMU_X86 OR CONFIG_BOARD_NATIVE_POSIX)
         target_compile_options(app PRIVATE
 	        -DCONFIG_LTE_NEIGHBOR_CELLS_MAX=10)
 endif()
+
+# Enable Link Time Optimization for the application code
+if(CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CONFIG_SIZE_OPTIMIZATIONS)
+        target_compile_options(app PRIVATE "-flto")
```
tejlmand, line 45: applying this locally on an app is a good starting point.

Using it solely on sources in app shall still link other Zephyr libraries as regular libs.
Just note that `lto` on linking will be inferred:
> GCC automatically performs link-time optimization if any of the objects involved were compiled with the -flto command-line option. 

Remember to careful investigated that referenced but KEEP sections in the object file are not pruned, for example functions referenced through SYS_INIT.

applications/asset_tracker_v2/CMakeLists.txt
```diff
@@ -39,3 +39,8 @@ if (CONFIG_BOARD_QEMU_X86 OR CONFIG_BOARD_NATIVE_POSIX)
         target_compile_options(app PRIVATE
 	        -DCONFIG_LTE_NEIGHBOR_CELLS_MAX=10)
 endif()
+
+# Enable Link Time Optimization for the application code
+if(CMAKE_C_COMPILER_ID STREQUAL "GNU" AND CONFIG_SIZE_OPTIMIZATIONS)
```
tejlmand, line 44: I think that in order to easily disable lto on compilation units, in case users notice unexpected behavior, then create a app specific setting here: https://github.com/nrfconnect/sdk-nrf/blob/main/applications/asset_tracker_v2/Kconfig

in the form of (feel free to adjust name / location, etc.:
```
config ATV_V2_LTO_OPTIMIZATION
        bool "Compile app library with lto"
        default y if SIZE_OPTIMIZATIONS
        help
          Compile source code in app library with lto.
```

tejlmand (changes requested):

applications/asset_tracker_v2/CMakeLists.txt
balaji-nordic: I have investigated this in two ways.

- I had a look at the object dump of zephyr.elf. I noticed that the functions referenced through SYS_INIT do get included in the final binary. The structs named `__init_FUNCTION` do get an entry in the final hex.  The `__used` keyword in https://github.com/zephyrproject-rtos/zephyr/blob/44e691e20e0b323ad67227ee072bd4166c75d894/include/zephyr/init.h#L189 ensures this.
- Performed a test locally to see that the button presses do get sent to the cloud. The button initialization happens during init (they are registered using SYS_INIT here https://github.com/nrfconnect/sdk-nrf/blob/a46fec90210e725382d8eded1871013e0561ac1a/applications/asset_tracker_v2/src/modules/ui_module.c#L659 )

applications/asset_tracker_v2/Kconfig
```diff
@@ -19,6 +19,14 @@ config ASSET_TRACKER_V2_APP_VERSION_MAX_LEN
 	int "Maximum length of the application firmware version"
 	default 150
 
+config ASSET_TRACKER_V2_LTO
+	bool "Enable link time optimization"
+	default y if SIZE_OPTIMIZATIONS
+	help
+		Compile the asset tracker application code with link time optimization enabled. This
```
peknis, line 26:
```suggestion
		Compile the Asset Tracker application code with link time optimization enabled. This
```

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -134,7 +134,10 @@ This section provides detailed lists of changes by :ref:`application <applicatio
 nRF9160: Asset Tracker v2
 -------------------------
 
-|no_changes_yet_note|
+* Updated:
+
+  * Enabled link time optimization to reduce the flash size of the application. This can be disabled
+  by using the Kconfig option :kconfig:option:`CONFIG_ASSET_TRACKER_V2_LTO`.
```
peknis, line 140:
```suggestion
  * Enabled link time optimization to reduce the flash size of the application.
    You can disable this using the Kconfig option :kconfig:option:`CONFIG_ASSET_TRACKER_V2_LTO`.
```

peknis (changes requested):

peknis (changes requested):

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -134,7 +134,9 @@ This section provides detailed lists of changes by :ref:`application <applicatio
 nRF9160: Asset Tracker v2
 -------------------------
 
-|no_changes_yet_note|
+* Updated:
+
+  * Enabled link time optimization to reduce the flash size of the application. You can disable this using the Kconfig option :kconfig:option:`CONFIG_ASSET_TRACKER_V2_LTO`.
```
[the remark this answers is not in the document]
balaji-nordic: I see. Done. :)

peknis (approved):

bjda (approved):

jhn-nordic (approved):

applications/asset_tracker_v2/Kconfig
```diff
@@ -19,6 +19,14 @@ config ASSET_TRACKER_V2_APP_VERSION_MAX_LEN
 	int "Maximum length of the application firmware version"
 	default 150
 
+config ASSET_TRACKER_V2_LTO
```
jostn, line 22: Other libraries/applications may also enable LTO. Should it be mentioned that LTO during linking is applied to the entirety of the files compiled with LTO?

balaji-nordic: Thanks for the comment @jostn  . Do you mean something like this 

> If object files containing GIMPLE bytecode are stored in a library archive, say libfoo.a, it is possible to extract and use them in an LTO link if you are using a linker with plugin support. To create static libraries suitable for LTO, use gcc-ar and gcc-ranlib instead of ar and ranlib; to show the symbols of object files with GIMPLE bytecode, use gcc-nm. Those commands require that ar, ranlib and nm have been compiled with plugin support. At link time, use the flag -fuse-linker-plugin to ensure that the library participates in the LTO optimization process:

from https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html  ? If so, I would leave the compiler documentation help developers with this additional info. Perhaps pointing them to this compiler doc may help? But I am afraid that the web links tend to get broken. So I prefer that we leave it to (advanced developers) to find this out themselves.

applications/asset_tracker_v2/Kconfig
```diff
@@ -19,6 +19,14 @@ config ASSET_TRACKER_V2_APP_VERSION_MAX_LEN
...
+	bool "Enable link time optimization"
+	default y if SIZE_OPTIMIZATIONS
+	help
+		Compile the Asset Tracker application code with link time optimization enabled. This
+		option is only applied for the application code and not the libraries and external modules
+		that are linked in.
```
tejlmand, line 28:
```suggestion
	  Compile the Asset Tracker application code with link time optimization enabled. This
	  option is only applied for the application code and not the libraries and external modules
	  that are linked in.
```
Just a single `<TAB>` and then two spaces in the Kconfig help text.

tejlmand (approved):
Looks good.
Noticed a minor nit in the Kconfig help text.

PR #11357, opened 2023-05-26
lib: identity_key: Do panic instead of assert

vili-nordic:
In case of failures in the identity key library we want to panic, and not use asserts that can be compiled out.

### DIFF
--- a/lib/identity_key/identity_key.c
+++ b/lib/identity_key/identity_key.c
@@ -27,7 +27,10 @@ static void generate_random_secp256r1_private_key(uint8_t *key_buff)
 
 	/* Initialize PSA Crypto */
 	status = psa_crypto_init();
-	__ASSERT(status == PSA_SUCCESS, "psa_crypto_init failed! Error: %d", status);
+	if (status != PSA_SUCCESS) {
+		printk("psa_crypto_init failed! Error: %d", status);
+		k_panic();
+	}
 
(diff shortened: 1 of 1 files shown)


lemrey (dismissed):
yeah `assert` is not an error check macro :)
but `k_panic` is not a reasonable behavior in case of errors either. That litelly crashes the application, all of it. You should return an error from this library and let the application decide what is the most sensible thing to do to handle this error instead.

vili-nordic:
> yeah `assert` is not an error check macro :) but `k_panic` is not a reasonable behavior in case of errors either. That litelly crashes the application, all of it. You should return an error from this library and let the application decide what is the most sensible thing to do to handle this error instead.

We will create a follow task up for this. For now, this is just to make implementation consistent with documentation (independent of CONFIG_ASSERT). We also have another library (hw_unique_key) with the same issue.

Vge0rge (approved):

lemrey:
I removed the -1 but a failure to initialize a library is not a reason to panic the kernel; panicking prevents the application from doing anything, blinking an LED, printing an error message for the user etc. it is the exact opposite of what it should do, trying to fail gracefully.

vili-nordic:
We will look into this later in NCSDK-21839

frkv (approved):
LGTM

PR #11374, opened 2023-05-30
Bluetooth: mesh: add Kconfigs for NLC profiles

akredalen:
Added Kconfig options for NLC profiles for redefining default values and/or ranges for corresponding mesh Kconfig options.


subsys/bluetooth/mesh/Kconfig
```diff
@@ -51,6 +51,53 @@ endif # BT_MESH_RELAY
...
 
+config BT_MESH_NLC_COMMON_PROFILE
+	bool "Enable common NLC performance configurations"
+	default n
+	depends on BT_MESH
+	help
+	  Enables common performance configurations required for the
+	  NLC Basic Scene Selector, Dimming Control, Energy Monitor,
+	  Occupancy Sensor and Ambient Light Sensor profiles. For the
+	  NLC Basic Lightness Controller, use BT_MESH_NLC_LIGHTNESS_CTRL
```
PavelVPV, line 62:
```suggestion
	  NLC Basic Lightness Controller, use CONFIG_BT_MESH_NLC_LIGHTNESS_CTRL
```

subsys/bluetooth/mesh/Kconfig
```diff
@@ -51,6 +51,53 @@ endif # BT_MESH_RELAY
...
+
+config BT_MESH_NLC_LIGHTNESS_CTRL
+	bool "Enable NLC Basic Lightness Controller"
+	default n
+	depends on BT_MESH
+	help
+	  Enables required performance configurations for the
+	  NLC Basic Lightness Controller. The NLC Basic Lightness Controller
+	  share the same performance configurations as for the common NLC profiles
+	  (BT_MESH_NLC_COMMON_PROFILE) but with more strict requirements for a
```
PavelVPV, line 73:
```suggestion
	  (CONFIG_BT_MESH_NLC_COMMON_PROFILE) but with more strict requirements for a
```

subsys/bluetooth/mesh/Kconfig
```diff
@@ -51,6 +51,53 @@ endif # BT_MESH_RELAY
...
+	depends on BT_MESH
+	help
+	  Enables common performance configurations required for the
+	  NLC Basic Scene Selector, Dimming Control, Energy Monitor,
+	  Occupancy Sensor and Ambient Light Sensor profiles. For the
+	  NLC Basic Lightness Controller, use BT_MESH_NLC_LIGHTNESS_CTRL
+	  instead.
+
+config BT_MESH_NLC_LIGHTNESS_CTRL
+	bool "Enable NLC Basic Lightness Controller"
```
PavelVPV, line 66:
```suggestion
	bool "Enable NLC Basic Lightness Controller performance configuration"
```

PavelVPV:
> According to the DoD, the task was to solve the redefinition of default values and ranges in the sample .config-files, which is why I ended up creating a common config option for 5 out of 6 NLC profiles. If we want to add for example the required models as `select`'s to the NLC configs as well, then it would make more sense to have one config for each. Let me know what you think.

Then, perhaps, it would make sense to add `_PERF_` suffix for the new options to make clear that these options only touch performance part, but not models?
`BT_MESH_NLC_COMMON` -> `BT_MESH_NLC_PERF_COMMON`.
`BT_MESH_NLC_LIGHTNESS_CTRL` -> `BT_MESH_NLC_PERF_LIGHTNESS_CTRL`.

subsys/bluetooth/mesh/Kconfig
```diff
@@ -51,6 +51,68 @@ endif # BT_MESH_RELAY
 
 endif
 
+config BT_MESH_NLC_PERF_SUPPORT
+	bool "Enable NLC performance configurations"
+	help
+	  Enables performance configurations required for NLC profiles.
```
PavelVPV, line 57:
```suggestion
	  Enables performance configurations required for NLC profiles to be qualifiable.
```
?

subsys/bluetooth/mesh/Kconfig
```diff
@@ -51,6 +51,68 @@ endif # BT_MESH_RELAY
...
+	  enable CONFIG_BT_MESH_NLC_PERF_LIGHTNESS_CTRL.
+
+choice BT_MESH_NLC_PERF_OPTIONS
+	prompt "NLC configuration options"
+	depends on BT_MESH_NLC_PERF_SUPPORT
+
+config BT_MESH_NLC_PERF_DEFAULT
+	bool "Default"
+	help
+	  Enables performance configurations required for the NLC Basic Lightness Controller,
```
PavelVPV, line 69: `NLC Basic Lightness Controller` got here accidentally?

subsys/bluetooth/mesh/Kconfig
```diff
@@ -51,6 +51,68 @@ endif # BT_MESH_RELAY
 
 endif
 
+config BT_MESH_NLC_PERF_SUPPORT
```
PavelVPV, line 54: A bit uncertain now after looking the whole code. May be instead of `_SUPPORT` it should be `_CONF` (configuration). If you disagree, then leave it as is.

PavelVPV (commented):
Looks better with the choice I think. Few minor questions. Rest looks OK to me.

subsys/bluetooth/mesh/Kconfig
akredalen: Yes, sorry, copy-paste.

subsys/bluetooth/mesh/Kconfig
ludvigsj: Should we wait to mention this until they are actually possible to qualify?

PavelVPV: Right, good point

ludvigsj (changes requested):
See conversation above, we should remove the "qualifiable" part until qualification is actually possible. Looks good otherwise.

ludvigsj (changes requested):
I thought of a thing: Now that we have the NLC profiles user guide page, should we mention these options on that page so that people know that they should use them?

doc/nrf/protocols/bt_mesh/nlc.rst
```diff
@@ -7,6 +7,9 @@ Bluetooth® Networked Lighting Control (NLC) profiles are a set of device profil
 The NLC profiles can be used to implement interoperable network controlled lighting setups, including sensors, light fixtures, energy monitoring, scene selectors and dimmer controls.
 Each of the profiles specifies a set of models and a set of performance parameters.
 
+Enable the :c:macro:`CONFIG_BT_MESH_NLC_PERF_CONF` macro to set the required performance configurations for the NLC profiles.
+When implementing the NLC Basic Lightness Controller profile individually or in conjunction with other NLC profiles, also enable the :c:macro:`CONFIG_BT_MESH_NLC_PERF_LIGHTNESS_CTRL` macro.
```
mia-ko, line 11:
```suggestion
Enable the :kconfig:option:`CONFIG_BT_MESH_NLC_PERF_CONF` Kconfig option to set the required performance configurations for the NLC profiles.
When implementing the NLC Basic Lightness Controller profile, individually or in conjunction with other NLC profiles, also enable the :kconfig:option:`CONFIG_BT_MESH_NLC_PERF_LIGHTNESS_CTRL` Kconfig option.
```

mia-ko (changes requested):

mia-ko (approved):

ludvigsj (approved):

PR #11383, opened 2023-05-31
clang-format: Sync clang-format file with Zephyr

SebastianBoe:
cp ~/ncs/zephyr/.clang-format ~/ncs/nrf/.clang-format.

AFAIK we copied Zephyr's .clang-format file a long time ago and have since been maintaining a copy of it.

It's not clear to me why we are maintaining a copy instead of syncing with Zephyr.

I'm assuming this is a mistake.

In this commit we sync with Zephyr.

AFAIK we do not intentionally maintain different formatting from Zephyr.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,80 +1,41 @@
-# SPDX-License-Identifier: GPL-2.0
+# SPDX-License-Identifier: Apache-2.0
 #
-# clang-format configuration file. Intended for clang-format >= 4.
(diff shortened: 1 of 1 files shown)


gmarull (approved):

SebastianBoe:
relates to https://github.com/nrfconnect/sdk-ic-next/pull/401#issuecomment-1557349427

carlescufi (approved):

PR #11390, opened 2023-06-01
lib: edge_impulse: Update to new Kconfig options

emob-nordic:
Update dependencies in edge impulse to avoid warnings about options being deprecated.


lib/edge_impulse/Kconfig
```diff
@@ -6,9 +6,9 @@
 
 menuconfig EDGE_IMPULSE
 	bool "Enable Edge Impulse"
-	depends on CPLUSPLUS
+	depends on CPP
 	depends on STD_CPP11
-	depends on LIB_CPLUSPLUS
+	depends on EXTERNAL_LIBCPP
```
MarekPieta, line 11: The `CONFIG_LIB_CPLUSPLUS` selects `CONFIG_REQUIRES_FULL_LIBCPP` (which would enable `CONFIG_GLIBCXX_LIBCPP`). I wonder if we can use `CONFIG_EXTERNAL_LIBCPP` instead here (is the external C++ standard library actually provided in our case)?
Seemingly everything builds without it so consider double checking with Edge Impulse team if it's still needed (keep in mind that it may be used e.g. only by subset of machine learning algorithms).

MarekPieta (approved):

nvlsianpu (approved):

doc/nrf/external_comp/edge_impulse.rst
```diff
@@ -79,18 +79,17 @@ Complete the following steps to configure the building process:
...
-   * :kconfig:option:`CONFIG_NEWLIB_LIBC_FLOAT_PRINTF`
 
    .. note::
       The :kconfig:option:`CONFIG_FPU` Kconfig option is implied by default if floating point unit (FPU) is supported by the hardware.
       Using FPU speeds up calculations.
 
 #. Make sure that the :kconfig:option:`CONFIG_FP16` Kconfig option is disabled.
    The Edge Impulse library is not compatible with half-precision floating point support introduced in Zephyr.
+#. The Edge Impulse library requires full C++ standard library implementation.
+   Because of that :kconfig:option:`CONFIG_EDGE_IMPULSE` option selects :kconfig:option:`CONFIG_REQUIRES_FULL_LIBCPP` option.
```
divipillai, line 92:
```suggestion
   The Edge Impulse library requires a full C++ standard library implementation.
   Because of that, the :kconfig:option:`CONFIG_EDGE_IMPULSE` option selects the :kconfig:option:`CONFIG_REQUIRES_FULL_LIBCPP` option.
```
Since, the step is not a user action, removed the sentence from the numbered list.

divipillai (approved):

PR #11410, opened 2023-06-02
doc: Thread: created types page, moved sed vs ssed

wiba-nordic:
Added dedicated pages for Thread device types, as well as SED and SSED.

### DIFF
--- a/doc/nrf/protocols/thread/device_types.rst
+++ b/doc/nrf/protocols/thread/device_types.rst
@@ -1,0 +1,133 @@
+.. _thread_device_types:
+
+Thread device types
+###################
+
+.. contents::
+   :local:
+   :depth: 2
+
+Thread devices can be either Full Thread Devices (FTD) or Minimal Thread Devices (MTD).
+The devices are divided based on their ability to exist in a network without a parent, determined by whether the device maintains a routing table (doesn't need a parent) or not (needs a parent).
+FTDs are the devices that maintain a routing table, while MTDs do not, forwarding all their messages to a parent instead.
+
+Another type of Thread device classification is based on the device's ability to route messages:
+
+* Routers can route messages, and they can function as parents to other Thread devices.
+* End Devices cannot route messages, and they cannot function as parents.
+
+Based on this classification, FTDs can be either Routers or End Devices, while MTDs can only be End Devices.
+
+For additional information about Thread device types, see the documentation for `Device Types on OpenThread portal`_.
+
+.. _thread_types_ftd:
+
+Full Thread Devices
+*******************
+
+In order to maintain the routing table, FTDs keep their radio on at all times.
+An important difference between FTDs and MTDs is that FTDs subscribe to special "all-routers" multicast addresses (``ff02::2`` and ``ff03::2``, see `IPv6 multicast addressing`_).
+Because of these reasons, they typically consume more power than MTDs.
+
+Full Thread Devices can be further divided into three categories:
+
+* Router
+* Full End Device (FED)
+* Router Eligible End Device (REED)
+
+.. list-table:: Full Thread Device categories
+   :widths: 15 10 10 15
+   :header-rows: 1
+
+   * -
+     - Router
+     - Full End Device
+     - Router Eligible End Device
+   * - Maintains a routing table
+     - Yes
+     - Yes
+     - Yes
+   * - Routes messages
+     - Yes
+     - No
+     - When working as a Router
+   * - Can extend the network
+     - Yes
+     - No
+     - When working as a Router
+
+Routers can be regarded as the backbone of a Thread network.
+They maintain the routing table and forward messages to other devices.
+Routers, unlike End Devices, can also be used to extend the network range and a Router is required for a Minimal End Device to join the network.
+The maximum number of Routers in a single Thread network is 32.
+
+FEDs maintain a routing table, but they can't route messages.
+This means they don't require a parent (Router) to function, but they cannot become parents themselves.
+
+REEDs maintain a routing table and, unlike FEDs, they can become Routers if needed.
+If a REED is the only device in range of an End Device trying to join the network, it will promote itself to a Router (see `Joining an existing Thread network`_ in the OpenThread documentation for more information).
+Conversely, when a Router has no children it can downgrade itself to a REED.
+
+.. _thread_types_mtd:
+
+Minimal Thread Devices
+**********************
+
+Minimal Thread Devices (MTDs) are devices that do not maintain a routing table and are typically low-power devices that are not always on.
+They can only be End Devices, and they always need a parent to function.
+They forward all their messages to their parent.
+
+Minimal Thread Devices can be further divided into three categories:
+
+* Minimal End Device (MED)
+* Sleepy End Device (SED)
+* Synchronized Sleepy End Device (SSED)
+
+.. list-table:: Minimal Thread Device categories
+   :widths: 15 10 10 15
+   :header-rows: 1
+
+   * -
+     - Minimal End Device
+     - Sleepy End Device
+     - Synchronized Sleepy End Device
+   * - Maintains a routing table
+     - No
+     - No
+     - No
+   * - Radio on
+     - Always
+     - Periodically
+     - Periodically
+   * - New messages
+     - Parent immediately forwards the messages
+     - The SED polls for new messages when it wakes up
+     - Parent forwards the messages during designated transmission window
+   * - No new messages
+     - No data transmission
+     - Parent indicates no pending messages
+     - No data transmission
+
+MEDs are the most basic MTDs, and their radio is always on.
+They don't keep a routing table, but otherwise operate like :ref:`FEDs <thread_types_ftd>`.
+
+SEDs try to limit their power consumption by sleeping most of the time, waking up periodically to poll for messages from their parent.
+After waking up, they send a data request to their parent.
+If the parent has any pending messages, it will send them to the SED.
+Otherwise, the parent will send a response indicating no pending messages.
+
+SSEDs operate similarly to Sleepy End Devices, but they are synchronized with their parent.
+They wake up at the same time as their parent, eliminating the need for polling for messages.
+If the parent has messages for the SSED, it sends them during the designated transmission window.
+The SSED allows the transmission to finish if radio activity is detected during the transmission window.
+Conversely, if there is no radio activity during the specified duration of transmission window, this indicates that the parent has no messages for the SSED and the SSED returns to sleep.
+The SSED synchronization results in lower power consumption compared to an SED, primarily because the SSED doesn't need to poll for messages, keeping transmission windows short.
+
+For more information about SSED activity, see the :ref:`thread_sed_ssed` page.
+
+.. _thread_types_configuring:
+
+Type configuration
+******************
+
+See :ref:`thread_ug_device_type` for information about how to configure Thread device type.
--- a/doc/nrf/protocols/thread/sed_ssed.rst
+++ b/doc/nrf/protocols/thread/sed_ssed.rst
@@ -1,0 +1,64 @@
+.. _thread_sed_ssed:
+
+Sleepy End Device types in Thread
+#################################
+
+Sleepy End Devices (SEDs) and Synchronized Sleepy End Devices (SSEDs) are both Minimal Thread Devices (MTDs).
+Unlike Full Thread Devices (FTDs), MTDs do not maintain a routing table and are typically low-power devices that are not always on.
+
+For more information, see the :ref:`thread_types_mtd` section in the :ref:`thread_device_types` documentation.
+
+Sleepy End Device
+*****************
+
+SEDs are MTDs that sleep most of the time in order to minimize the power consumption.
+They communicate with the Thread network by occasionally polling the parent Router for any pending data.
+
+The :kconfig:option:`CONFIG_OPENTHREAD_POLL_PERIOD` Kconfig handles the SED configuration by configuring the polling period.
+A higher polling frequency results in lower latency (better responsiveness), but also higher power consumption.
+
+The polling period can also be configured in runtime.
+See the ``pollperiod`` command in the OpenThread `CLI reference <OpenThread CLI Reference - pollperiod command>`_.
+
+Synchronized Sleepy End Device
+******************************
+
+SSEDs are MTDs that are further optimized for power consumption.
+A Thread SSED is synchronized with its parent Router and uses the radio only at scheduled intervals.
+It does this by using the :ref:`thread_ug_supported_features_csl` feature (introduced as one of the `Thread 1.2 Base Features`_).
+During those intervals, the device waits for the Router to send it any data related to the desired device activity.
+This reduces the network traffic (since there is no polling) and the power consumption (since the radio is off most of the time).
+
+An SSED does require sending packets occasionally to keep synchronization with the Router.
+However, unlike a regular SED, an SSED does not actively communicate with the Router by polling and it goes into the idle mode between the scheduled activity periods.
+If there is no application-related traffic for an extended period of time, the SSED sends a data poll request packet to synchronize with the parent.
+
+The |NCS| provides the following Kconfig options that let you enable CSL and specify the CSL parameters:
+
+* :kconfig:option:`CONFIG_OPENTHREAD_CSL_RECEIVER` - Enables SSED child mode.
+* :kconfig:option:`CONFIG_OPENTHREAD_CSL_AUTO_SYNC` - Enables the CSL autosynchronization feature.
+* :kconfig:option:`CONFIG_OPENTHREAD_CSL_TIMEOUT` - Sets the default CSL timeout in seconds.
+* :kconfig:option:`CONFIG_OPENTHREAD_CSL_CHANNEL` - Sets the default CSL channel.
+  This option corresponds to the ``csl channel`` CLI parameter.
+
+The following Kconfig options affect the size of the receive window, and thus also affect the device's power consumption:
+
+* :kconfig:option:`CONFIG_OPENTHREAD_CSL_RECEIVE_TIME_AHEAD` - Sets the CSL receiver wake up margin in microseconds.
+* :kconfig:option:`CONFIG_OPENTHREAD_MIN_RECEIVE_ON_AHEAD` - Sets the minimum receiving time before start of MAC header.
+* :kconfig:option:`CONFIG_OPENTHREAD_MIN_RECEIVE_ON_AFTER` - Sets the minimum receiving time after start of MAC header.
+* :kconfig:option:`CONFIG_OPENTHREAD_PLATFORM_CSL_UNCERT` - Sets the fixed uncertainty of the device for scheduling CSL Transmissions in units of 10 microseconds.
+
+Additionally, you must configure the ``period`` CLI parameter to enable CSL.
+For more information on using the CLI to configure parameters for CSL, see the ``csl`` command in the OpenThread `CLI reference <OpenThread CLI Reference - csl command_>`_.
+
+Comparison of SED and SSED
+**************************
+
+Compared to an SED, an SSED has no drawbacks for transmission and provides reduced power consumption and network traffic.
+This means you should configure your SED devices as SSEDs whenever possible.
+You can see the difference in power consumption on the :ref:`Thread power consumption <thread_power_consumption>` page.
+
+.. figure:: overview/images/thread_sed_ssed_comparison.svg
+   :alt: Comparison of Thread SED and Thread SSED radio activity
+
+   Comparison of Thread SED and Thread SSED radio activity
--- a/doc/nrf/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases/release-notes-changelog.rst
@@ -427,4 +427,7 @@ cJSON
 Documentation
 =============
 
-* Added a page on :ref:`ug_wireless_coexistence` in :ref:`protocols`.
+Added:
+
+* A page on :ref:`ug_wireless_coexistence` in :ref:`protocols`.
+* Pages on :ref:`thread_device_types` and :ref:`thread_sed_ssed` to the :ref:`ug_thread` documentation.
--- a/samples/matter/window_covering/README.rst
+++ b/samples/matter/window_covering/README.rst
@@ -45,9 +45,19 @@ SSED device type
 
 The SSED device type was created for the window covering devices to optimize the power usage of the device and communication pattern with the parent.
 
-.. include:: /protocols/thread/overview/commissioning.rst
-   :start-after: thread_ssed_description_start
-   :end-before: thread_ssed_description_end
+A Thread Synchronized Sleepy End Device (SSED) is synchronized with its parent router and uses the radio only at scheduled intervals, by using the :ref:`thread_ug_supported_features_csl` feature introduced as one of the `Thread 1.2 Base Features`_.
+During those intervals, the device waits for the router to send it any data related to the desired device activity.
+The SSED does require sending packets occasionally to keep synchronization with the router.
+However, unlike a regular SED, an SSED does not actively communicate with the router by polling and goes into the idle mode between the scheduled activity periods.
+If there is no application-related traffic for an extended period of time, the SSED sends a data poll request packet to synchronize with the parent.
+Compared to a standard SED, the SSED features further reduce energy consumption of the device and generate less data traffic.
+
+.. figure:: ../../../doc/nrf/protocols/thread/overview/images/thread_sed_ssed_comparison.svg
+   :alt: Comparison of Thread SED and Thread SSED radio activity
+
+   Comparison of Thread SED and Thread SSED radio activity
+
+See the :ref:`thread_sed_ssed` page for more information.
 
 Sample testing
 ==============
--- a/doc/nrf/links.txt
+++ b/doc/nrf/links.txt
@@ -80,6 +80,8 @@
 .. _`nRF52820 CMakeLists.txt`: https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/arm/nrf52833dk_nrf52820/CMakeLists.txt
 
 .. _`OpenThread CLI Reference`: https://github.com/openthread/openthread/blob/main/src/cli/README.md
+.. _`OpenThread CLI Reference - csl command`: https://github.com/openthread/openthread/blob/main/src/cli/README.md#csl
+.. _`OpenThread CLI Reference - pollperiod command`: https://github.com/openthread/openthread/blob/main/src/cli/README.md#pollperiod
 .. _`Commissioner CLI commands`: https://github.com/openthread/openthread/blob/main/src/cli/README_COMMISSIONER.md
 .. _`Joiner CLI commands`: https://github.com/openthread/openthread/blob/main/src/cli/README_JOINER.md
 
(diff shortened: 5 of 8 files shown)


doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,84 @@
+.. _thread_device_types:
+
+Thread device types
+###################
+
+.. contents::
+   :local:
+   :depth: 2
+
+Thread devices can be either Full Thread Devices (FTD) or Minimal End Devices (MTD).
```
pjasinski990, line 10: should me Minimal Thread Devices (MTD), mb

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Routers can be regarded as the backbone of a Thread network.
+They maintain the routing table and forward messages to other devices.
+A router is required for a Minimal End Device to join the network.
+Routers, unlike End Devices, can also be used to extend the network range.
+
+Both Routers and REEDs have routing capabilities, while FEDs do not.
+REEDs function similar to FEDs, but they can become Routers if needed.
+If a REED is the only device in range of an End Device trying to join the network it will promote itself to a Router (see `Joining an existing Thread network`_ in the OpenThread documentation for more information).
+Conversely, when a Router has no children it can downgrade itself to a REED.
```
pjasinski990, line 44: I am now thinking we could add a sentence informing that the limit of routers is 32 for a single network (enforced by 5-bit router id encoding in RLOC16)

doc/nrf/protocols/thread/device_types.rst
wiba-nordic: Good catch.

doc/nrf/protocols/thread/device_types.rst
wiba-nordic: Added mention of maximum number, but left the bit about RLOC16 out.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,85 @@
...
+
+SEDs try to limit their power consumption by sleeping most of the time, waking up periodically to poll for messages from their parent.
+After waking up, they send a data request to their parent.
+If the parent has any pending messages, it will send them to the SED.
+Otherwise, the parent will send a response indicating no pending messages.
+
+SSEDs operate similarly to Sleepy End Devices, but they are synchronized with their parent.
+They wake up at the same time as their parent, eliminating the need for polling for messages.
+If the parent has messages for the SSED, it sends them during the designated transmission window and the SSED allows the transmission to finish if radio activity is detected during the transmission window.
+Conversely, if there is no radio activity during the specified transmission window duration (indicating the parent has no messages for the SSED), the SSED returns to sleep.
```
greg-fer, line 75:
```suggestion
Conversely, if there is no radio activity during the specified duration of the transmission window, this indicates that the parent has no messages for the SSED and the SSED returns to sleep.
```

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,85 @@
...
+This synchronization results in lower power consumption, primarily because the SSED doesn't need to poll for messages, keeping transmission windows short.
```
greg-fer, line 76:
```suggestion
The SSED synchronization results in lower power consumption compared with the SED, primarily because the SSED doesn't need to poll for messages, keeping transmission windows short.
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
+.. _thread_sed_ssed:
+
+Sleepy End Device types in Thread
+#################################
+
+Sleepy End Devices (SEDs) and Synchronized Sleepy End Devices (SSEDs) are both :ref:`Minimal Thread Devices (MTDs)<thread_types_mtd>`.
+For more general information about Thread device types, see the documentation for :ref:`thread_device_types`.
```
greg-fer, line 7:
```suggestion
Sleepy End Devices (SEDs) and Synchronized Sleepy End Devices (SSEDs) are both Minimal Thread Devices (MTDs).
Unlike Full Thread Devices (FTDs), MTDs do not maintain a routing table and are typically low-power devices that are not always on.

For more general information about Thread device types, see the documentation for :ref:`thread_device_types`.
```
I would include some extra info and remove one of the links, so that we don't push the reader to go to other page if he needs a refresher on the differences between MTDs and FTDs.

doc/nrf/protocols/thread/device_types.rst
greg-fer, line 76: Also here I would suggest a table to present the differences.

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+
+Sleepy End Device
+*****************
+
+SEDs are MTDs that sleep most of the time in order to minimize the power consumption.
+They communicate with the Thread network by occasionally polling the parent router for any pending data.
```
greg-fer, line 13:
```suggestion
They communicate with the Thread network by occasionally polling the parent Router for any pending data.
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+
+The following Kconfig option is for SED configuration:
+
+* :kconfig:option:`CONFIG_OPENTHREAD_POLL_PERIOD` - Configures the polling period.
+  Higher polling frequency results in lower latency (better responsiveness), but also higher power consumption.
+
+The polling period can also be configured in runtime. See the ``pollperiod`` command in the OpenThread `CLI reference <OpenThread CLI Reference - pollperiod command>`_.
```
greg-fer, line 20:
```suggestion
The :kconfig:option:`CONFIG_OPENTHREAD_POLL_PERIOD` Kconfig option handles the SED configuration by configuring the polling period.
A higher polling frequency results in a lower latency (that is, better responsiveness), but also a higher power consumption.

The polling period can also be configured at runtime.
See the ``pollperiod`` command in the OpenThread `CLI reference <OpenThread CLI Reference - pollperiod command>`_.
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+
+Synchronized Sleepy End Device
+******************************
+
+SSEDs are MTDs that are further optimized for power consumption.
+A Thread SSED is synchronized with its parent router and uses the radio only at scheduled intervals.
```
greg-fer, line 26:
```suggestion
A Thread SSED is synchronized with its parent Router and uses the radio only at scheduled intervals.
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+It does this by using the :ref:`thread_ug_supported_features_csl` feature introduced as one of the `Thread 1.2 Base Features`_.
```
greg-fer, line 27:
```suggestion
It does this by using the :ref:`thread_ug_supported_features_csl` feature (introduced as one of the `Thread 1.2 Base Features`_).
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+During those intervals, the device waits for the router to send it any data related to the desired device activity.
```
greg-fer, line 28:
```suggestion
During those intervals, the device waits for the Router to send it any data related to the desired device activity.
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+This reduces the network traffic since there is no polling as well as the power consumption since the radio is off most of the time.
```
greg-fer, line 29:
```suggestion
This reduces both the network traffic (since there is no polling) and the power consumption (since the radio is off most of the time).
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+
+An SSED does require sending packets occasionally to keep synchronization with the router.
+However, unlike a regular SED, an SSED does not actively communicate with the router by polling and goes into the idle mode between the scheduled activity periods.
+If there is no application-related traffic for an extended period of time, the SSED sends a data poll request packet to synchronize with the parent.
```
greg-fer, line 33:
```suggestion
An SSED does require sending packets occasionally to keep synchronization with the Router.
However, unlike a regular SED, an SSED does not actively communicate with the Router by polling and it goes into the idle mode between the scheduled activity periods.
If there is no application-related traffic for an extended period of time, the SSED sends a data poll request packet to synchronize with the parent.
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+
+The |NCS| provides Kconfig options that let you enable CSL and specify the CSL parameters:
```
greg-fer, line 35:
```suggestion
The |NCS| provides the following Kconfig options that let you enable CSL and specify the CSL parameters:
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+
+* :kconfig:option:`CONFIG_OPENTHREAD_CSL_RECEIVER` - Enables SSED child mode.
+* :kconfig:option:`CONFIG_OPENTHREAD_CSL_AUTO_SYNC` - Enables CSL autosync.
```
greg-fer, line 38:
```suggestion
* :kconfig:option:`CONFIG_OPENTHREAD_CSL_AUTO_SYNC` - Enables the CSL autosynchronization feature.
```

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,63 @@
...
+
+The following Kconfig options affect the size of the receive window, and thus also affect the device's power consumption:
+
+* :kconfig:option:`CONFIG_OPENTHREAD_CSL_RECEIVE_TIME_AHEAD` - Sets the CSL receiver wake up margin in microseconds.
+* :kconfig:option:`CONFIG_OPENTHREAD_MIN_RECEIVE_ON_AHEAD` - Sets the minimum receiving time before start of MAC header.
+* :kconfig:option:`CONFIG_OPENTHREAD_MIN_RECEIVE_ON_AFTER` - Sets the minimum receiving time after start of MAC header.
+* :kconfig:option:`CONFIG_OPENTHREAD_PLATFORM_CSL_UNCERT` - Sets the fixed uncertainty of the device for scheduling CSL Transmissions in units of 10 microseconds.
+
+Additionally, you must configure the ``period`` CLI parameter to enable CSL.
+For more information on using the CLI to configure parameters for CSl, see the ``csl`` command in the OpenThread `CLI reference <OpenThread CLI Reference - csl command_>`_.
```
greg-fer, line 51:
```suggestion
For more information on using the CLI to configure parameters for CSL, see the ``csl`` command in the OpenThread `CLI reference <OpenThread CLI Reference - csl command_>`_.
```

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -365,4 +365,7 @@ cJSON
 Documentation
 =============
 
-* Added a page on :ref:`ug_wireless_coexistence` in :ref:`protocols`.
+Added:
+
+* A page on :ref:`ug_wireless_coexistence` in :ref:`protocols`.
+* Pages on :ref:`thread_device_types` and :ref:`thread_sed_ssed` to the :ref:`ug_thread` documentation.
```
greg-fer, line 371: I think this could be featured in Protocols/Thread section, actually.

wiba-nordic: I was kind of 50/50 on it, seeing as it's purely a documentation addition.

doc/nrf/protocols/thread/sed_ssed.rst
wiba-nordic: Moved the links around a little.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,85 @@
...
+A router is required for a Minimal End Device to join the network.
+Routers, unlike End Devices, can also be used to extend the network range.
+The maximum number of routers in a single Thread network is 32.
+
+Both Routers and REEDs have routing capabilities, while FEDs do not.
+REEDs function similar to FEDs, but they can become Routers if needed.
+If a REED is the only device in range of an End Device trying to join the network it will promote itself to a Router (see `Joining an existing Thread network`_ in the OpenThread documentation for more information).
+Conversely, when a Router has no children it can downgrade itself to a REED.
+
+FEDs don't require a parent to function, but they can't become parents (Routers) themselves.
```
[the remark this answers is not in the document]
wiba-nordic: I rearranged the information. Will try to mull a table that makes sense.

wiba-nordic:
Added tables, though I'm not happy with the entry for No new messages for MEDs. If anyone has better ideas for that one, I'm all ears.

greg-fer:
> Added tables, though I'm not happy with the entry for No new messages for MEDs. If anyone has better ideas for that one, I'm all ears.

I like the table formatting.

greg-fer (commented):
@wiba-nordic , are you planning any more changes wrt my other comments?

greg-fer (approved):
@LuDuda , can you check one more time and approve?

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -0,0 +1,64 @@
...
+* :kconfig:option:`CONFIG_OPENTHREAD_PLATFORM_CSL_UNCERT` - Sets the fixed uncertainty of the device for scheduling CSL Transmissions in units of 10 microseconds.
+
+Additionally, you must configure the ``period`` CLI parameter to enable CSL.
+For more information on using the CLI to configure parameters for CSL, see the ``csl`` command in the OpenThread `CLI reference <OpenThread CLI Reference - csl command_>`_.
+
+Comparison of SED and SSED
+**************************
+
+Compared to an SED, an SSED has no drawbacks for transmission and provides reduced power consumption and network traffic.
+This means you should configure your SED devices as SSEDs whenever possible.
```
LuDuda, line 58: I would definitely not advise to use SSED always but only for cases where we need a small latency (small csl interval) and want to have similar or smaller power consumption than SED has (e.g. window coverings). For sensors that report a new measurement each e.g. 30-60 seconds, is it justified to use CSL?

The common misconception is that SSED role is just better version of SED role. This is not true, it was designed for specific use cases and I would still expect vast majority of devices to use SED role rather than SSED role. 

@edmont could you comment here? I would like to provide a clear recommendation when usage of SSED is better. Unfortunately, we need to account the possibility of being a child of a router that has very large uncertainty - which would make solution much worse than SED.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+They don't keep a routing table, but otherwise operate like :ref:`FEDs <thread_types_ftd>`.
+
+SEDs try to limit their power consumption by sleeping most of the time, waking up periodically to poll for messages from their parent.
+After waking up, they send a data request to their parent.
+If the parent has any pending messages, it will send them to the SED.
+Otherwise, the parent will send a response indicating no pending messages.
+
+SSEDs operate similarly to Sleepy End Devices, but they are synchronized with their parent.
+They wake up at the same time as their parent, eliminating the need for polling for messages.
+If the parent has messages for the SSED, it sends them during the designated transmission window.
```
LuDuda, line 121: In practice, often there might be more frames that are buffered on Router side. In such case the first frame is received in designated transmission window, but the rest are "pulled" using regular Data Polls (exactly like SED).

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+     - When working as a Router
+   * - Can extend the network
+     - Yes
+     - No
+     - When working as a Router
+
+Routers can be regarded as the backbone of a Thread network.
+They maintain the routing table and forward messages to other devices.
+Routers, unlike End Devices, can also be used to extend the network range and a Router is required for a Minimal End Device to join the network.
+The maximum number of Routers in a single Thread network is 32.
```
LuDuda, line 62: "active Router"

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+
+FEDs maintain a routing table, but they can't route messages.
+This means they don't require a parent (Router) to function, but they cannot become parents themselves.
```
LuDuda, line 65: This is not correct. FEDs operate as End Devices and require parent (Router) to function.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+Thread device types
+###################
+
+.. contents::
+   :local:
+   :depth: 2
+
+Thread devices can be either Full Thread Devices (FTD) or Minimal Thread Devices (MTD).
+The devices are divided based on their ability to exist in a network without a parent, determined by whether the device maintains a routing table (doesn't need a parent) or not (needs a parent).
+FTDs are the devices that maintain a routing table, while MTDs do not, forwarding all their messages to a parent instead.
```
LuDuda, line 12: FTDs may also be in a "Child" state in the Thread network - forwarding all their messages to parent. I guess this sentence might be differently understood.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+Based on this classification, FTDs can be either Routers or End Devices, while MTDs can only be End Devices.
+
+For additional information about Thread device types, see the documentation for `Device Types on OpenThread portal`_.
+
+.. _thread_types_ftd:
+
+Full Thread Devices
+*******************
+
+In order to maintain the routing table, FTDs keep their radio on at all times.
```
LuDuda, line 28: Main reason why FTDs (but in fact also MTDs (like MED or SED during attachment, scanning)) - is that they need to be ready to receive messages asynchronously (from other devices). Maintaining routing table is a separate topic.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+An important difference between FTDs and MTDs is that FTDs subscribe to special "all-routers" multicast addresses (``ff02::2`` and ``ff03::2``, see `IPv6 multicast addressing`_).
+Because of these reasons, they typically consume more power than MTDs.
```
LuDuda, line 30: The subscription to the address won't directly result in higher power consumption. I would not mixed these two things.

It is true that the due to keeping the radio on all the time, device consumes more power, and needs to be mains powered.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+     - Full End Device
+     - Router Eligible End Device
+   * - Maintains a routing table
+     - Yes
+     - Yes
+     - Yes
+   * - Routes messages
+     - Yes
+     - No
+     - When working as a Router
```
LuDuda, line 53: This is a bit misleading. When REED operates as Router then it is called Router. Basically you cannot build the Thread device to operate as "Router" or "REED" only.. device will be in one of these states depending on the network condition.

So here i would put: No

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+   * - Can extend the network
+     - Yes
+     - No
+     - When working as a Router
```
LuDuda, line 57: This one is also controversial like one above.

doc/nrf/protocols/thread/device_types.rst
LuDuda: In general, maybe it would be worth to change this section a bit. 

From roles perspective FTD can be either:
 - REED / Router / Leader  (any of these depending on the network topology)
 - FED (which effectively is always a child)
 
 By reading this section, i have an impression that it is possible to configure device to be a REED or Router (Active Router).

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+Routers can be regarded as the backbone of a Thread network.
+They maintain the routing table and forward messages to other devices.
+Routers, unlike End Devices, can also be used to extend the network range and a Router is required for a Minimal End Device to join the network.
+The maximum number of Routers in a single Thread network is 32.
+
+FEDs maintain a routing table, but they can't route messages.
+This means they don't require a parent (Router) to function, but they cannot become parents themselves.
+
+REEDs maintain a routing table and, unlike FEDs, they can become Routers if needed.
+If a REED is the only device in range of an End Device trying to join the network, it will promote itself to a Router (see `Joining an existing Thread network`_ in the OpenThread documentation for more information).
```
LuDuda, line 68: nit: maybe "it will proactively request promotion to a Router role"   (REED may only request Leader for promotion)

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+Conversely, when a Router has no children it can downgrade itself to a REED.
```
LuDuda, line 69: It's true, but also simplification. The Thread specification provides a few other possibilities when Routers should downgrade to REED role. Basically it's all about the network topology, so maybe:

"Depending on the network topology conditions, Router may downgrade itself to a REED role (e.g. when does not have any children and sufficient number of routers are already in the network)."

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+
+   * -
+     - Minimal End Device
+     - Sleepy End Device
+     - Synchronized Sleepy End Device
+   * - Maintains a routing table
+     - No
+     - No
+     - No
+   * - Radio on
```
LuDuda, line 98: Maybe "Radio characteristic"
"Always enabled"
"Periodically enabled"

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+     - Always
+     - Periodically
+     - Periodically
+   * - New messages
+     - Parent immediately forwards the messages
+     - The SED polls for new messages when it wakes up
+     - Parent forwards the messages during designated transmission window
```
LuDuda, line 105: ... or when SSED explicitly polls data.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+   * - No new messages
+     - No data transmission
+     - Parent indicates no pending messages
+     - No data transmission
```
LuDuda, line 109: I think this is true. However :) in order to keep synchronization, we need to send messages from time to time (to respect CSL timeout).

Update: I see you have mentioned that in some more detailed documentation.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+
+MEDs are the most basic MTDs, and their radio is always on.
+They don't keep a routing table, but otherwise operate like :ref:`FEDs <thread_types_ftd>`.
```
LuDuda, line 112: This is not fully acurate. The difference between MTD and FTD is not about the routing table only. It's about certain functionalities/capabilites which FTD has, but MTD has not, like ability to resolve IPv6 addresses, handling address registration, maintaining routing table, maintaining links with neighbors and so on.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+
+SEDs try to limit their power consumption by sleeping most of the time, waking up periodically to poll for messages from their parent.
+After waking up, they send a data request to their parent.
+If the parent has any pending messages, it will send them to the SED.
+Otherwise, the parent will send a response indicating no pending messages.
+
```
LuDuda, line 118: I think it would be worth to mention how SED and SSEDs transmit messages (not data polls) to parent e.g. that it may happen any time, since Routers have their radio enabled all the time.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -0,0 +1,133 @@
...
+SSEDs operate similarly to Sleepy End Devices, but they are synchronized with their parent.
+They wake up at the same time as their parent, eliminating the need for polling for messages.
```
LuDuda, line 120: What does it mean that the parent "wakes up"? Parent never goes to sleep. It may may rather send the non-polled frame only in designated transmission window.

doc/nrf/protocols/thread/sed_ssed.rst
LuDuda: In general e.g. from HomeKit or Matter perspective, it would be great that potential user will read this section and will be able to choose the SED or SSED role depending on its device type / communication patters.

samples/matter/window_covering/README.rst
```diff
@@ -45,9 +45,19 @@ SSED device type
...
 
-.. include:: /protocols/thread/overview/commissioning.rst
-   :start-after: thread_ssed_description_start
-   :end-before: thread_ssed_description_end
+A Thread Synchronized Sleepy End Device (SSED) is synchronized with its parent router and uses the radio only at scheduled intervals, by using the :ref:`thread_ug_supported_features_csl` feature introduced as one of the `Thread 1.2 Base Features`_.
+During those intervals, the device waits for the router to send it any data related to the desired device activity.
+The SSED does require sending packets occasionally to keep synchronization with the router.
+However, unlike a regular SED, an SSED does not actively communicate with the router by polling and goes into the idle mode between the scheduled activity periods.
+If there is no application-related traffic for an extended period of time, the SSED sends a data poll request packet to synchronize with the parent.
+Compared to a standard SED, the SSED features further reduce energy consumption of the device and generate less data traffic.
```
LuDuda, line 53: "MAY further reduce energy" - it's not granted ;)

LuDuda, line 53: "MAY further reduce energy" - it's not granted ;)

LuDuda (commented):
First of all, thank you for your work on describing the Thread roles, this is indeed very important for our customers! I would like to "Request Changes" as in some places information might be more accurate, the rest are nits.

doc/nrf/protocols/thread/sed_ssed.rst
edmont: I totally agree, SSEDs are not intended to fully replace SEDs. As a rule of thumb it would be beneficial (in terms of energy and spectrum usage) to used SSEDs when the application requires to poll more often than 20 seconds. But this depends on many factors, like routers CSL Accuracy. KRKNWK-13546 should address this topic more in detail.

greg-fer:
> First of all, thank you for your work on describing the Thread roles, this is indeed very important for our customers! I would like to "Request Changes" as in some places information might be more accurate, the rest are nits.

@LuDuda , let me introduce your changes in a separate PR.

doc/nrf/protocols/thread/device_types.rst
greg-fer: I leave it up to @pjasinski990 if we want to mention this child state and where.

greg-fer:
https://github.com/nrfconnect/sdk-nrf/pull/11619

PR #11411, opened 2023-06-02
ci: clang-format: enforce code formatting

gmarull:
Add a new workflow that enforces code to be formatted using the rules specified in the `.clang-format` file. Because `git-clang-format` is used, only the changed lines are checked. This allows for a gradual codebase formatting.

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,22 @@
+name: Check code formatting
+
+on: pull_request
+
+env:
+  CLANG_FORMAT_VERSION: "16.0"
+
+jobs:
+  check-code-formatting:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install clang-format
+        run: pip install clang-format~=${CLANG_FORMAT_VERSION}
+
+      - name: Run clang-format
+        run: git clang-format origin/${{ github.base_ref }} --verbose


gmarull:
@umapraseeda should we add some documentation on how to install/run it locally?

SebastianBoe (changes requested):
It's cheaper to for a single developer to run clang-format once to reformat the entire code base than for dozens of developers to run it hundreds of times to gradually format the code base.

gmarull:
> It's cheaper to for a single developer to run clang-format once to reformat the entire code base than for dozens of developers to run it hundreds of times to gradually format the code base.

I'd be a bit more careful with sdk-nrf, we need to discuss the implications of this treewide change.

lemrey (changes requested):
This is new? `checkpatch` has always been enforced but `clang-format` has never been enforced. I do not believe we should enforce it either; it doesn't always format the code as one would reasonably expect to.

SebastianBoe:
"I'd be a bit more careful with sdk-nrf, we need to discuss the implications of this treewide change."
Introducing this action is a treewide change IMHO.

Now I need to run clang-format if I touch a source line.

gmarull:
> "I'd be a bit more careful with sdk-nrf, we need to discuss the implications of this treewide change." Introducing this action is a treewide change IMHO.
> 
> Now I need to run clang-format if I touch a source line.

Not always, if that line is properly formatted already. Also, you can run `git clang-format` and it’ll automatically format just your changes.

SebastianBoe:
It's not OK IMHO to introduce a test that fails for unrelated changes to the patch I am writing.

e.g. a variable rename from FOO to BAR does not make the formatting worse, but it will fail in CI due to this action.

PR #11417, opened 2023-06-02
samples: modem_trace_flash: Add shell support

jorgenmk:
Add shell support for triggering trace dump. Will be used in test automation.


jhn-nordic (approved):

jorgenmk:
Will use #11095 instead

PR #11418, opened 2023-06-02
applications: nrf5340_audio: Store prev vol on vol adjust

alexsven:
- Fixes bug where previous volume was not stored
- OCT-2588

### DIFF
--- a/applications/nrf5340_audio/src/modules/hw_codec.c
+++ b/applications/nrf5340_audio/src/modules/hw_codec.c
@@ -24,7 +24,9 @@
 LOG_MODULE_REGISTER(hw_codec, CONFIG_MODULE_HW_CODEC_LOG_LEVEL);
 
 #define VOLUME_ADJUST_STEP_DB 3
-#define BASE_10 10
+#define BASE_10		      10
+
+static uint32_t prev_volume_reg_val = OUT_VOLUME_DEFAULT;
 
 static cs47l63_t cs47l63_driver;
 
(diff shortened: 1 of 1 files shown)


Frodevan (approved):
Verified working!

andvib (approved):

applications/nrf5340_audio/src/modules/hw_codec.c
```diff
@@ -73,59 +75,53 @@ int hw_codec_volume_set(uint8_t set_val)
...
-		return ret;
-	}
-
-	volume_reg_val &= CS47L63_OUT1L_VOL_MASK;
-
-	/* The adjustment is in dB, 1 bit equals 0.5 dB,
-	 * so multiply by 2 to get increments of 1 dB
-	 */
-	int32_t new_volume_reg_val = volume_reg_val + (adjustment_db * 2);
+		new_volume_reg_val = prev_volume_reg_val;
```
erikrobstad, line 95: Is it necessary to write to HW codec if adjustment is 0?

erikrobstad: Can return 0 here if not necessary

applications/nrf5340_audio/src/modules/hw_codec.c
```diff
@@ -73,59 +75,53 @@ int hw_codec_volume_set(uint8_t set_val)
...
+	} else {
+		uint32_t volume_reg_val;
 
-	if (new_volume_reg_val < 0) {
-		LOG_WRN("Volume at MIN (-64dB)");
-		new_volume_reg_val = 0;
+		ret = cs47l63_read_reg(&cs47l63_driver, CS47L63_OUT1L_VOLUME_1, &volume_reg_val);
+		if (ret) {
+			LOG_ERR("Failed to get volume from cs47l63");
```
erikrobstad, line 101:
```suggestion
			LOG_ERR("Failed to get volume from CS47l63");
```

erikrobstad (changes requested):

applications/nrf5340_audio/src/modules/hw_codec.c
alexsven: The 0 is used when starting up the chip and will indicate that the volume should be set to the value it had when it was shut down, or the default value when booting up.

alexsven: It's also described in the header file for this function

applications/nrf5340_audio/src/modules/hw_codec.c
alexsven: You forgot the L ;) Changed and pushed

applications/nrf5340_audio/src/modules/hw_codec.c
erikrobstad: Aha, IMO it would be better to write the initial value directly in hw_codec_default_conf_enable(), but this is good enough

erikrobstad (approved):

PR #7532, opened 2022-05-09
workflow: Run native_posix tests and generate code coverage in github actions

balaji-nordic:
- Run native_posix tests (which currently include unit tests) in sdk-nrf in github actions. 
- Generate code coverage and publish the summary as a PR comment.
image
- Coverage report (html) available as artifact of the job
image

- Since the workflow is not yet in the repo, it will not have permissions to post comment or upload the code coverage artifact. To see how this will work, refer https://github.com/balaji-nordic/sdk-nrf/pull/5 
- Future work:
  - Make the action fail if the code coverage is not adequate OR if the PR reduces the coverage on main
  - Replace this with a more feature-rich tool like sonarcloud. See [this conversation](https://teams.microsoft.com/l/message/19:e7aa575b8c2a4438aecc6daa5b380ef8@thread.skype/1662723812041?tenantId=28e5afa2-bf6f-419a-8cf6-b31c6e9e5e8d&groupId=bbf3a6ae-b458-4b7b-9c01-2d2881d8b2fb&parentMessageId=1662723812041&teamName=NCS&channelName=General&createdTime=1662723812041&allowXTenantAccess=false)

Signed-off-by: Balaji Srinivasan <balaji.srinivasan@nordicsemi.no>

PR #11421, opened 2023-06-02
workflow: Add workflow that builds and runs on native_posix

balaji-nordic:
Builds everything that is allowed to be run on native_posix and also
runs them. This therefore runs samples, applications and the unit tests
in the repository.

### DIFF
--- a/.github/workflows/native_posix_run.yml
+++ b/.github/workflows/native_posix_run.yml
@@ -1,0 +1,43 @@
+name: Build and run on native_posix
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: zephyrprojectrtos/ci:v0.26.4
+    strategy:
+      matrix:
+        # Parrallelize the build by running using twister's subset parameter
+        # The size of this should match the environment variable SUBSET_SIZE below.
+        subset: [1, 2, 3, 4, 5]
+    env:
+      # Size of the subset array above.
+      SUBSET_SIZE: 5
+      CMAKE_PREFIX_PATH: /opt/toolchains
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          path: ncs/nrf
+
+      # Additional dependencies for building NCS unit tests
+      - name: Install deps
+        run: |
+          apt-get update
+          apt-get install -y ruby-full
+          pip3 install zcbor==0.5.1
+
+      - name: West init and update
+        run: |
+          west init -l ncs/nrf
+          cd ncs
+          west update --narrow -o=--depth=1
+
+      - name: Build and run on native_posix
+        shell: bash
+        run: |
+          ncs/zephyr/scripts/twister -p native_posix --subset ${{ matrix.subset }}/${{ env.SUBSET_SIZE }} -v -i -T ncs/nrf/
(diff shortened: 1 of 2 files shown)


balaji-nordic:
Re-attempting a smaller and simpler version of https://github.com/nrfconnect/sdk-nrf/pull/7532 .

.github/workflows/native_posix_run.yml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: zephyrprojectrtos/ci:latest
```
jangalda-nsc, line 11: isn't better to use specific version of this image? It may happen there is a new image version which require changes in zephyr repo which are not pulled into sdk-zephyr yet

.github/workflows/native_posix_run.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: West init and update
+        run: |
+          west init -l ncs/nrf
+          cd ncs
+          west update --narrow -o=--depth=1
+
+      - name: Build and run on native_posix
+        shell: bash
+        run: |
+          ncs/zephyr/scripts/twister -p native_posix --subset ${{ matrix.subset }}/${{ env.SUBSET_SIZE }} -v -i -T ncs/nrf/
```
jangalda-nsc, line 43: do we need store some artifacts to enable debugging issues, or having output of this command is enough

jangalda-nsc (approved):
2 questions, they are not PR blockers

jangalda-nsc:
@karhama could you please take a look?

.github/workflows/native_posix_run.yml
balaji-nordic: I really is better to use a specific version of the image. In fact, it is a recommended practice from docker as well. I was tempted to use the [v0.26.4](https://hub.docker.com/layers/zephyrprojectrtos/ci/v0.26.4/images/sha256-4b50d0c970fae7e2cd62a1b0a781d44436a203458c07d7d301631b6560882899?context=explore) tag but I was not entirely sure how zephyr's images are maintained. For example, will they get deleted in the future and so on. Hence I did not go for that approach. But definitely a good point.

balaji-nordic: On second thought, I think it is more risky to depend on `latest` than to depend on a tag. So I will revise this workflow to use a tag instead.

.github/workflows/native_posix_run.yml
balaji-nordic: I have used the `-i` and `-v` options to make it print a verbose output to help debugging. But I am open to suggestions on which artifacts to store. Perhaps we can start storing artifacts as the need arises?

DatGizmo (approved):

karhama:
I tried to check if this is already executed in jenkins side and it seems it is - so basically for me it looks this is a subset of what Jenkins executes for PRs. For a short while it is probably ok to have this kind of setup but for long term we should make our mind and choose which to use.

balaji-nordic:
> I tried to check if this is already executed in jenkins side and it seems it is - so basically for me it looks this is a subset of what Jenkins executes for PRs. For a short while it is probably ok to have this kind of setup but for long term we should make our mind and choose which to use (jenkins or github actions).

Yes. I agree we should keep this in one place. I prefer to run native_posix tests on github actions to give more visibility to the tests. We could remove the jenkins tests if running on github actions is proven to work well (after say a month or so)

balaji-nordic:
Closing this as we could not yet conclude on whether or not this is a good idea.

PR #11424, opened 2023-06-03
lib: nrf_modem_lib: lte_connectivity: Add CEREG listening

glarsennordic:
lte_connectivity now monitors CEREG notifications to more accurately determine whether connectivity is available.

Includes a configurable timeout controlling how long the device can be without a serving cell before serving cell is considered lost.


glarsennordic:
Note for future self: Modify commit to include ticket ID

glarsennordic:
This PR is marked as a draft because it is missing updates to unit tests, but a review now just to confirm that this approach is valid / makes sense would be immensely useful

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -263,17 +385,40 @@ static void pdn_event_handler(uint8_t cid, enum pdn_event event, int reason)
...
+			/* Cancel the registration timeout and mark serving cell as available. */
+			LOG_DBG("Registered to serving cell");
+			k_work_cancel_delayable(&reg_timeout_work);
+			update_has_cell(true);
+			break;
+		default:
+			LOG_DBG("Not registered to serving cell");
+			/* Start a timeout to mark the serving cell as lost. */
+			reg_timeout_schedule();
+			break;
```
trantanen, line 407: I'll discuss this approach more next week when few of my colleagues are back in the office.
Meanwhile, can you describe the problems you face with the PDN activated/deactivated events. I suspect they are coming too late but how much is that?
I assume your application is trying to send data and then connection is lost, which is informed late to the application via `PDN_EVENT_DEACTIVATED`.

trantanen (commented):
Just few questions at this point

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
tokangas: At least `+CEREG: 2` (`LTE_LC_NW_REG_SEARCHING`) should not be considered as loss of connectivity immediately, because modem may find a new (or the same) suitable cell quickly. Guidance from the modem team has been that the application side should rely on the PDN events to detect whether connectivity is available or not. I think that the modem most likely already has a similar timer which is being implemented here, but of course that's not configurable by the application.

I would also like to understand what is the actual problem with the PDN deactivated event?

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -64,6 +75,102 @@ void nrf_modem_fault_handler(struct nrf_modem_fault_info *fault_info)
 	fatal_error_notify_and_disconnect();
 }
 
+
+//TODO: Consider reordering these to be below the connectivity timeout and above the reg timeout
+static void become_active()
+{
+	LOG_DBG("Becoming active");
+	net_if_dormant_off(iface_bound);
```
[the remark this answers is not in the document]
glarsennordic: @rlubos can confirm, but my recollection is that this should be `net_if_dormant`. @rlubos do you recall the reason why we went with that?

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
glarsennordic: > can you describe the problems you face with the PDN activated/deactivated events. I suspect they are coming too late but how much is that?

We have been simulating network loss by placing our devices into RF chambers, and the PDN deactivated event never came, even after many hours of waiting.

My intuition for this is that the PDN status indicates whether there is a logical channel opened for the device, and the CEREG status represents whether there is a PHY layer channel opened for the device

Losing the cell does not immediately imply loss of the logical channel since if cell is regained in a "timely" fashion, the device may resume using the logical channel / PDN carrier it had been using before with no issue

So, in **_theory_**, yes, PDN events are the most "accurate" from the perspective of the carrier -- so long as a PDN is available, the carrier is allowing the device to connect to the internet.

However, from an application perspective, it may be desirable to consider the connection lost much sooner. An application may not want to wait for a non-configurable amount of time potentially lasting hours for the PDN carrier to be officially considered lost. An application may want to consider even a few minutes of loss-of-cell as a disconnection from the network.

These changes allow that behavior to be configured -- applications can rely solely on PDN if they want, or they can set up a time out to their specifications based on what would be considered to them an abnormal amount of time without a serving cell

glarsennordic: > At least +CEREG: 2 (LTE_LC_NW_REG_SEARCHING) should not be considered as loss of connectivity immediately

Regarding CEREG: 2 specifically: Good point, I think searching generally is pretty fast, so it would probably suffice to just have this state ignored completely by the timeout system. I will make that change in the coming days

trantanen: Can you provide us modem logs from the case where PDN deactivation is not seen or takes tens of minutes?
And did you see somewhere that the connection was really lost in an RF chamber. I've used one which still doesn't fully disconnect the device but it keeps connecting and disconnecting so the RF chamber is leaking. It would be good to see modem logs to know what's going on.

glarsennordic: Discussed the aforementioned logs offline, and yes PDN is definitely not coming in, connection is definitely being lost in the RF chamber, and all of this is expected behavior.

@tokangas I've adjusted the event logic to ignore CEREG 2, which I think should do the trick. I'll be testing with this patch over on #11519

lib/nrf_modem_lib/lte_connectivity/Kconfig
```diff
@@ -33,6 +33,17 @@ config LTE_CONNECTIVITY_CONNECT_TIMEOUT_SECONDS
 	  When the timeout is disabled, the library relies solely on the modem's internal
 	  connection handling to upkeep the connection.
 
+config LTE_CONNECTIVITY_REGISTRATION_TIMEOUT_SECONDS
```
trantanen, line 36: Given CEREG 2 is ignored, I'll provoke a discussion whether we need a timeout at all? :)

tokangas: That's a good question. Modem will send `+CEREG: 4` after the search has been finished and nothing was found. The next search starts depending on the periodic search configuration. With the default configuration, I think the delay after the first search is only 10 seconds and starts increasing from that. With this configuration, it would be possible to wait for several searches to be done before determining that LTE connectivity has been lost. But I'm not sure if that's necessary. With `+CEREG: 2` ignored, I guess the timeout could be removed and the implementation simplified.

glarsennordic: Oooh, I'm of two minds on this. For my purposes a long timeout is not needed, and removing the timeout would be nice.

But as @tokangas points out, I could definitely see a customer wanting to maybe wait for a few searches before officially timing out.

Plus, the timeout setting is a convenient way for the customer to disable this behavior altogether, for instance if they want to rely solely on PDN state.

Does any of this change your mind @trantanen? If not then I can remove the timeout and we can always throw it back in if customers ask for it.

This would have the side effect of making CEREG tracking mandatory, though -- no way to pay attention only to PDN -- which admittedly seems way less likely to be that interesting for a customer IMO

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.h
```diff
@@ -36,10 +36,39 @@ enum lte_connectivity_if_down_options {
...
+#define LTE_CONNECTIVITY_REG_TIMEOUT_INSTANT 0
+
 /** @brief Option types used to specify library behavior. */
 enum lte_connectivity_options {
 	/** Sets desired behavior when lte_connectivity_disable() / net_if_down() is called. */
 	LTE_CONNECTIVITY_IF_DOWN = 1,
+
+	/** Registration timeout.
+	 *
+	 *  After becoming unregisterred the serving cell, wait this long (in seconds) before
```
peknis, line 49:
```suggestion
	 *  After becoming unregisterred, the serving cell waits this long (in seconds) before
```

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.h
```diff
@@ -36,10 +36,39 @@ enum lte_connectivity_if_down_options {
...
+	 *  considering the serving cell lost.
+	 *
+	 *  The value type is int. Use LTE_CONNECTIVITY_REG_TIMEOUT_NONE to disable the
+	 *  timeout, and LTE_CONNECTIVITY_REG_TIMEOUT_INSTANT to make the timeout
```
peknis, line 53:
```suggestion
	 *  timeout and LTE_CONNECTIVITY_REG_TIMEOUT_INSTANT to make the timeout
```

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.h
```diff
@@ -36,10 +36,39 @@ enum lte_connectivity_if_down_options {
...
+	 */
+	LTE_CONNECTIVITY_REG_TIMEOUT = 2,
+};
+
+
+struct lte_connectivity_state {
+	/* Tracks the requested behavior when the network interface is brought down. */
+	enum lte_connectivity_if_down_options if_down_setting;
+
+	/* Tracks how long (in seconds) to to tolerate serving cell loss. */
```
peknis, line 64:
```suggestion
	/* Tracks how long (in seconds) to tolerate serving cell loss. */
```

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.h
```diff
@@ -36,10 +36,39 @@ enum lte_connectivity_if_down_options {
...
+	int reg_timeout;
+
+	/* Tracks whether a pdn bearer is currently active. */
```
peknis, line 67:
```suggestion
	/* Tracks whether a PDN bearer is currently active. */
```

peknis (changes requested):

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
trantanen: This comment can be closed as we've confirmed from modem team that we should be monitoring `CEREG: 4` to determine that network is unavailable.

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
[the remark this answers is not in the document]
trantanen: This should still be clarified. Reading `dormant` vs. `carrier` state descriptions, I would have selected `carrier` to be used here but there may well be some reasoning which I don't know as I haven't actively used or familiarized myself with the net_if APIs.

lib/nrf_modem_lib/lte_connectivity/Kconfig
trantanen: > Does any of this change your mind @trantanen? 
> Oooh, I'm of two minds on this.

Same here :) I was just initiating the conversation as I wasn't sure either whether we would need it but...

> If not then I can remove the timeout and we can always throw it back in if customers ask for it.

...unless we come up with a very valid use case, this is the reason why I would leave it out for now. If we add it now, there's a high chance it's one of those dozens of Kconfig's that Zephyr+NCS has that nobody ever use.

> This would have the side effect of making CEREG tracking mandatory, though admittedly disabling it seems like an unlikely desire

This is what I've been thinking, i.e., what's the drawback for the client if net_if is set to dormant state when network is not available? I'm not net_if guru so I'm curious.
And again, if this becomes a problem, we can add a Kconfig.

glarsennordic: In fact, thinking on it, disabling CEREG monitoring basically renders `lte_connectivity` unable to ***ever*** detect network loss if the cell network is totally unreachable, so I'm not sure that **_any_** customers would ever find it valuable. They may as well just not use conn_mgr at that point.

Let's ditch the timeout then and fully adopt CEREG monitoring, you've convinced me

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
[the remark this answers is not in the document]
glarsennordic: I thought @rlubos had mentioned to me at some point that `dormant` is more appropriate generally speaking, but looking back on messages I don't see that interaction


In this specific case I actually wonder whether we could use **_both_** -- PDN could update `dormant` and CEREG could update `carrier`

That would not only match roughly with the respective roles of PDN and CEREG state, but might also significantly simplify this PR. I'll play around with it.

@rlubos do you agree this would be an appropriate use of these two functions?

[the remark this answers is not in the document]
glarsennordic: Well, the state-tracking simplifications I was hoping for don't really pan out; `lte_connectivity` needs a way to know when to start the connection timeout, and passing off the state tracking problem to `net_if` by using both `carrier` and `dormant` doesn't accomplish that.


I'm gonna defer to @rlubos 's final judgement on whether it should be communicated with `dormant` vs with `carrier`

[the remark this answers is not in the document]
glarsennordic: Ah, offloading the problem to `net_if` very nearly works, but creates an unavoidable concurrency problem. Intelligently triggering timeouts requires checking `net_if_is_up` and there is no way to prevent another thread from calling `net_if_down` between that check and the corresponding timeout initiation. 

Explicitly tracking the state has the very strong advantage of avoiding this kind of concurrency trap, so I am tempted to stick with the current approach

I defer to @rlubos on whether we should use `dormant` or `carrier` for that purpose

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.h
glarsennordic: The affected comment has been removed

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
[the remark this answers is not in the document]
trantanen: > In this specific case I actually wonder whether we could use **_both_** -- PDN could update `dormant` and CEREG could update `carrier`

When I read the description for `dormant` and `carrier`, I had this option in mind as well. But let's wait for @rlubos comments.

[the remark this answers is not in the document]
rlubos: `carrier` state represents whether the interface is physically able to transmit packets (so yes, it could be considered L1 related. `dormant` is a bit vague, as conditions for entering/leaving this state can differ between interface types (joining Thread networking or associating with Wi-FI AP). The Linux documentation says:
> Interface is L1 up, but waiting for an external event, f.e. for a protocol to establish.

That said, I think that for LTE connectivity `dormant` **should** be used regardless. But there's no really a problem to use `carrier` as well if it makes sense (I defer to your knowledge about LTE connectivity) and would make the state handling easier. From the application perspective, there's really no difference.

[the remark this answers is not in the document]
trantanen: The way I understand `carrier` and `dormant` is that the former is more like the physical connection and the latter is logical. Physical connection is kind of clear for different technologies and CEREG status indicates this with the modem.

As @rlubos said, dormant is rather vague but seeing many places referring to Wi-Fi association or authentication, I'd say this would match with LTE registration, or PDN connectivity, which are kind of same as you cannot transmit data without either one of them.

I've browsed https://github.com/zephyrproject-rtos/zephyr/pull/51240, which introduced `carrier` (`NET_IF_LOWER_UP`) and `dormant` (`NET_IF_DORMANT`) states in the commit message as follows:

* Introduce a new interface flag: NET_IF_LOWER_UP, along with
  corresponding helper functions. The flag should be set/cleared on an
  interface by a network driver/L2 to signalize physical readiness of an
  interface to transmit data (for example cable plugged in).

* Introduce a new interface flag: NET_IF_DORMANT, along with
  corresponding helper functions. The flag should be set on an
  iterface when the interface is not ready to transmit application data,
  for example still not joined a Wi-Fi network.


Based on all above, I'd say that LTE cell lost would clearly be `carrier_off` and LTE PDN connectivity matches rather well with `dormant` state.
But as @rlubos said, these do not make difference to applications.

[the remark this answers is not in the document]
trantanen: > Even if we are CEREG 4 or PDN DEACT, the modem is still capable of transmitting and receiving packets were there someone in range to receive/transmit them, so carrier counts as being up IMO.

If `carrier` is considered up when we've received CEREG 4, then I'm just thinking about these states in a wrong manner. You cannot really transmit/receive anything when CEREG 4 (cell lost) is indicated. If the net_if semantics are such that you should still say that `carrier` is on in that state, then I do agree we shouldn't set carrier on. As said, I do not know the intended semantics of these states but sounds like `carrier` has nothing to do whether we have the real physical connection in place to transmit data instantly. Yes, you could still do a send() when CEREG 4 has been received but data may only go to the modem buffers. It won't reach the destination.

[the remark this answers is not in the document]
glarsennordic: > If the net_if semantics are such that you should still say that carrier is on in that state, then I do agree we shouldn't set carrier on

I am at this point pretty certain that is the case, so if we agree on this I will close this thread as resolved

lib/nrf_modem_lib/lte_connectivity/Kconfig
glarsennordic: Resolving since it is implemented

plskeggs (approved):
Do you plan to update the changelog?

glarsennordic:
Whoops, thank you for the reminder, I had forgotten! @plskeggs

peknis (approved):

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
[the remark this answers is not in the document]
trantanen: I'll leave this completely to you and maybe @rlubos to decide whether you use carrier or dormant states as I'm not an expert on what the meaning of the states are and seems I've made wrong assumptions already. And as was mentioned earlier, from application point of view it doesn't make a difference.

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -34,16 +34,22 @@ BUILD_ASSERT((CONFIG_NET_CONNECTION_MANAGER_STACK_SIZE >= 1024),
...
 /* Delayable work used to handle LTE connection timeouts. */
 static K_WORK_DELAYABLE_DEFINE(connection_timeout_work, connection_timeout_work_fn);
 
 /* Local reference to the network interface the l2 connection layer is bound to. */
 static struct net_if *iface_bound;
 
-/* Variable used to specify behavior when the network interface is brought down. */
-static uint8_t net_if_down_options = LTE_CONNECTIVITY_IF_DOWN_MODEM_SHUTDOWN;
+/* Container and protection mutex for internal state of the connectivity binding. */
+K_MUTEX_DEFINE(internal_state_lock);
```
jtguggedal, line 49: Nit: can be `static`

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -263,17 +335,45 @@ static void pdn_event_handler(uint8_t cid, enum pdn_event event, int reason)
...
+		/* Searching for a new cell, do not consider this cell loss unless it
+		 * fails (which will generate a new LTE_LC_EVT_NW_REG_STATUS event with
+		 * an unregistered status).
+		 */
+		break;
+	default:
+		LOG_DBG("Not registered to serving cell");
+		/* Mark the serving cell as lost. */
+		update_has_cell(false);
+			break;
```
jtguggedal, line 362: Nit: indentation seems off

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.h
```diff
@@ -42,6 +42,18 @@ enum lte_connectivity_options {
 	LTE_CONNECTIVITY_IF_DOWN = 1,
 };
 
+
+struct lte_connectivity_state {
```
jtguggedal, line 46: Looks like this type is only needed in one file, so could optionally be moved from the header file

jtguggedal (approved):

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -64,6 +70,99 @@ void nrf_modem_fault_handler(struct nrf_modem_fault_info *fault_info)
 	fatal_error_notify_and_disconnect();
 }
 
+static void become_active(void)
```
simensrostad, line 73: Some comments for these functions would be nice, this and `become_dormant` actually mean in this context.

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -263,17 +335,45 @@ static void pdn_event_handler(uint8_t cid, enum pdn_event event, int reason)
...
 	int ret;
 	int timeout = CONFIG_LTE_CONNECTIVITY_CONNECT_TIMEOUT_SECONDS;
-	int persistent = IS_ENABLED(CONFIG_LTE_CONNECTIVITY_CONNECTION_PERSISTENCE);
 
 	net_if_dormant_on(if_conn->iface);
 
-	/* Default auto features off. */
+	/* Apply requested flag overrides */
 	if (!IS_ENABLED(CONFIG_LTE_CONNECTIVITY_AUTO_CONNECT)) {
+
```
simensrostad, line 376:
```suggestion
```

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -34,16 +34,22 @@ BUILD_ASSERT((CONFIG_NET_CONNECTION_MANAGER_STACK_SIZE >= 1024),
 
 /* Forward declarations */
 static void connection_timeout_work_fn(struct k_work *work);
+static void connection_timeout_schedule(void);
+
 int lte_connectivity_disconnect(struct conn_mgr_conn_binding *const if_conn);
 
+
```
simensrostad, line 41:
```suggestion
```

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -34,16 +34,22 @@ BUILD_ASSERT((CONFIG_NET_CONNECTION_MANAGER_STACK_SIZE >= 1024),
 
 /* Forward declarations */
 static void connection_timeout_work_fn(struct k_work *work);
+static void connection_timeout_schedule(void);
+
 int lte_connectivity_disconnect(struct conn_mgr_conn_binding *const if_conn);
```
simensrostad, line 39:
```suggestion
```

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
```diff
@@ -34,16 +34,22 @@ BUILD_ASSERT((CONFIG_NET_CONNECTION_MANAGER_STACK_SIZE >= 1024),
 
 /* Forward declarations */
 static void connection_timeout_work_fn(struct k_work *work);
+static void connection_timeout_schedule(void);
+
```
simensrostad, line 38:
```suggestion
```

tests/lib/nrf_modem_lib/lte_connectivity/src/lte_connectivity_test.c
```diff
@@ -24,6 +24,14 @@
 extern int unity_main(void);
 
 static pdn_event_handler_t pdn_event_handler_callback;
+static lte_lc_evt_handler_t lte_lc_event_handler_callback;
+
+/* Initial condition tracking */
+static bool initial_conditions_observed;
```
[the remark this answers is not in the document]
glarsennordic: This test is specifically to check what the flags are set to at boot time, and since these tests should be written so that they succeed no matter what order they are called in, there is no way to guarantee that the flags haven't been mutated by other tests except to cache them during setup

tests/lib/nrf_modem_lib/lte_connectivity/src/lte_connectivity_test.c
```diff
@@ -365,80 +435,221 @@ void test_options_get_should_return_error_if_wrong_option_name_is_used(void)
...
+
+/* Verify that the iface is inactive before PDN and CEREG are ready, becomes active
+ * afterwards, and becomes inactive again if either is lost.
+ *
+ * Verify also that LTE_LC_NW_REG_SEARCHING is ignored
+ *
+ * k_sleep calls are not strictly necessary since iface state changes are instantaneous,
+ * but we perform them so that the NET_MGMT event queue is not overflowed during the test.
+ */
+void test_iface_pdn_and_cereg_tracking(void)
```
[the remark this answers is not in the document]
glarsennordic: Sure, I can split it up. I thought about doing so for some of these tests, but I thought the added code bloat would not be worth the marginal gains of test specificity -- especially since failed tests indicate what line they failed on.

This is your suite, though, so I'll defer to your preferences

tests/lib/nrf_modem_lib/lte_connectivity/src/lte_connectivity_test.c
```diff
@@ -55,72 +69,70 @@ static void network_interface_down_option_set(enum lte_connectivity_if_down_opti
...
-	}
+	/* Stop ignoring mocked functions*/
+	__cmock_lte_lc_func_mode_set_StopIgnore();
+	__cmock_nrf_modem_lib_shutdown_StopIgnore();
 
-	if (net_if_ipv4_get_global_addr(net_if, NET_ADDR_PREFERRED)) {
-		TEST_ASSERT_EQUAL(0, net_if_config_ipv4_put(net_if));
-	}
+	/* Prepare to track/mock modem_at_scanf calls for the upcoming test */
+	mock_nrf_modem_at_Init();
```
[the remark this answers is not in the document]
glarsennordic: I had thought that since `mock_nrf_modem_at` is a custom mock of ours named to look and function similarly to genuine cmock mocks, that a call to `mock_nrf_modem_at_Init` and `_Verify` are required

The github diff kinda butchers this, but both calls were already included before I made any modifications to these tests

tokangas (approved):

jhirsi (approved):

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
glarsennordic: I'm assuming you meant that I should remove the newline

tests/lib/nrf_modem_lib/lte_connectivity/src/lte_connectivity_test.c
```diff
@@ -322,38 +379,51 @@ void test_options_get_should_return_success(void)
...
+	size_t len = sizeof(option);
 
 	int ret = conn_mgr_if_get_opt(net_if, LTE_CONNECTIVITY_IF_DOWN, (void *)&option, &len);
 
 	TEST_ASSERT_EQUAL(0, ret);
 }
 
-void test_options_get_should_return_correct_option_value(void)
+/* Verify that conn_mgr_if_set_opt successfully changes the IF_DOWN option */
+void test_options_set_should_change_option_value(void)
```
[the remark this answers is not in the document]
glarsennordic: I can split into two testcases if you like, but the first set of `TEST_ASSERT_EQUAL` calls are confirming initial conditions

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
simensrostad: I don't think the forward declaration is needed when it is a public function exposed in `lte_connectivity,h`

simensrostad (approved):

lib/nrf_modem_lib/lte_connectivity/lte_connectivity.c
glarsennordic: Oh duh, you're quite right

I wonder if these members should be static since they are not called outside this file?

At any rate, that's outside the scope of this PR so I will leave them public and remove the forward declaration

PR #11425, opened 2023-06-03
lib: date_time: Fix AUTO_UPDATE for non-LTE builds

glarsennordic:
Currently, if CONFIG_LTE_LINK_CONTROL is not enabled, the date_time auto_update loop will not self-start unless CONFIG_DATE_TIME_AUTO_UPDATE is disabled.

This means that CONFIG_DATE_TIME_AUTO_UPDATE has the opposite effect intended if CONFIG_LTE_LINK_CONTROL is not enabled.

After this commit, automatic date_time updates are always enabled if CONFIG_DATE_TIME_AUTO_UPDATE is enabled, regardless of whether CONFIG_LTE_LINK_CONTROL is.

If CONFIG_LTE_LINK_CONTROL is enabled, this will be done by registering a callback with lte_lc_register_handler.

In all other cases, the auto_update loop will be started by calling date_time_core_schedule_update, and from then on the date_time library will periodically query its configured time source.

### DIFF
--- a/lib/date_time/date_time_core.c
+++ b/lib/date_time/date_time_core.c
@@ -146,11 +146,13 @@ void date_time_lte_ind_handler(const struct lte_lc_evt *const evt)
 
 void date_time_core_init(void)
 {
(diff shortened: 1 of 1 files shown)


glarsennordic:
Note for future self: Modify commit to include ticket ID

PR #11429, opened 2023-06-05
applications: nrf5340_audio: Move MCS discover for headset

alexsven:
- MCS discover moved to lc3_config_cb
- Will now do discovery even if stream is paused
- OCT-2582


alexsven:
Appears that the clang-format file in NCS has changed, so some of the changes in this PR is because of that.

erikrobstad (approved):

Frodevan (approved):
Verified working!

PR #11430, opened 2023-06-05
Move nrf_security from the nrfxlib repo to the nrf repo

SebastianBoe:
Move nrf_security from the nrfxlib repo to the nrf repo to reduce the need for manifest PRs and reduce the amount of west manifest conflicts.

https://nordicsemi.atlassian.net/browse/NCSDK-21941


samples/wifi/shell/README.rst
```diff
@@ -221,7 +221,7 @@ Dependencies
 
 This sample uses the following `sdk-nrfxlib`_ library:
```
sachinthegreen, line 222:
```suggestion
This sample uses the following library:
```

sachinthegreen (changes requested):

doc/CMakeLists.txt
```diff
@@ -83,7 +83,17 @@ function(add_docset name)
...
+  set(DOCSET_ENV
+	DOXYGEN_EXECUTABLE=${DOXYGEN_EXECUTABLE}
+	# The nrf_security nrfxlib Kconfig sources willl source nrf
+    # Kconfigs and therefore need to know where ZEPHYR_NRF_MODULE_DIR
+    # is. It's not clear how other Kconfig env vars are added and why
+    # this needed to be added as a special case here. There could be a
+    # better way to do this, but the author does not know how.
+	ZEPHYR_NRF_MODULE_DIR=${NRF_BASE}
+  )
+
```
gmarull, line 96: this looks like a hack, why is this needed? If nrf is sourcing nrfxlib, why is an nrfxlib component sourcing nrf?

gmarull (changes requested):

doc/CMakeLists.txt
joerchan: NCSDK-22101 is to follow up on the hack

gmarull: So the question is, where in Kconfig is this used so that it affects documentation?

SebastianBoe: related PR: https://github.com/nrfconnect/sdk-nrf/pull/11475

gmarull: and why is this needed? Isn't this already sourced in nrf?

SebastianBoe: It is not sourced in nrf.

It is a hack. We need to preserve the Kconfig order to maintain the same defaults values of MBEDTLS configurations.

We will follow this up later by getting rid of the default order dependencies, but right now this would take too long time and we want to get nrf_security moved as soon as possible so that people can continue to work on nrf_security outside of nrfxlib.

tejlmand (changes requested):
need some discussion / better reasoning for this split.

sachinthegreen (approved):

PR #11433, opened 2023-06-05
applications: Fix configuration loading

nordicjm:
Fixes configuration loading, needed for overlaying application configuration to function correctly.

### DIFF
--- a/applications/nrf_desktop/CMakeLists.txt
+++ b/applications/nrf_desktop/CMakeLists.txt
@@ -6,19 +6,10 @@
 
 cmake_minimum_required(VERSION 3.20.0)
 
-
-if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/configuration/${BOARD}")
-  message(FATAL_ERROR
-          "Board ${BOARD} is not supported.\n"
-          "Please make sure board specific configuration files are added to "
-          "${CMAKE_CURRENT_SOURCE_DIR}/configuration/${BOARD}")
-endif()
-
 ################################################################################
 
 # The application uses the configuration/<board> scheme for configuration files.
-set(APPLICATION_CONFIG_DIR "configuration/\${BOARD}")
-
+set(APPLICATION_CONFIG_DIR "${CMAKE_CURRENT_SOURCE_DIR}/configuration/\${BOARD}")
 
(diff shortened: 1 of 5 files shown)


kkasperczyk-no (approved):

applications/nrf_desktop/CMakeLists.txt
```diff
@@ -6,26 +6,24 @@
...
-
+set(APPLICATION_CONFIG_DIR "${CMAKE_CURRENT_SOURCE_DIR}/configuration/\${BOARD}")
 
 find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 project("nRF52 Desktop"
         VERSION 0.1)
 
 ################################################################################
 
+if (NOT EXISTS "${APPLICATION_CONFIG_DIR}")
```
MarekPieta, line 20: We already pass the `APPLICATION_CONFIG_DIR` to Zephyr build system before the actual check. Can the fatal error be reported there in this case?

MarekPieta: Also from what I understood we expect that Zephyr build system may modify content of the `APPLICATION_CONFIG_DIR` (that's why the check was moved). Consider logging the actually used configuration directory (`APPLICATION_CONFIG_DIR`) in that case.

nordicjm: It already does, e.g.: `Merged configuration '/tmp/bb/nrf/applications/nrf_desktop/configuration/nrf52840dk_nrf52840/prj.conf'`

applications/zigbee_weather_station/CMakeLists.txt
```diff
@@ -6,18 +6,10 @@
 
 cmake_minimum_required(VERSION 3.20.0)
 
-
-if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/configuration/${BOARD}")
```
tomchy, line 10: Could you explain, why this check is considered as an issue?

tomchy (changes requested):

applications/zigbee_weather_station/CMakeLists.txt
nordicjm: Zephyr already checks that APPLICATION_CONFIG_DIR exists and will error out if it does not

tomchy (approved):

PR #11447, opened 2023-06-06
tests: app_event_manager new api

michalek-no:
migration to Ztest new API.


MarekPieta (approved):

gopiotr (approved):
I tested it locally on my hardware and everything works properly.

PR #11448, opened 2023-06-06
tests: edge_impulse new api

michalek-no:
migration to Ztest new API.

### DIFF
--- a/tests/lib/edge_impulse/src/main.cpp
+++ b/tests/lib/edge_impulse/src/main.cpp
@@ -139,7 +139,7 @@ static void result_ready_cb(int err)
 	}
 }
 
-static void test_init(void)
+static void *test_init(void)
 {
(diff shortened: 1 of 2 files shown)


tests/lib/edge_impulse/src/main.cpp
```diff
@@ -430,7 +431,7 @@ static void test_data_isr(void)
 	zassert_ok(err, "Cannot take semaphore");
 }
 
-static void setup_fn(void)
+static void setup_fn(void*)
```
MarekPieta, line 434: btw. should we name the argument here? You may also use `ARG_UNUSED` to clearly mark that it's not going to be used in the function.

MarekPieta (approved):

gopiotr (approved):
I tested it locally on my hardware and everything works properly.

PR #11452, opened 2023-06-07
samples: wifi: sr_coex: update valid build options

muraliThokala:
Update valid build options for both DK and EK.


rado17:
> As twister is now only build, do we need 7001 combinations? Won't 7002 combinations cover all cases.

Having 7001 combinations is good to have since there are a couple of places where we have 7001 specific code.

rado17 (approved):

PR #11457, opened 2023-06-07
lib: modem_key_mgmt: update modem_key_mgmt.c

MechanicV1:
Added support to return 1 when buffer length doesnot match certificate length

### DIFF
--- a/doc/nrf/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases/release-notes-changelog.rst
@@ -313,6 +313,10 @@ Modem libraries
 
   * Updated the library to allow a ``PDP_type``-only configuration in the :c:func:`pdn_ctx_configure` function.
 
+* :ref:`nrf_modem_lib_readme`:
+
+   * Updated the :c:func:`modem_key_mgmt_cmp` function to return ``1`` if the buffer length does not match the certificate length.
+
 Libraries for networking
 ------------------------
 
--- a/lib/modem_key_mgmt/modem_key_mgmt.c
+++ b/lib/modem_key_mgmt/modem_key_mgmt.c
@@ -188,7 +188,7 @@ int modem_key_mgmt_cmp(nrf_sec_tag_t sec_tag,
 		       const void *buf, size_t len)
 {
 	int err;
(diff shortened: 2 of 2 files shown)


lib/modem_key_mgmt/modem_key_mgmt.c
```diff
@@ -199,16 +199,25 @@ int modem_key_mgmt_cmp(nrf_sec_tag_t sec_tag,
...
+	}
+
+	end = strchr(begin, '\"');
+	if (!end) {
+		return -ENOENT;
+	}
+
+	if (end - begin != len) {
+		return -ENOMEM;
 	}
```
lemrey, line 218: The check is correct, but the error is wrong. This function return 0 on match, 1 on mismatch (similar to `memcmp` semantics), or a negative value on errors. So in this case, where the length of the credential in modem does not match the length of the credential the application is checking against, we should return 1.

eivindj-nordic (approved):
LGTM, I think we need a changelog entry, though.

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -284,7 +284,10 @@ Debug libraries
 Modem libraries
 ---------------
 
-|no_changes_yet_note|
+* Modem key Management library:
+
+   * Updated :c:func:`modem_key_mgmt_cmp` function to return 1 if buffer length does not match the certificate length.
```
divipillai, line 289:
```suggestion
* :ref:`nrf_modem_lib_readme`:
  
  * Updated the :c:func:`modem_key_mgmt_cmp` function to return ``1`` if the buffer length does not match the certificate length.
```

lemrey (approved):

divipillai (approved):

cvinayak:
[12:58] Rathakrishnan, Nirmal
    looks like github is not picking the latest run again, https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/job/PR-11457/
​[12:58] Rathakrishnan, Nirmal
    its good to merge Chettimada, Vinayak

PR #11460, opened 2023-06-07
samples: matter: disable Wi-Fi LMP in the case of OTA

markaj-nordic:
... for samples with switchable app variant feature.
This has been already done for all other Wi-Fi based Matter samples.

### DIFF
--- a/samples/matter/common/src/ota_multi_image_processor_impl.cpp
+++ b/samples/matter/common/src/ota_multi_image_processor_impl.cpp
@@ -19,6 +19,10 @@ CHIP_ERROR OTAMultiImageProcessorImpl::PrepareDownload()
 
 	TriggerFlashAction(ExternalFlashManager::Action::WAKE_UP);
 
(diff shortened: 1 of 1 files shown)


Damian-Nordic (approved):

samples/matter/common/src/ota_multi_image_processor_impl.cpp
```diff
@@ -19,6 +19,11 @@ CHIP_ERROR OTAMultiImageProcessorImpl::PrepareDownload()
 
 	TriggerFlashAction(ExternalFlashManager::Action::WAKE_UP);
 
+#ifdef CONFIG_WIFI_NRF700X
+	CHIP_ERROR err = WiFiManager::Instance().SetLowPowerMode(false);
```
Damian-Nordic, line 23: Actually, why not use `ReturnErrorOnFailure`?

LuDuda (approved):

PR #11465, opened 2023-06-08
applications: nrf5340_audio: Remove NULL char from broadcast name

andvib:
Removes the NULL character from the end of the broadcast name string before including it in the advertising data. As the string is defined from a macro it will always be null terminated so no checks are needed.

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/le_audio_bis_gateway.c
+++ b/applications/nrf5340_audio/src/bluetooth/le_audio_bis_gateway.c
@@ -234,8 +234,8 @@ static int adv_create(void)
 		broadcast_id = CONFIG_BT_AUDIO_BROADCAST_ID_FIXED;
 	}
 
-	ext_ad[0] = (struct bt_data)BT_DATA_BYTES(BT_DATA_BROADCAST_NAME,
-						  CONFIG_BT_AUDIO_BROADCAST_NAME);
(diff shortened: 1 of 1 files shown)


applications/nrf5340_audio/src/bluetooth/le_audio_bis_gateway.c
```diff
@@ -234,8 +234,8 @@ static int adv_create(void)
 		broadcast_id = CONFIG_BT_AUDIO_BROADCAST_ID_FIXED;
 	}
 
-	ext_ad[0] = (struct bt_data)BT_DATA_BYTES(BT_DATA_BROADCAST_NAME,
-						  CONFIG_BT_AUDIO_BROADCAST_NAME);
+	ext_ad[0] = (struct bt_data)BT_DATA(BT_DATA_BROADCAST_NAME, CONFIG_BT_AUDIO_BROADCAST_NAME,
```
koffes, line 237: This change is fine, but may this apply in other places as well?

koffes (approved):

gWacey (approved):

alexsven (approved):

applications/nrf5340_audio/src/bluetooth/le_audio_bis_gateway.c
andvib: Broadcast name isn't used anywhere else, AFAICT no string are added in a similar matter - but there might very well be strings I'm not aware of

PR #11475, opened 2023-06-09
compliance: Add ZEPHYR_NRF_MODULE_DIR env var for running compliance

SebastianBoe:
Add ZEPHYR_NRF_MODULE_DIR env var for running compliance.

ZEPHYR_NRF_MODULE_DIR will be used by nrfxlib/Kconfig.nrfxlib, but in
general, any Kconfig source file could refer to any module dir.

kconfig.cmake exports each ZEPHYR_*_MODULE_DIR to Kconfig, but
checkcompliance does not so we add this env var to align with how the
build system invokes Kconfig.

I don't know how to add all MODULES in a clean way so we just add the
one we (will soon) need.

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -71,6 +71,7 @@ jobs:
       if: contains(github.event.pull_request.user.login, 'dependabot[bot]') != true
       run: |
         export ZEPHYR_BASE="$(dirname "$(pwd)")/zephyr"
+        export ZEPHYR_NRF_MODULE_DIR="$(dirname "$(pwd)")/nrf"
         # debug
         ls -la
         git log --pretty=oneline | head -n 10


gmarull:
> ZEPHYR_NRF_MODULE_DIR is used by nrfxlib/Kconfig.nrfxlib, but in general, any Kconfig source file could refer to any module dir.

it's not:

https://github.com/nrfconnect/sdk-nrfxlib/blob/aa953e5b5777bf9f406b546d152e581f9b4e3a63/Kconfig.nrfxlib#L1-L23

so where then?

SebastianBoe:
I apologize, that was an error in the git commit message that got fixed, but I forgot to update the PR description.

SebastianBoe:
Updated PR description to match git commit message

.github/workflows/compliance.yml
```diff
@@ -71,6 +71,7 @@ jobs:
       if: contains(github.event.pull_request.user.login, 'dependabot[bot]') != true
       run: |
         export ZEPHYR_BASE="$(dirname "$(pwd)")/zephyr"
+        export ZEPHYR_NRF_MODULE_DIR="$(dirname "$(pwd)")/nrf"
```
tejlmand, line 74: as `nrf` is not given a module name here: https://github.com/nrfconnect/sdk-nrf/blob/main/zephyr/module.yml
then it obtains it's name from the path, hence they are in sync, but we should consider if a more generic approach is better.
I don't foresee the `nrf` folder to be moved or renamed, but in general having compliance relying on internal repo structure can make refactoring hard and give hard to understand errors.

A generic approach can be obtained by running `package_helper.cmake` from :
```
cmake -Ssamples/basic/blinky -Bbuild_compliance -DMODULES=zephyr_module -P cmake/package_helper.cmake
```
which will produce `build_compliance/zephyr_modules.txt` where each line is:
```
"<module_name>":"<module_dir>":"<cmake_dir>"
```
So the first two parts can be extracted and used for env settings.

tejlmand (approved):
accepted, but a more generic approach would be desired for long term.

.github/workflows/compliance.yml
SebastianBoe: I tried, but eventually gave up.

The module_name has "-", which doesn't work well with env var names I think.

cmake -S$ZEPHYR_BASE/samples/basic/blinky -Bbuild_compliance -DMODULES=zephyr_module -P cmake/package_helper.cmake
awk -F \" '{ print "export ZEPHYR_" toupper($2) "_MODULE_DIR=" $4 }' build_compliance/zephyr_modules.txt > ZEPHYR_x_MODULE_DIR_env_vars.sh
source ZEPHYR_x_MODULE_DIR_env_vars.sh

ZEPHYR_x_MODULE_DIR_env_vars.sh: line 3: export: `ZEPHYR_WFA-QT-CONTROL-APP_MODULE_DIR=/home/sebo/ncs/modules/lib/wfa-qt-control-app': not a valid identifier

SebastianBoe:
Closing until a separate discussion is resolved.

PR #11484, opened 2023-06-12
tests: event_manager_proxy: Migrate to new ZTest API

rakons:
This commit migrates the test to new ZTest API.
Dependency between tests removed by moving initialization to the system init level.

JIRA: NCSDK-21738


gopiotr (approved):
I tested it locally on my hardware setup and it pass. Thank you for your work.

PR #11496, opened 2023-06-12
doc: sml: support for companion chips

gautesl:
Added functionality to include companion chip dependencies as footnotes in the SML tables.

Since the generation of the tables and the data used in the tables are done separately, the doc preview of this PR will not reflect the changes. Instead, these are screenshots of a local build:
image

image

### DIFF
--- a/doc/_scripts/software_maturity/software_maturity_scanner.py
+++ b/doc/_scripts/software_maturity/software_maturity_scanner.py
@@ -1,4 +1,7 @@
 """
+SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+Copyright (c) 2023 Nordic Semiconductor ASA
+
 NCS Software Maturity Level Scanner
 ===================================
 
@@ -113,8 +116,6 @@ is experimental and the others are not, the feature will still be marked as
 experimental. If this is an undesired outcome, the feature should be split
 into two separate features, one with the experimental symbols and the other
 with the supported ones.
-
-Copyright (c) 2022 Nordic Semiconductor ASA
 """
 
 from enum import IntEnum


doc/_scripts/software_maturity/software_maturity_scanner.py
```diff
@@ -28,6 +28,15 @@
   with a value allows all occurrences of the symbol where the value differs
   from the given value.
 
+  In cases where a feature is dependent on a companion chip to work, the feature
+  entry can be expanded to a dictionary containing the fields "rule" and
+  "boards_and_shields" where "rule" is the original KConfig rule and
+  "boards_and_shields" is a list of KConfig options indicating that the
+  required companion chip is present. If at least one sample is found
+  where none of the listed companion chip options are present, the feature
+  is marked as indipendent of the companion chips for the given SoC. Otherwise
```
krish2718, line 37:
```suggestion
  is marked as independent of the companion chips for the given SoC. Otherwise
```

doc/_extensions/software_maturity_table.py
```diff
@@ -223,9 +251,15 @@ def run(self):
 
             for soc in all_socs:
                 entry = nodes.entry()
-                if soc in maturity_list["Experimental"]:
+                footnote_target = None
+
+                if soc in experimental:
+                    if self.minimum_version(version, "0.2.0") and experimental[soc]:
```
krish2718, line 257: As we are auto-detecting the format of `boards_and_shields` as `list` or `dict` this is already backwards compatible, why do we need this version check here? The version of the script is anyway bumped to `0.2.0`, so, won't this always be true?

doc/_scripts/software_maturity/software_maturity_scanner.py
```diff
@@ -551,24 +627,39 @@ def generate_tables(
...
-        for tech, rule in top_table_info.items():
-            rule = parse_rule(rule)
+        for tech, rule_set in top_table_info.items():
+            parsed_rule_set = parse_rule_set(rule_set)
             # Top level technologies are currently only Supported/Not supported
-            if evaluate_rule(kconf.variables, rule):
-                soc_sets[tech].add(soc)
+            match, dependent_on = evaluate_rule_set(kconf.variables, parsed_rule_set)
+            if match:
+                # Dependencies are added together, or cleared if at least on instance has no dependencies
```
krish2718, line 635:
```suggestion
                # Dependencies are added together, or cleared if at least one instance has no dependencies
```

krish2718 (approved):

carlescufi (changes requested):
Please add the Nordic 5-clause license to all .py files

carlescufi (approved):

PR #11521, opened 2023-06-14
doc improvement: matter: update docset info for zoomin

greg-fer:
Edited the name, html heading, and the docset subtitle for Matter. Reason: clearly differentiate the Matter SDK docset from the NCS docs in Zoomin, regardless of placement.

### DIFF
--- a/doc/matter/conf.py
+++ b/doc/matter/conf.py
@@ -20,7 +20,7 @@ ZEPHYR_BASE = utils.get_projdir("zephyr")
 
 # General configuration --------------------------------------------------------
 
-project = "Matter"
+project = "Matter SDK"
 copyright = "2020-2023, Matter Contributors"
 author = "Matter Contributors"
 version = "1.1.0"
@@ -42,7 +42,7 @@ html_static_path = [str(NRF_BASE / "doc" / "_static")]
 html_last_updated_fmt = "%b %d, %Y"
 html_show_sourcelink = True
 html_show_sphinx = False
-html_title = "Matter documentation (nRF Connect SDK)"
+html_title = "Matter SDK documentation (nRF Connect SDK)"
 
 html_theme_options = {
     "docset": "matter",


greg-fer:
@Avalei , @umapraseeda , @b-gent , setting up this PR following our meeting.

@gmarull , I know you're not a fan of these changes that make this docset inconsistent with other docsets, but I'd like to make them stand out more in Zoomin, including its search.

doc/matter/conf.py
```diff
@@ -42,12 +42,12 @@
...
 html_show_sourcelink = True
 html_show_sphinx = False
-html_title = "Matter documentation (nRF Connect SDK)"
+html_title = "Matter SDK documentation (nRF Connect platform)"
 
 html_theme_options = {
     "docset": "matter",
     "docsets": utils.ALL_DOCSETS,
-    "subtitle": "nRF Connect SDK",
+    "subtitle": "nRF Connect platform",
```
greg-fer, line 50: FYI: This refers to the Matter implementation platforms: https://project-chip.github.io/connectedhomeip-doc/guides/index.html

gmarull, line 50: I don't think this is the path to follow, let's discuss this first.

gmarull (changes requested):

DeidreCasey:
I am fine with the changes in this PR. As far as I remember, after discussing this at length with Zoomin, we cannot separate the external doc sets more visually like we do now in NCS. The only viable option would be to upload them all separately as individual doc sets, but that is not possible for us. I think changing this title is a good compromise and should help improve a user's navigation clarity and awareness between doc sets.

@umapraseeda @b-gent any opinions?

doc/matter/conf.py
wiba-nordic: Where specifically is this tag used used? I don't see any mention of "nRF Connect SDK" on the linked page, and I'm not sure the linked page is part of our build of the Matter docs.

greg-fer: No no, the other way around: `nRF Connect platform` refers to the nRF Connect platform on https://project-chip.github.io/connectedhomeip-doc/guides/index.html

`nRF Connect SDK` is this string, which I want changed to `... platform`:
image

greg-fer:
> I am fine with the changes in this PR. As far as I remember, after discussing this at length with Zoomin, we cannot separate the external doc sets more visually like we do now in NCS. The only viable option would be to upload them all separately as individual doc sets, but that is not possible for us. I think changing this title is a good compromise and should help improve a user's navigation clarity and awareness between doc sets.
> 
> @umapraseeda @b-gent any opinions?

Given this input, @gmarull , do you think we can implement the changes from this PR at this stage?

doc/matter/conf.py
gmarull: Discussed today: subtitle was designed to make it clear that some doc sets come from upstream, but that what the user is browsing is actually part of nRF Connect SDK. Introducing a new concept, "platform", may confuse users.

greg-fer: Right, fair point. I've reverted this particular line.

doc/matter/conf.py
```diff
@@ -42,7 +42,7 @@
 html_last_updated_fmt = "%b %d, %Y"
 html_show_sourcelink = True
 html_show_sphinx = False
-html_title = "Matter documentation (nRF Connect SDK)"
+html_title = "Matter SDK documentation (nRF Connect platform)"
```
gmarull, line 45:
```suggestion
html_title = "Matter SDK documentation (nRF Connect SDK)"
```

DeidreCasey (approved):

wiba-nordic (approved):

gmarull (approved):

PR #11523, opened 2023-06-14
lib: nrf_modem: move modem trace initialization as early as possible

eivindj-nordic:
The modem will drop traces if they are not handled before the shared trace memory is full. Initialize the trace module as early as possible to start processing traces earlier.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -639,6 +639,7 @@ Modem libraries
     * The trace backends can now return ``-EAGAIN`` if the write operation can be retried.
     * Fixed a rare bug that caused a deadlock between two threads when one thread sent data while the other received a lot of data quickly.
     * The ``SO_IP_ECHO_REPLY``, ``SO_IPV6_ECHO_REPLY``, ``SO_TCP_SRV_SESSTIMEO`` and ``SO_SILENCE_ALL`` socket option levels to align with the modem option levels.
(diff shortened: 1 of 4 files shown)


eivindj-nordic:
>  There's a couple of things that happen first, the logging of the firmware UUID, which we can move after the callbacks have been invoked, and other callbacks, correct?

Correct.

> So what we want is to run one specific callback before all others?

Yes, more or less. The other callbacks might do modem operations that fill up the trace memory.

lemrey (approved):
Nice!

tokangas (approved):
Great, I was just chatting to @lemrey about doing something like this. 👍

divipillai (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -623,6 +623,7 @@ Modem libraries
     * The trace backends can now return ``-EAGAIN`` if the write operation can be retried.
     * Fixed a rare bug that caused a deadlock between two threads when one thread sent data while the other received a lot of data quickly.
     * The ``SO_IP_ECHO_REPLY``, ``SO_IPV6_ECHO_REPLY``, ``SO_TCP_SRV_SESSTIMEO`` and ``SO_SILENCE_ALL`` socket option levels to align with the modem option levels.
+    * The :ref:`modem_trace_module` is now initialized before the callbacks registered using the  :c:macro:`NRF_MODEM_LIB_ON_INIT` macro are called.
```
MirkoCovizzi, line 626:
```suggestion
    * The :ref:`modem_trace_module` is now initialized before the callbacks registered using the :c:macro:`NRF_MODEM_LIB_ON_INIT` macro are called.
```

MirkoCovizzi (approved):

PR #11531, opened 2023-06-14
Bluetooth: Mesh: Free RAM for mesh samples

Andrewpini:
Reduce stack size and buffers on majority of mesh samples.


samples/bluetooth/mesh/chat/prj.conf
```diff
@@ -10,7 +10,7 @@ CONFIG_NCS_SAMPLES_DEFAULTS=y
 CONFIG_LOG_MODE_DEFERRED=y
 
 # General configuration
-CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=2048
+CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=1800
```
alxelax, line 13: We got an agreement to add a reason as a comment if the default value is changed to something custom.
Does it have the sense to add it here?

alxelax: here and below (to not generate trash from the comments)

alxelax (commented):
Seems Ok to me but need to clarify the situation with comments.

samples/bluetooth/mesh/chat/prj.conf
ludvigsj: This should not be needed at all, default is 2048 (nrf/subsys/bluetooth/controller/Kconfig:
```
# SYSTEM_WORKQUEUE_STACK_SIZE is declared in Zephyr and also here for a second
# time, because Central side Bluetooth sample needs a different workqueue size
# when the SoftDevice Controller is used.
config SYSTEM_WORKQUEUE_STACK_SIZE
	int
	default 2048 if BT_LL_SOFTDEVICE
```
)

Andrewpini: I reduced it to make chat sample fit on 52832. After adding the mesh settings thread it overflowed the RAM. I can add a comment decribing this.

alxelax: Hm... did you try to configure publication over the gatt with this amount of memory?

ludvigsj: Just disable FLASH_SHELL instead. It will be enabled here because FLASH && SHELL and is "more obviously not needed" in my opinion at least. And saves you 8k instead of 200 bytes.

ludvigsj:
Commit message is wrong now as it still mentions reducing workqueue stack in chat sample

samples/bluetooth/mesh/light/prj.conf
```diff
@@ -32,6 +31,7 @@ CONFIG_BT_BUF_ACL_TX_SIZE=37
 CONFIG_BT_OBSERVER=y
 CONFIG_BT_PERIPHERAL=y
 CONFIG_BT_SETTINGS=y
+CONFIG_BT_RX_STACK_SIZE=2700
```
alxelax, line 34: How about adding explanation here?

samples/bluetooth/mesh/light_switch/prj.conf
```diff
@@ -32,6 +30,7 @@ CONFIG_BT_BUF_ACL_TX_SIZE=37
 CONFIG_BT_OBSERVER=y
 CONFIG_BT_PERIPHERAL=y
 CONFIG_BT_SETTINGS=y
+CONFIG_BT_RX_STACK_SIZE=2700
```
alxelax, line 33: and here?

alxelax (approved):

ludvigsj (changes requested):
See my comment about commit message

ludvigsj (approved):

PR #11534, opened 2023-06-14
Register as a Native Wi-Fi driver

krish2718:


drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
```diff
@@ -489,6 +489,19 @@ static int wifi_nrf_drv_main_zep(const struct device *dev)
...
+
+static struct wifi_mgmt_ops wifi_nrf_mgmt_ops = {
+	.scan = wifi_nrf_disp_scan_zep,
+#ifdef CONFIG_NET_STATISTICS_WIFI
+	.get_stats = wifi_nrf_stats_get,
+#endif /* CONFIG_NET_STATISTICS_WIFI */
+	.set_power_save = wifi_nrf_set_power_save,
+	.set_twt = wifi_nrf_set_twt,
+	.reg_domain = wifi_nrf_reg_domain,
+	.get_power_save_config = wifi_nrf_get_power_save_config,
```
rlubos, line 501: It's not clear to me - why are those operations defined here as well as in the `sdk-hostap` `zephyr/src/supp_main.c`?

krish2718: Ideally for Native Wi-Fi, WPA supplicant has to implement these and should take care of calling respective driver ops, but as that framework is missing, for now, we are using the off_api itself but this detail is hidden, but might change in the future. For others its as if WPA supplicant is implementing these ops. 

Added a comment in sdk-hostap to reflect the same, may be I should add one here as well.

VivekUppunda (approved):

sachinthegreen (approved):

sachinthegreen (commented):
Do we need a Changelog entry for this?

krish2718:
> Do we need a Changelog entry for this?

I don't think so, as this is transparent to applications.

rlubos:
This needs a rebase now (there was an earlier sdk-zephyr PR that had been merged).

rlubos (approved):

PR #11543, opened 2023-06-15
tests: edge_impulse new api with shuffling

michalek-no:
tests are run multiple times with random order


tests/lib/edge_impulse/src/main.cpp
```diff
@@ -399,19 +404,20 @@ ZTEST(suite0, test_data_thread)
 	zassert_ok(err, "Cannot start prediction");
 	err = k_sem_take(&test_sem, EI_TEST_SEM_TIMEOUT);
 	zassert_ok(err, "Cannot take semaphore");
+	err = k_thread_join(&thread, K_MSEC(TEST_THREAD_SLEEP_MS * EI_TEST_THREAD_WINDOW_SHIFT));
```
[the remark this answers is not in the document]
MarekPieta: Also, the prediction will not start (and semaphore will not be given) until the last batch of data is provided by the thread (so current way of calculating timeout here is misleading)

MarekPieta (changes requested):

tests/lib/edge_impulse/src/main.cpp
```diff
@@ -14,6 +14,12 @@
 #define EI_TEST_THREAD_WINDOW_SHIFT		2
 #define EI_TEST_WINDOW_SHIFT_CB			1
 
+#define TEST_THREAD_SLEEP_MS  10
+static bool test_init_once = true;
+static size_t timer_fn_calls;
+
+void  ei_run_classifier_mock_init(void);
```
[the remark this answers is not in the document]
michalek-no: prediction_idx needs clearing to keep resulting floats in range

tests/lib/edge_impulse/src/main.cpp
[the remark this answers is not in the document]
michalek-no: Does it mean if we assume no other delays (which isn't true) spawned thread should finish at most single TEST_THREAD_SLEEP_MS after prediction starts?

tests/lib/edge_impulse/src/main.cpp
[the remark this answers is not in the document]
MarekPieta: Thanks for the clarification.

tests/lib/edge_impulse/src/main.cpp
```diff
@@ -399,26 +406,28 @@ ZTEST(suite0, test_data_thread)
 	zassert_ok(err, "Cannot start prediction");
 	err = k_sem_take(&test_sem, EI_TEST_SEM_TIMEOUT);
 	zassert_ok(err, "Cannot take semaphore");
+	err = k_thread_join(&thread, K_MSEC( TEST_THREAD_SLEEP_MS * 21.37));
```
emob-nordic, line 409: ??? please don't troll

tests/lib/edge_impulse/src/main.cpp
```diff
@@ -141,6 +145,10 @@ static void result_ready_cb(int err)
 
 static void *test_init(void)
 {
+	static bool init_once;
+	if(init_once) return NULL;
+	init_once = true;
```
emob-nordic, line 150: Why do we need that? Isn't test_init called once?

``` suggestion
	static bool init_once;

	if(init_once) {
		return NULL;
	}

	init_once = true;
```

emob-nordic (changes requested):

tests/lib/edge_impulse/src/main.cpp
michalek-no: yes, double init of this part is not allowed

emob-nordic (approved):

tests/lib/edge_impulse/src/main.cpp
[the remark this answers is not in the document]
MarekPieta: At most single TEST_THREAD_SLEEP_MS after prediction **ends** (the data input thread has lower priority than ei_wrapper's processing thread, so right after input data is ready, the ei_wrapper starts running unitl finish, then the data input thread will have to only wait for the one last time before finishing).

tests/lib/edge_impulse/CMakeLists.txt
```diff
@@ -35,3 +35,4 @@ target_sources(app PRIVATE src/main.cpp)
 # Test uses ei_test_params.h file from edge_impuse_zip directory to verify if
 # ei_wrapper properly forwards the data between application and EI library.
 target_include_directories(app PRIVATE src/edge_impulse_zip/)
+target_include_directories(app PRIVATE src/edge_impulse_zip/edge-impulse-sdk/)
```
MarekPieta, line 38: Consider placing the header file in `src/edge_impulse_zip/`. The `edge-impulse-sdk` subdirectory contains source files for Edge Impulse library and it should not be directly included by the test (test is expected to access the EI library only via EI wrapper API - except for the newly introduced clearing API).

tests/lib/edge_impulse/src/edge_impulse_zip/edge-impulse-sdk/ei_run_classifier_mock.h
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef _EI_RUN_CLASSIFIER_MOCK_H_
+#define _EI_RUN_CLASSIFIER_MOCK_H_
+
+void  ei_run_classifier_mock_init(void);
```
MarekPieta, line 10: Double space

tests/lib/edge_impulse/src/main.cpp
```diff
@@ -399,26 +409,28 @@ ZTEST(suite0, test_data_thread)
 	zassert_ok(err, "Cannot start prediction");
 	err = k_sem_take(&test_sem, EI_TEST_SEM_TIMEOUT);
 	zassert_ok(err, "Cannot take semaphore");
+	err = k_thread_join(&thread, K_MSEC( TEST_THREAD_SLEEP_MS * 777));
```
MarekPieta, line 412: Please remove space before `TEST_THREAD_SLEEP_MS`

tests/lib/edge_impulse/src/main.cpp
```diff
@@ -434,8 +446,11 @@ ZTEST(suite0, test_data_isr)
 static void setup_fn(void *unused)
 {
 	ARG_UNUSED(unused);
+
 	bool cancelled;
 	int err = ei_wrapper_clear_data(&cancelled);
+	prediction_idx = 0;
+	ei_run_classifier_mock_init();
```
MarekPieta, line 453: Please take the semaphore here to ensure we do not start test with already given semaphore (it would make unit tests more robust)

MarekPieta (changes requested):

tests/lib/edge_impulse/src/main.cpp
```diff
@@ -434,11 +446,16 @@ ZTEST(suite0, test_data_isr)
...
 
 	zassert_false(cancelled, "Prediction was not cancelled");
 	zassert_ok(err, "Cannot clear data");
+	err = k_sem_take(&test_sem, EI_TEST_SEM_TIMEOUT);
```
MarekPieta, line 457: This can only act as an optional cleanup action. In most cases semaphore would not be available here, so you are expected to get `-EAGAIN` (spurious semaphore give should not happen unless the previous test does something fishy). I would also shorten the timeout a bit (I would suggest e.g. 20 ms to avoid blocking test for too long).

MarekPieta (changes requested):

MarekPieta (approved):

emob-nordic (approved):

PR #11544, opened 2023-06-15
nrf_security: Allow enabling PSA crypto without RNG support

joerchan:
Allow enabling PSA crypto (psa_crypto_init) without having RNG support enabled. This is consistent with Oberon PSA core example PSA crypto driver wrapper and their recommendation.

NCSDK-22138


vili-nordic:
commit message says with RNG enabled, when it probably should say without. Otherwise lgtm

vili-nordic (approved):

SebastianBoe (approved):
I suppose this is safe since psa_driver_wrapper_get_random still returns NOT_SUPPORTED for this configuration.

Vge0rge (approved):

PR #11549, opened 2023-06-16
lib: location: Add support for GCI cell

trantanen:
Support added for GCI (surrounding) cells in cellular positioning. `NCELLMEAS=2` is used by default. If we don’t find X cells, do `NCELLMEAS=4,Y`, where `Y=5-X`.

X is requested cell count and run-time configurable through `cell_count` member of `struct location_cellular_config` and build time configurable through `CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_CELL_COUNT`.

Cellular timeout will change to apply only for NCELLMEAS operation ignoring the cloud communication. Same change for Wi-Fi timeout.

GCI search can be disabled completely by setting `CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_CELL_COUNT=0`

Jira: NCSDK-21877

### DIFF
--- a/doc/nrf/libraries/modem/location.rst
+++ b/doc/nrf/libraries/modem/location.rst
@@ -47,6 +47,9 @@ The supported location methods are as follows:
 * Cellular positioning
 
   * Uses :ref:`lte_lc_readme` for getting a list of nearby cellular base stations.
+  * Neighbor cell measurement is performed with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_EXTENDED_COMPLETE` search type.
+    A GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_LIGHT` search type is performed if the previous search did not find enough cells.
+    For more details on GCI search, see :c:member:`location_cellular_config.cell_count`.
   * The ``cloud location`` method handles sending cell information to the selected location service and getting the calculated location back to the device.
 
 * Wi-Fi positioning
@@ -182,6 +185,7 @@ when :c:func:`location_config_defaults_set` function is called:
 * :kconfig:option:`CONFIG_LOCATION_REQUEST_DEFAULT_GNSS_VISIBILITY_DETECTION`
 * :kconfig:option:`CONFIG_LOCATION_REQUEST_DEFAULT_GNSS_PRIORITY_MODE`
 * :kconfig:option:`CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_TIMEOUT`
+* :kconfig:option:`CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_CELL_COUNT`
 * :kconfig:option:`CONFIG_LOCATION_REQUEST_DEFAULT_WIFI_TIMEOUT`
 
 Usage
--- a/include/modem/location.h
+++ b/include/modem/location.h
@@ -339,6 +339,9 @@ struct location_cellular_config {
 	 * @ref location_config_defaults_set function is called and can be changed
 	 * at build time with CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_TIMEOUT configuration.
 	 *
+	 * Timeout only applies to neighbor cell search, not the cloud communication.
+	 * Timeout for the entire location request specified in @ref location_config structure
+	 * is still valid.
 	 * When CONFIG_LOCATION_SERVICE_EXTERNAL is enabled, this timeout stops when
 	 * event LOCATION_EVT_CLOUD_LOCATION_EXT_REQUEST is sent. However, timeout specified in
 	 * @ref location_config structure is still valid.
@@ -354,6 +357,28 @@ struct location_cellular_config {
 	 * This parameter is ignored when CONFIG_LOCATION_SERVICE_EXTERNAL is enabled.
 	 */
 	enum location_service service;
+
+	/**
+	 * @brief Number of cells to be requested for cellular positioning.
+	 *
+	 * @details Default value is 4. It is applied when
+	 * @ref location_config_defaults_set function is called and can be changed
+	 * at build time with CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_CELL_COUNT configuration.
+	 *
+	 * Maximum value is 15.
+	 *
+	 * Zero indicates that only normal neighbor cell search is performed but no GCI search.
+	 *
+	 * If there are less than requested number of neighbor cells (including current cell),
+	 * GCI (surrounding) cells are requested also.
+	 *
+	 * Note that even if there are a lot of cells available, the number of cells
+	 * used for positioning may be lower than the requested number of cells due to
+	 * the behavior of the search algorithm. Also, the number of cells used for
+	 * positioning may be higher than the requested number of cells if there are
+	 * more neighbor cells (including current cell).
+	 */
+	uint8_t cell_count;
 };
 
 /** Wi-Fi positioning configuration. */
@@ -365,6 +390,10 @@ struct location_wifi_config {
 	 * @details Default value is 30000 (30 seconds). It is applied when
 	 * @ref location_config_defaults_set function is called and can be changed
 	 * at build time with CONFIG_LOCATION_REQUEST_DEFAULT_WIFI_TIMEOUT configuration.
+	 *
+	 * Timeout only applies to Wi-Fi scan, not the cloud communication.
+	 * Timeout for the entire location request specified in @ref location_config structure
+	 * is still valid.
 	 *
 	 * When CONFIG_LOCATION_SERVICE_EXTERNAL is enabled, this timeout stops when
 	 * event LOCATION_EVT_CLOUD_LOCATION_EXT_REQUEST is sent. However, timeout specified in
--- a/lib/location/location_core.h
+++ b/lib/location/location_core.h
@@ -36,6 +36,12 @@ struct location_request_info {
 
 	/** Whether to perform fallback for current location request processing. */
 	bool execute_fallback;
+
+	/**
+	 * Device uptime when location request timer expires.
+	 * This is used in cloud location method to calculate timeout for the cloud operation.
+	 */
+	int64_t timeout_uptime;
 };
 
 struct location_method_api {
--- a/lib/location/method_cloud_location.c
+++ b/lib/location/method_cloud_location.c
@@ -30,7 +30,7 @@ struct method_cloud_location_start_work_args {
 	struct k_work work_item;
 	const struct location_wifi_config *wifi_config;
 	const struct location_cellular_config *cell_config;
-	int64_t starting_uptime_ms;
+	int64_t locreq_timeout_uptime;
 };
 
 static struct method_cloud_location_start_work_args method_cloud_location_start_work;
@@ -44,18 +44,12 @@ static void method_cloud_location_positioning_work_fn(struct k_work *work)
 	const struct location_cellular_config *cell_config = work_data->cell_config;
 	struct wifi_scan_info *scan_wifi_info = NULL;
 	struct lte_lc_cells_info *scan_cellular_info = NULL;
-	int64_t scan_start_time;
 	int32_t used_timeout_ms;
 	int err = 0;
 #if defined(CONFIG_LOCATION_METHOD_WIFI)
 	struct k_sem wifi_scan_ready;
 
 	k_sem_init(&wifi_scan_ready, 0, 1);
-#endif
-#if defined(CONFIG_LOCATION_METHOD_CELLULAR)
-	struct k_sem ncellmeas_ready;
-
-	k_sem_init(&ncellmeas_ready, 0, 1);
 #endif
 
 	if (wifi_config != NULL && cell_config != NULL) {
@@ -68,7 +62,6 @@ static void method_cloud_location_positioning_work_fn(struct k_work *work)
 	}
 
 	location_core_timer_start(used_timeout_ms);
-	scan_start_time = k_uptime_get();
 
 #if defined(CONFIG_LOCATION_METHOD_WIFI)
 	if (wifi_config != NULL) {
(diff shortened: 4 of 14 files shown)


lib/location/method_cloud_location.c
```diff
@@ -118,12 +116,11 @@ static void method_cloud_location_positioning_work_fn(struct k_work *work)
 #else
 	struct location_data location;
 	struct location_data location_result = { 0 };
-	int64_t scan_time;
 	struct cloud_service_pos_req params = {
 		.cell_data = scan_cellular_info,
 		.wifi_data = scan_wifi_info,
 		.service = (cell_config != NULL) ? cell_config->service : wifi_config->service,
-		.timeout_ms = used_timeout_ms
+		.timeout_ms = SYS_FOREVER_MS
```
trantanen, line 123: I think we need something other than infinite here to avoid getting stuck if data transfer has issues.
Opinions?

samples/nrf9160/modem_shell/src/location/location_shell.c
```diff
@@ -130,6 +132,7 @@ static struct option long_options[] = {
 	{ "gnss_cloud_pvt", no_argument, 0, LOCATION_SHELL_OPT_GNSS_LOC_CLOUD_PVT },
 	{ "cellular_timeout", required_argument, 0, LOCATION_SHELL_OPT_CELLULAR_TIMEOUT },
 	{ "cellular_service", required_argument, 0, LOCATION_SHELL_OPT_CELLULAR_SERVICE },
+	{ "cellular_count", required_argument, 0, LOCATION_SHELL_OPT_CELLULAR_COUNT },
```
trantanen, line 135: Correct naming would be `cellular_cell_count` but sounds a bit weird and long. I'm ready to change if consistency is preferred.

lib/location/scan_cellular.c
```diff
@@ -17,10 +17,22 @@
...
 static struct lte_lc_ncell neighbor_cells[CONFIG_LTE_NEIGHBOR_CELLS_MAX];
+static struct lte_lc_cell gci_cells[CONFIG_LTE_NEIGHBOR_CELLS_MAX];
 static struct lte_lc_cells_info scan_cellular_info = {
-	.neighbor_cells = neighbor_cells
+	.neighbor_cells = neighbor_cells,
+	.gci_cells = gci_cells
 };
-static struct k_sem *scan_cellular_ready;
+
+/* Indicates when whole cellular scan operation is completed. This is given by caller. */
```
tokangas, line 26: "given" could be replaced with a different word here to avoid confusion, because we're talking about a semaphore which is given/taken. It could be for example "owned".

lib/location/scan_cellular.c
```diff
@@ -99,23 +128,70 @@ int scan_cellular_start(struct k_sem *ncellmeas_ready)
...
+	scan_cellular_cell_count = scan_cellular_cell_count - 1 - scan_cellular_info.ncells_count;
+
+	if (scan_cellular_cell_count > 0) {
+
+		/* Initiate GCI search in a worker. GCI search cannot be initiated here
+		 * because previous NCELLMEAS hasn't been completed yet as giving semaphore
+		 * in lte_lc event handler takes the execution here too quickly.
+		 *
+		 * This function is running in location work queue so GCI search is
+		 * added to system work queue to avoid operation getting stuck.
```
tokangas, line 149: As discussed offline, GCI search can take a long time and it would be better not to execute it in the system work queue.

samples/nrf9160/modem_shell/src/location/location_shell.c
tokangas: I was going to propose `cellular_cell_count` before I read your comment. It is longer, but it is also more correct. To me `cellular_count` sounds a bit weird. 😃 But somebody else can also comment...

lib/location/scan_cellular.c
trantanen: It will just be triggered in system work queue by sending AT command so this would be fine.
However, it would be nice to be able to avoid use of system work queue here for code readability sake.

samples/nrf9160/modem_shell/src/location/location_shell.c
trantanen: If there is someone who thinks cellular_cell_count is fine, I'll change it so will do it :)

include/modem/location.h
```diff
@@ -354,6 +357,29 @@ struct location_cellular_config {
...
+	 *
+	 * @details Default value is 4. It is applied when
+	 * @ref location_config_defaults_set function is called and can be changed
+	 * at build time with CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_CELL_COUNT configuration.
+	 *
+	 * Maximum value is 15.
+	 *
+	 * Zero indicates that only normal neighbor search is performed but no GCI search.
+	 *
+	 * Number of cells to be requested for cellular positioning. If there are less than
```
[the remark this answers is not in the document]
trantanen: True, I extended @brief description at some point and it apparently became the same as this sentence :)

lib/location/method_cloud_location.c
trantanen: As discussed, I'll check if we could set here the remaining time from global location request timeout.

doc/nrf/libraries/modem/location.rst
```diff
@@ -47,6 +47,9 @@ The supported location methods are as follows:
 * Cellular positioning
 
   * Uses :ref:`lte_lc_readme` for getting a list of nearby cellular base stations.
+  * Neighbor cell measurement is performed with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_EXTENDED_COMPLETE` search type.
+    Then, GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_LIGHT` search type is performed if previous search did not find enough cells.
+    More details on how GCI search is done can be seen from :c:member:`location_cellular_config.cell_count`.
```
peknis, line 52:
```suggestion
    A GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_LIGHT` search type is performed if the previous search did not find enough cells.
    For more details on GCI search, see :c:member:`location_cellular_config.cell_count`.
```

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -329,6 +329,10 @@ Modem libraries
 
 * Added the :ref:`modem_battery_readme` library that obtains battery voltage information or notifications from a modem.
 
+* :ref:`lib_location` library:
+
+  * Use of neighbor search modified to use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
```
peknis, line 334:
```suggestion
  * Neighbor cell search is modified to use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
```

include/modem/location.h
```diff
@@ -339,6 +339,9 @@ struct location_cellular_config {
 	 * @ref location_config_defaults_set function is called and can be changed
 	 * at build time with CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_TIMEOUT configuration.
 	 *
+	 * Timeout only applies to neighbor search, not the cloud communication.
```
peknis, line 342:
```suggestion
	 * Timeout only applies to neighbor cell search, not the cloud communication.
```

include/modem/location.h
```diff
@@ -354,6 +357,28 @@ struct location_cellular_config {
...
+	/**
+	 * @brief Number of cells to be requested for cellular positioning.
+	 *
+	 * @details Default value is 4. It is applied when
+	 * @ref location_config_defaults_set function is called and can be changed
+	 * at build time with CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_CELL_COUNT configuration.
+	 *
+	 * Maximum value is 15.
+	 *
+	 * Zero indicates that only normal neighbor search is performed but no GCI search.
```
peknis, line 370:
```suggestion
	 * Zero indicates that only normal neighbor cell search is performed but no GCI search.
```

include/modem/location.h
```diff
@@ -354,6 +357,28 @@ struct location_cellular_config {
...
+	 *
+	 * If there are less than requested number of neighbor cells (including current cell),
+	 * GCI (surrounding) cells are requested also.
+	 *
+	 * Notice that even if there would be a lot of cells available, the number of cells
```
peknis, line 375:
```suggestion
	 * Note that even if there are a lot of cells available, the number of cells
```

peknis (changes requested):

lib/location/method_cloud_location.c
trantanen: Timeout handling modified

lib/location/method_cloud_location.c
```diff
@@ -78,7 +76,7 @@ static void method_cloud_location_positioning_work_fn(struct k_work *work)
 
 #if defined(CONFIG_LOCATION_METHOD_CELLULAR)
 	if (cell_config != NULL) {
-		err = scan_cellular_start(&ncellmeas_ready);
+		err = scan_cellular_start(&ncellmeas_ready, cell_config->cell_count);
```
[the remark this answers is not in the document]
trantanen: Semaphore has been removed from `scan_cellular_start` parameter.

lib/location/scan_cellular.c
trantanen: Mutex implementation added that seemed to resolve these issues

tests/lib/location/src/location_test.c
```diff
@@ -48,6 +48,10 @@ static const char ncellmeas_resp[] =
 	"%NCELLMEAS:0,\"00011B07\",\"26295\",\"00B7\",2300,7,63,31,"
 	"150344527,2300,8,60,29,0,2400,11,55,26,184\r\n";
 
+static const char ncellmeas_gci_resp[] =
+	"%NCELLMEAS:0,\"00023906\",\"24405\",\"5A00\",64,1830518,6400,116,53,16,1840265,1,0,"
+	"\"001E3904\",\"24405\",\"5A00\",65535,0,6400,42,48,7,1840265,0,0\r\n";
+
```
jhirsi, line 54: Action for later, we need to test the max.

lib/lte_link_control/lte_lc.c
```diff
@@ -1461,9 +1464,12 @@ int lte_lc_neighbor_cell_measurement(struct lte_lc_ncellmeas_params *params)
 			LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_COMPLETE),
 		 "Invalid argument, API does not accept enum values directly anymore");
 
+	k_mutex_lock(&ncellmeas_ongoing_mutex, K_SECONDS(10));
```
jhirsi, line 1467: Should we print a warning trace in case of timeout?

tests/lib/location/src/location_test.c
trantanen: Yep, should do that also for normal neighbors...

lib/location/location_core.h
```diff
@@ -36,6 +36,12 @@ struct location_request_info {
 
 	/** Whether to perform fallback for current location request processing. */
 	bool execute_fallback;
+
+	/**
+	 * Device uptime when location request timer expires.
+	 * This is used in cloud location method to calculate timeout for rest operation.
```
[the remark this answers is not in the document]
trantanen: It means REST. Changed to "the cloud operation".

lib/lte_link_control/lte_lc.c
```diff
@@ -68,6 +68,7 @@ static char rai_param[2] = CONFIG_LTE_RAI_REQ_VALUE;
 /* Requested NCELLMEAS params */
 static struct lte_lc_ncellmeas_params ncellmeas_params;
 static bool ncellmeas_ongoing;
+K_MUTEX_DEFINE(ncellmeas_ongoing_mutex);
```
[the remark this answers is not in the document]
jtguggedal: I guess it depends on what happens in the event handlers. Maybe a higher priority thread is waiting and gets to run and issues a new ncell request before the flag is cleared. Would be interesting to understand what is actually going on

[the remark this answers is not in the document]
trantanen: Commit description says what I know. When I'm printing ncellmeas_ongoing, it looks like it should be cleared in the event handler before another thread issues new `lte_lc_neighbor_cell_measurement()`. I cannot figure out why it behaves the way it does even with volatile and atomic_t.

lib/lte_link_control/lte_lc.c
trantanen: There is a warning now with semaphore implementation.

lib/lte_link_control/lte_lc.c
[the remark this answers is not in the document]
trantanen: Mutex doesn't work nicely because you cannot ask for it's state. I've now used semaphore and seems to work nicely. If I set K_NO_WAIT to semaphore timeout, it'll fail as earlier but works nicely with 1s waiting time. Few milliseconds would probably be enough but 1s should be fine.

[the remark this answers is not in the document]
trantanen: I need to see if this implementation without 2nd ncellmeas triggered from a separate work queue is something I can use in unit tests as they seem to fail now.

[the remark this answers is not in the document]
trantanen: Unit tests are now also updated and the issues were related to the fact I forgot to give the semaphore in ncellmeas cancel.

peknis (approved):

lib/lte_link_control/lte_lc.c
```diff
@@ -67,7 +67,8 @@ static char rai_param[2] = CONFIG_LTE_RAI_REQ_VALUE;
 
 /* Requested NCELLMEAS params */
 static struct lte_lc_ncellmeas_params ncellmeas_params;
-static bool ncellmeas_ongoing;
+/* Sempahore value 1 means ncellmeas is not ongoing, and 0 means it's ongoing. */
+K_SEM_DEFINE(ncellmeas_ongoing_sem, 1, 1);
```
tokangas, line 71: Just a suggestion, but the name of the semaphore would be more logical if it would be something like `ncellmeas_idle_sem`. Now it's somewhat counter-intuitive that the name is "ongoing", but value 1 means idle.

tokangas (approved):

jhirsi (approved):

lib/lte_link_control/lte_lc.c
trantanen: Renamed.
I also found an error case where semaphore wasn't given and made a test for that too.

PR #11551, opened 2023-06-16
nrf_security: User supplied PRNG

vili-nordic:
Add option to allow application to provide random data.

ref: NCSDK-21958

### DIFF
--- a/subsys/nrf_security/Kconfig.psa
+++ b/subsys/nrf_security/Kconfig.psa
@@ -88,6 +88,11 @@ config PSA_ITS_ENCRYPTED
 	help
 	  Enables authenticated encryption for PSA Internal Trusted Storage files
 
+config PSA_CRYPTO_DRIVER_ALG_PRNG_TEST
+	bool
+	help
+	  Internal option used for testing. User provided function will be called
+	  to get random bytes.
 
 menuconfig PSA_NATIVE_ITS
 	bool "PSA native Internal Trusted Storage"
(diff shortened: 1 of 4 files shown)


subsys/nrf_security/Kconfig.psa
```diff
@@ -88,6 +88,11 @@ config PSA_ITS_ENCRYPTED
 	help
 	  Enables authenticated encryption for PSA Internal Trusted Storage files
 
+config PSA_CRYPTO_DRIVER_ALG_PRNG_TEST
+	bool
+	prompt "Application provided PRNG"
```
vili-nordic, line 93: ticket says promptless, but I can't set the option in prj.conf then.

Unsure how this was intended

mswarowsky (approved):

vili-nordic:
Another question is what we will do with TF-M. Users/testers can't really supply their function to be linked in there?

subsys/nrf_security/Kconfig.psa
```diff
@@ -88,6 +88,10 @@ config PSA_ITS_ENCRYPTED
 	help
 	  Enables authenticated encryption for PSA Internal Trusted Storage files
 
+config PSA_CRYPTO_DRIVER_ALG_PRNG_TEST
+	bool
+	help
+		Internal option used for testing.
```
Vge0rge, line 94: Nit: You can expand just a bit here if you want.

Vge0rge (approved):

subsys/nrf_security/Kconfig.psa
joerchan: More nit: Indent: 1 tab + 2 spaces

SebastianBoe:
"Another question is what we will do with TF-M. Users/testers can't really supply their function to be linked in there?"

TF-M users can supply application-specific CMake variables.
One option would be to supply a static inline header implementation as is done for user-specific mbedtls configs.

So from driver wrapper do:

#ifdef TFM_TEST_PRNG_HEADER
#include TFM_TEST_PRNG_HEADER
#endif

And from TF-M do

if(TFM_TEST_PRNG_HEADER)
add_definitions(TFM_TEST_PRNG_HEADER=${TFM_TEST_PRNG_HEADER})
endif()

Or something like this.

SebastianBoe (approved):

PR #11560, opened 2023-06-19
net: nrf_provisioning: Fix http request parameter names

juhaylinen:
Update parameter names for maximum TX and RX response sizes to match the latest API.

Increase Http TX buffer size.


tests/subsys/net/lib/nrf_provisioning/src/main.c
```diff
@@ -388,14 +388,14 @@ static int rest_client_request_url_valid(struct rest_client_req_context *req_ctx
 		} else if (strncmp(query_items[idx], "cver=", strlen("cver=")) == 0) {
 			info.cversion = &(query_items[idx][strlen("cver=")]);
 			TEST_ASSERT_GREATER_OR_EQUAL_INT(1, atoi(info.cversion));
-		} else if (strncmp(query_items[idx], "txSize=", strlen("txSize=")) == 0) {
-			info.txSize = &(query_items[idx][strlen("txSize=")]);
+		} else if (strncmp(query_items[idx], "txMaxSize=", strlen("txMaxSize=")) == 0) {
```
jarlamsa, line 391: Rather than using hardcoded string values, would it be better to define a string constant and use those instead?

jarlamsa (commented):
Comment about the strings used in the query.

tests/subsys/net/lib/nrf_provisioning/src/main.c
jarlamsa: Nvm, noticed just now that this was the test code.

jarlamsa (approved):

SeppoTakalo (approved):

PR #11563, opened 2023-06-19
net: nrf_provisioning: Add CoAP support to provisioning client

juhaylinen:
Add CoAP support to nRF provisioning client.
Remove own callback types for CoAP, Http and provisioning codec.
Remove library initialization with SYS_INIT.
Add kconfig CONFIG_NRF_PROVISIONING_AUTO_INIT to initialize the library after the modem library is initialized.


subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+	/* Set up TLS peer verification */
+	enum {
+		NONE = 0,
+		OPTIONAL = 1,
+		REQUIRED = 2,
+	};
+
+	verify = REQUIRED;
+
+	err = setsockopt(fd, SOL_TLS, TLS_PEER_VERIFY, &verify, sizeof(verify));
```
SeppoTakalo, line 105:
```suggestion
	verify = TLS_PEER_VERIFY_REQUIRED;

	err = zsock_setsockopt(fd, SOL_TLS, TLS_PEER_VERIFY, &verify, sizeof(verify));
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+	if (err) {
+		LOG_ERR("Failed to setup peer verification, err %d", errno);
+		return err;
+	}
+
+	/* Associate the socket with the security tag
+	 * we have provisioned the certificate with.
+	 */
+	err = setsockopt(fd, SOL_TLS, TLS_SEC_TAG_LIST, tls_sec_tag, sizeof(tls_sec_tag));
```
SeppoTakalo, line 114:
```suggestion
	err = zsock_setsockopt(fd, SOL_TLS, TLS_SEC_TAG_LIST, tls_sec_tag, sizeof(tls_sec_tag));
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+		return err;
+	}
+
+	if (IS_ENABLED(CONFIG_NRF_PROVISIONING_COAP_TLS_SESSION_CACHE)) {
+		session_cache = TLS_SESSION_CACHE_ENABLED;
+	} else {
+		session_cache = TLS_SESSION_CACHE_DISABLED;
+	}
+
+	err = setsockopt(fd, SOL_TLS, TLS_SESSION_CACHE,
```
SeppoTakalo, line 126:
```suggestion
	err = zsock_setsockopt(fd, SOL_TLS, TLS_SESSION_CACHE,
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+					&session_cache, sizeof(session_cache));
+	if (err) {
+		LOG_ERR("Failed to set TLS_SESSION_CACHE option: %d", errno);
+		return err;
+	}
+
+	err = setsockopt(fd, SOL_TLS, TLS_HOSTNAME, COAP_HOST, strlen(COAP_HOST));
```
SeppoTakalo, line 133:
```suggestion
	err = zsock_setsockopt(fd, SOL_TLS, TLS_HOSTNAME, COAP_HOST, strlen(COAP_HOST));
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+	if (err) {
+		LOG_ERR("Failed to setup TLS hostname, err %d", errno);
+		return err;
+	}
+
```
SeppoTakalo, line 138:
```suggestion
                       /* Enable connection ID */
                       uint32_t dtls_cid = TLS_DTLS_CID_ENABLED;

                       err = zsock_setsockopt(dl->fd, SOL_TLS, TLS_DTLS_CID, &dtls_cid,
                                              sizeof(dtls_cid));

```

Maybe control that by Kconfig, or document that we require Modem FW 1.3.5

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+	return 0;
+}
+
+static int socket_connect(int *const fd)
+{
+	static struct addrinfo hints;
```
SeppoTakalo, line 144:
```suggestion
	static struct zsock_addrinfo hints;
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+#elif defined(CONFIG_NET_IPV6)
+	hints.ai_family = AF_INET6;
+#elif defined(CONFIG_NET_IPV4)
+	hints.ai_family = AF_INET;
+#else
+	hints.ai_family = AF_UNSPEC;
+#endif /* defined(CONFIG_NET_IPV6) && defined(CONFIG_NET_IPV4) */
+
+	hints.ai_socktype = SOCK_DGRAM;
+	st = getaddrinfo(COAP_HOST, PEER_PORT, &hints, &address);
```
SeppoTakalo, line 161:
```suggestion
	st = zsock_getaddrinfo(COAP_HOST, PEER_PORT, &hints, &address);
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+	LOG_DBG("fd: %d", *fd);
+
+	if (st != 0) {
+		LOG_ERR("Unable to resolve address");
+		ret = -EFAULT;
+		goto clean_up;
+	}
+
+	sa = address->ai_addr;
+	inet_ntop(sa->sa_family,
```
SeppoTakalo, line 172:
```suggestion
	zsock_inet_ntop(sa->sa_family,
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+		  (void *)&((struct sockaddr_in *)sa)->sin_addr,
+		  peer_addr,
+		  INET6_ADDRSTRLEN);
+	LOG_DBG("getaddrinfo() %s", peer_addr);
+
+	*fd = socket(address->ai_family, address->ai_socktype, IPPROTO_DTLS_1_2);
```
SeppoTakalo, line 178:
```suggestion
	*fd = zsock_socket(address->ai_family, address->ai_socktype, IPPROTO_DTLS_1_2);
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+
+	/* Setup TLS socket options */
+	ret = tls_setup(*fd);
+	if (ret) {
+		LOG_ERR("Failed to setup TLS socket option");
+		ret = -EACCES;
+		goto clean_up;
+	}
+
+	if (connect(*fd, address->ai_addr, address->ai_addrlen) < 0) {
```
SeppoTakalo, line 193:
```suggestion
	if (zsock_connect(*fd, address->ai_addr, address->ai_addrlen) < 0) {
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+		}
+		goto clean_up;
+	}
+	LOG_INF("Connected");
+
+clean_up:
+
+	if (ret) {
+		if (*fd > -1) {
+			(void)close(*fd);
```
SeppoTakalo, line 208:
```suggestion
			(void)zsock_close(*fd);
```

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,636 @@
...
+
+	return ret;
+}
+
+static int socket_close(int *const fd)
+{
+	int ret;
+
+	if (*fd > -1) {
+		ret = close(*fd);
```
SeppoTakalo, line 221:
```suggestion
		ret = zsock_close(*fd);
```

SeppoTakalo (changes requested):
Looks OK.
I would prefer to use `zsock_` prefixed socket commands, so we don't depend on POSIX_NAMES Kconfig. Some applications might use it, some don't. We need to leave that option to users of that library.

Then I think we need to consider whether using DTLS CID would help there. Do we always need to close the socket to process the content?

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
plskeggs: @eivindj-nordic suggested in my PR for the nrfcloud CoAP library that I not worry about Kconfigs or checking the modem firmware version, but just always try to set it, and if it fails with EOPNOTSUPP then you know CID is not available.

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,643 @@
...
+		return ret;
+	}
+
+	initialized = true;
+
+	return 0;
+}
+
+/* Setup TLS options on a given socket */
+static int tls_setup(int fd)
```
plskeggs, line 83: If you intend to support either `SOCK_DGRAM` or `SOCK_STREAM` (UDP or TCP), then `tls_setup()` is fine.  But if this is really only intended for `SOCK_DGRAM`, then `dtls_setup()` seems more appropriate.

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,643 @@
...
+		} else {
+			LOG_DBG("Operation succesfull");
+			coap_ctx->response_len = 0;
+		}
+		if (last_block) {
+			LOG_DBG("Last packet received");
+		}
+	}
+
+	k_sem_give(&coap_response);
```
plskeggs, line 274: Do you expect that block transfers will occur during any part of provisioning, or will it always be within a single block?

You might want to only give the semaphore when last_block is true, if multiple blocks might occur. Otherwise, your send_coap_request() function will exit after the first block of a multiblock transfer, which might not be what you want. Further, the memcpy into the rx_buf would need to account for that.

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,643 @@
...
+
+	if (payload != NULL) {
+		client_request.payload = (uint8_t *)payload;
+		client_request.len = strlen(payload);
+	}
+
+	while (coap_client_req(client, coap_ctx->connect_socket, NULL, &client_request, -1) ==
+	       -EAGAIN) {
+		LOG_DBG("CoAP client busy");
+		k_sleep(K_MSEC(500));
```
plskeggs, line 300: Should there be a loop counter here that exits after some max count?  Otherwise this could hang forever.

subsys/net/lib/nrf_provisioning/src/nrf_provisioning_coap.c
```diff
@@ -0,0 +1,643 @@
...
+
+static int send_coap_request(struct coap_client *client, uint8_t method, const char *path,
+			     const char *payload,
+			     struct nrf_provisioning_coap_context *const coap_ctx)
+{
+	struct coap_client_request client_request = {
+		.method = method,
+		.confirmable = true,
+		.path = path,
+		.fmt = COAP_CONTENT_FORMAT_APP_CBOR,
```
plskeggs, line 285: Why does this function accept only a const char *payload, but it sets the content format to CBOR?

juhaylinen:
@plskeggs Thank you for the review. I've addressed the findings. Please re-review

plskeggs (approved):
LGTM

SeppoTakalo (approved):

jarlamsa (approved):

SeppoTakalo:
I removed the changelog tag, because provisioning client will be announced separately with cloud releases. It has no entry in the changelog yet.

PR #11564, opened 2023-06-19
wifi: Tie the Wi-Fi on/off to interface up/down

krish2718:
This saves power when Wi-Fi is enabled but unused.

### DIFF
--- a/drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
+++ b/drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
@@ -383,10 +383,27 @@ out:
 	return status;
 }
 
+enum wifi_nrf_status wifi_nrf_fmac_dev_rem_zep(struct wifi_nrf_drv_priv_zep *drv_priv_zep)
+{
+	struct wifi_nrf_ctx_zep *rpu_ctx_zep = NULL;
+
+	rpu_ctx_zep = &drv_priv_zep->rpu_ctx_zep;
+#ifdef CONFIG_NRF700X_RADIO_TEST
+	wifi_nrf_fmac_dev_deinit_rt(rpu_ctx_zep->rpu_ctx);
+	wifi_nrf_fmac_dev_rem_rt(rpu_ctx_zep->rpu_ctx);
+#else
+	wifi_nrf_fmac_dev_deinit(rpu_ctx_zep->rpu_ctx);
+	wifi_nrf_fmac_dev_rem(rpu_ctx_zep->rpu_ctx);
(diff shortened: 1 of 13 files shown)


drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
```diff
@@ -465,25 +464,12 @@ static int wifi_nrf_drv_main_zep(const struct device *dev)
...
-			__func__);
-		goto fmac_deinit;
-	}
 #ifndef CONFIG_NRF700X_RADIO_TEST
 	k_work_init_delayable(&vif_ctx_zep->scan_timeout_work,
 			      wifi_nrf_scan_timeout_work);
 #endif /* !CONFIG_NRF700X_RADIO_TEST */
 
 	return 0;
-fmac_deinit:
```
VivekUppunda, line 481: Hi Chaitanya,
I see that most of the init functions internally invoke the free of resources in the init routines itself and removing this code makes sense. But, would it make more sense to create deinitialization routines to specifically be called during init failures rather than have each init routine clean after itself. I guess it is a coding paradigm, just a point i was pondering.

VivekUppunda (approved):

rado17 (approved):

drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
krish2718: In Zephyr the interface management is not run time i.e., there is no `de-init` in https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/net/net_if.h#L2530.

drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
```diff
@@ -561,6 +587,11 @@ int wifi_nrf_if_stop_zep(const struct device *dev)
 	}
 
 	vif_ctx_zep->if_op_state = WIFI_NRF_FMAC_IF_OP_STATE_DOWN;
+
+	/* Remove the default VIF */
+	if (wifi_nrf_fmac_get_num_vifs(rpu_ctx_zep->rpu_ctx) == 0) {
+		wifi_nrf_fmac_dev_rem(rpu_ctx_zep->rpu_ctx);
```
krish2718, line 593: This crucial part is never reached, as we don't do `del_vif` in stop, so, no power savings. So, planning to move the all functionality that communicates with RPU from init to start/stop, that way this code will be reachable.

sachinthegreen (approved):

rlubos (approved):

PR #11565, opened 2023-06-19
doc: emds: document formula for time estimation

ludvigsj:
This adds documentation of the formula used to compute the required storage time at shutdown in the worst-case.

### DIFF
--- a/doc/nrf/libraries/others/emds.rst
+++ b/doc/nrf/libraries/others/emds.rst
@@ -107,6 +107,51 @@ To initialize the emergency data storage, complete the following steps:
 #. Call :c:func:`emds_prepare`.
 #. Create interrupt or other functionality that will call :c:func:`emds_store`.
 
+.. _emds_readme_application_integration:
+
+Application integration
+***********************
+
+When using EMDS in an application, you need to know the worst case scenario for how long power is required to be available.
+This knowledge makes it possible for you to make good design choices ensuring enough backup power to reach this time requirement.
+
+The easiest way of computing an estimate of the time required to store all entries, in a worst case scenario, is to call the :c:func:`emds_store_time_get` function.
+This function returns a worst-case storage time estimate in microseconds (µs) for a given application.
+For this to work, Kconfig options :kconfig:option:`CONFIG_EMDS_FLASH_TIME_BASE_OVERHEAD_US`, :kconfig:option:`CONFIG_EMDS_FLASH_TIME_ENTRY_OVERHEAD_US` and :kconfig:option:`CONFIG_EMDS_FLASH_TIME_WRITE_ONE_WORD_US` need to be set as described in the `Implementation`_ section.
+The :c:func:`emds_store_time_get` function estimates the required worst-case time to store :math:`n` entries using the following formula:
+
+.. math::
+
+   t_\text{store} = t_\text{base} + \sum_{i = 1}^n \left(t_\text{entry} + t_\text{word}\left(\left\lceil\frac{s_\text{ate}}{s_\text{block}}\right\rceil + \left\lceil\frac{s_i}{s_\text{block}}\right\rceil \right)\right)
+
+where :math:`t_\text{base}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_BASE_OVERHEAD_US`, :math:`t_\text{entry}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_ENTRY_OVERHEAD_US` and :math:`t_\text{word}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_WRITE_ONE_WORD_US`.
+:math:`s_i` is the size of the :math:`i`\ th entry in bytes and :math:`s_\text{block}` is the number of bytes in one word of flash.
+These can be found by looking at datasheets, driver documentation, and the configuration of the application.
+:math:`s_\text{ate}` is the size of the allocation table entry used by the EMDS flash module, which is 8 B.
+
+Example of time estimation
+==========================
+
+Using the formula from the previous section, you can estimate the time required to store all entries for the :ref:`bluetooth_mesh_light_lc` sample running on the nRF52840.
+The following values can be inserted into the formula:
+
+*  Set :math:`t_\text{base}` = 9000 µs.
+   This is the worst case overhead when a store is triggered in the middle of an erase on nRF52840 with :kconfig:option:`CONFIG_SOC_FLASH_NRF_PARTIAL_ERASE` enabled in the sample by default, and should be adjusted when using other configurations.
+*  Set :math:`t_\text{entry}` = 300 µs and :math:`t_\text{word}` = 41 µs. *Note: These values are valid only for this specific chip and configuration, and should be computed for the specific configuration whenever using EMDS.*
+*  The sample uses two entries, one for the RPL with 255 entries (:math:`s_i` = 2040 B) and one for the lightness state (:math:`s_i` = 3 B).
+*  The flash write block size :math:`s_\text{block}` in this case is 4 B, and the ATE size :math:`s_\text{ate}` is 8 B.
+
+This gives the following formula to compute estimated storage time:
+
+.. math::
+   \begin{aligned}
+   t_\text{store} = 9000\text{ µs} &+ \left( 300\text{ µs} + 41\text{ µs} \times \left( \left\lceil\frac{8\text{ B}}{4\text{ B}}\right\rceil + \left\lceil\frac{2040\text{ B}}{4\text{ B}}\right\rceil \right) \right) \\
+   &+ \left( 300\text{ µs} + 41\text{ µs} \times \left( \left\lceil\frac{8\text{ B}}{4\text{ B}}\right\rceil + \left\lceil\frac{3\text{ B}}{4\text{ B}}\right\rceil \right) \right) \\
+   &= 30715\text{ µs}
+   \end{aligned}
+
+Calling the :c:func:`emds_store_time_get` function in the sample automatically computes the result of the formula and returns 30715.
+
 Limitations
 ***********
     The power-fail comparator for the nRF528xx cannot be used with EMDS, as it will prevent the NVMC from performing write operations to flash.
--- a/samples/bluetooth/mesh/light_ctrl/README.rst
+++ b/samples/bluetooth/mesh/light_ctrl/README.rst
@@ -171,8 +171,10 @@ FEM support
 Emergency data storage
 ======================
 
-To build this sample with support for emergency data storage, set ``OVERLAY_CONFIG`` to :file:`overlay-emds.conf`.
+To build this sample with support for emergency data storage (EMDS), set ``OVERLAY_CONFIG`` to :file:`overlay-emds.conf`.
 This will save replay protection list (RPL) data and some of the :ref:`bt_mesh_lightness_srv_readme` data to the emergency data storage instead of to the :ref:`settings_api`.
+When using EMDS, certain considerations need to be taken regarding hardware choices in your application design.
+See :ref:`emds_readme_application_integration` in the EMDS documentation for more information.
 
 See :ref:`cmake_options` for instructions on how to add this option.
 For more information about using configuration overlay files, see :ref:`zephyr:important-build-vars` in the Zephyr documentation.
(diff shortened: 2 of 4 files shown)


doc/nrf/libraries/others/emds.rst
```diff
@@ -94,6 +94,36 @@ EMDS can store data within a guaranteed time, based on the amount of data being
...
+*  Using the formula to compute estimated storage time:
+
+.. math::
+   \begin{aligned}
+   t_\text{store} = 9000\text{ µs} &+ \left( 300\text{ µs} + 41\text{ µs} \times \left( \left\lceil\frac{8\text{ B}}{4\text{ B}}\right\rceil + \left\lceil\frac{2040\text{ B}}{4\text{ B}}\right\rceil \right) \right) \\
+   &+ \left( 300\text{ µs} + 41\text{ µs} \times \left( \left\lceil\frac{8\text{ B}}{4\text{ B}}\right\rceil + \left\lceil\frac{3\text{ B}}{4\text{ B}}\right\rceil \right) \right) \\
+   &= 30715\text{ µs}
+   \end{aligned}
+
+*  Calling :c:func:`emds_store_time_get` in the sample returns 30715, by automatically computing the result of this formula.
```
mia-ko, line 125:
```suggestion

This gives the following formula to compute estimated storage time:

.. math::
   \begin{aligned}
   t_\text{store} = 9000\text{ µs} &+ \left( 300\text{ µs} + 41\text{ µs} \times \left( \left\lceil\frac{8\text{ B}}{4\text{ B}}\right\rceil + \left\lceil\frac{2040\text{ B}}{4\text{ B}}\right\rceil \right) \right) \\
   &+ \left( 300\text{ µs} + 41\text{ µs} \times \left( \left\lceil\frac{8\text{ B}}{4\text{ B}}\right\rceil + \left\lceil\frac{3\text{ B}}{4\text{ B}}\right\rceil \right) \right) \\
   &= 30715\text{ µs}
   \end{aligned}

Calling the :c:func:`emds_store_time_get` function in the sample automatically computes the result of the formula and returns 30715.
```

mia-ko (changes requested):
Not sure if Requirements is the correct heading for this content. Usage? What do you think?

ludvigsj:
> Not sure if Requirements is the correct heading for this content. Usage? What do you think?

Hm... My reasoning for putting it in the "Requirements" section was that this computation is something you'd need to do quite early when developing an application, as it affects hardware design decisions (size of cap/battery). It's a "how to figure out your case-dependent requirements before deciding (how) to implement EMDS in your application" more than "how to use EMDS".

doc/nrf/libraries/others/emds.rst
```diff
@@ -94,6 +94,40 @@ EMDS can store data within a guaranteed time, based on the amount of data being
...
+This function returns a worst-case storage time estimate in microseconds (µs) for a given application.
+For this to work, Kconfig options :kconfig:option:`CONFIG_EMDS_FLASH_TIME_BASE_OVERHEAD_US`, :kconfig:option:`CONFIG_EMDS_FLASH_TIME_ENTRY_OVERHEAD_US` and :kconfig:option:`CONFIG_EMDS_FLASH_TIME_WRITE_ONE_WORD_US` need to be set as described in the `Implementation`_ section.
+The :c:func:`emds_store_time_get` function estimates the required worst-case time to store :math:`n` entries using the following formula:
+
+.. math::
+
+   t_\text{store} = t_\text{base} + \sum_{i = 1}^n \left(t_\text{entry} + t_\text{word}\left(\left\lceil\frac{s_\text{ate}}{s_\text{block}}\right\rceil + \left\lceil\frac{s_i}{s_\text{block}}\right\rceil \right)\right)
+
+where :math:`t_\text{base}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_BASE_OVERHEAD_US`, :math:`t_\text{entry}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_ENTRY_OVERHEAD_US` and :math:`t_\text{word}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_WRITE_ONE_WORD_US`.
```
PavelVPV, line 106: I would suggest to elaborate a bit more a need in `T base` (that it covers case when erase was started right before `emds_store()` call).

PavelVPV, line 106: Could you please also elaborated a bit more Kconfig help strings (or here) for `CONFIG_EMDS_FLASH_TIME_BASE_OVERHEAD_US` and `CONFIG_EMDS_FLASH_TIME_WRITE_ONE_WORD_US` and say that default values are taken from the datasheet for NVMC of the corresponding IC.

doc/nrf/libraries/others/emds.rst
```diff
@@ -94,6 +94,40 @@ EMDS can store data within a guaranteed time, based on the amount of data being
...
+:math:`s_i` is the size of the :math:`i`\ th entry in bytes, :math:`s_\text{ate}` is the size of an allocation table entry in bytes and :math:`s_\text{block}` is the number of bytes in one word of flash.
+These can be found by looking at datasheets, driver documentation, and the configuration of the application.
+
+Example of time estimation
+==========================
+
+Using the formula from the previous section, you can estimate the time required to store all entries for the :ref:`bluetooth_mesh_light_lc` sample running on the nRF52840.
+The following values can be inserted into the formula:
+
+*  Set :math:`t_\text{base}` = 9000 µs, :math:`t_\text{entry}` = 300 µs and :math:`t_\text{word}` = 41 µs. *Note: These values are valid only for this specific chip and configuration, and should be computed for the specific configuration whenever using EMDS.*
```
PavelVPV, line 116: Elaborate why `T base` is 9000 (partial erase is enabled in the sample)?

doc/nrf/libraries/others/emds.rst
```diff
@@ -94,6 +94,40 @@ EMDS can store data within a guaranteed time, based on the amount of data being
...
+*  The sample uses two entries, one for RPL (:math:`s_i` = 2040 B) and one for the lightness state (:math:`s_i` = 3 B).
```
PavelVPV, line 117: Specify number of RPL entries used in the sample?

doc/nrf/libraries/others/emds.rst
```diff
@@ -94,6 +94,40 @@ EMDS can store data within a guaranteed time, based on the amount of data being
...
+The :c:func:`emds_store_time_get` function estimates the required worst-case time to store :math:`n` entries using the following formula:
+
+.. math::
+
+   t_\text{store} = t_\text{base} + \sum_{i = 1}^n \left(t_\text{entry} + t_\text{word}\left(\left\lceil\frac{s_\text{ate}}{s_\text{block}}\right\rceil + \left\lceil\frac{s_i}{s_\text{block}}\right\rceil \right)\right)
+
+where :math:`t_\text{base}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_BASE_OVERHEAD_US`, :math:`t_\text{entry}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_ENTRY_OVERHEAD_US` and :math:`t_\text{word}` is the value specified by :kconfig:option:`CONFIG_EMDS_FLASH_TIME_WRITE_ONE_WORD_US`.
+:math:`s_i` is the size of the :math:`i`\ th entry in bytes, :math:`s_\text{ate}` is the size of an allocation table entry in bytes and :math:`s_\text{block}` is the number of bytes in one word of flash.
+These can be found by looking at datasheets, driver documentation, and the configuration of the application.
```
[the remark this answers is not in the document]
ludvigsj: To be honest I have no idea how you are supposed to find the ate size in a generic way. Unless we assume that people will always use the same driver/fs. The by far simplest way of figuring all this out is just to run the estimation function from the application itself, so I've been vague on these details on purpose.

doc/nrf/libraries/others/emds.rst
```diff
@@ -94,6 +94,40 @@ EMDS can store data within a guaranteed time, based on the amount of data being
...
+
+Example of time estimation
```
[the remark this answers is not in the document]
ludvigsj: Yes, probably. I'll add it.

doc/nrf/libraries/others/emds.rst
ludvigsj: That is covered in the `Implementation` section, which is where you learn how to set these KConfig settings. I didn't want to repeat the information.

ludvigsj: Covered in the `Implementation` section.

doc/nrf/libraries/others/emds.rst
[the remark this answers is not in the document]
PavelVPV: It is constant (as long as the struct doesn't change). We can also add static assert on the size and add a comment to update the documentation if the struct size changes.

doc/nrf/libraries/others/emds.rst
ludvigsj: And to be honest I find it a bit weird that these even have "default" values? It is going to vary from case to case

PavelVPV: Yeah, true, they should depend on platform. I'm also not sure why they are configurable.

mia-ko:
> > Not sure if Requirements is the correct heading for this content. Usage? What do you think?
> 
> Hm... My reasoning for putting it in the "Requirements" section was that this computation is something you'd need to do quite early when developing an application, as it affects hardware design decisions (size of cap/battery). It's a "how to figure out your case-dependent requirements before deciding (how) to implement EMDS in your application" more than "how to use EMDS".

I see :)
I would suggest to use Application integration section then. Starting with a little bit more information on why do this calculation. Like you described it for me.

When developing your application, you would...
Use this library's function xx to... (help you with the specific task...)

ludvigsj:
I'll run the sample with the EMDS overlay to ensure that my _ASSERT doesn't break it, DNM until that is done

samples/bluetooth/mesh/light_ctrl/README.rst
```diff
@@ -173,6 +173,8 @@ Emergency data storage
 
 To build this sample with support for emergency data storage, set ``OVERLAY_CONFIG`` to :file:`overlay-emds.conf`.
 This will save replay protection list (RPL) data and some of the :ref:`bt_mesh_lightness_srv_readme` data to the emergency data storage instead of to the :ref:`settings_api`.
+When using EMDS, certain conciderations need to be taken with regards to hardware choices in your application design.
+See :ref:`emds_readme` for more information.
```
PavelVPV, line 177: This is probably a nit-picking, but if I as a user didn't know anything about EMDS, then I would start reading about it from the beginning. But if I knew about EMDS, then it would make more sense if this sentence point me to the exact subsection with the calculations instead of forcing me to find it among other text.

PavelVPV (approved):

samples/bluetooth/mesh/light_ctrl/README.rst
mia-ko, line 177:
```suggestion
To build this sample with support for emergency data storage (EMDS), set ``OVERLAY_CONFIG`` to :file:`overlay-emds.conf`.
This will save replay protection list (RPL) data and some of the :ref:`bt_mesh_lightness_srv_readme` data to the emergency data storage instead of to the :ref:`settings_api`.
When using EMDS, certain conciderations need to be taken regarding hardware choices in your application design.
See :ref:`emds_readme` for more information.
```

mia-ko: I agree with Pavel.
If considerations we are talking about are all within the Application integration section in EMDS, then it would be good to add label for that section and reference it directly.

mia-ko (changes requested):

mia-ko (commented):
We would also need a changelog entry :)

ludvigsj:
> We would also need a changelog entry :)

About adding documentation? Is that a thing we do?

PavelVPV (approved):

mia-ko:
> > We would also need a changelog entry :)
> 
> About adding documentation? Is that a thing we do?

We might do that, usually in a Documentation section like here for NCS 2.4.0 release: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/releases/release-notes-2.4.0.html#documentation
Very small things usually don't make it to the list, but sections of a certain size is probably a good idea to mention.

E.g.
* Updated:

  *  The :ref:`emds_readme` library documentation with :ref:`emds_readme_application_integration` section about the formula used to compute the required storage time at shutdown in a worst-case scenario.

PR #11566, opened 2023-06-19
drivers: wifi: Store RPU firmware in external flash on nRF5340 and nRF52840

krish2718:
The internal flash utilization is almost full for Matter based
    applications, so, free up memory by storing the firmware patches in the
    external flash on supported platforms, for now only nRF5340 & nRF52840
    are supported, as nRF7002 has limitations access external flash using SPI.
    
    Due to XIP limitations we need to inform flash driver when XIP is being
    used, so, move the FW patch API to a separate C file to avoid using
    other RODATA to minimize this XIP management.

### DIFF
--- a/drivers/wifi/nrf700x/CMakeLists.txt
+++ b/drivers/wifi/nrf700x/CMakeLists.txt
@@ -55,6 +55,7 @@ zephyr_library_sources(
 	zephyr/src/zephyr_work.c
 	zephyr/src/timer.c
 	zephyr/src/zephyr_fmac_main.c
+	zephyr/src/zephyr_fw_load.c
 	zephyr/src/qspi/src/device.c
 	zephyr/src/qspi/src/rpu_hw_if.c
 	zephyr/src/qspi/src/ficr_prog.c
@@ -107,3 +108,8 @@ zephyr_library_sources_ifdef(CONFIG_NRF700X_ON_SPI
 zephyr_library_sources_ifdef(CONFIG_NRF700X_UTIL
 	zephyr/src/zephyr_wifi_util.c
 )
+
+if (CONFIG_NRF_WIFI_PATCHES_EXT_FLASH)
+# Run patches from the external flash (XIP). No need to copy.
+zephyr_code_relocate(FILES zephyr/src/zephyr_fw_load.c LOCATION EXTFLASH_RODATA NOCOPY)
+endif()
--- a/drivers/wifi/nrf700x/Kconfig.nrf700x
+++ b/drivers/wifi/nrf700x/Kconfig.nrf700x
@@ -26,6 +26,23 @@ config NRF_WIFI_IF_AUTO_START
 	bool
 	default y
 
+config NRF_WIFI_PATCHES_EXT_FLASH
+	bool "Store nRF700x FW patches in external flash"
+	# nRF7002 supports SPI based external flash access with no XIP
+	# Disable until Matter DFU changes are ready
+	# default y if BOARD_NRF5340DK_NRF5340_CPUAPP || BOARD_NRF52840DK_NRF52840
+	# For accessing external flash
+	select FLASH
+	# For relocation code to external flash
+	select XIP
+	select BUILD_NO_GAP_FILL
+	select CODE_DATA_RELOCATION
+	select HAVE_CUSTOM_LINKER_SCRIPT
+
+config CUSTOM_LINKER_SCRIPT
+	string "Custom linker script for nRF700x FW patches"
+	default "${ZEPHYR_BASE}/../nrf/drivers/wifi/nrf700x/rpu_fw_patches.ld"
+
 config NRF_WIFI_LOW_POWER
 	bool "Enable low power mode in nRF Wi-Fi chipsets"
 	default y
--- a/drivers/wifi/nrf700x/rpu_fw_patches.ld
+++ b/drivers/wifi/nrf700x/rpu_fw_patches.ld
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/**
+ * @file
+ * @brief Custom Linker command/script file
+ *
+ * Custom Linker script for the Cortex-M platforms.
+ */
+
+#include <zephyr/linker/sections.h>
+#include <zephyr/devicetree.h>
+
+#include <zephyr/linker/linker-defs.h>
+#include <zephyr/linker/linker-tool.h>
+
+#if CONFIG_BOARD_NRF5340DK_NRF5340_CPUAPP || CONFIG_BOARD_NRF52840DK_NRF52840
+/*
+ * nRF53/52 series ship an external flash that can be used for XIP using QSPI/SPI.
+ *
+ * Note: In nRF7002 external flash using is accessible only using SPI but there is no
+ *       support for XIP, so, relocation cannot be used.
+ */
+#if CONFIG_BOARD_NRF5340DK_NRF5340_CPUAPP
+#define EXTFLASH_BASE_ADDR 0x10000000
+#define EXTFLASH_SIZE 0x800000
+#elif CONFIG_BOARD_NRF52840DK_NRF52840
+#define EXTFLASH_BASE_ADDR 0x12000000
+#define EXTFLASH_SIZE 0x800000
+#endif /* CONFIG_BOARD_NRF5340DK_NRF5340_CPUAPP */
+
+#if USE_PARTITION_MANAGER && PM_EXTERNAL_FLASH_ADDRESS
+#include <pm_config.h>
+#define EXTFLASH_ADDRESS (EXTFLASH_BASE_ADDR + PM_EXTERNAL_FLASH_ADDRESS)
+#undef EXTFLASH_SIZE
+#define EXTFLASH_SIZE (PM_EXTERNAL_FLASH_SIZE)
+#else
+#define EXTFLASH_ADDRESS (EXTFLASH_BASE_ADDR)
+#endif /* USE_PARTITION_MANAGER && PM_EXTERNAL_FLASH_ADDRESS */
+
+MEMORY
+{
+     EXTFLASH (wx) : ORIGIN = EXTFLASH_ADDRESS, LENGTH = EXTFLASH_SIZE
+}
+
+#endif /* CONFIG_BOARD_NRF5340DK_NRF5340_CPUAPP || CONFIG_BOARD_NRF52840DK_NRF52840 */
+
+#include <zephyr/arch/arm/aarch32/cortex_m/scripts/linker.ld>
--- a/drivers/wifi/nrf700x/zephyr/src/zephyr_fw_load.c
+++ b/drivers/wifi/nrf700x/zephyr/src/zephyr_fw_load.c
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/**
+ * @brief File containing FW load functions for Zephyr.
+ *
+ * For NRF QSPI NOR special handling is needed for this file as all RODATA of
+ * this file is stored in external flash, so, any use of RODATA has to be protected
+ * by disabling XIP and enabling it again after use. This means no LOG_* macros
+ * (buffered) or buffered printk can be used in this file, else it will crash.
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <zephyr/kernel.h>
+#include <zephyr/device.h>
+#if defined(CONFIG_NRF_WIFI_PATCHES_EXT_FLASH) && defined(CONFIG_NORDIC_QSPI_NOR)
+#include <zephyr/drivers/flash/nrf_qspi_nor.h>
+#endif /* CONFIG_NRF_WIFI_PATCHES_EXT_FLASH */
+
+#include <zephyr_fmac_main.h>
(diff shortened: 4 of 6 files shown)


drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -12,6 +12,14 @@ config WIFI_NRF700X
 	select NET_L2_ETHERNET_MGMT if NETWORKING && NET_L2_ETHERNET
 	select WIFI_USE_NATIVE_NETWORKING if NETWORKING
 	select EXPERIMENTAL if !SOC_SERIES_NRF53X && !SOC_SERIES_NRF91X
+	# For relocation code to external flash
```
markaj-nordic, line 15: If not already done, I would suggest to add another Kconfig, like CONFIG_WIFI_PATCHES_EXT_FLASH, which would select all dependent configs. We would then decide in the application code weather or not we want to offload the internal flash by default or just in some specific configurations.

krish2718: Is there a downside in always using external flash? If not, we can make this default, else, I can protect this with a Kconfig.

markaj-nordic: Well, probably most of (I dare to say all) Matter customers are including external flash in their end products, but I guess we cannot assume this is always the case, especially for smaller applications, leveraging from our WiFi implementation.

krish2718: Sure, but I have enabled this feature only for nRF5340, see the custom linker script, so, it won't be applicable for anything else. Anyways, `CUSTOM_LINKER_SCRIPT` will always be set even if this feature isn't supported e.g., on nRF7002 which is undesirable, added a Kconfig.

drivers/wifi/nrf700x/rpu_fw_patches.ld
```diff
@@ -0,0 +1,34 @@
...
+/* TODO: In nRF7002 external flash using SPI isn't supported by nrfjprog yet */
+#if CONFIG_BOARD_NRF5340DK_NRF5340_CPUAPP
+
+/*
+ * nRF53 series is shipping a QSPI/SPI external flash mapped at 0x1000_0000 that can
+ * be used for XIP
+ */
+MEMORY
+{
+     EXTFLASH (wx) : ORIGIN = 0x10000000, LENGTH = 0x800000
```
Damian-Nordic, line 29: When partition manager is used, a part of the external flash may be reserved for some user partitions. For example, Matter samples have two partitions to store DFU images. I think you may want to enable this feature only when the partition manager is enabled and then do sth like:
```
#include <pm_config.h>
#define EXTFLASH_ADDRESS (PM_EXTERNAL_FLASH_ADDRESS + 0x10000000)
#define EXTFLASH_SIZE (PM_EXTERNAL_FLASH_ADDRESS)
...
EXTFLASH (wx) : ORIGIN = EXTFLASH_ADDRESS, LENGTH = EXTFLASH_SIZE
```
to make sure that the patches are put in the unreserved area of the external flash.

Or fall back to the entire area if the partition manager is disabled.

drivers/wifi/nrf700x/CMakeLists.txt
```diff
@@ -107,3 +107,13 @@ zephyr_library_sources_ifdef(CONFIG_NRF700X_ON_SPI
 zephyr_library_sources_ifdef(CONFIG_NRF700X_UTIL
 	zephyr/src/zephyr_wifi_util.c
 )
+
+# RPU FW header file is included in the zephyr_fmac_main.c
+# Run ext_code from the external flash (XIP). No need to copy.
+zephyr_code_relocate(FILES zephyr/src/zephyr_fmac_main.c LOCATION EXTFLASH_RODATA NOCOPY)
```
Damian-Nordic, line 113: What happens if this is run on nRF7002 DK and the code is relocated, but the linker script does not put the sections in the external flash? Will they be put normally in the internal flash?

drivers/wifi/nrf700x/zephyr/src/ext_flash_init_qspi.c
```diff
@@ -0,0 +1,105 @@
...
+
+	nrf_qspi_ifconfig1_set(NRF_QSPI, &qspi_phy_48mhz);
+
+	/* Enable XiP */
+	nrf_qspi_xip_set(NRF_QSPI, true);
+
+	return 0;
+}
+
+SYS_INIT(qspi_ext_mem_init, POST_KERNEL, CONFIG_KERNEL_INIT_PRIORITY_OBJECTS);
```
Damian-Nordic, line 105: This code seems to assume that you don't use Zephyr's QSPI NOR driver, right? In Zephyr's `nrf_qspi_nor.c`, I can see some code that prevents the flash from init/uninitializing when XPI is enabled, so in such a case you should probably use `nrf_qspi_nor_xip_enable` syscall to make the Zephyr's driver aware of the XIP.

drivers/wifi/nrf700x/rpu_fw_patches.ld
krish2718: Thanks, this is what I was worried about, this clarifies my doubt :).

drivers/wifi/nrf700x/zephyr/src/ext_flash_init_qspi.c
krish2718: Yes, this is just to initialise QSPI else any external flash access will cause a crash. I am not sure if just enabling nrfx flash driver is enough or not? If that works, then we can get rid of this implementation. Let me check that.

drivers/wifi/nrf700x/rpu_fw_patches.ld
Damian-Nordic: I think so. Just spotted a mistake - `PM_EXTERNAL_FLASH_ADDRESS` => PM_EXTERNAL_FLASH_SIZE in the second last line.

drivers/wifi/nrf700x/zephyr/src/ext_flash_init_qspi.c
Damian-Nordic: Not nrfx flash driver, but Zephyr's QSPI driver (which wraps the former; it is enabled with `CONFIG_NORDIC_QSPI_NOR`). Then only `nrf_qspi_nor_xip_enable` call should be needed as the Zephyr's driver should initialize the QSPI peripheral properly on boot. Summoning @anangl to confirm.

krish2718: Using `select NORDIC_QSPI_NOR` and removing the custom initialization doesn't seem to work it still crashes, checking with @anangl on this.

drivers/wifi/nrf700x/zephyr/src/zephyr_fw_load.c
```diff
@@ -0,0 +1,63 @@
...
+	status = wifi_nrf_fmac_fw_load(rpu_ctx,
+				       &fw_info);
+
+	if (status != WIFI_NRF_STATUS_SUCCESS) {
+		printf("%s: wifi_nrf_fmac_fw_load failed\n", __func__);
+		goto out;
+	}
+out:
+#ifdef CONFIG_NRF_WIFI_PATCHES_EXT_FLASH
+	nrf_qspi_nor_xip_enable(flash_dev, false);
```
krish2718, line 59: @anangl do we need a delay here to complete any transactions? In your reference sample there is a sleep of 2secs, but  that will increase the Wi-Fi delay, so, only want to keep minimum delay only if required.

drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -20,6 +20,25 @@ config NRF_WIFI_IF_AUTO_START
...
+	select SPI_NOR if BOARD_NRF7002DK_NRF5340_CPUAPP
+	# For relocation code to external flash
+	select XIP
+	select BUILD_NO_GAP_FILL
+	select CODE_DATA_RELOCATION
+	select HAVE_CUSTOM_LINKER_SCRIPT
+
+if NRF_WIFI_PATCHES_EXT_FLASH
+config CUSTOM_LINKER_SCRIPT
+	string "Custom linker script for nRF700x FW patches"
```
Damian-Nordic, line 38: Better to omit this line in an override to get an error when the original Kconfig name changes.

drivers/wifi/nrf700x/rpu_fw_patches.ld
```diff
@@ -0,0 +1,47 @@
...
+#define EXTFLASH_SIZE 0x800000
+#elif CONFIG_BOARD_NRF52840DK_NRF52840
+#define EXTFLASH_BASE_ADDR 0x12000000
+#define EXTFLASH_SIZE 0x800000
+#endif
+
+#if USE_PARTITION_MANAGER
+#include <pm_config.h>
+#define EXTFLASH_ADDRESS (EXTFLASH_BASE_ADDR + PM_EXTERNAL_FLASH_ADDRESS)
+#define EXTFLASH_SIZE (PM_EXTERNAL_FLASH_SIZE)
```
Damian-Nordic, line 37: Doesn't it raise a warning since you redefine the macro without #undef?

Damian-Nordic (approved):
Looking good!

drivers/wifi/nrf700x/rpu_fw_patches.ld
```diff
@@ -0,0 +1,47 @@
...
+#include <zephyr/devicetree.h>
+
+#include <zephyr/linker/linker-defs.h>
+#include <zephyr/linker/linker-tool.h>
+
+/*
+ * nRF53/52 series ship an external flash that can be used for XIP using QSPI/SPI.
+ *
+ * TODO: In nRF7002 external flash using SPI isn't supported by nrfjprog yet, so,
+ * the code builds but flashing fails.
```
anangl, line 24: XIP is only possible with QSPI.

drivers/wifi/nrf700x/zephyr/src/zephyr_fw_load.c
```diff
@@ -0,0 +1,62 @@
...
+#endif /* CONFIG_NRF_WIFI */
+	/* Load the FW patches to the RPU */
+	status = wifi_nrf_fmac_fw_load(rpu_ctx,
+				       &fw_info);
+
+	if (status != WIFI_NRF_STATUS_SUCCESS) {
+		printf("%s: wifi_nrf_fmac_fw_load failed\n", __func__);
+		goto out;
+	}
+out:
```
anangl, line 56: This looks a bit weird.

drivers/wifi/nrf700x/zephyr/src/zephyr_fw_load.c
anangl: No delay is needed. That sleep in the sample is not related to this call.

krish2718: ok, thanks.

drivers/wifi/nrf700x/rpu_fw_patches.ld
krish2718: ok, will remove nRF7002 references from this for now, till tools support is there and then we can think of implementing the same using flash API.

drivers/wifi/nrf700x/rpu_fw_patches.ld
krish2718: Well, I hadn't tested with `PARTITION_MANAGED_ENABLED`.

drivers/wifi/nrf700x/rpu_fw_patches.ld
markaj-nordic: So how will we implement this for nRF7002 DK (assuming the command line tools are ready) which uses SPI to talk to the external flash? Is it possible without XIP?

markaj-nordic: Ah ok, sorry, @krish2718 you mentioned the raw flash API.

markaj-nordic (approved):

sachinthegreen (approved):

VivekUppunda (approved):

drivers/wifi/nrf700x/zephyr/src/zephyr_fw_load.c
```diff
@@ -0,0 +1,62 @@
...
+	fw_info.lmac_patch_pri.size = sizeof(wifi_nrf_lmac_patch_pri_bimg);
+	fw_info.lmac_patch_sec.data = wifi_nrf_lmac_patch_sec_bin;
+	fw_info.lmac_patch_sec.size = sizeof(wifi_nrf_lmac_patch_sec_bin);
+	fw_info.umac_patch_pri.data = wifi_nrf_umac_patch_pri_bimg;
+	fw_info.umac_patch_pri.size = sizeof(wifi_nrf_umac_patch_pri_bimg);
+	fw_info.umac_patch_sec.data = wifi_nrf_umac_patch_sec_bin;
+	fw_info.umac_patch_sec.size = sizeof(wifi_nrf_umac_patch_sec_bin);
+
+#if defined(CONFIG_NRF_WIFI_PATCHES_EXT_FLASH) && defined(CONFIG_NORDIC_QSPI_NOR)
+	nrf_qspi_nor_xip_enable(flash_dev, true);
```
nordicjm, line 47: This is not very friendly if in the future XIP support will be added to ncs

nordicjm: If a user wants to execute code from QSPI by relocating their code there, and also use the nrf7002 and store the firmware on QSPI, this is going to cause a problem because it will disable XIP after it's finished loading

krish2718: I guess the idea is for all users to call enable XIP before using XIP, so, that should work even for multiple users.

PR #11569, opened 2023-06-20
nrf_security: Fix regression when using builtin enabled PSA crypto

frkv:
There is a regression where legacy APIs for:
 - ECP (ECDH/ECDSA)
 - AES
 - AEAD
 - RSA
 - Hash are pulled in from nrf_cc3xx library or nrf_oberon library regardless if nrf_cc3xx or nrf_oberon enables said feature in PSA crypto driver configurations. This commit enforces stricter rules to how _ALT is enabled based on enabled PSA crypto driver configurations for said drivers.

Note: _ALT based mechanisms are in the process of being removed but they are still a dependency for TLS/DTLS based APIs.

ref: NCSDK-21768


mswarowsky (approved):

subsys/nrf_security/Kconfig.legacy
```diff
@@ -76,81 +76,81 @@ config MBEDTLS_AES_DECRYPT_ALT
...
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_ALG_CHACHA20_POLY1305_CC3XX
 	default y
 	depends on MBEDTLS_CHACHA20_ALT
 
 config MBEDTLS_DHM_ALT
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_HAS_RSA_SUPPORT_CC3XX
```
joerchan, line 118: DHM is RSA?
From the mbedtls docs this looks to be "diffie hellmann merkle", should this be ECC instead?

frkv: yes

subsys/nrf_security/Kconfig.legacy
```diff
@@ -76,81 +76,81 @@ config MBEDTLS_AES_DECRYPT_ALT
...
 config MBEDTLS_CMAC_ALT
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_HAS_CMAC_SUPPORT_CC3XX
 	default y
 
 config MBEDTLS_CCM_ALT
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_HAS_CCM_SUPPORT_CC3XX
```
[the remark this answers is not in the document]
frkv: Fixed with force push

subsys/nrf_security/Kconfig.legacy
```diff
@@ -76,81 +76,81 @@ config MBEDTLS_AES_DECRYPT_ALT
...
 config MBEDTLS_AES_ALT
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_HAS_CIPHER_SUPPORT_CC3XX
 	default y
 
 config MBEDTLS_CMAC_ALT
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_HAS_CMAC_SUPPORT_CC3XX
```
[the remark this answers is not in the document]
frkv: Fixed with force push

subsys/nrf_security/Kconfig.legacy
```diff
@@ -76,81 +76,81 @@ config MBEDTLS_AES_DECRYPT_ALT
...
+	depends on OBERON_BACKEND || PSA_CRYPTO_DRIVER_HAS_ECC_SUPPORT_OBERON || \
+		   CC3XX_BACKEND || PSA_CRYPTO_DRIVER_HAS_ECC_SUPPORT_CC3XX
 	default y
 
 config MBEDTLS_ECDSA_SIGN_ALT
 	bool
-	depends on OBERON_BACKEND || PSA_CRYPTO_DRIVER_OBERON || \
-		   CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on OBERON_BACKEND || PSA_CRYPTO_DRIVER_HAS_ECC_SUPPORT_OBERON || \
+		   CC3XX_BACKEND || PSA_CRYPTO_DRIVER_HAS_ECC_SUPPORT_CC3XX
```
Vge0rge, line 147: Isn't that better fitted PSA_CRYPTO_DRIVER_ALG_ECDSA_CC3XX ? (Maybe also we can also use PSA_CRYPTO_DRIVER_ALG_DETERMINISTIC_ECDSA_CC3XX additionally)

Because the ECC_SUPPORT can also just mean ECDH

joerchan:
@frkv What is the status of this?

joerchan (changes requested):
Needs rebase after the Oberon Kconfig changes.

subsys/nrf_security/Kconfig.legacy
frkv: Kconfig changed to new type (ocrypto Kconfig rewrite)

subsys/nrf_security/Kconfig.legacy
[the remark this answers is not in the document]
frkv: Kconfig changed to new type (ocrypto Kconfig rewrite)

joerchan:
Something must have gone wrong in the change to the Oberon configs:
`/rsa.h:133:10: fatal error: rsa_alt.h` This seems very related to the change.

subsys/nrf_security/Kconfig.legacy
```diff
@@ -76,81 +76,82 @@ config MBEDTLS_AES_DECRYPT_ALT
...
-	depends on OBERON_BACKEND || PSA_CRYPTO_DRIVER_OBERON || \
-		   CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on OBERON_BACKEND || PSA_NEED_OBERON_CHACHA20_POLY1305 || \
+		   CC3XX_BACKEND || PSA_NEED_CC3XX_CHACHA20_POLY1305
 	default y
 
 config MBEDTLS_CHACHAPOLY_ALT
 	bool
-	depends on CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on CC3XX_BACKEND || PSA_NEED_CC3XX_CHACHA20_POLY1305
```
Vge0rge, line 112: Don't we need the same dependencies as the MBEDTLS_POLY1305_ALT here?
Meaning we don't have Oberon the same way as we have it there.

subsys/nrf_security/Kconfig.legacy
```diff
@@ -76,81 +76,82 @@ config MBEDTLS_AES_DECRYPT_ALT
...
-		   CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on OBERON_BACKEND || PSA_NEED_OBERON_ECDH_DRIVER || \
+		   CC3XX_BACKEND || PSA_NEED_CC3XX_ECDH_DRIVER
 	default y
 
 config MBEDTLS_ECDSA_GENKEY_ALT
 	bool
-	depends on OBERON_BACKEND || PSA_CRYPTO_DRIVER_OBERON || \
-		   CC3XX_BACKEND || PSA_CRYPTO_DRIVER_CC3XX
+	depends on OBERON_BACKEND || PSA_NEED_OBERON_ECDSA_DRIVER || \
```
Vge0rge, line 141: For ECDSA I think that what you need is:
PSA_NEED_CC3XX_KEY_PAIR_SECP || PSA_NEED_CC3XX_KEY_PAIR_SECK1
|| PSA_NEED_OBERON_KEY_PAIR_DRIVER

These are relevant to the key generation. So you can just remove the ECDSA_DRIVER, as it is right now it doesn't provide key generation anyway, it only provides the operations I think.

subsys/nrf_security/src/drivers/nrf_cc3xx/Kconfig
```diff
@@ -320,6 +338,12 @@ config PSA_NEED_CC3XX_RSA_PKCS1V15_CRYPT
 	depends on PSA_WANT_ALG_RSA_PKCS1V15_CRYPT
 	depends on PSA_USE_CC3XX_ASYMMETRIC_DRIVER
 
+config PSA_NEED_CC3XX_RSA_DRIVER
```
Vge0rge, line 341: If you want to keep this please add a comment to say that it was only created for some legacy integration. Otherwise you can just move it's depends on the legacy Kconfig file. I am ok with both as long as it has a comment to explain it in case that we keep it.

PR #11571, opened 2023-06-20
manifest: Update nrfxlib with possibility to split mpsl and fem

ahasztag:
Update nrfxlib revision to allow using mpsl and fem as separate libraries.


ankuns (approved):

jciupis (approved):

ahasztag:
@tejlmand Is it OK for you as it is now?

PR #11572, opened 2023-06-20
bluetooth: hids: Increase max HID report map size from 255 to 512 bytes

jacobszpz:
In the bluetooth HID service, the size of a report descriptor is held by a uint8_t, which means that any report descriptors can only have 255 bytes at most. However, the maximum report descriptor size is actually given by the maximum GATT attribute length, which is 512 bytes. This commit replaces the uint8_t for a uint16_t and adds an assertion that enforces the limit set by the Bluetooth specification.

### DIFF
--- a/subsys/bluetooth/services/hids.c
+++ b/subsys/bluetooth/services/hids.c
@@ -14,6 +14,7 @@
 
 #include <zephyr/bluetooth/bluetooth.h>
 #include <zephyr/bluetooth/conn.h>
+#include <zephyr/bluetooth/att.h>
 #include <zephyr/bluetooth/gatt.h>
 #include <zephyr/bluetooth/hci.h>
 #include <zephyr/bluetooth/uuid.h>
@@ -878,6 +879,11 @@ int bt_hids_init(struct bt_hids *hids_obj,
 		 const struct bt_hids_init_param *init_param)
 {
 	LOG_DBG("Initializing HIDS.");
+
+	if (init_param->rep_map.size > BT_ATT_MAX_ATTRIBUTE_LEN) {
+		LOG_WRN("Report map size exceeds max ATT attribute length");
+		return -EMSGSIZE;
+	}
 
 	hids_obj->pm.evt_handler = init_param->pm_evt_handler;
 	hids_obj->cp.evt_handler = init_param->cp_evt_handler;
--- a/include/bluetooth/services/hids.h
+++ b/include/bluetooth/services/hids.h
@@ -340,7 +340,7 @@ struct bt_hids_rep_map {
 	uint8_t const *data;
 
 	/** Size of the map. */
-	uint8_t size;
+	uint16_t size;
 };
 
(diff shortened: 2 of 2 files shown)


alwa-nordic (approved):

jacobszpz:
Changed the commit author name to pass the PR checks

subsys/bluetooth/services/hids.c
```diff
@@ -879,6 +880,9 @@ int bt_hids_init(struct bt_hids *hids_obj,
 {
 	LOG_DBG("Initializing HIDS.");
 
+	__ASSERT(init_param->rep_map.size <= BT_ATT_MAX_ATTRIBUTE_LEN,
```
KAGA164, line 883: This is an API function. I think that assert could be to restrictive here. Please consider returning error code here

jacobszpz: I have changed it to an error code now

alwa-nordic (approved):

subsys/bluetooth/services/hids.c
```diff
@@ -879,6 +880,11 @@ int bt_hids_init(struct bt_hids *hids_obj,
 {
 	LOG_DBG("Initializing HIDS.");
 
+	if (init_param->rep_map.size > BT_ATT_MAX_ATTRIBUTE_LEN) {
+		LOG_WRN("Report map size exceeds max ATT attribute length");
+		return BT_GATT_ERR(BT_ATT_ERR_INVALID_ATTRIBUTE_LEN);
```
alwa-nordic, line 885: Using `BT_GATT_ERR` here is incorrect. The return type of the surrounding function is a negative errno value, like `-EMSGSIZE`. Feel free to use a different errno value if it makes more sense.

```suggestion
		return -EMSGSIZE;
```

alwa-nordic (changes requested):

subsys/bluetooth/services/hids.c
jacobszpz: Change implemented, thanks for the review

KAGA164 (approved):

jacobszpz:
Greetings,

What is left to do to merge this PR?

Thanks

KAGA164:
> Greetings,
> 
> What is left to do to merge this PR?
> 
> Thanks

Hi @jacobszpz, our CI needs to be green. Could you try to rebase?

jacobszpz:
I synced the fork, is this okay?

jacobszpz:
Just rebased instead, please let me know what else is needed

jacobszpz:
I cannot see why the CI failed, so if there's anything wrong please let me know

KAGA164:
> I cannot see why the CI failed, so if there's anything wrong please let me know

@jacobszpz I afraid that you need to rebase one more time. We had the repositories upmerge yesterday also we solved issue with CI under your PR. We will merge your PR as soon as CI will be green

jacobszpz:
CI is okay now, should I rebase?

PR #11576, opened 2023-06-20
Expsing chan survey

martintv:


subsys/bluetooth/controller/Kconfig
```diff
@@ -107,6 +107,11 @@ config BT_CTLR_SDC_LLPM
 	  Low Latency Packet Mode (LLPM) is a Nordic proprietary addition
 	  which lets the application use connection intervals down to 1 ms.
 
+config BT_CTLR_SDC_CHANNEL_SURVEY
+	bool "Enable channel survey"
+	help
```
rugeGerritsen, line 112:
```suggestion
	select EXPERIMENTAL
	help
```

subsys/bluetooth/controller/hci_internal.c
```diff
@@ -1201,6 +1201,8 @@ static uint8_t vs_cmd_put(uint8_t const * const cmd,
 		return sdc_hci_cmd_vs_conn_event_extend((void *)cmd_params);
 	case SDC_HCI_OPCODE_CMD_VS_QOS_CONN_EVENT_REPORT_ENABLE:
 		return sdc_hci_cmd_vs_qos_conn_event_report_enable((void *)cmd_params);
+	case SDC_HCI_OPCODE_CMD_VS_QOS_CHANNEL_SURVEY_ENABLE:
```
rugeGerritsen, line 1204: Please add some ifdefs around here to ensure that the feature is not linked in when not enabled

rugeGerritsen (changes requested):
Also remember to update `vs_zephyr_supported_commands` in `hci_internal.c`

rugeGerritsen (approved):

wiba-nordic (approved):

subsys/bluetooth/controller/hci_driver.c
```diff
@@ -679,6 +679,13 @@ static int configure_supported_features(void)
 		}
 	}
 
+	if (IS_ENABLED(CONFIG_BT_CTLR_SDC_QOS_CHANNEL_SURVEY)) {
+		err = sdc_support_qos_channel_survey();
```
gopiotr, line 683: Shouldn't be there additional `#ifdefs` to avoid warning during building when feature is not enabled:
```
/home/redbeard/ncs/nrf/subsys/bluetooth/controller/hci_driver.c:684:9: error: implicit declaration of function ‘sdc_support_qos_channel_survey’ [-Werror=implicit-function-declaration]
  684 |   err = sdc_support_qos_channel_survey();
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
?

For example:
```c
#ifdef CONFIG_BT_CTLR_SDC_QOS_CHANNEL_SURVEY
	if (IS_ENABLED(CONFIG_BT_CTLR_SDC_QOS_CHANNEL_SURVEY)) {
		err = sdc_support_qos_channel_survey();
		if (err) {
			return -ENOTSUP;
		}
	}
#endif
```

nordicjm: That code is superfluous because the #ifdef and if do the same thing (one is dropped because ifdef is not true, other is dropped because it evaluates to if (0)) so the fix should be that the header be included so that the function definition is always available

gopiotr: @nordicjm - yes, you are right. Moreover I discovered, that my problem came from not updated dragoon repository revision:
https://github.com/nrfconnect/sdk-nrf/blob/8633d05d4d445fe1ab329275849c6320a5bac998/west.yml#L172-L179

When I updated revision to current HEAD (and apply west update), then my problem with lack of linking disappear. So, there is no need to add extra code - just dragoon revision in west manifest should be updated.

PR #11587, opened 2023-06-21
Bluetooth: Mesh: Fix bug DTT spanning multiple elements

Balaklaka:
If a Generic Default Transition Time Server model is not present on the main element of the model, then the Generic Default Transition Time Server model instance that is present on the element with the largest address that is smaller than the address of the main element of the node shall be used; if no model instance is present on any element with an address smaller than the address of the main element, then the Generic Default Transition Time Server is not supported.

### DIFF
--- a/doc/nrf/libraries/bluetooth_services/mesh/gen_dtt_srv.rst
+++ b/doc/nrf/libraries/bluetooth_services/mesh/gen_dtt_srv.rst
@@ -7,9 +7,10 @@ Generic Default Transition Time Server
    :local:
    :depth: 2
 
-The DTT Server provides a common way to specify the state transition time for other models on the same element.
-If other generic models on the same element receive state change commands without transition parameters, they will use the default transition time specified by the DTT Server model.
-This way, the DTT Server can define a consistent transition time for all states on their elements, without depending on client configurations.
+The DTT Server provides a common way to specify the state transition time for other models on any element.
+If a DTT Server is not present on the model's element, use the DTT Server model instance that is present on the element with the largest address that is smaller than the address of the given element.
+This way, if other generic models on any element receive state change commands without transition parameters, they will use the default transition time specified by the DTT Server model.
+The DTT Server can then define a consistent transition time for all states on their elements, without depending on client configurations.
 
 Configuration
 =============
--- a/doc/nrf/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases/release-notes-changelog.rst
@@ -314,6 +314,8 @@ Bluetooth libraries and services
 
     * The :kconfig:option:`BT_MESH_MODEL_SRV_STORE_TIMEOUT` Kconfig option, that is controlling timeout for storing of model states, is replaced by the :kconfig:option:`BT_MESH_STORE_TIMEOUT` Kconfig option.
 
+  * Fixed an issue where the :ref:'bt_mesh_dtt_srv_readme' model could not be found for models spanning multiple elements.
+
 Bootloader libraries
 --------------------
 
--- a/include/bluetooth/mesh/gen_dtt_srv.h
+++ b/include/bluetooth/mesh/gen_dtt_srv.h
@@ -107,21 +107,18 @@ void bt_mesh_dtt_srv_set(struct bt_mesh_dtt_srv *srv, uint32_t transition_time);
 int bt_mesh_dtt_srv_pub(struct bt_mesh_dtt_srv *srv,
 			struct bt_mesh_msg_ctx *ctx);
 
-/** @brief Find the Generic DTT server in a given element.
+/** @brief Find the Generic DTT server over multiple elements.
  *
- * @param[in] elem Element to find the DTT server in.
+ * If the DTT Server is not present on the given element of the model, then the
+ * try to find the DTT Server model instance that is present on the element with
+ * the largest address that is smaller than the address of the given element.
+ *
+ * @param[in] elem Element to start search of the DTT server in.
  *
  * @return A pointer to the DTT server instance, or NULL if no instance is
  * found.
  */
-static inline struct bt_mesh_dtt_srv *
-bt_mesh_dtt_srv_get(const struct bt_mesh_elem *elem)
-{
-	struct bt_mesh_model *model = bt_mesh_model_find(
-		elem, BT_MESH_MODEL_ID_GEN_DEF_TRANS_TIME_SRV);
-
-	return (struct bt_mesh_dtt_srv *)(model ? model->user_data : NULL);
-}
+struct bt_mesh_dtt_srv *bt_mesh_dtt_srv_get(const struct bt_mesh_elem *elem);
 
 /** @brief Get the default transition parameters for the given model.
  *
@@ -134,8 +131,13 @@ static inline bool
 bt_mesh_dtt_srv_transition_get(struct bt_mesh_model *model,
 			       struct bt_mesh_model_transition *transition)
 {
-	struct bt_mesh_dtt_srv *srv =
-		bt_mesh_dtt_srv_get(bt_mesh_model_elem(model));
+	struct bt_mesh_dtt_srv *srv;
+
+	if (IS_ENABLED(CONFIG_BT_MESH_DTT_SRV)) {
+		srv = bt_mesh_dtt_srv_get(bt_mesh_model_elem(model));
+	} else {
+		srv = NULL;
+	}
 
 	transition->time = srv ? srv->transition_time : 0;
 	transition->delay = 0;
--- a/subsys/bluetooth/mesh/gen_dtt_srv.c
+++ b/subsys/bluetooth/mesh/gen_dtt_srv.c
@@ -7,6 +7,7 @@
 #include <string.h>
 #include <bluetooth/mesh/gen_dtt_srv.h>
 #include "model_utils.h"
+#include "mesh/access.h"
 
 static void encode_status(struct net_buf_simple *buf, uint32_t transition_time)
 {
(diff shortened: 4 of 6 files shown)


doc/nrf/libraries/bluetooth_services/mesh/gen_dtt_srv.rst
```diff
@@ -7,9 +7,10 @@ Generic Default Transition Time Server
    :local:
    :depth: 2
 
-The DTT Server provides a common way to specify the state transition time for other models on the same element.
-If other generic models on the same element receive state change commands without transition parameters, they will use the default transition time specified by the DTT Server model.
-This way, the DTT Server can define a consistent transition time for all states on their elements, without depending on client configurations.
+The DTT Server provides a common way to specify the state transition time for other models on any element.
+If the DTT Server is not present on the main element of the model, then the DTT Server model instance that is present on the element with the largest address that is smaller than the address of the main element shall be used.
```
ludvigsj, line 11: What is `main element of the model`? I don't think I've heard of models having a main element before?

include/bluetooth/mesh/gen_dtt_srv.h
```diff
@@ -107,20 +108,32 @@ void bt_mesh_dtt_srv_set(struct bt_mesh_dtt_srv *srv, uint32_t transition_time);
...
  * found.
  */
 static inline struct bt_mesh_dtt_srv *
 bt_mesh_dtt_srv_get(const struct bt_mesh_elem *elem)
 {
-	struct bt_mesh_model *model = bt_mesh_model_find(
-		elem, BT_MESH_MODEL_ID_GEN_DEF_TRANS_TIME_SRV);
-
-	return (struct bt_mesh_dtt_srv *)(model ? model->user_data : NULL);
+	struct bt_mesh_model *model = NULL;
```
alxelax, line 125: This is quite not optimal to have complex functions as inline. I'd recommend moving function body to *.c file and leave here only declaration.
```
struct bt_mesh_dtt_srv *bt_mesh_dtt_srv_get(const struct bt_mesh_elem *elem);
```

Balaklaka: That was my first thought too, but then we have to add this model to all the samples as they only uses this two inline functions.

doc/nrf/libraries/bluetooth_services/mesh/gen_dtt_srv.rst
```diff
@@ -7,9 +7,10 @@ Generic Default Transition Time Server
    :local:
    :depth: 2
 
-The DTT Server provides a common way to specify the state transition time for other models on the same element.
-If other generic models on the same element receive state change commands without transition parameters, they will use the default transition time specified by the DTT Server model.
-This way, the DTT Server can define a consistent transition time for all states on their elements, without depending on client configurations.
+The DTT Server provides a common way to specify the state transition time for other models on any element.
+If the DTT Server is not present on the element of the model, then use the DTT Server model instance that is present on the element with the largest address that is smaller than the address of the given element.
```
mia-ko, line 11: You removed main.
I was just wondering if you by main element meant root element?

include/bluetooth/mesh/gen_dtt_srv.h
ludvigsj: What about
```
#ifdef BT_MESH_DTT_SRV
struct bt_mesh_dtt_srv *
bt_mesh_dtt_srv_get(const struct bt_mesh_elem *elem);
#else
static inline struct bt_mesh_dtt_srv *
bt_mesh_dtt_srv_get(const struct bt_mesh_elem *elem)
{
        return NULL;
}
#endif
```
?

ludvigsj: Or you could `#ifdef` everywhere the function gets called, that's perhaps less "magic"

doc/nrf/libraries/bluetooth_services/mesh/gen_dtt_srv.rst
Balaklaka: No, not root element. The element that the model is defined on.

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -313,6 +313,7 @@ Bluetooth libraries and services
   * Updated:
 
     * The :kconfig:option:`BT_MESH_MODEL_SRV_STORE_TIMEOUT` Kconfig option, that is controlling timeout for storing of model states, is replaced by the :kconfig:option:`BT_MESH_STORE_TIMEOUT` Kconfig option.
+    * Fixed an issue where :ref:'bt_mesh_dtt_srv_readme' could not be found for models spanning multiple elements.
```
mia-ko, line 316:
```suggestion
* Fixed an issue where the :ref:'bt_mesh_dtt_srv_readme' model could not be found for models spanning multiple elements.
```

doc/nrf/libraries/bluetooth_services/mesh/gen_dtt_srv.rst
mia-ko, line 11:
```suggestion
If a DTT Server is not present on the model's element, use the DTT Server model instance that is present on the element with the largest address that is smaller than the address of the given element.
```

mia-ko (changes requested):

include/bluetooth/mesh/gen_dtt_srv.h
```diff
@@ -134,8 +131,11 @@ static inline bool
 bt_mesh_dtt_srv_transition_get(struct bt_mesh_model *model,
 			       struct bt_mesh_model_transition *transition)
 {
-	struct bt_mesh_dtt_srv *srv =
-		bt_mesh_dtt_srv_get(bt_mesh_model_elem(model));
+	struct bt_mesh_dtt_srv *srv = NULL;
+
+#ifdef CONFIG_BT_MESH_DTT_SRV
+	srv = bt_mesh_dtt_srv_get(bt_mesh_model_elem(model));
+#endif
```
PavelVPV, line 138: Just to avoid unnecessary initialization if compiled with server support and keep compiling the whole code even if server support is not enabled.
```suggestion
	struct bt_mesh_dtt_srv *srv;

	if (IS_ENABLED(CONFIG_BT_MESH_DTT_SRV)) {
		srv = bt_mesh_dtt_srv_get(bt_mesh_model_elem(model));
	} else {
		srv = NULL;
	}
```

PavelVPV (commented):
Also please fix this code in scene server:
https://github.com/nrfconnect/sdk-nrf/blob/293fc456416ba942097ffce4cb0af382af624612/subsys/bluetooth/mesh/scene_srv.c#L834-L843

include/bluetooth/mesh/gen_dtt_srv.h
alxelax: It is better to create empty stub instead of linking never used functionality.

Balaklaka:
> Also please fix this code in scene server:
> 
> https://github.com/nrfconnect/sdk-nrf/blob/293fc456416ba942097ffce4cb0af382af624612/subsys/bluetooth/mesh/scene_srv.c#L834-L843

I have a little bit hard of seeing what to fix in this code. It will try to find the TRANS_TIME_SRV in model element, and then search backwards in elements. What am I missing?

subsys/bluetooth/mesh/gen_dtt_srv.c
```diff
@@ -188,3 +189,27 @@ int bt_mesh_dtt_srv_pub(struct bt_mesh_dtt_srv *srv,
...
+
+struct bt_mesh_dtt_srv *bt_mesh_dtt_srv_get(const struct bt_mesh_elem *elem)
+{
+	const struct bt_mesh_comp *comp = bt_mesh_comp_get();
+	uint16_t index;
+
+	index = elem->addr - comp->elem[0].addr;
+	if (index >= comp->elem_count) {
+		return NULL;
+	}
```
PavelVPV, line 201: This should not be needed. We assume that only correct `elem` is passed to the function.

ludvigsj (approved):

doc/nrf/releases/release-notes-changelog.rst
mia-ko: You probably missed that I changed the indentation here :)
Move this line outside the Updated list (on the same level as Updated list).
Usually, the order of changes would be:
Added
Updated
Fixed
Removed

subsys/bluetooth/mesh/gen_dtt_srv.c
Balaklaka: Line 198 is needed or the scene test will fail. And we can not really assume anything as this is a public function call.

PavelVPV: It is like assuming that NULL pointer will never be passed to the API (but `bt_mesh_model_elem` returns NULL pointer of model wasn't found). It is fine to check invalid parameters if they configure something, but I'm not sure where a user can get invalid elem pointer unless explicitly create one and use it. Right, 198 line can stay.

doc/nrf/releases/release-notes-changelog.rst
Balaklaka: I do not think that has been done before, I think they have been under updeted, but I will update it.

PavelVPV (approved):

doc/nrf/releases/release-notes-changelog.rst
Balaklaka: Fixed

mia-ko (approved):

PR #11588, opened 2023-06-21
net: azure_iot_hub: Return error if setting backend is not ready

jtguggedal:
Fix an issue where azure_iot_hub_dps_*_delete() APIs would return zero if -ENOENT was returned by settings_delete(). It was wrongly assumed that the API returned -ENOENT in case the record didn't exist and hence could not be deleted. This is not the case. -ENOENT is returned if the settings backend has not registered with the settings API.

This patch fixes the issue by propagating -ENOENT to the application.

Fixes CIA-996


simensrostad (approved):
lgtm

jtguggedal:
@simensrostad I updated the commit with a test case and running tests only on native_posix. Please have a look.

subsys/net/lib/azure_iot_hub/src/azure_iot_hub_dps.c
```diff
@@ -968,26 +968,19 @@ int azure_iot_hub_dps_hostname_delete(void)
 	int err;
 
 	err = settings_delete(DPS_SETTINGS_KEY "/" DPS_SETTINGS_HOSTNAME_KEY);
-	if (err == -ENOENT) {
-		LOG_DBG("No hostname was stored");
-		goto exit_on_success;
-	} else if (err) {
+	if (err) {
 		LOG_ERR("Failed to delete Azure IoT Hub hostname, error: %d", err);
```
simensrostad, line 973: unrelated but nice to do: Could remove these spaces, the rest of the formatting in this file returns right after the log message.

jtguggedal: Done

tests/subsys/net/lib/azure_iot_hub/dps/src/azure_iot_hub_dps_test.c
```diff
@@ -144,6 +140,19 @@ static void dps_handler(enum azure_iot_hub_dps_reg_status state)
 
 /* Tests */
 
+void test_azure_iot_hub_dps_reset_init_order(void)
```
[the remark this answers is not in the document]
jtguggedal: Yeah, this was a bit lazy. I've split it in two:
* `test_azure_iot_hub_dps_reset_before_init()` 
* `test_azure_iot_hub_dps_reset_after_init()`

fnawratil (approved):

cvinayak:
could this PR be rebased, seems a bit old now?

PR #11589, opened 2023-06-21
802154 phy test use ipc uart

ankuns:
For `nrf5340dk_nrf5340_cpunet` target the shell is directed to
remote_shell. This approach frees serial box of nRFf5340 cpunet and
allows to use other peripherals with the same peripheral id.

This PR contains also bugfixes required this change.


samples/peripheral/802154_phy_test/modules/app_rpc/zephyr/Kconfig
```diff
@@ -0,0 +1,11 @@
+#
+# Copyright (c) 2021 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config APP_RPC
```
ahasztag, line 7: I see that this a general module, which could also be used in other applications, and that is also why this is `APP_RPC` and not `PTT_APP_RPC` - however in this case maybe we should consider moving it, for example to `nrf5340/remote_shell`.

FrancescoSer (approved):

samples/peripheral/802154_phy_test/boards/nrf5340dk_nrf5340_cpunet.overlay
```diff
@@ -0,0 +1,5 @@
+/ {
```
jciupis, line 1: Missing license

samples/peripheral/802154_phy_test/boards/nrf21540dk_nrf52840.conf
```diff
@@ -0,0 +1 @@
+CONFIG_SHELL_PROMPT_UART=">"
```
jciupis, line 1: Missing license. I know this file contains one trivial line but I'm assuming we always have to publish licensed files.

jciupis (approved):
Looks good! Two minor comments below.

ahasztag (approved):
I've left just one comment just to consider, however apart from that everythings seems ok

samples/peripheral/802154_phy_test/boards/nrf5340dk_nrf5340_cpunet.overlay
ankuns: amended

samples/peripheral/802154_phy_test/boards/nrf21540dk_nrf52840.conf
ankuns: amended

PR #11590, opened 2023-06-21
lib: nrf_modem: allow modem shutdown in flight mode

eivindj-nordic:
Allow the modem to be shut down when in flight mode (CFUN=4).


tokangas (approved):
Looks good to me. According to the modem team, both functional modes 0 and 4 should be fine before shutdown.

divipillai (approved):

lemrey (approved):

PR #11594, opened 2023-06-21
doc: new top level hierarchy

greg-fer:
Created new top level hierarchy pages.
Moved existing pages without editing their contents. Updated include paths and redirects.
This is the first task of the rework planned in NCSDK-22319. Page tags and links will be updated in the next PR. NCSDK-22320.

----

This PR only touches on conceptual documentation that is not tied directly to code.
For more information, see https://nordicsemi.atlassian.net/wiki/spaces/TECHDOC/pages/203165695/2023-06-19+-+NCS+New+top+sections

----

- [x] Make the doc build green
- [x] Make the compliance checker happy --> Created follow-up tasks for image resizing, as it is not in scope of this PR: NCSIDB-1026, NCSIDB-1027, and NCSIDB-1028
- [x] Add changelog entry once the approvals are in


gmarull (approved):

greg-fer:
Following feedback from several people, renamed "Hardware guides" to a more accurate "Device configuration guides", as they "Working with" guides or other guides in that section are more about configuring specific devices rather than SoCs and their specifications (as in hardware guides on Infocenter).

Avalei:
Couple quick organizational comments:

## Configuring and building 
Could we reorganize this like the following? I was thinking it might make more sense to put the hardware-related topics first as contextual information before configuration.

- Board support
- Pin control
- Build and configuration system
- Building and programming an application
- Modifying an application
- Multi-image builds
- Bootloaders and Device Firmware Updates

## Testing and optimizing
Could we change "How to use RTT" to "How to connect using RTT"? It would be more consistent with the other headings in this section.

## Device configuration guides
Can we put the "Working with RF front-end modules" after the 52 series one? I think the main device guides should go first.

## About this documentation
Can we  put "Documentation build process" right before "Building the nRF Connect SDK documentation"? I think it makes sense to group those similar topics near each other.

Avalei (approved):

ihansse (approved):

greg-fer:
@Avalei , thanks for the comment. I've implemented all suggestions except:

- Change "How to use RTT" to "How to connect using RTT" -- this I will do in NCSDK-22325
- Put Board and Configuration System page after Board support and Pin control -- I want to keep the overview page at the very top, as it introduces concepts discuss also on Pin control

umapraseeda:
Looks good! Thanks @greg-fer !

umapraseeda (approved):

pdunaj (approved):

PR #11603, opened 2023-06-22
hw_unique_key: Change TF-M to use the HW_UNIQUE_KEY configuration

joerchan:
Change the TF-M hardware unique key configuration to also use the
HW_UNIQUE_KEY option.
Add configuration for TF-M crypto builtin key support, this option
is disabled in minimal mode since we don't need protected storage or
key derivation in this mode.
Refactor crypto keys to compile out HUK support when disabled and remove
the dummy crypto keys.

Use a header file for the derive_key function declaration that is
used by main and implemented in two different source files.


modules/trusted-firmware-m/Kconfig
```diff
@@ -15,6 +15,20 @@
 
 if BUILD_WITH_TFM
 
+config TFM_CRYPTO_BUILTIN_KEYS
+	bool
+	prompt "TF-M crypto builtin keys" if !TFM_PROFILE_TYPE_MINIMAL
```
Vge0rge, line 20: Nit 2: Maybe we can just depends on !TFM_PROFILE_TYPE_MINIMAL then? To having two ifs?

lib/hw_unique_key/Kconfig
```diff
@@ -30,12 +31,13 @@ config HW_UNIQUE_KEY_LOAD
 config HW_UNIQUE_KEY
 	bool "Hardware Unique Keys (HUK)"
 	depends on HAS_HW_NRF_CC310 || HAS_HW_NRF_CC312
-	depends on NRF_CC3XX_PLATFORM
+	depends on NRF_CC3XX_PLATFORM || BUILD_WITH_TFM
 	depends on NRF_SECURITY
-	depends on !ARM_NONSECURE_FIRMWARE
-	depends on MPU_ALLOW_FLASH_WRITE
+	depends on MPU_ALLOW_FLASH_WRITE || BUILD_WITH_TFM
+	depends on !BUILD_WITH_TFM || TFM_CRYPTO_BUILTIN_KEYS
```
mswarowsky, line 37: Why do we have here two times `OR BUILD_WITH_TFM`  and one with`NOT BUILD_WITH_TFM`

Shouldn't the `default y if BUILD_WITH_TFM` be enough?  as all three cases here don't make a difference?

mswarowsky:
Typo in commit message 
`TFC-M`

modules/tfm/tfm/boards/common/crypto_keys.c
```diff
@@ -173,6 +176,7 @@ enum tfm_plat_err_t tfm_plat_builtin_key_get_lifetime_and_slot(mbedtls_svc_key_i
 
 enum tfm_plat_err_t tfm_plat_load_builtin_keys(void)
 {
+#if defined(CONFIG_HW_UNIQUE_KEY) || defined(TFM_PARTITION_INITIAL_ATTESTATION)
```
Vge0rge, line 179: Why did do you want to guard only the defines here?  If none of these are selected the function will just return success without any wanting that something is wrong. I suggest to move guard with the whole function with this and return an error if none of these is chosen.

lib/hw_unique_key/Kconfig
joerchan: the `OR BUILD_WITH_TFM` is the get rid of dependencies when TF-M is enabled, the `NOT BUILD_WITH_TFM` is to add dependencies when TF-M is enabled.

An alternative to this is to instead add it all as a separate entry:

```
config HW_UNIQUE_KEY
	bool "Hardware Unique Keys (HUK)"
	depends on HAS_HW_NRF_CC310 || HAS_HW_NRF_CC312
	depends on NRF_CC3XX_PLATFORM
	depends on NRF_SECURITY
	depends on !ARM_NONSECURE_FIRMWARE
	depends on MPU_ALLOW_FLASH_WRITE
        select FPROTECT if HAS_HW_NRF_ACL
	imply NRFX_NVMC if HAS_HW_NRF_ACL
	
if BUILD_WITH_TFM
config HW_UNIQUE_KEY
      bool "Hardware Unique Key HUK"
      depends on TFM_CRYPTO_BUILTIN_KEYS
      default y
```

mswarowsky: Ahh ok I see, thanks for explaining.

modules/tfm/tfm/boards/common/tfm_hal_platform.c
```diff
@@ -44,8 +44,7 @@ static enum tfm_hal_status_t crypto_platform_init(void)
 		return TFM_HAL_ERROR_BAD_STATE;
 	}
 
-#if CRYPTO_KEY_DERIVATION_MODULE_ENABLED && \
-    !defined(PLATFORM_DEFAULT_CRYPTO_KEYS)
+#ifdef CONFIG_HW_UNIQUE_KEY_RANDOM
 	if (!hw_unique_key_are_any_written()) {
 		SPMLOG_INFMSG("Writing random Hardware Unique Keys to the KMU.\r\n");
 		hw_unique_key_write_random();
```
Vge0rge, line 50: FYI this needs to be updated now with the return code.  hw_unique_key_write_random will now return an error code.

modules/tfm/tfm/boards/common/crypto_keys.c
joerchan: I want to return success if none of these are enabled, otherwise the system will not boot.
But I can move the ifdef guard around the entire thing, I need it only do prevent the warnings of  local variable defined but not used.

modules/trusted-firmware-m/Kconfig
joerchan: Yeah, I can do that also.

modules/tfm/tfm/boards/common/tfm_hal_platform.c
joerchan: You already did, rebase handled the rest.

lib/hw_unique_key/Kconfig
```diff
@@ -30,12 +31,13 @@ config HW_UNIQUE_KEY_LOAD
 config HW_UNIQUE_KEY
 	bool "Hardware Unique Keys (HUK)"
 	depends on HAS_HW_NRF_CC310 || HAS_HW_NRF_CC312
-	depends on NRF_CC3XX_PLATFORM
+	depends on NRF_CC3XX_PLATFORM || BUILD_WITH_TFM
 	depends on NRF_SECURITY
-	depends on !ARM_NONSECURE_FIRMWARE
-	depends on MPU_ALLOW_FLASH_WRITE
+	depends on MPU_ALLOW_FLASH_WRITE || BUILD_WITH_TFM
```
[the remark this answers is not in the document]
joerchan: I think I will skip this suggestion, I don't find that cleaner.

mswarowsky (approved):

modules/tfm/tfm/boards/common/crypto_keys.c
Vge0rge: I see, just for later it makes sense to ask them to add an ifdef  in the caller of this function.

Vge0rge (approved):

PR #10434, opened 2023-02-24
net: l2_wifi: Add connect call with stored credentials.

rado17:
[SHEL-1230] Implement 'connect with stored credentials' facility.

PR #11604, opened 2023-06-22
lib: nrf_modem: `lte_connectivity` -> `lte_net_if`

lemrey:
- Renamed; this isn't really a library per-se but rather a component of the integration layer that adds support for Zephyr's API.
  `lte_connectivity` was not a good name, because LTE connectivity is possible regardless of this component.
- Removed (deprecated) `NRF_MODEM_LIB_NET_IF_AUTO_START`; this shouldn't be supported any longer in NCS.
- Removed header `lte_connectivity.h`. I will add its contents to libmodem's documentation as part of this PR.

Missing doc & changelog updates.


lib/nrf_modem_lib/netif/netif.c
```diff
@@ -452,3 +450,14 @@ int lte_connectivity_options_get(struct conn_mgr_conn_binding *const if_conn, in
 
 	return 0;
 }
+
+/* Bind L2 connectity APIs.
```
eivindj-nordic, line 454: nit:
```suggestion
/* Bind L2 connectivity APIs.
```

lib/nrf_modem_lib/nrf91_sockets.c
```diff
@@ -1132,18 +1137,10 @@ NET_DEVICE_OFFLOAD_INIT(nrf91_socket, "nrf91_socket",
...
-	.disconnect = lte_connectivity_disconnect,
-	.set_opt = lte_connectivity_options_set,
-	.get_opt = lte_connectivity_options_get,
-};
-
-CONN_MGR_CONN_DEFINE(LTE_CONNECTIVITY, &conn_api);
-CONN_MGR_BIND_CONN(nrf91_socket, LTE_CONNECTIVITY);
-#endif /* CONFIG_LTE_CONNECTIVITY */
+#if defined(CONFIG_NRF_MODEM_LIB_NETIF)
+extern struct conn_mgr_conn_api conn_api;
```
eivindj-nordic, line 1141: I'm wondering if this struct, as included as extern, should have a more explicit name, e.g. `nrf_modem_lib_netif_conn_api `etc.?

tests/lib/nrf_modem_lib/lte_connectivity/CMakeLists.txt
```diff
@@ -24,23 +24,23 @@ cmock_handle(${ZEPHYR_NRFXLIB_MODULE_DIR}/nrf_modem/include/nrf_gai_errors.h)
 cmock_handle(${ZEPHYR_NRFXLIB_MODULE_DIR}/nrf_modem/include/nrf_modem_os.h)
```
eivindj-nordic, line 24: nit: the folder with these tests should be renamed.

glarsennordic:
What are your thoughts on naming it `lte_netif` just to emphasize that it is specifically integrating the two?

I know you have placed it in the `modem_lib` folder, but I still suspect that in the long run `lte_netif` will make communication easier than just `nrf_modem_lib/netif` -- especially since we have `net_if` to contend with in Zephyr

lib/nrf_modem_lib/netif/netif.c
glarsennordic: In fact I would go so far as to suggest 

`/* Bind conn_mgr APIs */`

`L2 connectivity` is I think a mistaken reference to an old name

lib/nrf_modem_lib/nrf91_sockets.c
```diff
@@ -1133,17 +1138,9 @@ NET_DEVICE_OFFLOAD_INIT(nrf91_socket, "nrf91_socket",
...
-	.connect = nrf_modem_lib_netif_connect,
-	.disconnect = nrf_modem_lib_netif_disconnect,
-	.set_opt = nrf_modem_lib_netif_options_set,
-	.get_opt = nrf_modem_lib_netif_options_get,
-};
-
+extern struct conn_mgr_conn_api conn_api;
 CONN_MGR_CONN_DEFINE(NRF_MODEM_LIB_NETIF, &conn_api);
 CONN_MGR_BIND_CONN(nrf91_socket, NRF_MODEM_LIB_NETIF);
-#endif /* CONFIG_NRF_MODEM_LIB_NETIF */
```
glarsennordic: nit: I personally find these closure comments `/* CONFIG_NRF_MODEM_LIB_NETIF */` helpful, even for short `#if` spans

glarsennordic (approved):

glarsennordic:
Ah, one final thing to consider: It would be nice if both this and its Wi-Fi counterpart ended up with consistent naming:

https://github.com/nrfconnect/sdk-nrf/pull/10434

Currently it is named `l2_wifi_conn`

lib/nrf_modem_lib/nrf91_sockets.c
lemrey: good point

lemrey:
> What are your thoughts on naming it lte_netif just to emphasize that it is specifically integrating the two?'

That's a good idea.

> Also, since nrf91_sockets.c is also implementing a lot of stuff related to Zephyr's API, should this be moved under the umbrella as well? I guess it does stick mostly to the socket provider implementation, so maybe not

I think it's good to keep the two things separated, socket offloading and `net_if` wrapper; If possible I'd put the connection manager macros in the netif file, but the linker complains and I haven't been able to untangle the macros yet :)

lemrey:
> Ah, one final thing to consider: It would be nice if both this and its Wi-Fi counterpart ended up with consistent naming:
> 
> #10434
> 
> Currently it is named `l2_wifi_conn` or `L2_WIFI_CONNECTIVITY`

yes, definitely

lib/nrf_modem_lib/nrf91_sockets.c
lemrey: ok, I can add it back.

lib/nrf_modem_lib/netif/netif.c
lemrey: ok, will update

lib/nrf_modem_lib/netif/netif.c
```diff
@@ -452,3 +450,14 @@ int lte_connectivity_options_get(struct conn_mgr_conn_binding *const if_conn, in
...
+ * extern in nrf91_sockets.c
+ */
+struct conn_mgr_conn_api conn_api = {
+	.init = nrf_modem_lib_netif_init,
+	.connect = nrf_modem_lib_netif_connect,
+	.disconnect = nrf_modem_lib_netif_disconnect,
+	.set_opt = nrf_modem_lib_netif_options_set,
+	.get_opt = nrf_modem_lib_netif_options_get,
+};
```
rlubos, line 463: Shouldn't the actual Zephyr `net_if` definition for nRF91 also be moved here, if we want to abstract Zephyr integration stuff? Currently done in https://github.com/nrfconnect/sdk-nrf/blob/main/lib/nrf_modem_lib/nrf91_sockets.c#L1129

simensrostad (approved):
Looks good to me

lib/nrf_modem_lib/netif/netif.c
lemrey: It could be done but we'd need to extern all the nrf91 socket APIs, right?

glarsennordic:
> now that that option is deprecated we should add an unconditional call to lte_lc_init there instead, which would let us decouple this component from lte_lc and make lte_net_if_enable just call nrf_modem_lib_init

That seems reasonable to me -- @simensrostad do you have thoughts?

> Now onto _IF_DOWN_DISCONNECT and _IF_DOWN_SHUTDOWN; these options change the behavior of net_if_down, in a way that makes it no longer the opposite of what net_if_up does, which instead always just turns on the modem. I argue that we can remove that option altogether, and let net_if_up and net_if_down be symmetrical in all cases and just manipulate the modem power status.

I don't think we can get rid of this flag -- it arises from a fundamental ambiguity about what the nature of `net_if_up` and `net_if_down` should be for the LTE iface

**From one perspective**, `net_if_up` and `net_if_down` should affect only whether the modem is "ready" to try to connect to LTE, without affecting anything else (such as GNSS).

In that case, `_IF_DOWN_SHUTDOWN` isn't acceptable because it also shuts down `GNSS`. In this case, the call to `modem_init` is technically not quite appropriate, but included out of necessity / convenience. From this perspective, `net_if_up` and `net_if_down` are already *effectively* symmetric because their main purpose was never anything to do with `modem_init` in the first place -- they are *effectively* no-op functions, other than the convenience init.

**From another perspective**, `net_if_up` and `net_if_down` ought to do ***something***, so why not make them affect whether the modem as a whole has power? This violates the principle that `net_if_up/net_if_down` should only affect networking state, but some customers may not care about this, and are happy to gain the extra functionality.

In that case, the call to `modem_init` is not an incidental necessity, but the main purpose of `net_if_up`, and so it makes sense to have `net_if_down` call the corresponding `shutdown` function.

Because both of these perspectives have their merits, I think it is best to keep both options.

lemrey:
>  I'm not sure we should get rid of this flag -- it arises from a fundamental ambiguity about what the nature of net_if_up and net_if_down should be for the LTE iface

I think it's bad to have ambiguity as to what an API does; we should settle on what we want those to do for the LTE modem, and remove flags that modify the API behavior.

If `net_if_up` and `net_if_down` should just let the modem be "ready" to connect to the LTE network without affecting GNSS, then they should do nothing beyond setting the UP flag if the modem is on (or return an error otherwise), and we should accept that the power state of the modem can't be managed with those API.

Otherwise, we accept that they manage the power status.

> Because both of these perspectives have their merits, I think it is probably reasonable to keep both options

I do not agree with this, we should settle on what semantics we want for `net_if_up` and `net_if_down` when we talk about the LTE modem. If we let API change their behavior based on flags, it will be confusing both by itself, since we don't know what the API do anymore, and together with the other APIs that already exist since the user won't know what to us.

lemrey:
I think the last point is important, let me clarify what I mean about not knowing what APIs to use.
This is a problem both for NCS maintainers and for NCS users.

These are all the different ways to turn on the modem:
- do nothing (`NRF_MODEM_LIB_SYS_INIT` (deprecated))
- do nothing (`NET_IF_AUTO_START` (deprecated by this PR))
- call `nrf_modem_lib_init()`
- call `net_if_up()`

To turn off the modem:
- do nothing (fail to attach, `NET_IF_AUTO_DOWN`, with `_MODEM_SHUTDOWN` flag set)
- call `nrf_modem_lib_shutdown()`
- call `net_if_down()` (with `_MODEM_SHUTDOWN` flag set)

To attach to the LTE network:
- do nothing (do any of the 4 actions in "turn on modem", with `LTE_AUTO_INIT_AND_CONNECT` (deprecated))
- call `net_if_up()` (with `NET_IF_AUTO_CONNECT`)
- call `nrf_modem_at_printf("AT+CFUN=1")`
- call `lte_lc_init_and_connect()` (and `lte_lc_connect()`)
- call `conn_mgr_connect()`

To detach from the LTE network:
- call `nrf_modem_at_printf("AT+CFUN=4")`
- call `lte_lc_offline()`
- call `net_if_down()` (with `_LTE_DISCONNECTED` flag)
- call `conn_mgr_disconnect()`

I might have missed some, you can see why.

The issue I see here is that there are too many APIs/ways to do the same thing; this is only further exacerbated by APIs that change behavior based on flags.

lib/nrf_modem_lib/Kconfig.modemlib
```diff
@@ -10,28 +10,6 @@ config HEAP_MEM_POOL_SIZE
...
-	bool "Network interface auto-start"
-	depends on NRF_MODEM_LIB_NET_IF
-	default y
-	help
-	  If this option is enabled, the modem is automatically initialized from the main context
-	  at SYS init.
-	  Initialization of the modem might take some time, so bear in mind that the main context
-	  will be blocked for that duration.
-	  If this option is disabled the modem must be manually initialized by bringing
-	  the network interface's administrative state up by calling net_if_up().
```
glarsennordic: Sorry I missed this in my earlier reviews, I was confused by the naming and description -- Do not remove this, it has nothing to do with `NRF_MODEM_LIB_SYS_INIT`

It is a convenience KConfig that controls the value of `NET_IF_NO_AUTO_START` for the LTE iface. We definitely don't want to get rid of it. 

The description in KConfig is a bit misleading -- It's technically correct because `net_if_up` does initialize the modem, and  `NET_IF_NO_AUTO_START` causes `net_if_up` to be called by Zephyr on boot, but the setting is really about controlling `NET_IF_NO_AUTO_START`.

glarsennordic (changes requested):

lib/nrf_modem_lib/Kconfig.modemlib
glarsennordic: See this line for where it is used:
https://github.com/nrfconnect/sdk-nrf/blob/b88446c2298cc8bc4dd4fabcdb12a03d5cf8de32/lib/nrf_modem_lib/nrf91_sockets.c#L1107

glarsennordic:
I am not really convinced that it is detrimental to allow the `net_if_down` behavior of the integration to be customizable. Nor am I personally convinced we are suffering from an over-abundance of ways to do things.

We have three pretty nice abstraction levels, and the user is free to choose which level of abstraction is appropriate for their application:

1. Direct modem interaction (`nrf_modem_at_printf` and `nrf_modem_lib_xxx`)
2. lte_lc
3. conn_mgr/Zephyr (net_if_up, net_if_down, net_if_connect, net_if_disconnect)

Everything else you list are just flags which cause some combination of `net_if_up, net_if_down, net_if_connect, net_if_disconnect` to be fired automatically when conditions are met.

I personally feel that as long as the descriptions are written clearly, that shouldn't be too confusing. Admittedly, the description for `NRF_MODEM_LIB_NET_IF_AUTO_START` definitely does need to be made clearer IMO. It even fooled me.

I also strongly hesitate to conclude on a single behavior from among the two options. Generally the samples that the Portland team is responsible for don't actually need `net_if_down` to ever be called, so it's not something that currently affects us.

But I would not want to choose `IF_DOWN_SHUTDOWN` and then be caught in a situation that both requires GNSS and necessitates using  `conn_mgr_all_if_down`.

I would also not want to choose `IF_DOWN_DISCONNECT` and then be caught in a situation where `net_if_down` shutting down the modem is extremely convenient.

If push comes to shove, I'd probably choose `IF_DOWN_DISCONNECT` over `IF_DOWN_SHUTDOWN` simply because it is less invasive.

And to be clear, worst case to me would be removing the `modem_init` call in `net_if_up` -- Whatever we choose, I think that should stay.

Because it isn't clear to me which of the above two scenarios is more likely, I am inclined to suggest that we postpone this decision until later, and just stick to the decision we made when we discussed this same topic many months ago

lemrey:
**Automatically turning on the modem in `SYS_INIT`, _is bad idea_.**

Turning on the modem is an operation that blocks for up to two minutes due to the time it takes to execute the update.
In a real product (DFU-enabled), it is not acceptable to block that long before even getting a chance to run any code, and that is it.
Imagine your device getting a firmware a update and then, boom, screen goes black for 2 minutes.
No blinking LEDs, no messages on screen, no sound; the device appears dead for minutes.

I cannot see a good reason why a product developer would want that.
What have they achieved, beyond avoid calling one parameter-less function?
Even that minuscule "advantage" is immediately defeated by the fact that the application needs to know the result of the initialization, in case it failed, and so the user would need to _add_ code for that purpose.

Finally, if the user is really inclined, they can just declare a one-line `SYS_INIT` routine to do this themselves anyway.
Certainly we shouldn't encourage that, given all the above reasons.

Automatically initializing the modem in `SYS_INIT` is counterproductive, and not really usable in a real product, which is why we had just deprecated in NCS v2.4.0...

To conclude, if `net_if_up` and `net_if_down` turn on/off the modem then `NET_IF_NO_AUTO_START` should always be set (and we can remove the Kconfig option to control that).

Then onto the flags (shutdown/lte-disconnect) and Kconfig (auto-connect, auto-down).

> we have three pretty nice abstraction levels, and the user is free to choose which level of abstraction is appropriate for their application 

That is what I am trying to achieve with this PR, currently, like I have shown above, there are 4 ways to do most things, for example attach/detach from the network:
1) libmodem native (AT command)
2) ncs native (`lte_lc`)
3) Zephyr native (`conn_mgr`)
4) `net_if_up` with Kconfig / `net_if_down` with flag

The fourth way consist of `net_if_up` with a Kconfig and `net_if_down` with a flag (which is in addition to the Zephyr APIs `conn_mgr_if_connect` and `conn_mgr_if_disconnect`, right?). Note that these calls may be explicit in the application, or not, depending on more Kconfig options (`AUTO_INIT`, `AUTO_DOWN`). Furthermore, `net_if_up` and `net_if_down` can also be used to turn on the modem (without Kconfig/flags), and to turn off the modem (with a flag).

As you can see, the `net_if_` APIs can basically be used to do anything based on a Kconfig/flag combination. There is no clear semantic to what they do, and I see a problem with that. If you were to read `net_if_up` in the code you'd have no idea what that call does. Is it turning on the modem? Maybe, if it's not already initialized. Is it attaching to LTE? maybe, if you have a Kconfig set. And `net_if_down`, will it turn off the modem? Maybe, if you have set a flag. Will it detach from LTE? Maybe, if you set the other flag.

This is not at all a clear abstraction; it creates confusion as to what these functions do each time they are called.

I think we should at the very least remove the flags to change the behavior of `net_if_down`.
If the user wants to turn off the modem they can call `net_if_down`, if they want to detach they can call `conn_mgr_if_disconnect`.

There is a dedicated API to detach, so what do we achieve by extending another API with a flag to let it do the same thing?

This would at least remove the ambiguity as to what `net_if_down` does (and consequently `NET_IF_AUTO_DOWN`).
Parallelly, we can have `net_if_up` turn on the modem and `conn_mgr_if_connect` attach to the network.

Again, there is a dedicate API to attach, so what do we achieve by extending `net_if_up` with a KConfig to let it do the same thing? `NET_IF_AUTO_CONNECT` makes `net_if_up` ambiguous, so I would like to understand one thing: what are the trying to achieve/solve with this Kconfig? Why should we support `NET_IF_AUTO_CONNECT` for this specific network interface?

lib/nrf_modem_lib/Kconfig.modemlib
lemrey: > It's technically correct because net_if_up does initialize the modem, and NET_IF_NO_AUTO_START causes net_if_up to be called by Zephyr on boot

There is a big problem with that. I have tried to explain it in the comment below.

glarsennordic:
I can unblock this if you remove the changes that aren't related to the rename

We discussed offline that the other changes require a group meeting before moving forward

glarsennordic:
@rlubos I can't use `:ref:` at the moment because conn_mgr still doesn't have docs, so I am just using an inline literal for now instead

PR #11612, opened 2023-06-23
samples: bluetooth: mesh: dfu: Add DFU srv shell cmds for cancelling dfu

PavelVPV:
It is possible that DFU Client and DFU Server states get desynchronized and cancelling DFU procedure from DFU Client using mesh dfd-cancel message is not possible. When running dfu target sample, it may get stuck in phase 2 and won't let to start a new DFU procedure.

This commit enables DFU server shell commands in the target sample to let a user cancel the ongoing DFU procedure.

### DIFF
--- a/doc/nrf/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases/release-notes-changelog.rst
@@ -206,6 +206,10 @@ Bluetooth mesh samples
 
   * Fixed an issue with the sample not fitting into RAM size on the nrf52dk_nrf52832 board.
 
+* :ref:`ble_mesh_dfu_target` sample:
+
+  * Enabled Firmware Update Server shell commands allowing users to cancel DFU on the Target node.
+
 Cryptography samples
 --------------------
 
--- a/samples/bluetooth/mesh/dfu/target/README.rst
+++ b/samples/bluetooth/mesh/dfu/target/README.rst
@@ -181,6 +181,13 @@ The firmware distribution process then succeeds on the Distributor side, if the 
 
 For more information about the firmware distribution process, see :ref:`zephyr:bluetooth_mesh_dfu`.
 
+Interacting with the sample through shell
+-----------------------------------------
+
+This sample enables the Firmware Update Server model shell commands using the :kconfig:option:`CONFIG_BT_MESH_SHELL_DFU_SRV` Kconfig option.
+This allows cancelling of an ongoing DFU transfer using the ``mesh models dfu srv rx-cancel`` shell command, if the ongoing DFU transfer gets stalled.
+See :ref:`bluetooth_mesh_shell` for information about Bluetooth mesh shell commands.
+
 Logging
 =======
 
--- a/samples/bluetooth/mesh/dfu/target/prj.conf
+++ b/samples/bluetooth/mesh/dfu/target/prj.conf
@@ -65,6 +65,19 @@ CONFIG_BT_MESH_DFU_SRV=y
 CONFIG_BT_MESH_DFU_METADATA=y
 CONFIG_BT_MESH_SAR_CFG_SRV=y
 
+# Enable DFU Server shell commands to let user cancel DFU from shell
+CONFIG_BT_MESH_SHELL=y
+CONFIG_BT_MESH_SHELL_DFU_SRV=y
+# Disable other shell commands that get enabled automatically
+CONFIG_BT_MESH_SHELL_BLOB_IO_FLASH=n
+CONFIG_BT_MESH_SHELL_BLOB_SRV=n
+CONFIG_BT_MESH_SHELL_DFU_METADATA=n
+CONFIG_BT_MESH_SHELL_GATT_PROXY=n
+CONFIG_BT_MESH_SHELL_TEST=n
+CONFIG_BT_MESH_SHELL_PROV=n
+# Flash shell module uses 8k of RAM for testing buffers, disable to save RAM
+CONFIG_FLASH_SHELL=n
+
 # Reconfigure transport SAR for DFU
 # Increase probability of delivering segmented messages
 CONFIG_BT_MESH_SAR_TX_UNICAST_RETRANS_COUNT=0x5


samples/bluetooth/mesh/dfu/target/prj.conf
```diff
@@ -65,6 +65,17 @@ CONFIG_BT_MESH_DFU_SRV=y
...
+# Enable DFU Server shell commands to let user cancel DFU from shell
+CONFIG_BT_MESH_SHELL=y
+CONFIG_BT_MESH_SHELL_DFU_SRV=y
+# Disable other shell commands that get enabled automatically
+CONFIG_BT_MESH_SHELL_BLOB_IO_FLASH=n
+CONFIG_BT_MESH_SHELL_BLOB_SRV=n
+CONFIG_BT_MESH_SHELL_DFU_METADATA=n
+CONFIG_BT_MESH_SHELL_GATT_PROXY=n
+CONFIG_BT_MESH_SHELL_TEST=n
+CONFIG_BT_MESH_SHELL_PROV=n
```
ludvigsj, line 77: Check if we want to disable `CONFIG_FLASH_SHELL` as well, if both `CONFIG_FLASH` and `CONFIG_SHELL` is enabled implicitly after your change. It consumes 8k of RAM and we don't need it.

doc/nrf/releases/release-notes-changelog.rst
```diff
@@ -206,6 +206,10 @@ Bluetooth mesh samples
 
   * Fixed an issue with the sample not fitting into RAM size on the nrf52dk_nrf52832 board.
 
+* :ref:`ble_mesh_dfu_target` sample:
+
+  * Enabled Firmware Update Server shell commands to let a user cancel DFU on the target node.
```
mia-ko, line 211:
```suggestion
  * Enabled Firmware Update Server shell commands allowing users to cancel DFU on the Target node.
```

samples/bluetooth/mesh/dfu/target/README.rst
```diff
@@ -181,6 +181,13 @@ The firmware distribution process then succeeds on the Distributor side, if the
 
 For more information about the firmware distribution process, see :ref:`zephyr:bluetooth_mesh_dfu`.
 
+Interacting with the sample through shell
+-----------------------------------------
+
+This sample enables the Firmware Update Server model shell commands using the :kconfig:`CONFIG_BT_MESH_SHELL_DFU_SRV` option.
```
mia-ko, line 187:
```suggestion
This sample enables the Firmware Update Server model shell commands using the :kconfig:option:`CONFIG_BT_MESH_SHELL_DFU_SRV` Kconfig option.
```

samples/bluetooth/mesh/dfu/target/README.rst
```diff
@@ -181,6 +181,13 @@ The firmware distribution process then succeeds on the Distributor side, if the
...
+This allows to cancel the ongoing DFU transfer using `mesh models dfu srv rx-cancel` shell command if it gets stalled.
```
mia-ko, line 188:
```suggestion
This allows cancelling of an ongoing DFU transfer using the ``mesh models dfu srv rx-cancel`` shell command, if the ongoing DFU transfer gets stalled.

```

mia-ko (changes requested):

samples/bluetooth/mesh/dfu/target/README.rst
```diff
@@ -181,6 +181,13 @@ The firmware distribution process then succeeds on the Distributor side, if the
 
 For more information about the firmware distribution process, see :ref:`zephyr:bluetooth_mesh_dfu`.
 
+Interacting with the sample through shell
+-----------------------------------------
+
+This sample enables the Firmware Update Server model shell commands using the :kconfig:option:`CONFIG_BT_MESH_SHELL_DFU_SRV` Kconfig option.
+This allows canelling of an ongoing DFU transfer using the ``mesh models dfu srv rx-cancel`` shell command, if the ongoing DFU transfer gets stalled.
```
ludvigsj, line 188:
```suggestion
This allows cancelling of an ongoing DFU transfer using the ``mesh models dfu srv rx-cancel`` shell command, if the ongoing DFU transfer gets stalled.
```

ludvigsj (changes requested):

mia-ko (approved):

ludvigsj (approved):

alxelax:
Hm... What should a customer do if desynchronization happens in deployed devices in "the field"?
These changes are Ok but they seem mostly like developer hook to solve the problem in the workplace rather than the final product solution.
I guess, we need to provide an approach to how to do this from the mobile phone(or other customer front-end solution).
Probable custom command?

alxelax (approved):

ludvigsj:
> > Hm... What should a customer do if desynchronization happens in deployed devices in "the field"? These changes are Ok but they seem mostly like developer hook to solve the problem in the workplace rather than the final product solution. I guess, we need to provide an approach to how to do this from the mobile phone(or other customer front-end solution). Probably custom command?
> 
> It is possible to cancel DFU on any device by sending Firmware Update Cancel message. This is possible to do from the distributor sample as well, I missed that. So perhaps this change is not needed at all.

No, this is not possible. The distributor needs to be in a very specific state for it to allow you to send this message, it doesn't just allow you to send the message randomly to a given device with the current shell interface. Basically, it has to be in the middle of a transfer to that address for it to allow you to send the cancel message. Of course, if you end up out of sync, then you'll never be able to get back into this state, because the transfer will never start again because the target will return "busy" all the time until you cancel (which you can't do because you lost your state).

PavelVPV:
I will close the PR. Perhaps, when I created the issue, we didn't have DFU client shell commands, but now it is possible to call `mesh models dfu cli cancel <addr>` to cancel DFU on a specific target node. I think this may lead to wrong understanding of mesh DFU if this command will exist in the sample. I'll look if we need to additionally mention the canceling command on the dfu client side or in the dfu procedure guide.

PavelVPV:
> > > Hm... What should a customer do if desynchronization happens in deployed devices in "the field"? These changes are Ok but they seem mostly like developer hook to solve the problem in the workplace rather than the final product solution. I guess, we need to provide an approach to how to do this from the mobile phone(or other customer front-end solution). Probably custom command?
> > 
> > 
> > It is possible to cancel DFU on any device by sending Firmware Update Cancel message. This is possible to do from the distributor sample as well, I missed that. So perhaps this change is not needed at all.
> 
> No, this is not possible. The distributor needs to be in a very specific state for it to allow you to send this message, it doesn't just allow you to send the message randomly to a given device with the current shell interface. Basically, it has to be in the middle of a transfer to that address for it to allow you to send the cancel message. Of course, if you end up out of sync, then you'll never be able to get back into this state, because the transfer will never start again because the target will return "busy" all the time until you cancel (which you can't do because you lost your state).

This is the shell command implementation which allows to call `bt_mesh_dfu_cli_cancel` for a specific device:
https://github.com/zephyrproject-rtos/zephyr/blob/a7498c39e2d34d33a463fc861f1f39f0a626b8da/subsys/bluetooth/mesh/shell/dfu.c#L788-L814

When `ctx` is provided (like in `cmd_dfu_tx_cancel`), `bt_mesh_dfu_cli_cancel` function will send the Firmware Update Cancel message to a specific device bypassing all states:
https://github.com/zephyrproject-rtos/zephyr/blob/a7498c39e2d34d33a463fc861f1f39f0a626b8da/subsys/bluetooth/mesh/dfu_cli.c#L1065-L1083

Firmware Update Server model will drop all states once it receives the message (unless it is already in idle state):
https://github.com/zephyrproject-rtos/zephyr/blob/a7498c39e2d34d33a463fc861f1f39f0a626b8da/subsys/bluetooth/mesh/dfu_srv.c#L374-L387

ludvigsj:
> I will close the PR. Perhaps, when I created the issue, we didn't have DFU client shell commands, but now it is possible to call `mesh models dfu cli cancel <addr>` to cancel DFU on a specific target node. I think this may lead to wrong understanding of mesh DFU if this command will exist in the sample. I'll look if we need to additionally mention the canceling command on the dfu client side or in the dfu procedure guide.

I _think_ that when this issue was created, the way the shell worked was that if you enabled DFD server, the DFD server took ownership of the DFU client, and the DFU client commands would become unavailable from the shell. If that is no longer the case, then I agree, we don't need this PR.

PavelVPV:
> > I will close the PR. Perhaps, when I created the issue, we didn't have DFU client shell commands, but now it is possible to call `mesh models dfu cli cancel <addr>` to cancel DFU on a specific target node. I think this may lead to wrong understanding of mesh DFU if this command will exist in the sample. I'll look if we need to additionally mention the canceling command on the dfu client side or in the dfu procedure guide.
> 
> I _think_ that when this issue was created, the way the shell worked was that if you enabled DFD server, the DFD server took ownership of the DFU client, and the DFU client commands would become unavailable from the shell. If that is no longer the case, then I agree, we don't need this PR.

Yes, now the commands are untied from the model instance. A user can choose any dfu client instance instantiated on a device.

PR #11617, opened 2023-06-26
various: Enclose string if checks in quotes

nordicjm:
Fixes possible issues whereby string variables are checked without being properly encased in quote marks.

This is needed because unescaped variables will be treated as variable names, then errors will be emitted when they are dereferenced and those variables do not exist.

An example of what happens with bad code when configuring a build:
```
   *************************
   * Running CMake for ncp *
   *************************

CMake Error at CMakeLists.txt:10 (if):
  if given arguments:

    "STREQUAL" "nrf52840dongle_nrf52840"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
CMake Error at cmake/modules/sysbuild_extensions.cmake:405 (message):
  CMake configure failed for Zephyr project: ncp
```


Damian-Nordic:
Could you improve the PR description? That is, what kind of issues this might cause? At least I wasn't aware this might be wrong so would like to learn something :)

tomchy (approved):

applications/matter_weather_station/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set(APPLICATION_CONFIG_DIR "${CMAKE_CURRENT_SOURCE_DIR}/configuration/\${BOARD}"
 set(multiprotocol_rpmsg_KCONFIG_ROOT "\\\${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/config/nrfconnect/chip-module/Kconfig.multiprotocol_rpmsg.root")
 set(hci_rpmsg_KCONFIG_ROOT "\\\${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/config/nrfconnect/chip-module/Kconfig.hci_rpmsg.root")
 
-if(OVERLAY_CONFIG STREQUAL "overlay-factory_data.conf")
+if("${OVERLAY_CONFIG}" STREQUAL "overlay-factory_data.conf")
```
SebastianBoe, line 17: How is this a possible issue? :)

https://cmake.org/cmake/help/latest/command/if.html#strequal

nordicjm: Not all of them are issues, it is the ones using variables outside of quotes that are issues, I've updated the rest so that the style throughout the code remains uniform.

nordicjm:
@Damian-Nordic I have added a description and example to the original message.

carlescufi (changes requested):
The change itself looks good to me, but could you explain in the commit message why this is necessary? Right now it is not clear to me, and I'd like to understand it. Thanks!

samples/bluetooth/direct_test_mode/CMakeLists.txt
```diff
@@ -5,11 +5,11 @@
 #
 cmake_minimum_required(VERSION 3.20.0)
 
-if (DEFINED SHIELD AND ${BOARD} STREQUAL "nrf5340dk_nrf5340_cpunet")
+if(DEFINED SHIELD AND "${BOARD}" STREQUAL "nrf5340dk_nrf5340_cpunet")
   string(REPLACE " " ";" SHIELD_LIST "${SHIELD}")
   foreach(sh IN_LIST SHIELD_LIST)
```
KAGA164, line 10: There is a typo here, this worked previously by coincident. Now with quotes it does not work because of this bug.
Could you fix it also?
```suggestion
  foreach(sh IN LISTS SHIELD_LIST)
```

samples/peripheral/radio_test/CMakeLists.txt
```diff
@@ -6,11 +6,11 @@
 
 cmake_minimum_required(VERSION 3.20.0)
 
-if (DEFINED SHIELD AND ${BOARD} STREQUAL "nrf5340dk_nrf5340_cpunet")
+if(DEFINED SHIELD AND "${BOARD}" STREQUAL "nrf5340dk_nrf5340_cpunet")
   string(REPLACE " " ";" SHIELD_LIST "${SHIELD}")
   foreach(sh IN_LIST SHIELD_LIST)
```
KAGA164, line 11: There is a typo here, this worked previously by coincident. Now with quotes it does not work because of this bug.
Could you fix it also?
```suggestion
  foreach(sh IN LISTS SHIELD_LIST)
```

KAGA164 (approved):
Looks good. It event fix some CI toolchain issues

joerchan (approved):

frkv:
will the IN_LIST => IN LISTS issue be resolved? it looks like an actual error

samples/bluetooth/direct_test_mode/CMakeLists.txt
nordicjm: `IN_LIST` is correct, as per https://cmake.org/cmake/help/v3.3/command/if.html

nordicjm:
> will the IN_LIST => IN LISTS issue be resolved? it looks like an actual error

`IN_LIST` is the correct command, `IN_LISTS` does not exist, as per https://cmake.org/cmake/help/v3.3/command/if.html

KAGA164:
@nordicjm @frkv 
> > will the IN_LIST => IN LISTS issue be resolved? it looks like an actual error
> 
> `IN_LIST` is the correct command, `IN_LISTS` does not exist, as per https://cmake.org/cmake/help/v3.3/command/if.html

We have `foreach` statement here not `if`. https://cmake.org/cmake/help/v3.3/command/foreach.html

samples/bluetooth/direct_test_mode/CMakeLists.txt
nordicjm: Fixed

samples/peripheral/radio_test/CMakeLists.txt
nordicjm: Fixed

frkv:
> > will the IN_LIST => IN LISTS issue be resolved? it looks like an actual error
> 
> `IN_LIST` is the correct command, `IN_LISTS` does not exist, as per https://cmake.org/cmake/help/v3.3/command/if.html

`IN LISTS` is not with an underscore, and this is not an if-statement.

frkv (approved):
LGTM

carlescufi (approved):

PR #11619, opened 2023-06-26
doc: thread: SSED and device type updates

greg-fer:
Updated SED vs SSED docs with clarifications about use cases. Updated Thread device type page to fix inaccuracies. KRKNWK-16099 and PR #11410.

### DIFF
--- a/doc/nrf/protocols/thread/device_types.rst
+++ b/doc/nrf/protocols/thread/device_types.rst
@@ -8,15 +8,38 @@ Thread device types
    :depth: 2
 
 Thread devices can be either Full Thread Devices (FTD) or Minimal Thread Devices (MTD).
-The devices are divided based on their ability to exist in a network without a parent, determined by whether the device maintains a routing table (doesn't need a parent) or not (needs a parent).
-FTDs are the devices that maintain a routing table, while MTDs do not, forwarding all their messages to a parent instead.
+Generally speaking, the devices are classified based on the following criteria:
 
-Another type of Thread device classification is based on the device's ability to route messages:
+* What role they have in a Thread network - whether they are router-capable (Router, Leader) or can be just an End Device.
+* What is the communication scheme - whether the device keeps its radio enabled all the time or sporadically (allowing the device to sleep and save power).
+* How they route messages - whether it maintains a routing table (can route messages) or not (does not route messages).
+* Which Thread management functionalities they implement - whether they perform them by itself (for example, address resolution or address registration) or rely on the parent.
+
+The role and routing criteria can be used to classify the Thread devices as follows:
 
 * Routers can route messages, and they can function as parents to other Thread devices.
 * End Devices cannot route messages, and they cannot function as parents.
 
-Based on this classification, FTDs can be either Routers or End Devices, while MTDs can only be End Devices.
+Based on all these criteria, the Thread devices can be classified as follows:
+
+.. list-table:: Thread device classification
+   :header-rows: 1
+
+   * -
+     - Role
+     - Communication scheme
+     - Message routing method
+     - Thread management functionalities
+   * - Full Thread Devices (FTDs)
+     - Routers (Leaders) or End Devices
+     - Keep their radios enabled all the time.
+     - Maintain a routing table.
+     - Routers perform them by itself. End Devices classified as FTDs have some management functionalities.
+   * - Minimal Thread Devices (MTDs)
+     - End Devices
+     - May not have their radio enabled at all times.
+     - Do not maintain a routing table and forward all their messages to the parent.
+     - Rely on the parent to perform them.
 
 For additional information about Thread device types, see the documentation for `Device Types on OpenThread portal`_.
 
@@ -25,48 +48,50 @@ For additional information about Thread device types, see the documentation for 
 Full Thread Devices
 *******************
 
-In order to maintain the routing table, FTDs keep their radio on at all times.
-An important difference between FTDs and MTDs is that FTDs subscribe to special "all-routers" multicast addresses (``ff02::2`` and ``ff03::2``, see `IPv6 multicast addressing`_).
-Because of these reasons, they typically consume more power than MTDs.
+FTDs keep their radio enabled at all times to be ready to receive messages asynchronously.
+They typically consume more power than MTDs and are mains-powered, which allows them to have the following functionalities that MTDs do not have, among others:
 
-Full Thread Devices can be further divided into three categories:
+* Ability to resolve IPv6 addresses
+* Handling address registration
+* Maintaining routing table
+* Maintaining links with neighbors
+
+Another important difference between FTDs and MTDs is that FTDs subscribe to special "all-routers" multicast addresses (``ff02::2`` and ``ff03::2``, see `IPv6 multicast addressing`_).
+
+Full Thread Devices can be further divided into the following roles, which depend on the topology of the network:
 
 * Router
-* Full End Device (FED)
-* Router Eligible End Device (REED)
+* Full End Device (FED) or Router Eligible End Device (REED)
 
-.. list-table:: Full Thread Device categories
-   :widths: 15 10 10 15
+.. list-table:: Full Thread Device network role categories
    :header-rows: 1
 
    * -
      - Router
-     - Full End Device
-     - Router Eligible End Device
+     - Full End Device (FED) or Router Eligible End Device (REED)
    * - Maintains a routing table
-     - Yes
      - Yes
      - Yes
    * - Routes messages
      - Yes
      - No
-     - When working as a Router
    * - Can extend the network
      - Yes
      - No
-     - When working as a Router
 
 Routers can be regarded as the backbone of a Thread network.
 They maintain the routing table and forward messages to other devices.
 Routers, unlike End Devices, can also be used to extend the network range and a Router is required for a Minimal End Device to join the network.
-The maximum number of Routers in a single Thread network is 32.
+The maximum number of active Routers in a single Thread network is 32.
 
-FEDs maintain a routing table, but they can't route messages.
-This means they don't require a parent (Router) to function, but they cannot become parents themselves.
+FEDs are End Devices that maintain a routing table, but they can't route messages.
+FEDs are End Devices with FTDs capabilities to perform certain Thread management functionalities on behalf of their parent (Router), such as address resolution, effectively reducing the parent's computation power.
+They are child devices that require a parent to function, but they cannot become parents themselves.
 
-REEDs maintain a routing table and, unlike FEDs, they can become Routers if needed.
-If a REED is the only device in range of an End Device trying to join the network, it will promote itself to a Router (see `Joining an existing Thread network`_ in the OpenThread documentation for more information).
-Conversely, when a Router has no children it can downgrade itself to a REED.
+REEDs are End Devices that maintain a routing table and can be promoted to be Routers if needed.
+If a REED is the only device in range of an End Device trying to join the network, it will proactively request promotion to the Router role (see `Joining an existing Thread network`_ in the OpenThread documentation for more information).
+
+Depending on the network topology conditions, a Router may downgrade itself to a REED role, for example when it does not have any children and there is already a sufficient number of Routers in the network.
 
 .. _thread_types_mtd:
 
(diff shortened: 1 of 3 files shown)


doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -55,7 +55,15 @@ Comparison of SED and SSED
 **************************
 
 Compared to an SED, an SSED has no drawbacks for transmission and provides reduced power consumption and network traffic.
-This means you should configure your SED devices as SSEDs whenever possible.
+
+Based on the SED and SSED characteristics, use SSED for cases that meet the following criteria:
+
+* You need a small latency (small CSL interval).
+  As a rule of thumb, use SSED when the application requires to poll more often than every 20 seconds, but make sure to make detailed measurements for your use case.
+* You want to have similar or smaller power consumption than with SED.
```
pjasinski990, line 63: From what I understand, SSEDs are viable to use only when the added code and complexity are justified by the lower power usage. Frequent polling (eg. at least every 20s as proposed by Eduardo) makes the device a candidate to be a viable SSED, given the device and parent uncertainty (among other parameters) is good enough - so the decision whether to use SSED still requires some application-specific data.
We could write something along those lines - the more frequent polling, the better candidate the device is to be SSED, as long as other parameters don't make the transmission window's size abysmal.
The important thing here is also that parent devices from other vendors might have big uncertainty, which will result in a big transmission window regardless of how low the uncertainty is on the child (SSED).

greg-fer: @LuDuda , any opinions about Piotr's proposal?

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -8,15 +8,19 @@ Thread device types
...
 Thread devices can be either Full Thread Devices (FTD) or Minimal Thread Devices (MTD).
-The devices are divided based on their ability to exist in a network without a parent, determined by whether the device maintains a routing table (doesn't need a parent) or not (needs a parent).
-FTDs are the devices that maintain a routing table, while MTDs do not, forwarding all their messages to a parent instead.
+The devices are divided based on the following criteria:
 
-Another type of Thread device classification is based on the device's ability to route messages:
+* How they exist in a Thread network - whether the device keeps its radio enabled all the time.
+* How they route messages - whether it maintains a routing table (can route messages) or not (does not route messages).
+
+FTDs are the devices that keep their radios enabled all the time and maintain a routing table, while MTDs do not have their radio enabled at all times and they do not maintain a routing table, consequently forwarding all their messages to a parent instead.
```
LuDuda, line 16:
```suggestion
FTDs are the devices that keep their radios enabled all the time and maintain a routing table, while MTDs may have their radio enabled at all times and they do not maintain a routing table, consequently forwarding all their messages to a parent instead.
```

MED is one of MTD role which keeps its receiver on all the time - so "may" is more propriate.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -8,15 +8,19 @@ Thread device types
    :depth: 2
 
 Thread devices can be either Full Thread Devices (FTD) or Minimal Thread Devices (MTD).
-The devices are divided based on their ability to exist in a network without a parent, determined by whether the device maintains a routing table (doesn't need a parent) or not (needs a parent).
-FTDs are the devices that maintain a routing table, while MTDs do not, forwarding all their messages to a parent instead.
+The devices are divided based on the following criteria:
 
-Another type of Thread device classification is based on the device's ability to route messages:
+* How they exist in a Thread network - whether the device keeps its radio enabled all the time.
+* How they route messages - whether it maintains a routing table (can route messages) or not (does not route messages).
```
LuDuda, line 14: What about this?

```suggestion
* What role they have in a Thread network - whether they are router-capable (Router, Leader) or can be just an end-device
* What is the communication scheme - whether the device keeps its radio enabled all the time or sporadically (allowing device to sleep and save power)
* How they route messages - whether it maintains a routing table (can route messages) or not (does not route messages).
* Which Thread management functionalities they implement - whether they perform for example address resolution or address registration by itself or rely on the parent
```

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -25,48 +29,50 @@ For additional information about Thread device types, see the documentation for
...
+
+* Ability to resolve IPv6 addresses
+* Handling address registration
+* Maintaining routing table
+* Maintaining links with neighbors
 
-Full Thread Devices can be further divided into three categories:
+Another important difference between FTDs and MTDs is that FTDs subscribe to special "all-routers" multicast addresses (``ff02::2`` and ``ff03::2``, see `IPv6 multicast addressing`_).
+
+Full Thread Devices can be further divided into the following categories, which depend on the topology of the network:
```
LuDuda, line 42:
```suggestion
Full Thread Devices can be further divided into the following roles, which depend on the topology of the network:
```

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -25,48 +29,50 @@ For additional information about Thread device types, see the documentation for
...
 
 Routers can be regarded as the backbone of a Thread network.
 They maintain the routing table and forward messages to other devices.
 Routers, unlike End Devices, can also be used to extend the network range and a Router is required for a Minimal End Device to join the network.
-The maximum number of Routers in a single Thread network is 32.
+The maximum number of active Routers in a single Thread network is 32.
 
-FEDs maintain a routing table, but they can't route messages.
-This means they don't require a parent (Router) to function, but they cannot become parents themselves.
+FEDs are End Devices that maintain a routing table, but they can't route messages.
```
LuDuda, line 69:
```suggestion
FEDs are End Devices with FTDs capabilities to perform certain Thread management functionalities such as address resolution. Effectively reducing it's router computation power"
```

In last sentence.. i want to say that FED relieves its partent from doing things on behalf of the child. This is the main advantage of FED.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -95,31 +101,36 @@ Minimal Thread Devices can be further divided into three categories:
...
 
-MEDs are the most basic MTDs, and their radio is always on.
-They don't keep a routing table, but otherwise operate like :ref:`FEDs <thread_types_ftd>`.
+MEDs are the most basic MTDs, and their radio is always enabled.
+They don't keep a routing table and have a :ref:`more limited range of functionalities than FEDs <thread_types_ftd>`.
 
 SEDs try to limit their power consumption by sleeping most of the time, waking up periodically to poll for messages from their parent.
 After waking up, they send a data request to their parent.
 If the parent has any pending messages, it will send them to the SED.
+(This can happen at any time, since Routers have their radio enabled all the time.)
```
LuDuda, line 123: nit: Does it make sence to have this information together with previous sentence? I mean...

"[..] to the SED (this can happen at any time..."

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -55,7 +55,15 @@ Comparison of SED and SSED
...
+
+For example, one specific market application of SSEDs could be window coverings.
```
LuDuda, line 65:
```suggestion
For example, one specific market application of SSEDs could be window coverings. For this use-case blinds need to be very responsive e.g. latency within 500ms, while data is very rarely exchanged between controller device and window covering device.
```

doc/nrf/protocols/thread/sed_ssed.rst
LuDuda: Good suggestions from Piotr. In general I'm a bit afraid to "recommend" SSED role blindly for manufacturers even if latency is less than ~20s. 

The whole Thread/Matter ecosystem is not yet fully ready. I would say that the default option is SED. However, SSED is great when it's parent has small CSL uncertainty as well accuracy. We don't have an example in SDK, but customer may use information about parent's CSL parameters (uncertainty and accuracy) to switch to SSED role or not. This can be done just after attachment dynamically.

In general what Piotr suggested is great. E.g. that SSED is viable option and may decrease power consumption in case the CSL interval is small (~<20s) as well as parent has decent CSL parameters.

LuDuda (commented):
@greg-fer this is great, thank you!

I put few last comments, but it looks nice now.

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -95,31 +119,35 @@ Minimal Thread Devices can be further divided into three categories:
...
-They wake up at the same time as their parent, eliminating the need for polling for messages.
+They wake up at designated transmission windows agreed with their their parent, which eliminates the need for polling for messages.
 If the parent has messages for the SSED, it sends them during the designated transmission window.
 The SSED allows the transmission to finish if radio activity is detected during the transmission window.
+
+.. note::
+    If the length of the message exceeds the length of the transmission window, the first frame is received in the designated transmission window, but the rest is transmitted using regular Data Polls, exactly like SED.
+
 Conversely, if there is no radio activity during the specified duration of transmission window, this indicates that the parent has no messages for the SSED and the SSED returns to sleep.
 The SSED synchronization results in lower power consumption compared to an SED, primarily because the SSED doesn't need to poll for messages, keeping transmission windows short.
```
edmont, line 152:
```suggestion
The SSED synchronization usually results in lower power consumption compared to an SED, primarily because the SSED doesn't need to poll for messages, keeping transmission windows short.
```

doc/nrf/protocols/thread/sed_ssed.rst
edmont, line 1: I believe that apart from the power consumption perspective we should also comment about the radio channel usage, i.e., for environments where many sleepy devices are expected it would be beneficial to use SSEDs to reduce over the air traffic. This is probably not an argument for the residential market, but could have some impact on more specific use cases.

greg-fer:
@pjasinski990 , @edmont , @LuDuda , please kindly check again when the doc build completes.

doc/nrf/protocols/thread/sed_ssed.rst
```diff
@@ -55,7 +55,19 @@ Comparison of SED and SSED
 **************************
 
 Compared to an SED, an SSED has no drawbacks for transmission and provides reduced power consumption and network traffic.
-This means you should configure your SED devices as SSEDs whenever possible.
+
+Based on the SED and SSED characteristics, use SSED for cases that meet one or more of the following criteria:
+
+* You need a small latency (ensured by small poll period for SEDs or small CSL interval for SSEDs).
+  As a general rule, SSED is a viable option and may decrease power consumption if the CSL interval is small (less than 20 seconds) and the parent has decent CSL parameters.
```
edmont, line 62:
```suggestion
  As a general rule, SSED is a viable option and may decrease power consumption if the CSL Period is small (less than 20 seconds) and the parent has decent CSL Accuracy.
```

doc/nrf/protocols/thread/device_types.rst
```diff
@@ -95,33 +119,37 @@ Minimal Thread Devices can be further divided into three categories:
...
+They wake up at designated transmission windows agreed with their their parent, which eliminates the need for polling for messages.
 If the parent has messages for the SSED, it sends them during the designated transmission window.
 The SSED allows the transmission to finish if radio activity is detected during the transmission window.
+
+.. note::
+    If the length of the message exceeds the length of the transmission window, the first frame is received in the designated transmission window, but the rest is transmitted using regular Data Polls, exactly like SED.
+
 Conversely, if there is no radio activity during the specified duration of transmission window, this indicates that the parent has no messages for the SSED and the SSED returns to sleep.
-The SSED synchronization results in lower power consumption compared to an SED, primarily because the SSED doesn't need to poll for messages, keeping transmission windows short.
+The SSED synchronization usually results in lower power consumption compared to an SED, primarily because the SSED doesn't need to poll for messages, keeping transmission windows short.
```
edmont, line 152:
```suggestion
The SSED synchronization results in lower power consumption compared to an SED in some scenarios, primarily because the SSED doesn't need to poll for messages, keeping transmission windows short.
```
Sorry for going over this again, but we should be clear that SSED is not better than SED in power consumption as a general rule.

pjasinski990 (approved):

greg-fer:
Applied your edits, @edmont

edmont (approved):

LuDuda (approved):

PR #11626, opened 2023-06-27
lib: at_host: Use zephyr LOG for outgoing modem responses

simensrostad:
Use zephyr LOG for outgoing modem responses.
This ensures that responses received the modem are scheduled in turn by the zephyr logging system.

Previously these logs could cut zephyr LOG output because its pushed directly on UART. This lead to tests failing because the expected output was corrupted by these modem responses.

### DIFF
--- a/lib/at_host/at_host.c
+++ b/lib/at_host/at_host.c
@@ -48,6 +48,11 @@ static struct k_work cmd_send_work;
 
 static inline void write_uart_string(const char *str)
 {
+	if (IS_ENABLED(CONFIG_LOG_BACKEND_UART)) {
+		LOG_RAW("%s", str);
+		return;
+	}
+
 	/* Send characters until, but not including, null */
 	for (size_t i = 0; str[i]; i++) {
 		uart_poll_out(uart_dev, str[i]);


lib/at_host/at_host.c
```diff
@@ -48,10 +48,7 @@ static struct k_work cmd_send_work;
 
 static inline void write_uart_string(const char *str)
 {
-	/* Send characters until, but not including, null */
-	for (size_t i = 0; str[i]; i++) {
-		uart_poll_out(uart_dev, str[i]);
-	}
+	LOG_RAW("%s", str);
```
jorgenmk, line 51: This can be moved into `response_handler` and `write_uart_string` can be deleted?

simensrostad: `write_uart_string` is also called in `cmd_send`

simensrostad:
Updated the fix to push the modem response to UART if RTT is used as the logging backend. This is to ensure that the AT host will still be compatible with external tools such as the cellular monitor, even if RTT is selected as the logging backend.

jtguggedal (approved):
I think this should do the trick 👍

lib/at_host/at_host.c
```diff
@@ -48,6 +48,11 @@ static struct k_work cmd_send_work;
 
 static inline void write_uart_string(const char *str)
 {
+	if (IS_ENABLED(CONFIG_LOG_BACKEND_UART) {
```
jorgenmk, line 51: Missing closing bracket. Otherwise approved

jorgenmk (approved):

balaji-nordic (approved):

bjda (approved):

PR #11627, opened 2023-06-27
samples: bluetooth: peripheral_uart: Increase the main stack size

maje-emb:
This commit increases the size of the main stack for prj_minimal. 
This fixes a stack overflow on boot.

Ref: NCSDK-21709


grochu (approved):
Looks good.

samples/bluetooth/peripheral_uart/prj_minimal.conf
```diff
@@ -94,7 +94,7 @@ CONFIG_BT_HCI_TX_STACK_SIZE_WITH_PROMPT=y
 CONFIG_BT_HCI_TX_STACK_SIZE=640
 CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=512
 CONFIG_MPSL_WORK_STACK_SIZE=256
-CONFIG_MAIN_STACK_SIZE=640
+CONFIG_MAIN_STACK_SIZE=860
```
dchat-nordic, line 97: I would use 864, because the AAPCS requires double word alignment. So with 860 the 4 bytes is wasted or misused.

dchat-nordic (changes requested):

samples/bluetooth/peripheral_uart/prj_minimal.conf
maje-emb: Good point.

PR #11628, opened 2023-06-27
Bluetooth: Mesh: Def sysworkq stack size samples

Andrewpini:
To avoid Kconfig dependencies changing system workq stack size for different platforms, explicitly define this stack size for all mesh samples.


PavelVPV (approved):
I assume samples system tests are passing?

alxelax (approved):

ludvigsj (approved):

PR #11635, opened 2023-06-28
applications: nrf5340_audio: Disconnect during MCS discovery as WRN

andvib:
Disconnects during MCS discovery will return BT_ATT_ERR_UNLIKELY which will be printed as error. As this is a normal use case, message will be printed as warning instead.


erikrobstad (approved):
The compliance warning must be fixed :)

alexsven (approved):

koffes (approved):

PR #11637, opened 2023-06-28
"Won't fix" known issues page filter

gautesl:
Added two new directives to add a filter dropdown to a page. All elements that should be able to be filtered away must contain the "hideable" html class, as well as some class that the filter directive can recognize.

`page-filter` is the main directive which adds generic filtering. When using this directive, all filterable html classes must be given in its body as a tuple (class, displayname). The directive accepts two options:
`name`: Used as a html identifier and to display the "All x" option of the dropdown.
`default`: The default value of the dropdown. Defaults to "all".

`version-filter` is a subclass of the `page-filter` directive which automatically populates the dropdown with all release versions. This directive also adds the hideable class to all elements that already contain the vX-X-X style classes, and creates clickable version tags in the desired container when specifying the `container` option.

Examples of use:
```rst
.. page-filter::
  :name: components

  ble_controller  BLE Controller
  nrfcloud   nRFCloud
  tf-m       TF-M
  ble_mesh   BLE MESH

.. version-filter::
  :default: v2-4-0
  :container: dl/dt
```

This PR also replaces the existing filter function on the known issues page with the new `version-filter` directive, and adds a new "Won't fix" filter.

image

IMPORTANT: These issues are randomly selected for previewing purposes and do NOT reflect actual unresolvable issues!
No wontfix labels are added in this PR, and must be added at a later stage. This means that filtering for "Won't fix" issues currently will produce an empty page.


greg-fer:
@gautesl , this looks very nice, and thank you also for simplifying the handling of the long version list.

I have two suggestions:
- Can we have a "Won't fix" tag in red added to the list of version tags whenever we add a `wontfix`?

And a couple of questions:
- What does `dl/dt` stand for?
- Do I understand correctly that `version-filter` can be set to some other value than the latest released version, for example 2.3.0 instead of 2.4.0, yet 2.4.0 will still be available in the list?

@umapraseeda , this needs coordination with @avalei with regard to Zoomin.

gautesl:
> @gautesl , this looks very nice, and thank you also for simplifying the handling of the long version list.
> 
> I have a suggestion:
> 
> * Can we have a "Won't fix" tag in red added to the list of version tags whenever we add a `wontfix`?
> 
> And a couple of questions:
> 
> * What does `dl/dt` stand for?
> * Do I understand correctly that `version-filter` can be set to some other value than the latest released version, for example 2.3.0 instead of 2.4.0, yet 2.4.0 will still be available in the list?
> 
> @umapraseeda , this needs coordination with @avalei with regard to Zoomin.

Thank you for suggestions!
I am currently working on a `wontfix` tag, so that will come.

- `dl/dt` is the html node where the labels will be populated into (so that it matches the current system).
- The default version can be any version available from [versions.json](https://github.com/nrfconnect/sdk-nrf/blob/main/doc/versions.json).

greg-fer:
> A correct distribution of the wontfix label is a prerequisite for removing the DNM label and upgrading the PR from a draft PR.

@gautesl , can we have the `wontfix` in `page-filter/components` but without populating the tag to the issues? I'm afraid that if we decide to populate known issues with the `wontfix` tag in this PR, it will dilute the purpose of this PR and _significantly_ lengthen its review. I think adding the tag could be handled in the known issues review before each release.

gautesl:
Added a visible "Won't fix" tag to to the versiontag list. The tags are created with a new option to the `:page-filter:` directive, i.e. `:page-filter: [("wontfix", "Won't fix")]` will look for html elements with the "wontfix" class and insert tags containing the text "Won't fix".

@greg-fer After the new tag has been deemed satisfactory and the exact wording (now Won't fix - Fixable) has been selected, we can remove the test tags and only insert the dropdown component. It will be a "useless" filter until issues are populated with wontfix-tags (or what they will be called), but that can come later.

doc/nrf/releases_and_maturity/known_issues.rst
```diff
@@ -21,65 +21,27 @@ A known issue can list one or both of the following entries:
...
+.. version-filter::
+  :default: v2-4-2
+  :container: dl/dt
+  :tags: [("wontfix", "Won't fix")]
+
+.. page-filter::
+  :name: issues
+
+  wontfix    Won't fix
+  !wontfix   Fixable
```
umapraseeda, line 33: @wbober, could you confirm if this wording is good for the filter text? Do you have any other suggestions?

greg-fer: Do we need `Fixable`? AFAIK, all known issues are fixable until tagged with `Won't Fix`, so why do we need a tag for `!wontfix`? It will only confuse the readers, who will now face issues that are fixable, issues that won't be fixed, and issues that have neither of these tags... Not to mention that `Fixable` could change to `Won't Fix` and we would have to explain why this change happened.

gautesl: @greg-fer There is no visible "Fixable" tag, this is only for the dropdown showing the filter options

wbober: > Do we need `Fixable`? AFAIK, all known issues are fixable until tagged with `Won't Fix`, so why do we need a tag for `!wontfix`? It will only confuse the readers, who will now face issues that are fixable, issues that won't be fixed, and issues that have neither of these tags... Not to mention that `Fixable` could change to `Won't Fix` and we would have to explain why this change happened.

I'm with @greg-fer on that. I'd just settle on "All" and "Won't fix" filters.

gautesl: "Fixable" has now been removed as a filter option, and the all wontfix-tags has been removed (a screenshot of how it looks can be found in the PR description instead)

gautesl:
Removed all wontfix-tags used for testing and the "Fixable" filtering option. Filtering on "Won't fix" will now yield an empty page. Removed the DNM tag since no false tags are present.

greg-fer (approved):

greg-fer:
Thank you so much @gautesl for implementing this!

doc/_extensions/static/page_filter.js
```diff
@@ -0,0 +1,171 @@
...
+    .filter(val => val !== "all");
+
+  if (filters.length == 0) return;
+
+  var negativeFilters = filters
+    .filter(val => val.charAt(0) === "!")
+    .map(val => val.substring(1));
+  filters = filters.filter(val => val.charAt(0) !== "!")
+
+  let selector = "";
```
gmarull, line 24: nit: this file should likely make consistent usage of var/let

umapraseeda (approved):

doc/_extensions/static/page_filter.js
```diff
@@ -0,0 +1,171 @@
+const all_dropdowns = []
```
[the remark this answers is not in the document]
gautesl: This was already treated as a module except for the name, fixed now!

doc/_extensions/static/page_filter.js
gautesl: Fixed, thanks!

gmarull (approved):

PR #11639, opened 2023-06-28
manifest: sdk-zephyr: Pull CoAP library update

tokangas:
Use Zephyr internal zsock_ calls to remove dependency to NET_SOCKETS_POSIX_NAMES.


tokangas:
Closing, manifest update included in #11654 which needs this change.

PR #11640, opened 2023-06-28
ci: docbuild: some zoomin updates

gmarull:


gmarull:
@Avalei can you add the `ncs-latest` _tag_ to the subjectscheme?

Avalei (approved):
Subject scheme is now updated with `ncs-latest`.

PR #11642, opened 2023-06-28
applications: nrf5340_audio: Expand SD card module

MariusVaardal:
To avoid unnecessary opening
and closing of files


applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -30,12 +30,13 @@ static struct fs_mount_t mnt_pt = {
...
 };
 
-int sd_card_list_files(char *path)
+int sd_card_list_files(char *path, char *buf, size_t buf_size)
 {
 	int ret;
 	struct fs_dir_t dirp;
 	static struct fs_dirent entry;
 	char abs_path_name[PATH_MAX_LEN + 1] = SD_ROOT_PATH;
+	bool write_to_buffer = (buf != NULL);
```
koffes, line 39: No need to create a temporary variable here. Just use `buf != NULL` on line 75.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -71,12 +72,19 @@ int sd_card_list_files(char *path)
 		if (entry.name[0] == 0) {
 			break;
 		}
-
-		if (entry.type == FS_DIR_ENTRY_DIR) {
-			LOG_INF("[DIR ] %s", entry.name);
-		} else {
-			LOG_INF("[FILE] %s", entry.name);
+		if (write_to_buffer) {
+			size_t used_buf_size = strlen(buf);
```
koffes, line 76: NB strlen will only work if the buf is` \0` terminated.

koffes (changes requested):

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -222,3 +230,72 @@ int sd_card_init(void)
...
 }
+
+static volatile bool seg_read_started = false;
+static struct fs_file_t f_seg_read_entry;
+
+int sd_card_segment_read_open(char const *const filename, const char *path_to_file)
+{
+	char abs_path_name[PATH_MAX_LEN + 1] = SD_ROOT_PATH;
+	strcat(abs_path_name, path_to_file);
+	printk("abs path name:\t%s\n", abs_path_name);
```
koffes, line 241: printk is used a lot in Zephyr, but we use LOG_DBG, LOG_INF, LOG_WRN and LOG_ERR

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -222,3 +230,72 @@ int sd_card_init(void)
...
+	int ret;
```
koffes, line 242: `int ret` is usually declared at the top of any function.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -222,3 +230,72 @@ int sd_card_init(void)
...
+
+	if (!sd_init_success) {
+		return -ENODEV;
+	}
+
+	if (strlen(filename) > CONFIG_FS_FATFS_MAX_LFN) {
+		LOG_ERR("Filename is too long");
+		return -FR_INVALID_NAME;
```
koffes, line 250: Here we may need to discuss. The `FR_INVALID_NAME` you have chosen is a good decision, but this collides with the standard error defines.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -222,3 +230,72 @@ int sd_card_init(void)
...
+		LOG_ERR("Read file failed");
+		return ret;
+	}
+
+	*size = ret;
+
+	return 0;
+}
+
+int sd_card_segment_read_close(void)
```
koffes, line 286: Some renaming required. There is `sd_card_segment_read` and `sd_card_segment_read_close`. One is a subset of the other. 
I see your logic, but xxx_read_close is to verbs at the same time which can be confusing. 
To be discussed.

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -60,4 +62,32 @@ int sd_card_read(char const *const filename, char *const data, size_t *size);
...
 int sd_card_init(void);
 
+/**@brief   Open file on SD card
+ * param[in] filename   Name of file to open
+ * param[in] path_to_file Path to file
+ * @retval  0 on success
+ *              -ENODEV SD init faile. SD likely not inserted
+ *      Otherwise, error from undelying drivers
+ */
+int sd_card_segment_read_open(char const *const filename, const char *path_to_file);
```
koffes, line 72: I would not expect either the filename nor the path_to_file to change, so:
```suggestion
int sd_card_segment_read_open(char const *const filename, char const * const path_to_file);
```

applications/nrf5340_audio/src/modules/sd_card.c
koffes: Why is the word `segment` used? I thought this was reading a single file?

koffes: I think I see after going through the entire PR. If the file is too large, you read out a part/segment of the given file?

koffes (changes requested):

applications/nrf5340_audio/src/modules/sd_card.c
alexsven: And this should probably be a LOG_DBG

applications/nrf5340_audio/src/modules/sd_card.c
alexsven: There is one called ENAMETOOLONG that can be used

applications/nrf5340_audio/src/modules/sd_card.c
alexsven: Suggestion: sd_card_segment_open(), sd_card_segment_read(), sd_card_segment_close()

alexsven (changes requested):

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -24,13 +24,15 @@ LOG_MODULE_REGISTER(sd_card, CONFIG_MODULE_SD_CARD_LOG_LEVEL);
 static const char *sd_root_path = "/SD:";
 static FATFS fat_fs;
 static bool sd_init_success;
+static volatile bool seg_read_started;
```
alexsven, line 27:
```suggestion
static volatile bool segment_opened;
```

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,126 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+{
+	int ret;
+
+	if (!seg_read_started) {
+		return -EBUSY;
+	}
+
+	ret = fs_close(&f_seg_read_entry);
+	if (ret) {
+		LOG_ERR("Close file failed");
```
alexsven, line 304: Please add return value in the prints:
```suggestion
		LOG_ERR("Close file failed: %d", ret);
```

Goes for all the error prints in this file

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -9,17 +9,19 @@
...
 #include <stddef.h>
 
-/**@brief Print out the contents under SD card root path
+/**@brief Print out the contents under SD card root path and write the content to buffer.
  *
  * @param[in] path	Path of the folder which going to list
  *			If assigned path is null, then listing the contents under root
  *			If assigned path doesn't exist, an error will be returned
- *
+ * @param[in, out] buf Buffer where data is written. If set to NULL, it will be ignored.
```
alexsven, line 17: This is only an [out] as far as I can see

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -52,6 +54,51 @@ int sd_card_write(char const *const filename, char const *const data, size_t *si
  */
 int sd_card_read(char const *const filename, char *const data, size_t *size);
 
+/**@brief   Open file on SD card
+ * param[in] filename   Name of file to open
+ * param[in] path_to_file Path to file
```
alexsven, line 59:
```suggestion
 * @param[in] filename   Name of file to open
 * @param[in] path_to_file Path to file
```

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -52,6 +54,51 @@ int sd_card_write(char const *const filename, char const *const data, size_t *si
...
+ * @retval  0 on success
+ *              -ENODEV SD init failed. SD likely not inserted
+ *      Otherwise, error from underlying drivers
+ */
+int sd_card_segment_open(char const *const filename, char const *const path_to_file);
+
+/**@brief   Read segment on the open file on the SD card
+ * param[in] data   Where the read data is stored
```
alexsven, line 67:
```suggestion
 * @param[in] data   Where the read data is stored
```

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -52,6 +54,51 @@ int sd_card_write(char const *const filename, char const *const data, size_t *si
...
+/**@brief   Close the file opened by the sd_card_segment_read_open function
+ * @retval  0 on success
+ *              -EBUSY Segment read operation has not started
+ *      Otherwise, error from underlying drivers
+ */
+int sd_card_segment_close(void);
+
+/**@brief   Peek at data ahead of current file position without changing file position
+ * param[in, out] data Location to where the read data should be stored
+ * param[in, out] size Size of data to be read
```
alexsven, line 87:
```suggestion
 * @param[in, out] data Location to where the read data should be stored
 * @param[in, out] size Size of data to be read
```

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -52,6 +54,51 @@ int sd_card_write(char const *const filename, char const *const data, size_t *si
...
+ * @retval  0 on success
+ *              -EBUSY Segment read operation has not started
+ *		Otherwise, error from underlying drivers
+ */
+int sd_card_segment_peek(char *const data, size_t *size);
+
+/**@brief   Jump over segment from current file position
+ * param[in] size   Size of segment to skip
```
alexsven, line 95:
```suggestion
 * @param[in] size   Size of segment to skip
```

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -52,6 +54,51 @@ int sd_card_write(char const *const filename, char const *const data, size_t *si
...
+ * param[in] path_to_file Path to file
+ * @retval  0 on success
+ *              -ENODEV SD init failed. SD likely not inserted
+ *      Otherwise, error from underlying drivers
+ */
+int sd_card_segment_open(char const *const filename, char const *const path_to_file);
+
+/**@brief   Read segment on the open file on the SD card
+ * param[in] data   Where the read data is stored
+ * @param[in,out] size	Number of bytes to be read from file
```
alexsven, line 68:
```suggestion
 * @param[in, out] size	Number of bytes to be read from file
```

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,126 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	if (ret < 0) {
+		LOG_ERR("Read file failed");
+		return ret;
+	}
+	ret = fs_seek(&f_seg_read_entry, offset, 0);
+	if (ret < 0) {
+		LOG_ERR("Fs seek failed");
+		return ret;
+	}
+	*size = ret;
```
alexsven, line 273: As far as I see fs_seek sets ret to 0 if the operation was successful, so why change size to 0 here?

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,126 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+
+	return 0;
+}
+
+int sd_card_segment_skip(const size_t *size)
```
alexsven, line 278: Why send the pointer to size here? It is just dereferenced when using fs_seek and it should probably be called offset and not size

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,126 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+		LOG_ERR("Read file failed");
+		return ret;
+	}
+
+	*size = ret;
+
+	return 0;
+}
+
+int sd_card_segment_peek(char *const data, size_t *size)
```
alexsven, line 248: Doesn't seem that size needs to be a pointer here

applications/nrf5340_audio/src/modules/sd_card.h
alexsven: How is data an out? Is the pointer changed? You can probably just remove all the [in] and [in, out] from this file to keep it simpler

alexsven (changes requested):

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,125 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+int sd_card_segment_skip(const size_t size)
+{
+	int ret;
+
+	if (!segment_opened) {
+		return -EBUSY;
+	}
+	ret = fs_seek(&f_seg_read_entry, size, FS_SEEK_CUR);
+	if (ret < 0) {
+		LOG_ERR("Fs seek failed. Return value: %d\n", ret);
```
alexsven, line 286: No need for \n when using LOG_XXX

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -52,6 +54,51 @@ int sd_card_write(char const *const filename, char const *const data, size_t *si
...
+ *			indicates that the file is empty
+ * @retval  0 on success
+ *              -ENODEV SD init failed. SD likely not inserted
+ *      Otherwise, error from underlying drivers
+ */
+int sd_card_segment_read(char *const data, size_t *size);
+
+/**@brief   Peek at data ahead of current file position without changing file position
+ * @param[in, out] data Location to where the read data should be stored
+ * @param[in, out] size Size of data to be read
```
alexsven, line 80: Size is not an out

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,125 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	if (!segment_opened) {
+		return -EBUSY;
+	}
+
+	offset = fs_tell(&f_seg_read_entry);
+	if (offset < 0) {
+		LOG_ERR("Fs tell failed");
+		return offset;
+	}
+	ret = fs_read(&f_seg_read_entry, data, size);
```
alexsven, line 263: ret here is how many bytes you have read, what if ret doesn't match the size that was requested?

applications/nrf5340_audio/src/modules/sd_card.h
alexsven: I still suggest removing all the [in] and [in, out] from this file.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,125 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	if (ret < 0) {
+		LOG_ERR("Read file failed");
+		return ret;
+	}
+	ret = fs_seek(&f_seg_read_entry, offset, 0);
+	if (ret < 0) {
```
alexsven, line 269:
```suggestion
	}
	
	ret = fs_seek(&f_seg_read_entry, offset, 0);
	if (ret < 0) {
```

applications/nrf5340_audio/src/modules/sd_card.c
alexsven, line 263:
```suggestion
	}
	
	ret = fs_read(&f_seg_read_entry, data, size);
```

applications/nrf5340_audio/src/modules/sd_card.c
alexsven: This can be done multiple places in this file. Newline after `}`, unless the next line starts with a `}` itself

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,125 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
 	return 0;
 }
 
+int sd_card_segment_open(char const *const filename, char const *const path_to_file)
+{
+	int ret;
+
```
alexsven, line 196: Should probably check if segment is already opened here

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,125 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	*size = ret;
+
+	return 0;
+}
+
+int sd_card_segment_peek(char *const data, size_t size)
+{
+	int ret;
+
```
alexsven, line 251: Remove newline here

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,125 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	return 0;
+}
+
+int sd_card_segment_close(void)
+{
+	int ret;
+
+	if (!segment_opened) {
+		return -EBUSY;
+	}
```
alexsven, line 299:
```suggestion
                LOG_WRN("Segment is already closed");
		return -EALREADY;
	}
```

alexsven (changes requested):

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -24,19 +24,25 @@ LOG_MODULE_REGISTER(sd_card, CONFIG_MODULE_SD_CARD_LOG_LEVEL);
 static const char *sd_root_path = "/SD:";
 static FATFS fat_fs;
 static bool sd_init_success;
+static volatile bool seg_read_started;
```
koffes, line 27: Why is this volatile?

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -62,6 +68,8 @@ int sd_card_list_files(char *path)
 			return ret;
 		}
 	}
+
+	size_t used_buf_size = 0;
```
koffes, line 72: Declare this at the beginning of the function.

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -9,26 +9,28 @@
...
- *
+ * @param buf Buffer where data is written. If set to NULL, it will be ignored.
+ * @param buf_size Buffer size
  * @return	0 on success.
  *              -ENODEV SD init failed. SD card likely not inserted
+ *              -EINVAL Failed to append to buffer
  *		Otherwise, error from underlying drivers
  */
-int sd_card_list_files(char *path);
+int sd_card_list_files(char *path, char *buf, size_t buf_size);
```
koffes, line 24: How does the calling function know how much has been written to the buffer?
Suggest having a in + out pointer here which [in]: The size of the buffer, and then on [out] the number of bytes written to the buffer.

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -9,26 +9,28 @@
...
 
 /**@brief Write data from buffer into the file
  *
  * @note If the file already exists, data will be appended to the end of the file.
  *
- * @param[in] filename	Name of the target file for writing, the default location is the
+ * @param filename	Name of the target file for writing, the default location is the
```
koffes, line 30: Why has the [in] and [out] been removed? These are mandatory.

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -9,26 +9,28 @@
...
  *
- * @param[in] path	Path of the folder which going to list
+ * @param path	Path of the folder which going to list
  *			If assigned path is null, then listing the contents under root
  *			If assigned path doesn't exist, an error will be returned
- *
+ * @param buf Buffer where data is written. If set to NULL, it will be ignored.
+ * @param buf_size Buffer size
  * @return	0 on success.
  *              -ENODEV SD init failed. SD card likely not inserted
```
koffes, line 20: Missing documentation for added return codes.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -90,17 +105,17 @@ int sd_card_list_files(char *path)
...
-	int ret;
 
 	if (!sd_init_success) {
 		return -ENODEV;
 	}
 
 	if (strlen(filename) > CONFIG_FS_FATFS_MAX_LFN) {
 		LOG_ERR("Filename is too long");
-		return -FR_INVALID_NAME;
+		return -ENAMETOOLONG;
```
koffes, line 118: This looks like a non-standard error code. Where is this defined?

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +195,136 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
 }
 
+int sd_card_segment_open(char const *const filename, char const *const path_to_file)
+{
+	int ret;
+
+	if (seg_read_started) {
+		LOG_ERR("Segment read has already started");
+		/* Which value should I return here? */
+		return -1;
```
koffes, line 205: Use a defined error code to return.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +195,136 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	}
+
+	if (strlen(path_to_file) + strlen(abs_path_name) > PATH_MAX_LEN) {
```
koffes, line 208: For clarity:
```suggestion
	if ((strlen(path_to_file) + strlen(abs_path_name)) > PATH_MAX_LEN) {
```

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +195,136 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+		LOG_ERR("Filepath is too long");
+		return -EINVAL;
+	}
+
+	char abs_path_name[PATH_MAX_LEN + 1] = SD_ROOT_PATH;
+	strcat(abs_path_name, path_to_file);
+	LOG_INF("abs path name:\t%s", abs_path_name);
+
+	if (!sd_init_success) {
```
koffes, line 217: This should likely be the first check in this function. If this is wrong, nothing will work.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +195,136 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+
+	*size = ret;
+
+	return 0;
+}
+
+int sd_card_segment_peek(char *const data, size_t *size)
+{
+	int ret;
+	off_t offset;
```
koffes, line 262: What is off_t?

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +195,136 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	strcat(abs_path_name, filename);
+	fs_file_t_init(&f_seg_read_entry);
+
+	ret = fs_open(&f_seg_read_entry, abs_path_name, FS_O_READ);
+	if (ret) {
+		LOG_ERR("Open file failed");
+		return ret;
+	}
+
+	seg_read_started = true;
```
koffes, line 235: We need to discuss here. We cannot have a seg_read_started here. What if another thread wants to open another segment? This approach does not make this thread safe.

koffes (changes requested):

applications/nrf5340_audio/src/modules/sd_card.h
MariusVaardal: https://github.com/nrfconnect/sdk-nrf/pull/11642#discussion_r1270267041

applications/nrf5340_audio/src/modules/sd_card.c
koffes: Will close this after our latest changes have been merged.

rick1082 (approved):

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,54 +8,98 @@
...
- * @return	0 on success.
- *              -ENODEV SD init failed. SD card likely not inserted
+ * @param[in]		filename	Name of the target file for reading, the default location is
+ *					the root directoy of SD card, accept absolute path under
+ *					root of SD card
+ * @param[out]		data		The buffer which will be filled by read file contents
+ * @param[in, out]	size		Pointer to the number of bytes which wait to be read from
+ *					the file The actual read size will be returned If the actual
+ *					read size is 0, there will be a warning message which
+					indicates that the file is empty
```
alexsven, line 58: This section clearly needs periods between the sentences ;)

alexsven (changes requested):

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -71,12 +73,19 @@ int sd_card_list_files(char *path)
...
-		} else {
-			LOG_INF("[FILE] %s", entry.name);
+		if (buf != NULL) {
+			size_t remaining_buf_size = *buf_size - used_buf_size;
+			ssize_t len = snprintk(
+				&buf[used_buf_size], remaining_buf_size, "[%s]\t%s\n",
+				entry.type == FS_DIR_ENTRY_DIR ? "DIR " : "FILE", entry.name);
+			used_buf_size += len;
+
+			if ((len < 0) || (len >= remaining_buf_size)) {
```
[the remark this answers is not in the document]
MariusVaardal: It returns the number of bytes that would have been written to the buffer if the buffer was sufficiently large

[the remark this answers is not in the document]
MariusVaardal: As far as I understand, it does not return error codes

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,69 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	}
+
+	return 0;
+}
+
+int sd_card_read(char *const data, size_t *size, struct fs_file_t *f_seg_read_entry)
+{
+	int ret;
+
+	ret = fs_read(f_seg_read_entry, data, *size);
```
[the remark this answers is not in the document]
MariusVaardal: The f_seg_read_entry pointer must be initialized to avoid getting an error code inside of fs_read.

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,99 @@
...
- *		Otherwise, error from underlying drivers
+ * @retval	0 on success.
+ * @retval	-EPERM SD card read is ongoing somewhere else.
+ * @retval	-ENODEV SD init failed. SD card likely not inserted.
+ * @retval	-EINVAL Failed to append to buffer.
+ * @retval	-FR_INVALID_NAME Path is too long.
+ * @retval	Otherwise, error from underlying drivers.
  */
-int sd_card_list_files(char *path);
+int sd_card_list_files(const char *path, char *buf, size_t *buf_size);
```
koffes, line 30:
```suggestion
int sd_card_list_files(char const * const path, char *buf, size_t *buf_size);
```

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,99 @@
...
+ * @param[in, out]	size	Number of bytes to be read from file.
+ *				The actual read size will be returned.
+ *				If the actual read size is 0, there will be a warning message.
+ *				which indicates that the file is empty.
+ *
+ * @retval	0 on success.
+ * @retval	-ENODEV SD init failed. SD likely not inserted.
+ * @retval	Otherwise, error from underlying drivers.
+ */
+int sd_card_read(char *const data, size_t *size, struct fs_file_t *f_seg_read_entry);
```
koffes, line 89: since the data is being read into `data`, suggest removing the const, although this will work.

koffes (changes requested):

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,100 @@
...
-/**@brief Print out the contents under SD card root path
+/**@brief	Print out the contents under SD card root path and write the content to buffer.
+ *		This function is not thread safe
  *
- * @param[in] path	Path of the folder which going to list
- *			If assigned path is null, then listing the contents under root
- *			If assigned path doesn't exist, an error will be returned
+ * @param[in]		path		Path of the folder which is going to be listed.
+ *					If assigned path is null, then listing the contents under
+ *					root If assigned path doesn't exist, an error will be
```
alexsven, line 18:
```suggestion
 *					root. If assigned path doesn't exist, an error will be
```

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,100 @@
 #define _SD_CARD_H_
 
 #include <stddef.h>
+#include <zephyr/fs/fs.h>
 
-/**@brief Print out the contents under SD card root path
+/**@brief	Print out the contents under SD card root path and write the content to buffer.
+ *		This function is not thread safe
```
alexsven, line 14:
```suggestion
 *		This function is not thread safe.
```

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,100 @@
...
+ * @retval	Otherwise, error from underlying drivers.
+ */
+int sd_card_open(char const *const filename, struct fs_file_t *f_seg_read_entry);
+
+/**@brief	Read segment on the open file on the SD card.
+ *
+ * @param[out]		buf	Pointer to the buffer to write the read data into.
+ * @param[in, out]	size	Number of bytes to be read from file.
+ *				The actual read size will be returned.
+ *				If the actual read size is 0, there will be a warning message.
```
alexsven, line 83:
```suggestion
 *				If the actual read size is 0, there will be a warning message
```

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,100 @@
...
+ *				which indicates that the file is empty.
+ *
+ * @retval	0 on success.
+ * @retval	-ENODEV SD init failed. SD likely not inserted.
+ * @retval	Otherwise, error from underlying drivers.
+ */
+int sd_card_read(char *data, size_t *size, struct fs_file_t *f_seg_read_entry);
```
alexsven, line 90:
```suggestion
int sd_card_read(char *buf, size_t *size, struct fs_file_t *f_seg_read_entry);
```

alexsven: ``f_seg_read_entry`` is not documented here

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,100 @@
...
+
+/**@brief	Close the file opened by the sd_card_segment_read_open function.
+ *
+ * @retval	0 on success.
+ * @retval	-EBUSY Segment read operation has not started.
+ * @retval	Otherwise, error from underlying drivers.
  */
-int sd_card_read(char const *const filename, char *const data, size_t *size);
+int sd_card_close(struct fs_file_t *f_seg_read_entry);
```
alexsven, line 98: Missing documentation for ``f_seg_read_entry``

alexsven (changes requested):
It is generally the reviewer that resolves the conversations after the fixes has been done. Still some things that should be fixed, and two items that was probably hidden from before. Click `Load more...` to view all comments

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,69 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	if (ret < 0) {
+		LOG_ERR("Read file failed");
```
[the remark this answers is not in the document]
alexsven: Still not fixed

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +190,69 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	return 0;
+}
+
+int sd_card_close(struct fs_file_t *f_seg_read_entry)
+{
+	int ret;
+
+	ret = fs_close(f_seg_read_entry);
+	if (ret) {
+		LOG_ERR("Close file failed");
```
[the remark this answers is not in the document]
alexsven: Still not fixed

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,107 @@
...
+/**@brief	Read segment on the open file on the SD card.
+ *
+ * @param[out]		buf			Pointer to the buffer to write the read data into.
+ * @param[in, out]	size			Number of bytes to be read from file.
+ *						The actual read size will be returned.
+ *						If the actual read size is 0, there will be a
+ *						warning message which indicates that the file is
+ *						empty.
+ * @param[in, out]	f_seg_read_entry	Pointer to the location where the file should be
+ *entered.
```
alexsven, line 88: Check formatting. Also, since this is [in.out] does that mean that the pointer is updated to the end of the file once done? If so, that should be described here

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,107 @@
...
+ * @retval	0 on success.
+ * @retval	-ENODEV SD init failed. SD card likely not inserted.
+ * @retval	Otherwise, error from underlying drivers.
+ */
+int sd_card_open_read_close(char const *const filename, char *const buf, size_t *size);
+
+/**@brief	Open file on SD card.
+ *
+ * @param[in]		filename		Name of file to open.
+ * @param[in, out]	f_seg_read_entry	Pointer to the location where the file should be
```
alexsven, line 69: Is this pointer changed once open is done? If so, the documentation should describe it

alexsven: Same goes for all the documentation of f_seg_read_entry it seems

alexsven (changes requested):

alexsven (approved):

applications/nrf5340_audio/src/modules/sd_card.h
```diff
@@ -8,55 +8,113 @@
...
- * @param[in] data	The buffer which will be filled by read file contents
- * @param[in,out] size	Pointer to the number of bytes which wait to be read from the file
- *			The actual read size will be returned
- *			If the actual read size is 0, there will be a warning message which
- *			indicates the file is empty
- * @return	0 on success.
- *              -ENODEV SD init failed. SD card likely not inserted
- *		Otherwise, error from underlying drivers
+ * @param[in]		filename	Name of the target file for reading, the default location is
+ *					the root directoy of SD card, accept absolute path under
```
koffes, line 52: That doc must mention that only one open can be done at a time.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -24,18 +24,25 @@ LOG_MODULE_REGISTER(sd_card, CONFIG_MODULE_SD_CARD_LOG_LEVEL);
 static const char *sd_root_path = "/SD:";
 static FATFS fat_fs;
 static bool sd_init_success;
+static bool sd_read_ongoing;
```
koffes, line 27: Rename to sd_operation_ongoing.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -24,18 +24,25 @@ LOG_MODULE_REGISTER(sd_card, CONFIG_MODULE_SD_CARD_LOG_LEVEL);
...
 	int ret;
 	struct fs_dir_t dirp;
 	static struct fs_dirent entry;
 	char abs_path_name[PATH_MAX_LEN + 1] = SD_ROOT_PATH;
+	size_t used_buf_size = 0;
+
+	if (sd_read_ongoing) {
+		LOG_ERR("SD read ongoing");
+		return -EPERM;
+	}
```
koffes, line 46: Set the variable sd_oper_ongoing and to false at all returns.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +197,78 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+
+	fs_file_t_init(f_seg_read_entry);
+
+	ret = fs_open(f_seg_read_entry, abs_path_name, FS_O_READ);
+	if (ret) {
+		LOG_ERR("Open file failed: %d", ret);
+		return ret;
+	}
+
+	sd_read_ongoing = true;
```
koffes, line 237: Move to line 219.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +197,78 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+
+	return 0;
+}
+
+int sd_card_read(char *buf, size_t *size, struct fs_file_t *f_seg_read_entry)
+{
+	int ret;
+
```
koffes, line 245: Need to check that sd_read_ongoing is true.

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -180,6 +197,78 @@ int sd_card_read(char const *const filename, char *const data, size_t *size)
...
+	*size = ret;
+
+	return 0;
+}
+
+int sd_card_close(struct fs_file_t *f_seg_read_entry)
+{
+	int ret;
+
```
koffes, line 260: Need to check that sd_read_ongoing is true.

koffes (changes requested):

koffes:
Change the static variable to a semaphore.

koffes:
We are almost there. The semaphore cannot wait forever as this may cause the system to hang with no error message.
Suggest:
`#define K_SEM_OPER_TIMEOUT_MS 500` and use this instead of `K_FOREVER`:

`k_sem_take(&m_sem_sd_oper_ongoing, K_MSEC(K_SEM_OPER_TIMEOUT_MS));`

koffes (approved):

koffes:
Replaced by: https://github.com/nrfconnect/sdk-nrf/pull/12024

PR #11644, opened 2023-06-28
boards: thingy91_nrf52840: enable bt controller based on BT

bbilas:
Enable bluetooth controller and NOP event when BT is enabled by the user.

Signed-off-by: Bartosz Bilas <b.bilas@grinn-global.com>


bbilas:
@tejlmand gentle ping :)

PR #11650, opened 2023-06-29
doc: matter: add step to update module

greg-fer:
Added a step to configuration process about updating Matter module. This to make sure the CHIP Tool version is the latest one when building. NCSDK-22351.

### DIFF
--- a/doc/nrf/protocols/matter/getting_started/testing/thread_one_otbr.rst
+++ b/doc/nrf/protocols/matter/getting_started/testing/thread_one_otbr.rst
@@ -41,6 +41,14 @@ To configure and use Thread Border Router and Matter controller on the same devi
 
 .. rst-class:: numbered-step
 
+Update the Matter module
+========================
+
+Update the Matter module to the latest version by :ref:`gs_updating`.
+If you are working with a newly installed |NCS| or a specific release of the |NCS|, you can skip this step.
+
+.. rst-class:: numbered-step
+
 Program the sample
 ==================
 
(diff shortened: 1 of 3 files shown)


doc/nrf/protocols/matter/getting_started/testing/thread_one_otbr.rst
```diff
@@ -41,6 +41,14 @@ To configure and use Thread Border Router and Matter controller on the same devi
 
 .. rst-class:: numbered-step
 
+Update the Matter module
+========================
+
+Update the Matter module to the latest version by :ref:`gs_updating`.
+If you are working with a newly installed |NCS| or a specific release of the |NCS|, you can skip this step.
```
ArekBalysNordic, line 48: I wonder whether customers should skip this step when they're using a specific release. I can imagine the case when the customer has NCS 2.3.0 revision installed and calls for example: `git checkout v2.4.0` (update to a specific release) within sdk-nrf git repository and then forget to call `west update`. In particular cases, they will build an example, but the sdk-connectedhomeip won't be updated and some problems may occur. Maybe it is worth adding something like `... or you have installed a specific release of the |NCS|...` because it suggests that the customer went through the full installation of NCS SDK.

LuDuda: Oh.. But doing `west update` is like a MUST for everything in NCS when changing the sdk-nrf revision.
I find this step a bit ambiguous - for example what is a "Matter module" (probably sdk-connectedhomeip)?

I don't know if i understand the initial problem. Maybe we just need to make sure that user will use or build chip-tool from the proper nRF Connect revision. I thought that the request was to update the "Chip-tool" installation guide with additional sentence.

greg-fer: >  I think that the only thing that could be worth pointing out is that you need to make sure to build again if you change NCS version

I'll add this to NCSDK-22324

greg-fer:
Closing. The information about rebuilding after getting a new NCS version will be added in the Build and configuration guide for the whole NCS.

PR #11654, opened 2023-06-29
CoAP support for Location library and modem_shell

tokangas:
This PR adds nRF Cloud CoAP support to the Location library and modem_shell sample.

Commits from the nRF Cloud library CoAP PR #11535 are still included because the changes have not been merged yet. Only the last three commits belong to this PR.


tony-le-24 (approved):
LGTM with unit test changes

plskeggs (approved):
LGTM.  Lots of CI issues though.

tokangas:
> LGTM. Lots of CI issues though.

Yes, I'll go through them today. It is starting to be a bit complicated to support all different assistance configurations (nRF Cloud MQTT/REST/CoAP, LwM2M, SUPL and external). Luckily we do have the CI, which builds all configurations automatically...

jayteemo (approved):

lib/location/cloud_service/cloud_service_nrf.c
```diff
@@ -8,18 +8,23 @@
 #include <zephyr/logging/log.h>
 #if defined(CONFIG_NRF_CLOUD_MQTT)
 #include <net/nrf_cloud_location.h>
-#else
+#elif defined(CONFIG_NRF_CLOUD_REST)
 #include <net/nrf_cloud_rest.h>
+#else
```
glarsennordic, line 13: Can this be `elif defined(CONFIG_NRF_CLOUD_COAP)` ?

It's not worth blocking the PR if that's not practical, but it seems wise to be explicit. If we push forward on this, can that change be in a followup PR?

lib/location/cloud_service/cloud_service_nrf.c
```diff
@@ -107,13 +112,20 @@ int cloud_service_nrf_pos_get(
...
 	const struct nrf_cloud_rest_location_request loc_req = {
 		.cell_info = params->cell_data,
 		.wifi_info = params->wifi_data
 	};
+	struct nrf_cloud_location_result result;
 
+#if defined(CONFIG_NRF_CLOUD_REST)
 	LOG_DBG("Sending positioning request (REST)");
 	err = nrf_cloud_rest_location_get(&rest_ctx, &loc_req, &result);
+#else
```
glarsennordic, line 125: also ideally be explicit about COAP here

lib/location/method_gnss.c
```diff
@@ -271,6 +278,7 @@ static void method_gnss_nrf_cloud_agps_request(void)
 		return;
 	}
 	rest_ctx.auth = (char *)jwt_buf;
+#endif
```
glarsennordic, line 281: I'm gonna stop pointing this out -- just generally being explicit about COAP guards seems like a good idea

samples/cellular/modem_shell/src/cloud/cloud_coap_shell.c
```diff
@@ -0,0 +1,147 @@
...
+#include <net/nrf_cloud_coap.h>
+#include <zephyr/shell/shell.h>
+#include "mosh_print.h"
+
+BUILD_ASSERT(IS_ENABLED(CONFIG_NRF_CLOUD_COAP));
+
+BUILD_ASSERT(!IS_ENABLED(CONFIG_MOSH_CLOUD_LWM2M));
+
+extern struct k_work_q mosh_common_work_q;
+extern const struct shell *mosh_shell;
```
glarsennordic, line 18: This seems to be unused?

samples/cellular/modem_shell/src/gnss/gnss.c
```diff
@@ -113,7 +119,8 @@ static struct k_work get_pgps_data_work;
 	!defined(CONFIG_LWM2M_CLIENT_UTILS_LOCATION_ASSIST_AGPS)) || \
 	(defined(CONFIG_NRF_CLOUD_PGPS) && \
 	 !defined(CONFIG_LWM2M_CLIENT_UTILS_LOCATION_ASSIST_PGPS))) && \
-	!defined(CONFIG_NRF_CLOUD_MQTT)
+	!defined(CONFIG_NRF_CLOUD_MQTT) && \
+	defined(CONFIG_NRF_CLOUD_REST)
```
glarsennordic, line 123: I think this boolean logic needs to be split up and given some explanatory comments, maybe in followup PR. For instance it isn't clear from this why enabling MQTT should disable specifically the JWT buffer

samples/cellular/modem_shell/src/gnss/gnss.c
```diff
@@ -864,6 +877,11 @@ static void get_pgps_data_work_fn(struct k_work *work)
 #else
```
glarsennordic, line 877: It is super unclear which #if statement this belongs to

glarsennordic (approved):

lib/location/method_gnss.c
plskeggs: In this case, the whole function is wrapped in:
```
#elif defined(CONFIG_NRF_CLOUD_REST) || defined(CONFIG_NRF_CLOUD_COAP)
...
#endif
so the `#if defined(CONFIG_NRF_CLOUD_REST) #else...` at 281 is fine.

File: samples/cellular/modem_shell/src/gnss/gnss.c (hunk 5 above)
Comment (plskeggs) [reply samples/cellular/modem_shell/src/gnss/gnss.c:123]: I'll leave that for @tokangas.

File: lib/location/method_gnss.c [hunk 7]
```diff
@@ -302,7 +310,11 @@ static void method_gnss_nrf_cloud_agps_request(void)
 		sizeof(agps_rest_data_buf),
 		0};
 
+#if defined(CONFIG_NRF_CLOUD_REST)
 	err = nrf_cloud_rest_agps_data_get(&rest_ctx, &request, &result);
+#else
```
glarsennordic, line 315: Guard for COAP could be explicit here as well

samples/cellular/modem_shell/src/gnss/gnss.c
```diff
@@ -667,6 +675,7 @@ static void get_agps_data(struct k_work *item)
 		.rx_buf = rx_buf,
 		.rx_buf_len = sizeof(rx_buf),
 	};
+#endif
```
glarsennordic, line 678: Guard for COAP could be explicit here

samples/cellular/modem_shell/src/gnss/gnss.c
```diff
@@ -696,7 +705,11 @@ static void get_agps_data(struct k_work *item)
 		request.net_info = &net_info;
 	}
 
+#if defined(CONFIG_NRF_CLOUD_REST)
 	err = nrf_cloud_rest_agps_data_get(&rest_ctx, &request, &result);
+#else
```
glarsennordic, line 710: And here

glarsennordic (approved):
Looks even better

Since we are now explicitly guarding for COAP, I wonder if there should be some kind of assertion making sure that at least one of the options is selected

Possibly that is not necessary though, up to you

samples/cellular/modem_shell/src/gnss/gnss.c
glarsennordic: Hmm, I commented on an endif here. That seems like a mistake on my part. Possibly I grabbed the wrong line, one sec.

glarsennordic: IDK why I thought this was a coap branch, it clearly isn't. My bad.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -451,7 +457,13 @@ Modem libraries
...
 * :ref:`lib_location` library:
 
-  * Neighbor cell search is modified to use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
+  * Added:
+
+    * Added support for accessing nRF Cloud services using CoAP through the :ref:`lib_nrf_cloud_coap` library.
+
+  * Updated:
+
+    * Neighbor cell search is modified to use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
```
wiba-nordic, line 466:
```suggestion
    * Neighbor cell search to use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
```

wiba-nordic (changes requested):

wiba-nordic (approved):

tokangas:
> Since we are now explicitly guarding for COAP, I wonder if there should be some kind of build-time assertion making sure that at least one of the options is selected

@glarsennordic There already are build asserts in the Location library and modem_shell:

https://github.com/nrfconnect/sdk-nrf/blob/6ea4370e595f18143de9272e3d61dfe29b806861/lib/location/method_gnss.c#L34
https://github.com/nrfconnect/sdk-nrf/blob/6ea4370e595f18143de9272e3d61dfe29b806861/samples/cellular/modem_shell/src/gnss/gnss.c#L59

Or did you mean something else?

samples/cellular/modem_shell/src/gnss/gnss.c
tokangas: @glarsennordic I agree that this is ugly. The complexity comes from the fact that REST is not always used to fetch assistance data even though `CONFIG_NRF_CLOUD_REST` is enabled. For example, if LwM2M is enabled, it's preferred over REST and without this nasty flagging, a compilation warning would be generated for unused REST buffers. The same happens also if both `CONFIG_NRF_CLOUD_MQTT` and `CONFIG_NRF_CLOUD_REST` are enabled. The purpose of this flagging is to enable REST buffers only when REST is used, so we don't get compilation warnings with any of the supported configurations.

It would be simpler if only one transport could be enabled at a time, but currently mosh supports some combinations. There was even some discussion about adding even more supported combinations, but that would increase the complexity even more. I'm not sure what is the point, because for example if both MQTT and REST are enabled, I think that everything goes over MQTT anyway, so REST support can't be used for much. Probably to update the device shadow or perform JITP, but I think that's all.

Then again, if such combinations are supported by our libraries and make sense on some level, then it would be good to also test those.

If you like, I could add a comment to explain what the flagging does and why it's necessary.

glarsennordic: Yes, comments would be a good addition.

Beyond comments, I was also suggesting that you could split the boolean logic into expository statements.

I haven't tested, but I think this should work:

```
#define AGPS_USES_REST  (defined(CONFIG_NRF_CLOUD_REST) && \ 
			!defined(CONFIG_NRF_CLOUD_MQTT) && \
			 defined(CONFIG_NRF_CLOUD_AGPS) && \
			!defined(CONFIG_LWM2M_CLIENT_UTILS_LOCATION_ASSIST_AGPS))
#define PGPS_USES_REST  (defined(CONFIG_NRF_CLOUD_REST) && \ 
			!defined(CONFIG_NRF_CLOUD_MQTT) && \
			 defined(CONFIG_NRF_CLOUD_PGPS) && \
			!defined(CONFIG_LWM2M_CLIENT_UTILS_LOCATION_ASSIST_PGPS))
	
#if (AGPS_USES_REST || PGPS_USES_REST)
/* JWT and recieve buffers will be needed if either AGPS or PGPS uses REST */
static char jwt_buf[600];
static char rx_buf[2048];
#endif
```

glarsennordic:
> Or did you mean something else?

Nope, I hadn't seen those, my bad!

samples/cellular/modem_shell/src/gnss/gnss.c
tokangas: @glarsennordic Your proposal caused compilation warnings, so I had to change it a bit. Created PR #11875, let's see if all configurations can still be built without warnings.

PR #11659, opened 2023-06-30
nrf_desktop: update wake_up spelling to wakeup

greg-fer:
Updated occurrences in docs and code as discussed with owners. wake_up() function stays unchanged.
NCSDK-11637.


applications/nrf_desktop/doc/ble_latency.rst
```diff
@@ -39,7 +39,7 @@ You can set the option :ref:`CONFIG_DESKTOP_BLE_LOW_LATENCY_LOCK <config_desktop
 This speeds up sending the first HID report after not sending a report for some connection intervals.
 Enabling this option increases the power consumption - the connection latency is kept low unless the device is in the low power mode.
 
-You can use the :ref:`CONFIG_DESKTOP_BLE_LATENCY_PM_EVENTS <config_desktop_app_options>` Kconfig option to enable or disable handling of the power management events, such as :c:struct:`power_down_event` and :c:struct:`wake_up_event`.
+You can use the :ref:`CONFIG_DESKTOP_BLE_LATENCY_PM_EVENTS <config_desktop_app_options>` Kconfig option to enable or disable handling of the power management events, such as :c:struct:`power_down_event` and :c:struct:`wakeup_event`.
```
divipillai, line 42: The :c:struct:`wake_up_event` struct links to the API docs of the "power_event.h" file (like the power_down_event), which is present in the CAF library: 
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/libraries/caf/caf_overview.html#caf-power-events

Should we make the changes in the .rst, .c , and .h files of the CAF libraries? For example, For CAF libraries, the wake_up_event struct is mentioned in the following rst files:  known_issues.rst, buttons.rst, click_detector.rst, leds.rst, power_manager.rst , and sensor_manager.rst (also in release-notes-1.9.0.rst).

MarekPieta: If we want to apply the update, we must do it in all of the places. I am concerned about braking existing API and forcing customers to align with the change. If we agree that change is worth braking the API, I can push commit updating remaining places :)

divipillai: Also, if we decide to apply the changes in all the places, there are two figures of CAF libraries that would require updates. I can work on the figures :).

PR #11664, opened 2023-06-30
tests: nrf_modem_lib_trace: use FFF

MirkoCovizzi:
This commit replaces the custom mock implementation of `nrf_modem_at_printf` with fake function implementations using FFF.

### DIFF
--- a/tests/lib/nrf_modem_lib/nrf_modem_lib_trace/src/main.c
+++ b/tests/lib/nrf_modem_lib/nrf_modem_lib_trace/src/main.c
@@ -8,6 +8,7 @@
 #include <unity.h>
 #include <zephyr/toolchain/common.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/fff.h>
 #include <syscalls/rand32.h>
 #include <modem/nrf_modem_lib.h>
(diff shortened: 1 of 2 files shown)


tests/lib/nrf_modem_lib/nrf_modem_lib_trace/src/main.c
```diff
@@ -63,15 +62,48 @@ void nrf_modem_lib_trace_callback(enum nrf_modem_lib_trace_event evt)
 	callback_evt = evt;
 }
 
+static int nrf_modem_at_printf_trace_level_valid(const char *fmt, va_list args)
+{
+	TEST_ASSERT_EQUAL_STRING(TRACE_LEVEL_FMT, fmt);
+
+	int val = va_arg(args, int);
+
+	TEST_ASSERT_TRUE(
```
eivindj-nordic, line 71: We should be explicit here to test that the level is actually what we expect, e.g. `NRF_MODEM_LIB_TRACE_LEVEL_FULL`, not just one of the allowed ones.

eivindj-nordic (approved):

lemrey (approved):
Looks good

PR #11669, opened 2023-07-03
Bluetooth: controller: increase the default number of sync buffers

jori-nordic:
At the time of writing, the SDC cannot receive chained PDUs with only two buffers. It needs at least three.

With two, it can still sync to advertisers that have data that fits a single PDU though, so 2 can still be the minimum.


subsys/bluetooth/controller/Kconfig
```diff
@@ -169,7 +169,7 @@ config BT_CTLR_SDC_SCAN_BUFFER_COUNT
 
 config BT_CTLR_SDC_PERIODIC_SYNC_BUFFER_COUNT
 	int "Number of periodic synchronization receive buffers"
-	default 2
+	default 3
 	range 2 20
 	help
 	  The buffers are used for processing incoming packets
```
rugeGerritsen, line 175: Can you also document that 3 buffers are needed for chaining?

jori-nordic: where? in the kconfig help below?

rugeGerritsen (approved):

gopiotr (approved):

PR #11673, opened 2023-07-03
nrf_security: openthread config for mbedtls debug

pjasinski990:
Adds config that makes it possible to enable mbedtls debug configs with openthread. Currently it's impossible to select MBEDTLS_DEBUG if not using CUSTOM_OPENTHREAD_SECURITY (as it depends on MBEDTLS, which is promptless).


samples/openthread/cli/overlay-logging.conf
```diff
@@ -37,3 +37,7 @@ CONFIG_LOG_BACKEND_RTT=y
 # Disable UART logging backend
 # CONFIG_LOG_BACKEND_UART=n
 CONFIG_SHELL_LOG_BACKEND=n
+
+# mbedTLS logs
```
maciejbaczmanski, line 41:
```suggestion
# Enable Mbed TLS logging
```
Also shouldn't we move it a bit higher in the file? maybe before `# Enable OpenThread logging` ?

maciejbaczmanski (commented):
Thanks!
One comment, but it applies for all `.conf` files

maciejbaczmanski (approved):

edmont (approved):

canisLupus1313 (approved):

subsys/nrf_security/Kconfig.tls
```diff
@@ -439,3 +439,9 @@ endmenu
 
 
 endif # MBEDTLS_TLS_LIBRARY
+
+config OPENTHREAD_MBEDTLS_DEBUG
```
joerchan, line 443: Move this to "subsys/net/openthread" please.

joerchan (changes requested):

subsys/net/openthread/Kconfig
```diff
@@ -52,6 +52,12 @@ config OPENTHREAD_NRF_SECURITY
 config OPENTHREAD_MBEDTLS_LIB_NAME
 	default "mbedtls_external" if OPENTHREAD_NRF_SECURITY
 
+config OPENTHREAD_MBEDTLS_DEBUG
+	bool "MbedTLS logs for OpenThread"
+	select MBEDTLS
```
joerchan, line 57: Nit: Maybe a bit much for MBEDTLS_DEBUG to enable the MBEDTLS feature itself.

This is already done by OPENTHREAD_MBEDTLS, I'm thinking that should be enough.

joerchan (approved):

subsys/net/openthread/Kconfig
pjasinski990: The problem is we use OPENTHREAD_NRF_SECURITY (not OPENTHREAD_MBEDTLS). MBEDTLS is not selected in our default configuration.

PR #11691, opened 2023-07-04
.github: labeler.yaml: Change path to changelog after it moved

oyvindronningstad:
The changelog moved to a new subdirectory, update this so the PR labeler works properly (doesn't add the changelog-required label when the changelog has been updated).

NCSDK-22320

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -15,7 +15,7 @@
   - "**/*.rst"
 
 "changelog-entry-required":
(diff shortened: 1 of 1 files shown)


oyvindronningstad:
I tested it here, by adding a change to the changelog, but it didn't work, likely because the change needs to be in main to start being used by the labeler.

greg-fer (approved):
Thanks for this, @oyvindronningstad !

rlubos (approved):

PR #11692, opened 2023-07-04
nrf_security: Fix psa_driver_wrapper_get_builtin_key

SebastianBoe:
Bugfix for psa_driver_wrapper_get_builtin_key. Presumably introduced in 32ff463bebffff02ccb3e9b7b7285c72a3612db4 through a bad conflict resolution.

The consequences of the bug are unknown.


vili-nordic (approved):

joerchan:
> The consequences of the bug are unknown.

Pretty sure the function is only called with this argument inside TF-M, so I don't see this an at issue

Vge0rge (approved):

joerchan (approved):
I applied the patch manually, because the patch does not apply on our version of this file.
So this is just my mistake.

mswarowsky (approved):

PR #11698, opened 2023-07-05
test_spec: Move integration test coverage from PR to nightly

SebastianBoe:
Run less tests for every PR that patches nrf_security.

This reduces CI load, speeds up PR regression testing, and reduces the amount of time investigating unrelated CI issues.

In my experience it is rare for these tests to catch regressions in nrf_security.

As far as I can tell, any test failures that slip through the cracks will be caught in nightly so there is no loss of test coverage for releases.

It does not scale to integration test everything against everything on every PR.


shanthanordic (approved):
@Frodevan @krga2022 @tony-le-24

magnev (approved):

PR #11707, opened 2023-07-05
Migrate Zigbee unit tests to new ZTest API [KRKNWK 16665]

sebastiandraus:


tests/subsys/zigbee/osif/nvram/src/main.c
```diff
@@ -65,7 +67,7 @@ static void test_zb_nvram_erase(void)
 	}
 }
 
-static void test_zb_nvram_write(void)
+ZTEST(osif_test, test_zb_nvram_write)
 {
```
tomchy, line 71: (un)fortunately the new ZTEST API does not allow to rely on the order of tests (they are executed alphabetically or in a random order), so this test requires a small setUp function that will erase the area before writing to it.

gopiotr (changes requested):
Please remove in this PR (in separate commit) Zigbee tests from quarantine file:
https://github.com/nrfconnect/sdk-nrf/blob/57cccdab05311a13aaf6d8c329913b3fff27fda1/scripts/quarantine.yaml#L23-L28
Thanks to this we would be able to build zigbee tests in CI dedicated for this PR

mariuszpos (approved):

tests/subsys/zigbee/osif/nvram/src/main.c
sebastiandraus: Good catch, fix already applied :)

sebastiandraus:
> Please remove in this PR (in separate commit) Zigbee tests from quarantine file:
> 
> https://github.com/nrfconnect/sdk-nrf/blob/57cccdab05311a13aaf6d8c329913b3fff27fda1/scripts/quarantine.yaml#L23-L28
> 
> Thanks to this we would be able at least build zigbee tests in CI dedicated for this PR

Done!

tests/subsys/zigbee/osif/serial/serial_async_api/src/main.c
```diff
@@ -324,48 +331,6 @@ void test_main(void)
 	/* Enable zigbee callbacks implemented in zigbee_scheduler mock. */
 	zigbee_enable();
```
gopiotr, line 332: Is there anything against moving this `zigbee_enable()` and those `printk` to testsuite setup function `configure_test_suite`? Thanks to this you could remove `test_main` function from this test.

tests/subsys/zigbee/osif/serial/serial_basic_api/src/main.c
```diff
@@ -167,24 +173,6 @@ void test_main(void)
 	/* Enable zigbee callbacks implemented in zigbee_scheduler mock. */
 	zigbee_enable();
```
gopiotr, line 174: The same issue as in `serial_basic_api`.

gopiotr (changes requested):

tests/subsys/zigbee/osif/serial/serial_async_api/src/main.c
sebastiandraus: Good idea, already applied :)

gopiotr (approved):

tomchy (approved):

PR #11708, opened 2023-07-05
lib: nrf_modem: trace: fixes for uart trace backend++

eivindj-nordic:
If the UART trace backend has flow control enabled, the trace backend might abort the current send, hence 0 bytes will be written. Return an error in this case to not assert on the trace backend returning 0 bytes.


lemrey:
Isn't that an issue with how we define the `trace_backend_write` interface?
Why do we allow that to return `0`? Is it necessary to differentiate between `0` and errors?

eivindj-nordic:
> Isn't that an issue with how we define the `trace_backend_write` interface? Why do we allow that to return `0`? Is it necessary to differentiate between `0` and errors?

It depends. The alternative is to let the trace backend loop forever in this case instead, and log the warning there. But this will happen in some cases, where we do not want to assert.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -381,7 +381,11 @@ Modem libraries
 
 * :ref:`nrf_modem_lib_readme`:
 
-  * Updated the :c:func:`nrf_modem_lib_shutdown` function to allow the modem to be in flight mode (``CFUN=4``) when shutting down the modem.
+  Updated:
+
+    * The :c:func:`nrf_modem_lib_shutdown` function to allow the modem to be in flight mode (``CFUN=4``) when shutting down the modem.
+    * The trace backends can now return ``-EAGAIN`` if the write operation can be retried.
+    * The trace backend deinit function is expected to succeed and can no longer return an error.
```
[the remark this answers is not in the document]
eivindj-nordic: There is no function called `trace_backend_deinit`, but there is a `deinit` field in the `nrf_modem_lib_trace_backend` struct with a template for the function signature. I can add `deinit` in backticks.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -381,7 +381,11 @@ Modem libraries
 
 * :ref:`nrf_modem_lib_readme`:
 
-  * Updated the :c:func:`nrf_modem_lib_shutdown` function to allow the modem to be in flight mode (``CFUN=4``) when shutting down the modem.
+  Updated:
+
+    * The :c:func:`nrf_modem_lib_shutdown` function to allow the modem to be in flight mode (``CFUN=4``) when shutting down the modem.
+    * The trace backends can now return ``-EAGAIN`` if the write operation can be retried.
+    * The trace backends ``deinit`` function is expected to succeed and can no longer return an error.
```
divipillai, line 388: @eivindj-nordic : There is one more entry : _Added CEREG event tracking to lte_connectivity_ that should be part of :ref:`nrf_modem_lib_readme`: . Could you move this entry here.

Thanks for fixing this.

eivindj-nordic: done

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -381,7 +381,13 @@ Modem libraries
...
 * :ref:`nrf_modem_lib_readme`:
 
-  * Updated the :c:func:`nrf_modem_lib_shutdown` function to allow the modem to be in flight mode (``CFUN=4``) when shutting down the modem.
+  * Added CEREG event tracking to ``lte_connectivity``.
+
+  Updated:
+
+    * The :c:func:`nrf_modem_lib_shutdown` function to allow the modem to be in flight mode (``CFUN=4``) when shutting down the modem.
+    * The trace backends can now return ``-EAGAIN`` if the write operation can be retried.
+    * The trace backends ``deinit`` function is expected to succeed and can no longer return an error.
```
divipillai, line 390:
```suggestion
  * Added CEREG event tracking to ``lte_connectivity``.

  * Updated:

    * The :c:func:`nrf_modem_lib_shutdown` function to allow the modem to be in flight mode (``CFUN=4``) when shutting down the modem.
    * The trace backends can now return ``-EAGAIN`` if the write operation can be retried.
    * The trace backends ``deinit`` function is expected to succeed and can no longer return an error.
```
nit

divipillai (approved):

nordic-mare (approved):

include/modem/trace_backend.h
```diff
@@ -40,11 +40,8 @@ struct nrf_modem_lib_trace_backend {
 
 	/**
 	 * @brief Deinitialize the compile-time selected trace backend.
-	 *
-	 * @return 0 If the operation was successful.
-	 *         Otherwise, a (negative) error code is returned.
 	 */
-	int (*deinit)(void);
+	void (*deinit)(void);
```
lemrey, line 44: I think it's best to keep it returning an `int`, even if current implementations can't fail. That is still a possibility if they change or new ones are added; by removing the return value and the log it's impossible to tell if this operation failed.

lemrey (commented):
One small comment otherwise looks good

include/modem/trace_backend.h
eivindj-nordic: updated

lemrey (approved):

PR #11719, opened 2023-07-06
drivers: wifi: Configure BSS max idle time

rado17:
Pass default BSS max idle time to UMAC.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -110,7 +110,7 @@ manifest:
     - name: hostap
       repo-path: sdk-hostap
       path: modules/lib/hostap
-      revision: 19cc5bdb6d482ba2ca65006fb7bc9a33db72ff45
+      revision: 42a92ba947785b35ad4e85cbf6e121e4d9a3bd50
       userdata:
         ncs:
           upstream-url: https://w1.fi/cgit/hostap/
--- a/drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fw/host_rpu_umac_if.h
+++ b/drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fw/host_rpu_umac_if.h
@@ -1529,6 +1529,7 @@ struct nrf_wifi_umac_assoc_info {
 	signed char control_port;
 	unsigned int prev_bssid_flag;
 	unsigned char prev_bssid[NRF_WIFI_ETH_ADDR_LEN];
+	unsigned short bss_max_idle_time;
 } __NRF_WIFI_PKD;
 
 /**
--- a/drivers/wifi/nrf700x/osal/fw_if/umac_if/src/default/fmac_api.c
+++ b/drivers/wifi/nrf700x/osal/fw_if/umac_if/src/default/fmac_api.c
@@ -881,6 +881,10 @@ enum wifi_nrf_status wifi_nrf_fmac_assoc(void *dev_ctx,
 		connect_common_info->nrf_wifi_flags |= NRF_WIFI_CONNECT_COMMON_INFO_PREV_BSSID;
 	}
 
+	if (assoc_info->bss_max_idle_time) {
+		connect_common_info->maxidle_insec = assoc_info->bss_max_idle_time;
+	}
+
 	status = umac_cmd_cfg(fmac_dev_ctx,
 			      assoc_cmd,
 			      sizeof(*assoc_cmd));
--- a/drivers/wifi/nrf700x/zephyr/src/zephyr_wpa_supp_if.c
+++ b/drivers/wifi/nrf700x/zephyr/src/zephyr_wpa_supp_if.c
@@ -822,6 +822,10 @@ int wifi_nrf_wpa_supp_associate(void *if_priv, struct wpa_driver_associate_param
 		assoc_info.use_mfp = NRF_WIFI_MFP_REQUIRED;
 	}
 
+	if (params->bss_max_idle_period) {
+		assoc_info.bss_max_idle_time = params->bss_max_idle_period;
+	}
+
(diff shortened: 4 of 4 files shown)


drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -284,4 +284,12 @@ config NRF700X_RSSI_STALE_TIMEOUT_MS
 	  value as driver does not store it and requires RPU to provide the
 	  info.
 
+config NRF700X_BSS_MAX_IDLE_TIME
+	int "BSS max idle timeout in seconds"
+	default 300
```
ajayparida, line 289: Missing the range(min , max) of values.

drivers/wifi/nrf700x/osal/fw_if/umac_if/src/cmd.c
```diff
@@ -169,6 +169,7 @@ enum wifi_nrf_status umac_cmd_init(struct wifi_nrf_fmac_dev_ctx *fmac_dev_ctx,
 #endif /* !CONFIG_NRF700X_RADIO_TEST */
 
 	umac_cmd_data->op_band = op_band;
+	umac_cmd_data->maxidle_insec = CONFIG_NRF700X_BSS_MAX_IDLE_TIME;
```
ajayparida, line 172: Do we need to check the range here

ajayparida (changes requested):

drivers/wifi/nrf700x/osal/fw_if/umac_if/src/cmd.c
rado17: Putting a range restriction in Kconfig should be enough. That blocks the build if an out-of-range default value is configured in Kconfig.

drivers/wifi/nrf700x/Kconfig.nrf700x
rado17: Done.

ajayparida (approved):

drivers/wifi/nrf700x/osal/fw_if/umac_if/src/cmd.c
krish2718, line 172: Shouldn't this be coming from WPA supplicant?

rado17: Looks like supplicant doesn't support setting bss_max_idle_time for station. Also, current requirement is to pass the max idle timeout at bootup. So passing it here seems correct.

krish2718: It does support but it is not passed as part of `wpa_driver_associate_params` in `associate` callback and passing it in the bootup is not quite right, it has to be part of the associate/connect command.

krish2718 (changes requested):
As discussed offline, lets

1. Move this to `associate` as this has to be per-association (needs UMAC patch)
2. Move Kconfig to WPA supplicant and patch `wpa_driver_associate_params` (Can be an upstream patch)
3. Add BSS_IDLE_MGMT_OFFLOAD to decide who does the BSS Idle management (WPA supplicant or UMAC)

drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -301,4 +301,14 @@ config NRF700X_RSSI_STALE_TIMEOUT_MS
 	  value as driver does not store it and requires RPU to provide the
 	  info.
 
+config BSS_IDLE_MGMT_OFFLOAD
+	int
```
krish2718, line 305: This can just be a bool and IMHO this need not be exposed to users, we can just decide this in the code.

drivers/wifi/nrf700x/osal/fw_if/umac_if/src/cmd.c
```diff
@@ -170,6 +170,8 @@ enum wifi_nrf_status umac_cmd_init(struct wifi_nrf_fmac_dev_ctx *fmac_dev_ctx,
 
 	umac_cmd_data->op_band = op_band;
 
+	umac_cmd_data->mgmt_buff_offload = CONFIG_BSS_IDLE_MGMT_OFFLOAD;
```
krish2718, line 173: This is wrong, IIRC, `mgmt_buff_offload` was for whether UMAC needs RX buffers from host or not. This is for BSS IDLE management.

krish2718 (changes requested):

drivers/wifi/nrf700x/Kconfig.nrf700x
rado17: In that case, will remove this from here and define it as a macro in header.

drivers/wifi/nrf700x/osal/fw_if/umac_if/src/cmd.c
rado17: True. I read this wrong. I will remove this for now since BSS_MAX_IDLE_TIME_OFFLOAD is yet to be supported in UMAC.

drivers/wifi/nrf700x/Kconfig.nrf700x
rado17: Removing this since UMAC is yet to add support for this.

krish2718 (approved):

west.yml
```diff
@@ -110,7 +110,7 @@ manifest:
     - name: hostap
       repo-path: sdk-hostap
       path: modules/lib/hostap
-      revision: 8e2e758b0296e0f6e8a2261884dca15da37d96a8
+      revision: pull/108/head
```
krish2718, line 113: Update to `372c30e40a977c1dbfd6b559eb1e22d447c21e3e`

rado17: Done

PR #11720, opened 2023-07-06
samples: Bluetooth: encrypt llpm sample

martintv:
It makes most sense to demonstrate llpm on an ecrypted link. Also setting down conection interval to the smallest bluetooth allows. 7.5ms. Since encyrpting the link takes quite a lot of packets back and forth, this will speed up setup. But also it makes sense that the sample compares the best bluetooth can do with the best llpm can do.

### DIFF
--- a/samples/bluetooth/llpm/src/main.c
+++ b/samples/bluetooth/llpm/src/main.c
@@ -27,8 +27,8 @@
 
 #define DEVICE_NAME	CONFIG_BT_DEVICE_NAME
 #define DEVICE_NAME_LEN (sizeof(DEVICE_NAME) - 1)
-#define INTERVAL_MIN    0x50     /* 80 units,  100 ms */
-#define INTERVAL_MAX    0x50     /* 80 units,  100 ms */
+#define INTERVAL_MIN    0x6     /* 6 units,  7.5 ms */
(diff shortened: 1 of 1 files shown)


samples/bluetooth/llpm/src/main.c
```diff
@@ -414,10 +412,24 @@ static void test_run(void)
 	}
 }
 
+void security_changed(struct bt_conn *conn, bt_security_t level,
+				 enum bt_security_err err)
+{
+	printk("Security changed: level %i, err: %i\n", level, err);
+
+	/*Start service discovery when link is encrypted*/
```
jori-nordic, line 420: can you add a check that `err` is non-zero and disconnect/assert if that's the case?

martintv: done

martintv:
image
capture from running sample in bsim on top of this PR

jori-nordic:
@martintv you might want to pass the cmdline option `-RealEncryption=1` to the bsim devices, else bt_encrypt will not really encrypt the data, that's why it appears in cleartext after the `LL_START_ENC`s are exchanged.

jori-nordic (approved):

alwa-nordic (approved):

PR #11721, opened 2023-07-06
manifest: update zephyr with nrf5340 net core crash fix

ankuns:
This commit brings in zephyr containing the fix against crashes of nrf5340 net core when wfi instruction was followed by access to the memory.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -59,7 +59,7 @@ manifest:
     # https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/modules.html
     - name: zephyr
       repo-path: sdk-zephyr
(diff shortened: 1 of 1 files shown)


west.yml
```diff
@@ -59,7 +59,7 @@ manifest:
     # https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/modules.html
     - name: zephyr
       repo-path: sdk-zephyr
-      revision: 52222dc79450304fe4a32e5e0e77b9a460a93a86
+      revision: pull/1231/head
```
rlubos, line 62: `77ed2a43ed4f52fd9ce19c71525617ac2a9948a7`

rlubos (approved):

PR #11726, opened 2023-07-06
drivers: wifi: Support for forced Passive scan

ajayparida:
Default scan mode is set as Active scan. User can force passive scan from Kconfig or from shell using 'passive' option. Once set it will override regulatory settings and forces all channels to be passive.


krish2718 (commented):
For the 1st commit, can you do a `git revert` of old commit that introduced the change.

krish2718 (approved):

sachinthegreen (approved):

PR #11731, opened 2023-07-06
Rtc pretick

pdunaj:


samples/nrf5340/empty_app_core/src/main.c
```diff
@@ -39,8 +40,26 @@ void disable_ram_and_wfi(register volatile uint32_t *reg_begin,
 void main(void)
 {
 	/* Power off RAM and suspend CPU */
```
hubertmis, line 42: This comment applies only to `else` now

subsys/net_core_monitor/app_core.c
```diff
@@ -0,0 +1,69 @@
...
+	uint32_t reset_cnt = mem >> 16;
+
+	LOG_DBG("mem: 0x%08X", mem);
+
+	if (reset_cnt != prev_reset_cnt) {
+		prev_reset_cnt = reset_cnt;
+		ret = -EFAULT;
+	} else {
+		diff = (cnt - prev_cnt) & BIT_MASK(16);
+		if (diff == 0 || diff > 3) {
```
hubertmis, line 47: Why 3? If this function was called from a thread with low priority Net Core could increment its counter much more than 3 between two invocations of this function.

pdunaj: The feeding interval is long. If app cannot schedule threat one can increase the interval.

Still you are right as I don't see any reason we have `3` as a limit. I think this was added to detect the overflow but it should be done neater. Maybe `> UINT16_MAX` or maybe simply use `uint16_t` for cnt and prev_cnt and let overflow be handled within type.
FYI  @maje-emb .

hubertmis:
I'm thinking if we could create a solution that would not require periodic wake-ups of App or Net cores. Maybe something with IPC instead of shared memory?

PR #11733, opened 2023-07-06
net: nrf_provisioning: Add kconfig to save the command id

juhaylinen:
Add new kconfig NRF_PROVISIONING_SAVE_CMD_ID. When enabled the latest command id is saved to storage after successful Finished command.

### DIFF
--- a/subsys/net/lib/nrf_provisioning/Kconfig
+++ b/subsys/net/lib/nrf_provisioning/Kconfig
@@ -38,6 +38,11 @@ config NRF_PROVISIONING_WITH_CERT
 	help
 	  Includes the root certificate used by the server side and provisions it if needed.
 
+config NRF_PROVISIONING_SAVE_CMD_ID
+	bool "Save the latest command id to storage"
+	help
+	  Saves the latest command id to storage after successful Finished command.
+
 rsource "Kconfig.nrf_provisioning_http"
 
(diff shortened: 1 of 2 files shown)


subsys/net/lib/nrf_provisioning/Kconfig
```diff
@@ -38,6 +38,12 @@ config NRF_PROVISIONING_WITH_CERT
 	help
 	  Includes the root certificate used by the server side and provisions it if needed.
 
+config NRF_PROVISIONING_SAVE_CMD_ID
+	bool "Save the latest command id to storage"
+	default n
```
MarkusLassila, line 43: Should this be y, as we were already doing it? I tried to look what was being done with the stored value and to me it seemed that it was only used for informational purposes at the moment. If there is no real usage, then the default value does not matter.

jarlamsa: Btw, defaulting to n is redundant as it is by default n.

juhaylinen: Before this change the command id was saved to storage and loaded when the client starts. After some internal discussion, we decided to change this to optional and not to enable by default. This allows the customer to re-send the full provisioning configuration.

MarkusLassila (approved):

jarlamsa (approved):

PR #11734, opened 2023-07-06
tests: net: lib: nrf_cloud: add fota_common tests

jayteemo:
Add test suite for nrf_cloud_fota_common.c.
IRIS-4790


tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,716 @@
...
+	zassert_equal(IS_ENABLED(CONFIG_NRF_MODEM), ret, MDM_DELTA_ASSERT_STR);
+
+	ret = nrf_cloud_fota_is_type_enabled(NRF_CLOUD_FOTA_MODEM_FULL);
+	zassert_equal(IS_ENABLED(CONFIG_NRF_CLOUD_FOTA_FULL_MODEM_UPDATE) &&
+		      IS_ENABLED(CONFIG_NRF_MODEM),
+		      ret,
+		      FMFU_ASSERT_STR);
+
+	ret = nrf_cloud_fota_is_type_enabled(NRF_CLOUD_FOTA_TYPE__INVALID);
+	zassert_false(ret, "return should be false when invalid FOTA type is provided");
```
glarsennordic, line 81: Each of these could arguably be a separate test case

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,716 @@
...
+
+#define MDM_DELTA_ASSERT_STR "DELTA MODEM FOTA enabled should correspond to the value of: " \
+			     "CONFIG_NRF_MODEM"
+
+#define FMFU_ASSERT_STR "FULL MODEM FOTA enabled should correspond to the value of: " \
+			"CONFIG_NRF_CLOUD_FOTA_FULL_MODEM_UPDATE + CONFIG_NRF_MODEM"
+
+ZTEST(nrf_cloud_fota_common_test, test_nrf_cloud_fota_is_type_enabled)
+{
+	Z_TEST_SKIP_IFNDEF(CONFIG_FOTA_DOWNLOAD);
```
glarsennordic, line 57: I think you can remove this skip statement, these functions should still give valid results if `CONFIG_FOTA_DOWNLOAD` isn't enabled

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,716 @@
...
+
+ZTEST(nrf_cloud_fota_common_test, test_nrf_cloud_bootloader_fota_slot_set_get_s0_get_fail)
+{
+	Z_TEST_SKIP_IFNDEF(CONFIG_FOTA_DOWNLOAD);
+	struct nrf_cloud_settings_fota_job job;
+
+	job.bl_flags = NRF_CLOUD_FOTA_BL_STATUS_CLEAR;
+	job.type = NRF_CLOUD_FOTA_BOOTLOADER;
+
+	fota_download_s0_active_get_fake.custom_fake = fake_fota_download_s0_active_get__fails;
```
glarsennordic, line 92: Does FFF automatically reset custom fakes at the start of each test?

glarsennordic: It seems possibly not -- and if that is the case, we will need to explicitly do so

Calling `RESET_FAKE(fota_download_s0_active_get);` and similar for all fakes from the `run_before` function would suffice. But definitely double check that this is actually needed

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,716 @@
...
+ZTEST(nrf_cloud_fota_common_test, test_nrf_cloud_bootloader_fota_slot_fail)
+{
+	int ret = nrf_cloud_bootloader_fota_slot_set(NULL);
+
+	zassert_equal(-EINVAL, ret, "return should be -EINVAL when job is NULL");
+}
+
+ZTEST(nrf_cloud_fota_common_test, test_nrf_cloud_bootloader_fota_slot_fota_not_enabled)
+{
+	Z_TEST_SKIP_IFDEF(CONFIG_FOTA_DOWNLOAD);
```
glarsennordic, line 111: I could be misremembering, but I've run into troubles in the past where skipping a test marks the test suite as a failure. Is that a risk with Z_TEST_SKIP_IFDEF?

I think I only ran into that with Unity/Cmock. I think it might be different for ZTEST

glarsennordic (commented):
Stopping review here since I suspect with some added comments reviewing will go faster

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,716 @@
...
+
+		if ((fota_type == NRF_CLOUD_FOTA_MODEM_FULL) ||
+		    (fota_type == NRF_CLOUD_FOTA_MODEM_DELTA)) {
+			zassert_true(nrf_cloud_fota_is_type_modem(fota_type),
+				     "Modem FOTA type not detected");
+		} else {
+			zassert_false(nrf_cloud_fota_is_type_modem(fota_type),
+				      "Non-modem FOTA type detected as modem type");
+		}
+	}
```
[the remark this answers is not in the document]
glarsennordic: > if a non-modem type is updated, this test won't need to be updated, which seems like a good thing?

From my perspective this is actually a (very mildly) bad thing -- I think tests are most useful when they serve as a checksum for developer intent. So it is better, from that perspective, to explicitly list the intent of each individual FOTA type.

Emphasis on very mildly, in this case. Don't take this message as me insisting, I'm just explaining my reasoning. For something as minor as this it seems fine to leave it as is if that's what you prefer

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,716 @@
...
+}
+
+#define APP_FOTA_ASSERT_STR "APP FOTA enabled should correspond to the value of: " \
```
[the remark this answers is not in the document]
glarsennordic: You can definitely get compliance to accept very long messages, but you will need to break them into multi line strings

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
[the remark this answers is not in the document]
jayteemo: yes i understand the intent, but this test is for testing modem types, so i will leave it as is.

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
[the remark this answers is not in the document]
jayteemo: hmm i thought i tried that... but maybe i didn't do it correctly.  i'll try again.

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
glarsennordic: Some folks insist upon it, even when it makes the test cases significantly more complex! The idea is that each unit test should focus on a single minimal function. This theoretically makes the test easier to maintain, understand, and interpret in logs.

IMO splitting tests up like that is definitely desirable, but only worthwhile when it does not come at a significant cost to test complexity / readability / redundancy.

In this case I can't see any costs to doing it, so I lean toward splitting it

But this is another one of those situations where it's really not a big deal, so if you want to keep it as all one test, it isn't too huge of a bother for me

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
glarsennordic: > does simply setting the fake function to the desired value before use have some unintended consequence?

The main fear is this: If you, or anyone after you, ever accidentally writes a test which depends on which fake has been loaded, and forgets to explicitly set the fake in the test, then that test will have differing behavior depending on the order in which tests are executed

Such a test is very undesirable because it may appear fine until some subtle change is made in the tests that run before it, and then it may suddenly fail

So generally it is desirable to do what you can to guarantee that the starting conditions for each test is totally independent of what happened in the tests that ran before it. And so you would want to reset your fakes between each test

jayteemo: so, no unintended consequences... just good practice to reset the fakes ✔️ 

looks like other projects do the following in setup:
```
/* Register resets */
DO_FOREACH_FAKE(RESET_FAKE);
/* reset common FFF internal structures */
FFF_RESET_HISTORY();
```
so i'll add that to `run_before `.

glarsennordic: My bad, I was kind of conceiving of it as an unintended consequence, but it is more of a good practice thing yeah

I think that `DO_FOREACH_FAKE` thing is specific to the modem team's unit tests, they had to actually write a custom macro for that

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,716 @@
...
+
+/* This function runs after each completed test */
+static void run_after(void *fixture)
+{
+	ARG_UNUSED(fixture);
+}
+
+ZTEST_SUITE(nrf_cloud_fota_common_test, NULL, NULL, run_before, run_after, NULL);
+
+ZTEST(nrf_cloud_fota_common_test, test_nrf_cloud_fota_is_type_modem)
```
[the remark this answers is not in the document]
jayteemo: i have added comments, in english language, for each test.

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,823 @@
...
+	int ret = nrf_cloud_pending_fota_job_process(&job, &reboot_required);
+
+	zassert_equal(0, ret, "return should be 0 when app FOTA is already confirmed");
+	zassert_equal(NRF_CLOUD_FOTA_VALIDATE_UNKNOWN, job.validate,
+		      "validate status should be UNKNOWN on when write is already confirmed");
+}
+
+/* Verify that nrf_cloud_pending_fota_job_process succeeds when a pending
+ * APP FOTA validation is UNKNOWN because CONFIG_MCUBOOT_IMG_MANAGER is not enabled.
+ */
```
glarsennordic, line 342:
```suggestion
/* Verify that nrf_cloud_pending_fota_job_process succeeds when and sets the job validation status to UNKNOWN if
 * provided a pending APP FOTA job and CONFIG_MCUBOOT_IMG_MANAGER is not enabled.
 */
 ```
 
 Also, can you explain (to me at least, not sure yet if it needs to be in the test comments) why UNKNOWN is expected in this case?

File: tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c [hunk 9]
```diff
@@ -0,0 +1,823 @@
...
+	job.validate = NRF_CLOUD_FOTA_VALIDATE_PENDING;
+	job.type = NRF_CLOUD_FOTA_APPLICATION;
+
+	boot_is_img_confirmed_fake.custom_fake = fake_boot_is_img_confirmed__true;
+
+	int ret = nrf_cloud_pending_fota_job_process(&job, &reboot_required);
+
+	zassert_equal(0, ret, "return should be 0 when app FOTA is already confirmed");
+	zassert_equal(NRF_CLOUD_FOTA_VALIDATE_UNKNOWN, job.validate,
+		      "validate status should be UNKNOWN on when write is already confirmed");
```
glarsennordic, line 337: is `on when` a typo?

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,823 @@
...
+	zassert_equal(0, ret, "return should be 0 when app FOTA is processed as failure");
+	zassert_equal(true, reboot_required,
+		      "reboot_required should be true when app FOTA is processed as failure");
+	zassert_equal(NRF_CLOUD_FOTA_VALIDATE_FAIL, job.validate,
+		      "validate status should be FAIL when app FOTA is processed as failure");
+}
+
+/* Verify that nrf_cloud_pending_fota_job_process succeeds when a pending
+ * APP FOTA validation is UNKNOWN.
+ */
```
glarsennordic, line 321: I want to rephrase this as `Verify that nrf_cloud_pending_fota_job_process succeeds and sets validate status to UNKNOWN if XXX`

But the intended condition XXX is unclear to me

Is UNKNOWN just always expected for APP FOTA jobs? How does `boot_is_img_confirmed` relate to it?

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,823 @@
...
+	job.type = NRF_CLOUD_FOTA_BOOTLOADER;
+	fota_download_s0_active_get_fake.custom_fake = fake_fota_download_s0_active_get__s0_active;
+
+	int ret = nrf_cloud_bootloader_fota_slot_set(&job);
+
+	zassert_equal(0, ret,
+		      "return should be 0");
+	zassert_equal(NRF_CLOUD_FOTA_BL_STATUS_S0_FLAG_SET|NRF_CLOUD_FOTA_BL_STATUS_S0_WAS_ACTIVE,
+		      job.bl_flags,
+		      "bl_flags should be set and active");
```
glarsennordic, line 203: Did you mean `S0 flag` instead of `bl_flags` here?

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,823 @@
...
+	ret = nrf_cloud_bootloader_fota_slot_set(&job);
+
+	zassert_equal(0, ret,
+		      "return should be 0 when flag already set");
+	zassert_equal(NRF_CLOUD_FOTA_BL_STATUS_S0_FLAG_SET, job.bl_flags,
+		      "bl_flags should be NRF_CLOUD_FOTA_BL_STATUS_S0_FLAG_SET");
+
+}
+
+/* Verify that nrf_cloud_bootloader_fota_slot_set returns success and the flag is set.
```
glarsennordic, line 184: Can you specify the precondition for expecting this behavior?

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,823 @@
...
+	job.bl_flags = NRF_CLOUD_FOTA_BL_STATUS_CLEAR;
+	job.type = NRF_CLOUD_FOTA_BOOTLOADER;
+
+	int ret = nrf_cloud_bootloader_fota_slot_set(&job);
+
+	zassert_equal(-ENOTSUP, ret,
+		      "return should be -ENOTSUP when CONFIG_FOTA_DOWNLOAD is not enabled");
+}
+
+/* Verify that nrf_cloud_bootloader_fota_slot_set returns success and no flag change.
```
glarsennordic, line 154: Can you specify the precondition for expecting this behavior?

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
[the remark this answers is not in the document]
glarsennordic: Thanks for adding those, overall they are a very nice addition

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
jayteemo: https://github.com/nrfconnect/sdk-nrf/blob/2006366ca2bc8437bbebbcb7072413abe064dd3f/subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_common.c#L170

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,823 @@
...
+/* Verify that nrf_cloud_pending_fota_job_process succeeds when a pending
+ * FMFU FOTA validation FAILS because the external flash device is not set.
+ * The reboot flag should become true.
+ * This test must be called before nrf_cloud_fota_fmfu_dev_set executes successfully.
+ */
+ZTEST(nrf_cloud_fota_common_test, test_06_nrf_cloud_pending_fota_job_process_fmfu_validate_fail)
+{
+	Z_TEST_SKIP_IFNDEF(CONFIG_NRF_CLOUD_FOTA_FULL_MODEM_UPDATE);
+
+#if defined(CONFIG_NRF_CLOUD_FOTA_FULL_MODEM_UPDATE)
```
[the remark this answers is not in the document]
jayteemo: i moved them into their own file.

tests/subsys/net/lib/nrf_cloud/fota_common/src/fmfu.c
```diff
@@ -0,0 +1,240 @@
...
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+#include "main.c"
+
+/* Verify that nrf_cloud_fota_fmfu_dev_set returns failure when the function parameter is NULL. */
+ZTEST(nrf_cloud_fota_common_test, test_00_nrf_cloud_fota_fmfu_dev_set_null_fdev)
+{
+	Z_TEST_SKIP_IFDEF(CONFIG_DFU_TARGET_FULL_MODEM_USE_EXT_PARTITION);
```
plskeggs, line 11: Why is this skipped if CONFIG_DFU_TARGET_FULL_MODEM_USE_EXT_PARTITION=y?

plskeggs (approved):

tests/subsys/net/lib/nrf_cloud/fota_common/src/fmfu.c
jayteemo: https://github.com/nrfconnect/sdk-nrf/blob/2006366ca2bc8437bbebbcb7072413abe064dd3f/subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_common.c#L54

because it is testing the code in that if statement

plskeggs: Got it.

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
```diff
@@ -0,0 +1,546 @@
...
+	ret = nrf_cloud_bootloader_fota_slot_set(&job);
+
+	zassert_equal(0, ret,
+		      "return should be 0");
+	zassert_equal(NRF_CLOUD_FOTA_BL_STATUS_S0_FLAG_SET, job.bl_flags,
+		      "bl_flags should indicate s0 set and s0 inactive");
+}
+
+/* Verify that nrf_cloud_pending_fota_job_process fails for an unknown FOTA job type. */
+ZTEST(nrf_cloud_fota_common_test, test_nrf_cloud_pending_fota_job_process_unknwn_type)
```
tony-le-24, line 223: is this intended or typo? `test_nrf_cloud_pending_fota_job_process_**unknwn**_type`?

tony-le-24 (approved):
LGTM!

tests/subsys/net/lib/nrf_cloud/fota_common/src/main.c
jayteemo: probably intended, but not necessary... i'll add the `o` 😅

PR #11739, opened 2023-07-07
net: nrf_provisioning: Add platform

juhaylinen:
Add 9160dk to allowed platforms.


plskeggs:
See Teams comment from today.

plskeggs (approved):

PR #11740, opened 2023-07-07
doc: nrf_security name change

mia-ko:
Changing the name of Nordic Security Module
to remove the module reference
(as it is a subsystem and not actually a module)

### DIFF
--- a/doc/nrf/libraries/nrf_security/doc/drivers.rst
+++ b/doc/nrf/libraries/nrf_security/doc/drivers.rst
@@ -1,13 +1,13 @@
 .. _nrf_security_drivers:
 
-nrf_security drivers
+nRF Security drivers
 ####################
 
 .. contents::
    :local:
    :depth: 2
 
-The nrf_security module supports multiple enabled PSA drivers at the same time.
+The nRF Security subsystem supports multiple enabled PSA drivers at the same time.
 This mechanism is intended to extend the available feature set of hardware-accelerated cryptography or to provide alternative implementations of the PSA Crypto APIs.
 
 You can enable a cryptographic feature or algorithm using PSA Crypto API configurations that follow the format ``PSA_WANT_ALG_XXXX``.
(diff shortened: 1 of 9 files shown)


doc/nrf/libraries/nrf_security/doc/drivers.rst
```diff
@@ -16,7 +16,7 @@ Enabling more than one PSA driver might add support for additional key sizes or
 
 It is possible to disable specific features on the PSA driver level to optimize the code size.
 
-The nrf_security module supports the following PSA drivers:
+The nrf_security supports the following PSA drivers:
```
joerchan, line 19:
```suggestion
The nRF Security subsystem supports the following PSA drivers:
```

doc/nrf/libraries/nrf_security/doc/drivers.rst
```diff
@@ -7,7 +7,7 @@ nrf_security drivers
    :local:
    :depth: 2
 
-The nrf_security module supports multiple enabled PSA drivers at the same time.
+The nrf_security supports multiple enabled PSA drivers at the same time.
```
joerchan, line 10:
```suggestion
The nRF Security subsystem supports multiple enabled PSA drivers at the same time.
```

subsys/nrf_security/Kconfig
```diff
@@ -32,13 +32,13 @@ config NORDIC_SECURITY_BACKEND
 
 config NRF_SECURITY
 	bool
-	prompt "Use nrf_security module" if !PSA_PROMPTLESS
+	prompt "Use nrf_security" if !PSA_PROMPTLESS
```
joerchan, line 35: I'm wondering if we should change the wording to "Enable" here.

```suggestion
	prompt "Enable nRF Security" if !PSA_PROMPTLESS
```

joerchan (changes requested):
I don't see the need to use "nrf_security" in the documentation, I suggest we get rid of it.

mswarowsky (approved):
With the changes suggested by @joerchan lgtm

subsys/nrf_security/Kconfig
mia-ko: Sure, it's more in line with other config options as well.

joerchan (approved):

melwee01 (approved):

PR #11748, opened 2023-07-07
wifi: Fix clock include

krish2718:
Build fixes.


muraliThokala (approved):

VivekUppunda (approved):

samples/wifi/scan/src/main.c
```diff
@@ -217,7 +219,7 @@ int main(void)
 
 	net_mgmt_add_event_callback(&wifi_shell_mgmt_cb);
 
-#ifdef CLOCK_FEATURE_HFCLK_DIVIDE_PRESENT
+#if defined(CLOCK_FEATURE_HFCLK_DIVIDE_PRESENT) || NRF_CLOCK_HAS_HFCLK192M
```
ajayparida, line 222: Better to use
if defined(CLOCK_FEATURE_HFCLK_DIVIDE_PRESENT) || defined(NRF_CLOCK_HAS_HFCLK192M)
Apllicable for rest of the files in this PR.

krish2718: Oops, its merged, but this is the way its used elsewhere (See: `modules/hal/nordic_haltium/nrfx/drivers/include/nrfx_clock.h`), don't really know why `defined` is skipped though.

bama-nordic: I really think `defined` should not be skipped - @krish2718 maybe you can raise a PR with this correction?

krish2718: Wanted to keep the same conditions across NCS.

krish2718: Here's the reason, it shouldn't be just `defined()` as the condition should only be included if its `1`, see https://github.com/zephyrproject-rtos/hal_nordic/blob/master/nrfx/hal/nrf_clock.h#L76

PR #11750, opened 2023-07-08
Llpm sample connect directly to 1ms

martintv:
samples: bluetooth: Connect directly to 1ms in llpm sample
    
This improves speed of setting up encryption and doing service discovery by alot.

doing the minimum now in this PR, we can probably also get rid of the vs hci command  (SDC_HCI_OPCODE_CMD_VS_CONN_UPDATE) and use the zephyr funciton to change connection interval now. but that will be in a later PR

### DIFF
--- a/samples/bluetooth/llpm/README.rst
+++ b/samples/bluetooth/llpm/README.rst
@@ -106,33 +106,32 @@ After programming the sample to both development kits, test it by performing the
 
    - The central outputs the following information::
 
-       Press any key to set LLPM short connection connection interval (1 ms)
+       Press any key to start measuring transmission latency
 
    - The peripheral outputs the following information::
 
        Press any key to start measuring transmission latency
 
 #. Press a key in the terminal that is connected to the peripheral.
-#. Observe the terminal connected to the peripheral. The latency measurements are printed in the terminal.
-   The latency is expected to be shorter than the default connection interval::
+#. Observe the terminal connected to the peripheral.
+   The latency measurements are printed in the terminal.
+   The latency is expected to be around 1 ms.
 
-       Transmission Latency: 80917 (us), CRC mismatches: 0
+       Transmission Latency: 1083 (us), CRC mismatches: 0
 
 #. Press a key in the terminal that is connected to the central.
-#. Observe the connection gets updated to LLPM connection interval (1 ms) on both sides::
-
-       Connection interval updated: LLPM (1 ms)
-
 #. Observe the terminal connected to the peripheral.
    The measured latency on the peripheral becomes approximate 1 ms::
 
        Transmission Latency: 1098 (us), CRC mismatches: 0
 
-#. Press a key in the terminal that is connected to the central.
-#. Observe the terminal connected to the central.
-   The measured latency on the central remains approximate 1 ms::
+#. Observe the central switches to standard 7.5 ms interval right after performing latency measurements.
 
-       Transmission Latency: 1235 (us), CRC mismatches: 0
+#. Observe the terminal connected to the central.
+   The latency is higher now.
+
+   Connection interval updated: 7.5 ms
+   Transmission Latency: 4592 (us), CRC mismatches: 0
 
 .. msc::
    hscale = "1.3";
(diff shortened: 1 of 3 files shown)


alwa-nordic (approved):

samples/bluetooth/llpm/src/main.c
```diff
@@ -367,25 +369,13 @@ static const struct bt_latency_client_cb latency_client_cb = {
...
-		printk("Press any key to set LLPM short connection interval (1 ms)\n");
-		console_getchar();
-
-		if (enable_llpm_short_connection_interval()) {
-			printk("Enable LLPM short connection interval failed\n");
-			return;
-		}
-	}
-
 	printk("Press any key to start measuring transmission latency\n");
```
KAGA164, line 379: Please update also sample documentation with changes here

samples/bluetooth/llpm/src/main.c
```diff
@@ -409,6 +399,24 @@ static void test_run(void)
 		}
 
 		memset(&llpm_latency, 0, sizeof(llpm_latency));
+		if (conn_info.role == BT_CONN_ROLE_CENTRAL) {
+			static int counter;
+			static uint16_t interval_us = INTERVAL_LLPM_US;
+
+			counter++;
+			if (counter == 20) {
```
KAGA164, line 407: Why 20? Could you explain in comment or use define instead of magic number?

samples/bluetooth/llpm/src/main.c
```diff
@@ -230,6 +229,9 @@ static void le_param_updated(struct bt_conn *conn, uint16_t interval,
 {
 	if (interval == INTERVAL_LLPM) {
 		printk("Connection interval updated: LLPM (1 ms)\n");
+	} else {
+		__ASSERT_NO_MSG(interval == 6);
```
KAGA164, line 233: Maybe you could keep this define `#define INTERVAL_MIN    0x6` and use it here

KAGA164 (changes requested):
Please update the sample doc and fix typos in the commit message

samples/bluetooth/llpm/src/main.c
martintv: done

samples/bluetooth/llpm/src/main.c
martintv: addded comment, but open t changing it some other way. like not doing this switching between intervals even.

samples/bluetooth/llpm/src/main.c
martintv: done

KAGA164 (approved):

jori-nordic (approved):

samples/bluetooth/llpm/README.rst
```diff
@@ -106,33 +106,29 @@ After programming the sample to both development kits, test it by performing the
...
-
 #. Observe the terminal connected to the peripheral.
    The measured latency on the peripheral becomes approximate 1 ms::
 
        Transmission Latency: 1098 (us), CRC mismatches: 0
 
-#. Press a key in the terminal that is connected to the central.
-#. Observe the terminal connected to the central.
-   The measured latency on the central remains approximate 1 ms::
+#. After the central has done a few measurments it switches to standard 7.5ms interval
```
annakielar, line 127: 1. Is the second part of this sentence dependent on the results of these measurements? Or does it happen always after doing measurements? What kind of measurements are done here? "A few measurements" is too vague.

2. The first sentence in each step should contain only actions for end users, hence this sentence should be rephrased. Is this something that can be observed/checked/verified by end user?
For example:
```suggestion
#. Observe the central switches to standard 7.5 ms interval right after performing latency measurements.

```

samples/bluetooth/llpm/README.rst
```diff
@@ -106,33 +106,29 @@ After programming the sample to both development kits, test it by performing the
...
+#. Observe the terminal connected to the central, that the latency is now higher.
```
annakielar, line 128:
```suggestion
#. Observe the terminal connected to the central.
   The latency is higher now.
```

annakielar (changes requested):

samples/bluetooth/llpm/README.rst
martintv: it is hardcoded to 20, but I didn't want to write that in the doc so that we have to worry about keeping doc and implementation in sync. The switch is not dependent on anything no. It does it regardless of meansurment results.

annakielar (approved):

rlubos:
There's no more manifest changes in the PR, removed DNM.

PR #11758, opened 2023-07-10
lib: modem_antenna: add support for MIPI RFFE configuration

bjda:
Adds Kconfig entries and logic to support the external MIPI RFFE interface.

### DIFF
--- a/lib/modem_antenna/Kconfig
+++ b/lib/modem_antenna/Kconfig
@@ -41,6 +41,36 @@ config MODEM_ANTENNA_AT_COEX0
 	  Defines the AT%XCOEX0 command to be sent to the modem. Leave
 	  empty if this command should not be sent.
 
+config MODEM_ANTENNA_AT_MIPIRFFEDEV
+	string "AT%XMIPIRFFEDEV command"
+	help
+	  Defines the AT%XMIPIRFFEDEV command to be sent to the modem. Leave
+	  empty if this command should not be sent.
+
+config MODEM_ANTENNA_AT_MIPIRFFECTRL_INIT
+	string "AT%XMIPIRFFECTRL command (INIT)"
+	help
+	  Defines the AT%XMIPIRFFECTRL command to be sent to the modem for INIT phase.
+	  Leave empty if this command should not be sent.
+
+config MODEM_ANTENNA_AT_MIPIRFFECTRL_ON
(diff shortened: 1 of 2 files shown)


lib/modem_antenna/Kconfig
maxd-nordic, line 1: Note: I think we should set `MODEM_ANTENNA_AT_MAGPIO` and such in the board files and not here as a default anymore.

bjda: Agreed

tokangas (approved):

lib/modem_antenna/Kconfig
tokangas: Are you planning to do something for this? Just wondering how that works in practice, because the default values for `MODEM_ANTENNA_AT_MAGPIO` and `MODEM_ANTENNA_AT_COEX0` depend also on the selected antenna (`MODEM_ANTENNA_GNSS_ONBOARD`/`MODEM_ANTENNA_GNSS_EXTERNAL`).

bjda: For a new branch I have a section like this in Kconfig.defconfig:

```KConfig
config MODEM_ANTENNA
	depends on NRF_MODEM_LIB
	default y

if MODEM_ANTENNA

config MODEM_ANTENNA_AT_MIPIRFFEDEV
	default "AT%XMIPIRFFEDEV=1,4,71,198,248"
config MODEM_ANTENNA_AT_MIPIRFFECTRL_INIT
	default "AT%XMIPIRFFECTRL=1,0,1,28,248"
config MODEM_ANTENNA_AT_MIPIRFFECTRL_ON
	default "AT%XMIPIRFFECTRL=1,1,1,28,56,13,0,0,8,8,715,4,4,770,12,12,829,11,11,863,130,130,892,1,1,939,129,129,978,26,26,1042,8,8,1118,4,4,1270,12,12,1386,14,14,1523,130,130,2200"
config MODEM_ANTENNA_AT_MIPIRFFECTRL_OFF
	default "AT%XMIPIRFFECTRL=1,2,1,28,184"
config MODEM_ANTENNA_AT_MIPIRFFECTRL_PWROFF
	default "AT%XMIPIRFFECTRL=1,3,1,28,184"

endif # MODEM_ANTENNA
```

In this case internal/external GPS does not matter. If it did, conditions could be added for each default.

jhn-nordic (approved):

PR #11760, opened 2023-07-10
treewide: Replace non-standard path values with module variables

nordicjm:
Fixes instances of things like ${ZEPHYR_BASE}../some/path with actual variable names which are generated by Zephyr pointing to the module paths automatically.


SebastianBoe (approved):
+1 assuming we get a green CI.

But I think we will not get a green CI.

AFAIK ZEPHYR_*_MODULE_DIR vars are not available to the TF-M build system and it looks like we are adding this symbol to some TF-M build scripts. But if CI goes green then I am mistaken.

nordicjm:
> +1 assuming we get a green CI.
> 
> But I think we will not get a green CI.
> 
> AFAIK ZEPHYR_*_MODULE_DIR vars are not available to the TF-M build system and it looks like we are adding this symbol to some TF-M build scripts. But if CI goes green then I am mistaken.

The base build task did run and complete when I had this in draft, though I cancelled the further downstream jobs.

joerchan:
@nordicjm @SebastianBoe This should run fine, the argument passed to the TF-M build system is also renamed.

joerchan (approved):

maxd-nordic (approved):

MarekPieta (approved):

carlescufi:
scancode check is irrelevant, it's just a CMakeLists file.

PR #11761, opened 2023-07-10
doc: nsib: add mention of static partitioning req

greg-fer:
Added mentions of static partitioning being required for NSIB. Edited the structure of bootloader_adding.rst to make it clearer. NCSIDB-964.


greg-fer:
@umapraseeda , FYI.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -117,8 +117,10 @@ The upgradable bootloader carries out the same signature and metadata verificati
 Also, it can upgrade both itself and the following image in the boot sequence, which, in most cases, is an application.
 
 .. caution::
-   You should add a second-stage bootloader only when necessary by the design or firmware upgrade needs.
-   Adding the second stage bootloader for no reason will lead to a degradation of the system's overall security, as attackers can exploit bugs that may exist in either bootloader.
+   * You should add a second-stage bootloader only when necessary by the design or firmware upgrade needs.
+     Adding the second stage bootloader for no reason will lead to a degradation of the system's overall security, as attackers can exploit bugs that may exist in either bootloader.
+   * When adding a second-stage bootloader, make sure to define :ref:`static partition for the immutable bootloader <ug_pm_static>`.
```
hellesvik-nordic, line 122: Static partitioning is not only needed for second-stage bootloaders, but also for single-stage bootloaders.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
```diff
@@ -286,35 +296,41 @@ The following section describes how to add it to your secure bootloader chain.
...
 ==========================================
 
-To use MCUboot as an upgradable bootloader, the application must already use the |NSIB| as the immutable bootloader.
-MCUboot is added to its boot chain by including the :kconfig:option:`CONFIG_BOOTLOADER_MCUBOOT` Kconfig option with either the build command or in the application's :file:`prj.conf` file:
+To use MCUboot as an upgradable bootloader to your application, complete the following steps:
 
-.. code-block::
+1. :ref:`Add |NSIB| as the immutable bootloader <ug_bootloader_adding_immutable_b0>`.
+#. Define :ref:`static partition for the immutable bootloader <ug_pm_static>`.
+   This is important because adding an upgradable bootloader to an application changes its :ref:`flash memory partition layout <ug_bootloader_flash_mcuboot>` to accommodate first and second image slots for both immutable and upgradable bootloaders.
```
hellesvik-nordic, line 303: You do not have to use a static partition for developement, it becomes important for production.
Also, the first build you do, you can use dynamic partitioning. As long as all subsequent builds use the same partitioning as was dynamically created.
Dynamic partitioning is so much easier that I  generally recommend starting by doing this, to see how the system generated the partitioning. Then you can create static partitioning from what you learned.
Can I suggest that we mention static partition later on in this list instead?

MarekPieta: Maybe we could mention that user can play around with dynamic partitions as long as DFU is not used. We need to keep in mind that the dynamic partitions may not be backwards compatible and upgrading firmware may lead to strange issues when dynamic partition maps are not interoperable.

samples/bootloader/README.rst
```diff
@@ -167,8 +168,8 @@ This would result, for example, in a flash memory layout like the following, whe
 Pre-signed variants
 -------------------
 
-When two slots are present, two images must be built.
-One that is executable from slot 0, and the other one from slot 1.
+When two slots are present, either for the application or for a second-stage bootloader, two images must be built.
+One that is executable from slot 0, and the other one from slot 1, both of which must have :ref:`static partition defined <ug_pm_static>`.
```
MarekPieta, line 172: Maybe we keep the common part in `_ug_bootloader` and link there for information about architecture? User should be aware of the basics anyway.

scripts/partition_manager/partition_manager.rst
```diff
@@ -717,11 +717,14 @@ The sizes of each partition are determined by the associated :file:`pm.yml` file
 Static configuration
 ********************
 
+The Partition Manager provides static configuration that lets you define static partitions.
+This feature can be used with single-image applications, but its most common use case is with multi-image ones.
```
MarekPieta, line 721: The static configuration can be used in general when Partition Manager manages memory partitions. I would not go deeper into details here (e.g. Partition Manager for single-image applications requires additional configuration and it's not used by default).

scripts/partition_manager/partition_manager.rst
```diff
@@ -717,11 +717,14 @@ The sizes of each partition are determined by the associated :file:`pm.yml` file
...
+
 By default, the Partition Manager dynamically places the partitions in memory.
-However, if you have a deployed product that consists of multiple images, where only a subset of the included images can be upgraded through a firmware update mechanism, the upgradable images must be statically configured.
+However, there are cases where a deployed product can consist of multiple images, and only a subset of these images can be upgraded through a firmware update mechanism.
+In such cases, you must statically configure the upgradable images using the Partition Manager's static configuration.
```
MarekPieta, line 725: `you must rely on static configuration of partitions to ensure that all of the built images are compatible.`

samples/bootloader/README.rst
MarekPieta: We could also inform there that user must statically define all of the partitions used in a given application configuration.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
hellesvik-nordic: > Maybe we could mention that user can play around with dynamic partitions as long as DFU is not used.

I think this one points to the important difference between DFU and bootloaders.
And as you say, this limitation is only relevant for DFU.
Therefore, it could be argued that we do not need to mention static partitioning here, but leave it for a "how to enable DFU" section? 
However, I think I like it being mentioned here still.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
```diff
@@ -286,35 +296,41 @@ The following section describes how to add it to your secure bootloader chain.
 Adding MCUboot as an upgradable bootloader
 ==========================================
 
-To use MCUboot as an upgradable bootloader, the application must already use the |NSIB| as the immutable bootloader.
-MCUboot is added to its boot chain by including the :kconfig:option:`CONFIG_BOOTLOADER_MCUBOOT` Kconfig option with either the build command or in the application's :file:`prj.conf` file:
+To use MCUboot as an upgradable bootloader to your application, complete the following steps:
 
-.. code-block::
+1. :ref:`Add |NSIB| as the immutable bootloader <ug_bootloader_adding_immutable_b0>`.
+#. Define :ref:`static partition for the immutable bootloader <ug_pm_static>`.
```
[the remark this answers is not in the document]
hellesvik-nordic: > user may even generate dynamic partition map for use-case with 2 bootloaders and use it as a base for the static partition map

For what its worth: I think this is the easiest way to get started with static partitioning

samples/bootloader/README.rst
hellesvik-nordic: > We could also inform there that user must statically define all of the partitions used in a given application configuration.

Not all partitions. For upgradable bootloaders it goes like this:

B0: should be statically partitioned on release
S0: should be statically partitioned on release
S1: should be statically partitioned on release
mcuboot_primary: can be dymanically partitioned
mcuboot_secondary: can be dymanically partitioned

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
MarekPieta: In theory yes. But I wonder if there are practical use-cases of using bootloader without DFU mechanism (apart from simple local checks on desk). That's why I was wondering about simplifying this part.

samples/bootloader/README.rst
MarekPieta: You are right. I slightly oversimplified it here.
Still, the application must be interoperable with the second stage bootloader so that partition maps for the second stage bootloader and application must remain consistent.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
greg-fer: Let's keep this info next to where we mention flash memory partitions.

greg-fer:
@hellesvik-nordic , @MarekPieta , missed some comments. Pushed latest updates only now.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
hellesvik-nordic: > Do we have use-cases where bootloader without DFU mechanism is used?
Not the most widely used, but for devices without a KMU, the [HUK ](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/keys/hw_unique_key/README.html#overview )is protected by a bootloader before the application runs:
However, [PSA Certified Level 1](https://www.psacertified.org/app/uploads/2022/06/JSADEN001-PSA_Certified_Level_1-2.2-REL-01.pdf) requires upgradability, so using HUK without PSA Certifiication seems quite niche.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -96,14 +97,17 @@ The verification provided by this bootloader is recommended and suitable for all
 * Metadata verification - Checks that the images are compatible.
 
 .. caution::
-   You must generate and use your own signing keys while in development and before deploying when using either MCUboot or the |NSIB| as an immutable bootloader.
-   See :ref:`ug_fw_update_development_keys` for more information.
+   * You must generate and use your own signing keys while in development and before deploying when using either MCUboot or the |NSIB| as an immutable bootloader.
+     See :ref:`ug_fw_update_development_keys` for more information.
+   * Make sure to define :ref:`static partition for the immutable bootloader <ug_pm_static>` for production devices.
+     Failing to do so will keep the dynamic allocation of partitions in memory, which can cause overwriting of the memory slots and can irreversibly break the device firmware update procedure.
```
MarekPieta, line 103: Maybe we could link to complete explanation described below instead of using this sentence to prevent confusion (_ug_bootloader_flash_static_requirement)?

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -186,3 +190,17 @@ For example, the partition layout for :file:`zephyr/samples/hello_world` using M
...
+Static partition requirement for DFU
+------------------------------------
+
+As long as you are not using Device Firmware Updates (DFU), you can use the dynamic allocation of memory partitions for either immutable or upgradable bootloaders.
+
+However, if you want to perform DFU, you must :ref:`define a static partition map <ug_pm_static>`.
+This is important for different reasons, including:
+
+* If the dynamically generated partitions change between builds, the newly generated application images may no longer be compatible with the bootloader programmed to device.
+* Adding an upgradable bootloader to an application changes its :ref:`flash memory partition layout <ug_bootloader_flash_mcuboot>` to accommodate first and second image slots for both immutable and upgradable bootloaders.
```
MarekPieta, line 205: Maybe we could explain this in detail in the first point:
`If the dynamically generated partitions change between builds, the newly generated application images may no longer be compatible with the bootloader programmed to device.  In that case, the newly build application image may not be bootable by the bootloader programmed onto the device.`

MarekPieta: In that case we may no longer need to keep 2nd and 3rd bullet points (they refer to the same root cause of the issue).

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
```diff
@@ -59,10 +59,18 @@ To ensure that the immutable bootloader occupies as little flash memory as possi
 
 See :ref:`ug_bootloader_config` for more information about using Kconfig fragments.
 
+.. caution::
+   If you want to perform DFU, remember to :ref:`define a static partition map <ug_pm_static>` for the immutable bootloader.
```
MarekPieta, line 63: Maybe we can keep the note right under `Adding a bootloader chain`?
It's a common requirement no matter which bootloader is going to be used.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
```diff
@@ -223,10 +231,23 @@ To build :doc:`mcuboot:index-ncs` with a Zephyr or |NCS| sample, enable the :kco
...
+----------------------------------------------
+
+The following sections describe different configuration options available for MCUboot as an immutable bootloader.
+
+.. _ug_bootloader_adding_partitioning:
+
+Generating partition map
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you plan to use DFU, you can generate a dynamic partition map at this stage and use it as a base for the :ref:`ug_bootloader_flash_static_requirement` for the production device.
```
MarekPieta, line 244: You can follow the same tip for other bootloaders.

samples/bootloader/README.rst
```diff
@@ -49,7 +49,8 @@ The NSIB implements a simple and reliable :term:`Root of Trust (RoT)` for a secu
 
    Each slot has a version number, and the bootloader will select the slot with the latest version.
 
-   For more information about creating a second stage bootloader, see :ref:`ug_bootloader_adding_upgradable`.
+   :ref:`Static configuration for defining static partitions <ug_pm_static>` must be used when using DFU.
```
MarekPieta, line 52: Should it actually belong here? The section seems to focus on how bootloader works, not on configuration

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -96,14 +97,17 @@ The verification provided by this bootloader is recommended and suitable for all
 * Metadata verification - Checks that the images are compatible.
 
 .. caution::
-   You must generate and use your own signing keys while in development and before deploying when using either MCUboot or the |NSIB| as an immutable bootloader.
-   See :ref:`ug_fw_update_development_keys` for more information.
+   * You must generate and use your own signing keys while in development and before deploying when using either MCUboot or the |NSIB| as an immutable bootloader.
```
MarekPieta, line 100: btw. This part contains generic information about bootloaders in the NCS. Maybe we can make it part of a generic introduction ?

samples/bootloader/README.rst
```diff
@@ -183,6 +184,12 @@ When using the ``nrf52840dk_nrf52840`` board, this would produce a flash memory
 
    B0 flash memory layout with MCUboot
 
+.. note::
+   When using DFU, and especially for production-ready devices, these parititions must be configured in the following way to meet :ref:`ug_bootloader_flash_static_requirement`:
+
+   * *B0*, *S0*, *S1* - These slots must be statically partitioned.
+   * *mcuboot_primary* and *mcuboot_secondary* - These slots can be dymanically partitioned.
```
MarekPieta, line 191: This is a bit tricky, because we still must ensure compatibility of the application with partitions known to MCUboot bootloader

hellesvik-nordic: How about 
`These slots can be dymanically partitioned. Keep in mind that MCUboot and application knowledge of partitions must match.` ?

samples/bootloader/README.rst
greg-fer: Moved to Configuration.

samples/bootloader/README.rst
greg-fer: Moved to configuration.

greg-fer:
@hellesvik-nordic , @MarekPieta , please have another look. As mentioned on Teams, bigger changes will be tackled in separate tasks.

MarekPieta (approved):

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -65,23 +65,43 @@ There are two implementations currently supported:
...
+.. _ug_bootloader_flash_static_requirement:
+
+Static partition requirement for DFU
+------------------------------------
+
+By default, the Partition Manager generates the partition map dynamically.
+As long as you are not using Device Firmware Updates (DFU), you can use the dynamic generation of memory partitions.
+
+However, if you want to perform DFU, you must :ref:`define a static partition map <ug_pm_static>`.
+This is important because the dynamically generated partitions change between builds.
```
hellesvik-nordic, line 82: Add can, as it is not certain?
`generated partitions can change between builds`

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -65,23 +65,43 @@ There are two implementations currently supported:
...
+The newly generated application images may no longer be compatible with the bootloader programmed to the device.
```
hellesvik-nordic, line 83: Can I suggest that we specify that it is partitions that are not compatible?
'The newly generated application images may no longer have partitions which are compatible with the bootloader partitions.'
I was not able to phrase this sentence in a way that keeps the ' programmed to the device' part

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -65,23 +65,43 @@ There are two implementations currently supported:
...
+   You can use this file as a base for your static partition map.
+
+The memory partitions that must be defined in the static partition map depend on the selected bootloader chain.
+For details, see :ref:`ug_bootloader_flash`.
 
 .. _immutable_bootloader:
 
-First-stage immutable bootloader
-================================
+Immutable bootloader
```
hellesvik-nordic, line 98: This now looks like a sub-header of "Static partition requirement for DFU".
And I beleive that it is its own section right?

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -103,7 +123,7 @@ The :ref:`bootloader capabilities table <app_bootloaders_support_table>` indicat
 
 .. _upgradable_bootloader:
 
-Second-stage Upgradable bootloader
+Second-stage upgradable bootloader
```
hellesvik-nordic, line 126: This now looks like a sub-header of "Static partition requirement for DFU".
And I beleive that it is its own section right?

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
```diff
@@ -7,7 +7,7 @@ Adding a bootloader chain
    :local:
    :depth: 2
 
-You can add bootloader chains to an application in the following ways:
+You can add bootloader chain to an application in the following ways:
```
hellesvik-nordic, line 10: a chain?

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
```diff
@@ -286,35 +297,39 @@ The following section describes how to add it to your secure bootloader chain.
...
+      -DCONFIG_SECURE_BOOT=y \
+      -DCONFIG_BOOTLOADER_MCUBOOT=y
 
-MCUboot can use the cryptographic functionality exposed by the immutable bootloader, reducing the flash memory usage for MCUboot to less than 16 kB.
-To enable this configuration, apply both the :file:`prj_minimal.conf` Kconfig project file and the :file:`external_crypto.conf` Kconfig fragment for the MCUboot image:
+   |how_to_configure|
 
-.. code-block::
+#. Optionally, you can configure MCUboot to use the cryptographic functionality exposed by the immutable bootloader and reduce the flash memory usage for MCUboot to less than 16 kB.
+   To enable this configuration, apply both the :file:`prj_minimal.conf` Kconfig project file and the :file:`external_crypto.conf` Kconfig fragment for the MCUboot image:
```
hellesvik-nordic, line 314: Maybe we should mention that users in this case can use [CONFIG_PM_PARTITION_SIZE_MCUBOOT](https://github.com/nrfconnect/sdk-nrf/blob/d5eb58b3ce993e460e18b0119a19a4e7b70ec126/modules/mcuboot/boot/zephyr/Kconfig#L21) on the child image to redcue the partition size to match?

hellesvik-nordic: And yes, this configuration is not to find in Kconfig search, as MCUboot is an "application" and not part of the Kconfig system as such. Have been discussed in tech-pubs teams before I think

hellesvik-nordic: We should mention [CONFIG_PM_PARTITION_SIZE_MCUBOOT](https://github.com/nrfconnect/sdk-nrf/blob/d5eb58b3ce993e460e18b0119a19a4e7b70ec126/modules/mcuboot/boot/zephyr/Kconfig#L21) in the "Customizing the bootloader" section as well, as this configuration is not mentioned once in the docs. [Search tests](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/search.html?q=+CONFIG_PM_PARTITION_SIZE_MCUBOOT).
But that is maybe for another PR?

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
MarekPieta: btw. The static partition map is also important if we program a precompiled hex as a child image instead of building it from source (yesterday evening I found this use-case)

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
MarekPieta: Maybe:
`The newly generated application images may no longer use partition map which is compatible with the partition map used by the bootloader programmed to the device.`?

scripts/partition_manager/partition_manager.rst
```diff
@@ -718,10 +718,10 @@ Static configuration
 ********************
 
 By default, the Partition Manager dynamically places the partitions in memory.
-However, if you have a deployed product that consists of multiple images, where only a subset of the included images can be upgraded through a firmware update mechanism, the upgradable images must be statically configured.
```
hellesvik-nordic, line 721: Here we remove the general statement ` the upgradable images must be statically configured`, and leave the reader to infer the limitation from the example following this.
I think we should still give a statement that in such a case, upgradable images must have partitions that are static/matches the bootloader.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
MarekPieta: The `CONFIG_PM_PARTITON_SIZE_MCUBOOT` is used only if we use dynamic partition map. Maybe we could describe this possibility in Partition Manager docs and link there.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
hellesvik-nordic: I had not thought about that, good point

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
hellesvik-nordic: I agree that using "partition map" is good here.
However, keeping the `programmed to the device` at the end makes the sentence quite "heavy". Is it nessecary to specify that the bootloader is already programmed, or can we expect readers to undestand that from context?
Nitpick, add "a": `no longer use a partition map`

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
hellesvik-nordic: Agreed.
Partition Manager docs should definetly meantion CONFIG_PM_PARTITION_SIZE_<CHILD_IMAGE> and list the common variants of this

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
greg-fer: It is not a sub-header of Static partition requirement, because it has higher-level header formatting === vs lower-level --- of that section.

hellesvik-nordic: But it looks smaller:
bilde

greg-fer:
Converting to draft given the amount of comments.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -65,23 +65,43 @@ There are two implementations currently supported:
...
+By default, the Partition Manager generates the partition map dynamically.
+As long as you are not using Device Firmware Updates (DFU), you can use the dynamic generation of memory partitions.
+
+However, if you want to perform DFU, you must :ref:`define a static partition map <ug_pm_static>`.
+This is important because the dynamically generated partitions change between builds.
+The newly generated application images may no longer be compatible with the bootloader programmed to the device.
+In such case, the newly built application image may not be bootable by the bootloader programmed onto the device.
 
 .. note::
+   For detailed information about the memory layout used for the build, see the partition configuration in the :file:`partitions.yml` file, located in the build folder directory, or run ``ninja partition_manager_report``.
```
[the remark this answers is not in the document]
greg-fer: Will be done in https://nordicsemi.atlassian.net/browse/NCSDK-22712

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
greg-fer: There was an issue with the Architecture heading, actually. Fixed.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
greg-fer: Fixed as mentioned above.

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
greg-fer: https://nordicsemi.atlassian.net/browse/NCSDK-22733

scripts/partition_manager/partition_manager.rst
greg-fer: New phrasing:
> By default, the Partition Manager dynamically places the partitions in memory.
However, there are cases where a deployed product can consist of multiple images, and only a subset of these images can be upgraded through a firmware update mechanism.
In such cases, the upgradable images must have partitions that are static or are matching the bootloader partition map.
For example, if a device includes a non-upgradable first-stage bootloader and an upgradable application, the application image to be upgraded must be linked to the same address as the one that is deployed.
For this purpose, the Partition Manager provides the static configuration to define static partitions.

hellesvik-nordic: How about `In such cases, the upgradable images must have partitions that are static in order to match the bootloaders partition map.`
?

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -65,23 +65,43 @@ There are two implementations currently supported:
   .. figure:: images/bootloader_memory_layout.svg
      :alt: Memory layout
 
-For detailed information about the memory layout, see the partition configuration in the :file:`partitions.yml` file, located in the build folder directory, or run ``ninja partition_manager_report``.
+By default, building an application with any bootloader configuration creates a :ref:`multi-image build <ug_multi_image>`, where the :ref:`partition_manager` manages its memory partitions.
+Bootloaders are built as child images.
```
hellesvik-nordic, line 69: If we are to be very correct, the sentence `Bootloaders are built as child images.` as a standalone is not technically true.
However, if we see it in the contect of the preceding sentence, I think it is fine.
Maybe refer to make sentences more connected?
`In this case, bootloaders are built as child images.`

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader.rst
```diff
@@ -65,23 +65,43 @@ There are two implementations currently supported:
...
+When building an application with :ref:`Cortex-M Security Extensions (CMSE) enabled <app_boards_spe_nspe_cpuapp_ns>`, then :ref:`Trusted Firmware-M (TF-M) <ug_tfm>` is also built with the image automatically.
```
hellesvik-nordic, line 70: Remove "also" from "is also built with the image automatically."
This does not refer to previous sentence

scripts/partition_manager/partition_manager.rst
```diff
@@ -718,10 +718,11 @@ Static configuration
...
+However, there are cases where a deployed product can consist of multiple images, and only a subset of these images can be upgraded through a firmware update mechanism.
+In such cases, the upgradable images must have partitions that are static or are matching the partition map used bootloader programmed onto the device.
```
hellesvik-nordic, line 722: `are static or are matching` : It does not help if they are static (pm_static.yml) if you change the pm_static.yml in between the builds. Can I suggest that we use "and" `are static and are matching`.
`partition map used bootloader` ---> `partition map used by the bootloader`

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
```diff
@@ -286,35 +297,39 @@ The following section describes how to add it to your secure bootloader chain.
...
 
-   west build -b nrf52840dk_nrf52840 zephyr/samples/hello_world -- \
-   -DCONFIG_BOOTLOADER_MCUBOOT=y \
-   -DCONFIG_SECURE_BOOT=y \
-   -Dmcuboot_CONF_FILE=prj_minimal.conf \
-   -Dmcuboot_OVERLAY_CONFIG=external_crypto.conf
+   .. code-block::
+
+      west build -b nrf52840dk_nrf52840 zephyr/samples/hello_world -- \
```
hellesvik-nordic, line 318: Testing this command with v2.4.0, I get the error
```
CMake Error at C:/ncs/v2.4.0/nrf/modules/mcuboot/CMakeLists.txt:292 (file):
  file STRINGS file "C:/ncs/v2.4.0/nrf/modules/mcuboot/prj_minimal.conf"
  cannot be read.
```
I also got a colleguea to test, and they say the same

greg-fer: I haven't touched this command, just changed formatting style around it. Can you open a documentation bug for the developer who added it? It's a "Won't Fix" for this PR's scope for me.

hellesvik-nordic: Created https://nordicsemi.atlassian.net/browse/NCSIDB-1056

hellesvik-nordic (approved):

doc/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.rst
kittydepa: Attempt to address these points here: https://github.com/nrfconnect/sdk-nrf/pull/15018

PR #11770, opened 2023-07-11
applications: connectivity_bridge: Redirect logs to RTT

maxd-nordic:
This patch makes sure that logs are definitely routed to RTT. Previously, BLE logs would be routed towards the nRF91, causing parsing errors.


bjda (approved):
Makes sense to me

jhn-nordic (approved):

PR #11777, opened 2023-07-11
samples: matter: remove af-gen-event.h

Damian-Nordic:
The file was not auto-generated and had to be created manually but it is not longer needed in the current version of Matter.


LuDuda (approved):

Damian-Nordic:
@doki-nordic The license checker was rejecting my PR as it only removed files. Could you please take a look at the script change?

kkasperczyk-no (approved):

PR #11780, opened 2023-07-11
libs: keys: Use logging instead of printk.

joerchan:
Use logging instead of printk in keys libraries.


SebastianBoe (approved):

mswarowsky (approved):
Nice usability change

PR #11798, opened 2023-07-13
doc: nrf5340_audio: split docs into multiple pages

greg-fer:
Split the application documentation into multiple pages. OCT-2610.


greg-fer:
@gmarull , in this PR, I'd like to use `options-from-kconfig` in a doc subdirectory rather than in a README directly next to the .kconfig file.

Is there an option to define the `rel-path` for `options-from-kconfig`, so that the extension goes one dir level up to retrieve the Kconfig data?

applications/nrf5340_audio/doc/feature_support.rst
```diff
@@ -0,0 +1,15 @@
...
+nRF5340 Audio feature support
+#############################
+
+The following table lists features of the nRF5340 Audio application and their respective limitations and maturity level.
+For an explanation of the maturity levels, see :ref:`Software maturity levels <software_maturity>`.
+
+.. note::
+   Features not listed are not supported.
+
+.. include:: /releases_and_maturity/software_maturity.rst
```
erikrobstad, line 13: We don't need it to have the ".. toggle::" now

applications/nrf5340_audio/README.rst
```diff
@@ -1064,83 +793,6 @@ You can test upgrading the firmware on both cores at the same time on a headset
...
-
-Do not use the default MCUboot key for end products.
-See :ref:`ug_fw_update` and :ref:`west-sign` for more information.
-
-To create your own app that supports DFU, you can use the `nRF Connect Device Manager`_ libraries for Android and iOS.
-
-Changing default values
-=======================
-
-Given the requirements for the Coordinated Set Identification Service (CSIS), make sure to change the Set Identity Resolving Key (SIRK) value when adapting the application.
```
erikrobstad, line 1142: Looks like this part has been removed now?

erikrobstad (changes requested):

alexsven:
I see that none of the default values shows up for the different Kconfigs, I know they didn't do that before this PR either but is there a way to parse that and show it in the doc?

alexsven (approved):
Looks better now, not as overwhelming when first opening up the documentation

rick1082 (approved):

erikrobstad (approved):

greg-fer:
> I see that none of the default values shows up for the different Kconfigs, I know they didn't do that before this PR either but is there a way to parse that and show it in the doc?

I'll ask the people who coordinate this part of the docs.

gmarull:
> > I see that none of the default values shows up for the different Kconfigs, I know they didn't do that before this PR either but is there a way to parse that and show it in the doc?
> 
> I'll ask the people who coordinate this part of the docs.

are these sample level Kconfigs? If so, this is not supported.

PR #11803, opened 2023-07-13
net: wifi: Changes needed to run Zephyr net samples with Wi-Fi

rlubos:
This PR introduces a set of changes needed to run upstream Zephyr networking samples with nRF7002. For a list of samples, see https://github.com/nrfconnect/sdk-zephyr/pull/1242 - samples with an overlay added have been successfully tested with `nrf7002dk_nrf5340_cpuapp`.

This PR comprises mostly of changes needed to automatically enable features essential for Wi-Fi operation when an appropriate board is selected and networking is enabled. The current feature auto-enablement is pretty aggressive so in most cases, the only change needed on the sample side is to configure SSID/password. 

Originally, I was hoping we could rely on automatic configuration only, and avoid overlays in `sdk-zephyr`, but it soon turned out that in some cases an overlay config is unavoidable (for example when the sample overwritten the default `POSIX_MAX_FDS`). Since we'd need to carry downstream patch anyway because of this, I've decided to add an overlay to all of the required samples, to clearly indicate which samples have been tested and are supported. This means we can relax the current approach to automatically enable certain features a bit and move configuration to the overlay files. If you think something is too much (like for example enabling Connection Manager automatically is a bit of a gray area for me) please speak up, this PR is, among others, to gather feedback.

As for the remaining patches:
* ae933961e5abd4e228fa9d72f31be69562a508ca - upstream networking samples rely on the NET_CONFIG library (and its initial boot delay) heavily, therefore we should not skip that part. I propose to revert that change, and set timeout to 0 in `sdk-nrf` samples only (not sure if it's really needed though)
* b92a6745741f9d83c836ec60950ccee6ebdaaf8e - in case static SSID/password configuration is used, we don't really need a backend to store credentials, it only adds overall configuration complexity (you need to enable settings/nvs/flash etc. for each and every sample). Therefore I suggest to introduce a dummy variant of the backend, enabled automatically when static credential configuration is used.


boards/arm/nrf7002dk_nrf5340/Kconfig.defconfig
```diff
@@ -71,6 +71,25 @@ config FLASH_LOAD_SIZE
 
 endif # BOARD_NRF7002DK_NRF5340_CPUAPP_NS
 
+if NETWORKING
```
krish2718, line 74: We should do this for nRF7000/nRF7001 also.

drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -297,4 +297,27 @@ config NRF700X_RSSI_STALE_TIMEOUT_MS
 	  value as driver does not store it and requires RPU to provide the
 	  info.
 
+if NETWORKING
+
+# Finetune defaults for certain system components used by the driver
+config HEAP_MEM_POOL_SIZE
+	default 120000
```
krish2718, line 304: Is this based on experiments to see lowest possible where Wi-Fi works? This might not work, `153000` is the number that we come up with , of course, we have a plan to run more tests and come up with a better number.

rlubos: It's just an arbitrary value that worked for me, I'm open to setting something more appropriate as well.

boards/arm/nrf7002dk_nrf5340/Kconfig.defconfig
rlubos: Yes, definitely.

rlubos: @krish2718  Now looking at this, I would add a similar defconfig for nR7001 DK/shield, but what about nRF7000 shield? `WPA_SUPP` is not needed for sure, but do `NETWORKING`/`NET_L2_ETHERNET` need to be enabled for scan only? Or is `WIFI`/`WIFI_NRF700X` enough?

krish2718: `WIFI_NRF700X` and `NET_L2_ETHERNET` both are needed.

samples/wifi/scan/prj.conf
```diff
@@ -7,6 +7,7 @@ CONFIG_WIFI=y
 CONFIG_WIFI_NRF700X=y
 CONFIG_NET_L2_WIFI_MGMT=y
 CONFIG_HEAP_MEM_POOL_SIZE=25000
+CONFIG_WPA_SUPP=n
```
glarsennordic, line 10: I think mqtt_multi_service and location sample WiFi scan-only overlays will need this change as well, as well as any other samples that have scan-only overlays for nrf9160. Or am I mistaken?

drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -302,4 +302,27 @@ config NRF700X_RSSI_STALE_TIMEOUT_MS
...
+	default 16
+
+config NET_TX_STACK_SIZE
+	default 4096
+
+config NET_RX_STACK_SIZE
+	default 4096
+
+config NET_TC_TX_COUNT
+	default 1
```
glarsennordic, line 324: Does NET_TX_RX_COUNT not need to be adjusted?

To be honest I am still a bit in the dark on what this setting controls and why it needs to be adjusted for our wifi drivers

subsys/net/l2_wifi_if_conn/Kconfig
```diff
@@ -6,6 +6,7 @@
 menuconfig L2_WIFI_CONNECTIVITY
 	bool "L2 Wi-Fi Connectivity"
 	depends on NET_CONNECTION_MANAGER
+	default y if WPA_SUPP
```
glarsennordic, line 9: This is the one bit I am worried about. It would be nice if we could be more surgical about this.

> This means we can relax the current approach to automatically enable certain features a bit and move configuration to the overlay files.

I wonder if it would be acceptable to place `CONFIG_L2_WIFI_CONNECTIVITY=y` in the nrf700x overlay files

We have no intention of ever up-streaming those overlays, and furthermore the overlays won't function properly unless `CONFIG_L2_WIFI_CONNECTIVITY` winds up enabled, so it seems ***somewhat*** reasonable to enable `CONFIG_L2_WIFI_CONNECTIVITY` there.

Yes, this would mean that the nrf700x overlays would be referencing a config option from sdk-nrf, but since we don't intend to upstream those overlays, and since those overlays won't work unless used from within sdk-nrf, this seems like not such a big deal

Thoughts @rlubos @mbolivar-nordic ?

rlubos:
I've encountered some issues with automatic Wi-Fi enablement on the EK...
* For a reason unknown to me, `modem_shell`, even though disabled WPA_SUPP explicitly, could not turn the option out. The same configuration (i. e. setting `CONFIG_WPA_SUPP=n` in the overlay) worked fine with the `location` sample, however in case of `modem_shell` the symbol behaved as if it was y-selected.
* In case of matter `lock` sample, one configuration builds the sample with EK enabled, but disables Wi-Fi:
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/lock/sample.yaml#L78
The way `CONFIG_CHIP_WIFI` is constructed however (selecting Wi-Fi configs if enabled) caused that CONFIG_WIFI etc. remained enabled for that configuration (CC @markaj-nordic)

For now, I've decided to drop the commit that enabled Wi-Fi for DK/EK automatically and moved the configs to the overlays in sdk-zephyr.

rlubos:
@richabp The doc page content/location is just my proposal, I'm fully open for even a complete redesign, whatever you suggest.

samples/wifi/scan/prj.conf
glarsennordic: Oh, my bad

subsys/net/l2_wifi_if_conn/Kconfig
glarsennordic: > I don't really think, that someone would enable both connection manager and supplicant with no intention of using it for Wi-Fi management.

You're right, I missed that this specifically targets the supplicant and not the wifi driver as a whole. So my concerns are definitely moot.

I'd still probably consider it marginally cleaner to enable directly in the overlays, but you're right, there is truly not any issue with doing it the way you've done it so feel free to ignore that personal preference if you think it is best

glarsennordic (approved):

doc/nrf/samples/wifi_zephyr.rst
```diff
@@ -0,0 +1,48 @@
...
+* :ref:`sockets-http-get`
+* :ref:`sntp-client-sample`
+* :ref:`syslog-net-sample`
+* :ref:`telnet-console-sample`
+
+Building and running
+********************
+
+All of the listed samples have been provided with an overlay file, ``overlay-nrf700x.conf``, which
+allows to configure the sample to run with nRF700x driver.
```
ioannisg, line 31: Is "nRF700x driver" the right name? IIRC, we use "Wi-Fi driver" in NCS, but up to finding a better name if needed.

doc/nrf/samples/wifi_zephyr.rst
```diff
@@ -0,0 +1,48 @@
+.. _wifi_zephyr_samples:
+
+Wi-Fi: Zephyr networking samples
+################################
+
+In addition to nRF Connect SDK samples, it is also possible to run selected Zephyr networking
+samples with Wi-Fi.
+The following list specifies samples that are currently supported with the nRF700x driver:
```
ioannisg, line 8: same comment here.

ioannisg: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/protocols/wifi/index.html
(Wi-Fi driver is the name, at least in this page)

doc/nrf/samples/wifi_zephyr.rst
```diff
@@ -0,0 +1,48 @@
+.. _wifi_zephyr_samples:
+
+Wi-Fi: Zephyr networking samples
+################################
+
+In addition to nRF Connect SDK samples, it is also possible to run selected Zephyr networking
```
ioannisg, line 6: should we mention "upstream Zephyr" samples here, to indicate that these are maintained in the upstream project?
e.g. "it is also possible to run selected TCP/IP networking samples provided and maintained as part of  the upstream Zephyr project"?

ioannisg (commented):
some very minor doc comments/questions from me

doc/nrf/samples/wifi_zephyr.rst
rlubos: I was inspired by "Working with nRF700x" section where it referred "nRF700x driver": https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/device_guides/working_with_nrf/nrf70/developing/constrained.html#id9
But I can switch to Wi-Fi driver if it's more commonly used.

doc/nrf/samples/wifi_zephyr.rst
rlubos: I'll update this section, however I'd drop the `TCP/IP` part and stick to `networking samples`. IMO it's narrowing too much, we also support CoAP/DNS etc which are UDP.

doc/nrf/samples/wifi_zephyr.rst
richabp: Not sure if **nRF700X Wi-Fi driver** could be the alternative as used in https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/drivers/nrf700x.html.

markaj-nordic:
> * In case of matter `lock` sample, one configuration builds the sample with EK enabled, but disables Wi-Fi:
>       https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/lock/sample.yaml#L78
>       The way `CONFIG_CHIP_WIFI` is constructed however (selecting Wi-Fi configs if enabled) caused that CONFIG_WIFI etc. remained enabled for that configuration (CC @markaj-nordic)

Discussed f2f, will be discussed and potentially addressed (dedicated configuration for nRF53/nRF7002EK + Thread -> Matter switchable WiFi/Thread door lock use case).

doc/nrf/samples/wifi_zephyr.rst
ioannisg: I am considering us phasing out the nRF700X term. Marketing has already switched to mostly used nRF70 Series, @richabp. Keep Wi-Fi driver for now, IMHO.

doc/nrf/samples/wifi_zephyr.rst
ioannisg: That's fine. For me TCP/IP stack refers to the stack model name, not just the two protocols (i.e. it is another way to refer to the internet protocol suite)

ioannisg (approved):
thanks

doc/nrf/samples/wifi_zephyr.rst
richabp: @ioannisg Sure!

doc/nrf/samples/wifi_zephyr.rst
```diff
@@ -0,0 +1,48 @@
...
+* :ref:`sockets-http-get`
+* :ref:`sntp-client-sample`
+* :ref:`syslog-net-sample`
+* :ref:`telnet-console-sample`
+
+Building and running
+********************
+
+All of the listed samples have been provided with an overlay file, ``overlay-nrf700x.conf``, which
+allows to configure the sample to run with Wi-Fi driver.
```
richabp, line 31:
```suggestion
An overlay file, ``overlay-nrf700x.conf`` is provided to all Zephyr samples, which configures the sample to run with the Wi-Fi driver.
```

doc/nrf/samples/wifi_zephyr.rst
```diff
@@ -0,0 +1,48 @@
...
+
+To build Zephyr samples for the nRF7002 DK, use the ``nrf7002dk_nrf5340_cpuapp`` build target.
+The following is an example of the CLI command:
+
+.. code-block:: console
+
+   west build -b nrf7002dk_nrf5340_cpuapp -- -DOVERLAY_CONFIG=overlay-nrf700x.conf
+
+To build for the nRF7002 EK with nRF5340 DK, use the ``nrf5340dk_nrf5340_cpuapp`` build target with
+the ``SHIELD`` CMake option set to ``nrf7002ek``.
```
richabp, line 41:
```suggestion
To build for the nRF7002 EK with nRF5340 DK, use the ``nrf5340dk_nrf5340_cpuapp`` build target with the ``SHIELD`` CMake option set to ``nrf7002ek``.
```

richabp:
> @richabp The doc page content/location is just my proposal, I'm fully open for even a complete redesign, whatever you suggest.

@rlubos Shouldn't it be in the Wi-Fi folder?

richabp (commented):
Some minor suggestions.

bama-nordic (approved):

doc/nrf/samples/wifi_zephyr.rst
```diff
@@ -0,0 +1,48 @@
...
+* :ref:`sockets-echo-sample`
+* :ref:`async-sockets-echo-sample`
+* :ref:`sockets-echo-client-sample`
+* :ref:`sockets-echo-server-sample`
+* :ref:`sockets-http-get`
+* :ref:`sntp-client-sample`
+* :ref:`syslog-net-sample`
+* :ref:`telnet-console-sample`
+
+Building and running
```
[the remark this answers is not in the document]
rlubos: I've added a Configuration section with details on how to configure Wi-Fi credentials.

The Dependencies section IMO doesn't make much sense, as we're not documenting a single example here, each sample has its own dependencies.

As for requirements, I'm a bit puzzled here. We typically use `.. table-from-sample-yaml::` for this, but here that's not the case, as there's is no corresponding `sample.yaml` with this documentation.

rlubos:
Thanks @richabp, I've applied your comments, and added "Configuration" section.

> > @richabp The doc page content/location is just my proposal, I'm fully open for even a complete redesign, whatever you suggest.
> 
> @rlubos Shouldn't it be in the Wi-Fi folder?

Did you mean `samples/wifi` here? I wasn't sure if we can put documentation there, as we'd be adding doc file in the sample folder structure w/o a corresponding sample, but if that's fine I can move it there. Unless you meant something else.

rlubos:
Thanks @richabp, I've applied your suggestions

richabp (approved):

krish2718 (approved):

rlubos:
As the Zephyr part of the supplicant was moved to `sdk-nrf`, the PR towards `sdk-hostap` was no longer applicable. Instead, I've moved the two commits to this repository.

PR #11811, opened 2023-07-14
nrf_security: Enable PSA Key types in samples and configurations

joerchan:
Update sdk-zephyr manifest to include PSA key types as user configurations.

Enable PSA key types in all samples, tests and crypto configurations.

test-sdk-nrf: joerchan-11811


doc/nrf/libraries/nrf_security/doc/driver_config.rst
```diff
@@ -30,6 +30,79 @@ If multiple drivers are enabled, the first ordered item in this table takes prec
 
 Enabling or disabling PSA driver specific configurations controls the support for a given algorithm, per driver.
 
+Key type configuration
+**********************
+
+To enable block or stream cipher algorithms, set ore or more of the following Kconfig options:
```
SebastianBoe, line 36:
```suggestion
To enable block or stream cipher algorithms, set one or more of the following Kconfig options:
```

doc/nrf/libraries/nrf_security/doc/driver_config.rst
```diff
@@ -30,6 +30,55 @@ If multiple drivers are enabled, the first ordered item in this table takes prec
...
+To enable block or stream cipher algorithms, set one or more of the following Kconfig options:
+
++-----------------------+-------------------------------------------------------------+
+| Key type              | Configuration option                                        |
++=======================+=============================================================+
+| AES                   | :kconfig:option:`CONFIG_PSA_WANT_KEY_TYPE_AES`              |
++-----------------------+-------------------------------------------------------------+
+| Chacha20              | :kconfig:option:`CONFIG_PSA_WANT_KEY_TYPE_CHACHA20`         |
++-----------------------+-------------------------------------------------------------+
+| ECC Key Pair          | :kconfig:option:`CONFIG_PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY`   |
```
SebastianBoe, line 45:
```suggestion
| ECC Key Pair          | :kconfig:option:`CONFIG_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR`   |
```

joerchan: Thanks.

doc/nrf/libraries/nrf_security/doc/driver_config.rst
```diff
@@ -30,6 +30,55 @@ If multiple drivers are enabled, the first ordered item in this table takes prec
 
 Enabling or disabling PSA driver specific configurations controls the support for a given algorithm, per driver.
 
+Key type configuration
+**********************
+
+To enable block or stream cipher algorithms, set one or more of the following Kconfig options:
```
SebastianBoe, line 36: AFAIK RSA does not use a block or stream cipher.

Perhaps we should describe this in a different way:

Algorithms use keys with some flexibility in which keys can be used for which operations.
But to use a PSA key it's corresponding Kconfig option must first be enabled.

The Kconfig options follow a specific naming pattern: the PSA key PSA_KEY_TYPE_<type> is enabled by the Kconfig option CONFIG_PSA_WANT_KEY_TYPE_<type>.

For information about which keys can be used with specific algorithms, please refer to the PSA Crypto API documentation

doc/nrf/libraries/nrf_security/doc/driver_config.rst
```diff
@@ -392,7 +441,7 @@ To configure the Hash algorithms, set one or more of the following Kconfig optio
 +-----------------------+---------------------------------------------------+
 | Hash algorithm        | Configuration option                              |
 +=======================+===================================================+
-| SHA-1                 | :kconfig:option:`CONFIG_PSA_WANT_ALG_SHA_1`       |
+| SHA-1 (unsafe)        | :kconfig:option:`CONFIG_PSA_WANT_ALG_SHA_1`       |
```
SebastianBoe, line 444:
```suggestion
| SHA-1 (weak)        | :kconfig:option:`CONFIG_PSA_WANT_ALG_SHA_1`       |
```

same elsewhere. See Georgios's review.

samples/crypto/aes_ccm/prj.conf
```diff
@@ -16,4 +16,5 @@ CONFIG_MBEDTLS_ENABLE_HEAP=y
 CONFIG_MBEDTLS_HEAP_SIZE=8192
 
 CONFIG_PSA_WANT_GENERATE_RANDOM=y
+CONFIG_PSA_WANT_KEY_TYPE_AES=y
 CONFIG_PSA_WANT_ALG_CCM=y
```
SebastianBoe, line 20: Our drivers only support the AES key for CCM.

Would it be possible for us to select the AES key from the CCM algorithm?

Or maybe some other mechanism to not make this "choice" that is not really a choice.
That would mean less config work for our users.

image

tests/tfm/tfm_psa_test/prj.conf
```diff
@@ -16,7 +16,13 @@ CONFIG_TFM_ISOLATION_LEVEL=2
 # RNG
 CONFIG_PSA_WANT_GENERATE_RANDOM=y
 
-# AES Ciphers
+# Keys
+CONFIG_PSA_WANT_KEY_TYPE_AES=y
+CONFIG_PSA_WANT_KEY_TYPE_CHACHA20=y
+CONFIG_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR=y
+CONFIG_PSA_WANT_KEY_TYPE_RSA_KEY_PAIR=y
```
SebastianBoe, line 23: Should we just enable MBEDTLS_CIPHER_ALL_ENABLED maybe?
Or does it have some unwanted side-effects?

Maybe that would require less maintenance here.

SebastianBoe: same for regression test

SebastianBoe (changes requested):
changes requested

samples/crypto/aes_ccm/prj.conf
frkv: Maybe this is a case for the Kconfig "imply" mechanism?

tests/tfm/tfm_psa_test/prj.conf
joerchan: These are defined as part of Zephyrs TLS configuration, I don't want to mix in those here.

I don't see this as a maintenance issue, but we can add PSA_WANT_ALL configurations in the future if we need.

samples/crypto/aes_ccm/prj.conf
```diff
@@ -16,4 +16,5 @@ CONFIG_MBEDTLS_ENABLE_HEAP=y
 CONFIG_MBEDTLS_HEAP_SIZE=8192
```
SebastianBoe, line 16: Hi, using this branch I built this sample and it built fine.

But to my surprise disabling CONFIG_PSA_WANT_KEY_TYPE_AES did not produce a build-time error.

IMHO we cannot introduce this option without validating it's correctness at build time.

samples/crypto/aes_ccm/prj.conf
SebastianBoe: Hi, I did some experiments to see if this would work and it appears to work as intended.

https://github.com/SebastianBoe/fw-nrfconnect-nrf-1/commit/adc80da258298d5b9142759b056be1f48b846f4f

Here we introduce a promptless default value for the AES key type that is processed early in the configuration to override upstream configs. This follows prior-art (openthread).

I agree we should not patch upstream config files with our platform-specific limitations.

"I find it simplest to have a one-to-one mapping between Kconfig symbols to header defines, instead of having implicit symbols in this way."

I agree with this and in the proposed patch we would still have a one-to-one mapping between Kconfig symbols and header defines.

"Also with the PSA_WANT_AES_KEY_SIZE_128/192/256 it becomes kind of strange to not have the key type for AES IMO."

I agree that we should have a Kconfig symbol for the AES key type, in the proposed patch we still have such a symbol.

doc/nrf/libraries/nrf_security/doc/driver_config.rst
mswarowsky: Agree the PSA Crypto API doc also uses the word weak for it I think it would be nice to be aligned with them, and we use weak in L460/461
https://armmbed.github.io/mbed-crypto/html/api/ops/hashes.html

doc/nrf/libraries/nrf_security/doc/driver_config.rst
```diff
@@ -402,11 +451,14 @@ To configure the Hash algorithms, set one or more of the following Kconfig optio
 +-----------------------+---------------------------------------------------+
 | SHA-512               | :kconfig:option:`CONFIG_PSA_WANT_ALG_SHA_512`     |
 +-----------------------+---------------------------------------------------+
-| MD5                   | :kconfig:option:`CONFIG_PSA_WANT_ALG_MD5`         |
+| MD5 (unsafe)          | :kconfig:option:`CONFIG_PSA_WANT_ALG_MD5`         |
```
mswarowsky, line 454:
```suggestion
| MD5 (weak)          | :kconfig:option:`CONFIG_PSA_WANT_ALG_MD5`         |
```

doc/nrf/libraries/nrf_security/doc/driver_config.rst
```diff
@@ -416,7 +468,7 @@ You can use the following PSA driver-specific configurations for fine-grained co
 +-----------------------+---------------------------------------------------------------+---------------------------------------------------------------+
 | Hash algorithm        |  nrf_cc3xx driver support                                     | nrf_oberon driver support                                     |
 +=======================+===============================================================+===============================================================+
-| SHA-1                 | :kconfig:option:`CONFIG_PSA_CRYPTO_DRIVER_ALG_SHA_1_CC3XX`    | :kconfig:option:`CONFIG_PSA_CRYPTO_DRIVER_ALG_SHA_1_OBERON`   |
+| SHA-1 (unsafe)        | :kconfig:option:`CONFIG_PSA_CRYPTO_DRIVER_ALG_SHA_1_CC3XX`    | :kconfig:option:`CONFIG_PSA_CRYPTO_DRIVER_ALG_SHA_1_OBERON`   |
```
mswarowsky, line 471:
```suggestion
| SHA-1 (weak)        | :kconfig:option:`CONFIG_PSA_CRYPTO_DRIVER_ALG_SHA_1_CC3XX`    | :kconfig:option:`CONFIG_PSA_CRYPTO_DRIVER_ALG_SHA_1_OBERON`   |
```

doc/nrf/libraries/nrf_security/doc/driver_config.rst
```diff
@@ -426,9 +478,9 @@ You can use the following PSA driver-specific configurations for fine-grained co
 +-----------------------+---------------------------------------------------------------+---------------------------------------------------------------+
 | SHA-512               | Not supported                                                 | :kconfig:option:`CONFIG_PSA_CRYPTO_DRIVER_ALG_SHA_512_OBERON` |
 +-----------------------+---------------------------------------------------------------+---------------------------------------------------------------+
-| MD5                   | Not supported                                                 | Not supported                                                 |
+| MD5 (unsafe)          | Not supported                                                 | Not supported                                                 |
```
mswarowsky, line 481:
```suggestion
| MD5 (weak)          | Not supported                                                 | Not supported                                                 |
```

doc/nrf/libraries/nrf_security/doc/driver_config.rst
mswarowsky, line 36: Shall we just say "cryptographic algorithms", as then it includes asymmetric crypto, and the key agreement algorithms as well
```suggestion
To enable cryptographic algorithms, set one or more of the following Kconfig options:
```

mswarowsky (changes requested):

samples/crypto/aes_ccm/prj.conf
SebastianBoe: Will be fixed in Georgios's PR.

mswarowsky (approved):

doc/nrf/libraries/nrf_security/doc/driver_config.rst
```diff
@@ -30,6 +30,55 @@ If multiple drivers are enabled, the first ordered item in this table takes prec
...
++-----------------------+-------------------------------------------------------------+
+| DH Public key         | :kconfig:option:`CONFIG_PSA_WANT_KEY_TYPE_DH_PUBLIC_KEY`    |
++-----------------------+-------------------------------------------------------------+
+
+Key type support
+================
+
+Key type support for each driver:
+
++-----------------------+---------------------------+----------------------------+
```
b-gent, line 63: This table doesn't render properly (additional + sign needed on lines 67, 69, 71 (...) to indicate a separate cell)
image

b-gent (changes requested):

b-gent (approved):
lgtm now

rlubos:
LwM2M tests passed on previous run, current diff was only with doc changes, merging.

PR #11815, opened 2023-07-17
test: unity: add 64 bit support

lemrey:
Add support for 64 bits macros.


lemrey:
It enables support for 64 bit types in Unity, which we use in libmodem.

lemrey:
This only affect tests, so I am removing the `changelog-entry-required`

eivindj-nordic (approved):

MirkoCovizzi (approved):

PR #11821, opened 2023-07-18
net: lib: nrf_cloud: move dev status flag out of codec

jayteemo:
Move the flag that controls the device status shadow update out of the codec and into the fsm.
IRIS-5485


glarsennordic (approved):
Looks great

plskeggs (approved):

plskeggs:
Any need for a changelog entry?  Seems like this does not modify the public API.

jayteemo:
> Any need for a changelog entry? Seems like this does not modify the public API.

i saw no need, no functionality changes, just internal cleanup.

PR #11823, opened 2023-07-18
manifest: sdk-hostap: Pull fix for scan busy after disconnect

krish2718:
Pull in the fix to abort ongoing scan (if any) during disconnect.


VivekUppunda (approved):

krish2718:
Manifest update will happen in https://github.com/nrfconnect/sdk-nrf/pull/11820

PR #11826, opened 2023-07-18
Bluetooth: add le power control HCI vs cmd APR handling

KyraLengfeld:
The vs command set_power_control_apr_handling can be used to enable utilization of remote APR (acceptible power reduction), received with the power control response from peer, to adjust the local tx power.

For this the sdk-nrfxlib revision is updated as well.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -140,7 +140,7 @@ manifest:
     - name: nrfxlib
       repo-path: sdk-nrfxlib
       path: nrfxlib
-      revision: 16721d9c09b2e37fc805fa992a7537211bd0836b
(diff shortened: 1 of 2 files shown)


Pieloaf (approved):

west.yml
```diff
@@ -140,7 +140,7 @@ manifest:
     - name: nrfxlib
       repo-path: sdk-nrfxlib
       path: nrfxlib
-      revision: 16721d9c09b2e37fc805fa992a7537211bd0836b
+      revision: pull/1055/head
```
rlubos, line 143: `83892a4e56b9dfc4cfcc954e62cfb65cec13509c`

PR #11829, opened 2023-07-18
drivers: wifi: Fix the interface check

krish2718:
commit 34f31d55bd("drivers: wifi: Fix IP event handling") missed checking for validity of VIF context, so, it still assumes the interface is Wi-Fi and goes ahead, causing a crash.

Also, convert the check to a helper function to be used in multiple places.

### DIFF
--- a/drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
+++ b/drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
@@ -95,6 +95,26 @@ const char *wifi_nrf_get_drv_version(void)
 	return NRF700X_DRIVER_VERSION;
 }
 
+/* If the interface is not Wi-Fi then errors are expected, so, fail silently */
+struct wifi_nrf_vif_ctx_zep *wifi_nrf_get_vif_ctx(struct net_if *iface)
+{
+	struct wifi_nrf_vif_ctx_zep *vif_ctx_zep = NULL;
+	struct wifi_nrf_ctx_zep *rpu_ctx = &rpu_drv_priv_zep.rpu_ctx_zep;
+
+	if (!iface || !rpu_ctx) {
+		return NULL;
+	}
+
+	for (int i = 0; i < ARRAY_SIZE(rpu_ctx->vif_ctx_zep); i++) {
+		if (rpu_ctx->vif_ctx_zep[i].zep_net_if_ctx == iface) {
+			vif_ctx_zep = &rpu_ctx->vif_ctx_zep[i];
+			break;
+		}
+	}
+
+	return vif_ctx_zep;
+}
+
 void wifi_nrf_event_proc_scan_start_zep(void *if_priv,
 					struct nrf_wifi_umac_event_trigger_scan *scan_start_event,
 					unsigned int event_len)
--- a/drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
+++ b/drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
@@ -178,18 +178,10 @@ static void ip_maddr_event_handler(struct net_mgmt_event_callback *cb,
 				   struct net_if *iface)
 {
 	struct wifi_nrf_vif_ctx_zep *vif_ctx_zep = NULL;
-	const struct device *dev = NULL;
 	struct net_eth_addr mac_addr;
 	struct nrf_wifi_umac_mcast_cfg *mcast_info = NULL;
 	enum wifi_nrf_status status;
 	uint8_t mac_string_buf[sizeof("xx:xx:xx:xx:xx:xx")];
-
-	dev = net_if_get_device(iface);
-
-	if (!dev) {
-		LOG_ERR("%s: dev is NULL\n", __func__);
-		return;
-	}
 
 	for (int i = 0; i < ARRAY_SIZE(rpu_ctx->vif_ctx_zep); i++) {
 		if (rpu_ctx->vif_ctx_zep[i].zep_net_if_ctx == iface) {
--- a/drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
+++ b/drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
@@ -207,7 +207,7 @@ static void ip_maddr_event_handler(struct net_mgmt_event_callback *cb,
 	    (mgmt_event == NET_EVENT_IPV4_MADDR_DEL)) {
 		if ((cb->info == NULL) ||
 		    (cb->info_length != sizeof(struct in_addr))) {
-			return;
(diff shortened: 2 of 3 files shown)


drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
```diff
@@ -95,6 +95,38 @@ const char *wifi_nrf_get_drv_version(void)
...
+
+	for (int i = 0; i < ARRAY_SIZE(rpu_ctx->vif_ctx_zep); i++) {
+		if (rpu_ctx->vif_ctx_zep[i].zep_net_if_ctx == iface) {
+			vif_ctx_zep = &rpu_ctx->vif_ctx_zep[i];
+			break;
+		}
+	}
+
+	if (!vif_ctx_zep) {
+		LOG_DBG("%s: Event received for an interface that is not "
```
glarsennordic, line 123: For multi-iface hardware this is going to create a log message for every IPv4 event raised on any non-WiFi iface

You mention `If the interface is not Wi-Fi then errors are expected`, but this seems wrong to me -- It is very much expected that these events will be raised on non-Wi-Fi ifaces, so I'm not sure this should be a debug log

drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
```diff
@@ -191,11 +191,9 @@ static void ip_maddr_event_handler(struct net_mgmt_event_callback *cb,
 		return;
```
glarsennordic, line 191: Do I understand correctly that this function is meant to only handle IP_MADDR events?

That is, the only events you are interested in are:

1. NET_EVENT_IPV4_MADDR_ADD
2. NET_EVENT_IPV4_MADDR_DEL
3. NET_EVENT_IPV6_MADDR_ADD
4. NET_EVENT_IPV6_MADDR_DEL

If so, then I would recommend adding a check to confirm that it really is one of these four events being raised. Currently, this event callback is being fired for all events in the `NET_EVENT_IPV4` and `NET_EVENT_IPV6` event groups.

This is because `net_mgmt` does not prevent event false positives, but only false negatives. So for instance:


```
net_mgmt_init_event_callback(&ip_maddr4_cb,
		     ip_maddr_event_handler,
		     NET_EVENT_IPV4_MADDR_ADD | NET_EVENT_IPV4_MADDR_DEL);
```

is guaranteed to fire for `NET_EVENT_IPV4_MADDR_ADD` and `NET_EVENT_IPV4_MADDR_DEL`, but it may also fire for any other `NET_EVENT_IPV4` event whose event code happens to fall within the bitmask

I think the full list of events that particular bitmask matches is

1. NET_EVENT_IPV4_CMD_ADDR_ADD
2. NET_EVENT_IPV4_CMD_ADDR_DEL
3. NET_EVENT_IPV4_CMD_MADDR_ADD
4. NET_EVENT_IPV4_CMD_MADDR_DEL
5. NET_EVENT_IPV4_CMD_ROUTER_ADD
6. NET_EVENT_IPV4_CMD_ROUTER_DEL
7. NET_EVENT_IPV4_CMD_DHCP_START


See the network management docs for more details:
https://docs.zephyrproject.org/latest/connectivity/networking/api/net_mgmt.html

Namely:

> Event set filtering allows false positives for events that have the same layer and layer code. A callback handler function must check the event code (passed as an argument) against the specific network events it will handle, regardless of how many events were in the set passed to [net_mgmt_init_event_callback()](https://docs.zephyrproject.org/latest/connectivity/networking/api/net_mgmt.html#c.net_mgmt_init_event_callback).

glarsennordic (changes requested):

drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
krish2718: True, but debugs are disabled by default, this is just a FYI, but if that is too much noise I can remove them.

glarsennordic: I'm not sure how much noise it will be, I worry mostly that it will be misleading debug output, especially given the `If the interface is not Wi-Fi then errors are expected` comment, I'd recommend at least removing that comment since it is IMO a bit misleading

If you add the IPV4 event filtering I mention in my other comment, then I think the event overlap will be reduced and this won't be so noisy so it's probably fine

drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
glarsennordic: I did see those two if statements, but consider for example if `NET_EVENT_IPV4_CMD_DHCP_START` is fired

`wifi_nrf_fmac_set_mcast_addr` will still fire even though `NET_EVENT_IPV4_CMD_DHCP_START` is not one of the four events you are interested in, because those if statements serve only to configure the arguments that will be passed to it.

You need, for instance, an additional else statement which returns out of the event handler if it is none of the four target events

drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
krish2718: Well, after rethinking changed my mind, not much value in those prints, so, removed them now.

drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
glarsennordic: You're right, return early makes way more sense -- either way, you need something which actually cancels handling the event if it's not one of the four target events

krish2718: Also, there is a memory leak, so, fixing that as well :)

drivers/wifi/nrf700x/zephyr/src/zephyr_net_if.c
```diff
@@ -196,6 +196,13 @@ static void ip_maddr_event_handler(struct net_mgmt_event_callback *cb,
...
 	}
 
+	if (mgmt_event != NET_EVENT_IPV4_MADDR_ADD &&
+	    mgmt_event != NET_EVENT_IPV4_MADDR_DEL &&
+	    mgmt_event != NET_EVENT_IPV6_MADDR_ADD &&
+	    mgmt_event != NET_EVENT_IPV6_MADDR_DEL) {
+		return;
+	}
```
glarsennordic: I'd put this even before the call to `wifi_nrf_get_vif_ctx` and `net_if_get_device calls since it will save you doing those lookups for spurious events

VivekUppunda (approved):

rado17 (approved):

glarsennordic (approved):
LGTM

PR #11831, opened 2023-07-19
boards: Align the nRF21540 EK names with the new convention

maje-emb:
According to the new naming convention there should be no '_' characters in the board names.


MarekPieta (approved):

dchat-nordic (approved):

KAGA164 (commented):
I tested these changes. They looks good. I think that it is worth to put note about this shield name change in the changelog since it affects build command for many subsystems

grochu (approved):
Looks good

alexsven (approved):

PR #11841, opened 2023-07-20
applications: nrf_desktop: Add peers cache to HID forward module

MarekPieta:
PR introduces:

- Configuration channel peers cache functionality to the HID forward module. The option allows host tools to retrieve information about change of the connected peripherals without need of complete rediscovery.
- Support for the peers cache in the HID configurator script.
- Configuration channel implementation cleanups.
- Fix for invalid assertion in config event logging implementation.

Jira: NCSDK-17179

### DIFF
--- a/applications/nrf_desktop/src/modules/hid_forward.c
+++ b/applications/nrf_desktop/src/modules/hid_forward.c
@@ -756,7 +756,7 @@ static void handle_config_channel_peers_req(const struct config_event *event)
 	BUILD_ASSERT(ARRAY_SIZE(peripherals) + CFG_CHAN_BASE_ID <=
 		     CFG_CHAN_UNUSED_PEER_ID);
 
-	static uint8_t cur_per;
+	static uint8_t cur_per_idx;
 
 	switch (event->status) {
 	case CONFIG_STATUS_INDEX_PEERS:
@@ -767,41 +767,35 @@ static void handle_config_channel_peers_req(const struct config_event *event)
 			}
 		}
 
-		cur_per = 0;
+		cur_per_idx = 0;
 		send_nodata_response(event, CONFIG_STATUS_SUCCESS);
 		break;
 
 	case CONFIG_STATUS_GET_PEER:
 	{
-		struct hids_peripheral *per;
-
-		if (cur_per < ARRAY_SIZE(peripherals)) {
-			per = &peripherals[cur_per];
-		} else {
-			per = NULL;
-		}
-
-		while ((cur_per < ARRAY_SIZE(peripherals)) &&
-		       (per->cfg_chan_id == CFG_CHAN_UNUSED_PEER_ID)) {
-			cur_per++;
-			if (cur_per < ARRAY_SIZE(peripherals)) {
-				per = &peripherals[cur_per];
-			}
-		}
+		struct hids_peripheral *per = NULL;
 
 		BUILD_ASSERT(sizeof(per->hwid) == HWID_LEN);
+
+		while (cur_per_idx < ARRAY_SIZE(peripherals)) {
+			struct hids_peripheral *cur_per = &peripherals[cur_per_idx++];
+
+			if (cur_per->cfg_chan_id != CFG_CHAN_UNUSED_PEER_ID) {
+				per = cur_per;
+				break;
+			}
+		}
 
 		struct config_event *rsp = generate_response(event,
 						    HWID_LEN + sizeof(uint8_t));
 		size_t pos = 0;
 
-		if (cur_per < ARRAY_SIZE(peripherals)) {
+		if (per) {
 			memcpy(&rsp->dyndata.data[pos],
 			       per->hwid, sizeof(per->hwid));
 			pos += sizeof(per->hwid);
 
 			rsp->dyndata.data[pos] = per->cfg_chan_id;
-			cur_per++;
 		} else {
 			pos += HWID_LEN;
 
--- a/applications/nrf_desktop/src/modules/hid_forward.c
+++ b/applications/nrf_desktop/src/modules/hid_forward.c
@@ -85,6 +85,7 @@ struct hids_peripheral {
 static struct subscriber subscribers[CONFIG_USB_HID_DEVICE_COUNT];
 static bt_addr_le_t peripheral_address[CONFIG_BT_MAX_PAIRED];
 static struct hids_peripheral peripherals[CONFIG_BT_MAX_CONN];
+static uint8_t peripheral_cache[CONFIG_BT_MAX_CONN];
 static bool suspended;
 
 
@@ -568,6 +569,10 @@ static int register_peripheral(struct bt_gatt_dm *dm, const uint8_t *hwid,
 	__ASSERT_NO_MSG(hwid_len == HWID_LEN);
 	memcpy(per->hwid, hwid, hwid_len);
 
+	peripheral_cache[per_id]++;
+	/* An odd number is assigned to a connected peripheral. */
+	__ASSERT_NO_MSG(peripheral_cache[per_id] & 0x01);
+
 	LOG_INF("Peripheral %p registered and linked to %p", (void *)per,
 		(void *)get_subscriber(per));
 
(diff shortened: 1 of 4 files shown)


MarekPieta:
Fixed compliance issues.

applications/nrf_desktop/src/modules/hid_forward.c
```diff
@@ -997,6 +1023,12 @@ static void disconnect_peripheral(struct hids_peripheral *per)
 	if (per->cfg_chan_rsp) {
 		submit_forward_error_rsp(per, CONFIG_STATUS_WRITE_FAIL);
 	}
+
+	size_t per_id = per - &peripherals[0];
+
+	peripheral_cache[per_id]++;
+	/* A disconnected peripheral is marked with an even number. */
+	__ASSERT_NO_MSG(!(peripheral_cache[per_id] & 1));
```
kapi-no, line 1031: Convert constant `1` to define (e.g. CONNECTED_PEER_DISTINGUISHER or any other name that could hint what it is used for). Also it would be good to define constant as hex `0x01` as it is mainly used in the bitmask operations

applications/nrf_desktop/src/modules/hid_forward.c
```diff
@@ -997,6 +1023,12 @@ static void disconnect_peripheral(struct hids_peripheral *per)
 	if (per->cfg_chan_rsp) {
 		submit_forward_error_rsp(per, CONFIG_STATUS_WRITE_FAIL);
 	}
+
+	size_t per_id = per - &peripherals[0];
```
kapi-no, line 1027: Consider defining a simple function for calculating per_id from the per instances (e.g. per_to_per_id). Could it be used also in other places? (If so, i would be great to add a dedicated commit with a small refactor before adding this functionality).

scripts/hid_configurator/NrfHidDevice.py
```diff
@@ -377,10 +390,23 @@ def _read_device_info(dev, recipient):
...
+                                                                   0,
+                                                                   ConfigStatus.GET_PEERS_CACHE,
+                                                                   None)
+        if success:
+            return " ".join("0x{:02x}".format(b) for b in fetched_data)
+        else:
+            return None
+
     @staticmethod
     def _get_connected_peers(dev, recipient):
```
kapi-no, line 406: why do we attach also peers cache results to this function? Can't we have two standalone functions one for getting connected peers and for getting peers cache?

scripts/hid_configurator/NrfHidDevice.py
```diff
@@ -211,7 +213,18 @@ def open_devices(vid):
                         dir_devs[hwid] = discovered_dev
                         dev_active = True
 
-                peers = NrfHidDevice._get_connected_peers(dev, LOCAL_RECIPIENT)
+                peers, peers_cache = NrfHidDevice._get_connected_peers(dev, LOCAL_RECIPIENT)
```
kapi-no, line 216: If you separate functions, you can poll peer cache here and then after handling it you can poll for connected peers

applications/nrf_desktop/src/modules/hid_forward.c
```diff
@@ -773,35 +778,31 @@ static void handle_config_channel_peers_req(const struct config_event *event)
...
-			if (cur_per < ARRAY_SIZE(peripherals)) {
-				per = &peripherals[cur_per];
+
+			if (per->cfg_chan_id != CFG_CHAN_UNUSED_PEER_ID) {
+				break;
 			}
-		}
 
-		BUILD_ASSERT(sizeof(per->hwid) == HWID_LEN);
+			per = NULL;
```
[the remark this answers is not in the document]
MarekPieta: It would. Changed. I also renamed `cur_per_idx` to `cur_per` to prevent confusion

applications/nrf_desktop/src/modules/hid_forward.c
MarekPieta: Changed to `0x01`, thanks for the tip.
I think comment is more meaningful than constant name here (it's difficult to find meaningful name), so I would prefer to keep it as is (also the code is used only in assertions).

applications/nrf_desktop/src/modules/hid_forward.c
MarekPieta: Currently this is the only place where we use it (there is another function that finds peripheral by configuration channel recipient and it's used in other places). Can we keep it as is?

scripts/hid_configurator/NrfHidDevice.py
MarekPieta: I prefer to keep it together (from the end-user perspective cache value is useless standalone).  Later on the `_get_connected_peers` could use peers cache internally to avoid rediscovery if possible (currently we do not maintain list of peers so it's not used).

MarekPieta: If we add a link between peripherals and dongle (in similar way as it's done in fwupd), we would update it

scripts/hid_configurator/NrfHidDevice.py
MarekPieta: I would prefer to get peers cache after indexing peers to prevent accidentally triggering assertion failure (the one in config_event logging implementation) on peripherals with older firmware.

MarekPieta: Also, as stated in previous comment, later on we could use peers cache only internally in the function.

applications/nrf_desktop/src/modules/hid_forward.c
```diff
@@ -813,6 +812,26 @@ static void handle_config_channel_peers_req(const struct config_event *event)
...
+
+		size_t pos = 0;
+		struct config_event *rsp = generate_response(event,
+							     CONFIG_CHANNEL_FETCHED_DATA_MAX_SIZE);
+
+		memcpy(&rsp->dyndata.data[pos], peripheral_cache, sizeof(peripheral_cache));
+		pos += sizeof(peripheral_cache);
+
+		memset(&rsp->dyndata.data[pos], 0x00, CONFIG_CHANNEL_FETCHED_DATA_MAX_SIZE - pos);
+		pos = CONFIG_CHANNEL_FETCHED_DATA_MAX_SIZE;
```
mkapala-nordic, line 828: nit: i think this line is unnecessary as `pos` is not further used here.

mkapala-nordic (approved):

applications/nrf_desktop/src/modules/hid_forward.c
MarekPieta: To be honest I preferred to update it anyway (it does not cost much and data consistency is ensured :))

applications/nrf_desktop/src/modules/hid_forward.c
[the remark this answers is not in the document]
kapi-no: oh, yes. i accidentally named in the same way as indexing variable

applications/nrf_desktop/src/modules/hid_forward.c
kapi-no: yes

scripts/hid_configurator/NrfHidDevice.py
kapi-no: It is not useless as you make many operations in your open_devices function without using the connected peers variable.

And according to cache encoding, you can deduce the number of connected peers without polling the connected peers

MarekPieta: Yes, the peers cache can be very useful for the **rediscovery**, but it would not be too useful before initial discovery (keep in mind that current implementation of the HID configurator script does not store any information about peers after script finishes execution so peers are fully rediscovered each time script is called by the user).

The only information provided by reading peers cache before the initial discovery is the number of connected peers (keep in mind that a given peer may also "move" in the peripherals array on dongle side - it may be linked with various index of the peers cache). So that actually with the current implementation, the peers cache is mainly read for debug purposes here (there is no dedicated use-case for it at this point). Still I would prefer to keep the functionality testable in scope of NCS.

kapi-no: you can keep it testable, I am just talking about two dedicated API calls instead of having one that returns two values

MarekPieta: If I do it, I would need to get peers cache unconditionally. Current implementation allows me to avoid doing it for peripherals (early return on index peers if functionality is not supported). That helps to avoid assertion failures for peripherals with older firmware. Maybe we could keep it like this? IMHO, it's nothing bad in returning 2 values - especially if they are tightly coupled like in our case.

kapi-no: ok

kapi-no (approved):

PR #11847, opened 2023-07-20
lib: date_time: use millisecond precision in ntp api

syscrypt:
Hi, 
the ntp part of the date_time lib ignores the fraction field returned by the ntp protocol, which leads to an accuracy within one second. I am aware that the date_time api docs mention that the fetched time is a UTC date time stamp so I guess this is expected behavior. But it is a little bit confusing that the api `date_time_now` method expects an `int64_t` pointer to a var called `unix_time_ms` which then has an accuracy of a second. 

Is it possible to update the ntp part to provide a millisecond accuracy as in my pr?

Regards

### DIFF
--- a/lib/date_time/date_time_ntp.c
+++ b/lib/date_time/date_time_ntp.c
@@ -89,6 +89,7 @@ static bool is_connected_to_lte(void)
 int date_time_ntp_get(int64_t *date_time_ms)
 {
 	int err;
+	uint32_t milliseconds;
 
 #if defined(CONFIG_LTE_LINK_CONTROL)
 	if (!is_connected_to_lte()) {
@@ -108,7 +109,9 @@ int date_time_ntp_get(int64_t *date_time_ms)
 			continue;
 		}
 		LOG_DBG("Time obtained from NTP server %s", servers[i]);
-		*date_time_ms = (int64_t)sntp_time.seconds * 1000;
+
+		milliseconds = (uint32_t)(((double) sntp_time.fraction / UINT32_MAX) * 1000);
+		*date_time_ms = (int64_t)sntp_time.seconds * 1000 + milliseconds;
 		return 0;
 	}
 


tokangas:
Thanks for your contribution, @syscrypt. The implementation seems to have been like this already in the first version. I don't know the reason, but I don't see why we should ignore the fraction.

Your change looks good to me, however you need to sign the CLA and the commit description needs some fine tuning to pass the compliance checks.

rlubos:
Please follow the commit message guidelines (https://docs.zephyrproject.org/latest/contribute/guidelines.html#commit-message-guidelines), in short:
* The commit message needs some change description in the body
* it needs a signoff entry

syscrypt:
Thanks for the quick turnaround, I have adjusted the commit.

lib/date_time/date_time_ntp.c
```diff
@@ -108,7 +109,9 @@ int date_time_ntp_get(int64_t *date_time_ms)
 			continue;
 		}
 		LOG_DBG("Time obtained from NTP server %s", servers[i]);
-		*date_time_ms = (int64_t)sntp_time.seconds * 1000;
+
+		milliseconds = (uint32_t)(((double) sntp_time.fraction / UINT32_MAX) * 1000);
```
trantanen, line 113: I think a comment would be nice to say that the fraction part is not milliseconds, microseconds or nanoseconds but rather fractions of a second based on 32 bit unsigned integer maximum value.
I know that's what the code does or "says" but given this is very weird unit, it would be good to have a comment alarming the reader about it.

trantanen (commented):
> But it is a little bit confusing that the api `date_time_now` method expects an `int64_t` pointer to a var called `unix_time_ms` which then has an accuracy of a second.

Zephyr has a convention of using time as `int64_t` in milliseconds so regardless of what would make sense in a particular API, this convention is still followed in many places, such as Location library API.

> Hi, the ntp part of the date_time lib ignores the fraction field returned by the ntp protocol, which leads to an accuracy within one second.

While this change makes sense and is fine for me, we will always have an accuracy issue because there is data transfer delay in returning the NTP time from the server and then getting it through the device's SW architecture. So this change removes some inaccuracy but there is still a random delay which can be seconds or even tens of seconds in some cases depending on cellular network conditions.

tokangas:
@syscrypt You still need to sign the CLA, see https://github.com/nrfconnect/sdk-nrf/pull/11847#issuecomment-1644053251.

PR #11849, opened 2023-07-21
samples: matter: Redundant FFS configurations

jnsgsr:
Remove or change configurations where FFS is not applicable for device. FFS is specific for Amazon ecosystem where there is support only for Light Bulbs (and light switch but not of type that we support), therefor only Light Bulb project is correct to have FFS feature and those configurations we should maintain.

test-sdk-nrf: PR-615


jnsgsr:
The system integration tests pass when CI_LIB branch was used https://jenkins-ncs.nordicsemi.no/job/buran/job/jaga/job/sdk-nrf/job/PR-11849/

Damian-Nordic (approved):
Can you update the PR descriptions with why those configurations have been removed? So that we know the reason in the future?

markaj-nordic (approved):

doublemis1 (approved):

PR #11853, opened 2023-07-21
applications: nrf_desktop: Change ts resolution to us in motion sim

mkapala-nordic:
Make the module rely on the cycle count and change timestamp resolution used in the Motion Simulated module from milliseconds to microseconds to increase accuracy.

Jira: NCSDK-22778

### DIFF
--- a/applications/nrf_desktop/src/hw_interface/Kconfig.motion
+++ b/applications/nrf_desktop/src/hw_interface/Kconfig.motion
@@ -80,13 +80,14 @@ config DESKTOP_MOTION_RIGHT_KEY_ID
 	  ID of key use for generating right motion.
 
 config DESKTOP_MOTION_SIMULATED_EDGE_TIME
-	int "Time for transition between two points in a trajectory [ms]"
+	int "Time for transition between two points in a trajectory [us]"
 	depends on DESKTOP_MOTION_SIMULATED_ENABLE
-	range 8 1000000
-	default 16
+	range 8 1000000000
+	default 16384
 	help
 	  The simulated movement data will be tracing predefined path, an eight-sided polygon.
 	  Must be power of two (calculations speedup).
+	  The resolution may be limited by the precision of the system's hardware clock.
 
 config DESKTOP_MOTION_SIMULATED_SCALE_FACTOR
 	int "Scale factor for given shape"
--- a/applications/nrf_desktop/src/hw_interface/motion_simulated.c
+++ b/applications/nrf_desktop/src/hw_interface/motion_simulated.c
@@ -72,12 +72,22 @@ static void motion_event_send(int16_t dx, int16_t dy)
 	APP_EVENT_SUBMIT(event);
 }
 
+static uint64_t get_timestamp(void)
+{
+	/* Use higher resolution if available. */
+	if (IS_ENABLED(CONFIG_TIMER_HAS_64BIT_CYCLE_COUNTER)) {
+		return k_cyc_to_us_near64(k_cycle_get_64());
+	} else {
+		return k_cyc_to_us_near64(k_cycle_get_32());
+	}
+}
+
 static void generate_motion_event(void)
 {
 	BUILD_ASSERT((edge_time & (edge_time - 1)) == 0,
 			 "Edge time must be power of 2");
 
-	uint32_t t = k_uptime_get_32();
+	uint64_t t = get_timestamp();
 	size_t v1_id = (t / edge_time) % ARRAY_SIZE(coords);
 	size_t v2_id = (v1_id + 1) % ARRAY_SIZE(coords);
 
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -222,6 +222,7 @@ nRF Desktop
     This is done to prevent the device from powering down and waking up multiple times when an USB cable is connected.
   * Disabled ``CONFIG_BOOT_SERIAL_IMG_GRP_HASH`` in MCUboot bootloader release configurations of boards that use nRF52820 SoC.
     This is done to reduce the memory consumption.
+  * To improve the accuracy, the generation of simulated movement data in the :ref:`nrf_desktop_motion` now uses a timestamp in microseconds based on the cycle count (either :c:func:`k_cycle_get_32` or :c:func:`k_cycle_get_64` function depending on the :kconfig:option:`CONFIG_TIMER_HAS_64BIT_CYCLE_COUNTER` Kconfig option).
 
 Thingy:53: Matter weather station
 ---------------------------------


kapi-no (approved):
do we need changelog entry for this code improvement?

applications/nrf_desktop/src/hw_interface/Kconfig.motion
```diff
@@ -80,7 +80,7 @@ config DESKTOP_MOTION_RIGHT_KEY_ID
 	  ID of key use for generating right motion.
 
 config DESKTOP_MOTION_SIMULATED_EDGE_TIME
-	int "Time for transition between two points in a trajectory [ms]"
+	int "Time for transition between two points in a trajectory [us]"
```
kapi-no, line 83: We may want to mention us resolution limitations if there are any

applications/nrf_desktop/src/hw_interface/Kconfig.motion
```diff
@@ -80,7 +80,7 @@ config DESKTOP_MOTION_RIGHT_KEY_ID
...
 	depends on DESKTOP_MOTION_SIMULATED_ENABLE
 	range 8 1000000
 	default 16
```
kapi-no, line 86: shouldn't we adjust the range and defaults here and multiply them by 1000?

applications/nrf_desktop/src/hw_interface/motion_simulated.c
```diff
@@ -72,12 +72,22 @@ static void motion_event_send(int16_t dx, int16_t dy)
 	APP_EVENT_SUBMIT(event);
 }
 
+static int64_t get_timestamp(void)
```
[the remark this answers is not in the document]
MarekPieta: Or maybe we could divide it by period for whole shape here to avoid overflows later on? (that could also allow us to properly handle overflows for 32 bit value here). Otherwise we may need to use bigger data types to prevent overflows later on.

MarekPieta (changes requested):

applications/nrf_desktop/src/hw_interface/motion_simulated.c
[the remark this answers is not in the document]
MarekPieta: I was also thinking a bit about possible overflows during calculations in `generate_motion_event` (we now use value with more bits for timestamps). Took a closer look at the implementation - it seems that modulo operation would prevent overflows later on.

applications/nrf_desktop/src/hw_interface/Kconfig.motion
mkapala-nordic: Added note about precision limitation.

applications/nrf_desktop/src/hw_interface/Kconfig.motion
mkapala-nordic: increased default value to `16ms` aligned to `2^x` and adjusted higher range by `* 1000`  so it is as big as it was before

applications/nrf_desktop/src/hw_interface/motion_simulated.c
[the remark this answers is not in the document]
mkapala-nordic: > btw. Why not `uint64_t`? Both underlying APIs rely on unsigned values

you are right, changed to `uint64_t`.



> I was also thinking a bit about possible overflows during calculations in `generate_motion_event` (we now use value with more bits for timestamps). Took a closer look at the implementation - it seems that modulo operation would prevent these overflows for bigger timestamps.
> 
> Only the system's HW clock may overflow - optionally we could improve the implementation by handling HW clock overflow in this function and make the `get_timestamp` function it return value from 0 to time of drawing whole trajectory (so that we could keep trajectory continuous also after a clock overflow).

IMO the system's HW clock overflow is a nit corner case here. 
Theoretically, if my calculations are correct, it could happen after approx. 36h of continous module work (assuming `SYS_CLOCK_HW_CYCLES_PER_SEC` is 32kHz and timestamp is in `uint32_t`).
Moreover, it possibly would be just one jump in trajectory.
I would desist from any other changes here.

mkapala-nordic:
> do we need changelog entry for this code improvement?

added short release note.

mkapala-nordic:
> do we need changelog entry for this code improvement?

added short release note

kapi-no (approved):

applications/nrf_desktop/src/hw_interface/motion_simulated.c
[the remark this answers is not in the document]
MarekPieta: I would not insist. Let's keep it as is then.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -222,6 +222,7 @@ nRF Desktop
     This is done to prevent the device from powering down and waking up multiple times when an USB cable is connected.
   * Disabled ``CONFIG_BOOT_SERIAL_IMG_GRP_HASH`` in MCUboot bootloader release configurations of boards that use nRF52820 SoC.
     This is done to reduce the memory consumption.
+  * Motion generation in the Motion Simulated module is now based on the cycle count and uses a timestamp in microseconds instead of a timestamp in milliseconds to improve accuracy.
```
MarekPieta, line 225: Add ref to module's doc page, please. also consider linking to `k_cycle_get_32` and `k_cycle_get_64` APIs to make it clear.

MarekPieta (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -222,6 +222,7 @@ nRF Desktop
     This is done to prevent the device from powering down and waking up multiple times when an USB cable is connected.
   * Disabled ``CONFIG_BOOT_SERIAL_IMG_GRP_HASH`` in MCUboot bootloader release configurations of boards that use nRF52820 SoC.
     This is done to reduce the memory consumption.
+  * Simulated movement data generation in the :ref:`nrf_desktop_motion` is now based on the cycle count (:c:func:`k_cycle_get_32` or :c:func:`k_cycle_get_64` function depending on the :kconfig:option:`CONFIG_TIMER_HAS_64BIT_CYCLE_COUNTER` Kconfig option) and uses a timestamp in microseconds instead of a timestamp in milliseconds to improve accuracy.
```
annwoj, line 225:
```suggestion
  * Generating simulated movement data in the :ref:`nrf_desktop_motion` is now based on the cycle count (either :c:func:`k_cycle_get_32` or :c:func:`k_cycle_get_64` function depending on the :kconfig:option:`CONFIG_TIMER_HAS_64BIT_CYCLE_COUNTER` Kconfig option).
  Additionally, timestamp is outputted in microseconds instead of milliseconds to improve the accuracy.
```
Rephrased a bit and split the sentence to enhance the flow, please let me know if this affects the meaning in any way.

mkapala-nordic: Information about changing timestamp to microseconds and using cycle count is connected with each other (as timestamp in us depends on the usage of cycle count) so it changes meaning a bit.

Now I think that we could drop the information of previously used "timestamp in milliseconds" to simplify it and maybe rearrange the original sentence to be more precise about the change in implementation that happened.
Something like:
```suggestion
  * The generation of simulated movement data in the :ref:`nrf_desktop_motion` now uses a timestamp in microseconds based on the cycle count (either :c:func:`k_cycle_get_32` or :c:func:`k_cycle_get_64` function depending on the :kconfig:option:`CONFIG_TIMER_HAS_64BIT_CYCLE_COUNTER` Kconfig option) to improve the accuracy.
```

What do you think?

annwoj: I think it's better :) though I still feel too much is happening at the end of the sentence. What if we moved the last part of the sentence to the beginning? In other words:
To improve the accuracy, the generation of simulated movement data in the :ref:`nrf_desktop_motion` now uses a timestamp in microseconds based on the cycle count (either :c:func:`k_cycle_get_32` or :c:func:`k_cycle_get_64` function depending on the :kconfig:option:`CONFIG_TIMER_HAS_64BIT_CYCLE_COUNTER` Kconfig option).  (?)

mkapala-nordic: Sure, even better!

annwoj (approved):

PR #11857, opened 2023-07-24
application: serial_lte_modem: Read device local time

junqingzou:
Add `#XCCLK` command to get device local time. As Zephyr does not support local time conversion, get UTC by `AT%CCLK?` then use time zone and daylight_saving info to calculate local time.

### DIFF
--- a/applications/serial_lte_modem/doc/Generic_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/Generic_AT_commands.rst
@@ -505,6 +505,61 @@ Test command
 
 The test command is not supported.
 
+Device local time #XCCLK
+========================
+
+The ``#XCCLK`` command gets the real-time clock of the device.
+
+Set command
+-----------
+
+The set command returns the local time, the time zone and daylight-saving time info.
+
+Syntax
+~~~~~~
+
+::
+
+   #XCCLK
+
+Response syntax
+~~~~~~~~~~~~~~~
+
+::
+
+   #XCCLK: <time>,<time_zone>,<daylight_saving_time>
+
+The ``<time>`` value returns a string indicating local time in the format "yy/MM/dd,hh:mm:ss+zz", where the characters, from left to right, indicate year, month, day, hour, minutes, seconds.
+
+The ``<time_zone>`` value returns an integer indicating the difference, expressed in quarters of an hour, between the local time and GMT (value range from -48 to +48 and 99 for "not set").
+
+The ``<daylight_saving_time>`` value returns an integer indicating below:
+
+* ``0`` - No adjustment of daylight-saving time.
+* ``1`` - +1 hour adjustment of daylight-saving time.
+* ``2`` - +2 hours adjustment of daylight-saving time.
+
+Example
+~~~~~~~
+
+::
+
+  AT#XCCLK
+
+  #XCCLK: "23/07/24,12:27:43",+36,0
+
+  OK
+
+Read command
+------------
+
+The read command is not supported.
+
+Test command
+------------
+
+The test command is not supported.
+
 Native TLS CMNG #XCMNG
 ======================
 
--- a/applications/serial_lte_modem/src/slm_at_commands.c
+++ b/applications/serial_lte_modem/src/slm_at_commands.c
@@ -6,12 +6,14 @@
 #include <ctype.h>
 #include <stdio.h>
 #include <string.h>
+#include <time.h>
 #include <zephyr/init.h>
 #include <zephyr/drivers/uart.h>
 #include <zephyr/kernel.h>
 #include <zephyr/logging/log_ctrl.h>
 #include <zephyr/sys/reboot.h>
 #include <zephyr/sys/util.h>
+#include <zephyr/sys/timeutil.h>
 #include <zephyr/types.h>
 #include <dfu/dfu_target.h>
 #include <modem/at_cmd_parser.h>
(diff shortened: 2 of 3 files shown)


applications/serial_lte_modem/doc/Generic_AT_commands.rst
```diff
@@ -505,6 +505,61 @@ Test command
...
+   #XCCLK
+
+Response syntax
+~~~~~~~~~~~~~~~
+
+::
+
+   #XCCLK: <time>,<time_zone>,<daylight_saving_time>
+
+The ``<time>`` value returns a string indicating local time in the format "yy/MM/dd,hh:mm:ss+zz", where the characters, from left to right, indicate year, month, day, hour, minutes, seconds.
```
wiba-nordic, line 532:
```suggestion
The ``<time>`` value returns a string indicating local time in the format "yy/MM/dd,hh:mm:ss+zz".
From left to right, the characters indicate the year, month, day, hour, minutes, and seconds.
```
The list leaves "zz" undefined.

applications/serial_lte_modem/doc/Generic_AT_commands.rst
```diff
@@ -505,6 +505,61 @@ Test command
...
+
+The ``<time_zone>`` value returns an integer indicating the difference, expressed in quarters of an hour, between the local time and GMT (value range from -48 to +48 and 99 for "not set").
```
wiba-nordic, line 534:
```suggestion
The ``<time_zone>`` value returns an integer indicating the difference between the local time and GMT.
The difference is expressed in quarters of an hour, so the value ranges from ``-48`` to ``+48``, with the value ``99`` used to indicate the difference has not been set.
```

applications/serial_lte_modem/doc/Generic_AT_commands.rst
```diff
@@ -505,6 +505,61 @@ Test command
...
+
+The ``<daylight_saving_time>`` value returns an integer indicating below:
```
wiba-nordic, line 536:
```suggestion
The ``<daylight_saving_time>`` value returns an integer indicating one of the following:
```

applications/serial_lte_modem/doc/Generic_AT_commands.rst
```diff
@@ -505,6 +505,61 @@ Test command
...
+
+* ``0`` - No adjustment of daylight-saving time.
+* ``1`` - +1 hour adjustment of daylight-saving time.
+* ``2`` - +2 hours adjustment of daylight-saving time.
```
wiba-nordic, line 540:
```suggestion
* ``0`` - No adjustment for daylight saving time.
* ``1`` - +1 hour adjustment for daylight saving time.
* ``2`` - +2 hours adjustment for daylight saving time.
```
Grammatically daylight-saving time would be correct, but the convention is to write it without the dash.

wiba-nordic (changes requested):

VTPeltoketo:
Clear and straightforward addition. Firstly I just wondered why the user cannot use AT+CCLK and make the transform by him/herself. But then I noticed, that our CCLK is not standard one because the time zone is missing. Do you know why?
Moreover, the daylight mode is not supported in CCLK according to the standard 3GPP TS 27.007 V18.3.0 (2023-06)

Summa summarum: Do we need this feature in SLM or can we dig the local time from existing AT-commands?  

Current AT command using modem version 1.3.5:
AT+CCLK?
+CCLK: "23/07/31,07:31:47"

junqingzou:
> Clear and straightforward addition. Firstly I just wondered why the user cannot use AT+CCLK and make the transform by him/herself. But then I noticed, that our CCLK is not standard one because the time zone is missing. Do you know why? Moreover, the daylight mode is not supported in CCLK according to the standard 3GPP TS 27.007 V18.3.0 (2023-06)
> 
> Summa summarum: Do we need this feature in SLM or can we dig the local time from existing AT-commands?
> 
> Current AT command using modem version 1.3.5: AT+CCLK? +CCLK: "23/07/31,07:31:47"

It should, unless that network does not give the info to UE right after attach.
at+cclk?
+CCLK: "23/07/31,08:16:01+36"
OK

What's missing is daylight_saving_time (DST), which is added in Nordic proprietary %CCLK.

Zephyr does not support local date time as it has no time zone info.

VTPeltoketo:
ok, but when the %CCLK has all the needed information, do we need #XCCLK because the local time calculation is so simple?

junqingzou:
> ok, but when the %CCLK has all the needed information, do we need #XCCLK because the local time calculation is so simple?
If MCU is not Zephyr-based, we calculate for MCU with Zephyr API.
Originaly prepared as reference code for SoC-based application, i.e. how to calculate local datetime based on modem info.

ChanduBoyapati:
Hi Guys,
As per customer requirement, they need local time instead of UTC time. In India, we don't have day light saving mode and this command looks fine as per customer requirement. Our competitors have local time option in their modem. So, better we have this option in SLM and if required we will keep as internal command.

VTPeltoketo:
Just to clarify the need to myself: The UTC-local time conversion is very simple, just two lines of mathematics.

Is the root reason functions timeutil_timegm() and gmtime_r() which we can easily use inside SLM, but maybe not by the customer?

junqingzou:
> timeutil_timegm

API timeutil_timegm() is Zephyr time represemtation API,

junqingzou:
> 

API timeutil_timegm() is Zephyr time represemtation API, gmtime_r() is standard libc API. They are not the root reason, however, as any SoC app nRF91 could call them. For external MCU, there could be couterpart APIs to do the conversion to and from standard C time_t structure.

Since mode already has +CCLK and %CCLK, let's drop the idea of another #XCCLK to avoid confusion.

@ChanduBoyapati please ask your customer to do the calculation on MCU side following similar logic. The nrf_modem_at_scanf() could be replaced by sscanf() using the %CCLK? response as input.

VTPeltoketo:
Ok, thank you

PR #11862, opened 2023-07-24
samples: crypto: psa_tls: Don't test secure-only nrf91

SebastianBoe:
Don't test secure-only nrf91 with psa_tls. NCS applications must use non-secure nrf91 due to a HW limitation in nrf91.

For this reason all nrf9160 samples test only against nrf91_ns and not nrf91 secure-only.

The only reason to test secure-only would be to test a library that would be used by a bootloader, but I can't imagine a bootloader would be doing psa_tls so skip this testing.

Omitting this testing will reduce CI load and decrease maintenance cost.


mswarowsky (approved):
Good idea

Vge0rge (approved):
I added @stephen-nordic  here since it's useful for him to know.

SebastianBoe:
waiting for review from @magnev

SebastianBoe:
this change was not very popular

PR #11881, opened 2023-07-26
samples: cellular: modem_shell: Use snippet for enabling UART traces

tokangas:
Updated sample.yaml to use nrf91-modem-trace-uart snippet for enabling modem UART traces.

Updated disabling of UART1 to use Zephyr PM API when the Zephyr UART trace backend is used.

Fixed possible deadlock when disabling UART1 using direct register access.


moksanen (approved):
lgtm

trantanen (approved):

PR #11886, opened 2023-07-26
TF-M: Fix and document TF-M build issue with printf and heap

joerchan:
Fix undefined reference to 'end' when disabling secure output from the
TF-M image.

When secure output is disabled the printf function is not overriden
in TF-M and the libc variant has references to heap functions.
Make sure that the printf function is overriden for this case also and
drop the output message.

NCSDK-22818

### DIFF
--- a/doc/nrf/releases_and_maturity/known_issues.rst
+++ b/doc/nrf/releases_and_maturity/known_issues.rst
@@ -3579,6 +3579,11 @@ Trusted Firmware-M (TF-M)
 
 The issues in this section are related to the TF-M implementation in the |NCS|.
 
+.. rst-class:: v2-4-0
+
+NCSDK-22818: TF-M does not build with :kconfig:option:`CONFIG_TFM_PROFILE_TYPE_MINIMAL` disabled and :kconfig:option:`CONFIG_TFM_LOG_LEVEL_SILENCE` enabled
+  This fails with the error: message "undefined reference to 'end'" caused by printf being included in the build.
+
 .. rst-class:: v2-4-1 v2-4-0 v2-3-0 v2-2-0 v2-1-4 v2-1-3 v2-1-2 v2-1-1 v2-1-0 v2-0-2 v2-0-1 v2-0-0 v1-9-2 v1-9-1 v1-9-0 v1-8-0 v1-7-1 v1-7-0 v1-6-1 v1-6-0 v1-5-2 v1-5-1 v1-5-0
 
 NCSDK-22629: TF-M might fail to reset when using nrfjprog version 10.22.x on nRF9160 platforms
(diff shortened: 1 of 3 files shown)


doc/nrf/releases_and_maturity/known_issues.rst
```diff
@@ -3550,6 +3550,11 @@ Trusted Firmware-M (TF-M)
 
 The issues in this section are related to the TF-M implementation in the |NCS|.
 
+.. rst-class:: v2-4-0
+
+NCSDK-22818: TF-M does not build with :kconfig:option:`CONFIG_TFM_PROFILE_TYPE_MINIMAL` disabled and :kconfig:option:`CONFIG_TFM_LOG_LEVEL_SILENCE` enabled.
+  This will fail with the error: message "undefined reference to 'end'" caused by printf being included in the build
```
melwee01, line 3556:
```suggestion
  This fails with the error: message "undefined reference to 'end'" caused by printf being included in the build.
```

Can we change those single quotation marks ' ' to backticks ` ` ? Looks a bit odd as is.

doc/nrf/releases_and_maturity/known_issues.rst
```diff
@@ -3550,6 +3550,11 @@ Trusted Firmware-M (TF-M)
...
+
 .. rst-class:: v2-4-1 v2-4-0 v2-3-0 v2-2-0 v2-1-4 v2-1-3 v2-1-2 v2-1-1 v2-1-0 v2-0-2 v2-0-1 v2-0-0 v1-9-2 v1-9-1 v1-9-0 v1-8-0 v1-7-1 v1-7-0 v1-6-1 v1-6-0 v1-5-2 v1-5-1 v1-5-0
 
 NCSDK-22629: TF-M might fail to reset when using nrfjprog version 10.22.x on nRF9160 platforms
```
melwee01, line 3560:
```suggestion
NCSDK-22629: TF-M might fail to reset when using nrfjprog version 10.22.x on nRF9160 platforms.
```

melwee01 (commented):
Nothing I'd consider big enough to block merging.

doc/nrf/releases_and_maturity/known_issues.rst
joerchan: Not sure that is wise, backticks are intepreted differently and could have an affect on the doc build.

doc/nrf/releases_and_maturity/known_issues.rst
joerchan: Ok, I can fix this up also.

melwee01 (approved):

lemrey (approved):

PR #11888, opened 2023-07-26
wifi: modules: hostap: Notify connection transient states

rado17:
Pull changes to notify transient states while staion is connecting to a network.


modules/hostap/src/supp_events.c
```diff
@@ -1,18 +1,112 @@
...
+			break;
+		}
+	}
+
+	event_data->event = event;
+
+	switch (event) {
+	case CONNECTED:
+		mac = data->connected.bssid;
+		ret = sscanf(status+strlen("CTRL-EVENT-CONNECTED - Connection to"),
```
krish2718, line 43: where are the checks for `ret`?

rado17: Done.

krish2718 (approved):

sachinthegreen (approved):

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,158 @@
...
+
+	data = (union supp_mon_event_data *)event_data->data;
+
+	for (int i = 0; i < ARRAY_SIZE(event_map); i++) {
+		if (strncmp(status, event_map[i], strlen(event_map[i])) == 0) {
+			event = i;
+			break;
+		}
+	}
+
```
krish2718, line 47: No check if the event isn't found?

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,158 @@
 #include "includes.h"
 #include "common.h"
 
-int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event, int status)
+#define MAC_STR_FORMAT "%2hhx:%2hhx:%2hhx:%2hhx:%2hhx:%2hhx"
+
+char *event_map[] = {
```
krish2718, line 14: static and also better to use proper namespaces i.e., prefix here

modules/hostap/src/supp_events.h
```diff
@@ -23,7 +23,9 @@ enum net_event_wpa_supp_cmd {
 	NET_EVENT_WPA_SUPP_CMD_IFACE_ADDED,
 	NET_EVENT_WPA_SUPP_CMD_IFACE_REMOVING,
 	NET_EVENT_WPA_SUPP_CMD_IFACE_REMOVED,
-	NET_EVENT_WIFI_CMD_MAX
+	NET_EVENT_WPA_SUPP_CMD_COUNT,
+	NET_EVENT_WPA_SUPP_PRIV_START = NET_EVENT_WPA_SUPP_CMD_COUNT,
```
krish2718, line 27: We don't need `priv` concept here, its only needed to abstract NM in upstream.

krish2718 (changes requested):

modules/hostap/src/supp_events.c
rado17: Done

modules/hostap/src/supp_events.c
rado17: Done

modules/hostap/src/supp_events.h
rado17: Removed priv.

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,164 @@
...
+	}
+
+	event_data->event = event + NET_EVENT_WPA_SUPP_CMD_COUNT + 1;
+
+	switch (event_data->event) {
+	case WPA_SUPP_CONNECTED:
+		mac = data->connected.bssid;
+		ret = sscanf(status+strlen("CTRL-EVENT-CONNECTED - Connection to"),
+			MAC_STR_FORMAT, &mac[0], &mac[1], &mac[2], &mac[3], &mac[4], &mac[5]);
+		if (ret <= 0) {
```
krish2718, line 60: I guess this can be moved outside switch case, no?

modules/hostap/src/supp_events.h
```diff
@@ -41,7 +42,99 @@ enum net_event_wpa_supp_cmd {
...
+
+	struct supp_event_network_removed {
+		unsigned int id;
+	} network_removed;
+
+	char supp_event_str[NM_WIFI_EVENT_STR_LEN];
+};
+
+enum supp_event_num {
+	WPA_SUPP_WIFI_CMD_NM_EVENT = NET_EVENT_WPA_SUPP_CMD_COUNT,
```
krish2718, line 113: The prefix should be WPA_SUPP_EVENT

modules/hostap/src/supp_events.h
```diff
@@ -41,7 +42,99 @@ enum net_event_wpa_supp_cmd {
...
-int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event, int status);
+int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event, void *status,
+			size_t len);
 int generate_supp_state_event(const char *ifname, enum net_event_wpa_supp_cmd event, int status);
 
+#define REASON_CODE_LEN 32
+#define NM_WIFI_EVENT_STR_LEN 64
+#define ETH_ALEN 6
+
+union supp_mon_event_data {
```
krish2718, line 53: drop the `mon` to sync with other names

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,164 @@
...
+		break;
+	case WPA_SUPP_TERMINATING:
+	case WPA_SUPP_SCAN_STARTED:
+	case WPA_SUPP_SCAN_FAILED:
+	case WPA_SUPP_NETWORK_NOT_FOUND:
+	case WPA_SUPP_NETWORK_ADDED:
+	case WPA_SUPP_NETWORK_REMOVED:
+		strncpy(data->supp_event_str, wpa_supp_event_map[event],
+				 sizeof(data->supp_event_str));
+		event_data->data_len = strlen(data->supp_event_str);
```
krish2718, line 154: what about null terminator?

modules/hostap/src/supp_events.h
```diff
@@ -41,7 +42,99 @@ enum net_event_wpa_supp_cmd {
 #define NET_EVENT_WPA_SUPP_IFACE_REMOVING					\
 	(_NET_MGMT_WPA_SUPP_EVENT | NET_EVENT_WPA_SUPP_CMD_IFACE_REMOVING)
 
-int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event, int status);
+int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event, void *status,
+			size_t len);
 int generate_supp_state_event(const char *ifname, enum net_event_wpa_supp_cmd event, int status);
 
+#define REASON_CODE_LEN 32
```
krish2718, line 49: Currently, `NO_PSK_AVAILABLE` is the longest string, so, should we just limit to 17? We also need to add a note here to update this if we add a new call to `wpas_auth_failed`

krish2718: or just add a enum here and handle the string to enum converion in an helper, this way we abstract the internal implementation.

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,164 @@
...
+		if (ret <= 0) {
+			wpa_printf(MSG_ERROR, "%s Parse failed: %s",
+				  wpa_supp_event_map[event], strerror(errno));
+			break;
+		}
+		event_data->data_len = sizeof(data->connected);
+		break;
+	case WPA_SUPP_DISCONNECTED:
+		mac = data->disconnected.bssid;
+		ret = sscanf(status+strlen("CTRL-EVENT-DISCONNECTED bssid="),
```
krish2718, line 69: spaces around `+`

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,164 @@
...
+			break;
+		}
+		event_data->data_len = sizeof(data->disconnected);
+		break;
+	case WPA_SUPP_ASSOC_REJECT:
+		/* TODO */
+		break;
+	case WPA_SUPP_AUTH_REJECT:
+		mac = data->auth_reject.bssid;
+		ret = sscanf(status+strlen("CTRL-EVENT-AUTH-REJECT "), MAC_STR_FORMAT
```
krish2718, line 84: spaces around `+`

modules/hostap/src/supp_events.h
```diff
@@ -23,7 +23,8 @@ enum net_event_wpa_supp_cmd {
 	NET_EVENT_WPA_SUPP_CMD_IFACE_ADDED,
 	NET_EVENT_WPA_SUPP_CMD_IFACE_REMOVING,
 	NET_EVENT_WPA_SUPP_CMD_IFACE_REMOVED,
-	NET_EVENT_WIFI_CMD_MAX
+	NET_EVENT_WPA_SUPP_CMD_COUNT,
```
krish2718, line 26:
```
/* Has to be the last valid event */
NET_EVENT_WPA_SUPP_INT_EVENT
```

modules/hostap/src/supp_events.h
```diff
@@ -41,7 +42,99 @@ enum net_event_wpa_supp_cmd {
...
+	WPA_SUPP_SCAN_RESULTS,
+	WPA_SUPP_SCAN_FAILED,
+	WPA_SUPP_BSS_ADDED,
+	WPA_SUPP_BSS_REMOVED,
+	WPA_SUPP_NETWORK_NOT_FOUND,
+	WPA_SUPP_NETWORK_ADDED,
+	WPA_SUPP_NETWORK_REMOVED
+};
+
+#define NET_EVENT_WPA_SUPP_NM_EVENT					\
```
krish2718, line 131: Group this with other events and drop the NM, its only WPA supplicant here.

modules/hostap/src/supp_events.h
```diff
@@ -41,7 +42,99 @@ enum net_event_wpa_supp_cmd {
...
+	(_NET_MGMT_WPA_SUPP_EVENT | WPA_SUPP_WIFI_CMD_NM_EVENT)
+
+struct supp_status_data {
+	enum supp_event_num event;
+	void *data;
+	int data_len;
```
krish2718, line 137: size_t

modules/hostap/src/supp_events.h
```diff
@@ -41,7 +42,99 @@ enum net_event_wpa_supp_cmd {
...
+	WPA_SUPP_BSS_REMOVED,
+	WPA_SUPP_NETWORK_NOT_FOUND,
+	WPA_SUPP_NETWORK_ADDED,
+	WPA_SUPP_NETWORK_REMOVED
+};
+
+#define NET_EVENT_WPA_SUPP_NM_EVENT					\
+	(_NET_MGMT_WPA_SUPP_EVENT | WPA_SUPP_WIFI_CMD_NM_EVENT)
+
+struct supp_status_data {
```
krish2718, line 134: This isn't `status`, so, `supp_int_event_data` ?

krish2718 (changes requested):

modules/hostap/src/supp_events.c
rado17: Done

modules/hostap/src/supp_events.h
rado17: Done

modules/hostap/src/supp_events.h
rado17: Done

modules/hostap/src/supp_events.c
rado17: Added 1 byte for null terminator

modules/hostap/src/supp_events.c
rado17: Done

modules/hostap/src/supp_events.c
rado17: Done

modules/hostap/src/supp_events.h
rado17: Done

modules/hostap/src/supp_events.h
rado17: Done

modules/hostap/src/supp_events.h
rado17: Done

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,134 @@
...
+		strncpy(data->supp_event_str, wpa_supp_event_map[event],
+				 sizeof(data->supp_event_str));
+		event_data->data_len = strlen(data->supp_event_str) + 1;
+	default:
+		break;
+	}
+
+	if (ret <= 0) {
+		wpa_printf(MSG_ERROR, "%s Parse failed: %s",
+				wpa_supp_event_map[event_data->event], strerror(errno));
```
krish2718, line 126: aligment seems off.

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,134 @@
...
+}
+
+int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event,
+		 void *status, size_t len)
 {
 	const struct device *dev = device_get_binding(ifname);
 	struct net_if *iface = net_if_lookup_by_dev(dev);
+	union supp_event_data data;
+	struct supp_int_event_data event_data;
+	int ret;
```
krish2718, line 139: move this to the case where its used.

modules/hostap/src/supp_events.c
```diff
@@ -38,7 +174,6 @@ int generate_supp_state_event(const char *ifname, enum net_event_wpa_supp_cmd ev
 {
 	/* TODO: Replace device_get_binding. */
 	const struct device *dev = device_get_binding(ifname);
-
```
krish2718, line 41: unncessary whitespace, if you want to fix the warning that empty line after declarations then `iface` also needs to be cubbed here.

modules/hostap/src/supp_events.h
```diff
@@ -23,7 +23,9 @@ enum net_event_wpa_supp_cmd {
 	NET_EVENT_WPA_SUPP_CMD_IFACE_ADDED,
 	NET_EVENT_WPA_SUPP_CMD_IFACE_REMOVING,
 	NET_EVENT_WPA_SUPP_CMD_IFACE_REMOVED,
-	NET_EVENT_WIFI_CMD_MAX
+	NET_EVENT_WPA_SUPP_CMD_COUNT,
+	NET_EVENT_WPA_SUPP_CMD_MAX,
+	NET_EVENT_WPA_SUPP_CMD_INT_EVENT
```
krish2718, line 28: This has to be before the MAX and you can remove `NET_EVENT_WPA_SUPP_CMD_COUNT`.

modules/hostap/src/supp_events.c
```diff
@@ -21,10 +145,22 @@ int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event, int
...
+		wifi_mgmt_raise_connect_result_event(iface, *(int *)status);
 		break;
 	case NET_EVENT_WIFI_CMD_DISCONNECT_RESULT:
-		wifi_mgmt_raise_disconnect_result_event(iface, status);
+		wifi_mgmt_raise_disconnect_result_event(iface, *(int *)status);
+		break;
+	case NET_EVENT_WPA_SUPP_CMD_INT_EVENT:
+		event_data.data = &data;
+		ret = process_status(&event_data, (char *)status);
+		if (ret >= 0) {
```
krish2718, line 156: `0` is considered failure in `process_status`

modules/hostap/src/supp_events.c
```diff
@@ -21,10 +145,22 @@ int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event, int
...
+			net_mgmt_event_notify_with_info(NET_EVENT_WPA_SUPP_INT_EVENT,
+						iface, &event_data, sizeof(event_data));
+		} else {
+			if (ret != -ENOTSUP) {
+				wpa_printf(MSG_ERROR, "NM event notification failed\n");
```
krish2718, line 161: if you can move this log `process_status` and modify to say `WPA supplicant event is not supported`, then we can remove this branch. Please use WPA_S/WPA_SUPP instead of NM, NM was meant to be used in upstream zephyr to cover different types of NMs.

modules/hostap/src/supp_events.h
krish2718: my comment is applicable for all events in the enum, not just first one, sorry if its unclear.

krish2718 (changes requested):

modules/hostap/src/supp_events.c
rado17: This is removal of unnecessary whitespace.

modules/hostap/src/supp_events.c
rado17: This is to keep the difference between events not supported and the events which are supported but their parsing failed. Will replace NM with WPA_SUPP.

krish2718: Sure, but my suggestion of moving this to `process_status` still applies.

rado17: Right. Will remove this log. Eror case in `process_status` is already being logged.

modules/hostap/src/supp_events.h
rado17: Since we are reading reason code as string from event, I think adding enum to string translation is not required here. Will reduce the array size to 18 bytes.

modules/hostap/src/supp_events.c
rado17: Rectified.

modules/hostap/src/supp_events.c
krish2718: Can you please fix this in another commit? The whitespace here is needed but we need club declarations together before that.

modules/hostap/src/supp_api.c
```diff
@@ -121,8 +122,9 @@ static void supp_shell_connect_status(struct k_work *work)
 	if (ctrl->requested_op == CONNECT && wpa_s->wpa_state != WPA_COMPLETED) {
 		if (ctrl->connection_timeout > 0 && seconds_counter++ > ctrl->connection_timeout) {
 			_wpa_cli_cmd_v("disconnect");
+			conn_result = -ETIMEDOUT;
 			send_wifi_mgmt_event(wpa_s->ifname, NET_EVENT_WIFI_CMD_CONNECT_RESULT,
-				-ETIMEDOUT);
+						(void *)&conn_result, sizeof(int));
```
krish2718, line 127: alignment seems off

modules/hostap/src/supp_events.c
krish2718: not for `-ENOTSUP`

modules/hostap/src/supp_events.c
rado17: Actually, this was an inadvertent change, reverting this bit.

modules/hostap/src/supp_events.c
rado17: Logged the -ENOTSUP as well case in `process_status` .

modules/hostap/src/supp_api.c
rado17: I think function call overflowing to next line should be aligned with the left parentheses.

modules/hostap/src/supp_events.c
```diff
@@ -9,10 +9,134 @@
...
+
+	for (i = 0; i < ARRAY_SIZE(wpa_supp_event_map); i++) {
+		if (strncmp(status, wpa_supp_event_map[i], strlen(wpa_supp_event_map[i])) == 0) {
+			event = i;
+			break;
+		}
+	}
+
+	if (i >= ARRAY_SIZE(wpa_supp_event_map)) {
+		wpa_printf(MSG_ERROR, "Event not supported\n");
```
krish2718, line 50: better to log the event as well for debug.

modules/hostap/src/supp_events.c
```diff
@@ -21,10 +145,20 @@ int send_wifi_mgmt_event(const char *ifname, enum net_event_wifi_cmd event, int
...
 	case NET_EVENT_WIFI_CMD_CONNECT_RESULT:
-		wifi_mgmt_raise_connect_result_event(iface, status);
+		wifi_mgmt_raise_connect_result_event(iface, *(int *)status);
 		break;
 	case NET_EVENT_WIFI_CMD_DISCONNECT_RESULT:
-		wifi_mgmt_raise_disconnect_result_event(iface, status);
+		wifi_mgmt_raise_disconnect_result_event(iface, *(int *)status);
+		break;
+	case NET_EVENT_WPA_SUPP_CMD_INT_EVENT:
+		int ret;
```
krish2718, line 154: do we even need `ret` now?

krish2718 (changes requested):

modules/hostap/src/supp_events.c
rado17: Done

krish2718 (approved):

PR #11893, opened 2023-07-27
lib: sms: Reinitialize SMS registration after offline mode

trantanen:
When LTE is activated, reinitialize SMS registration with the LTE network if it had been registered earlier.

Jira: NCSIDB-1060

### DIFF
--- a/lib/sms/sms.c
+++ b/lib/sms/sms.c
@@ -11,6 +11,9 @@
 #include <errno.h>
 #include <modem/at_monitor.h>
 #include <nrf_modem_at.h>
+#if defined(CONFIG_LTE_LINK_CONTROL)
+#include <modem/lte_lc.h>
+#endif
 
 #include "sms_submit.h"
 #include "sms_deliver.h"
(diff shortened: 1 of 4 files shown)


lib/sms/sms.c
```diff
@@ -241,7 +244,7 @@ static int sms_init(void)
 	}
 	/* Parameters 1 to 4 should be 0 if no client is registered. */
 	if (cnmi_value1 != 0 || cnmi_value2 != 0 || cnmi_value3 != 0 || cnmi_value4 != 0) {
-		LOG_ERR("Only one SMS client can be registered");
+		LOG_WRN("Only one SMS client can be registered");
```
tokangas, line 247: You could check here if `sms_client_registered` is already `true`, in which case it's expected that +CNMI value is not `0,0,0,0` and the function could return `0`. Otherwise print an error and return `-EBUSY`. That would also simplify the condition on line 396.

trantanen: I made this change and added a debug log when the registration is in place so we know what happens.
I'll still add some unit test.

tokangas (approved):
Looks good to me.

jhirsi (approved):

PR #11900, opened 2023-07-28
nRF Cloud Multi Service Sample

plskeggs:
This integrates CoAP support into the nRF Cloud MQTT Multi Service Sample, with the goal of it becoming the nRF Cloud Multi Service Sample instead.  Eventually this will also move to the /net folder instead of /cellular once Wi-Fi support is integrated.


samples/cellular/nrf_cloud_mqtt_multi_service/overlay_coap.conf
```diff
@@ -0,0 +1,70 @@
...
+
+# Location Services configuration
+CONFIG_LOCATION_DATA_DETAILS=y
+
+# Date Time lib - Used by PGPS and main application
+CONFIG_DATE_TIME_AUTO_UPDATE=y
+CONFIG_DATE_TIME_MODEM=y
+CONFIG_DATE_TIME_NTP=y
+
+CONFIG_NRF_CLOUD_COAP_SERVER_HOSTNAME="coap.dev.nrfcloud.com"
```
plskeggs, line 68: TODO: change this to prod

samples/cellular/nrf_cloud_mqtt_multi_service/prj.conf
```diff
@@ -7,7 +7,7 @@
 # Log level
 # For more verbose and detailed log output, set the log level to
 # CONFIG_MQTT_MULTI_SERVICE_LOG_LEVEL_DBG=y instead.
-CONFIG_MQTT_MULTI_SERVICE_LOG_LEVEL_INF=y
+CONFIG_MULTI_SERVICE_LOG_LEVEL_DBG=y
```
plskeggs, line 10: TODO: revert this

samples/cellular/nrf_cloud_mqtt_multi_service/prj.conf
```diff
@@ -65,6 +65,7 @@ CONFIG_AT_HOST_LIBRARY=y
 
 # nRF Cloud
 CONFIG_NRF_CLOUD_MQTT=y # This also enables FOTA by implicitly setting CONFIG_NRF_CLOUD_FOTA=y
+CONFIG_NRF_CLOUD_FOTA=y
```
plskeggs, line 68: TODO:  remove, but update docs to indicate it can now be disabled (just like the other services this sample uses can already be disabled)

samples/cellular/nrf_cloud_mqtt_multi_service/prj.conf
```diff
@@ -84,7 +85,8 @@ CONFIG_NRF_CLOUD_LOCATION=y
 CONFIG_MODEM_INFO=y
 CONFIG_MODEM_INFO_ADD_NETWORK=y
 CONFIG_NRF_CLOUD_PGPS=y
-CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD=4
+CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD=2
+CONFIG_NRF_CLOUD_PGPS_NUM_PREDICTIONS=4
```
plskeggs, line 89: TODO: revert

samples/cellular/nrf_cloud_mqtt_multi_service/prj.conf
```diff
@@ -116,3 +118,9 @@ CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=n
 CONFIG_NRF_CLOUD_ALERT=y
 CONFIG_NRF_CLOUD_LOG_DIRECT=y
 CONFIG_NRF_CLOUD_LOG_OUTPUT_LEVEL=3
+
+CONFIG_STACK_SENTINEL=y
+CONFIG_HW_STACK_PROTECTION=y
+CONFIG_DEBUG_COREDUMP=y
+CONFIG_DEBUG_COREDUMP_BACKEND_LOGGING=y
+CONFIG_DEBUG_COREDUMP_MEMORY_DUMP_MIN=y
```
plskeggs, line 126: TODO: revert

samples/cellular/nrf_cloud_mqtt_multi_service/sample.yaml
```diff
@@ -1,10 +1,18 @@
...
       - thingy91_nrf9160_ns
     tags: ci_build
+  sample.nrf9160.nrf_cloud_multi_service.coap:
+    build_only: true
+    platform_allow: nrf9160dk_nrf9160_ns thingy91_nrf9160_ns
+    integration_platforms:
+      - nrf9160dk_nrf9160_ns
+      - thingy91_nrf9160_ns
+    extra_args: OVERLAY_CONFIG=overlay_coap.conf
+    tags: ci_build
```
plskeggs, line 18: TODO: add other entries here so other overlays we provide and support will be built by CI (e.g., FMFU)

samples/cellular/nrf_cloud_mqtt_multi_service/src/application.c
```diff
@@ -187,6 +107,7 @@ static void on_location_update(const struct location_event_data * const location
 	}
 }
 
+#if 0 // defined(CONFIG_AT_CMD_REQUESTS)
```
plskeggs, line 110: TODO: remove // here

samples/cellular/nrf_cloud_mqtt_multi_service/src/application.c
```diff
@@ -269,6 +190,44 @@ static void handle_at_cmd_requests(const struct nrf_cloud_data *const dev_msg)
...
+static void handle_at_cmd_requests(const struct nrf_cloud_data *const dev_msg)
+{
+}
+#endif /* CONFIG_AT_CMD_REQUESTS */
+
+#if defined(CONFIG_NRF_CLOUD_COAP)
+static void check_shadow(void)
+{
+	int err;
+	char buf[512];
```
plskeggs, line 203: TODO: change 512 to a # define

samples/cellular/nrf_cloud_mqtt_multi_service/src/application.c
```diff
@@ -343,20 +303,40 @@ void main_application_thread_fn(void)
...
-			LOG_INF("Sent test counter = %d", counter);
-			(void)send_sensor_sample("COUNT", counter++);
+			struct nrf_cloud_sensor_data data = {
+				.app_id = "COUNT",
+				.data_type = NRF_CLOUD_DATA_TYPE_INT,
+				.int_val = counter++
+			};
+
+			LOG_INF("Sent test counter = %d", data.int_val);
+			(void)send_device_message(&data);
```
plskeggs, line 329: TODO: test with CoAP

samples/cellular/nrf_cloud_mqtt_multi_service/src/application.c
```diff
@@ -343,20 +303,40 @@ void main_application_thread_fn(void)
...
+		}
+
+#if defined(CONFIG_NRF_CLOUD_COAP)
+#if defined(CONFIG_NRF_CLOUD_COAP_FOTA)
+		if (handle_fota_process() != -EAGAIN) {
```
plskeggs, line 334: TODO: allow this and shadow checking to be called after multiple sensor updates (lower frequency).

Alternatively, rework this loop to use a timer or deferred worker to sample everything at a separately configurable rate.

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c
```diff
@@ -237,7 +289,7 @@ int nrf_cloud_coap_location_get(struct nrf_cloud_rest_location_request const *co
 	if (!nrf_cloud_coap_is_connected()) {
 		return -EACCES;
 	}
-	static uint8_t buffer[256];
+	static uint8_t buffer[1024];
```
plskeggs, line 292: TODO: use a # define here

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
```diff
@@ -143,8 +154,20 @@ static bool connect_cloud(void)
...
 
+	int err;
+
+#if defined(CONFIG_NRF_CLOUD_MQTT)
 	/* Connect to nRF Cloud -- Non-blocking. State updates are handled in callbacks. */
-	int err = nrf_cloud_connect();
+	err = nrf_cloud_connect();
+#elif defined(CONFIG_NRF_CLOUD_COAP)
+	/* Connect via CoAP -- blocking. */
+	err = nrf_cloud_coap_connect();
```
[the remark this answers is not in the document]
glarsennordic: I guess this would require having `nrf_cloud_init` work for coap as well, but that seems doable

include/net/nrf_cloud_codec.h
```diff
@@ -48,12 +49,42 @@ enum nrf_cloud_enc_src {
 	NRF_CLOUD_ENC_SRC_PRE_ENCODED,
 };
 
+/** @brief Data types for nrf_cloud_sensor_data. */
+enum nrf_cloud_data_type {
```
jayteemo, line 53: can you add a type for none/un-initialized?
e.g.:  `NRF_CLOUD_DATA_TYPE_NONE`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -157,6 +162,31 @@ int nrf_cloud_obj_msg_init(struct nrf_cloud_obj *const obj, const char *const ap
...
+		}
+
+		size_t app_id_sz = strlen(app_id) + 1;
+		struct nrf_cloud_obj_coap_cbor *msg = nrf_cloud_calloc(1, sizeof(*msg));
+
+		if (!msg) {
+			return -ENOMEM;
+		}
+
+		/* Add the app_id */
```
jayteemo, line 178: initialize`msg->type` 
e.g.: `msg->type = NRF_CLOUD_DATA_TYPE_NONE;`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -673,6 +834,26 @@ int nrf_cloud_obj_pvt_add(struct nrf_cloud_obj *const obj,
 
 		return nrf_cloud_pvt_data_encode(pvt, obj->json);
 	}
+	case NRF_CLOUD_OBJ_TYPE_COAP_CBOR:
+	{
+		if (!obj->coap_cbor) {
+			return -ENOENT;
+		}
+
+		if (obj->coap_cbor->pvt) {
```
jayteemo, line 843: instead, check the union type to see if it has already been set:
e.g.: `if (obj->coap_cbor->type != NRF_CLOUD_DATA_TYPE_NONE)`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -673,6 +834,26 @@ int nrf_cloud_obj_pvt_add(struct nrf_cloud_obj *const obj,
...
+			return -ENOTEMPTY;
+		}
+
+		obj->coap_cbor->pvt = nrf_cloud_calloc(1, sizeof(*obj->coap_cbor->pvt));
+
+		if (!obj->coap_cbor->pvt) {
+			return -ENOMEM;
+		}
+
+		*obj->coap_cbor->pvt = *pvt;
```
jayteemo, line 853: set type:
`obj->coap_cbor->type = NRF_CLOUD_DATA_TYPE_PVT;`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -376,19 +467,44 @@ int nrf_cloud_obj_num_add(struct nrf_cloud_obj *const obj, const char *const key
...
+		}
+
+		size_t str_sz = strlen(val) + 1;
+		obj->coap_cbor->str_val = nrf_cloud_calloc(1, str_sz);
+
+		if (!obj->coap_cbor->str_val) {
+			return -ENOMEM;
+		}
+
+		memcpy(obj->coap_cbor->str_val, val, str_sz);
```
jayteemo, line 504: set type:
`obj->coap_cbor->type = NRF_CLOUD_DATA_TYPE_STR;`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -376,19 +467,44 @@ int nrf_cloud_obj_num_add(struct nrf_cloud_obj *const obj, const char *const key
...
 		return cJSON_AddStringToObjectCS(dest_json_get(obj, data_child),
 						 key, val) ? 0 : -ENOMEM;
 	}
+	case NRF_CLOUD_OBJ_TYPE_COAP_CBOR:
+	{
+		if (!obj->coap_cbor) {
+			return -ENOENT;
+		}
+
+		if (obj->coap_cbor->str_val) {
```
jayteemo, line 493: instead, check the union type to see if it has already been set:
e.g.: `if (obj->coap_cbor->type != NRF_CLOUD_DATA_TYPE_NONE)`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -366,6 +441,22 @@ int nrf_cloud_obj_num_add(struct nrf_cloud_obj *const obj, const char *const key
 		return cJSON_AddNumberToObjectCS(dest_json_get(obj, data_child),
 						 key, val) ? 0 : -ENOMEM;
 	}
+	case NRF_CLOUD_OBJ_TYPE_COAP_CBOR:
+	{
+		if (!obj->coap_cbor) {
+			return -ENOENT;
+		}
+
```
jayteemo, line 449: add union type check:
```
if (obj->coap_cbor->type != NRF_CLOUD_DATA_TYPE_NONE) {
        return -ENOTEMPTY;
}
```

samples/cellular/nrf_cloud_mqtt_multi_service/src/application.c
plskeggs: FOTA is not connection management.

The shadow is multiple things, including connection management, but also declaration of supported sensors, which are managed here.

My only concern with using threads everywhere is that each one needs a stack, and that consumes RAM.  But yes, I know that was your design.

include/net/nrf_cloud_codec.h
```diff
@@ -34,6 +34,7 @@ enum nrf_cloud_obj_type {
 	NRF_CLOUD_OBJ_TYPE__UNDEFINED,
 
 	NRF_CLOUD_OBJ_TYPE_JSON,
+	NRF_CLOUD_OBJ_TYPE_COAP_CBOR,
```
jayteemo, line 37: it might be useful to explicitly state that this object type is to be used for only one type (nrf_cloud_data_type) of nrf cloud coap message... bulk and custom/nested messages are not supported.
not sure if the comment should be added here, on `struct nrf_cloud_obj_coap_cbor` or on `NRF_CLOUD_OBJ_COAP_CBOR_DEFINE`

samples/cellular/nrf_cloud_mqtt_multi_service/src/main.c
```diff
@@ -39,6 +39,15 @@ K_THREAD_DEFINE(con_thread, CONFIG_CONNECTION_THREAD_STACK_SIZE, cloud_connectio
...
+	const char *protocol;
+
+	if (IS_ENABLED(CONFIG_NRF_CLOUD_MQTT)) {
+		protocol = "MQTT";
+	} else if (IS_ENABLED(CONFIG_NRF_CLOUD_COAP)) {
+		protocol = "CoAP";
+	}
+
+	LOG_INF("nRF Cloud %s Multi Service sample has started, version: %s",
+		protocol, CONFIG_APP_VERSION);
```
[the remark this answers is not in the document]
plskeggs: I thought it might be useful to reflect the connection type in the name, but it could be done some other way.

samples/cellular/nrf_cloud_mqtt_multi_service/src/handle_fota.c
```diff
@@ -0,0 +1,533 @@
+/*
```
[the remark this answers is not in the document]
plskeggs: It's temporary.  Yes I can rename it.

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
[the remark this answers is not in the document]
plskeggs: I do not think this is the time for this optimization.  I'll create a ticket for it.

samples/cellular/nrf_cloud_mqtt_multi_service/src/main.c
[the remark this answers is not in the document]
glarsennordic: Maybe a "currently using MQTT/CoAP" message immediately after `nRF Cloud Multi Service Sample has started`

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
```diff
@@ -115,8 +121,13 @@ void disconnect_cloud(void)
...
 	 */
 	LOG_INF("Disconnecting from nRF Cloud");
-	int err = nrf_cloud_disconnect();
+	int err;
 
+#if defined(CONFIG_NRF_CLOUD_MQTT)
+	err = nrf_cloud_disconnect();
+#elif defined(CONFIG_NRF_CLOUD_COAP)
+	err = nrf_cloud_coap_disconnect();
+#endif
```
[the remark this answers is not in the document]
plskeggs: I'll make a ticket.

include/net/nrf_cloud_codec.h
glarsennordic: In addition to explicitly stating it with a comment as you suggest, we could also name the type itself something like `NRF_CLOUD_OBJ_TYPE_CBOR_DATA_MSG`. At minimum, it doesn't seem necessary to include `COAP` in the title, at least

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
```diff
@@ -182,40 +205,64 @@ static void cloud_ready(void)
...
+		/* Include the application version */
+		.application_version = CONFIG_APP_VERSION
+	};
+	struct nrf_cloud_modem_info *mdm = &modem_info;
+#else
+	struct nrf_cloud_modem_info *mdm = NULL;
+#endif
 	struct nrf_cloud_device_status device_status = {
-		.modem = NULL,
+		.modem = mdm,
```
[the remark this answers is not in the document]
plskeggs: I'm not sure what you mean by "nrf_cloud can query this information itself."  From where?

I didn't design the device_info struct.  We can discuss redesigning it in the future.

This isn't "relying" on CONFIG_NRF_MODEM.  It will build with it disabled.

include/net/nrf_cloud_codec.h
jayteemo: this type is only for nRF Cloud **_CoAP_** CBOR messages...

include/net/nrf_cloud_codec.h
```diff
@@ -48,12 +49,42 @@ enum nrf_cloud_enc_src {
...
+	/** The int int_val field contains the data. */
+	NRF_CLOUD_DATA_TYPE_INT
+};
+
+/** @brief Object to support nRF Cloud CoAP CBOR messages */
+struct nrf_cloud_obj_coap_cbor {
+	char *app_id;
+	enum nrf_cloud_data_type type;
+	union {
+		struct nrf_cloud_data data;
```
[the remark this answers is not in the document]
plskeggs: no, I thought it was at first but it isn't.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
plskeggs: It's redundant since calloc() sets it to 0, and NRF_CLOUD_DATA_TYPE_NONE = 0, but if you want it for clarity I can.

jayteemo: yes, it would be good to set it explicitly.
i don't know what the value of `NRF_CLOUD_DATA_TYPE_NONE` since it doesn't exist yet... and maybe its value will be `-1` or `0xFFFFFFFF`.

glarsennordic (commented):
Took a cursory glance, I think this is viable

include/net/nrf_cloud_codec.h
jayteemo: i was explaining why i named it with the added COAP text and feel that it should stay.
we could potentially add a generic CBOR type later (which was kind of the original intent).

i don't think the JSON type is MQTT specific.

samples/cellular/nrf_cloud_mqtt_multi_service/src/application.c
plskeggs: it works

samples/cellular/nrf_cloud_mqtt_multi_service/src/application.c
plskeggs: done for both

samples/cellular/nrf_cloud_mqtt_multi_service/src/application.c
```diff
@@ -269,6 +190,44 @@ static void handle_at_cmd_requests(const struct nrf_cloud_data *const dev_msg)
...
 	(void)nrf_cloud_obj_free(&msg_obj);
 }
+#else
+static void handle_at_cmd_requests(const struct nrf_cloud_data *const dev_msg)
+{
+}
+#endif /* CONFIG_AT_CMD_REQUESTS */
+
+#if defined(CONFIG_NRF_CLOUD_COAP)
+static void check_shadow(void)
```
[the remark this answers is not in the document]
plskeggs: did something like this, different names though

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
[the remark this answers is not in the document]
plskeggs: However, it is a better user experience in the nRF Cloud portal if these areas are updated to reflect the name of the sample and how it is connected, rather than leave stale information there (potentially).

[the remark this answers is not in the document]
jayteemo: we have [CONFIG_NRF_CLOUD_SEND_DEVICE_STATUS](https://github.com/nrfconnect/sdk-nrf/blob/b712ba60fb59d184523f3da0a8265ceea758cf6b/subsys/net/lib/nrf_cloud/Kconfig#L46) and its related config options.  so this information is already sent on connect for M(QTT)MSS.
can they be implemented for CoAP too, so the application doesn't have to do this?
if not, they can be moved into the mqtt specific kconfig.

samples/cellular/nrf_cloud_mqtt_multi_service/src/fota_support_coap.c
```diff
@@ -0,0 +1,527 @@
...
+
+	LOG_INF("Checking for pending FOTA job");
+	ret = nrf_cloud_pending_fota_job_process(&pending_job, &reboot_required);
+
+	if ((ret == 0) && reboot_required) {
+		/* Save validate status and reboot */
+		(void)save_pending_job();
+
+		/* Sleep to display log message */
+		LOG_INF("Rebooting in %us...", PENDING_REBOOT_S);
```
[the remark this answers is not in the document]
jayteemo: i think you could just have one "fota_reboot" function for MQTT and CoAP:
e.g.:
```
void fota_reboot(const unsigned int delay_s, const bool error) 
{
	LOG_INF("Rebooting in %us%s", delay_s, error ? " due to error" : "...");

	if (error)
	{
		(void)lte_lc_deinit();
	}

	LOG_PANIC();
	k_sleep(K_SECONDS(delay_s));
	sys_reboot(SYS_REBOOT_COLD);
}
```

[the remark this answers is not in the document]
jayteemo: looks like this could be done in the rest fota sample too 😅

samples/cellular/nrf_cloud_mqtt_multi_service/src/fota_support_coap.c
```diff
@@ -0,0 +1,527 @@
...
+
+int coap_fota_process(void)
+{
+	int err;
+
+	/* If a FOTA job is in progress, handle it first */
+	if (validate_in_progress_job()) {
+		err = update_job_status();
+		if (err) {
+			LOG_ERR("Error; must reboot: %d", err);
```
jayteemo: `LOG_ERR("Error updating FOTA job status: %d", err);`

the reboot is mentioned in the `error_reboot` log msg.

samples/cellular/nrf_cloud_mqtt_multi_service/src/fota_support_coap.c
```diff
@@ -0,0 +1,527 @@
...
+{
+	/* Job operations can take up to 30s to be processed */
+	LOG_INF("Waiting %u seconds for job update to be processed by nRF Cloud...", JOB_WAIT_S);
+	cleanup();
+	k_sleep(K_SECONDS(JOB_WAIT_S));
+}
+
+static void error_reboot(void)
+{
+	LOG_INF("Rebooting due to error in %us...", ERROR_REBOOT_S);
```
jayteemo: `"Rebooting in %us due to error"`

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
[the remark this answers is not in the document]
plskeggs: OK, I can rework the CoAP library to do this, but I'd like to do it in a follow-on PR.

samples/cellular/nrf_cloud_mqtt_multi_service/src/main.c
```diff
@@ -33,12 +35,34 @@ K_THREAD_DEFINE(msg_thread, CONFIG_MESSAGE_THREAD_STACK_SIZE, message_queue_thre
...
+#endif
+
 /* main() is called from the main thread, which defaults to priority zero,
  * but for illustrative purposes we don't use it. main_application() could be called directly
  * from this function, rather than given its own dedicated thread.
  */
 int main(void)
 {
-	LOG_INF("nRF Cloud MQTT multi-service sample has started, version: %s", CONFIG_APP_VERSION);
+	LOG_INF("nRF Cloud Multi Service sample has started, version: %s",
```
jayteemo, line 54: "multi-service"
i think it should be hyphenated for consistency

samples/cellular/nrf_cloud_mqtt_multi_service/src/main.c
```diff
@@ -33,12 +35,34 @@ K_THREAD_DEFINE(msg_thread, CONFIG_MESSAGE_THREAD_STACK_SIZE, message_queue_thre
...
+		CONFIG_APP_VERSION);
```
jayteemo, line 55: i don't remember what the objections were to making this one log msg...
so ignore if no one likes this:

```
LOG_INF("nRF Cloud multi-service sample has started, version: %s, protocol: %s",
		CONFIG_APP_VERSION,
		IS_ENABLED(CONFIG_NRF_CLOUD_COAP) ? "CoAP" : "MQTT");
```

samples/cellular/nrf_cloud_mqtt_multi_service/src/main.c
```diff
@@ -33,12 +35,34 @@ K_THREAD_DEFINE(msg_thread, CONFIG_MESSAGE_THREAD_STACK_SIZE, message_queue_thre
...
+
+	const char *protocol;
+
+	if (IS_ENABLED(CONFIG_NRF_CLOUD_MQTT)) {
+		protocol = "MQTT";
+	} else if (IS_ENABLED(CONFIG_NRF_CLOUD_COAP)) {
+		protocol = "CoAP";
+	}
+
+	LOG_INF("Connected using %s", protocol);
```
jayteemo, line 65: the actual connection has not occurred yet, right?

samples/cellular/nrf_cloud_mqtt_multi_service/src/shadow_support_coap.c
```diff
@@ -0,0 +1,75 @@
...
+
+#define COAP_SHADOW_MAX_SIZE 512
+
+static int check_shadow(void)
+{
+	int err;
+	char buf[COAP_SHADOW_MAX_SIZE];
+
+	buf[0] = '\0';
+	LOG_INF("Checking for shadow delta...");
```
jayteemo: i think all the LOG_INFs should be LOG_DBGs.

samples/cellular/nrf_cloud_mqtt_multi_service/src/shadow_support_coap.c
```diff
@@ -0,0 +1,75 @@
...
+	err = nrf_cloud_coap_shadow_get(buf, sizeof(buf), true);
+	if (err == -EACCES) {
+		LOG_INF("Not connected yet.");
+	} else if (err) {
+		LOG_ERR("Failed to request shadow delta: %d", err);
+	} else {
+		size_t len = strlen(buf);
+
+		LOG_INF("Delta: len:%zd, %s", len, len ? buf : "None");
+		/* Do something with the shadow delta's JSON data, such
```
jayteemo: it doesn't look like MSS cares about the shadow deltas, so is this necessary?
if it is necessary, can a comment be added as to why?

in general, what happens if there are a bunch of shadow deltas and the device never acks them?

include/net/nrf_cloud_coap.h
```diff
@@ -245,6 +275,17 @@ int nrf_cloud_coap_shadow_device_status_update(const struct nrf_cloud_device_sta
  */
 int nrf_cloud_coap_shadow_service_info_update(const struct nrf_cloud_svc_info * const svc_inf);
 
+/**
+ * @brief Send an nRF Cloud Object
```
jayteemo, line 279: see comment in the c file:
update brief to say what the object requirements are.

include/net/nrf_cloud_codec.h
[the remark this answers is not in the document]
jayteemo: it can be removed then?

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
```diff
@@ -182,40 +206,60 @@ static void cloud_ready(void)
...
 		.svc = &service_info
 	};
 
+	if (updated) {
+		return; /* It is not necessary to do this more than once per reboot. */
```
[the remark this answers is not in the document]
plskeggs: Is that a request?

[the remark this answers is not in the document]
plskeggs: OK.

samples/cellular/nrf_cloud_mqtt_multi_service/src/fota_support_coap.c
```diff
@@ -0,0 +1,527 @@
...
+	struct dfu_target_fmfu_fdev fmfu_dev_inf = {
+		.size = 0,
+		.offset = 0,
+		/* CONFIG_DFU_TARGET_FULL_MODEM_USE_EXT_PARTITION is enabled, so no need
+		 * to specify the flash device here
+		 */
+		.dev = NULL
+	};
+
+	err = nrf_cloud_fota_fmfu_dev_set(&fmfu_dev_inf);
```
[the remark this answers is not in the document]
plskeggs: And then remove the struct dfu_target_fmfu_fdev I presume as well.

samples/cellular/nrf_cloud_mqtt_multi_service/src/fota_support_coap.c
```diff
@@ -0,0 +1,527 @@
...
+
+		/* Sleep to display log message */
+		LOG_INF("Rebooting in %us...", PENDING_REBOOT_S);
+		LOG_PANIC();
+		k_sleep(K_SECONDS(PENDING_REBOOT_S));
+		sys_reboot(SYS_REBOOT_COLD);
+	}
+}
+
+int coap_fota_init(void)
```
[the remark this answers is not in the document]
plskeggs: Yes, I'm planning to create a ticket for it.

samples/cellular/nrf_cloud_mqtt_multi_service/src/fota_support_coap.c
plskeggs: function is gone now

samples/cellular/nrf_cloud_mqtt_multi_service/src/main.c
plskeggs: Georges suggested a second line.  Doesn't matter to me.

samples/cellular/nrf_cloud_mqtt_multi_service/src/shadow_support_coap.c
plskeggs: We have requests to provide an example of how a customer could use the shadow delta to send a "command" to a CoAP device when the device sleeps a lot.

It would go here.  The note "Do something..." is a prompt to the reader.

I plan to improve this (future) by showing how they can parse out something that they send using curl, perhaps.  Unfortunately the frontend only allows for ending of the config section, which I'm not sure we want to encourage the customer to modify, since that is for ATv2.  Perhaps we need to designate a new section within the shadow that the front end will also allow edits for.

samples/cellular/nrf_cloud_mqtt_multi_service/src/shadow_support_coap.c
plskeggs: I want to keep the inf for the delta itself, if one is received.

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c
```diff
@@ -160,14 +160,49 @@ int nrf_cloud_coap_pgps_url_get(struct nrf_cloud_rest_pgps_request const *const
...
+int nrf_cloud_coap_obj_send(struct nrf_cloud_obj *const obj)
+{
+	if (!nrf_cloud_coap_is_connected()) {
+		return -EACCES;
+	}
+
+	if (!obj) {
+		return -EINVAL;
+	}
+
```
jayteemo, line 172: i think we should add the following check, at least for the type... up to you if you want to allow a pre-encoded buffer to be sent (i would say no, to reduce potential errors/misuse):
```
/* Only support sending of the CoAP CBOR type or a pre-encoded buffer. */
if ((obj->type != NRF_CLOUD_OBJ_TYPE_COAP_CBOR) &&
    (obj->enc_src != NRF_CLOUD_ENC_SRC_PRE_ENCODED) {
		return -ENOTSUP;
}
```

samples/cellular/nrf_cloud_mqtt_multi_service/src/shadow_support_coap.c
jayteemo: the config section can and should be used by any application.
no need to create a separate section.

include/net/nrf_cloud_coap.h
```diff
@@ -135,6 +136,35 @@ int nrf_cloud_coap_pgps_url_get(struct nrf_cloud_rest_pgps_request const *const
...
+ * @brief Send a preencoded JSON message to nRF Cloud.
+ *
+ *  The CoAP message is sent as a non-confirmable CoAP message.
+ *
+ * @param[in]     message    The string to send.
+ *
+ * @retval 0 If successful.
+ *          Otherwise, a (negative) error code is returned.
+ */
+int nrf_cloud_coap_json_message_send(const char *message);
```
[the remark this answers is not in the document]
plskeggs: It should get stored in the backend message store and be retrievable, same as via MQTT.

samples/cellular/nrf_cloud_multi_service/src/cloud_connection.c
```diff
@@ -366,7 +411,7 @@ static void cloud_event_handler(const struct nrf_cloud_evt *nrf_cloud_evt)
 										"Invalid");
 
 		/* Notify fota_support of the completed download. */
-		on_fota_downloaded();
+		fota_reboot(0, false);
```
jayteemo, line 414: maybe do `fota_reboot(5, false);` just so the reboot msg isn't "0s" 😅

samples/cellular/nrf_cloud_multi_service/src/shadow_support_coap.c
```diff
@@ -0,0 +1,75 @@
...
+	LOG_DBG("Checking for shadow delta...");
+	err = nrf_cloud_coap_shadow_get(buf, sizeof(buf), true);
+	if (err == -EACCES) {
+		LOG_DBG("Not connected yet.");
+	} else if (err) {
+		LOG_ERR("Failed to request shadow delta: %d", err);
+	} else {
+		size_t len = strlen(buf);
+
+		LOG_INF("Delta: len:%zd, %s", len, len ? buf : "None");
```
jayteemo, line 37: `I want to keep the inf for the delta itself, if one is received.`
should this log message be moved below into the `if (len) {` block (and the "None" removed)?

not a big deal in its current location... it will print every 5 minutes and let users know the thread is still running.

jayteemo (approved):

plskeggs:
The docs are partially complete.  @wiba-nordic @peknis I will be out until Tuesday 8/8/2023 so if you do a review, just know it is preliminary.

The CI tests pass except for the cloud tests, because they do not yet expect the folder rename and signon message change.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -305,7 +305,7 @@ Cellular samples (renamed from nRF9160 samples)
 
 * Added the :ref:`battery` sample to show how to use the :ref:`modem_battery_readme` library.
 
-* :ref:`nrf_cloud_mqtt_multi_service` sample:
+* :ref:`nrf_cloud_multi_service` sample:
```
wiba-nordic, line 308: Changelog should mention name change and added functionality. I recommend adding ``  * Renamed Cellular: nRF Cloud MQTT multi-service to Cellular: nRF Cloud multi-service.`` as the very first entry in the list for the name change.

wiba-nordic (changes requested):

samples/cellular/nrf_cloud_mqtt_multi_service/prj.conf
wiba-nordic: Was this added to the docs? Didn't notice, at least.

samples/cellular/nrf_cloud_mqtt_multi_service/src/main.c
glarsennordic: Being a single or separate lines is not particularly important to me, as long as the sample is named multi-service sample

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -24,14 +24,17 @@ The sample supports the following development kits:
 
 .. include:: /includes/external_flash_nrf91.txt
 
-.. _nrf_cloud_mqtt_multi_service_features:
+When using CoAP, modem firmware version 1.3.5 or above is recommended, in order to benefit from the power savings provided by DTLS Connection ID.
```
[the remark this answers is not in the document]
peknis: Or even "... use modem firmware version..."
As we tend to avoid using "we recommend".

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -561,16 +570,57 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+     python3 device_credentials_installer.py --ca CA0x3bc7f3b014a8ad492999c594f08bbc2fcffc5fd1_ca.pem \
+     --ca_key CA0x3bc7f3b014a8ad492999c594f08bbc2fcffc5fd1_prv.pem --id_str nrf- --id_imei -s -d --coap
+     ...
+     <- OK
+     Saving provisioning endpoint CSV file provision.csv...
+     Provisioning CSV file saved
+
+Once the script creates the provision.csv file, visit http://nrfcloud.com/#/provision-devices to provision the device in nRF Cloud.
+This registers the device with nRF Cloud, so that the device can use any nRF Cloud services via CoAP, MQTT, or REST connectivity.
```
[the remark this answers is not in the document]
peknis: via -> through
as we try to avoid Latinisms.

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -561,16 +570,57 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+.. _nrf_cloud_multi_service_building_coap:
+
+Building with CoAP Support
+==========================
+
+To build the sample to use CoAP instead of MQTT, use the ``-DOVERLAY_CONFIG=overlay_coap.conf`` options.
```
peknis, line 614:
```suggestion
To build the sample to use CoAP instead of MQTT, use the ``-DOVERLAY_CONFIG=overlay_coap.conf`` option.
```

peknis (changes requested):

samples/cellular/nrf_cloud_mqtt_multi_service/prj.conf
plskeggs: Done.

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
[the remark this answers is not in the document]
plskeggs: https://nordicsemi.atlassian.net/browse/IRIS-6791

samples/cellular/nrf_cloud_mqtt_multi_service/src/cloud_connection.c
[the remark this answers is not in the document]
plskeggs: https://nordicsemi.atlassian.net/browse/IRIS-6792

samples/cellular/nrf_cloud_mqtt_multi_service/src/message_queue.c
```diff
@@ -122,6 +126,11 @@ static int consume_device_message(void)
 	/* Send message */
 	ret = nrf_cloud_send(&mqtt_msg);
 
+#elif defined(CONFIG_NRF_CLOUD_COAP)
+	ret = nrf_cloud_coap_obj_send(queued_msg);
+
+#endif /* CONFIG_NRF_CLOUD_COAP */
```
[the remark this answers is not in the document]
plskeggs: https://nordicsemi.atlassian.net/browse/IRIS-6791

samples/cellular/nrf_cloud_mqtt_multi_service/src/fota_support_coap.c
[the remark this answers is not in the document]
plskeggs: https://nordicsemi.atlassian.net/browse/IRIS-6793

doc/nrf/links.txt
```diff
@@ -772,6 +772,11 @@
 .. _`CoAP response codes`: https://datatracker.ietf.org/doc/html/rfc7252#section-12.1.2
 
 .. _`RFC 7252 - The Constrained Application Protocol`: https://datatracker.ietf.org/doc/html/rfc7252
+.. _`RFC 7959 - Block-Wise Transfer in CoAP`: https://datatracker.ietf.org/doc/html/rfc7959
+.. _`RFC 7049 - Concise Binary Object Representation`: https://datatracker.ietf.org/doc/html/rfc7049
+.. _`RFC 8610 - Concise Data Definition Language (CDDL)`: https://datatracker.ietf.org/doc/html/rfc8610
+.. _`RFC 8132 - PATCH and FETCH Methods for CoAP`: https://datatracker.ietf.org/doc/html/rfc8132
+.. _`RFC 9146 - Connection Identifier for DTLS 1.2`: https://datatracker.ietf.org/doc/html/rfc9146
```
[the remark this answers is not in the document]
plskeggs: I plan to add them to the nRF Cloud CoAP library docs.

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +568,113 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+   See :ref:`configuration_device_id` to use other device ID formats.
+
+.. code-block:: console
+
+   python3 device_credentials_installer.py --ca CA0x3bc7f3b014a8ad492999c594f08bbc2fcffc5fd1_ca.pem \
+   --ca_key CA0x3bc7f3b014a8ad492999c594f08bbc2fcffc5fd1_prv.pem --id_str nrf- --id_imei -s -d --coap
+   ...
+   <- OK
+   Saving provisioning endpoint CSV file provision.csv...
+   Provisioning CSV file saved
```
wiba-nordic, line 658:
```suggestion
.. code-block:: console

   python3 device_credentials_installer.py --ca <certificate_serial>_ca.pem \
   --ca_key <certificate_serial>_prv.pem --id_str nrf- --id_imei -s -d --coap
   ...
   <- OK
   Saving provisioning endpoint CSV file provision.csv...
   Provisioning CSV file saved

Where ``<certificate_serial>`` is your certificate's serial number in hex.
```
The console output for the command isn't really necessary since we're only interested in the command.

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +568,113 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
 
+When using CoAP, the following additional configuration options are available:
+
+.. _CONFIG_COAP_SHADOW_CHECK_RATE_SECONDS:
+
+CONFIG_COAP_SHADOW_CHECK_RATE_SECONDS - How many seconds between requests for any change (delta) in the device shadow.
+
+.. _CONFIG_COAP_SHADOW_THREAD_STACK_SIZE:
+
+CONFIG_COAP_SHADOW_THREAD_STACK_SIZE - Size in bytes of the shadow check thread stack.
```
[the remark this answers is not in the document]
plskeggs: Thanks.

samples/cellular/nrf_cloud_multi_service/Kconfig
```diff
@@ -220,10 +221,96 @@ config AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH
...
+	default 60
+	help
+	  Sets the time to wait between each request to get any pending FOTA job.
+
+config COAP_FOTA_THREAD_STACK_SIZE
+	int "CoAP FOTA Thread Stack Size (bytes)"
+	default 2048
+	help
+	  Sets the stack size (in bytes) for the CoAP FOTA thread of the
+	  sample.
```
glarsennordic, line 285: These are all here temporarily until we integrate this into nrf_cloud itself, correct?

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
```diff
@@ -0,0 +1,73 @@
...
+CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD=2
+CONFIG_NRF_CLOUD_PGPS_NUM_PREDICTIONS=4
+
+# Location Services configuration
+CONFIG_LOCATION_DATA_DETAILS=y
+
+# Date Time lib - Used by PGPS and main application
+CONFIG_DATE_TIME_AUTO_UPDATE=y
+CONFIG_DATE_TIME_MODEM=y
+CONFIG_DATE_TIME_NTP=y
```
glarsennordic, line 69: Do these need to be explicitly enabled as compared to mainline `prj.conf`?

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
```diff
@@ -0,0 +1,73 @@
...
+CONFIG_NRF_CLOUD_LOG_DIRECT=y
+CONFIG_NRF_CLOUD_LOG_OUTPUT_LEVEL=3
+CONFIG_NRF_CLOUD_LOG_LOG_LEVEL_INF=y
+CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD=2
+CONFIG_NRF_CLOUD_PGPS_NUM_PREDICTIONS=4
+
+# Location Services configuration
+CONFIG_LOCATION_DATA_DETAILS=y
+
+# Date Time lib - Used by PGPS and main application
```
glarsennordic, line 66: "Used by PGPS" is a bit confusing since NRF_CLOUD_LOCATION is disabled

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
```diff
@@ -0,0 +1,73 @@
...
+
+# Network
+CONFIG_NET_SOCKETS_POSIX_NAMES=y
+CONFIG_NET_SOCKETS_TLS_SET_MAX_FRAGMENT_LENGTH=y
+CONFIG_NET_SOCKETS_OFFLOAD_PRIORITY=40
+
+# nRF Cloud
+CONFIG_NRF_CLOUD_MQTT=n
+CONFIG_NRF_CLOUD_FOTA=n
+CONFIG_NRF_CLOUD_LOCATION=n
```
glarsennordic, line 50: Did you mean for this to be disabled?

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
```diff
@@ -0,0 +1,73 @@
...
+
+# Location Services configuration
+CONFIG_LOCATION_DATA_DETAILS=y
+
+# Date Time lib - Used by PGPS and main application
+CONFIG_DATE_TIME_AUTO_UPDATE=y
+CONFIG_DATE_TIME_MODEM=y
+CONFIG_DATE_TIME_NTP=y
+
+CONFIG_NRF_CLOUD_COAP_SERVER_HOSTNAME="coap.nrfcloud.com"
```
glarsennordic, line 71: If this isn't the default value, maybe it should be?

samples/cellular/nrf_cloud_multi_service/src/fota_support.c
```diff
@@ -7,18 +7,24 @@
...
 #include <net/nrf_cloud.h>
 
 #include "fota_support.h"
 
-LOG_MODULE_REGISTER(fota_support, CONFIG_MQTT_MULTI_SERVICE_LOG_LEVEL);
+LOG_MODULE_REGISTER(fota_support, CONFIG_MULTI_SERVICE_LOG_LEVEL);
 
 /* Called from nRF Cloud event handler in connection.c */
-void on_fota_downloaded(void)
+void fota_reboot(const unsigned int delay_s, const bool error)
```
glarsennordic, line 18: Can you please create a new `on_fota_downloaded` function that takes no arguments, and then calls `fota_reboot(FOTA_REBOOT_DELAY_S, false);`?

Then make `FOTA_REBOOT_DELAY` a KConfig for the sample, or do something else so that it is defined here (speaking of which, it looks like you've defined it both in `cloud_connection.c` and in `fota_support_coap.c`)

Then the call to `fota_reboot` in `cloud_connection.c` can go back to being `on_fota_downloaded`, which IMO enforces better separation of concerns without actually changing your work too much

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
```diff
@@ -0,0 +1,73 @@
...
+
+# Logs
+CONFIG_MULTI_SERVICE_LOG_LEVEL_DBG=y
+CONFIG_LOG=y
+CONFIG_NET_LOG=y
+CONFIG_LOG_PRINTK=y
+CONFIG_COAP_LOG_LEVEL_INF=y
+CONFIG_LOCATION_LOG_LEVEL_DBG=y
+CONFIG_DATE_TIME_LOG_LEVEL_DBG=y
+CONFIG_NRF_CLOUD_COAP_LOG_LEVEL_DBG=y
```
glarsennordic, line 26: Not sure if you're aware you still have these here -- if you are that's fine, no need to remove them, just making sure

samples/cellular/nrf_cloud_multi_service/prj.conf
```diff
@@ -116,3 +116,4 @@ CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=n
 CONFIG_NRF_CLOUD_ALERT=y
 CONFIG_NRF_CLOUD_LOG_DIRECT=y
 CONFIG_NRF_CLOUD_LOG_OUTPUT_LEVEL=3
+CONFIG_TEST_COUNTER=y
```
glarsennordic, line 119: I have no problem with this, but is this intentional?

samples/cellular/nrf_cloud_multi_service/Kconfig
plskeggs: Probably not the last two (job check rate minutes and thread stack size), but the rest, yes.

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
plskeggs: Probably not.

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
plskeggs: `CONFIG_NRF_CLOUD_LOCATION` only applies to MQTT.

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
plskeggs: Yes, they do not apply to CoAP.

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
plskeggs: It is the default.  I change it to test on other stages, but I can remove it now.

samples/cellular/nrf_cloud_multi_service/src/application.c
```diff
@@ -39,6 +41,14 @@ BUILD_ASSERT(CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH >= AT_CMD_REQUEST_ERR_
 #define TEMP_ALERT_HYSTERESIS 1.5f
 #define TEMP_ALERT_LOWER_LIMIT (TEMP_ALERT_LIMIT - TEMP_ALERT_HYSTERESIS)
 
+#if defined(CONFIG_NRF_CLOUD_MQTT)
+#define MSG_OBJ_DEFINE(_obj_name) \
+	NRF_CLOUD_OBJ_JSON_DEFINE(_obj_name);
+#elif defined(CONFIG_NRF_CLOUD_COAP)
+#define MSG_OBJ_DEFINE(_obj_name) \
+	NRF_CLOUD_OBJ_COAP_CBOR_DEFINE(_obj_name);
+#endif
```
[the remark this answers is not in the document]
plskeggs: I blame Justin ;-)

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
plskeggs: I'll remove them

samples/cellular/nrf_cloud_multi_service/prj.conf
plskeggs: Is what intentional?  Enabling the test counter?  Why is it disabled?

I can remove it, but wondering why it is the way it is.

wiba-nordic (approved):

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
    Modem responses longer than this length will be replaced with an error code message (-NRF_E2BIG).
    Cannot be less than 40 bytes.
 
+When using CoAP, the following additional configuration options are available:
+
+.. _CONFIG_COAP_SHADOW_CHECK_RATE_SECONDS:
+
+CONFIG_COAP_SHADOW_CHECK_RATE_SECONDS - Rate to check for shadow changes
+   How many seconds between requests for any change (delta) in the device shadow.
```
wiba-nordic, line 608:
```suggestion
CONFIG_COAP_SHADOW_CHECK_RATE_SECONDS - Rate of checking for shadow changes (seconds)
   Sets the time (in seconds) between requests for any change (delta) in the device shadow.
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+.. _CONFIG_COAP_SHADOW_THREAD_STACK_SIZE:
+
+CONFIG_COAP_SHADOW_THREAD_STACK_SIZE - CoAP Shadow Thread Stack Size (bytes)
+   Sets the stack size (in bytes) for the shadow delta checking thread of the sample.
```
wiba-nordic, line 613:
```suggestion
   Sets the stack size (in bytes) for the thread that checks the shadow delta.
```

samples/cellular/nrf_cloud_multi_service/src/application.c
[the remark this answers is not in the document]
jayteemo: Not sure why this would go in nrf_cloud...
But put it wherever.

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+.. _CONFIG_NRF_CLOUD_COAP_FOTA:
+
+CONFIG_NRF_CLOUD_COAP_FOTA - Enable FOTA with CoAP
+   The sample periodically checks for pending FOTA jobs.
+   The sample performs the FOTA update when received.
```
wiba-nordic, line 619:
```suggestion
CONFIG_NRF_CLOUD_COAP_FOTA - FOTA with CoAP
   Enables periodical checks for pending FOTA jobs.
   The sample performs any received FOTA updates.
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+If :kconfig:option:`CONFIG_NRF_CLOUD_COAP_FOTA` is enabled, these options additional are available:
```
wiba-nordic, line 621:
```suggestion
If :kconfig:option:`CONFIG_NRF_CLOUD_COAP_FOTA` is enabled, these additional options are available:
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+.. _CONFIG_COAP_FOTA_DL_TIMEOUT_MIN:
+
+CONFIG_COAP_FOTA_DL_TIMEOUT_MIN - CoAP FOTA download timeout
+    The time in minutes allotted for a FOTA download to complete.
```
wiba-nordic, line 626:
```suggestion
CONFIG_COAP_FOTA_DL_TIMEOUT_MIN - CoAP FOTA download timeout (minutes)
    Sets the time (in minutes) allotted for a FOTA download to complete.
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+.. _CONFIG_COAP_FOTA_USE_NRF_CLOUD_SETTINGS_AREA:
+
+CONFIG_COAP_FOTA_USE_NRF_CLOUD_SETTINGS_AREA - Make FOTA compatible with other samples
+   Use the same settings area as the nRF Cloud FOTA library.
+
+.. _CONFIG_COAP_FOTA_SETTINGS_NAME:
+
+CONFIG_COAP_FOTA_SETTINGS_NAME - Settings identifier
+   Set the identifier for the CoAP FOTA storage if :kconfig:option:`CONFIG_COAP_FOTA_USE_NRF_CLOUD_SETTINGS_AREA` is not enabled.
```
wiba-nordic, line 636:
```suggestion
   Sets the identifier for the CoAP FOTA storage if :kconfig:option:`CONFIG_COAP_FOTA_USE_NRF_CLOUD_SETTINGS_AREA` is not enabled.
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+.. _CONFIG_COAP_FOTA_SETTINGS_KEY_PENDING_JOB:
+
+CONFIG_COAP_FOTA_SETTINGS_KEY_PENDING_JOB - Settings item key
+   Set the settings item key for pending FOTA job info if :kconfig:option:`CONFIG_COAP_FOTA_USE_NRF_CLOUD_SETTINGS_AREA` is not enabled.
```
wiba-nordic, line 641:
```suggestion
   Sets the settings item key for pending FOTA job info if :kconfig:option:`CONFIG_COAP_FOTA_USE_NRF_CLOUD_SETTINGS_AREA` is not enabled.
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+.. _CONFIG_COAP_FOTA_JOB_CHECK_RATE_MINUTES:
+
+CONFIG_COAP_FOTA_JOB_CHECK_RATE_MINUTES - FOTA job check interval (minutes)
+   How many minutes between requests for a FOTA job.
```
wiba-nordic, line 646:
```suggestion
   Sets the time (in minutes) between requests for a FOTA job.
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+.. _CONFIG_COAP_FOTA_THREAD_STACK_SIZE:
+
+CONFIG_COAP_FOTA_THREAD_STACK_SIZE - CoAP FOTA Thread Stack Size (bytes)
+   Sets the stack size (in bytes) for the FOTA job checking thread of the sample.
```
wiba-nordic, line 651:
```suggestion
   Sets the stack size (in bytes) for the thread that checks for a FOTA job.
```

wiba-nordic (changes requested):

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
 
 .. include:: /includes/build_and_run_ns.txt
 
-.. _nrf_cloud_mqtt_multi_service_building_wifi:
+.. _nrf_cloud_multi_service_coap_provisioning:
+
+Provisioning a CoAP device on nRF Cloud
+***************************************
+
+For more information, see `nRF Cloud Utilities documentation`_.
```
peknis, line 671: This might have a better place at the end of this section.

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+* :file:`<certificate_serial>_ca.pem`
+* :file:`<certificate_serial>_prv.pem`
+* :file:`<certificate_serial>_pub.pem`
+
+Where ``<certificate_serial>`` is your certificate's serial number in hex.
+You will use these files to sign device certificates.
+
+You only need to generate these three files once.
+They can be used to sign as many device certificates as you need.
```
peknis, line 691:
```suggestion
You can use them to sign as many device certificates as you need.
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -557,20 +600,131 @@ CONFIG_AT_CMD_REQUEST_RESPONSE_BUFFER_LENGTH - Length of AT command request resp
...
+
+Next, for each device you wish to provision, execute the following command in the same folder as the above command:
+
+.. note::
+   The options used below assume that the sample is built with the default :kconfig:option:`CONFIG_NRF_CLOUD_CLIENT_ID_SRC_IMEI` option enabled and :kconfig:option:`CONFIG_NRF_CLOUD_CLIENT_ID_PREFIX` option set to ``nrf-``.
```
peknis, line 696:
```suggestion
   The options used in the example command assume that the sample is built with the default :kconfig:option:`CONFIG_NRF_CLOUD_CLIENT_ID_SRC_IMEI` option enabled and :kconfig:option:`CONFIG_NRF_CLOUD_CLIENT_ID_PREFIX` option set to ``nrf-``.
```

samples/cellular/nrf_cloud_multi_service/README.rst
```diff
@@ -605,7 +759,17 @@ However, because JSON logs are large, you may want to edit the overlay file to c
 Deselect the :kconfig:option:`CONFIG_LOG_BACKEND_NRF_CLOUD_OUTPUT_TEXT` Kconfig option and select :kconfig:option:`CONFIG_LOG_BACKEND_NRF_CLOUD_OUTPUT_DICTIONARY` instead.
 See `Dictionary-based Logging`_ to learn how dictionary-based logging works, how the dictionary is built, and how to decode the binary log output.
 
-.. _nrf_cloud_mqtt_multi_service_dependencies:
+.. _nrf_cloud_multi_service_dependencies:
+
+References
+**********
+
+`RFC 7252 - The Constrained Application Protocol`_
```
peknis, line 767: Use bullet points.

samples/cellular/nrf_cloud_multi_service/src/fota_support.h
```diff
@@ -8,6 +8,20 @@
...
 #include <zephyr/kernel.h>
 
+/* Time in seconds to wait before rebooting after a FOTA update. */
+#define FOTA_REBOOT_DELAY_S	5
+#define PENDING_REBOOT_S	10
+#define ERROR_REBOOT_S		30
+#define FOTA_REBOOT_S		10
+
+/**
+ * @brief De-init LTE if error is true.
```
peknis, line 18: Not so sure about de-init. Should it rather be deinitialize?

peknis (changes requested):

samples/cellular/nrf_cloud_multi_service/README.rst
plskeggs: I specifically chose this location to emphasize that this provisioning must occur.  I do not want users to follow the "Building With CoAP Support" directions without properly provisioning the device.

wiba-nordic:
Looks good, but I'll check the final doc build before approval.

samples/cellular/nrf_cloud_multi_service/README.rst
peknis: Okay, but you might emphasize this a bit more, maybe saying something like: Read this first.

peknis (approved):

wiba-nordic (approved):

tony-le-24 (approved):

samples/cellular/nrf_cloud_multi_service/prj.conf
glarsennordic: When I originally added it, I conceived of it as an assistance feature for testing that the sample itself works, rather than as a demo feature, but I guess you could also view it as a demo feature, in which case changing the KConfig itself to be enabled by default might make sense

samples/cellular/nrf_cloud_multi_service/src/application.c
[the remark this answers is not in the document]
glarsennordic: > Not sure why this would go in nrf_cloud...

It's a generic object declarator that intelligently chooses between either the COAP or JSON object type, and to me that choice feels like the domain of nrf_cloud rather than of the application

I don't see a reason to **_force_** that choice to be made by nrf_cloud always, but it does seem like it could live in `nrf_cloud` as `NRF_CLOUD_OBJ_AUTO_DEFINE` 

This also makes me wonder: Should `nrf_cloud_data_type` instead be `nrf_cloud_obj_fixed_data_type` or something to that effect?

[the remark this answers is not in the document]
jayteemo: this is a demo application which now supports 2 build options.
i don't think many real world applications will be built that way.
having an "auto" macro seems like it would just lead to confusion and/or lack of awareness of what is being done.
but, again, put it wherever you want.  this is not worth a continued discussion on my part.

samples/cellular/nrf_cloud_multi_service/README.rst
plskeggs: I realized after viewing the built docs that the indent level of the Provisioning a CoAP device on nRF Cloud section was the same as the Building and running section.  I have swapped the two and it looks better to me now.

QA discovered a lack of documentation on a particular provisioning topic, so I added a 3 more sentences to the Provisioning a CoAP device on nRF Cloud.  If it needs changes, I can do a follow on PR.

samples/cellular/nrf_cloud_multi_service/src/application.c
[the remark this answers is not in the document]
glarsennordic: Sure, `application.h` is fine if you both feel that way, but maybe let's reconsider it down the line when COAP is more mature

samples/cellular/nrf_cloud_mqtt_multi_service/sample.yaml
plskeggs: https://nordicsemi.atlassian.net/browse/IRIS-6808

glarsennordic (approved):

PR #11904, opened 2023-07-28
subsys: fota_download: correcting URI check

VTPeltoketo:
- HASH checksums are used to validate changed URI

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -595,6 +595,12 @@ Libraries for networking
     * The :kconfig:option:`CONFIG_DOWNLOAD_CLIENT_MAX_HOSTNAME_SIZE` Kconfig option's default value to ``255``.
     * The :kconfig:option:`CONFIG_DOWNLOAD_CLIENT_MAX_FILENAME_SIZE` Kconfig option's default value to ``255``.
 
+* :ref:`lib_fota_download` library:
+
+  * Updated:
+
+    * The library now verifies whether the download started with the same URI and resumes the interrupted download.
+
 Libraries for NFC
 -----------------
 
--- a/subsys/net/lib/fota_download/Kconfig
+++ b/subsys/net/lib/fota_download/Kconfig
@@ -8,6 +8,7 @@ menuconfig FOTA_DOWNLOAD
 	bool "FOTA Download"
 	depends on DOWNLOAD_CLIENT
 	depends on DFU_TARGET
+	select SYS_HASH_FUNC32
 	imply FW_INFO
 
 if (FOTA_DOWNLOAD)
--- a/subsys/net/lib/fota_download/src/fota_download.c
+++ b/subsys/net/lib/fota_download/src/fota_download.c
@@ -6,6 +6,7 @@
 
 #include <zephyr/kernel.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/sys/hash_function.h>
 #include <net/fota_download.h>
 #include <net/download_client.h>
 #include <pm_config.h>
@@ -29,6 +30,8 @@ LOG_MODULE_REGISTER(fota_download, CONFIG_FOTA_DOWNLOAD_LOG_LEVEL);
 static fota_download_callback_t callback;
 static const char *dl_host;
 static const char *dl_file;
+static uint32_t dl_host_hash;
+static uint32_t dl_file_hash;
 static struct download_client dlc;
 static struct k_work_delayable  dlc_with_offset_work;
 static int socket_retries_left;
@@ -415,6 +418,8 @@ int fota_download_start_with_image_type(const char *host, const char *file,
 	const enum dfu_target_image_type expected_type)
 {
 	static int sec_tag_list[1];
+	uint32_t host_hash = 0;
+	uint32_t file_hash = 0;
 	int err = -1;
 
 	struct download_client_cfg config = {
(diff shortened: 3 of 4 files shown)


subsys/net/lib/fota_download/Kconfig
```diff
@@ -8,6 +8,8 @@ menuconfig FOTA_DOWNLOAD
 	bool "FOTA Download"
 	depends on DOWNLOAD_CLIENT
 	depends on DFU_TARGET
+	depends on SYS_HASH_FUNC32
+	depends on SYS_HASH_FUNC32_CHOICE_MURMUR3
```
SeppoTakalo, line 12:
```suggestion
```

I don't think we need to care what is the default hash function. MURMUR3 is btw the default on
```

choice SYS_HASH_FUNC32_CHOICE
	prompt "Default system-wide 32-bit hash function"
	default SYS_HASH_FUNC32_CHOICE_MURMUR3
	help
	  The default system-wide 32-bit hash function is sys_hash32().
```

But after all, this code works with any of the selected hash functions.

subsys/net/lib/fota_download/src/fota_download.c
```diff
@@ -29,6 +30,8 @@ LOG_MODULE_REGISTER(fota_download, CONFIG_FOTA_DOWNLOAD_LOG_LEVEL);
 static fota_download_callback_t callback;
 static const char *dl_host;
 static const char *dl_file;
+static uint32_t dl_host_sha256;
+static uint32_t dl_file_sha256;
```
SeppoTakalo, line 34:
```suggestion
static uint32_t dl_host_hash;
static uint32_t dl_file_hash;
```

You have misleading names here (and below). We use 32bit hash, not SHA256.

SeppoTakalo (changes requested):
Looks functionally correct.

Please remove the extra Kconfig dependency and rename the variables to `*_hash` instead of `*_sha256`

subsys/net/lib/fota_download/src/fota_download.c
```diff
@@ -415,6 +418,8 @@ int fota_download_start_with_image_type(const char *host, const char *file,
 	const enum dfu_target_image_type expected_type)
 {
 	static int sec_tag_list[1];
+	static uint32_t host_sha256;
+	static uint32_t file_sha256;
```
SeppoTakalo, line 422: Why these are `static`?
They don't have to be.

subsys/net/lib/fota_download/Kconfig
VTPeltoketo: Fixed

subsys/net/lib/fota_download/src/fota_download.c
VTPeltoketo: Fixed

subsys/net/lib/fota_download/src/fota_download.c
VTPeltoketo: Fixed

SeppoTakalo (approved):

subsys/net/lib/fota_download/Kconfig
```diff
@@ -8,6 +8,7 @@ menuconfig FOTA_DOWNLOAD
 	bool "FOTA Download"
 	depends on DOWNLOAD_CLIENT
 	depends on DFU_TARGET
+	depends on SYS_HASH_FUNC32
```
SeppoTakalo, line 11: Actually.. It looks like we should use `select` instead of `depends on`.

```
warning: FOTA_DOWNLOAD (defined at /jenkins_workspace/workspace/latest_sdk-
nrf_PR-11904/nrf/subsys/net/lib/fota_download/Kconfig:7) was assigned the value 'y' but got the
value 'n'. Check these unsatisfied dependencies: SYS_HASH_FUNC32 (=n). See
```

All build failures seem to relate to this one. The SYS_HASH_FUNC32 is not enabled in any of the build, so they will fail. Dependencies are not automatically selected.

VTPeltoketo: Ok, I changed it to "select"

tests/subsys/net/lib/fota_download/prj.conf
```diff
@@ -8,3 +8,4 @@ CONFIG_ZTEST_NEW_API=y
 CONFIG_FLASH=y
 CONFIG_FW_INFO=y
 CONFIG_TFM_BUILD_NS=y # In order to mock tfm_platform_s0_active
+CONFIG_SYS_HASH_FUNC32_CHOICE_MURMUR3=y
```
SeppoTakalo, line 11: I think it is enough to add

```suggestion
CONFIG_SYS_HASH_FUNC32=y
```

SeppoTakalo (approved):
Squash both of these commits together. It is basically just one change.

tests/subsys/net/lib/fota_download/prj.conf
VTPeltoketo: Fixed

sigvartmh (approved):
I assume this is either to speed up the check or to remove `strcmp` right? Anyways so far changes looks good to me.

VTPeltoketo:
@sigvartmh There were two reasons: 1) There was a bug in the original code, the pointers of strcmp pointed to the same address. 2) strcmp requires a second buffer of URI and the buffer is quite big. We are running out of RAM and second buffer is not a viable solution.

SeppoTakalo (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -592,6 +592,12 @@ Libraries for networking
     * The :kconfig:option:`CONFIG_DOWNLOAD_CLIENT_MAX_HOSTNAME_SIZE` Kconfig option's default value to ``255``.
     * The :kconfig:option:`CONFIG_DOWNLOAD_CLIENT_MAX_FILENAME_SIZE` Kconfig option's default value to ``255``.
 
+* :ref:`lib_fota_download` library:
+
+  * Updated:
+
+    * Verifies if download started with the same URI and resumes the interrupted download.
```
b-gent, line 599:
```suggestion
    * The library now verifies whether the download started with the same URI and resumes the interrupted download.
```

VTPeltoketo: Fixed

b-gent (approved):

PR #11910, opened 2023-07-31
applications: nrf5340_audio: SD card playback PR

MariusVaardal:


rick1082 (changes requested):
Nice improvement :) I leave couple comments, please check.

alexsven (changes requested):
Not done yet, but here are some comments

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,336 @@
...
+#include "pcm_mix.h"
+#include "audio_i2s.h"
+#include "sw_codec_lc3.h"
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(sd_card_playback, CONFIG_MODULE_SD_CARD_PLAYBACK_LOG_LEVEL);
+
+#define RING_BUF_SIZE_1920_BYTES    1920
+#define SD_CARD_PLAYBACK_STACK_SIZE 4096
+#define MAX_FILENAME_LEN	    32
```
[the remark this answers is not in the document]
MariusVaardal: Just a random number for now. What should it be?

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,336 @@
...
+#include <zephyr/sys/ring_buffer.h>
+#include <zephyr/shell/shell.h>
+#include "sd_card.h"
+#include "hw_codec.h"
+#include "sw_codec_lc3.h"
+#include "sw_codec_select.h"
+#include "pcm_stream_channel_modifier.h"
+#include "pcm_mix.h"
+#include "audio_i2s.h"
+#include "sw_codec_lc3.h"
```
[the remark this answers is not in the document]
MariusVaardal: No

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,336 @@
...
+	uint16_t signal_len_msb; /* Number of samples in signal, 16 MSB (>> 16) */
+};
+
+enum Audio_formats {
+	WAV,
+	LC3
+};
+
+RING_BUF_DECLARE(m_ringbuf_sound_data_lc3, RING_BUF_SIZE_1920_BYTES);
+K_SEM_DEFINE(m_sem_load_from_buf_lc3, 1, 1);
```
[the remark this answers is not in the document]
MariusVaardal: It gives the thread the chance to put valid data into the ringbuffer before the first time data from the ringbuffer is requested by the audio_datapath_stream_out function

applications/nrf5340_audio/src/modules/Kconfig
```diff
@@ -116,4 +120,9 @@ config BUTTON_PUBLISH_STACK_SIZE
 	default 300
 
 endmenu # Stack sizes
+menu "SD card playback"
+config SD_CARD_PLAYBACK
+	bool "If true, lc3 playback is enabled"
+	default n
+endmenu # Lc3 playback
```
rick1082, line 127: We might replace "lc3" to "LC3 encoded file"

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
rick1082: We can check if this is CONFIG_FS_FATFS_MAX_LFN, maybe that is the length which combines filename and path.

applications/nrf5340_audio/src/audio/audio_system.c
```diff
@@ -81,16 +82,20 @@ static void audio_headset_configure(void)
...
 	}
 #endif /* (CONFIG_STREAM_BIDIRECTIONAL) */
-
-	sw_codec_cfg.decoder.num_ch = SW_CODEC_MONO;
+	sw_codec_cfg.decoder.num_ch = SW_CODEC_ONE_CHANNEL;
+	sw_codec_cfg.decoder.num_inst = SW_CODEC_ONE_INST;
+	if (IS_ENABLED(CONFIG_SD_CARD_PLAYBACK)) {
+		sw_codec_cfg.decoder.num_ch += SW_CODEC_ONE_CHANNEL;
+		sw_codec_cfg.decoder.num_inst += SW_CODEC_ONE_INST;
+	}
```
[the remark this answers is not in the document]
rick1082: Agree, we might need to discuss how to handle this properly.

applications/nrf5340_audio/src/modules/Kconfig
MariusVaardal: I changed it to "sd card playback". I just forgot to change the name after I switched the module name from "LC3 playback"

MariusVaardal: Since it can also play .wav files

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+{
+	int ret;
+
+	sd_card_playback_thread_id = k_thread_create(
+		&sd_card_playback_thread_data, sd_card_playback_thread_stack,
+		SD_CARD_PLAYBACK_STACK_SIZE, (k_thread_entry_t)sd_card_playback_thread, NULL, NULL,
+		NULL, K_PRIO_PREEMPT(4), 0, K_NO_WAIT);
+	ret = k_thread_name_set(sd_card_playback_thread_id, "sd_card_playback");
+	if (ret < 0) {
+		LOG_ERR("Failed");
```
koffes, line 294: No need for a print here. `Failed` does not give the developer much additional information.

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+}
+
+/* Shell functions */
+static int cmd_play_wav_file(const struct shell *shell, size_t argc, char **argv)
+{
+	char file_loc[MAX_PATH_LEN + MAX_FILENAME_LEN] = "";
+
+	strcat(file_loc, playback_file_path);
+	strcat(file_loc, argv[1]);
+	sd_card_playback_wav(file_loc, 10, 16, SAMPLE_RATE_48K);
```
koffes, line 307: 10 and 16 are magic numbers. Must be defined or explained.

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+	return 0;
+}
+
+static int cmd_play_lc3_file(const struct shell *shell, size_t argc, char **argv)
+{
+	char file_loc[MAX_PATH_LEN + MAX_FILENAME_LEN] = "";
+
```
koffes, line 314: You may want to add a check that `argc == 2`

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+		return ret;
+	}
+	shell_print(shell, "%s", buf);
+	return 0;
+}
+
+/* Creating subcommands (level 1 command) array for command "demo". */
+SHELL_STATIC_SUBCMD_SET_CREATE(
+	sd_card_playback_cmd,
+	SHELL_COND_CMD(CONFIG_SHELL, play_lc3, NULL, "Play lc3 file.", cmd_play_lc3_file),
```
koffes, line 356: This is not strictly needed in this PR, but how much work would it be to have just a `play`, and based on the file name ending the system itself detects if it is a wav or LC3 file?

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,59 @@
...
+ */
+bool sd_card_playback_is_active(void);
+
+/**@brief	Play audio from .wav file from SD card. Only support for mono files.
+ *
+ * @param[in]	filename Name of file to be played. Path from root of SD card is accepted.
+ * @param[in]	frame_duration_ms Frame duration in milliseconds.
+ * @param[in]	bit_depth Bit depth.
+ * @param[in]	sample_rate Sample rate.
+ * @param[in]	audio_ch number of channels.
```
koffes, line 31: Remove

koffes (changes requested):

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -52,19 +57,25 @@ struct sw_codec_encoder {
 
 struct sw_codec_decoder {
 	bool enabled;
+	uint8_t audio_mode;
 	uint8_t num_ch;
 	enum audio_channel audio_ch; /* Only used if channel mode is mono */
```
rick1082, line 62: audio_mode, num_ch and audio_ch are very similar, we could improve this, or adding comment to explain the details in here.

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -39,8 +39,13 @@ enum sw_codec_select {
 
 enum sw_codec_num_ch {
 	SW_CODEC_ZERO_CHANNELS,
```
rick1082, line 41: Would there be a condition use zero channel?
If yes, SW_CODEC_ZERO_CHANNEL (remove S) would be good.
If not, then probably we can use

```
enum sw_codec_num_ch {
	SW_CODEC_ONE_CHANNEL = 1,  /* Only use one channel */
	SW_CODEC_TWO_CHANNELS, /* Use both channels */
};
```

I was wondering if we could use number directly instead of enumeration, but since we use switch case to catch this in later, seems the enumeration is the best choice in current stage.

rick1082 (changes requested):

rick1082 (changes requested):
Please let me add some comments in here
For now the PR is focusing on playing encoded stream from SD card, but would be great if we can put "play encoded stream from internal flash" into design consideration, for making the design more feasible. (This could be another task)

applications/nrf5340_audio/src/audio/audio_system.c
```diff
@@ -90,7 +91,10 @@ static void audio_headset_configure(void)
 	}
 #endif /* (CONFIG_STREAM_BIDIRECTIONAL) */
 
-	sw_codec_cfg.decoder.num_ch = SW_CODEC_MONO;
+	sw_codec_cfg.decoder.num_ch = SW_CODEC_ONE_CHANNEL;
+	if (IS_ENABLED(CONFIG_SD_CARD_PLAYBACK)) {
+		sw_codec_cfg.decoder.num_ch++;
```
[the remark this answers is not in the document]
MariusVaardal: I think I have covered all the cases

applications/nrf5340_audio/src/audio/sw_codec_select.c
```diff
@@ -61,7 +68,7 @@ int sw_codec_encode(void *pcm_data, size_t pcm_size, uint8_t **encoded_data, siz
 			}
 			break;
 		}
-		case SW_CODEC_STEREO: {
+		case SW_CODEC_TWO_CHANNELS: {
 			ret = sw_codec_lc3_enc_run(pcm_data_mono[AUDIO_CH_L], pcm_block_size_mono,
```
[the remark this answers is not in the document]
MariusVaardal: I agree that this is a good question

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -59,48 +71,72 @@ int sd_card_list_files(char *path)
...
 	struct fs_file_t f_entry;
 	char abs_path_name[PATH_MAX_LEN + 1] = SD_ROOT_PATH;
-	int ret;
+
+	if (k_sem_count_get(&m_sem_sd_oper_ongoing) <= 0) {
+		LOG_ERR("SD operation ongoing");
+		return -EPERM;
+	}
+
+	k_sem_take(&m_sem_sd_oper_ongoing, K_FOREVER);
```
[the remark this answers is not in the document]
MariusVaardal: Yes, this is simply because i didnt rebase to get the changes from my previous PR. I think it is better now. It is set to wait for 500 ms.

applications/nrf5340_audio/src/modules/sd_card_playback.c
MariusVaardal: So I shouldn't print the error code?

applications/nrf5340_audio/src/modules/sd_card_playback.c
MariusVaardal: Good question. I will take a look at it when I am done with the other stuff:)

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -24,20 +24,30 @@ LOG_MODULE_REGISTER(sd_card, CONFIG_MODULE_SD_CARD_LOG_LEVEL);
 static const char *sd_root_path = "/SD:";
 static FATFS fat_fs;
 static bool sd_init_success;
+K_SEM_DEFINE(m_sem_sd_oper_ongoing, 1, 1);
```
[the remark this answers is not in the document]
MariusVaardal: Yes. There was some concurrency issue when it was left out. It avoids all the concurrency issues for this module. But I think the issue might be resolved on a higher level. It might be possible if I remember correctly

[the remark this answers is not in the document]
MariusVaardal: *Might be possible to do it in the sd card playback module

applications/nrf5340_audio/src/audio/sw_codec_select.h
MariusVaardal: Yes, I will probably rename this at some point as Kristoffer also thought it was a bit vague. Hopefully this will make it more clear

applications/nrf5340_audio/src/modules/sd_card.c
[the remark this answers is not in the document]
rick1082: But would this stop us open and read multiple SD card file at the same time?

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -47,7 +47,11 @@ int sd_card_list_files(char const *const path, char *buf, size_t *buf_size)
 		return -EPERM;
 	}
 
-	k_sem_take(&m_sem_sd_oper_ongoing, K_MSEC(K_SEM_OPER_TIMEOUT_MS));
+	ret = k_sem_take(&m_sem_sd_oper_ongoing, K_MSEC(K_SEM_OPER_TIMEOUT_MS));
+
+	if (ret != 0) {
+		LOG_ERR("Sem take failed. Ret: %d", ret);
```
andvib, line 53: Is this debug code? If not - should it return and not execute the rest of the function?

applications/nrf5340_audio/src/audio/audio_system.c
```diff
@@ -52,7 +53,8 @@ static void audio_gateway_configure(void)
 
 #if (CONFIG_STREAM_BIDIRECTIONAL)
 	sw_codec_cfg.decoder.enabled = true;
-	sw_codec_cfg.decoder.num_ch = SW_CODEC_MONO;
+	sw_codec_cfg.decoder.num_ch = SW_CODEC_ONE_CHANNEL;
+	sw_codec_cfg.decoder.audio_mode = SW_CODEC_MONO;
```
andvib, line 57: I'm not convinced setting it here is enough as the number of channels is still changed later in the function. Maybe it's enough to do a check at the end of the function, and set MONO/STEREO based on the number of channels?

applications/nrf5340_audio/src/audio/sw_codec_select.c
```diff
@@ -51,7 +58,7 @@ int sw_codec_encode(void *pcm_data, size_t pcm_size, uint8_t **encoded_data, siz
 		}
 
 		switch (m_config.encoder.num_ch) {
```
andvib, line 60:
```suggestion
		switch (m_config.encoder.audio_mode) {
```
I think this solution works best if we do it for both gateway and headset, and encode and decode

andvib (changes requested):
Reviewed sw_codec part mostly

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+	for (uint32_t i = 0; i < lc3_playback_cfg.lc3_frames_num; i++) {
+		/* Skip the frame length info to get to the audio data */
+		ret = sd_card_read((char *)&lc3_playback_cfg.lc3_frame_length,
+				   &lc3_playback_cfg.lc3_frame_length_size, &f_seg_read_entry);
+		if (ret < 0) {
+			LOG_ERR("Error when trying to skip file on SD card. Return value: %d", ret);
+			return ret;
+		}
+		uint8_t lc3_frame[lc3_playback_cfg.lc3_frame_length];
+		size_t lc3_fr_len = lc3_playback_cfg.lc3_frame_length;
```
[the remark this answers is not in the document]
MariusVaardal: I dont know why, but I struggle to convert lc3_playback_cfg.lc3_frame_length into size *.

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+	ret = sd_card_close(&f_seg_read_entry);
+	if (ret < 0) {
+		LOG_ERR("Error when closing file. Return value: %d", ret);
+		return ret;
+	}
+	sd_card_playback_active = false;
+	return 0;
+}
+
+static int sd_card_playback_thread_play_wav(void)
```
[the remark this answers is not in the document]
MariusVaardal: I changed the names of the functions

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+
+	sd_card_playback_active = true;
+	while (sd_card_playback_active) {
+
+		ret = sd_card_read(pcm_wav_mono_frame, &pcm_wav_mono_frame_size, &f_seg_read_entry);
+		if (ret < 0) {
+			LOG_ERR("Error when trying to read file on SD card. Return value: %d", ret);
+			return ret;
+		}
+		if (pcm_wav_mono_frame_size < pcm_frame_size) {
```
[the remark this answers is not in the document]
MariusVaardal: It checks whether it is able to fill the 1ms pcm buffer with data from the file. If it is not, then it means that we have reached the end of the file and this can be the last iteration

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+		LOG_ERR("Error when closing file. Return value: %d", ret);
+		return ret;
+	}
+	sd_card_playback_active = false;
+	return 0;
+}
+
+static void sd_card_playback_thread(void *arg1, void *arg2, void *arg3)
+{
+	while (!sw_codec_is_initialized()) {
```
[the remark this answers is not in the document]
MariusVaardal: Hmm. I think one problem with this is that the sd card playback thread might start up before sw_codec is initialized. If i put the sd card playback stuff at the end of main without any delay, I think the check might be run before sw_codec has had the time to initialize.

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+}
+
+void sd_card_playback_lc3(char *filename)
+{
+	playback_file_format = LC3;
+	playback_file_name = filename;
+	k_sem_give(&m_sem_playback);
+}
+
+int sd_card_playback_mix_with_stream(void *const pcm_a, size_t pcm_a_size)
```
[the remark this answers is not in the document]
MariusVaardal: I dont know if I understand this. Do you think I should mix the audio data in the SD card playback module and not from audio datapath? In my head, this seems quite cumbersome. Then I would have to fetch data from the main BT stream and then mix it and then send it back before the next audio frame occurs.

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,59 @@
...
+	SAMPLE_RATE_48K = 48
+};
+
+/**@brief	Figure out whether or not the sd_card_playback module is active.
+ *
+ * @retval	true, if active. false, if not active.
+ */
+bool sd_card_playback_is_active(void);
+
+/**@brief	Play audio from .wav file from SD card. Only support for mono files.
```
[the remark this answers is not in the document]
MariusVaardal: It plays the file as though it contained mono data. The frequency is cut in half and the time it takes to go through the file doubles

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+		LOG_ERR("Audio Lc3 header read failed. Return value: %d", ret);
+		return ret;
+	}
+
+	pcm_frame_size = 2 * lc3_file_header.sample_rate * lc3_file_header.frame_ms / 1000;
+	lc3_playback_cfg.lc3_frames_num =
+		2 * ((lc3_file_header.signal_len_msb << 16) + lc3_file_header.signal_len_lsb) /
+		pcm_frame_size;
+
+	uint8_t pcm_mono_frame[pcm_frame_size];
```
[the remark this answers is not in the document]
MariusVaardal: I have to read the header before I know what pcm frame size is. But should I declare it at the top of the function and then use malloc to allocate the correct amount of bytes?

applications/nrf5340_audio/src/bluetooth/Kconfig.defaults
```diff
@@ -16,9 +16,6 @@ config BT_ECC
 config BT_EXT_ADV
 	default y
 
-config BT_DEVICE_NAME
```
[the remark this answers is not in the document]
MariusVaardal: No. I think the same config is set a few lines above:)

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -53,18 +60,24 @@ struct sw_codec_encoder {
...
+	enum sw_codec_select sw_codec;	 /* sw_codec to be used, e.g. LC3, etc */
 	struct sw_codec_decoder decoder; /* Struct containing settings for decoder */
 	struct sw_codec_encoder encoder; /* Struct containing settings for encoder */
-	bool initialized; /* Status of codec */
+	bool initialized;		 /* Status of codec */
 };
 
+/** @brief  Check if sw_codec is initialized
+ * @return	true, if initialized. false, otherwise
+ */
```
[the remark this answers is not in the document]
alexsven: Not fixed. Brief should be below /**. It should be changed in the entire file at the same time.

applications/nrf5340_audio/src/modules/Kconfig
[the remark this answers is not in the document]
alexsven: Has not been fixed

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
alexsven: Now it is 12, that seem a bit short. Is this due to restrictions from FATFS?

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,336 @@
...
+
+/*File structure of the LC3 encoded files*/
+struct lc3_binary_hdr_t {
+	uint16_t file_id;	 /* Constant value, 0xCC1C */
+	uint16_t hdr_size;	 /* Header size, 0x0012 */
+	uint16_t sample_rate;	 /* Sample frequency / 100 */
+	uint16_t bit_rate;	 /* Bit rate / 100 (total for all channels) */
+	uint16_t channels;	 /* Number of channels */
+	uint16_t frame_ms;	 /* Frame duration in ms * 100 */
+	uint16_t rfu;		 /* RFU value */
```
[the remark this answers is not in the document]
alexsven: Sorry, I meant that you could expand it in the comment, not the variable itself.

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
alexsven: Hmm, that doesn't sound right. When we are using semaphores the thread waiting for data should do just that, wait, until the data is ready. Will have to look at the code using the semaphore to see what can be done.

applications/nrf5340_audio/src/modules/sd_card.c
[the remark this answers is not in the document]
MariusVaardal: Yes

applications/nrf5340_audio/src/bluetooth/Kconfig.defaults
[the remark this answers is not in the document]
MariusVaardal: But I guess if it is wrong then another PR can fix it

alexsven (changes requested):
Had to end mid-review, but at least you have some comments to work on

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
MariusVaardal: OK. I dont really need the initial count to be 1. It works with an init count of 0.

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
MariusVaardal: And the function is a thread function, so i guess i cant return anything inside it(?).

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
MariusVaardal: It doestn work if I, in the line under, write this: 
ret = sd_card_read((char *)lc3_frame, (size_t *)&lc3_fr_len, &f_seg_read_entry);

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,402 @@
+#include "sd_card_playback.h"
+#include "sd_card.h"
+#include "sw_codec_lc3.h"
+#include "sw_codec_select.h"
+#include "pcm_mix.h"
+#include <stdint.h>
+#include <zephyr/sys/ring_buffer.h>
+#include <zephyr/shell/shell.h>
```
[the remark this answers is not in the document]
MariusVaardal: I had a discussion with Andreas and Kristoffer about the fact that the module can only play when the BT stream is up and running. We concluded that this could be fixed in the future since it would require a lot of extra work

applications/nrf5340_audio/prj.conf
```diff
@@ -61,3 +61,4 @@ CONFIG_SD_LOG_LEVEL_OFF=y
 
 # Suppress LOG_INF messages from hci_core
 CONFIG_BT_HCI_CORE_LOG_LEVEL_WRN=y
+CONFIG_SD_CARD_PLAYBACK=y
```
koffes, line 64: This should only be set if we are in DEBUG mode.

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -39,8 +39,13 @@ enum sw_codec_select {
 
 enum sw_codec_num_ch {
 	SW_CODEC_ZERO_CHANNELS,
-	SW_CODEC_MONO, /* Only use one channel */
-	SW_CODEC_STEREO, /* Use both channels */
+	SW_CODEC_ONE_CHANNEL,  /* Only use one channel */
+	SW_CODEC_TWO_CHANNELS, /* Use both channels */
+};
+
+enum sw_codec_audio_mode {
```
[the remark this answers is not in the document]
koffes: We did not discuss this in the meeting. @andvib have you two discussed this naming further?

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
koffes: Ah, then this is better.

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
koffes: We discussed this, and this will not be done in this PR.

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+{
+	int ret;
+
+	ret = sd_card_playback_header_read(playback_file_name);
+	if (ret < 0) {
+		LOG_ERR("Audio Lc3 header read failed. Return value: %d", ret);
+		return ret;
+	}
+
+	pcm_frame_size = 2 * lc3_file_header.sample_rate * lc3_file_header.frame_ms / 1000;
```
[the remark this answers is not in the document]
koffes: This has not been changed.

koffes (changes requested):

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,411 @@
...
+		return ret;
+	}
+
+	sd_card_playback_active = true;
+	while (sd_card_playback_active) {
+
+		ret = sd_card_read(pcm_wav_mono_frame, &pcm_wav_mono_frame_size, &f_seg_read_entry);
+		if (ret < 0) {
+			LOG_ERR("SD card read err: %d", ret);
+			return ret;
```
andvib, line 219: Should `sd_card_playback_active` be set to false before returning? So that a new stream can be started?

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,411 @@
...
+			sd_card_playback_active = false;
+		}
+		/* Wait until there is enough space in the ringbuffer */
+		if (sd_card_playback_active) {
+
+			ret = sd_card_playback_load_ringbuf(pcm_wav_mono_frame,
+							    pcm_wav_mono_frame_size);
+			if (ret < 0) {
+				LOG_ERR("Load ringbuf err: %d.", ret);
+				return ret;
```
andvib, line 231: Set `sd_card_playback_active` to false here as well?

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,411 @@
...
+			}
+		}
+	}
+	ret = sd_card_close(&f_seg_read_entry);
+	if (ret < 0) {
+		LOG_ERR("SD card close file err: %d", ret);
+		return ret;
+	}
+	sd_card_playback_active = false;
```
andvib, line 240: This shouldn't be have to be set to false here as the `while`-loop won't exit unless this is false.

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,411 @@
...
+	return 0;
+}
+
+int sd_card_playback_mix_with_stream(void *const pcm_a, size_t pcm_a_size)
+{
+	int ret;
+	uint8_t pcm_b[pcm_frame_size];
+	size_t write_size = pcm_frame_size;
+
+	ret = sd_card_playback_buffer_set(pcm_b, &write_size);
```
andvib, line 304:
```suggestion
	ret = sd_card_playback_buffer_set(pcm_b, &read_size);
```
... or am I just mixing up the read/write functions now?

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,411 @@
...
+	char file_loc[MAX_PATH_LEN] = "";
+
+	strcat(file_loc, playback_file_path);
+	strcat(file_loc, argv[1]);
+	sd_card_playback_wav(file_loc, wav_playback_cfg.frame_duration_ms,
+			     wav_playback_cfg.bit_depth, SAMPLE_RATE_48K);
+	return 0;
+}
+
+static int cmd_play_lc3_file(const struct shell *shell, size_t argc, char **argv)
```
andvib, line 351: Should the `sw_codec_is_initialized`-check be added here as well?

applications/nrf5340_audio/src/audio/sw_codec_select.h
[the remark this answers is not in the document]
andvib: We haven't, I think we need some input from you guys.

Would it be better if specify the mode is related to the channels? E.g. `sw_codec_channel_mode`?

Also - should we just remove `sw_codec_num_ch` since it's only being used as a number now, or should we leave it as an enum to have a guard for assigning to high numbers? If we keep it I think we must add `SW_CODEC_THREE_CHANNELS` as well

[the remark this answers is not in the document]
andvib: @MariusVaardal We had a small discussion on this offline, and our suggestion is that `sw_codec_num_ch` is renamed to `sw_codec_num_instances`, and `sw_codec_audio_mode` to `sw_codec_channel_mode`

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,411 @@
...
+	lc3_playback_cfg.lc3_frames_num =
+		2 * ((lc3_file_header.signal_len_msb << 16) + lc3_file_header.signal_len_lsb) /
+		pcm_frame_size;
+
+	uint8_t pcm_mono_frame[pcm_frame_size];
+
+	sd_card_playback_active = true;
+	for (uint32_t i = 0; i < lc3_playback_cfg.lc3_frames_num; i++) {
+
+		uint8_t lc3_frame[lc3_playback_cfg.lc3_frame_length];
```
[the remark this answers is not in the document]
MariusVaardal: There might be a better way of doing it, but I am using it as a uint16 some places. If I have a char buffer, then I think I have to struggle a bit because I need to fetch the info from two indices in the buffer

applications/nrf5340_audio/src/modules/sd_card_playback.c
MariusVaardal: Yes

applications/nrf5340_audio/src/modules/sd_card_playback.c
MariusVaardal: Don't know if it is necessary as it is checked in the function it calls

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,411 @@
...
+	size_t lc3_frame_length_size;
+};
+
+enum audio_formats {
+	SD_CARD_PLAYBACK_WAV,
+	SD_CARD_PLAYBACK_LC3,
+};
+
+RING_BUF_DECLARE(m_ringbuf_sound_data_lc3, CONFIG_SD_CARD_PLAYBACK_RING_BUF_SIZE);
+K_SEM_DEFINE(m_sem_load_from_buf_lc3, 0, 1);
```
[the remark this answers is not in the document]
andvib: Bit back and forth now, but with the refactor I realize that maybe `m_sem_lc3_ringbuf_space_available` is more correct for this semaphore

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
andvib: That should not matter in this scope actually. If you have a look at `sizeof` for both `char` and `uint8_t`, you'll see that they're both 1. So any manipulation you have to do to access the array as `uint16_t` will be the same for both (TLDR `char` and `uint8_t` are the same). Since the two functions that are using this array both take the pointer as a `char *` it's safe to define the array as a `char` - if any of the functions need to access it as `uint16_t` they will do the casting themselves (y)

applications/nrf5340_audio/src/modules/sd_card_playback.c
andvib: True. In that case you should make sure that this check is at the same level for both LC3 and WAV - right now the check is in `cmd_play` for WAV, but `sd_card_playback` for LC3. (I don't really have a preference for which of the two as long as it's the same for both)

applications/nrf5340_audio/src/audio/audio_system.c
```diff
@@ -89,7 +94,14 @@ static void audio_headset_configure(void)
 	}
 #endif /* (CONFIG_STREAM_BIDIRECTIONAL) */
 
-	sw_codec_cfg.decoder.num_ch = SW_CODEC_MONO;
+	sw_codec_cfg.decoder.num_ch = SW_CODEC_ONE_CHANNEL;
+	sw_codec_cfg.decoder.channel_mode = SW_CODEC_MONO;
+
+	if (IS_ENABLED(CONFIG_SD_CARD_PLAYBACK)) {
+		/* Need an extra decoder channel to decode data from SD card */
+		sw_codec_cfg.decoder.num_ch++;
```
andvib, line 102: Can we have the case where `num_ch` is 2, and we increment it to 3 here? The `sw_codec_num_ch` enum doesn't have a definition for 3, so we need to safeguard what happens in that case.

andvib (changes requested):

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -38,34 +38,50 @@ enum sw_codec_select {
...
 enum sw_codec_num_ch {
-	SW_CODEC_ZERO_CHANNELS,
-	SW_CODEC_MONO, /* Only use one channel */
-	SW_CODEC_STEREO, /* Use both channels */
+	SW_CODEC_ONE_CHANNEL = 1, /* Only use one channel */
+	SW_CODEC_TWO_CHANNELS,	  /* Use both channels */
+};
+
+enum sw_codec_channel_mode {
+	SW_CODEC_MONO,
```
koffes, line 46:
```suggestion
	SW_CODEC_MONO = 1,
```

applications/nrf5340_audio/src/audio/sw_codec_select.h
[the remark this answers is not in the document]
koffes: Ok, this is not done. Discuss tomorrow.

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,65 @@
...
+ * @retval	true, if active. false, if not active.
+ */
+bool sd_card_playback_is_active(void);
+
+/**
+ * @brief	Play audio from .wav file from SD card. Only support for mono files.
+ *
+ * @note	Only support for mono
+ *
+ * @param[in]	filename Name of file to be played. Path from root of SD card is accepted.
```
koffes, line 31: Missing doc for arguments.

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
koffes: Discuss tomorrow

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
koffes: Yes, that is correct.

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+	uint8_t pcm_b[pcm_frame_size];
+	size_t write_size = pcm_frame_size;
+
+	ret = sd_card_playback_buffer_set(pcm_b, &write_size);
+	if (ret < 0) {
+		LOG_ERR("Error when loading pcm data into buffer. Ret: %d", ret);
+		return ret;
+	}
+	if (write_size == pcm_frame_size) {
+		pcm_mix(pcm_a, pcm_a_size, pcm_b, pcm_frame_size, B_MONO_INTO_A_STEREO_L);
```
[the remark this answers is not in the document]
koffes: The return value is checked now, good, but if this fails, and err will be printed and the function will continue. Discuss if this error should be returned.

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,426 @@
...
+}
+
+int sd_card_playback_init(void)
+{
+	int ret;
+
+	sd_card_playback_thread_id = k_thread_create(
+		&sd_card_playback_thread_data, sd_card_playback_thread_stack,
+		CONFIG_SD_CARD_PLAYBACK_STACK_SIZE, (k_thread_entry_t)sd_card_playback_thread, NULL,
+		NULL, NULL, CONFIG_SD_CARD_PLAYBACK_THREAD_PRIORITY, 0, K_NO_WAIT);
```
koffes, line 342: To stay in sync with other names: 
```suggestion
		NULL, NULL, K_PRIO_PREEMPT(CONFIG_SD_CARD_PLAYBACK_THREAD_PRIO), 0, K_NO_WAIT);
```

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,362 @@
...
+}
+
+int sd_card_playback_init(void)
+{
+	int ret;
+
+	sd_card_playback_thread_id = k_thread_create(
+		&sd_card_playback_thread_data, sd_card_playback_thread_stack,
+		SD_CARD_PLAYBACK_STACK_SIZE, (k_thread_entry_t)sd_card_playback_thread, NULL, NULL,
+		NULL, K_PRIO_PREEMPT(4), 0, K_NO_WAIT);
```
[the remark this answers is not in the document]
koffes: We need to discuss the priority in relation to other threads.

applications/nrf5340_audio/src/modules/sd_card_playback.c
koffes: Any update on this? Though, I do not want to block the PR with this featurecreep.

applications/nrf5340_audio/src/modules/Kconfig
```diff
@@ -124,11 +128,34 @@ config VOLUME_MSG_SUB_STACK_SIZE
...
 config VOLUME_MSG_SUB_QUEUE_SIZE
 	int "Queue size for volume message subscriber"
 	default 4
 
 endmenu # Zbus
+
+#----------------------------------------------------------------------------#
+menuconfig SD_CARD_PLAYBACK
+	bool "If true, SD card playback is enabled"
+	default n
```
koffes, line 143:
```suggestion
menuconfig SD_CARD_PLAYBACK
	bool "If true, SD card playback is enabled"
	select EXPERIMENTAL
	default n
```

koffes (changes requested):

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,426 @@
...
+	int ret;
+	uint8_t pcm_b[pcm_frame_size];
+	size_t read_size = pcm_frame_size;
+
+	ret = sd_card_playback_ringbuf_read(pcm_b, &read_size);
+	if (ret < 0) {
+		LOG_ERR("Loading data into buffer err: %d", ret);
+		return ret;
+	}
+	if (read_size == pcm_frame_size) {
```
alexsven, line 323: Newline after { usually

alexsven (changes requested):
More to come, haven't finished sd_card.c

applications/nrf5340_audio/src/modules/sd_card_playback.c
MariusVaardal: I have not done any work on this feature yet

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,65 @@
...
+#define _SD_CARD_PLAYBACK_H_
+
+#include <zephyr/kernel.h>
+
+/**
+ * @brief	Sample rates supported by the sd_card_playback_wav-function.
+ */
+enum sd_playback_sample_rates {
+	SAMPLE_RATE_48K = 48000
+};
```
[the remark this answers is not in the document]
MariusVaardal: Hmm. I think there will be a problem in the audio_datapath_stream_out function if the ringbuffer in sd_card_playback.c contains anything other than 48k mono data. As far as I understand, it expects to get 48k mono pcm data from sd_card_playback, so the the audio will be played off twice as fast in audio datapath if the sample rate is 96k, for example

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,426 @@
...
+		LOG_ERR("Sw codec not initialized");
+		return -EACCES;
+	}
+	playback_file_format = SD_CARD_PLAYBACK_LC3;
+	playback_file_name = filename;
+	k_sem_give(&m_sem_playback);
+	return 0;
+}
+
+int sd_card_playback_mix_with_stream(void *const pcm_a, size_t pcm_a_size)
```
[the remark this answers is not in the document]
MariusVaardal: I added a check in sd_card_playback_mix_with_stream to ensure that sd_card_playback_active is true. This is not true before pcm_frame size is set

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -42,12 +42,12 @@ int sd_card_list_files(char const *const path, char *buf, size_t *buf_size)
...
 	size_t used_buf_size = 0;
 
-	if (k_sem_count_get(&m_sem_sd_oper_ongoing) <= 0) {
-		LOG_ERR("SD operation ongoing");
-		return -EPERM;
-	}
+	ret = k_sem_take(&m_sem_sd_oper_ongoing, K_MSEC(K_SEM_OPER_TIMEOUT_MS));
 
-	k_sem_take(&m_sem_sd_oper_ongoing, K_MSEC(K_SEM_OPER_TIMEOUT_MS));
+	if (ret < 0) {
```
erikrobstad, line 47:
```suggestion
	if (ret) {
```

applications/nrf5340_audio/src/modules/sd_card.c
```diff
@@ -127,12 +127,12 @@ int sd_card_open_write_close(char const *const filename, char const *const data,
 	struct fs_file_t f_entry;
 	char abs_path_name[PATH_MAX_LEN + 1] = SD_ROOT_PATH;
 
-	if (k_sem_count_get(&m_sem_sd_oper_ongoing) <= 0) {
-		LOG_ERR("SD operation ongoing");
-		return -EPERM;
-	}
+	ret = k_sem_take(&m_sem_sd_oper_ongoing, K_MSEC(K_SEM_OPER_TIMEOUT_MS));
```
erikrobstad, line 131: We usually do not have newlines between the function and the if (ret) check. Valid several places in this file

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,433 @@
+#include "sd_card_playback.h"
```
erikrobstad, line 1: A license must be added

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,433 @@
...
+static int sd_card_playback_ringbuf_read(uint8_t *buf, size_t *size)
+{
+	int ret;
+	uint16_t read_size;
+
+	ret = k_mutex_lock(&mtx_ringbuf, K_MSEC(1));
+	if (ret < 0) {
+		LOG_ERR("Unable to take mutex. Ret: %d", ret);
+		return ret;
+	}
```
erikrobstad, line 80: Newline after "}".
Valid several places in this file

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,433 @@
...
+	ret = sd_card_list_files(playback_file_path, buf, &buf_size);
+	if (ret < 0) {
+		shell_error(shell, "List files err: %d", ret);
+		return ret;
+	}
+	shell_print(shell, "%s", buf);
+	return 0;
+}
+
+/* Creating subcommands (level 1 command) array for command "demo". */
```
erikrobstad, line 424: Hmm, what does these two "command "demo"" comments mean? Maybe this is copied from button_handler.c?

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * @brief	Initialize sd_card_playback module. Create sd_card_playback thread.
+ *
+ * @retval	0 on success.
+ *		Otherwise, error from underlying drivers.
+ */
+int sd_card_playback_init(void);
+
+#endif
```
erikrobstad, line 75:
```suggestion
#endif /* _SD_CARD_PLAYBACK_H_ */
```

erikrobstad (changes requested):

alexsven (changes requested):

applications/nrf5340_audio/src/modules/CMakeLists.txt
```diff
@@ -14,3 +14,8 @@ target_sources(app PRIVATE
 	       ${CMAKE_CURRENT_SOURCE_DIR}/power_meas.c
 	       ${CMAKE_CURRENT_SOURCE_DIR}/sd_card.c
 )
+
+if (CONFIG_SD_CARD_PLAYBACK)
```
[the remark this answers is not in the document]
MariusVaardal: Yes

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,65 @@
...
+ * @note	Only support for mono
+ *
+ * @param[in]	filename	Name of file to be played. Path from root of SD card is accepted.
+ */
+int sd_card_playback_lc3(char *filename);
+
+/**
+ * @brief	Mix pcm data from sd_card_playback module with audio stream out.
+ *
+ * @param[in, out]	pcm_a	Buffer into which to mix pcm data from lc3_module.
```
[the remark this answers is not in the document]
MariusVaardal: But it is read by the pcm mix function

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,433 @@
...
+		return ret;
+	}
+
+	sd_card_playback_active = true;
+	while (sd_card_playback_active) {
+
+		ret = sd_card_read(pcm_wav_mono_frame, &pcm_wav_mono_frame_size, &f_seg_read_entry);
+		if (ret < 0) {
+			LOG_ERR("SD card read err: %d", ret);
+			sd_card_playback_active = false;
```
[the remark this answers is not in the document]
MariusVaardal: Good point

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,433 @@
...
+		2 * ((lc3_file_header.signal_len_msb << 16) + lc3_file_header.signal_len_lsb) /
+		pcm_frame_size;
+
+	uint8_t pcm_mono_frame[pcm_frame_size];
+
+	sd_card_playback_active = true;
+	for (uint32_t i = 0; i < lc3_playback_cfg.lc3_frames_num; i++) {
+
+		ret = sd_card_read((char *)&lc3_playback_cfg.lc3_frame_length,
+				   &lc3_playback_cfg.lc3_frame_length_size, &f_seg_read_entry);
```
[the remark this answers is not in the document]
MariusVaardal: Pretty sure they do. At least the ones from T2 labs.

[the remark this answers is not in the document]
MariusVaardal: I have sent a message to Rick. He will probably respond by next week

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,555 @@
...
+	}
+
+	ret = sd_card_playback_ringbuf_read(pcm_b, &read_size);
+	if (ret) {
+		LOG_ERR("Loading data into buffer err: %d", ret);
+		return ret;
+	}
+
+	if (read_size > 0) {
+		ret = pcm_mix(pcm_a, pcm_a_size, pcm_b, pcm_frame_size, B_MONO_INTO_A_STEREO_L);
```
alexsven, line 443: This is a segmentation fault waiting to happen. If sd_card_playback_ringbuf_read returns less than pcm_fram_size you will start to mix in data that has not been initialized. I suggest doing:
```suggestion
		ret = pcm_mix(pcm_a, pcm_a_size, pcm_b, read_size, B_MONO_INTO_A_STEREO_L);
```

If sd_card_playback_ringbuf_read returns less than pcm_frame_size you would get an error, but you don't check the return value in audio_datapath.c

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,65 @@
...
+ * @param[in]		size	Size of input buffer.
+ *
+ * @retval	0 on success.
```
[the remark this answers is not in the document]
alexsven: Has not been done

applications/nrf5340_audio/src/modules/sd_card_playback.h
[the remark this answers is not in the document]
alexsven: ok

alexsven (changes requested):

applications/nrf5340_audio/src/audio/audio_system.h
```diff
@@ -53,6 +53,13 @@ void audio_system_stop(void);
  */
 int audio_system_fifo_rx_block_drop(void);
 
+/**
+ * @brief       Initialize the audio_system
```
alexsven, line 57: I hope this function is not initializeing the audio_system ;)

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,555 @@
...
+	}
+
+	pcm_mono_frame_size = wav_file_header.byte_rate * FRAME_DURATION_MS / 1000;
+	/* This buffer must be allocated after pcm mono frame size is calculated */
+	uint8_t pcm_mono_frame[pcm_mono_frame_size];
+
+	pcm_frame_size = pcm_mono_frame_size;
+	/* wav_file_header.wav_size = file size - 8 */
+	audio_length_bytes = wav_file_header.wav_size + 8 - sizeof(wav_header_t);
+	n_iter = ceil((float)audio_length_bytes / (float)pcm_frame_size);
```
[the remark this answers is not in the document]
MariusVaardal: The idea is that the sd_card_read func just reads to the end of the file and writes to the pcm_mono_frame_size variable how many bytes it has read. I then use this variable in the function call to sd_card_playback_ringbuf_write() and in the pcm_mix function there seems to be support for mixing buffers of different sizes. I could of course also simply discard the last bit of data

applications/nrf5340_audio/src/audio/sw_codec_select.c
```diff
@@ -235,6 +240,8 @@ int sw_codec_uninit(struct sw_codec_config sw_codec_cfg)
 		LOG_ERR("Unsupported codec: %d", m_config.sw_codec);
 		return false;
 	}
+
+	m_config.initialized = false;
 	return 0;
```
erikrobstad, line 245: Newline before return here

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,555 @@
...
+	for (int i = 0; i < n_iter; i++) {
+		/* Read a chuck of audio data from file */
```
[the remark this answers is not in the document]
MariusVaardal: It is no particular reason. It lines up well with the size of the ringbuffer (20 ms, but this is configurable) and the read size from audio_datapath_stream_out (10 ms). I could double the chunk size and make it equal to the size of the ringbuffer.

applications/nrf5340_audio/src/modules/CMakeLists.txt
```diff
@@ -14,3 +14,8 @@ target_sources(app PRIVATE
...
+        target_sources(app PRIVATE
+                ${CMAKE_CURRENT_SOURCE_DIR}/sd_card_playback.c)
```
erikrobstad, line 20: Nitpick: The tabbing of this line is not the same as above

erikrobstad (changes requested):

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,555 @@
...
+		}
+
+		ret = sd_card_playback_ringbuf_write((char *)pcm_mono_frame, pcm_mono_write_size);
+		if (ret < 0) {
+			LOG_ERR("Load ringbuf err: %d", ret);
+			sd_card_playback_active = false;
+			break;
+		}
+	}
+
```
andvib, line 345: This got a bit messy with back and forth - but you can set `sd_card_playback_active` to `false` here in the same way you're doing it for WAV so they're as similar as possible

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
MariusVaardal: The reason is that it is easier to share code with the LC3 portion of the file, which uses 10 ms chunks

applications/nrf5340_audio/src/modules/sd_card_playback.c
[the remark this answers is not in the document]
alexsven: Never mind, I didn't see that sd_card_read would change the `pcm_mono_frame_size`

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,567 @@
...
+LOG_MODULE_REGISTER(sd_card_playback, CONFIG_MODULE_SD_CARD_PLAYBACK_LOG_LEVEL);
+
+#define MAX_PATH_LEN	    (CONFIG_FS_FATFS_MAX_LFN)
+#define LIST_FILES_BUF_SIZE 512
+#define FRAME_DURATION_MS   (CONFIG_AUDIO_FRAME_DURATION_US / 1000)
+
+#define WAV_FORMAT_PCM	    1
+#define WAV_SAMPLE_RATE_48K 48000
+
+#pragma pack(1)
```
alexsven, line 30: Any particular reason for wanting to override the compiler here? Same for the pragma below. This might cause the speed to decrease, and I don't necessarily see why we would need to alter the alignment of these structs

alexsven: Talked to Andreas and was told it was to easier read out the header of lc3/wav. Suggest using __packed instead, see le_audio.rx.h for an example

applications/nrf5340_audio/src/modules/CMakeLists.txt
erikrobstad: I was thinking like this, but you don't have to change unless other changes are needed as well (you don't need to push a new commit only for this) :)
image

erikrobstad (approved):

applications/nrf5340_audio/src/modules/sd_card_playback.c
```diff
@@ -0,0 +1,555 @@
...
+	ret = sd_card_open(playback_file_name, &f_seg_read_entry);
+	if (ret) {
+		LOG_ERR("Open SD card file err: %d", ret);
+		return ret;
+	}
+
+	/* Read the file header */
+	ret = sd_card_read((char *)&lc3_file_header, &lc3_file_header_size, &f_seg_read_entry);
+	if (ret < 0) {
+		LOG_ERR("Read SD card file err: %d", ret);
```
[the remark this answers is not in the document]
andvib: This looks to still be missing

andvib (changes requested):

rick1082 (approved):

koffes (approved):

alexsven (approved):

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef _SD_CARD_PLAYBACK_H_
+#define _SD_CARD_PLAYBACK_H_
+
+#include <zephyr/kernel.h>
+
+/**
+ * @brief	Figure out whether or not the sd_card_playback module is active.
```
greg-fer, line 13:
```suggestion
 * @brief	Check whether or not the SD card playback module is active.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @retval	true    Active.
+ * @retval	false   Not active.
+ */
+bool sd_card_playback_is_active(void);
+
+/**
+ * @brief	Play audio from .wav file from SD card. Only support for mono files.
```
greg-fer, line 21:
```suggestion
 * @brief	Play audio from .wav file from the SD card. Only support for mono files.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @note	Only support for 48k mono files
```
greg-fer, line 23:
```suggestion
 * @note	Supports only 48k mono files.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param[in]	filename	Name of file to be played. Path from root of SD card is accepted.
```
greg-fer, line 25:
```suggestion
 * @param[in]	filename	Name of file to be played. Path from the root of the SD card is accepted.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @retval	0       Success.
+ * @retval	-EACCES SW codec is not initialized.
+ */
+int sd_card_playback_wav(char *filename);
+
+/**
+ * @brief	Play audio from LC3 file from SD card.
```
greg-fer, line 33:
```suggestion
 * @brief	Play audio from an LC3 file from the SD card.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @note	Only support for mono
```
greg-fer, line 35:
```suggestion
 * @note	Supports only mono files.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param[in]	filename	Name of file to be played. Path from root of SD card is accepted.
```
greg-fer, line 37:
```suggestion
 * @param[in]	filename	Name of file to be played. Path from the root of the SD card is accepted.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @retval	0       Success.
+ * @retval	-EACCES SW codec is not initialized.
+ */
+int sd_card_playback_lc3(char *filename);
+
+/**
+ * @brief	Mix pcm data from sd_card_playback module with audio stream out.
```
greg-fer, line 45:
```suggestion
 * @brief	Mix the PCM data from the SD card playback module with the audio stream out.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param[in, out]	pcm_a	Buffer into which to mix pcm data from lc3_module.
```
greg-fer, line 47:
```suggestion
 * @param[in, out]	pcm_a	Buffer into which to mix PCM data from the LC3 module.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ * @param[in]		size	Size of input buffer.
```
greg-fer, line 48:
```suggestion
 * @param[in]		size_t	Size of the input buffer.
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @retval	0       Success.
+ * @retval      -EACCES SD card playback is not active.
+ * @retval      Otherwise, error from underlying drivers.
+ */
+int sd_card_playback_mix_with_stream(void *const pcm_a, size_t pcm_a_size);
+
+/**
+ * @brief	Initialize sd_card_playback module. Create sd_card_playback thread.
```
greg-fer, line 57:
```suggestion
 * @brief	Initialize the SD card playback module. Create the SD card playback thread.
```

applications/nrf5340_audio/src/modules/Kconfig
```diff
@@ -129,11 +133,35 @@ config VOLUME_MSG_SUB_STACK_SIZE
...
+#----------------------------------------------------------------------------#
+menuconfig SD_CARD_PLAYBACK
+	bool "Enable playback from SD card"
+	select EXPERIMENTAL
+	default n
+
+if SD_CARD_PLAYBACK
+
+config SD_CARD_PLAYBACK_STACK_SIZE
+	int "Stack size for SD card playback thread"
```
greg-fer, line 154:
```suggestion
	int "Stack size for the SD card playback thread"
```

applications/nrf5340_audio/src/modules/Kconfig
```diff
@@ -129,11 +133,35 @@ config VOLUME_MSG_SUB_STACK_SIZE
...
+	default 4096
+
+config SD_CARD_PLAYBACK_RING_BUF_SIZE
+	int "Size of ringbuffer for the SD card playback module"
```
greg-fer, line 158:
```suggestion
	int "Size of the ring buffer for the SD card playback module"
```

applications/nrf5340_audio/src/modules/Kconfig
```diff
@@ -129,11 +133,35 @@ config VOLUME_MSG_SUB_STACK_SIZE
...
+	default 960
+
+config SD_CARD_PLAYBACK_THREAD_PRIORITY
+	int "Priority for SD card playback thread"
```
greg-fer, line 162:
```suggestion
	int "Priority for the SD card playback thread"
```

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -65,7 +67,16 @@ struct sw_codec_config {
 	bool initialized;		 /* Status of codec */
 };
 
-/**@brief	Encode PCM data and output encoded data
+/**
+ * @brief	Check if sw_codec is initialized
```
greg-fer, line 71:
```suggestion
 * @brief	Check if the software codec is initialized.
```

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -65,7 +67,16 @@ struct sw_codec_config {
...
+ *
+ * @retval	true	SW codec is initialized.
+ * @retval	false	SW codec is not initialized.
+ */
+bool sw_codec_is_initialized(void);
+
+/**
+ * @brief	Encode PCM data and output encoded data
```
greg-fer, line 79:
```suggestion
 * @brief	Encode PCM data and output encoded data.
```
Periods are missing at the end of the API documentation sentences throughout the file. Fix globally please.

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -92,7 +104,8 @@ int sw_codec_encode(void *pcm_data, size_t pcm_size, uint8_t **encoded_data, siz
 int sw_codec_decode(uint8_t const *const encoded_data, size_t encoded_size, bool bad_frame,
 		    void **pcm_data, size_t *pcm_size);
 
-/**@brief	Uninitialize sw_codec and free allocated space
+/**
+ * @brief	Uninitialize sw_codec and free allocated space
```
greg-fer, line 108:
```suggestion
 * @brief	Uninitialize the software codec and free the allocated space.
```

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -102,7 +115,8 @@ int sw_codec_decode(uint8_t const *const encoded_data, size_t encoded_size, bool
  */
 int sw_codec_uninit(struct sw_codec_config sw_codec_cfg);
 
-/**@brief	Initialize sw_codec and statically or dynamically
+/**
+ * @brief	Initialize sw_codec and statically or dynamically
```
greg-fer, line 119:
```suggestion
 * @brief	Initialize the software codec and statically or dynamically
```

applications/nrf5340_audio/src/audio/sw_codec_select.h
```diff
@@ -102,7 +115,8 @@ int sw_codec_decode(uint8_t const *const encoded_data, size_t encoded_size, bool
...
  *		allocate memory to be used, depending on selected codec
```
greg-fer, line 120:
```suggestion
 *		allocate memory to be used, depending on the selected codec
```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef _SD_CARD_PLAYBACK_H_
+#define _SD_CARD_PLAYBACK_H_
+
```
greg-fer, line 9:
```suggestion

/**
 * @file
 * @defgroup sd_card_playback SD card playback
 * @{
 * @brief The SD card playback module for nRF5340 Audio.
 */

```

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,63 @@
...
+ */
+int sd_card_playback_mix_with_stream(void *const pcm_a, size_t pcm_a_size);
+
+/**
+ * @brief	Initialize sd_card_playback module. Create sd_card_playback thread.
+ *
+ * @return	0 on success, otherwise, error from underlying drivers.
+ */
+int sd_card_playback_init(void);
+
```
greg-fer, line 62:
```suggestion

/**
 * @}
 */

```

applications/nrf5340_audio/src/modules/sd_card_playback.h
greg-fer, line 45: This line is causing the doc build issue:

image

greg-fer (changes requested):

andvib (approved):

applications/nrf5340_audio/src/modules/sd_card_playback.h
```diff
@@ -0,0 +1,76 @@
...
+ * @brief	Mix the PCM data from the SD card playback module with the audio stream out.
+ *
+ * @param[in, out]	pcm_a	Buffer into which to mix PCM data from the LC3 module.
+ * @param[in]		size_t	Size of the input buffer.
+ *
+ * @retval	0       Success.
+ * @retval      -EACCES SD card playback is not active.
+ * @retval      Otherwise, error from underlying drivers.
+ */
+int sd_card_playback_mix_with_stream(void *const pcm_a, size_t pcm_a_size);
```
greg-fer, line 63: Not a code expert, but should the param[in] mention `pcm_a_size` and `size_t` be a `*size_t` instead?
This is related to the doc build issue.

greg-fer (approved):
Docs will be covered in https://nordicsemi.atlassian.net/browse/OCT-2790 in coordination with @andvib

PR #11912, opened 2023-07-31
modules: trusted-firmware-m: Require provisioning for IAT

Vge0rge:
Require provisioning when the initial attestation service(IAT) is enabled. Allowing IAT without any provisioning data is not useful and it increases unused code.

Ref: NCSDK-18032


SebastianBoe (approved):

joerchan (dismissed):
@Vge0rge There is a perhaps a task duplication, can you take a look at NCSDK-19927 and remove the workaround also?

PR #11924, opened 2023-08-01
samples: cellular: modem_shell: sock: Add packet number prefix

trantanen:
Sock tool: Add `--packet_number_prefix` option to include an index of the packet that is sent in periodic sending. This enables tracking of which send operations have been sent properly.

### DIFF
--- a/samples/cellular/modem_shell/src/sock/sock.c
+++ b/samples/cellular/modem_shell/src/sock/sock.c
@@ -50,6 +50,7 @@ extern struct k_work_q mosh_common_work_q;
 struct data_transfer_info {
 	struct k_work_delayable work;
 	int interval;
+	int packet_number_prefix;
 	void *parent; /* Type is sock_info */
 	bool data_format_hex; /* Print data in hex format vs. normal string */
 };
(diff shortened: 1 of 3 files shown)


samples/cellular/modem_shell/src/sock/sock.c
```diff
@@ -933,7 +948,22 @@ int sock_send_data(
...
+			if (packet_number_prefix) {
+				/* Make room for packet buffer prefix */
+				if (data_format_hex) {
+					mosh_error(
+						"Hexadecimal format cannot be used with "
+						"packet number prefix");
+					return -EINVAL;
+				}
+
+				sprintf(socket_info->send_buffer, "0000%s", data_out);
```
tokangas, line 961: The length of the send buffer is `data_out_length`, so this writes past the end of the buffer, right?

trantanen: Yep, this is something I was supposed to double check and seems I remembered the buffer allocation wrong.

tokangas (approved):

jhirsi (approved):

PR #11925, opened 2023-08-01
lib: nrf_cloud_coap: add nrf_cloud_coap_bytes_send

maxd-nordic:
This patch adds a function to nrf cloud coap that allows sending of app-specific raw bytes data.

### DIFF
--- a/include/net/nrf_cloud_coap.h
+++ b/include/net/nrf_cloud_coap.h
@@ -136,6 +136,18 @@ int nrf_cloud_coap_pgps_url_get(struct nrf_cloud_rest_pgps_request const *const 
 int nrf_cloud_coap_sensor_send(const char *app_id, double value, int64_t ts_ms);
 
 /**
+ * @brief Send raw bytes to nRF Cloud.
+ * @param[in]     app_id The app_id identifying the type of data.
+ * @param[in]     buf buffer with binary string.
+ * @param[in]     buf_len  length of buf in bytes.
+ * @param[in]     ts_ms  Timestamp the data was measured, or NRF_CLOUD_NO_TIMESTAMP.
+ *
+ * @retval 0 If successful.
+ *          Otherwise, a (negative) error code is returned.
+ */
+int nrf_cloud_coap_bytes_send(const char *app_id, uint8_t *buf, size_t buf_len, int64_t ts_ms);
+
+/**
  * @brief Send the device location in the @ref nrf_cloud_gnss_data PVT field to nRF Cloud.
  *
  *  The CoAP message is sent as a non-confirmable CoAP message. Only
--- a/subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c
+++ b/subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c
@@ -20,6 +20,7 @@
 #include "nrf_cloud_codec_internal.h"
 #include "nrf_cloud_mem.h"
 #include "coap_codec.h"
+#include "msg_encode.h"
 
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(nrf_cloud_coap, CONFIG_NRF_CLOUD_COAP_LOG_LEVEL);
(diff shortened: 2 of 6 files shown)


plskeggs:
I need to check with the backend team whether this will be processed correctly or discarded.

include/net/nrf_cloud_coap.h
```diff
@@ -135,6 +135,18 @@ int nrf_cloud_coap_pgps_url_get(struct nrf_cloud_rest_pgps_request const *const
  */
 int nrf_cloud_coap_sensor_send(const char *app_id, double value, int64_t ts_ms);
 
+/**
+ * @brief Send raw bytes to nRF Cloud.
+ * @param[in]     app_id The app_id identifying the type of data.
```
coderbyheart, line 140: *question:* why does this need an appId, after all raw implies that the cloud shouldn't care about it and app ID is a nRF Cloud specific feature that allows it to distinguish messages for showing them in the UI.

include/net/nrf_cloud_coap.h
```diff
@@ -135,6 +135,18 @@ int nrf_cloud_coap_pgps_url_get(struct nrf_cloud_rest_pgps_request const *const
...
+ * @param[in]     buf buffer with binary string.
+ * @param[in]     buf_len  length of buf in bytes.
+ * @param[in]     ts_ms  Timestamp the data was measured, or NRF_CLOUD_NO_TIMESTAMP.
```
coderbyheart, line 143: *question:* why enforce having a timestamp here? What if raw data is a collection of messages?

maxd-nordic: I was unsure if this might be needed for compatibility so I pulled it in. Can be removed if unnecessary.

include/net/nrf_cloud_coap.h
maxd-nordic: I would definitely prefer a more direct channel for custom messages. If we can support it in the backend, we should.

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c
```diff
@@ -160,6 +161,40 @@ int nrf_cloud_coap_pgps_url_get(struct nrf_cloud_rest_pgps_request const *const
 }
 #endif /* CONFIG_NRF_CLOUD_PGPS */
 
+int nrf_cloud_coap_bytes_send(const char *app_id, uint8_t *buf, size_t buf_len, int64_t ts_ms)
```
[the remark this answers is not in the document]
maxd-nordic: done

include/net/nrf_cloud_coap.h
plskeggs: This kind of architecture change needs to wait until all the parties involved are back from vacation.

theflintstone: But should we keep some kind of categorization way. The AppID provides info how to handle the message, what is the time-to-leave, who has subscribed it etc. For now the AppID is for UI but I see many other needs for it

I would understand that devices could send encrypted messages but still categorize those.

maxd-nordic:
I think it might be best to actually have another uri next to d2c and c2d: "raw". This could be used for sending/receiving raw byte messages. The bridge could put them in a base64 string and wrap into a json for further forwarding.

maxd-nordic:
Waiting for the solution of the nRF Cloud team.

plskeggs:
> I think it might be best to actually have another uri next to d2c and c2d: "raw". This could be used for sending/receiving raw byte messages. The bridge could put them in a base64 string and wrap into a json for further forwarding.

@maxd-nordic this is implemented now and available on dev.  We have not implemented nrf_cloud library support in NCS for it yet.  Would you re-open your PR and try it out?

You might consider using `COAP_CONTENT_FORMAT_APP_OCTET_STREAM` instead of `COAP_CONTENT_FORMAT_APP_CBOR` to be even more generic, but it's up to you.  You can skip using the CBOR encoding changes and the CDDL changes, and just send the raw payload, whatever it happens to be.

PR #11927, opened 2023-08-01
caf: modules: added a custom command to the shell

tony-le-24:
Added a custom command to simulate a button pressed event to the app_event_manager shell. This command then will be send using RTT in our nrf_cloud_rest_cell_location sample (with RTT Kconfigs enabled).


subsys/app_event_manager/app_event_manager_shell.c
```diff
@@ -166,3 +167,27 @@ SHELL_STATIC_SUBCMD_SET_CREATE(sub_app_event_manager,
...
+{
+	if (argc < 3) {
+		shell_error(shell, "Missing required parameter(s)\n");
+		return -EINVAL;
+	}
+
+	bool pressed = false;
+	struct button_event *event = new_button_event();
+
+	if (strcmp(argv[2], "true") == 0) {
```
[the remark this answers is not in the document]
tony-le-24: Yes, it can be anything.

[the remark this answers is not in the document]
glarsennordic: Agreed, this parameter being an integer would be preferred -- though I think you will need to use atoi before casting to bool

subsys/app_event_manager/app_event_manager_shell.c
```diff
@@ -166,3 +167,27 @@ SHELL_STATIC_SUBCMD_SET_CREATE(sub_app_event_manager,
 
 SHELL_CMD_REGISTER(app_event_manager, &sub_app_event_manager,
 		   "Application Event Manager commands", NULL);
+
+static int button_sim_handler(const struct shell *shell, size_t argc, char **argv)
+{
+	if (argc < 3) {
+		shell_error(shell, "Missing required parameter(s)\n");
```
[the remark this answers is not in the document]
tony-le-24: the `\n` is needed since `shell_error` doesn't created new line automatically.

subsys/app_event_manager/app_event_manager_shell.c
```diff
@@ -166,3 +167,27 @@ SHELL_STATIC_SUBCMD_SET_CREATE(sub_app_event_manager,
...
+	}
+
+	bool pressed = false;
+	struct button_event *event = new_button_event();
+
+	if (strcmp(argv[2], "true") == 0) {
+		pressed = true;
+	}
+
+	event->key_id = (uint16_t)argv[1];
```
glarsennordic, line 185: Can you actually just cast to uint16_t like this? I think you need to use atoi don't you?

subsys/app_event_manager/app_event_manager_shell.c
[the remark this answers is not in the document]
jayteemo: it looks like it does though.

subsys/app_event_manager/app_event_manager_shell.c
```diff
@@ -166,3 +167,27 @@ SHELL_STATIC_SUBCMD_SET_CREATE(sub_app_event_manager,
...
+	event->pressed = pressed;
+	APP_EVENT_SUBMIT(event);
+
+	return 0;
+}
+
+SHELL_CMD_REGISTER(button_sim, NULL,
```
glarsennordic, line 192: Can you integrate into the `SHELL_STATIC_SUBCMD_SET_CREATE` call above instead of creating a totally separate command?

plskeggs:
Looks good to me. You'll need to update the changelog and this page:
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/libraries/others/app_event_manager.html

subsys/app_event_manager/app_event_manager_shell.c
```diff
@@ -7,6 +7,7 @@
 #include <stdlib.h>
 #include <zephyr/shell/shell.h>
 #include <app_event_manager.h>
+#include <caf/events/button_event.h>
```
glarsennordic, line 10: I'm realizing this creates a backward dependency from `app_event_manager` to `CAF`

Maybe it would be better to create a CAF shell. Sorry I misled you on this one. If you go this route then of course you can ignore my comment about integrating into the `SHELL_STATIC_SUBCMD_SET_CREATE` call, and just move your additions to their own file somewhere in the CAF library

Though in that case you would ideally still use `SHELL_STATIC_SUBCMD_SET_CREATE` and make your button sim command a subcommand of the CAF shell root command

subsys/app_event_manager/app_event_manager_shell.c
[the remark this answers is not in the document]
tony-le-24: You're right, my bad. Thought it works the same as `shell_fprintf`.

subsys/app_event_manager/app_event_manager_shell.c
tony-le-24: it works with casting but will use atoi to be safer.

glarsennordic: It probably appears to work, but if I type `0` into that as-is it'll interpret that as the number 48 since the ASCII hex code for `0` is `0x30`

subsys/app_event_manager/app_event_manager_shell.c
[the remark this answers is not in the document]
jayteemo: oh right, just check for zero, no need for atoi in this case:
`event->pressed = argv[2] != '0'`

subsys/app_event_manager/app_event_manager_shell.c
plskeggs: creating a CAF shell is what I was anticipating...

glarsennordic: > creating a CAF shell is what I was anticipating...

I'm only suggesting a skeletal one with a single command. I don't think that is any less reasonable than adding a command supporting a single CAF event to the app_event_manager shell

glarsennordic: > creating a CAF shell is what I was anticipating...

Yeah, sorry about that. I encouraged Tony to add a command to this existing shell because it already exists, but forgot that `app_event_manager` and `CAF` are technically distinct, and the former is a dependency of the latter.

subsys/app_event_manager/app_event_manager_shell.c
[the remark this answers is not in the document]
tony-le-24: gotta use `strcmp` since `argv[2] != '0'` always gives 1, weird

jayteemo (approved):

subsys/caf/modules/shell.c
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+#include <zephyr/shell/shell.h>
+#include <app_event_manager.h>
+#include <caf/events/button_event.h>
+
+static int set_button_handler(const struct shell *shell, size_t argc, char **argv)
+{
+	if (argc < 3) {
+		shell_error(shell, "Missing required parameter(s)");
```
jayteemo, line 14: where is it documented what the required parameters are?

subsys/caf/modules/CMakeLists.txt
```diff
@@ -37,4 +37,7 @@ zephyr_library_sources_ifdef(CONFIG_CAF_SENSOR_DATA_AGGREGATOR sensor_data_aggre
 zephyr_library_sources_ifdef(CONFIG_CAF_SETTINGS_LOADER settings_loader.c)
 
 zephyr_library_sources_ifdef(CONFIG_CAF_BLE_SMP ble_smp.c)
+
+zephyr_library_sources_ifdef(CONFIG_CAF_SHELL shell.c)
```
glarsennordic, line 41: Do you mind naming it `caf_shell.c`?

subsys/caf/modules/Kconfig
```diff
@@ -17,3 +17,4 @@ rsource "Kconfig.power_manager"
 rsource "Kconfig.sensor_data_aggregator"
 rsource "Kconfig.sensor_manager"
 rsource "Kconfig.settings_loader"
+rsource "Kconfig.shell"
```
glarsennordic, line 20: Similarly, `Kconfig.caf_shell`

I'm not sure we're actually at risk of naming conflict, but it seems wise to just be explicit anyways

At minimum it'll make navigating to these source files easier

subsys/caf/modules/Kconfig.shell
```diff
@@ -0,0 +1,18 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menuconfig CAF_SHELL
+	bool "Shell module"
```
glarsennordic, line 8: Can you be more explicit here?

"Shell interface for triggering CAF events"

subsys/caf/modules/shell.c
```diff
@@ -0,0 +1,33 @@
...
+		return -EINVAL;
+	}
+
+	struct button_event *event = new_button_event();
+
```
glarsennordic, line 19: For completeness, can you raise an error complaining about not enough heap if *event is NULL?

glarsennordic (approved):
LGTM, only minor tweaks

glarsennordic:
Also please check the code owners for CAF to make sure we didn't miss them by accident since this code changed locations

subsys/caf/modules/shell.c
tony-le-24: will add some comments about those, thanks.

pdunaj (approved):
I think it's ok.
We actually wanted to add injecting events but there was never enough time to do it. We can start with this change.

I originally thought injecting could actually be done in type agnostic way. The arguments could be event name (used to extract type id from section) and raw event data (that would have to be formed by the caller based on type). Maybe we should add it...

subsys/caf/modules/shell.c
pdunaj: the default allocator of events checks that already

pdunaj:
@MarekPieta , can you have a look?

subsys/caf/modules/shell.c
```diff
@@ -0,0 +1,33 @@
...
+	event->key_id = (uint16_t)atoi(argv[1]);
+	event->pressed = *argv[2] != '0';
```
MarekPieta, line 21: Maybe we can define it simply as `y` or `n` and reject other values here (you could also ensure that parameter length is 1 here)? This seems more straightforward for booleans than using numbers.
btw. currently pressed would also be `false` e.g. for value of `01` here which may not be very intuitive.

@jayteemo what do you think about it ?

subsys/caf/modules/shell.c
```diff
@@ -0,0 +1,33 @@
...
+	return 0;
+}
+
+SHELL_STATIC_SUBCMD_SET_CREATE(sub_caf_shell_command,
+	SHELL_CMD_ARG(set_button, NULL, "Simulate a button event with button ID and state",
+				set_button_handler, 3, 0),
+	SHELL_SUBCMD_SET_END
+);
+
+SHELL_CMD_REGISTER(caf, &sub_caf_shell_command, "CAF Shell commands", NULL);
```
MarekPieta, line 33: Maybe we can call the set `caf_events` (`Submit a CAF event with user-defined parameters` as help). Then we could call each command simply by an event name (e.g. `button_event`).

subsys/caf/modules/shell.c
```diff
@@ -0,0 +1,33 @@
...
+
+	event->key_id = (uint16_t)atoi(argv[1]);
+	event->pressed = *argv[2] != '0';
+	APP_EVENT_SUBMIT(event);
+
+	return 0;
+}
+
+SHELL_STATIC_SUBCMD_SET_CREATE(sub_caf_shell_command,
+	SHELL_CMD_ARG(set_button, NULL, "Simulate a button event with button ID and state",
```
MarekPieta, line 28: `Submit a button_event with user-defined key ID and pressed state` ? To make it more in line with naming in CAF.

subsys/caf/modules/Kconfig.shell
```diff
@@ -0,0 +1,18 @@
...
+	depends on APP_EVENT_MANAGER
+	depends on SHELL
```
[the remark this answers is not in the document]
MarekPieta: Could we add Kconfig help to describe what the module can be used for? (e.g. what events can be handled currently)

MarekPieta (changes requested):

subsys/caf/modules/shell.c
jayteemo: `y` or `n` seems fine

glarsennordic: `y` or `n` and rejecting anything else seems reasonable to me as well

subsys/caf/modules/shell.c
glarsennordic: Can you also add some kind of help command / help messaging? Check out some of the other existing shells for examples of how to do this

subsys/caf/modules/shell.c
```diff
@@ -0,0 +1,33 @@
...
+static int set_button_handler(const struct shell *shell, size_t argc, char **argv)
+{
+	if (argc < 3) {
+		shell_error(shell, "Missing required parameter(s)");
+		return -EINVAL;
+	}
+
+	struct button_event *event = new_button_event();
+
+	event->key_id = (uint16_t)atoi(argv[1]);
```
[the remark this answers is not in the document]
tony-le-24: It's not intentional. I'll make the checks.

pdunaj (approved):

tony-le-24:
@peknis I added a new shell module for CAF. How should I change the docs? thanks

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -655,6 +655,12 @@ Common Application Framework (CAF)
 
   * Reduced verbosity of logs denoting allowed power states from ``info`` to ``debug``.
 
+* :ref:`caf_shell`:
+
+  * Added new module `caf_shell` as a shell interface for triggering CAF events.
```
peknis, line 660:
```suggestion
  * Added a new module ``caf_shell`` as a shell interface for triggering CAF events.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -655,6 +655,12 @@ Common Application Framework (CAF)
...
+  * Added :kconfig:option:`CONFIG_CAF_SHELL` for enabling the module.
```
peknis, line 661:
```suggestion
  * Added the :kconfig:option:`CONFIG_CAF_SHELL` Kconfig option for enabling the module.
```

peknis:
> @peknis I added a new shell module for CAF. How should I change the docs? thanks

I'd believe you need to write a piece of the new module in \ncs\nrf\doc\nrf\libraries\caf

peknis (changes requested):

tony-le-24:
> I'd believe you need to write a piece of the new module in \ncs\nrf\doc\nrf\libraries\caf

Thank you so much. I didn't know these exists. There are docs for other issues that I've been having too.

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,32 @@
...
+
+CAF: Shell module
+###################
+
+.. contents::
+   :local:
+   :depth: 2
+
+The shell module of the :ref:`lib_caf` (CAF) is responsible for executing defined shell commands to interact with other CAF modules and events.
+The module can handle a command for submitting CAF button events (press or release) at the moment and should be used along with the ::ref::`caf_buttons` module.
```
peknis, line 11:
```suggestion
The module handles commands for submitting CAF button events (press or release) and is supposed to be used with the :ref:`caf_buttons` module.
```

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,32 @@
...
+
+Configuration
+*************
+
+To use the module, you must enable the following Kconfig options:
+* :kconfig:option:`CONFIG_SHELL` - This is required since we need a shell for the module to receive commands.
```
peknis, line 17:
```suggestion
To use the module, enable the following Kconfig options:

* :kconfig:option:`CONFIG_SHELL` - This option enables a shell for the module to receive commands.
```

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,32 @@
...
+* :kconfig:option:`CONFIG_APP_EVENT_MANAGER` - This option enables the application event manager.
+
+Usage
+*****
+
+All shell commands in this module are subcommands of the ``caf_events`` command.
+To send a button event, you need to use the following syntax:
```
peknis, line 24:
```suggestion
To send a button event, use the following syntax:
```

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,32 @@
...
+
+  .. code-block:: console
+
+     caf_events set_button [buton_id] [state]
+
+The ``set_button`` subcommand has two required arguments:
+    * ``button_id`` - A single-digit numeric ID of the button.
+    * ``state`` - Button press state, can only be either ``y`` or ``n``.
```
peknis, line 32:
```suggestion
The ``set_button`` subcommand has two required arguments:

* ``button_id`` - A single-digit numeric ID of the button.
* ``state`` - Button press state, can only be either ``y`` or ``n``.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -686,6 +686,12 @@ Common Application Framework (CAF)
 
   * Reduced verbosity of logs denoting allowed power states from ``info`` to ``debug``.
 
+* :ref:`caf_shell`:
+
+  * Added a new module ``caf_shell`` as a shell interface for triggering CAF events.
+  * Added the :kconfig:option:`CONFIG_CAF_SHELL` Kconfig option for enabling the module.
+  * Added a shell command for submitting CAF events, and a subcommand for button event submission.
```
peknis, line 693:
```suggestion
* Added :ref:`caf_shell` for triggering CAF events.
```
As this is a totally new module, no need to say more at this point. Suggesting that you move this right under the Common Application Framework CAF heading.

peknis (changes requested):

peknis (approved):

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,34 @@
...
+The shell module of the :ref:`lib_caf` (CAF) is responsible for executing defined shell commands to interact with other CAF modules and events.
+The module handles commands for submitting CAF button events (press or release) and is supposed to be used with the :ref:`caf_buttons` module.
+
+Configuration
+*************
+
+To use the module, enable the following Kconfig options:
+
+* :kconfig:option:`CONFIG_SHELL` - This option enables a shell for the module to receive commands.
+* :kconfig:option:`CONFIG_APP_EVENT_MANAGER` - This option enables the application event manager.
```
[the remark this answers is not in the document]
MarekPieta: You can also mention that the module selects `CONFIG_CAF_BUTTON_EVENTS`

MarekPieta (changes requested):

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Usage
+*****
+
+All shell commands in this module are subcommands of the ``caf_events`` command.
+To send a button event, use the following syntax:
+
+  .. code-block:: console
+
+     caf_events set_button [buton_id] [state]
```
peknis, line 29:
```suggestion
     caf_events set_button [button_id] [state]
```

peknis (changes requested):

tony-le-24:
@MarekPieta Thank you so much for your detailed review. I'll update the docs today.

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,34 @@
...
+
+The ``set_button`` subcommand has two required arguments:
+
+* ``button_id`` - A single-digit numeric ID of the button.
```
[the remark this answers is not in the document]
tony-le-24: I think decimal would be easier. I'll make a comment about that.

subsys/caf/modules/caf_shell.c
```diff
@@ -0,0 +1,45 @@
...
+#include <caf/events/button_event.h>
+
+static const char set_button_cmd_help_str[] =
+	"Submit a button_event with user-defined key ID and pressed state\n"
+	"  Key ID           Single-digit numeric ID of the button\n"
+	"  Pressed state    'y' or 'n'\n";
+
+static int set_button_handler(const struct shell *shell, size_t argc, char **argv)
+{
+	if ((strlen(argv[1]) > 1) || !isdigit(*argv[1])) {
```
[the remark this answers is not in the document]
tony-le-24: I've been thinking about bigger key IDs, in case someone has a board with 20 LEDs. Using decimal would be easier to remember than hex (for me at least)

peknis (approved):

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,38 @@
+.. _caf_shell:
+
+CAF: Shell module
+###################
```
MarekPieta, line 4: Align underscore length

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,38 @@
...
+The module introduces a shell command for submitting events informing about button press or release (:c:struct:`button_event`).
+Submitting :c:struct:`button_event` related to the hardware button presses can be handled by the :ref:`caf_buttons`.
+
+Configuration
+*************
+
+To use the module, enable the following Kconfig options:
+
+* :kconfig:option:`CONFIG_CAF_SHELL` - This option is mandatory as it enables this module.
+* :kconfig:option:`CONFIG_SHELL` - This option enables a shell for the module to receive commands.
```
MarekPieta, line 20:
```suggestion
* :kconfig:option:`CONFIG_SHELL` - This option enables Zephyr shell that is used by the module to receive commands from user.
```

doc/nrf/libraries/caf/caf_shell.rst
```diff
@@ -0,0 +1,38 @@
...
+*****
+
+All shell commands of this module are subcommands of the :command:`caf_events` command.
+To send a :c:struct:`button_event`, use the following syntax:
+
+  .. code-block:: console
+
+     caf_events button_event [button_id] [pressed]
+
+The ``button_event`` subcommand has two required arguments:
```
MarekPieta, line 35:
```suggestion
The :command:`button_event` subcommand has two required arguments:
```

subsys/caf/modules/caf_shell.c
```diff
@@ -0,0 +1,49 @@
...
+	"  Key ID           Decimal numeric ID of the button\n"
+	"  Pressed state    'y' or 'n'\n";
+
+static int button_event_handler(const struct shell *shell, size_t argc, char **argv)
+{
+	/* The string part of button_id param, for invalidation check */
+	char *str_part;
+
+	/* Convert button_id from string to decimal number */
+	long button_id = strtol(argv[1], &str_part, 10);
```
MarekPieta, line 22: What do you think about using `0` instead of `10` as base to allow user to input values in hex too?

subsys/caf/modules/caf_shell.c
```diff
@@ -0,0 +1,49 @@
...
+
+	event->key_id = (uint16_t)button_id;
+	event->pressed = *argv[2] != 'n';
+	APP_EVENT_SUBMIT(event);
+
+	return 0;
+}
+
+SHELL_STATIC_SUBCMD_SET_CREATE(sub_caf_shell_command,
+	SHELL_CMD_ARG(button_event, NULL, set_button_cmd_help_str, button_event_handler, 3, 0),
```
MarekPieta, line 44: `set_button_cmd_help_str` -> `button_event_cmd_help_str` ?

subsys/caf/modules/caf_shell.c
MarekPieta: btw. Should we check if the `button_id` actually fits in `uint16_t` (lower than `UINT16_MAX`) and is not negative?

MarekPieta (changes requested):

MarekPieta (approved):

PR #11930, opened 2023-08-02
applications: Matter Bridge: architectural refactoring

markaj-nordic:
Major scope:

  - allow to inject Matter and non-Matter devices to be bridged
    into the BridgeManager object
  - added devices factory to support bridged devices creation
  - implemented a dedicated map facilitating device object life
    time management
  - added necessary BridgeManager modifications
  
Also:

  - Added missing virtual constructor in abstract
    BridedDeviceDataProvider class.
  - stop the timer in data provider when object is being destructed
  - added more logs

### DIFF
--- a/applications/matter_bridge/src/bridge_shell.cpp
+++ b/applications/matter_bridge/src/bridge_shell.cpp
@@ -4,10 +4,11 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 
+#include "bridge_manager.h"
+#include "bridged_device_factory.h"
+
 #include <zephyr/logging/log.h>
 #include <zephyr/shell/shell.h>
-
-#include "bridge_manager.h"
 
 LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
 
@@ -20,8 +21,19 @@ static int add_bridged_device(const struct shell *shell, size_t argc, char **arg
 		nodeLabel = argv[2];
 	}
 
-	CHIP_ERROR err =
-		GetBridgeManager().AddBridgedDevice(static_cast<BridgedDevice::DeviceType>(deviceType), nodeLabel);
+	auto *newBridgedDevice = BridgeFactory::GetBridgedDeviceFactory().Create(
+		static_cast<BridgedDevice::DeviceType>(deviceType), nodeLabel);
+
+	VerifyOrReturnError(newBridgedDevice != nullptr, -EPERM,
+			    shell_fprintf(shell, SHELL_INFO, "Cannot allocate Matter device of given type\n"));
+
+	auto *newDataProvider = BridgeFactory::GetDataProviderFactory().Create(
+		static_cast<BridgedDevice::DeviceType>(deviceType), BridgeManager::HandleUpdate);
+
+	VerifyOrReturnError(newDataProvider != nullptr, -EPERM, delete newBridgedDevice,
+			    shell_fprintf(shell, SHELL_INFO, "Cannot allocate data provider of given type\n"));
+
+	CHIP_ERROR err = GetBridgeManager().AddBridgedDevices(newBridgedDevice, newDataProvider);
 	if (err == CHIP_NO_ERROR) {
 		shell_fprintf(shell, SHELL_INFO, "Done\n");
 	} else if (err == CHIP_ERROR_INVALID_STRING_LENGTH) {
--- a/samples/matter/common/src/bridge/bridge_manager.cpp
+++ b/samples/matter/common/src/bridge/bridge_manager.cpp
@@ -6,21 +6,6 @@
 
 #include "bridge_manager.h"
 
-#if CONFIG_BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE
-#include "onoff_light.h"
-#include "onoff_light_data_provider.h"
-#endif
-
-#if CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE
-#include "temperature_sensor.h"
-#include "temperature_sensor_data_provider.h"
-#endif
-
-#if CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE
-#include "humidity_sensor.h"
-#include "humidity_sensor_data_provider.h"
-#endif
-
 #include <app-common/zap-generated/ids/Clusters.h>
 #include <app/reporting/reporting.h>
 #include <app/util/generic-callbacks.h>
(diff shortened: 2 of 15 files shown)


kkasperczyk-no (approved):

samples/matter/common/src/bridge/util/bridge_util.h
```diff
@@ -0,0 +1,124 @@
...
+   only be constructed by passing a user-defined initialized list with
+   <BridgedDevice::DeviceType, ConcreteDeviceCreator> pairs.
+   Then, Create() method can be used to obtain an instance of demanded
+   device type with all passed arguments forwarded to the underlying
+   ConcreteDeviceCreator.
+*/
+template <typename T, typename... Args> class DeviceFactory {
+public:
+	using ConcreteDeviceCreator = std::function<T *(Args...)>;
+	using CreationMap = std::map<BridgedDevice::DeviceType, ConcreteDeviceCreator>;
```
Damian-Nordic, line 28: Can we avoid using costly structures for containers that will likely have just a couple of elements? Probably a static array of key-value pairs would be good enough and required less memory.

Also, the current design introduces a reverse dependency - Bridge Manager includes a file from the bridge application. If you want to have a generic bridge module, why not to introduce a factory interface such as:
```
template <class T, class... Args>
class GenericFactory
{
    virtual T* Create(Args&&... args) = 0;
}

using GenericDeviceFactory = GenericFactory<BridgedDevice, BridgedDevice::DeviceType, const char* /* label */>;
...
```
and then an application would create a concrete implementation of the Create method that could be a simple switch-case (without the need for the costly map) and pass that implementation to the bridge manager upon initialization. Otherwise, it may be difficult to use the common bridge module if one must provide specific header files etc. in the application.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -165,48 +129,48 @@ CHIP_ERROR BridgeManager::AddDevice(BridgedDevice *device)
...
+BridgeManager::GetBridgedDevice(uint16_t index, const EmberAfAttributeMetadata *attributeMetadata, uint8_t *buffer)
 {
-	if (endpoint >= kMaxBridgedDevices) {
+	if (index >= kMaxBridgedDevices) {
 		return nullptr;
 	}
 
-	BridgedDevice *device = sBridgeManager.mBridgedDevices[endpoint];
-	if ((device == nullptr) || !buffer || !attributeMetadata) {
+	DevicePtr *device = const_cast<DevicePtr *>(&(sBridgeManager.mDevicesMap[index]->mDevice));
```
Damian-Nordic, line 139:
```suggestion
	BridgedDevice *device = sBridgeManager.mDevicesMap[index]->mDevice->get();
```
Otherwise, I cannot see how `device` could be null. Also, where is it checked that `mDevicesMap` contains `index`?

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -165,48 +129,48 @@ CHIP_ERROR BridgeManager::AddDevice(BridgedDevice *device)
...
 {
-	BridgedDevice *device = GetBridgedDevice(endpoint, attributeMetadata, buffer);
+	DevicePtr *device = GetBridgedDevice(index, attributeMetadata, buffer);
 
 	if (!device) {
 		return CHIP_ERROR_INVALID_ARGUMENT;
 	}
 
-	return device->HandleRead(clusterId, attributeMetadata->attributeId, buffer, maxReadLength);
+	return (*device)->HandleRead(clusterId, attributeMetadata->attributeId, buffer, maxReadLength);
```
Damian-Nordic, line 156: Should `GetBridgedDevice` return `BridgedDevice *` to avoid double pointer dereference?

Damian-Nordic (commented):
A few possible issues found but most comments are style suggestions

samples/matter/common/src/bridge/util/bridge_util.h
Damian-Nordic: We discussed f2f that there's a task to rework this to avoid using STL so feel free to ignore this comment for now.

samples/matter/common/src/bridge/bridge_manager.h
```diff
@@ -6,41 +6,62 @@
...
+#include "bridge_util.h"
 #include "bridged_device.h"
 #include "bridged_device_data_provider.h"
-#include <map>
 
 class BridgeManager {
 public:
 	void Init();
-	CHIP_ERROR AddBridgedDevice(BridgedDevice::DeviceType bridgedDeviceType, const char *nodeLabel);
+	CHIP_ERROR AddBridgedDevices(BridgedDevice *aDevice, BridgedDeviceDataProvider *aDataProvider);
```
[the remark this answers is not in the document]
Damian-Nordic: OK, it still represents a single physical device, so that's why it confused me, but I guess the current name is fine, but maybe in the future you could add some doxygen comment about that :).

samples/matter/common/src/bridge/bridge_shell.cpp
```diff
@@ -20,8 +21,17 @@ static int add_bridged_device(const struct shell *shell, size_t argc, char **arg
...
 
-	CHIP_ERROR err =
-		GetBridgeManager().AddBridgedDevice(static_cast<BridgedDevice::DeviceType>(deviceType), nodeLabel);
+	auto *newBridgedDevice = BridgeFactory::GetBridgedDeviceFactory().Create(
+		static_cast<BridgedDevice::DeviceType>(deviceType), nodeLabel);
+	auto *newDataProvider = BridgeFactory::GetDataProviderFactory().Create(
+		static_cast<BridgedDevice::DeviceType>(deviceType), BridgeManager::HandleUpdate);
+
+	if (!newBridgedDevice || !newDataProvider) {
+		shell_fprintf(shell, SHELL_INFO, "Cannot allocate device of given type\n");
```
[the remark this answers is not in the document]
markaj-nordic: That's right. I will either check these pointers separately or rework the factory as you suggested.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -48,57 +33,15 @@ void BridgeManager::Init()
...
 }
 
-CHIP_ERROR BridgeManager::AddBridgedDevice(BridgedDevice::DeviceType bridgedDeviceType, const char *nodeLabel)
+CHIP_ERROR BridgeManager::AddBridgedDevices(BridgedDevice *aDevice, BridgedDeviceDataProvider *aDataProvider)
 {
-	Platform::UniquePtr<BridgedDevice> device(nullptr);
-	Platform::UniquePtr<BridgedDeviceDataProvider> dataProvider(nullptr);
-
+	const char *nodeLabel = aDevice->GetNodeLabel();
 	if (nodeLabel && (strlen(nodeLabel) >= BridgedDevice::kNodeLabelSize)) {
```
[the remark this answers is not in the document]
markaj-nordic: Yes, I guess it should be moved to the BridgedDeviceFactory (before we allocate the object).

samples/matter/common/src/bridge/bridge_manager.cpp
markaj-nordic: Yep, originally I wanted to use a reference but then though we needed to handle the case when the provided arguments were invalid (buffer and attributeMetadata), but these are not used currently, and maybe won't be. I guess with `Contains()` taking care of the valid index we can drop the GetBridgedDevice() method and use the map directly to avoid a mess with uniqe pointers.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -108,46 +51,67 @@ CHIP_ERROR BridgeManager::RemoveBridgedDevice(uint16_t endpoint)
...
+CHIP_ERROR BridgeManager::AddDevices(BridgedDevice *aDevice, BridgedDeviceDataProvider *aDataProvider)
 {
 	uint8_t index = 0;
+	Platform::UniquePtr<BridgedDevice> device(aDevice);
+	Platform::UniquePtr<BridgedDeviceDataProvider> provider(aDataProvider);
+
+	/* Maximum number of Matter bridged devices is controlled inside mDevicesMap,
+	   but the data providers may be created independently, so let's ensure we do not
+	   violate the maximum number of supported instances. */
+	VerifyOrReturnError(++mNumberOfProviders <= kMaxDataProviders, CHIP_ERROR_INTERNAL);
```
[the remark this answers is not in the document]
markaj-nordic: You mean the if the` mDevicesMap` is full? I agree, we should check that in the first place.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -108,46 +51,67 @@ CHIP_ERROR BridgeManager::RemoveBridgedDevice(uint16_t endpoint)
...
+			mDevicesMap.Insert(index, DevicePair(std::move(device), std::move(provider)));
 			EmberAfStatus ret;
 			while (true) {
+				if (!mDevicesMap[index]) {
+					LOG_ERR("Cannot retrieve bridged device from index %d", index);
+					return CHIP_ERROR_INTERNAL;
+				}
+				auto &storedDevice = mDevicesMap[index]->mDevice;
+				if (!storedDevice) {
+					LOG_ERR("Bridged device stored under index %d is empty", index);
```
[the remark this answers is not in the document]
markaj-nordic: In theory something might have gone wrong when inserting into a map, but I guess can safely assume it should depend only in the input.

samples/matter/common/src/bridge/bridge_manager.cpp
markaj-nordic: I'll just remove the `GetBridgedDevice` for the sake of direct access to the map with []operator.

samples/matter/common/src/bridge/util/bridge_util.h
markaj-nordic: For now I will move the shell module to the `applications/matter_bridge` as it's suppose to be a part of the application.

ArekBalysNordic (approved):

samples/matter/common/src/bridge/bridge_manager.cpp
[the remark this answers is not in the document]
Damian-Nordic: I mean that if you increase mNumberOfProviders before checking if it's bigger than kMaxDataProviders, it may grow beyond the maximum (so then even if you release on device, the manager will still think it's full).

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -20,8 +21,19 @@ static int add_bridged_device(const struct shell *shell, size_t argc, char **arg
...
+	auto *newBridgedDevice = BridgeFactory::GetBridgedDeviceFactory().Create(
+		static_cast<BridgedDevice::DeviceType>(deviceType), nodeLabel);
+
+	VerifyOrReturnError(newBridgedDevice != nullptr, -EPERM,
+			    shell_fprintf(shell, SHELL_INFO, "Cannot allocate Matter device of given type\n"));
+
+	auto *newDataProvider = BridgeFactory::GetDataProviderFactory().Create(
+		static_cast<BridgedDevice::DeviceType>(deviceType), BridgeManager::HandleUpdate);
+
+	VerifyOrReturnError(newDataProvider != nullptr, -EPERM,
```
Damian-Nordic, line 33: My point was that `newBridgedDevice` should be released in the case of error here.

Damian-Nordic (approved):

samples/matter/common/src/bridge/bridge_manager.cpp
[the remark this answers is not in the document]
markaj-nordic: Fixed.

applications/matter_bridge/src/bridge_shell.cpp
markaj-nordic: Fixed.

markaj-nordic:
Also added an EXPERIMENTAL flag.

PR #11953, opened 2023-08-04
Tx support outside twt sp

ajayparida:
Support to extend TWT service period

### DIFF
--- a/drivers/wifi/nrf700x/Kconfig
+++ b/drivers/wifi/nrf700x/Kconfig
@@ -279,6 +279,22 @@ config NRF700X_RPU_PS_IDLE_TIMEOUT_MS
 	int "RPU power save idle timeout in milliseconds"
 	default 10
 
+config NRF700X_RPU_EXTEND_TWT_SP
+	bool "Enable extending TWT service period"
+	help
+	  In case frames accepted before beginning of SP are not
+	  transmitted before the SP completes then typically they are
+	  dropped to conform to SP window as per specification i.e., no
+	  transmission outside SP window.
+
+	  This feature mitigates the frame loss by transmitting even after SP
+	  completion by using standard contention mechanism which is allowed
+	  in specification but not recommended. As the device is actively transmitting
+	  beyond SP, the power consumption increases depending on the amount
+	  of traffic available at the start of SP.
+
+	  Please note that if a frame is sent after SP starts it will be queued and this
+	  mechanism is not used.
 endif
 
 config WIFI_FIXED_MAC_ADDRESS
--- a/drivers/wifi/nrf700x/osal/fw_if/umac_if/src/cmd.c
+++ b/drivers/wifi/nrf700x/osal/fw_if/umac_if/src/cmd.c
@@ -180,7 +180,11 @@ enum wifi_nrf_status umac_cmd_init(struct wifi_nrf_fmac_dev_ctx *fmac_dev_ctx,
 					CONFIG_NRF700X_REG_DOMAIN,
 					NRF_WIFI_COUNTRY_CODE_LEN);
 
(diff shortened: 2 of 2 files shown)


drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -279,6 +279,11 @@ config NRF700X_RPU_PS_IDLE_TIMEOUT_MS
 	int "RPU power save idle timeout in milliseconds"
 	default 10
 
+config NRF700X_RPU_EXTEND_TWT_SP
+	bool "Enable extending TWT service period"
+	help
+	  Enabling this feature extends TWT service period to send frames.
+	  It uses EDCA to avoid dropping of frames, so consumes more power.
```
krish2718, line 286: This needs more clarification to convey that `new frames` aren't send outside SP, only accepted frames before the SP are sent outside SP.

krish2718 (changes requested):
The commit order is wrong: Dependent commits should be first (oldest).

drivers/wifi/nrf700x/Kconfig.nrf700x
ajayparida: Updated help. Please check

drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -279,6 +279,13 @@ config NRF700X_RPU_PS_IDLE_TIMEOUT_MS
 	int "RPU power save idle timeout in milliseconds"
 	default 10
 
+config NRF700X_RPU_EXTEND_TWT_SP
+	bool "Enable extending TWT service period"
+	help
+	  Enabling this feature extends TWT service period.
+	  Only accepted frames during SP will be send.
```
rado17, line 286: Should be "will be sent."

drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -279,6 +279,13 @@ config NRF700X_RPU_PS_IDLE_TIMEOUT_MS
...
+	  New frames outside SP will not be send out.
```
rado17, line 287: Should be "will not be sent out."

rado17 (changes requested):

drivers/wifi/nrf700x/Kconfig.nrf700x
```diff
@@ -279,6 +279,13 @@ config NRF700X_RPU_PS_IDLE_TIMEOUT_MS
...
+	  It uses EDCA to avoid dropping of frames, so consumes more power.
```
krish2718, line 288:
```suggestion
     In case frames accepted before beginning of SP are not
     transmitted before the SP completes then typically they are
     dropped to conform to SP window as per specification i.e., no
     transmission outside SP window.
     
     This feature mitigates the frame loss by transmitting even after SP
     completion by using standard contention mechanism which is allowed
     in specification but not recommended. As the device is actively transmitting
     beyond SP, the power consumption increases depending on the amount
     of traffic available at the start of SP. 
     
     Please note that if a frame is sent after SP starts it will be queued and this 
     mechanism is not used.
```

drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fw/host_rpu_sys_if.h
```diff
@@ -780,12 +780,25 @@ enum op_band {
 	BAND_24G
 };
 
+#define TWT_EXTEND_SP_EDCA  0x1
```
krish2718, line 783:
```suggestion
#define TWT_TX_BEYOND_SP 0x1
```

No need to put the actual mechanism here.

krish2718 (changes requested):

rado17 (approved):

krish2718 (approved):

PR #11960, opened 2023-08-04
doc: security: Don't document how to flash a TF-M

SebastianBoe:
Don't document how to flash TF-M as it is the same as for any NCS build.

This documentation is expensive to maintain as it is referencing development environment specifics that are unrelated to TF-M and subject to change.


mswarowsky (approved):
Fine for me to remove that part

mia-ko (approved):

PR #11963, opened 2023-08-04
Bluetooth: Mesh: remove division by zero in ctl

alxelax:
If ctl range min is equal range max then Generic level is 0.

### DIFF
--- a/subsys/bluetooth/mesh/light_ctl_internal.h
+++ b/subsys/bluetooth/mesh/light_ctl_internal.h
@@ -30,9 +30,10 @@ static inline int16_t temp_to_lvl(struct bt_mesh_light_temp_srv *srv,
 {
 	uint16_t temp = CLAMP(raw_temp, srv->range.min, srv->range.max);
 
-	return ROUNDED_DIV((temp - srv->range.min) * UINT16_MAX,
-				(srv->range.max - srv->range.min)) -
-				LIGHT_TEMP_LVL_OFFSET;
+	return srv->range.max == srv->range.min ? 0
+						: ROUNDED_DIV((temp - srv->range.min) * UINT16_MAX,
+							      (srv->range.max - srv->range.min)) -
+							  LIGHT_TEMP_LVL_OFFSET;
 }
 
 void bt_mesh_light_temp_srv_set(struct bt_mesh_light_temp_srv *srv,


Balaklaka (approved):

subsys/bluetooth/mesh/light_ctl_internal.h
```diff
@@ -30,9 +30,10 @@ static inline int16_t temp_to_lvl(struct bt_mesh_light_temp_srv *srv,
 {
 	uint16_t temp = CLAMP(raw_temp, srv->range.min, srv->range.max);
 
-	return ROUNDED_DIV((temp - srv->range.min) * UINT16_MAX,
-				(srv->range.max - srv->range.min)) -
-				LIGHT_TEMP_LVL_OFFSET;
+	return srv->range.max == srv->range.min ? 0
+						: ROUNDED_DIV((temp - srv->range.min) * UINT16_MAX,
+							      (srv->range.max - srv->range.min)) -
+							  LIGHT_TEMP_LVL_OFFSET;
```
ludvigsj, line 36: Not really a fan of ternary expressions across four lines. Suggest this as more readable:

```suggestion
	if (srv->range.max == srv->range.min) {
		return 0;
	}

	return ROUNDED_DIV((temp - srv->range.min) * UINT16_MAX,
				(srv->range.max - srv->range.min)) -
				LIGHT_TEMP_LVL_OFFSET;
```

ludvigsj (changes requested):

subsys/bluetooth/mesh/light_ctl_internal.h
ludvigsj: Also, shouldnt the entire divisor be checked for zero, not just the `max - min` part? As in include the `LIGHT_TMP_LVL_OFFSET` when checking if the divisor is zero? Or am I missing something?

alxelax: I do not see how this is more readable since almost all ternary is located on a single line and the rest three lines are the same macro. Btw. this is how built-in ncs-sdk rules autoformatted. 

`LIGHT_TMP_LVL_OFFSET` is not part of the division macro.

ludvigsj: Oh, sorry, misread the parenthesis. I still think the `if` is more readable but I'm not going to stall the PR for this. Will approve

ludvigsj (approved):

PR #11964, opened 2023-08-04
applications: nrf_desktop: Fix returning error in MCUmgr callbacks

MarekPieta:
Change fixes returning error in MCUmgr callbacks. This prevents assertion failures when the error is returned by the application module.

Jira: NCSDK-22964

### DIFF
--- a/applications/nrf_desktop/src/modules/dfu_mcumgr.c
+++ b/applications/nrf_desktop/src/modules/dfu_mcumgr.c
@@ -69,7 +69,8 @@ static enum mgmt_cb_return mcumgr_img_mgmt_cb(uint32_t event,
 	k_work_reschedule(&dfu_timeout, DFU_TIMEOUT);
 	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) && dfu_lock_claim(&mcumgr_owner)) {
 		(void)k_work_cancel_delayable(&dfu_timeout);
-		return MGMT_ERR_EACCESSDENIED;
+		*rc = MGMT_ERR_EACCESSDENIED;
+		return MGMT_CB_ERROR_RC;
 	}
 
 	if (IS_ENABLED(CONFIG_MCUMGR_TRANSPORT_BT) &&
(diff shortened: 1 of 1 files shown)


kapi-no (approved):
is this regression present in NCS 2.4.0? Should we document it in the changelog?

mkapala-nordic (approved):

MarekPieta:
> is this regression present in NCS 2.4.0? Should we document it in the changelog?

No, the MCUmgr API was changed after NCS 2.4.0. Then I think no doc update is needed here (the module is even marked as experimental currently).

applications/nrf_desktop/src/modules/dfu_mcumgr.c
```diff
@@ -69,7 +69,8 @@ static enum mgmt_cb_return mcumgr_img_mgmt_cb(uint32_t event,
 	k_work_reschedule(&dfu_timeout, DFU_TIMEOUT);
 	if (IS_ENABLED(CONFIG_DESKTOP_DFU_LOCK) && dfu_lock_claim(&mcumgr_owner)) {
 		(void)k_work_cancel_delayable(&dfu_timeout);
-		return MGMT_ERR_EACCESSDENIED;
+		*rc = MGMT_ERR_EACCESSDENIED;
```
nordicjm, line 72: Not so sure this error really matches the use case, I mean theoretically it is access denied but the real reason is that the transfer is being done on another transport

MarekPieta: Help of this error code states that: `Access to specific function, command or resource denied`
I was using this error code to indicate that access to the secondary image slot is denied (application denies access to the secondary image slot, because another transport uses it). I preferred to avoid using other error code here to distinguish from errors returned by MCUmgr itself.

MarekPieta: Can we keep it, or could you suggest an error code that is more suitable here?

nordicjm: The problem is that this is an SMP error code, not a group error code. A new code should be added to https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/mgmt/mcumgr/grp/img_mgmt/img_mgmt.h which (when using SMP version 2) will give the client an exact error, for the legacy version then it needs to do a translation to an SMP error code and this might be acceptable. We are moving away from SMP version 1

MarekPieta: I see. Can we do it as follow-up action (I made a new Jira ticket for this one)? We need to introduce new error code to Zephyr and then cherry-pick it - that would delay fix on our end. Also, my current change does not introduce a regression - it keeps the return code as it was.

nordicjm: Yes, I will add a new code to zephyr now and next upmerge this code can be switched over to use it

MarekPieta: Thanks !

PR #11968, opened 2023-08-07
Psa ps

mswarowsky:
test-sdk-nrf: test-psa-ps


samples/crypto/persistent_key_usage/src/main.c
```diff
@@ -15,20 +15,91 @@
...
+#include <nrf_cc3xx_platform.h>
+#include <hw_unique_key.h>
+
+#ifndef HUK_HAS_KMU
+#include <zephyr/sys/reboot.h>
+#endif
+
+static int write_huk(void)
+{
+	int result = nrf_cc3xx_platform_init();
```
joerchan, line 76: nrf_cc3xx_platform_init should be handled by HW_CC3XX option, which also handles the case where entropy has not been enabled, i.e calling another init function.

subsys/secure_storage/src/aead/secure_backend_aead.c
```diff
@@ -0,0 +1,326 @@
...
+
+cleanup_objects:
+	/* Remove all object if an error occurs */
+	storage_remove_object(uid, prefix, SECURE_STORAGE_AEAD_BACKEND_FILENAME_SUFFIX_NONCE);
+	storage_remove_object(uid, prefix, SECURE_STORAGE_AEAD_BACKEND_FILENAME_SUFFIX_DATA);
+	storage_remove_object(uid, prefix, SECURE_STORAGE_AEAD_BACKEND_FILENAME_SUFFIX_FLAGS);
+
+cleanup:
+	memset(&additional_data, 0, sizeof(additional_data));
+	memset(nonce, 0, sizeof(nonce));
```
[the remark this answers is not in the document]
mswarowsky: good point, replaces all memset with a safe_memset

Vge0rge (approved):

subsys/secure_storage/src/aead/aead_ctr_nonce.c
```diff
@@ -0,0 +1,84 @@
...
+	/* Incrementing is implemented by using 2 64bit numbers */
+	g_nonce.low++;
+
+	/* On overflow */
+	if (g_nonce.low == 0) {
+		++g_nonce.high;
+	}
+
+	/* Copy low first */
+	memcpy(nonce, &g_nonce.low, MIN(nonce_len, sizeof(g_nonce.low)));
```
vili-nordic, line 55: the sizeof can be changed to sizeof(g_nonce) and we can skip the if under

subsys/secure_storage/src/aead/secure_backend_aead.c
```diff
@@ -0,0 +1,338 @@
...
+	additional_data.flags = data_flags;
+	additional_data.size = data_length;
+
+	status = secure_storage_aead_decrypt(key_buf, AEAD_KEY_SIZE, nonce, AEAD_NONCE_SIZE,
+					     (void *)&additional_data, sizeof(additional_data),
+					     aead_buf, object_data_size, data_buf,
+					     SECURE_STORAGE_MAX_ASSET_SIZE, &aead_out_size);
+
+	/* Clean up */
+	safe_memset(key_buf, sizeof(key_buf), 0, AEAD_KEY_SIZE);
```
vili-nordic, line 165: idk. there are many returns from this functions that will not result in this cleanup.

subsys/secure_storage/src/storage_backend_settings.c
```diff
@@ -0,0 +1,176 @@
...
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/settings/settings.h>
+
+#include "storage_backend.h"
+
+LOG_MODULE_REGISTER(internal_secure_storage_settings, CONFIG_SECURE_STORAGE_LOG_LEVEL);
+
+/* UID as uint64_t in hexadecimal representation length */
+#define SECURE_STORAGE_SETTINGS_BACKEND_FILENAME_UID_LENGTH (sizeof(uint64_t) * 2)
```
vili-nordic, line 20: unused define?

subsys/secure_storage/src/storage_backend_settings.c
```diff
@@ -0,0 +1,176 @@
...
+
+/* Storage pattern: prefix, uid low, uid high, suffix */
+#define SECURE_STORAGE_SETTINGS_BACKEND_FILENAME_PATTERN "%s%08x%08x%s"
+
+/* Max filename length aligned on Settings File backend max length */
```
vili-nordic, line 25: aligned *with*

vili-nordic (approved):

subsys/secure_storage/src/aead/secure_backend_aead.c
mswarowsky: True moved it to one clean up block at the end

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+
+config SECURE_STORAGE_AEAD_BACKEND_MAX_DATA_SIZE
+	int "AEAD backend maximum storage size"
+	depends on SECURE_STORAGE_AEAD_BACKEND
+	default 256
+	help
+	  This defines the maximum data size that can be stored.
+
+choice SECURE_STORAGE_AEAD_AEAD_CRYPTO_BACKEND
+	prompt "AEAD crypto backend"
```
joerchan, line 90: AEAD algorithm

This makes more sense to me,

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+if SECURE_STORAGE_AEAD_BACKEND
+
+config SECURE_STORAGE_AEAD_BACKEND_MAX_DATA_SIZE
+	int "AEAD backend maximum storage size"
+	depends on SECURE_STORAGE_AEAD_BACKEND
+	default 256
+	help
+	  This defines the maximum data size that can be stored.
+
+choice SECURE_STORAGE_AEAD_AEAD_CRYPTO_BACKEND
```
joerchan, line 89: AEAD is twice here.
~~Suggest using ALG instead of BACKEND.~~
BACKEND should be before AEAD (see above)

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+
+choice SECURE_STORAGE_BACKEND
+	prompt "Trust backend"
+	depends on SECURE_STORAGE
+	default SECURE_STORAGE_AEAD_BACKEND
+	help
+	  The backend that handles encryption, authentication or other means of
+	  trust validation of the stored assets.
+
+menuconfig SECURE_STORAGE_AEAD_BACKEND
```
joerchan, line 74: strange prefix order here, it seems there is a backend for AEAD, but AEAD is the backend.

I suggest SECURE_STORAGE_BACKEND_AEAD instead.
Same order for the rest of the options.

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+	int "AEAD backend maximum storage size"
+	depends on SECURE_STORAGE_AEAD_BACKEND
+	default 256
+	help
+	  This defines the maximum data size that can be stored.
+
+choice SECURE_STORAGE_AEAD_AEAD_CRYPTO_BACKEND
+	prompt "AEAD crypto backend"
+	depends on SECURE_STORAGE_AEAD_BACKEND
+	default SECURE_STORAGE_AEAD_BACKEND_CRYPTO_PSA_CHACHAPOLY
```
joerchan, line 92: In PSA PS with TF-M we by default use either CCM or GCM.

It seems odd to go for a third one here. Is there a specific reason to use chacha-poly here?

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+	depends on PSA_PROTECTED_STORAGE
+	default "ps/"
+	help
+	  This defines the prefix for Protected Storage object names.
+
+endif # PSA_PROTECTED_STORAGE
+
+menuconfig PSA_INTERNAL_TRUSTED_STORAGE
+	bool "Internal Trusted Storage"
+	depends on SECURE_STORAGE
```
joerchan, line 47: Redundant, you're under "if SECURE_STORAGE" already.

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+	string "Internal Trusted Storage object prefix"
+	depends on PSA_INTERNAL_TRUSTED_STORAGE
+	default "its/"
+	help
+	  This defines the prefix for Internal Trusted Storage object names.
+
+endif # PSA_INTERNAL_TRUSTED_STORAGE
+
+choice SECURE_STORAGE_BACKEND
+	prompt "Trust backend"
```
joerchan, line 67: Very short description.

Maybe "Level of trust"

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+	bool "PSA ChaChaPoly"
+	select PSA_WANT_ALG_CHACHA20_POLY1305
+	help
+	  Use PSA Crypto API's with the ChaChaPoly-1305 AEAD algorithm.
+
+endchoice # SECURE_STORAGE_AEAD_AEAD_CRYPTO_BACKEND
+
+choice SECURE_STORAGE_AEAD_BACKEND_NONCE
+	prompt "AEAD nonce implementation"
+	depends on SECURE_STORAGE_AEAD_BACKEND
```
joerchan, line 107: Redundant depends on, your already under "if SECURE_STORAGE_AEAD_BACKEND"

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+endchoice # SECURE_STORAGE_BACKEND
+
+choice SECURE_STORAGE_STORAGE_BACKEND
+	prompt "Storage backend"
+	depends on SECURE_STORAGE
+	default SECURE_STORAGE_STORAGE_SETTINGS_BACKEND
+	help
+	  Implementation used to store the assets
+
+config SECURE_STORAGE_STORAGE_SETTINGS_BACKEND
```
joerchan, line 166: SECURE_STORAGE_STORAGE_BACKEND_SETTINGS

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,175 @@
...
+	bool "Settings for storage"
```
joerchan, line 167: Settings storage backend

subsys/secure_storage/src/aead/aead_key_hash.c
```diff
@@ -0,0 +1,23 @@
...
+psa_status_t secure_storage_get_key(psa_storage_uid_t uid, uint8_t *key_buf, size_t key_length)
+{
+
+	size_t olen;
+
+	if (key_length < AEAD_KEY_SIZE) {
+		return PSA_ERROR_BUFFER_TOO_SMALL;
+	}
+
+	return psa_hash_compute(PSA_ALG_SHA_256, (uint8_t *)&uid, sizeof(uid), key_buf, key_length,
```
joerchan, line 21: It seems odd to me to get the key this way. Isn't HKDF key derivation better?

subsys/secure_storage/src/aead/aead_key_huk.c
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+#include <hw_unique_key.h>
+#include <nrf_cc3xx_platform.h>
+#include <zephyr/init.h>
+
+#include "aead_key.h"
+
+#ifdef HUK_HAS_KMU
+#define KEYSLOT HUK_KEYSLOT_MKEK
```
joerchan, line 14: Is MKEK or MEXT the right one here? MEXT says external storage, which seems like what PS is for.

subsys/secure_storage/src/aead/secure_backend_aead.c
```diff
@@ -0,0 +1,338 @@
...
+	psa_storage_uid_t uid;
+	psa_storage_create_flags_t flags;
+	size_t size;
+};
+
+/* Temporary AEAD encryption/decryption buffers */
+static uint8_t aead_buf[AEAD_MAX_BUF_SIZE];
+static uint8_t data_buf[SECURE_STORAGE_MAX_ASSET_SIZE];
+
+static void safe_memset(void *dest, const size_t dest_size, const uint8_t ch, const size_t n)
```
joerchan, line 54: I think we should use mbedtls_platform_zeroise instead

subsys/net/openthread/Kconfig
```diff
@@ -36,7 +36,9 @@ config OPENTHREAD_NRF_SECURITY_PSA
 	select MBEDTLS_ENABLE_HEAP
 	select MBEDTLS_TLS_LIBRARY
 	select MBEDTLS_PSA_CRYPTO_STORAGE_C if !BUILD_WITH_TFM
-	select PSA_NATIVE_ITS if !BUILD_WITH_TFM
+	select SECURE_STORAGE if !BUILD_WITH_TFM
+	select SETTINGS if !BUILD_WITH_TFM
```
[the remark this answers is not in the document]
mswarowsky: `SECURE_STORAGE_STORAGE_SETTINGS_BACKEND` is default selected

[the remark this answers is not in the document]
joerchan: If your default changes, then this gets invalid. It's not about if it is correct now, but how it will act when changes are made. Relying on defaults in different parts of the system creates something that is hard to maintain.

If SECURE_STORAGE_STORAGE_SETTINGS_BACKEND selected SETTINGS it would be different.

Now it is selecting an implicit dependency, and it is hard to see why.

samples/crypto/persistent_key_usage/boards/nrf52840dk_nrf52840.conf
```diff
@@ -2,13 +2,26 @@
 CONFIG_PSA_CRYPTO_DRIVER_OBERON=n
 CONFIG_PSA_CRYPTO_DRIVER_CC3XX=y
 
-# Note that NVS is only one of the options to enable persistent storage of keys
-# using the settings subsystem.
+# Using the Internal Trused Storage with the settings subsystem with the
```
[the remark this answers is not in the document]
mswarowsky: All the CONFIGS I took from the other samples t make the settings with littlefs to work,
By testing this, seems like th QPSI is not need so removed it and removed the empty lines below to making it more obvious that the comment applies to the whole block

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,189 @@
...
+	  be placed inside internal flash. Some examples of assets that require
+	  this are replay protection values for external storage and keys for
+	  use by components of the PSA Root of Trust.
+
+if PSA_INTERNAL_TRUSTED_STORAGE
+
+config PSA_INTERNAL_TRUSTED_STORAGE_PREFIX
+	string "Internal Trusted Storage object prefix"
+	depends on PSA_INTERNAL_TRUSTED_STORAGE
+	default "its/"
```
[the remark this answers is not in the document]
mswarowsky: I took that from the original design, which made this part of the interface, so it could be used by other implementations as well.
By moving it to the SECURE_STORAGE_STORAGE_SETTINGS_BACKEND part we can not distinguish between PS and ITS anymore which was whished at one point in this ever ongoing thing.

subsys/secure_storage/Kconfig
```diff
@@ -0,0 +1,189 @@
...
+	  way device intimate data and protects against physical attacks.
+	  It supports several storage back-ends and several secure
+	  implementation back-ends to provide various levels of trust.
+
+if PSA_PROTECTED_STORAGE
+
+config PSA_PROTECTED_STORAGE_PREFIX
+	string "Protected Storage object prefix"
+	depends on PSA_PROTECTED_STORAGE
+	default "ps/"
```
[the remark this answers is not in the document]
mswarowsky: See comment for ITS

subsys/secure_storage/Kconfig
mswarowsky: Was taken from the previous work that Backend was attached to the end of each choice, but I don't care, so changed it

subsys/secure_storage/Kconfig
mswarowsky: It was given to me with this ChaCha-Poly1305 to just continued using that one @Vge0rge maybe knows more of the historic background.

subsys/secure_storage/Kconfig
mswarowsky: Changed it to secure storage backend, the help text should give more info

subsys/secure_storage/src/aead/aead_key_huk.c
mswarowsky: I just copied that code from our sample https://github.com/nrfconnect/sdk-nrf/blob/main/samples/keys/hw_unique_key/src/derive_key.c#L15 

`_Master key encryption key (MKEK). It is used for deriving Key Encryption Keys (KEKs), which are used to encrypt other keys when these are stored. It is provided to CryptoCell when it is used._`
As this storage was intended to be used mainly to store keys MKEK seems all right

subsys/secure_storage/src/aead/secure_backend_aead.c
mswarowsky: I could not find that function, can you tell me what to include to get that function

joerchan: It's part of mbedtls/platform_util.h, I think you you should have access to it.

subsys/secure_storage/src/aead/aead_key_hash.c
joerchan: Hmm, settings storage is sometimes used to store keys, but that defeats the purpose of storing things encrypted in settings. Unless you just want to get this part working, however we didn't add an "unencrypted backend" so I guess we need this as our just-works solution?

I think giving this option a help text of "this is unsecure" would be good, to make sure nobody uses this by accident and thinks it provides them any good security.

subsys/secure_storage/Kconfig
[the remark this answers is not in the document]
joerchan: Ok, I see that this has been put into the interface now, and is not easily change-able.

Maybe consider removing the "/" here and "." from suffixes, and let the storage backend form the path using slash and dots instead?
It just seems very specific to the settings backend, so I'm not sure if it is re-usable to others.

subsys/secure_storage/src/aead/secure_backend_aead.c
mswarowsky: Ahh ok typo explains why I couldn't find it.. changed it

subsys/secure_storage/Kconfig
Vge0rge: The first thing that I can recall is to have the exact same implementation for both CC310 and CC312 devices. Since the CCM/GCM we had to two either have different implementations or use CCM 128 bits for everything which is not a good security practice.

I don't remember if we had other reasons to be honest but I can search if I can find more documents.

In any case that was a design decision back then and I would not propose to change the design in this PR.

subsys/secure_storage/Kconfig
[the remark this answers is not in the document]
mswarowsky: Ok removed the / not sure where the . in the prefix symbols is?

subsys/secure_storage/Kconfig
mswarowsky: And CC310 doesn't even have GCM support, or do I mess up something again here?

subsys/secure_storage/src/aead/aead_key_hash.c
mswarowsky: OK Added and unsecure and not recommended block in the help text

subsys/secure_storage/Kconfig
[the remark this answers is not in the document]
joerchan: They're in the source file: 
```
#define SECURE_STORAGE_BACKEND_AEAD_FILENAME_SUFFIX_SIZE  ".size"
#define SECURE_STORAGE_BACKEND_AEAD_FILENAME_SUFFIX_FLAGS ".flags"
#define SECURE_STORAGE_BACKEND_AEAD_FILENAME_SUFFIX_DATA  ".data"
#define SECURE_STORAGE_BACKEND_AEAD_FILENAME_SUFFIX_NONCE ".nonce"
```

subsys/net/openthread/Kconfig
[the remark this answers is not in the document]
joerchan: Can't select a choice, so disregard this comment.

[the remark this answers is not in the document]
mswarowsky: Grouped the dependencies into an own entry

subsys/secure_storage/Kconfig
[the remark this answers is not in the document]
mswarowsky: Ah these, ok I was only looking in the Kconfigs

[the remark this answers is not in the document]
mswarowsky: Ok then We have to change the interface again to request dies data specify, 
So either work with a suffix or have its own function for it.

PR #11971, opened 2023-08-07
tools: Update ccache version to 4.8.2

jangalda-nsc:
Ccache 4.4+ has support for using the Redis protocol. It allows to configure shared ccache database.

### DIFF
--- a/scripts/tools-versions-linux.yml
+++ b/scripts/tools-versions-linux.yml
@@ -23,7 +23,7 @@ zephyr-sdk:
     - x86_64-zephyr-elf
     - arm-zephyr-eabi
 ccache:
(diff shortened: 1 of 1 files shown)


scripts/tools-versions-linux.yml
```diff
@@ -23,7 +23,7 @@ zephyr-sdk:
     - x86_64-zephyr-elf
     - arm-zephyr-eabi
 ccache:
-  version: 3.7.7
+  version: 4.8.2
```
miha-nordic, line 26: In PR title is version 4.7.2. Is 4.8.2 correct?

miha-nordic (approved):

PR #11974, opened 2023-08-07
modules: hostap: Update POSIX_MAX_FDS default

rlubos:
Since hostap Kconfig move to sdk-nrf repository, the default for POSIX_MAX_FDS set in the supplicant's Kconfig takes precedence over the default set in sdk-nrf originally. This causes issues with Zephyr samples running with default setting for this config, as supplicant sets this config very tight.

I propose to increase the default value set by supplicant, to leave a few FDS entries for the applications to use. Otherwise, every single socket-based sample will need to override this config in its prj.conf file.


modules/hostap/Kconfig
```diff
@@ -121,7 +121,8 @@ config POSIX_MAX_FDS
 	# l2_packet - 1
 	# ctrl_iface - 2 * socketpairs = 4(local and global)
 	# z_wpa_event_sock - 1 socketpair = 2
-	default 7 if !POSIX_API
+	# 4 left for the applications running in default configuration
+	default 11 if !POSIX_API
```
krish2718, line 125: We already have a PR under review that increases this to 11, so, we might as well make the default 16.

rlubos: Hmm that commit increases the config only for that app, I suggest here that we simply increase the default set by `hostap`. Otherwise we'll need to update every single sample that wants to sue socket API, as currently the default leaves no FD for the application to use. WDYT?

As for the default value I've left 4 entries for the samples, just like the default value from Zephyr.

krish2718: Sure, I understand. What I am trying to say is default `11` might not be enough, so, proposing default as 16.

* hostap - 7
* QT App - 4
* Apps - 0

rlubos: Ah, ok. I'll increase the default then.

krish2718 (approved):

sachinthegreen (approved):

PR #11976, opened 2023-08-07
hw_unique_key: Create read Kconfig option

mswarowsky:
On nRF52840 to use the HUK the immustable bootloader is needed. By adding a new Kconfig option that should be used the key is just read but not written we ensure during build time that the images includes the immutable bootloader.


melwee01 (approved):

lib/hw_unique_key/Kconfig
```diff
@@ -63,3 +63,14 @@ config HW_UNIQUE_KEY_PARTITION_SIZE
 	help
 	  Size of the flash page for storing the HUK.
 	  Only needed when there is no KMU for storing keys (nRF52840).
+
+config READ_HW_UNIQUE_KEY
```
Vge0rge, line 67: I think that it is better to be consistent here. So I would propose to rename the symbol READ_HW_UNIQUE_KEY -> HW_UNIQUE_KEY_READ

doc/nrf/libraries/others/hw_unique_key.rst
```diff
@@ -78,7 +78,7 @@ Additionally, you can enable the :kconfig:option:`CONFIG_HW_UNIQUE_KEY_RANDOM` o
 See :ref:`configure_application` for information on how to enable the required configuration options.
 
 You can then use the HUKs through the APIs in the :ref:`CC3xx platform libraries<nrf_cc3xx_platform_readme>`.
-You can also derive a key using :c:func:`hw_unique_key_derive_key`.
+You can also enable the :kconfig:option:`READ_HW_UNIQUE_KEY` option to derive a key using :c:func:`hw_unique_key_derive_key`.
```
Vge0rge, line 81: So this is not exactly accurate I think. Since this Kconfig option will not change anything in the code, it will just enforce some requires is how I see it.  i think that it has value to enforce this in code maybe? So basically ifdef hw_unique_key_derive_key  (and any functions that it calls) with the new Kconfig.

mia-ko (changes requested):

lib/hw_unique_key/Kconfig
SebastianBoe: +1.

It is convention to start with the module, and then suffix with features of the module.

lib/hw_unique_key/Kconfig
```diff
@@ -6,7 +6,7 @@
 
 config HW_UNIQUE_KEY_SRC
 	bool
-	default y if HW_UNIQUE_KEY || HW_UNIQUE_KEY_LOAD
+	default y if HW_UNIQUE_KEY || HW_UNIQUE_KEY_LOAD || READ_HW_UNIQUE_KEY
```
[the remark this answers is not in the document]
mswarowsky: Yeah sure Georgios requested the same, so changing it

lib/hw_unique_key/Kconfig
mswarowsky: Changing it

doc/nrf/libraries/others/hw_unique_key.rst
mswarowsky: Ok yeah wasn't sure if I should go that strict, but sounds good for me

doc/nrf/libraries/others/hw_unique_key.rst
```diff
@@ -78,7 +78,7 @@ Additionally, you can enable the :kconfig:option:`CONFIG_HW_UNIQUE_KEY_RANDOM` o
 See :ref:`configure_application` for information on how to enable the required configuration options.
 
 You can then use the HUKs through the APIs in the :ref:`CC3xx platform libraries<nrf_cc3xx_platform_readme>`.
-You can also derive a key using :c:func:`hw_unique_key_derive_key`.
+You can also enable the :kconfig:option:`HW_UNIQUE_KEY_READ` option to derive a key using :c:func:`hw_unique_key_derive_key`.
```
mia-ko, line 81:
```suggestion
You can also enable the :kconfig:option:`CONFIG_HW_UNIQUE_KEY_READ` option to derive a key using :c:func:`hw_unique_key_derive_key`.
```

lib/hw_unique_key/Kconfig
```diff
@@ -63,3 +63,14 @@ config HW_UNIQUE_KEY_PARTITION_SIZE
...
+config HW_UNIQUE_KEY_READ
+	bool "Use HUK"
+	depends on !BUILD_WITH_TFM
+	select SECURE_BOOT if HAS_HW_NRF_ACL
+	select FPROTECT if HAS_HW_NRF_ACL
+	help
+	  Use the Hardware Unique Key (HUK) in read access to derive new keys.
+	  It will require an immutable bootloader with the CONFIG_HW_UNIQUE_KEY_LOAD
+	  option set on devices without a KMU.
+	  Note that the key has to be written before.
```
mia-ko, line 76: Btw, the key has to be written before what?

mia-ko (changes requested):

lib/hw_unique_key/Kconfig
mswarowsky: before it can be read. 
Is that enough or do you think that requires more details

lib/hw_unique_key/Kconfig
```diff
@@ -63,3 +63,14 @@ config HW_UNIQUE_KEY_PARTITION_SIZE
...
 	  Size of the flash page for storing the HUK.
 	  Only needed when there is no KMU for storing keys (nRF52840).
+
+config HW_UNIQUE_KEY_READ
+	bool "Use HUK"
+	depends on !BUILD_WITH_TFM
+	select SECURE_BOOT if HAS_HW_NRF_ACL
+	select FPROTECT if HAS_HW_NRF_ACL
+	help
+	  Use the Hardware Unique Key (HUK) in read access to derive new keys.
```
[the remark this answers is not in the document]
Vge0rge: If you prefer it actually, you can just rename it to HW_UNIQUE_KEY_USE or something like this.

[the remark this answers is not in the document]
mswarowsky: Hehe when we spoke at my desk you weren't a fan of use :D but sure we can rename that gonna change the description text.

lib/hw_unique_key/hw_unique_key.c
```diff
@@ -82,6 +82,7 @@ bool hw_unique_key_are_any_written(void)
 	return false;
 }
 
+#ifdef CONFIG_HW_UNIQUE_KEY_USE
```
Vge0rge, line 85: This function is used by the TF-M here:
https://github.com/nrfconnect/sdk-nrf/blob/8d165848f383a74e262ee66e2bdfad35aca1cb20/modules/tfm/tfm/boards/common/crypto_keys.c#L39

So I am expecting the option CONFIG_HW_UNIQUE_KEY_USE  should probably be propagated to TF-M for when it needs to use this function?

lib/hw_unique_key/Kconfig
mia-ko: I would maybe add more details. Is that something that user has to make sure that it's done? And how. Maybe reference to or mention of a function/option/procedure that is used, if any.

lib/hw_unique_key/hw_unique_key.c
mswarowsky: discussed offline

lib/hw_unique_key/Kconfig
mswarowsky: Oh was not aware that we include these procedure description in Kconfigs,

I added a reference to the Kconfigs used for that, I will describing the whole process here is a bit too much

mia-ko: Yeah, no need to describe the entire process.
It just needed some more info, because writing just simply that the key has to be written before it is read is kind of obvious and does not give that much of a useful info.
But writing that the key has to be written by XX before it can be read, or XX has to have completed and written the key before it can be read gives a clear instruction on what users should be looking out for.
Giving these references is useful I think, so if you are happy now, then I am as well :)

mswarowsky: Yeah makes sense, but I think the reference to the Kconfig that is needed for writing should be fine now

Vge0rge (approved):

mia-ko (approved):

mswarowsky:
Turned out we will run in cyclic Kconfig dependencies, So I give on on what was supposed to be a small fix...

PR #11977, opened 2023-08-08
applications: Added Bluetooth LE support to Matter Bridge

kkasperczyk-no:
Introduced the following changes:
* Splitted the configuration for simulated and Bluetooth LE bridged devices
* Implemented Bluetooth LE connectivity manager module being an engine for handling Bluetooth LE connections with bridged devices
* Implemented BLuetooth LE data provider module providing an API for the Bluetooth LE bridged device
* Implemented data provider example that uses Nordic LBS Bluetooth service and is represented as a Matter On/Off light.

To make the code review easier, here are several assumptions I made creating this PR:
* This is very initial version for Bluetooth LE support. We handle only happy case scenario, error handling, disconnections, etc. will be handled in the follow-up.
* The `BLEConnectivityManager` is agnostic module for handling Bluetooth LE connections and traffic without analyzing the exchanged data specific for some BLE service.
* The `BLEBridgedDeviceProvider` is an abstract module inheriting by the `BridgedDeviceDataProvider`. The idea for such dependency was that the final Bluetooth LE bridged device has to implement both `BridgedDeviceDataProvider` API to interpret Matter related data and `BLEBridgedDeviceProvider` to interpret Bluetooth LE related data.


applications/matter_bridge/src/app_task.cpp
```diff
@@ -53,6 +58,13 @@ FactoryResetLEDsWrapper<1> sFactoryResetLEDs{ { FACTORY_RESET_SIGNAL_LED } };
 bool sIsNetworkProvisioned = false;
 bool sIsNetworkEnabled = false;
 bool sHaveBLEConnections = false;
+
+#ifdef CONFIG_BRIDGED_DEVICE_BLE
+static struct bt_uuid *sUuidLbs = BT_UUID_LBS;
```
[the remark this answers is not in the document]
kkasperczyk-no: Yes, sure, I will remove it. I copied parts of the Bluetooth code written in C and this is the cause.

applications/matter_bridge/src/bridged_device_types/ble_onoff_light_data_provider.cpp
```diff
@@ -0,0 +1,180 @@
...
+		return -EINVAL;
+	}
+	mButtonCharacteristicHandle = gatt_desc->handle;
+
+	gatt_desc = bt_gatt_dm_desc_by_uuid(discoveredData, gatt_chrc, sUuidCcc);
+	if (!gatt_desc) {
+		LOG_ERR("No CCC descriptor found.");
+		return -EINVAL;
+	}
+	mCccHandle = gatt_desc->handle;
```
ArekBalysNordic, line 169: Nit: It seems that it can be simplified somehow using for.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
+
+using namespace chip;
+
+BLEConnectivityManager BLEConnectivityManager::sBLEConnectivityManager;
+
+BT_SCAN_CB_INIT(scan_cb, BLEConnectivityManager::FilterMatch, NULL, NULL, NULL);
+
+BT_CONN_CB_DEFINE(conn_callbacks) = { .connected = BLEConnectivityManager::ConnectionHandler,
```
ArekBalysNordic, line 25: I leave a note here, that if you define connection callbacks like here, they will be used also for Matter's BLE during the commissioning process. It isn't bad, but be aware of that.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -0,0 +1,310 @@
...
+			      .mConnParam));
+
+	/* TODO: obtain more information from advertising data to help user distinguish the devices (e.g. BT name). */
+
+	GetBLEConnectivityManager().mScannedDevicesCounter++;
+}
+
+void BLEConnectivityManager::ConnectionHandler(struct bt_conn *conn, uint8_t conn_err)
+{
+	const bt_addr_le_t *addr = bt_conn_get_dst(conn);
```
ArekBalysNordic, line 70: Here should be added a validation to check if we should use this part of the code because it also will be used during the commissioning process.

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -0,0 +1,64 @@
...
+	/* One BT connection is reserved for the Matter service purposes. */
+	static constexpr uint16_t kMaxCreatedDevices = CONFIG_BT_MAX_CONN - 1;
+	static constexpr uint8_t kMaxServiceUuids = CONFIG_BT_SCAN_UUID_CNT;
+
+	CHIP_ERROR Init(bt_uuid **serviceUuids, uint8_t serviceUuidsCount);
+	CHIP_ERROR Scan();
+	CHIP_ERROR StopScan();
+	CHIP_ERROR Connect(uint8_t index, BLEBridgedDevice::DeviceConnectedCallback callback, void *context,
+			   bt_uuid *serviceUuid);
+	BLEBridgedDevice *FindBLEBridgedDevice(struct bt_conn *conn);
```
ArekBalysNordic, line 35: Nit: It can be also defined as a reference and optional.

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -0,0 +1,64 @@
...
+	ScannedDevice mScannedDevices[kMaxScannedDevices];
+	BLEBridgedDevice mCreatedDevices[kMaxCreatedDevices];
+	bt_uuid *mServicesUuid[kMaxServiceUuids];
+	uint8_t mServicesUuidCount;
+
+	friend BLEConnectivityManager &GetBLEConnectivityManager();
+	static BLEConnectivityManager sBLEConnectivityManager;
+};
+
+inline BLEConnectivityManager &GetBLEConnectivityManager()
```
markaj-nordic, line 61: This is my subjective opinion so I leave the decision up to you @kkasperczyk-no but I prefer the singleton implementation with local static instance being returned by reference for the following reasons:
- less code
- more encapsulation
- might be less prone to the static initialization fiasco
- no need to additionally define the static object in the cpp file
- no need for friendship relationship

Also note that to achieve "the real" singleton pattern, we would need to make the constructor of the `BLEConnectivityManager` private.

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -46,6 +67,73 @@ static int add_bridged_device(const struct shell *shell, size_t argc, char **arg
...
+	context->deviceType = deviceType;
+
+	if (nodeLabel) {
+		strcpy(context->nodeLabel, nodeLabel);
+	}
+
+	BridgedDeviceDataProvider *provider = add_ble_provider(context->deviceType);
+
+	if (!provider) {
+		return -ENOMEM;
```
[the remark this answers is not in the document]
markaj-nordic: As discussed f2f, we may want to use a uniqe_ptr in this case.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+	return CHIP_NO_ERROR;
+}
+
+void BLEConnectivityManager::ScanTimeoutCallback(k_timer *timer)
+{
+	DeviceLayer::PlatformMgr().ScheduleWork(ScanTimeoutHandle, reinterpret_cast<intptr_t>(timer));
+}
+
+void BLEConnectivityManager::ScanTimeoutHandle(intptr_t context)
```
markaj-nordic, line 234: I think that ideally we should let the application to register this kind of callback and decide what to do with scan data. I can imagine the real device not wanting to interact with the console at all.

We may also think about single scan-result per callback, to let the application filter the particular device and (for instance) abort the scan once the device of interest is found (it would require API extension).

applications/matter_bridge/src/bridged_device_types/ble_onoff_light_data_provider.cpp
kkasperczyk-no: Not quite, the CCC descriptor doesn get char, but only descriptor. So we could simplify only led and button part, so create for loop to iterate through 2 elements :)

kkasperczyk-no: We could also create a list of handles, however I believe it would be harder to access them in other parts of the code, as every handle is used for other kind of interaction.

samples/matter/common/src/bridge/ble_bridged_device.h
```diff
@@ -0,0 +1,40 @@
...
+struct BLEBridgedDevice {
+	using DeviceConnectedCallback = void (*)(BLEBridgedDevice *device, struct bt_gatt_dm *discoveredData,
+						 bool discoverySucceeded, void *context);
+
+	bt_addr_le_t mAddr;
+	DeviceConnectedCallback mConnectedCallback;
+	void *mConnectedCallbackContext;
+	bt_uuid *mServiceUuid;
+	struct bt_conn *mConn;
+	BLEBridgedDeviceProvider *provider;
```
markaj-nordic, line 39:
```suggestion
	BLEBridgedDeviceProvider *mProvider;
```

samples/matter/common/src/bridge/ble_bridged_device.h
```diff
@@ -0,0 +1,40 @@
...
+	virtual void MatchBleDevice(BLEBridgedDevice *device) = 0;
+	virtual int ParseDiscoveredData(struct bt_gatt_dm *discoveredData) = 0;
+
+protected:
+	BLEBridgedDevice *mDevice;
+};
+
+struct BLEBridgedDevice {
+	using DeviceConnectedCallback = void (*)(BLEBridgedDevice *device, struct bt_gatt_dm *discoveredData,
+						 bool discoverySucceeded, void *context);
```
markaj-nordic, line 32: Maybe we can use `BluetoothConnectionContext` if this is an only use case instead of `void`, we would spare additional casts in the callback.

markaj-nordic (commented):
Second part of the review.

applications/matter_bridge/src/bridged_device_types/ble_onoff_light_data_provider.cpp
ArekBalysNordic: Ok, let's leave it as it is :)

applications/matter_bridge/src/bridged_device_types/ble_onoff_light_data_provider.h
```diff
@@ -0,0 +1,47 @@
...
+
+	static void NotifyAttributeChange(intptr_t context);
+	static void GattWriteCallback(struct bt_conn *conn, uint8_t err, struct bt_gatt_write_params *params);
+	static uint8_t GattNotifyCallback(struct bt_conn *conn, struct bt_gatt_subscribe_params *params,
+					  const void *data, uint16_t length);
+
+	bt_uuid *GetServiceUuid() override;
+	void MatchBleDevice(BLEBridgedDevice *device) override
+	{
+		mDevice = device;
```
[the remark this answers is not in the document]
kkasperczyk-no: Yeah, it was actually checked before invoking this method, but still it would be better to have a check inside as well. I changed method API to return error code and validate pointer.

applications/matter_bridge/src/bridged_device_types/ble_onoff_light_data_provider.cpp
```diff
@@ -0,0 +1,180 @@
...
+	case Clusters::OnOff::Attributes::OnOff::Id: {
+		memcpy(mGattWriteDataBuffer, buffer, sizeof(mOnOff));
+
+		mGattWriteParams.data = mGattWriteDataBuffer;
+		mGattWriteParams.offset = 0;
+		mGattWriteParams.length = sizeof(mOnOff);
+		mGattWriteParams.handle = mLedCharacteristicHandle;
+		mGattWriteParams.func = BleOnOffLightDataProvider::GattWriteCallback;
+
+		int err = bt_gatt_write(mDevice->mConn, &mGattWriteParams);
```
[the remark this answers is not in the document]
kkasperczyk-no: Hmm, fair point, it's not initialized in constructor, so it could happen. Fixed.

applications/matter_bridge/README.rst
```diff
@@ -107,29 +118,58 @@ You can use the following commands to control the bridge device:
...
-      * ``770`` - Temperature Sensor.
-      * ``775`` - Humidity Sensor.
+      * ``<device_type>`` is a Matter device type to use for the bridged device.
+      The argument is mandatory and accepts the following values:
 
-   * ``[node label]`` is a node label for the bridged device.
-     The argument is optional and you can use it to better personalize the device in your application.
+         * ``256`` - On/Off Light.
+         * ``770`` - Temperature Sensor.
+         * ``775`` - Humidity Sensor.
```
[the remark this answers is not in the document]
kkasperczyk-no: The numbers are device type ids from the Matter spec, so I thought it will be better to use them instead of creating some custom numbering. And the supported device types are displayed in the command help or if argument is not provided:

> uart:~$ matter_bridge add
add: wrong parameter count
add - Adds bridged device. 
      Usage: add <bridged_device_type> [node_label]
      * bridged_device_type - the bridged device's type, e.g. 256 - OnOff Light, 770 - TemperatureSensor, 775 - HumiditySensor
      * node_label - the optional bridged device's node label

applications/matter_bridge/README.rst
```diff
@@ -157,6 +197,23 @@ You can use the following commands to control the bridge device:
...
+
+      .. parsed-literal::
+         :class: highlight
+
+         I: Scan result:
+         I: -----------------------------
+         I: | Index |      Address      |
+         I: -----------------------------
+         I: | 0     | e6:11:40:96:a0:18 |
+
```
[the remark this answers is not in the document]
kkasperczyk-no: Ok

applications/matter_bridge/README.rst
```diff
@@ -258,6 +321,74 @@ After building the sample and programming it to your development kit, complete t
...
+      :class: highlight
+
+      uart:~$ matter_bridge add <device_type> <bluetooth_device_index>
+
+   The *<device_type>* is a Matter device type that will be used to represent a new bridged device in the Matter Data Model.
+   The following values are currently supported:
+
+   * ``256`` - On/Off Light
+   * ``770`` - Temperature Sensor
+   * ``775`` - Humidity Sensor
```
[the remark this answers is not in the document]
kkasperczyk-no: No, good catch, it's copy paste error from simulated version.

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -46,6 +67,73 @@ static int add_bridged_device(const struct shell *shell, size_t argc, char **arg
...
+	GetBLEConnectivityManager().Connect(bleDeviceIndex, bluetooth_device_connected, context,
+					    context->provider->GetServiceUuid());
+
+#else
+
+	if (argv[2]) {
+		nodeLabel = argv[2];
+	}
+
+#ifdef CONFIG_BRIDGED_DEVICE_SIMULATED
```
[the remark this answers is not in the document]
kkasperczyk-no: Why?

samples/matter/common/src/bridge/ble_bridged_device.h
kkasperczyk-no: I don't think we should do that. First of all the `BluetoothConnectionContext` contains `shell *` and the `BLEBridgedDevice` doesn't depend on the application implementation. Generally the application may want to pass some custom content to the callback context, so I think keeping it void * is a proper design.

applications/matter_bridge/README.rst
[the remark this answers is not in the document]
ArekBalysNordic: Oh, right, then everything is alright :)

applications/matter_bridge/src/bridge_shell.cpp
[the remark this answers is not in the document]
ArekBalysNordic: I think you can merge `#else` above with this ifdef and make `#elif CONFIG_BRIDGED_DEVICE_SIMULATED` and then remove one `#endif` at the bottom.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: > I think that ideally we should let the application to register this kind of callback and decide what to do with scan data

Totally agree, I wanted to implement it, but then I forgot to do that.

> We may also think about single scan-result per callback

Unfortunately Bluetooth API is not as good as Wi-Fi in this matter. In Bluetooth you will receive the same result from the same address multiple times (e.g. if you have 3 devices you will receive like ~30 calls in 10 s, 10 per each device), as it just reports the fact every time it captures advertisement. That's why I decided to create a list in the Bluetooth module and pass processed data to the application.

applications/matter_bridge/README.rst
```diff
@@ -36,9 +36,20 @@ The application supports bridging the following Matter device types:
...
+
 The bridge supports adding and removing bridged devices dynamically at application runtime using commands from a dedicated Matter bridge shell module.
 
+Bluetooth LE Bridged devices
+============================
+
+The application supports following Bluetooth LE services that are represented by the corresponding Matter device types:
+
+* :ref:`Nordic Led Button Service <lbs_readme>` represented by the Matter On/Off Light
+
```
greg-fer, line 52: I would rather have all of this as one section "Bridged devices", in which I would explain what the bridged devices are and what types are implemented by the sample. Then, I would move the Kconfigs to a subsection of Configuration "Enabling bridged devices".

applications/matter_bridge/README.rst
```diff
@@ -107,29 +118,58 @@ You can use the following commands to control the bridge device:
 
 * For adding a new bridged device to the bridge:
 
-   .. code-block:: console
+   * For simulated bridged device:
```
greg-fer, line 138: I will rework this list into a table in a separate commit.

applications/matter_bridge/README.rst
```diff
@@ -218,6 +275,12 @@ After building the sample and programming it to your development kit, complete t
    See `Commissioning the device`_ for more information.
    During the commissioning process, write down the value for the bridge node ID.
    This is going to be used in the next steps (*<bridge_node_ID>*).
+
+Complete the next steps, depending on used bridged devices configuration.
```
greg-fer, line 279: We can't interrupt the testing steps like this.
```suggestion
#. Complete the next steps, depending on your bridged device configuration.
```

applications/matter_bridge/README.rst
```diff
@@ -218,6 +275,12 @@ After building the sample and programming it to your development kit, complete t
...
+
+Testing with simulated bridged devices
```
greg-fer, line 281: I suggest we use tabs for these two sections.

applications/matter_bridge/README.rst
```diff
@@ -258,6 +321,74 @@ After building the sample and programming it to your development kit, complete t
...
+   The *<device_type>* is a Matter device type that will be used to represent a new bridged device in the Matter Data Model.
+   The following values are currently supported:
+
+   * ``256`` - On/Off Light
+   * ``770`` - Temperature Sensor
+   * ``775`` - Humidity Sensor
+
+   The *<bluetooth_device_index>* is a Bluetooth LE device index that was scanned in the previous step.
+
+   For example, if you want to add a new Bluetooth LE bridged device with index `0` that will be exposed as an On/Off Light endpoint, use the following command:
```
greg-fer, line 364:
```suggestion
   For example, if you want to add a new Bluetooth LE bridged device with index `0` that is exposed as an On/Off Light endpoint, use the following command:
```

applications/matter_bridge/README.rst
```diff
@@ -258,6 +321,74 @@ After building the sample and programming it to your development kit, complete t
...
+
+   .. parsed-literal::
+      :class: highlight
+
+      uart:~$ matter_bridge add 256 0
+
+   You should be able to see the following output on the terminal:
```
greg-fer, line 371: See the edit to line 336.

greg-fer (changes requested):
I would consider a changelog entry, but the call is yours.

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include "ble_bridged_device.h"
+#include "bridged_device_data_provider.h"
+
+#include <zephyr/bluetooth/addr.h>
+#include <zephyr/bluetooth/conn.h>
+#include <zephyr/bluetooth/gatt.h>
+#include <zephyr/kernel.h>
+
+class BLEConnectivityManager {
```
markaj-nordic, line 17: Looks like `struct` keyword is populated across the `BLEConnectivityManager`, so I suggest to remove all redundant occurrences simply to have less code overall.

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -46,6 +66,83 @@ static int add_bridged_device(const struct shell *shell, size_t argc, char **arg
...
+
+	Platform::UniquePtr<BluetoothConnectionContext> contextPtr(context);
+	contextPtr->shell = shell;
+	contextPtr->deviceType = deviceType;
+
+	if (nodeLabel) {
+		strcpy(contextPtr->nodeLabel, nodeLabel);
+	}
+
+	BridgedDeviceDataProvider *provider = create_ble_provider(contextPtr->deviceType);
```
markaj-nordic, line 119: We are using a `create_ble_provider()` which crates the concrete version of `BridgedDeviceDataProvider` (`BLEBridgedDeviceProvider`). I suggest to either have one generic factory creating `BridgedDeviceDataProvider` or make the `create_ble_provider` to return concrete type. The latter would require to change the 
`using BleDataProviderFactory = DeviceFactory<BridgedDeviceDataProvider, UpdateAttributeCallback>;` 
into 
`using BleDataProviderFactory = DeviceFactory<BLEBridgedDeviceProvider, UpdateAttributeCallback>;`

Then no casting would be needed below - also if we are sure that the downcast is valid probably the `static_cast` is better then `reinterpret_cast` (we cannot use `dynamic_cast` because of lack of RTTI support).

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -46,6 +66,83 @@ static int add_bridged_device(const struct shell *shell, size_t argc, char **arg
...
+
+	if (!provider) {
+		return -ENOMEM;
+	}
+
+	contextPtr->provider = reinterpret_cast<BLEBridgedDeviceProvider *>(provider);
+	CHIP_ERROR err = GetBLEConnectivityManager().Connect(bleDeviceIndex, bluetooth_device_connected, contextPtr->get(),
```
markaj-nordic: We are passing the context raw pointer to the ConnectivityManager::Connect and we release the uniqe_ptr if the operation succeeded, which is good. But then which entity is responsible for managing the context's life time? I would expect the ConnectivityManager, but eventually we land in `bluetooth_device_connected` callback, right? Should the context pointer be released there then?

markaj-nordic (commented):
Last part of the review. Apologies for being iterative.

applications/matter_bridge/src/bridge_shell.cpp
kkasperczyk-no: I think you put this comment without refreshing, as it was fixed earlier. The unique ptr is created in connected callback, so the context should be deleted properly.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -0,0 +1,310 @@
...
+{
+	GetBLEConnectivityManager().StopScan();
+
+	LOG_INF("Scan result:");
+	LOG_INF("-----------------------------");
+	LOG_INF("| Index |      Address      |");
+	LOG_INF("-----------------------------");
+	for (int i = 0; i < GetBLEConnectivityManager().mScannedDevicesCounter; i++) {
+		LOG_INF("| %d     | %02x:%02x:%02x:%02x:%02x:%02x |", i,
+			GetBLEConnectivityManager().mScannedDevices[i].mAddr.a.val[5],
```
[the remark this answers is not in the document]
kkasperczyk-no: I don't know, it's always a trade off. This way we would need to memcpy the address to the temporary variable to only print it.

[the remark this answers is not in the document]
kkasperczyk-no: Btw it was sort of changed after the most recent push. Please see if the comment is still applicable.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: As we discussed f2f, there is not much we can do to avoid it, as while being in the central role we cannot clearly say when we started or not (like it is done for peripheral role in case of advertising). For now checking if the address is present on a list of connected devices and returning in case it is not there seems to be good enough.

applications/matter_bridge/src/bridged_device_factory.h
```diff
@@ -76,23 +91,44 @@ inline BridgedDeviceFactory &GetBridgedDeviceFactory()
...
+			  return chip::Platform::New<SimulatedTemperatureSensorDataProvider>(clb);
+		  } },
 #endif
 	};
 	return sDeviceDataProvider;
 }
+#endif
+
+#ifdef CONFIG_BRIDGED_DEVICE_BLE
+inline BleDataProviderFactory &GetBleDataProviderFactory()
```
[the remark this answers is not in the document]
kkasperczyk-no: I believe we discussed this f2f? In theory it is be possible that someone would like to support more than 1 fabric at a time to bridge several different kinds of devices, so we cannot assume only 1 is valid.

samples/matter/common/src/bridge/ble_connectivity_manager.h
kkasperczyk-no: Do you mean something like:     

> static BLEConnectivityManager & Instance()
>     {
>         static BLEConnectivityManager sInstance;
>         return sInstance;
>     }

?

markaj-nordic: Yes.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
markaj-nordic: Ok, sounds fair.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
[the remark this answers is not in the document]
markaj-nordic: Ok, if we need additional `memcpy`, then we'd better leave it as it is.

applications/matter_bridge/src/bridge_shell.cpp
markaj-nordic: Right.

applications/matter_bridge/Kconfig
```diff
@@ -11,12 +11,58 @@ config BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE
...
+choice BRIDGED_DEVICE_IMPLEMENTATION
+	prompt "Bridged Device implementation"
+	default BRIDGED_DEVICE_SIMULATED
+
+	config BRIDGED_DEVICE_SIMULATED
+		bool "Simulated Bridged Device"
+		help
+		  Enables using simulated bridged devices that do not perform
+		  real operations, but emulate some exemplary data to mock the real
+		  devices behavior.
```
greg-fer, line 29:
```suggestion
		  Enables using simulated bridged devices that do not perform
		  real operations, but emulate exemplary data to imitate the real
		  device behavior.
```

greg-fer (commented):
Apologies for iterative review :(

applications/matter_bridge/Kconfig
```diff
@@ -11,12 +11,58 @@ config BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE
...
+
+	config BRIDGED_DEVICE_BLE
```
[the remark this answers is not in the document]
kkasperczyk-no: Changed to BRIDGED_DEVICE_BT

samples/matter/common/src/bridge/ble_connectivity_manager.h
kkasperczyk-no: @markaj-nordic I refactored `BLEConnectivityManager` and `BridgeManager` to follow design recommended by you.

kkasperczyk-no:
@markaj-nordic @ArekBalysNordic @greg-fer I applied multiple changes and squashed all commits. Please try to make one more round of review to get this PR done.

applications/matter_bridge/README.rst
```diff
@@ -29,15 +36,28 @@ Overview
...
-The application simulates the functionality of bridged devices by generating pseudorandom measurement results and changing the light state in a fixed time intervals.
-The bridge supports adding and removing bridged devices dynamically at application runtime using commands from a dedicated Matter bridge shell module.
+.. _matter_bridge_app_bridged_support:
+
+Bridged device support
+======================
+
+The application supports two bridged device configurations that are mutually exclusive:
+
+* Simulated bridged device - This configuration simulates the functionality of bridged devices by generating pseudorandom measurement results and changing the light state in a fixed time intervals.
```
greg-fer, line 53:
```suggestion
* Simulated bridged device - This configuration simulates the functionality of bridged devices by generating pseudorandom measurement results and changing the light state in fixed+time intervals.
```

applications/matter_bridge/README.rst
```diff
@@ -29,15 +36,28 @@ Overview
...
+* Bluetooth LE bridged device - This configuration allows to connect a real peripheral Bluetooth LE device to the Matter bridge and represent its functionalities using :ref:`Matter Data Model <ug_matter_overview_data_model>`.
+  The application supports the following Bluetooth LE services that are represented by corresponding Matter device types:
```
greg-fer, line 55:
```suggestion
  The application supports the following Bluetooth LE services:
```

applications/matter_bridge/README.rst
```diff
@@ -29,15 +36,28 @@ Overview
...
+
+  * Nordic Semiconductor's :ref:`LED Button Service <lbs_readme>` - represented by the Matter On/Off Light
```
greg-fer, line 57:
```suggestion
  * Nordic Semiconductor's :ref:`LED Button Service <lbs_readme>` - represented by the Matter On/Off Light device type.
```

applications/matter_bridge/README.rst
```diff
@@ -103,65 +123,116 @@ Button 2:
...
+      I: -----------------------------
+      I: | 0     | e6:11:40:96:a0:18 |
+
+Adding a simulated bridged device to the Matter bridge
+   Use the following command:
+
+   .. parsed-literal::
+      :class: highlight
+
+      matter_bridge add *<device_type>* *[node_label]*
```
greg-fer, line 151:
```suggestion
      matter_bridge add *<device_type>* *["node_label"]*
```

applications/matter_bridge/README.rst
```diff
@@ -103,65 +123,116 @@ Button 2:
...
+   * *<device_type>* is the Matter device type to use for the bridged device.
      The argument is mandatory and accepts the following values:
 
       * ``256`` - On/Off Light.
       * ``770`` - Temperature Sensor.
       * ``775`` - Humidity Sensor.
 
-   * ``[node label]`` is a node label for the bridged device.
-     The argument is optional and you can use it to better personalize the device in your application.
+   * *[node label]* is the node label for the bridged device.
```
greg-fer, line 162:
```suggestion
   * *["node label"]* is the node label for the bridged device.
```

applications/matter_bridge/README.rst
```diff
@@ -103,65 +123,116 @@ Button 2:
...
 
-         I: Adding OnOff Light bridged device
-         I: Added device to dynamic endpoint 3 (index=0)
+     * ``256`` - On/Off Light.
 
-* For removing a bridged device from the bridge:
+   * *<ble_device_index>* is the Bluetooth LE device index on the list returned by the ``scan`` command.
+     The argument is mandatory and accepts only the values returned by the ``scan`` command.
+
+   * *[node label]* is the node label for the bridged device.
```
greg-fer, line 189:
```suggestion
   * *["node label"]* is the node label for the bridged device.
```

applications/matter_bridge/README.rst
```diff
@@ -218,45 +289,101 @@ After building the sample and programming it to your development kit, complete t
...
+            For example, read the value of the *on-off* attribute from the *onoff* cluster using the following command:
 
-   .. parsed-literal::
-      :class: highlight
+            .. parsed-literal::
+               :class: highlight
+
+               ./chip-tool onoff read on-off *<bridge_node_ID>* *<bridged_device_endpoint_ID>*
+
+   .. tabs::
```
greg-fer, line 330:
```suggestion

```
Delete this line. It is duplicated with line 294.

applications/matter_bridge/README.rst
```diff
@@ -218,45 +289,101 @@ After building the sample and programming it to your development kit, complete t
...
+            .. parsed-literal::
+               :class: highlight
+
+               I: Adding OnOff Light bridged device
+               I: Added device to dynamic endpoint 3 (index=0)
+
+         #. Write down the value for the bridged device dynamic endpoint ID.
+            This is going to be used in the next steps (*<bridged_device_endpoint_ID>*).
+         #. Use the :doc:`CHIP Tool <matter:chip_tool_guide>` to read the value of an attribute from the bridged device endpoint.
+            For example, read the value of the *on-off* attribute from the *onoff* cluster using a following command:
```
greg-fer, line 381:
```suggestion
            For example, read the value of the *on-off* attribute from the *onoff* cluster using the following command:
```

greg-fer (changes requested):
Checked the documentation build. Spotted a couple of final issues.

greg-fer (approved):

ArekBalysNordic (approved):
LGTM

applications/matter_bridge/src/bridge_shell.cpp
markaj-nordic: Discussed f2f. Let's keep it as it is. I hope we will find cleaner solution when refactoring the factory.

markaj-nordic (approved):

PR #11981, opened 2023-08-08
samples: cellular: Updated samples with modem lib shutdown()

MechanicV1:
This PR adds nrf_modem_lib_shutdown() to gracefully shutdown modem on completion of the samples

### DIFF
--- a/samples/cellular/battery/src/main.c
+++ b/samples/cellular/battery/src/main.c
@@ -424,6 +424,10 @@ int main(void)
 
 		k_sleep(K_SECONDS(refresh_interval));
 	}
+	err = nrf_modem_lib_shutdown();
+	if (err) {
+		printk("Failed to shutdown modem lib, err %d\n", err);
+	}
 
 	return 0;
 }
--- a/samples/cellular/pdn/src/main.c
+++ b/samples/cellular/pdn/src/main.c
@@ -163,6 +163,10 @@ int main(void)
 	ifaddrs_print();
 
 	lte_lc_power_off();
+	err = nrf_modem_lib_shutdown();
+	if (err) {
+		printk("Failed to shutdown modem lib, err %d\n", err);
+	}
 	printk("Bye\n");
 
 	return 0;
--- a/samples/cellular/ciphersuites/src/main.c
+++ b/samples/cellular/ciphersuites/src/main.c
@@ -234,5 +234,10 @@ clean_up:
 
 	lte_lc_power_off();
 
+	err = nrf_modem_lib_shutdown();
+	if (err) {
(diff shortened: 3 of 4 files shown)


samples/cellular/battery/src/main.c
```diff
@@ -424,6 +424,11 @@ int main(void)
 
 		k_sleep(K_SECONDS(refresh_interval));
 	}
+	
+	err = nrf_modem_lib_shutdown();
```
eivindj-nordic, line 428: This is unreachable per now. The return statements (on error) in above while loop should shut down the modem library. Maybe add a `clean_up` goto as in the ciphersuites sample?

samples/cellular/coap_client/src/main.c
```diff
@@ -185,6 +185,18 @@ static void modem_configure(void)
 #endif /* defined(CONFIG_LTE_LINK_CONTROL) */
 }
 
+/**@brief Truns of LTE link and Shuts down modem to provide LTE link.
+ */
+static void modem_shutdown(void)
+{
+#if defined(CONFIG_LTE_LINK_CONTROL)
+	lte_lc_power_off();	
+#endif /* defined(CONFIG_LTE_LINK_CONTROL) */
```
eivindj-nordic, line 194: We should do a CFUN=0 here if the LTE LC is not enabled.

samples/cellular/pdn/src/main.c
```diff
@@ -163,6 +163,11 @@ int main(void)
 	ifaddrs_print();
 
 	lte_lc_power_off();
+	
+	err = nrf_modem_lib_shutdown();
+	if (err) {
+		printk("Failed to shutdown modem lib, err %d\n", err);
+	}
```
eivindj-nordic, line 170: nit: space after end of if bracket.

samples/cellular/coap_client/src/main.c
MechanicV1: nrf_modem_lib_shutdown() this does a cfun 0 after getting the status of device.

samples/cellular/coap_client/src/main.c
```diff
@@ -185,6 +185,19 @@ static void modem_configure(void)
 #endif /* defined(CONFIG_LTE_LINK_CONTROL) */
 }
 
+/**@brief Truns of LTE link and Shuts down modem to provide LTE link.
+ */
+static void modem_shutdown(void)
+{
+#if defined(CONFIG_LTE_LINK_CONTROL)
```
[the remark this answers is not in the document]
lemrey: Same thing for `modem_configure()` where again the call to `lte_lc_init_and_connect()` is `#ifdef`d.

PR #11983, opened 2023-08-08
pm/poweroff: import latest changes in Zephyr

gmarull:
CONFIG_PM is no longer available for existing nRF SoCs. It was only used to access system off; however, this functionality is now offered by the new independent `sys_poweroff()` API. This change allows saving close to 1K of ROM on all builds, since the PM subsystem is no longer required.


SebastianBoe (approved):

gmarull:
It looks like we have some Kconfig usage issues (incorrect usage of select) in CAF that lead to situations like

```
[2023-08-08T14:25:25.335Z] warning: CAF_POWER_MANAGER (defined at /jenkins_workspace/workspace/latest_sdk-nrf_PR-11983/nrf/subsys/caf/modules/Kconfig.power_manager:9) has direct dependencies POWEROFF && CAF with value n, but is currently being y-selected by the following symbols:

[2023-08-08T14:25:25.335Z]  - DESKTOP_POWER_MANAGER (defined at /jenkins_workspace/workspace/latest_sdk-nrf_PR-11983/nrf/applications/nrf_desktop/src/modules/Kconfig.caf_power_manager.default:9), with value y, direct dependencies y (value: y)
```

This likely never showed up because CONFIG_PM was always on.

@pdunaj

gmarull:
@maxd-nordic It looks like `subsys/debug/etb_trace/etb_trace_lp.c` relies on Kernel internals that are only available when CONFIG_PM=y, but this option is no longer supported for nRF SoCs. Can you please comment on that? We should not be using Kernel internals in the first place.

maxd-nordic:
@gmarull Yes, this feature relies on special handling just before and just after going to sleep. We didn't find a better way to do that back then. Is there another interface we could use? It's quite important to have this power-saving feature if one desires to use ETB in battery-backed production software. @jtguggedal

gmarull:
> @gmarull Yes, this feature relies on special handling just before and just after going to sleep. We didn't find a better way to do that back then. Is there another interface we could use? It's quite important to have this power-saving feature if one desires to use ETB in battery-backed production software. @jtguggedal

This likely needs upstream changes so that this information is exposed via a public API, also independent from PM.

gmarull:
@maxd-nordic we need a solution, this is blocking progress on other stuff. Can you open an issue upstream describing your needs? In the meantime this can be dropped, I guess. cc: @carlescufi

maxd-nordic:
> @maxd-nordic we need a solution, this is blocking progress on other stuff. Can you open an issue upstream describing your needs? In the meantime this can be dropped, I guess. cc: @carlescufi

https://github.com/zephyrproject-rtos/zephyr/issues/61413

MarekPieta (approved):
One small comment, in general looks good

pdunaj (approved):

PR #11989, opened 2023-08-09
mpsl: new way of doing calibration with ktimer

martintv:
instead of mpsl doing it.


martintv:
got merged in this: https://github.com/nrfconnect/sdk-nrf/pull/12147

PR #11990, opened 2023-08-09
nrf_security: Update user supplied RNG call

magnev:
Update user supplied PRNG functionality to also allow usage of default PRNG functions

ref: NCSDK-22996

### DIFF
--- a/subsys/nrf_security/src/psa_crypto_driver_wrappers.c
+++ b/subsys/nrf_security/src/psa_crypto_driver_wrappers.c
@@ -2435,24 +2435,28 @@ psa_status_t psa_driver_wrapper_init_random(psa_driver_random_context_t *context
 psa_status_t psa_driver_wrapper_get_random(psa_driver_random_context_t *context, uint8_t *output,
 					   size_t output_size)
 {
+	psa_status_t status;
 #if defined(PSA_CRYPTO_DRIVER_ALG_PRNG_TEST)
-	return prng_test_generate_random(output, output_size);
+	status = prng_test_generate_random(output, output_size);
+
+	if (status != PSA_ERROR_NOT_SUPPORTED) {
+		return status;
+	}
 #endif
 #if defined(PSA_CRYPTO_DRIVER_ALG_PRNG_CC3XX_PLATFORM)
-	int err;
 	size_t output_length;
 
 	/* Using internal context. */
 	(void)context;
 
 #if defined(PSA_CRYPTO_DRIVER_ALG_CTR_DRBG_CC3XX_PLATFORM)
-	err = nrf_cc3xx_platform_ctr_drbg_get(NULL, output, output_size, &output_length);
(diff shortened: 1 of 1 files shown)


SebastianBoe:
Hi, could you please elaborate on why we want to "also allow usage of default PRNG functions" ?

subsys/nrf_security/src/psa_crypto_driver_wrappers.c
```diff
@@ -2436,7 +2436,9 @@ psa_status_t psa_driver_wrapper_get_random(psa_driver_random_context_t *context,
 					   size_t output_size)
 {
 #if defined(PSA_CRYPTO_DRIVER_ALG_PRNG_TEST)
-	return prng_test_generate_random(output, output_size);
+	if (prng_test_generate_random(output, output_size) == PSA_SUCCESS) {
+		return PSA_SUCCESS;
+	}
```
SebastianBoe, line 2441: Let's be consistent with error handling in this function.

So,

int err;

err = bla_bla();

if(err == PSA_SUCCESS) {
return SUCCESS;
}

magnev:
> Hi, could you please elaborate on why we want to "also allow usage of default PRNG functions" ?

Hey, I want to use user defined RNG for a subset of test cases where it is required to produce deterministic output to be able to verify the functionality properly. For other test cases I still want to use the default PRNG functions that are selected by configs in   `psa_driver_wrapper_get_random()`

Do you want me to add this description to commit message ?

subsys/nrf_security/src/psa_crypto_driver_wrappers.c
magnev: Sure, I will update the code

subsys/nrf_security/src/psa_crypto_driver_wrappers.c
```diff
@@ -2436,7 +2436,9 @@ psa_status_t psa_driver_wrapper_get_random(psa_driver_random_context_t *context,
...
 #endif
 #if defined(PSA_CRYPTO_DRIVER_ALG_PRNG_CC3XX_PLATFORM)
 	int err;
```
SebastianBoe, line 2444: Usually fall-through is implemented with the driver returning PSA_ERROR_NOT_SUPPORTED and the wrapper checking for it, instead of checking ! SUCCESS.

Is it possible for the driver to return NOT_SUPPORTED so that we implement fall-through in a consistent way in the driver_wrappers?

SebastianBoe:
"Do you want me to add this description to commit message ?"
No, it's fine as it is, I just needed to understand the motivation to be able to review :)

subsys/nrf_security/src/psa_crypto_driver_wrappers.c
magnev: Yes, good point

SebastianBoe (approved):

mswarowsky (approved):

PR #11994, opened 2023-08-09
tools: adding pip requirements check workflow

DatGizmo:
* The workflow merges the requirements from nrf, mcuboot and zephyr
* Checks if the requirements-fixed and new generated file match
* Packages that where only in the current requirements-fixed are added to requirements-ci.txt
* Following packages will be added to requirements-fixed.txt with this change

    Deprecated
    Markdown
    PyGithub
    PyJWT
    PyNaCl
    azure-core
    azure-storage-blob
    capstone
    dill
    editdistance
    exceptiongroup
    gitlint-core
    graphviz
    grpcio
    importlib-resources
    isodate
    jsonschema-specifications
    lark
    libusb-package
    m2r2
    mistune
    msgpack
    pathspec
    pkgutil_resolve_name
    platformdirs
    pypng
    referencing
    rpds-py
    sphinx-copybutton
    sphinx-markdown-tables
    sphinx-ncs-theme
    sphinx-notfound-page
    sphinx-togglebutton
    sphinxcontrib-jquery
    sphinxcontrib-plantuml


scripts/requirements.txt
```diff
@@ -1,2 +1,4 @@
 -r requirements-base.txt
 -r requirements-build.txt
+-r requirements-ci.txt
```
thst-nordic, line 3: check with doc team to see if this is ok to edit.

scripts/requirements-ci.txt
```diff
@@ -0,0 +1,15 @@
+aenum
+bitarray
+chardet
+gitlint==0.18.0
```
thst-nordic, line 4: add 'gitpython' for the twister test-spec script

.github/workflows/pip-requirements.yml
```diff
@@ -0,0 +1,81 @@
...
+          OLD=ncs/nrf/scripts/requirements-fixed.txt
+          echo -e "$NEW\t\t\t\t\t"$OLD
+          diff -y $NEW $OLD
+
+      - name: Post summary
+        if: ${{ !cancelled() }}
+        run: |
+          if [[ "failure" == "${{ steps.diff.outcome }}" ]]; then
+            echo -e 'New requirements-fixed.txt file differs from the current file.
+            Please add the newly created file to the PR.' >> $GITHUB_STEP_SUMMARY
```
gmarull, line 66: This will annoy many people who do not care about the management of Python dependencies. It should be automated by a bot.

gmarull (changes requested):

.github/workflows/pip-requirements.yml
thst-nordic: We had planned for a bot account to push a fixup commit on top of the PR so that the dependencies would be upgraded/fixed in the same PR.     Then decided it wasn't worth the possible mess.  

I see from here: 
https://github.com/nrfconnect/sdk-nrf/commits/main/scripts/requirements-fixed.txt
it is changing once a month, so we should automate it :/

scripts/requirements-ci.txt
thst-nordic: do it in another PR

.github/workflows/pip-requirements.yml
thst-nordic: We have this automation working in a different PR coming soon.  We will merge this as is.

gmarull: please, revert this patch. PR was not approved.

PR #11995, opened 2023-08-09
net: lib: aws_fota: Improve error handling for FOTA jobs

fnawratil:
- Improves error handling by providing more fine grained errors on job parsing.

- Jobs are now marked as failed if document is invalid but the job id could be parsed. 

- A new error state is introduced to enableretries for marking jobs as failed.

- Additionally, the job update accepted/rejected topics are no longer subscribed to, as this is not necessary to receive the respective messages and thus adds unnecessary overhead.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -590,6 +590,7 @@ Libraries for networking
 
     * The :kconfig:option:`CONFIG_AWS_FOTA_HOSTNAME_MAX_LEN` Kconfig option has been replaced by the :kconfig:option:`CONFIG_DOWNLOAD_CLIENT_MAX_HOSTNAME_SIZE` Kconfig option.
     * The :kconfig:option:`CONFIG_AWS_FOTA_FILE_PATH_MAX_LEN` Kconfig option has been replaced by the :kconfig:option:`CONFIG_DOWNLOAD_CLIENT_MAX_FILENAME_SIZE` Kconfig option.
+    * AWS FOTA jobs are now marked as failed if the job document for the update is invalid.
 
 * :ref:`lib_azure_fota` library:
 
--- a/subsys/net/lib/aws_fota/include/aws_fota_json.h
+++ b/subsys/net/lib/aws_fota/include/aws_fota_json.h
@@ -37,6 +37,16 @@ extern "C" {
  */
 #define EXECUTION_OBJ_DECODED_BIT 2
 
+/** @brief AWS FOTA JSON parse result. */
+enum aws_fota_json_result {
+	AWS_FOTA_JSON_RES_SKIPPED = 1,		/*!< Not a FOTA job document, skipped */
+	AWS_FOTA_JSON_RES_SUCCESS = 0,		/*!< Job document parsed successfully */
+	AWS_FOTA_JSON_RES_INVALID_PARAMS = -1,	/*!< Input parameters invalid */
+	AWS_FOTA_JSON_RES_INVALID_JOB = -2,	/*!< Job document invalid, could not get job id */
+	AWS_FOTA_JSON_RES_INVALID_DOCUMENT = -3,/*!< FOTA update data invalid */
+	AWS_FOTA_JSON_RES_URL_TOO_LONG = -4,	/*!< Parts of URL too large for buffer */
+};
+
 /**
  * @brief Parse a given AWS IoT DescribeJobExecution response JSON object.
  *	  More information on this object can be found at https://docs.aws.amazon.com/iot/latest/developerguide/jobs-api.html#mqtt-describejobexecution
@@ -54,9 +64,7 @@ extern "C" {
  * @param[in] file_path_buf_size  Size of the output buffer for the "file" field
  * @param[out] version_number  Version number from the Job Execution data type.
  *
- * @return 0 if the Job Execution object is empty, 1 if Job Execution object was
- *	     correctly decoded, otherwise a negative error code is returned
- *	     identicating reason of failure.
+ * @return aws_fota_json_result specifying the result of the parse operation.
  **/
 int aws_fota_parse_DescribeJobExecution_rsp(const char *job_document,
 					    uint32_t payload_len,
--- a/subsys/net/lib/aws_fota/src/aws_fota.c
+++ b/subsys/net/lib/aws_fota/src/aws_fota.c
@@ -23,6 +23,7 @@ static enum internal_state {
 	STATE_INIT,
 	STATE_DOWNLOADING,
 	STATE_DOWNLOAD_COMPLETE,
+	STATE_ERROR,
 } internal_state = STATE_UNINIT;
 
 /* Enum used when parsing AWS jobs topics messages are received on. */
@@ -78,6 +79,8 @@ static char *state2str(enum internal_state state)
 		return "STATE_DOWNLOADING";
 	case STATE_DOWNLOAD_COMPLETE:
 		return "STATE_DOWNLOAD_COMPLETE";
+	case STATE_ERROR:
+		return "STATE_ERROR";
 	default:
 		return "Unknown";
 	}
@@ -95,13 +98,18 @@ static void internal_state_set(enum internal_state new_state)
 	internal_state = new_state;
 }
 
+static void set_current_job_id(uint8_t *job_id)
+{
+	strncpy(job_id_handling, job_id, sizeof(job_id_handling));
+	job_id_handling[sizeof(job_id_handling) - 1] = '\0';
+}
+
 static void reset_library(void)
 {
 	internal_state_set(STATE_INIT);
 	execution_status = AWS_JOBS_QUEUED;
 	download_progress = 0;
-	strncpy(job_id_handling, AWS_JOB_ID_DEFAULT, sizeof(job_id_handling));
-	job_id_handling[sizeof(job_id_handling) - 1] = '\0';
+	set_current_job_id(AWS_JOB_ID_DEFAULT);
 	LOG_DBG("Library reset");
 }
 
@@ -142,8 +150,7 @@ static enum jobs_topic topic_type_get(const char *incoming_topic, size_t topic_l
  *
  * @return 0 If successful otherwise a negative error code is returned.
  */
-static int get_published_payload(struct mqtt_client *client, uint8_t *write_buf,
-				 size_t length)
+static int get_published_payload(struct mqtt_client *client, uint8_t *write_buf, size_t length)
 {
 	uint8_t *buf = write_buf;
 	uint8_t *end = buf + length;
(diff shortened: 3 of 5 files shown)


subsys/net/lib/aws_fota/include/aws_fota_json.h
```diff
@@ -37,6 +37,16 @@ extern "C" {
  */
 #define EXECUTION_OBJ_DECODED_BIT 2
 
+/** @brief AWS FOTA JSON parse result. */
+enum aws_fota_json_result {
+	AWS_FOTA_JSON_RES_SKIPPED = 1,		/*!< Not a FOTA job document, skipped */
```
simensrostad, line 42:
```suggestion
	AWS_FOTA_JSON_RES_SKIPPED = 1,		/* Not a FOTA job document, skipped */
```

simensrostad: Nit: Not familiar with this formatting, what value does it add?

simensrostad: And why doesn't success start at 0 and the rest including `AWS_FOTA_JSON_RES_SKIPPED` in decreasing order? Why have both negative and positive values

subsys/net/lib/aws_fota/include/aws_fota_json.h
```diff
@@ -54,9 +64,7 @@ extern "C" {
  * @param[in] file_path_buf_size  Size of the output buffer for the "file" field
  * @param[out] version_number  Version number from the Job Execution data type.
  *
- * @return 0 if the Job Execution object is empty, 1 if Job Execution object was
- *	     correctly decoded, otherwise a negative error code is returned
- *	     identicating reason of failure.
+ * @return aws_fota_json_result specifying the result of the parse operation
```
simensrostad, line 67:
```suggestion
 * @return aws_fota_json_result specifying the result of the parse operation.
```

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -208,6 +215,25 @@ static int update_job_execution(struct mqtt_client *const client,
 	return 0;
 }
 
+/**
+ * @brief Update the job document of the current job to AWS_JOBS_FAILED and move to ERROR state.
+ *
+ * @param client Connected MQTT client instance
```
simensrostad, line 221:
```suggestion
 * @param client Connected MQTT client instance

```

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -208,6 +215,25 @@ static int update_job_execution(struct mqtt_client *const client,
...
+ * @return 0 If successful otherwise a negative error code is returned.
+ */
+static int set_current_job_failed(struct mqtt_client *const client)
+{
+	struct aws_fota_event aws_fota_evt;
```
simensrostad, line 226:
```suggestion
	struct aws_fota_event aws_fota_evt {
	    .id = AWS_FOTA_EVT_ERROR
	}
```

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -208,6 +215,25 @@ static int update_job_execution(struct mqtt_client *const client,
...
+
+	aws_fota_evt.id = AWS_FOTA_EVT_ERROR;
```
simensrostad, line 228:
```suggestion
```

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -248,40 +273,52 @@ static int get_job_execution(struct mqtt_client *const client,
...
 
-	if (err < 0) {
-		LOG_ERR("Error when parsing the json: %d", err);
-		goto cleanup;
-	} else if (err == 0) {
+	if (err == AWS_FOTA_JSON_RES_SKIPPED) {
 		LOG_DBG("Got only one field");
 		LOG_DBG("No queued jobs for this device");
 		return 0;
+	} else if (err < 0
```
simensrostad, line 280: Alternative formatting (suggestion):

```
	} else if ((err < 0) &&
		   (err != AWS_FOTA_JSON_RES_INVALID_DOCUMENT) &&
		   (err != AWS_FOTA_JSON_RES_URL_TOO_LONG)) {
		LOG_ERR("Error when parsing the json: %d", err);
		err = -ENODATA;
		goto cleanup;
	}
```

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -248,40 +273,52 @@ static int get_job_execution(struct mqtt_client *const client,
...
+	set_current_job_id(job_id_incoming);
+
+	/* Check if the update data is valid */
+	if (err == AWS_FOTA_JSON_RES_INVALID_DOCUMENT) {
+		LOG_ERR("Invalid FOTA update document: %d", err);
+		return set_current_job_failed(client);
+	} else if (err == AWS_FOTA_JSON_RES_URL_TOO_LONG) {
+		LOG_ERR("URL elements too long for buffer: %d", err);
+		return set_current_job_failed(client);
+	}
```
simensrostad, line 304: No need to check for the other parse errors?

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -739,21 +757,7 @@ static void http_fota_handler(const struct fota_download_evt *evt)
...
-		aws_fota_evt.id = AWS_FOTA_EVT_ERROR;
-
-		callback(&aws_fota_evt);
-		reset_library();
-
-		/* If the FOTA download fails it might be due to the image being deleted.
-		 * Try to get the next job if any exist.
-		 */
-		(void)aws_jobs_get_job_execution(client_internal, "$next", get_topic);
+		(void) set_current_job_failed(client_internal);
```
simensrostad, line 760:
```suggestion
		(void)set_current_job_failed(client_internal);
```

simensrostad: A comment why the return value is ignored why do nice.

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -361,6 +392,12 @@ static int job_update_accepted(struct mqtt_client *const client,
 		reset_library();
 	}
 		break;
+	case AWS_JOBS_FAILED: {
```
simensrostad, line 395: Are the curly braces needed in this case? I think thats only needed when declaring variables inside the switch case.

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -514,24 +553,31 @@ static int on_connack_evt(struct mqtt_client *const client)
...
 		break;
-	default:
+	case STATE_ERROR:
+		status = AWS_JOBS_FAILED;
 		break;
+	default:
+		return 0;
+	}
+
+
```
simensrostad, line 569:
```suggestion
```

simensrostad (commented):
Overall looks good, few nits.

subsys/net/lib/aws_fota/include/aws_fota_json.h
fnawratil: > Nit: Not familiar with this formatting, what value does it add?

That's doxygen syntax for putting documentation after members: https://www.doxygen.nl/manual/docblocks.html#memberdoc



> And why doesn't success start at 0 and the rest including `AWS_FOTA_JSON_RES_SKIPPED` in decreasing order? Why have both negative and positive values

Because `AWS_FOTA_JSON_RES_SKIPPED` is not an error. This mimics the existing implementation and the way it is done in `aws_fota_mqtt_evt_handler`.

subsys/net/lib/aws_fota/src/aws_fota.c
fnawratil: That's being handled above (in the section of your previous comment), where we deal with all other errors except `AWS_FOTA_JSON_RES_INVALID_DOCUMENT` and `AWS_FOTA_JSON_RES_URL_TOO_LONG`.

jtguggedal (commented):
Sorry about the late review, slipped through the cracks..

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -95,13 +98,18 @@ static void internal_state_set(enum internal_state new_state)
 	internal_state = new_state;
 }
 
+static void set_current_job_id(uint8_t *job_id)
+{
+	strncpy(job_id_handling, job_id, sizeof(job_id_handling));
+	job_id_handling[sizeof(job_id_handling) - 1] = '\0';
```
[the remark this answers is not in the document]
fnawratil: Since job IDs can be of arbitrary length between 1 and 64 bytes, I don't think any checks make sense here.
https://docs.aws.amazon.com/iot/latest/apireference/API_Job.html#iot-Type-Job-jobId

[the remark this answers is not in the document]
jtguggedal: I meant to check if the null-terminated byte overwrote the last character of the ID. Then we'd use the wrong ID in any subsequent operation. But i the absolute max length of an ID is 64 characters, then null-terminating at index 64 is always safe.

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -208,6 +215,25 @@ static int update_job_execution(struct mqtt_client *const client,
...
+static int set_current_job_failed(struct mqtt_client *const client)
+{
+	struct aws_fota_event aws_fota_evt;
+
+	aws_fota_evt.id = AWS_FOTA_EVT_ERROR;
+	callback(&aws_fota_evt);
+	internal_state_set(STATE_ERROR);
+	return update_job_execution(client,
+				    job_id_handling,
+				    sizeof(job_id_handling),
```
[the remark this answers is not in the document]
fnawratil: Since we are not actually using the buffer length for anything except for a string comparison and work with the null termination internally anyways, I would actually propose getting rid of this parameter and just using `sizeof(job_id_handling)` directly in the one `strncmp` in `update_job_execution()`.

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -208,6 +215,25 @@ static int update_job_execution(struct mqtt_client *const client,
 	return 0;
 }
 
+/**
```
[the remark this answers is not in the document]
jtguggedal: Sure, if it's useful for people to have it there, no problem

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -739,21 +760,9 @@ static void http_fota_handler(const struct fota_download_evt *evt)
...
-		aws_fota_evt.id = AWS_FOTA_EVT_ERROR;
-
-		callback(&aws_fota_evt);
-		reset_library();
-
-		/* If the FOTA download fails it might be due to the image being deleted.
-		 * Try to get the next job if any exist.
-		 */
-		(void)aws_jobs_get_job_execution(client_internal, "$next", get_topic);
+		/* Ignore return value, since we can't deal with a failed update here anyways. */
```
simensrostad, line 764: Why can't we deal with a failed update here? IMO not checking the return value of a function deserves a descriptive comment if its not super obvious.

simensrostad (approved):
LGTM :dog:

jtguggedal (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -590,6 +590,7 @@ Libraries for networking
 
     * The :kconfig:option:`CONFIG_AWS_FOTA_HOSTNAME_MAX_LEN` Kconfig option has been replaced by the :kconfig:option:`CONFIG_DOWNLOAD_CLIENT_MAX_HOSTNAME_SIZE` Kconfig option.
     * The :kconfig:option:`CONFIG_AWS_FOTA_FILE_PATH_MAX_LEN` Kconfig option has been replaced by the :kconfig:option:`CONFIG_DOWNLOAD_CLIENT_MAX_FILENAME_SIZE` Kconfig option.
+    * Jobs are now marked as failed if the job document for the update is invalid.
```
divipillai, line 593:
```suggestion
    * AWS FOTA Jobs are now marked as failed if the job document for the update is invalid.
```
Just to be specific.

divipillai (approved):

PR #11996, opened 2023-08-09
doc: protcols: Update Wi-Fi samples memory footprints

krish2718:
* Remove non-Wi-Fi sample MQTT on 9160
* Use "twister --size" over "size" to get accurate foot prints
* Use proper sample for location that uses Wi-Fi
* Add scan sample to 9160


visinor (approved):
Looks good to me.

richabp (commented):
Does it require a changelog entry?

Also, unable to see the doc preview.

sachinthegreen (approved):

ioannisg (approved):
Thanks @krish2718

krish2718:
> Does it require a changelog entry?
No, this is just updating the footprints.
 
> Also, unable to see the doc preview.

Yeah, not sure why though.

richabp (approved):

PR #11997, opened 2023-08-09
Bluetooth: Mesh: do not send time status often than 30 seconds

alxelax:
Commit adds restriction for the Time model to not send Time Status more often than 30 seconds if this is aresult of state update after Time Set or Time Status commands.


alxelax:
PR has been converted into draft. I want to implement unit tests for the changes.

subsys/bluetooth/mesh/time_srv.c
```diff
@@ -10,6 +10,7 @@
 #include "time_util.h"
 
 #define SUBSEC_STEPS 256U
+#define RESTRICTION_TIME 30000ll
```
ludvigsj, line 13:
```suggestion
#define STATUS_INTERVAL_MIN 30000ll
```

ludvigsj (changes requested):
I have a few questions.

 1. Why does the time relaying introduce randomness, whereas the time set response does not?
 2. If I read the current code correctly, the time set handling will just not send a response if it is less than 30 seconds since last response was sent. In my opinion, we should always send one response after the "last" time set, potentially having to delay this response (between 1 and 30 seconds, depending on when the last status was sent).
 3. Also, if I read correctly, the relay handling will first wait until 30 seconds have passed and then schedule a delayed status at 20-50 seconds later after this, meaning there will be 50-80 seconds between these two relays. Is this on purpuse? If so: why?

ludvigsj:
> Time Set will always send a response to the sender here: https://github.com/nrfconnect/sdk-nrf/pull/11997/files#diff-626675688a86ce1b261757ffd4f251e64c8b334403c2432102a99ec505fa8c32R273 The only state change publishing is restricted by 30 seconds limitation.

My point is, it should probably sent one more status update than it does now. Meaning, the behavior should be this, I think:

t = 0, time_set
t = 0, publishes time because of time_set
t = 10, time_set
t = 20, time_set
t = 30, publish time a final time to inform of latest state change.

The last step will not be done with the current code.

alxelax:
> > Time Set will always send a response to the sender here: https://github.com/nrfconnect/sdk-nrf/pull/11997/files#diff-626675688a86ce1b261757ffd4f251e64c8b334403c2432102a99ec505fa8c32R273 The only state change publishing is restricted by 30 seconds limitation.
> 
> My point is, it should probably sent one more status update than it does now. Meaning, the behavior should be this, I think:
> 
> t = 0, time_set t = 0, publishes time because of time_set t = 10, time_set t = 20, time_set t = 30, publish time a final time to inform of latest state change.
> 
> The last step will not be done with the current code.

I asked this question to @omkar3141 during implementation. Should the server skip or postpone publishing? I got answered it should skip. 
@omkar3141, could you clarify @ludvigsj's concern?

subsys/bluetooth/mesh/time_srv.c
```diff
@@ -243,6 +265,11 @@ static int handle_time_set(struct bt_mesh_model *model, struct bt_mesh_msg_ctx *
 		srv->time_update_cb(srv, ctx, BT_MESH_TIME_SRV_SET_UPDATE);
 	}
 
+	/* The time server shouldn't send out statuses more often than every 30 seconds. */
+	if ((srv->data.timestamp + RESTRICTION_TIME) < srv->data.sync.uptime) {
+		(void)bt_mesh_time_srv_time_status_send(srv, NULL);
+	}
+
```
omkar3141, line 272: I think this change is not needed. If 'client' is manually setting the time, then we do want client to be notified with status as soon as possible. Since statuses can be lost due to collisions, client should be able to retry again quickly.

subsys/bluetooth/mesh/time_srv.c
```diff
@@ -486,7 +520,10 @@ int bt_mesh_time_srv_time_status_send(struct bt_mesh_time_srv *srv,
 		return -EOPNOTSUPP;
 	}
 
-	return send_time_status(srv->model, ctx, k_uptime_get());
+	srv->model->pub->ttl = 0;
+	srv->data.timestamp = uptime;
+
```
omkar3141, line 525: I would rather move these changes to "send_time_status()". Because "bt_mesh_time_srv_time_status_send()" is meant for sending unsoliciated Time_Status messages and this function can get blocked if there is no correct role.

subsys/bluetooth/mesh/time_srv.c
```diff
@@ -243,6 +265,11 @@ static int handle_time_set(struct bt_mesh_model *model, struct bt_mesh_msg_ctx *
 		srv->time_update_cb(srv, ctx, BT_MESH_TIME_SRV_SET_UPDATE);
 	}
 
+	/* The time server shouldn't send out statuses more often than every 30 seconds. */
+	if ((srv->data.timestamp + RESTRICTION_TIME) < srv->data.sync.uptime) {
+		(void)bt_mesh_time_srv_time_status_send(srv, NULL);
```
omkar3141, line 270: It will be better to call 'send_time_status()' here, since, server shall always send response while handling Time_Set message.

omkar3141 (changes requested):
The spec does not have clarity about whether 30 second limit applies to all kinds of publishing or only to periodic publishing. The intention in the spec seems to be to prevent flooding of network by Time_Status messages if someone sets the publish period to a low value. We will ask for clarification.

Meanwhile it is enough to implement rate limitation for handing of 'Time_Status' messages (in `handle_time_status`). So, if other server is misbehaving, this server does not start misbehaving as well.

omkar3141 (changes requested):

subsys/bluetooth/mesh/time_srv.c
alxelax: Server always sends response. Please take a look at line 273: https://github.com/nrfconnect/sdk-nrf/pull/11997/files#diff-626675688a86ce1b261757ffd4f251e64c8b334403c2432102a99ec505fa8c32R273

This is restriction for publishing Time Status as result of state change.
Please read errata:
> A Time Server should publish a Time Status message **when the TAI Seconds, the Subsecond, or the Time Zone 
> Offset Current fields of the Time state of the Time Server (see Section 5.1.1) have been updated as a result 
> of processing a Time Set message** (see Section 5.2.1.2) or a Time Status message (see Section 5.2.1.3), and 
> **the value of the Time Role state (see Section 5.1.2) is 0x01 (Time Authority) or 0x02 (Time Relay)**, 
> or the Time Server obtains a new Time state (see Section 5.1.1) from a trusted OOB source (such as a GPS or a 
> mobile phone synchronized with a cellular network).
> **The two consecutive publications of the Time Status message should be separated by at least 30 seconds.** 

Seems this code does exactly what errata require. Restricts publishing on state change over Time set for Time Relay and Time Authority.

subsys/bluetooth/mesh/time_srv.c
alxelax: It is immediately notified. Please take a look at the next code line.

alxelax:
> The spec does not have clarity about whether 30 second limit applies to all kinds of publishing or only to periodic publishing. The intention in the spec seems to be to prevent flooding of network by Time_Status messages if someone sets the publish period to a low value. We will ask for clarification.
> 
> Meanwhile it is enough to implement rate limitation for handing of 'Time_Status' messages (in `handle_time_status`). So, if other server is misbehaving, this server does not start misbehaving as well.

Hm... Seems spec clarifies the time restriction only for one kind of publishing. This is state change publishing. I added periodic here to follow 30 seconds rules. I'll remove considering time in periodic publishing. If user wants to do DDoS attack of network over the time service, let's allow to do this. :)

subsys/bluetooth/mesh/time_srv.c
```diff
@@ -217,7 +226,20 @@ static int handle_time_status(struct bt_mesh_model *model, struct bt_mesh_msg_ct
...
+
+		/* Random delay has been already scheduled. */
+		if (k_work_delayable_is_pending(&srv->status_delay)) {
+			return 0;
+		}
+
+		uint8_t rnd;
+
+		bt_rand(&rnd, sizeof(uint8_t));
+		k_work_schedule(&srv->status_delay, K_SECONDS(20 + rnd % 30));
```
ludvigsj, line 242: Jitter is supposed to be 20-50 milliseconds, here you delay 20-50 seconds

alxelax:
Discussed IRL. For now, we decided to apply publishing time restriction rules for handling Time Status messages and postpone for Time Set messages. At the current moment, publishing the last Time model state depends on configured periodic publishing time of the Time model. If Time Set is received during 30 seconds restriction time it is not clear what is better to do: publish immediately or postpone until the end of the already started 30 seconds timeout. 

If clarification for the Time Set message is gotten from Bluetooth SIG it will be applied later as a separate task\PR.

omkar3141 (approved):

ludvigsj (approved):

PR #12003, opened 2023-08-10
net: lib: lwm2m_client_utils: Add delay after credential changed

SeppoTakalo:
This is a workaround for some Bootstrap servers that might not be fast enough to send generated credentials into the target servers.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -597,6 +597,7 @@ Libraries for networking
 
     * Support for using pre-provisioned X.509 certificates.
     * Support for using DTLS Connection Identifier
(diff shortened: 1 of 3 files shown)


SeppoTakalo:
@VTPeltoketo if the problem happens almost every time on your environment, can you verify with this, whether 15s helps to remedy the issue, or do we need longer timeout?

rlubos (approved):

juhaylinen (approved):

VTPeltoketo:
> 

If I understood right, the corresponding code is not used by SLM

SeppoTakalo:
Screenshot from 2023-08-10 12-53-12

Actually, this was not the fix we require.
Apparently the state machine changes on Zephyr side causes the socket to be closed too early, so we don't send ACK for the bootstrap finnish command.
This seem to randomly work, but I'm working on a fix for the Zephyr side.

SeppoTakalo:
Closing as proper fix is now submitted.

PR #12005, opened 2023-08-10
Bluetooth: Mesh: alternate location publishing

alxelax:
The Location model shall support model publication. When configured for publication, the Generic Location Server shall publish Generic Location Global Status messages if the server knows its global location or shall publish Generic Location Local Status messages if the server knows its local location. If configured for periodic publications, and both global and local locations are known, the Generic Location Server shall alternate between publishing its local and global locations in consecutive publish periods.

### DIFF
--- a/subsys/bluetooth/mesh/gen_loc_srv.c
+++ b/subsys/bluetooth/mesh/gen_loc_srv.c
@@ -24,11 +24,6 @@
 		.is_mobile = false, .time_delta = 0, .precision_mm = 4096000,  \
 	}
 
-static bool pub_in_progress(struct bt_mesh_loc_srv *srv)
-{
-	return k_work_delayable_is_pending(&srv->pub.timer);
-}
-
 /* Global location */
 
 static void rsp_global(struct bt_mesh_model *model,
@@ -64,6 +59,7 @@ static int global_set(struct bt_mesh_model *model, struct bt_mesh_msg_ctx *ctx,
 	struct bt_mesh_loc_global global;
 
 	bt_mesh_loc_global_decode(buf, &global);
+	srv->pub_state.is_global_available = 1;
 
 	srv->handlers->global_set(srv, ctx, &global);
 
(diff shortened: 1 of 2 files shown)


subsys/bluetooth/mesh/gen_loc_srv.c
```diff
@@ -194,37 +185,60 @@ const struct bt_mesh_model_op _bt_mesh_loc_setup_srv_op[] = {
...
-				       BT_MESH_LOC_OP_GLOBAL_STATUS);
-		bt_mesh_loc_global_encode(srv->pub.msg, &loc);
-	} else if (srv->pub_op == BT_MESH_LOC_OP_LOCAL_STATUS) {
-		struct bt_mesh_loc_local loc = LOC_LOCAL_DEFAULT;
+static void local_update_handler(struct bt_mesh_loc_srv *srv)
+{
+	struct bt_mesh_loc_local loc = LOC_LOCAL_DEFAULT;
 
-		srv->handlers->local_get(srv, NULL, &loc);
+	srv->pub_state.last = 1;
```
ludvigsj, line 203: I'm not a fan of these magic numbers `1` and `0`. Either create #defines or an enum for them, or make the field a bool called something like `last_was_local`.

ludvigsj (changes requested):

subsys/bluetooth/mesh/gen_loc_srv.c
ludvigsj: The other fields should also be `bool`, and renamed `local_available` and `global_availble`. Makes it easier to follow when reading the code without having the struct with doc comments available on the same screen.

alxelax: I do not see the reason to have `bool` for them and spend 3(probably more if considering potential structure alignment) bytes if it is possible to use bitfields.

ludvigsj: At least rename the fields with `_available`. It is not obvious when reading the code without having the struct definition at hand.

akredalen (approved):

PR #12007, opened 2023-08-10
https_client: Move the sample to `net/` and add connection manager and Wi-Fi support

jtguggedal:
* Move the https_client sample to the net/ folder as preparation
to support multiple networking technologies.
* Rework the network handling to use the connection manager.
* Add support for nRF7002DK.

test-sdk-nrf: sdk-nrf-pr-12007


lemrey:
I would rather this sample be copied to /net than moved there, we use this sample a lot when working with nRF9160, have automated tests in CI and so forth.

samples/net/https_client/src/main.c
```diff
@@ -17,23 +17,21 @@
...
 #define HTTP_HEAD_LEN		(sizeof(HTTP_HEAD) - 1)
-
 #define HTTP_HDR_END		"\r\n\r\n"
 
 #define RECV_BUF_SIZE		2048
 #define TLS_SEC_TAG		42
 
 /* Macros used to subscribe to specific Zephyr NET management events. */
 #define L4_EVENT_MASK		(NET_EVENT_L4_CONNECTED | NET_EVENT_L4_DISCONNECTED)
+#define IP_EVENT_MASK		(NET_EVENT_IPV4_ADDR_ADD | NET_EVENT_IPV6_ADDR_ADD | NET_EVENT_IPV6_CMD_DAD_SUCCEED)
```
rlubos, line 34: `NET_EVENT_IPV6_CMD_DAD_SUCCEED` -> `NET_EVENT_IPV6_DAD_SUCCEED`
Is it really needed though? Doesn't seem to be handled in `ip_event_handler()`.

rlubos:
> I would rather this sample be copied to /net than moved there, we use this sample a lot when working with nRF9160, have automated tests in CI and so forth.

Would it be a big issue to switch to using the "generalized" sample? Having two copies of the sample IMO woudl just increase the maintenance cost.

lemrey:
> > I would rather this sample be copied to /net than moved there, we use this sample a lot when working with nRF9160, have automated tests in CI and so forth.
> 
> Would it be a big issue to switch to using the "generalized" sample? Having two copies of the sample IMO woudl just increase the maintenance cost.

perhaps not, but I would like to discuss #11604 in a meeting and try to find a resolution first, before we start moving samples out of `samples/cellular` into `samples/networking`

lemrey:
@MechanicV1 this sample is moving too, can you please update the tests?

samples/net/https_client/src/main.c
```diff
@@ -168,7 +153,61 @@ int tls_setup(int fd)
...
+		break;
+	}
+}
+
+static void ip_event_handler(struct net_mgmt_event_callback *cb,
+			     uint32_t event,
+			     struct net_if *iface)
+{
+	switch (event) {
+	case NET_EVENT_IPV4_ADDR_ADD:
```
rlubos, line 189: Since you already register for those events, maybe it'd make sense to actually print the assigned address? `cb->info` can be casted to `(const struct in_addr *)`/`(const struct in6_addr *)` (depending on the event) and then printed out (`net_addr_ntop()` can be used to convert address to a printable form).

samples/net/https_client/src/main.c
```diff
@@ -168,7 +153,61 @@ int tls_setup(int fd)
...
+		break;
+	}
+}
+
+static void connectivity_event_handler(struct net_mgmt_event_callback *cb,
+				       uint32_t event,
+				       struct net_if *iface)
+{
+	if (event == NET_EVENT_CONN_IF_FATAL_ERROR) {
+		printk("Fatal error received from the connectivity layer, rebooting\n");
```
rlubos, line 205: `rebooting` - is that really the case here?

samples/net/https_client/src/main.c
```diff
@@ -304,8 +301,65 @@ int main(void)
...
+		return 0;
+	}
+
+	k_sem_take(&network_connected_sem, K_FOREVER);
+
+	send_http_request();
+
+	/* A delay is in soome cases needed for the TCP stack to wrap up its work before taking down
+	 * the network interface.
+	 */
```
rlubos, line 355: `in some cases` - actually, I think it's safe to assume here, that it's reasonable to add a small delay between socket close / interface shutdown in all cases, to allow the TCP stack to gracefully teardown the connection. I'd remove that ambiguity from the comment, and just mention here that the small delay here is for the TCP connection teardown.

rlubos (commented):
Looks good, just some minor suggestions

divipillai (approved):

samples/net/https_client/sample.yaml
```diff
@@ -1,13 +1,22 @@
...
+    integration_platforms:
+      - nrf7002dk_nrf5340_cpuapp
+    platform_allow: nrf7002dk_nrf5340_cpuapp
+    extra_configs:
+      - CONFIG_WIFI_CREDENTIALS_STATIC_SSID="test-ssid"
+      - CONFIG_WIFI_CREDENTIALS_STATIC_PASSWORD="test-password"
+    tags: ci_build
   sample.cellular.https_client.tfm-mbedtls:
     build_only: true
     extra_args: OVERLAY_CONFIG="overlay-tfm_mbedtls.conf"
```
balaji-nordic, line 22: I think we need another build configuration that builds with this overlay for wifi

samples/net/https_client/src/main.c
```diff
@@ -168,7 +153,61 @@ int tls_setup(int fd)
...
+			     uint32_t event,
+			     struct net_if *iface)
+{
+	switch (event) {
+	case NET_EVENT_L4_CONNECTED:
+		printk("Network connectivity established and IP address assigned\n");
+		on_net_event_l4_connected();
+		break;
+	case NET_EVENT_L4_DISCONNECTED:
+		printk("Network connectivity lost, no IP address assigned\n");
```
balaji-nordic, line 176:
```suggestion
		printk("Network connectivity lost\n");
```

samples/net/https_client/src/main.c
```diff
@@ -304,8 +301,65 @@ int main(void)
...
+	/* Connecting to the configured connectivity layer.
+	 * Wi-Fi or LTE depending on the board that the sample was built for.
+	 */
+	err = conn_mgr_all_if_up(true);
+	if (err) {
+		printk("conn_mgr_all_if_up, error: %d\n", err);
+		return err;
+	}
+
+	/* Provision certificates before connecting to the LTE network */
```
balaji-nordic, line 335:
```suggestion
	/* Provision certificates before connecting to the network */
```

balaji-nordic (changes requested):

samples/net/https_client/src/main.c
KajaKoren: Actually this handler is not needed for the purpose of the sample, so it will be removed.

samples/net/https_client/sample.yaml
KajaKoren: This is not needed, it does not make sense to build the sample with this overlay for wifi, as it is already using Mbed TLS.

rlubos (approved):
LGTM, but please take a look at the compliance results (https://github.com/nrfconnect/sdk-nrf/actions/runs/6482379213/job/17601777566?pr=12007), there are a few minor issues reported.

balaji-nordic (approved):
Please fix compliance and squash the last commit with another one.

SeppoTakalo (approved):

KajaKoren (approved):

PR #12010, opened 2023-08-10
manifest: Update MCUboot with imporant fixes from upstream

de-nordic:
Fixes to boot_set_next and functions matching flash area to images and slots.


de-nordic:
No longer needed.

PR #12022, opened 2023-08-11
net: lwm2m_client_utils: Fix race condition on Firmware object

jheiskan81:
net: lwm2m: Fix race condition on Firmware object

Fix possible race conditions when state and results are written
by locking the registry, so a first write does not cause Notify message
to be send too early.


jheiskan81:
@SeppoTakalo @juhaylinen @rlubos Please review this fix.

SeppoTakalo (approved):

rlubos (approved):
Looks good, but please fix the commit message (see compliance).

juhaylinen (approved):

SeppoTakalo (approved):

PR #12023, opened 2023-08-11
nrf_security: Check for existence of RNG driver at build time

SebastianBoe:
Check for existence of an RNG driver at build time.

This moves input validation from runtime to build time.

We take psa_driver_wrapper_get_random functions that just return "not supported" out of the build as this should be detected at build time, not runtime.


mswarowsky (approved):
I like the fail early approach

PR #12024, opened 2023-08-11
applications: nrf5340_audio: Expand SD card module

mVaardal:
To avoid unnecessary opening
and closing of files


koffes (approved):
This is the same code as in https://github.com/nrfconnect/sdk-nrf/pull/11642

PR #12034, opened 2023-08-14
Report nRF Cloud connection protocol and method to device shadow

plskeggs:
Indicate in the reported.connection.protocol and .method fields the network application protocol (MQTT, REST, or CoAP) as well as the network connection method (LTE or Wi-Fi).

Enhance information reported by the nrf_cloud_rest_cell_location sample to include this.

Jira: IRIS-6732


samples/cellular/nrf_cloud_rest_cell_location/src/main.c
```diff
@@ -390,13 +390,31 @@ static void do_location_card_enable(void)
 		.ui = &ui
 	};
 
+	struct nrf_cloud_modem_info mdm_inf = {
+		/* Include all available modem info. Change to NRF_CLOUD_INFO_NO_CHANGE to
```
jayteemo, line 394: the comment would be more correct i think if you changed to the following:
```
.network = NRF_CLOUD_INFO_SET,
.sim = NRF_CLOUD_INFO_SET,
```

samples/cellular/nrf_cloud_rest_cell_location/src/main.c
```diff
@@ -390,13 +390,31 @@ static void do_location_card_enable(void)
...
+	};
+
 	/* Keep the connection alive so it can be used by the initial cellular
 	 * positioning request.
 	 */
 	rest_ctx.keep_alive = true;
 
 	LOG_INF("Enabling location card...");
-	err = nrf_cloud_rest_shadow_service_info_update(&rest_ctx, device_id, &svc_inf);
+	err = nrf_cloud_rest_shadow_device_status_update(&rest_ctx, device_id, &dev_status);
```
glarsennordic, line 417: This commit mentions that the sample is now reporting whether we are using LTE or Wi-Fi, but I don't see a change here that would do that

If the sample reporting that is just a side-effect of another commit, then consider not mentioning it in the commit message here

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -968,8 +969,46 @@ int json_send_to_cloud(cJSON *const request)
...
+{
+	int ret;
+
+#if defined(CONFIG_NRF_CLOUD_MQTT)
+	ret = json_add_num_cs(connection_obj, NRF_CLOUD_JSON_KEY_KEEPALIVE,
+			      CONFIG_NRF_CLOUD_MQTT_KEEPALIVE);
+#else
+	ret = json_add_null_cs(connection_obj, NRF_CLOUD_JSON_KEY_KEEPALIVE);
+#endif
+	ret += json_add_str_cs(connection_obj, NRF_CLOUD_JSON_KEY_PROTOCOL,
```
jayteemo, line 996: something like this (and one for method):
```
#if defined(CONFIG_NRF_CLOUD_MQTT)
  #define NRF_CLOUD_JSON_VAL_CFGD_PROTO_VAL   NRF_CLOUD_JSON_VAL_PROTO_MQTT
#elif defined(CONFIG_NRF_CLOUD_REST)
  #define NRF_CLOUD_JSON_VAL_CFGD_PROTO_VAL   NRF_CLOUD_JSON_VAL_PROTO_REST
#elif defined(CONFIG_NRF_CLOUD_COAP)
  #define NRF_CLOUD_JSON_VAL_CFGD_PROTO_VAL   NRF_CLOUD_JSON_VAL_PROTO_COAP
#else
  #define NRF_CLOUD_JSON_VAL_CFGD_PROTO_VAL   "Unknown"
#endif
```
would make this section easier on the eyes

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -968,8 +969,46 @@ int json_send_to_cloud(cJSON *const request)
 
 	return err;
 }
+
+static int shadow_connection_reset(cJSON *connection_obj)
```
glarsennordic, line 973: After a couple re-reads I see now that `shadow_connection_reset` is meant to write appropriate JSON data so-as to cause connection info to be reset. So it's "reset the shadow connection info" -- but the title reads a bit more like "reset the shadow connection"

Maybe it could be named something like `shadow_connection_info_blank`.

Either way it would be good to have some function briefs and a couple more inline comments for `shadow_connection_reset` and `shadow_connection_update`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -695,7 +697,7 @@ int nrf_cloud_state_encode(uint32_t reported_state, const bool update_desired_to
 		ret += json_add_null_cs(pairing_obj, NRF_CLOUD_JSON_KEY_CFG);
 		ret += json_add_null_cs(reported_obj, NRF_CLOUD_JSON_KEY_STAGE);
 		ret += json_add_null_cs(reported_obj, NRF_CLOUD_JSON_KEY_TOPIC_PRFX);
-		ret += json_add_null_cs(connection_obj, NRF_CLOUD_JSON_KEY_KEEPALIVE);
+		ret += shadow_connection_reset(connection_obj);
```
glarsennordic, line 700: What would happen if we called `shadow_connection_update` here instead of `shadow_connection_reset`?

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -717,9 +719,8 @@ int nrf_cloud_state_encode(uint32_t reported_state, const bool update_desired_to
 		ret += json_add_null_cs(pairing_obj, NRF_CLOUD_JSON_KEY_CFG);
 		ret += json_add_null_cs(reported_obj, NRF_CLOUD_JSON_KEY_PAIR_STAT);
 
-		/* Report keepalive value. */
-		ret += json_add_num_cs(connection_obj, NRF_CLOUD_JSON_KEY_KEEPALIVE,
-				       CONFIG_NRF_CLOUD_MQTT_KEEPALIVE);
+		/* Report keepalive, protocol, and menthod values. */
```
glarsennordic, line 722: menthod -> method

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -1694,6 +1733,18 @@ int nrf_cloud_shadow_dev_status_encode(const struct nrf_cloud_device_status *con
 		parent_obj = cJSON_AddObjectToObjectCS(root_obj, NRF_CLOUD_JSON_KEY_REP);
 	}
 
+	cJSON *connection_obj = cJSON_AddObjectToObjectCS(parent_obj, NRF_CLOUD_JSON_KEY_CONN);
```
jayteemo, line 1736: i realize this is supposed to be temporary, but who knows how long it will remain...
Can you add a kconfig to disable adding the new connection data?
All the other device status stuff can be turned on/off via the `nrf_cloud_device_status` struct.

glarsennordic:
I forgot to mention: Consider renaming `connection->method` to something else, pending internal discussion

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
plskeggs: Correct, method will need to always be device-reported.

samples/cellular/nrf_cloud_rest_cell_location/src/main.c
plskeggs: It's a side effect.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
plskeggs: The verb needs to go on the right.  These are static functions which we normally do not provide briefs on.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
plskeggs: Why?  The original code here set all fields to null.  I do not see a reason to break the pattern.

glarsennordic: My thinking was it would let you get rid of the `shadow_connection_reset` function altogether

And thinking in terms of how BE processes this stuff, writing blank here before later writing the actual data doesn't seem like it serves much purpose, but possibly I am mistaken

glarsennordic: Sure, in that case you can still consider keeping `ret += json_add_null_cs(connection_obj, NRF_CLOUD_JSON_KEY_KEEPALIVE);` if you want to avoid `shadow_connection_reset`, but if we're planning to rewrite all this it isn't a huge deal to me either way

plskeggs: But, the point of setting the JSON keys to a null value is, if they do not get set to a non-null value, it removes the key and value from the shadow.  It's basically a way to delete a field in the shadow.

If the UA_PIN_WAIT state is reached, the fields are nulled.  If it never advances to the UA_PIN_COMPLETE state, the fields will be deleted in the shadow.  Otherwise they will be set to the proper values in the shadow.

jayteemo: i believe the values are null'd just to sanitize the shadow so it is in a known state before continuing.

but i think you can just set the entire `"connection"` object to NULL, instead of the stuff inside it.

plskeggs: The other fields set to null in nrf_cloud_state_encode() for STATE_UA_PIN_WAIT set every subfield to null.  I am only following the existing pattern here.

I separated out the shadow_connection_reset() stuff because I thought I might need it for REST and CoAP, but I do not.  So I can just remove this function and put the json_add_null_cs() calls in nrf_cloud_state_encode().

glarsennordic: > But, the point of setting the JSON keys to a null value is, if they do not get set to a non-null value, it removes the key and value from the shadow

I guessed that might be the intent, but I will point out that we wind up writing several of them blank in the `STATE_UA_PIN_COMPLETE` case if they aren't set, so writing them blank in `STATE_UA_PIN_WAIT` still comes off as pointless to me.

A couple of them I'm not super sure on the behavior of, but I'd put forward that in a hypothetical cleanup of this area, maybe instead of blanking first, and then writing values later, let's consider writing a function that just writes all device info and blanks anything not in use by the device in one fell swoop.

plskeggs: https://nordicsemi.atlassian.net/browse/IRIS-6835

plskeggs: I assume the author did this in STATE_UA_PIN_WAIT in case STATE_UA_PIN_COMPLETE was never reached.  But someone can figure that out when they do the above ticket.

samples/cellular/nrf_cloud_rest_cell_location/src/main.c
plskeggs: The commit message already clarifies this:
`The latter information is done automatically when using nrf_cloud_rest_shadow_device_status_update().`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -1694,6 +1735,18 @@ int nrf_cloud_shadow_dev_status_encode(const struct nrf_cloud_device_status *con
...
```
jayteemo, line 1738: since this function is for adding the device status as indicated by the provided struct, can you wrap the entire thing:
```
#if defined(CONFIG_NRF_CLOUD_REPORT_CONNECTION)
        cJSON *connection_obj = cJSON_AddObjectToObjectCS(parent_obj, NRF_CLOUD_JSON_KEY_CONN);
    
	if (connection_obj == NULL) {
		err = -ENOMEM;
		goto cleanup;
	}

	err = shadow_connection_info_update(connection_obj);
	if (err) {
		goto cleanup;
	}
#endif
```

or maybe it would make sense to add it as on option in the struct?

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -98,6 +97,26 @@ static const char *const job_status_strings[] = {
...
+#elif defined(CONFIG_NRF_CLOUD_REST)
+#define NRF_CLOUD_JSON_VAL_CFGD_PROTO_VAL NRF_CLOUD_JSON_VAL_PROTO_REST
+#elif defined(CONFIG_NRF_CLOUD_COAP)
+#define NRF_CLOUD_JSON_VAL_CFGD_PROTO_VAL NRF_CLOUD_JSON_VAL_PROTO_COAP
+#else
+#define NRF_CLOUD_JSON_VAL_CFGD_PROTO_VAL "Unknown"
+#endif
+
+/* Define a string represention what connection method we are using. */
+#if defined(CONFIG_NRF_MODEM_LIB)
```
jayteemo, line 112: can `CONFIG_WIFI`  and `CONFIG_NRF_MODEM_LIB` both be enabled?
i think that is a valid use case, but not sure if it works now?

plskeggs: Yes, they can be both enabled, but only for Wi-Fi scanning for location services, with LTE being the only fully operating network stack.

The Wi-Fi stack plus Mbed-TLS plus Zephyr socket and UDP stack are all too large to fit on the nrf916x with the modem and modemlib.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
plskeggs: This function is called by `nrf_cloud_shadow_device_status_update()` which is in turn called by a couple of different MQTT-based samples and applications.  I would like to leave it as is.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
plskeggs: Yes, but I do not think there is a way to determine it at runtime yet, correct?

glarsennordic: I haven't looked into it deeply, there are some potential avenues, but I'm not sure whether any of them would even be the right choice -- it's just something I think should be on our radar

plskeggs: https://nordicsemi.atlassian.net/browse/IRIS-6842

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
jayteemo: yes, my point is that you are adding new functionality to this function which is different from its original purpose.
it originally encoded device status as requested by the provided struct.
now it is adding a connection object.

plskeggs: True.  Maybe it would be better to report it inside device in the shadow:
```
    "reported": {
      "device": {
        "deviceInfo": {
          "modemFirmware": "mfw_nrf9160_1.3.5"
        },
        "serviceInfo": {
        ...
        },
        "connInfo": {
                "protocol": "MQTT",
                "method": "LTE"
         }
      }
   }

Comment (jayteemo) [reply subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c:1738]: i like that for this feature.

**Review by glarsennordic — APPROVED** [review]

File: subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c (hunk 8 above)
Comment (jayteemo) [reply subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c]: consider squashing in the following commit: 
https://github.com/nrfconnect/sdk-nrf/commit/35564b1676bfb6dfc0f0bdbf0e6940859c924fe4

it was easier than explaining 😅

**Review by peknis — APPROVED** [review]

File: subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c (hunk 8 above)
Comment (plskeggs) [reply subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c]: Thanks for doing the work.  I pulled it in, rebased / split / etc., and added use of it to the other samples.

**Review by jayteemo — APPROVED** [review]

File: samples/cellular/nrf_cloud_rest_cell_location/prj.conf [hunk 10]
```diff
@@ -38,7 +38,7 @@ CONFIG_MODEM_INFO=y
 CONFIG_MODEM_INFO_ADD_DEVICE=y
 CONFIG_MODEM_INFO_ADD_NETWORK=y
 CONFIG_MODEM_INFO_ADD_DATE_TIME=n
-CONFIG_MODEM_INFO_ADD_SIM=n
+CONFIG_MODEM_INFO_ADD_SIM=y
```
lincolnthomas, line 41: Is this relevant to this PR?

samples/cellular/nrf_cloud_multi_service/prj.conf
```diff
@@ -116,3 +116,4 @@ CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=n
 CONFIG_NRF_CLOUD_ALERT=y
 CONFIG_NRF_CLOUD_LOG_DIRECT=y
 CONFIG_NRF_CLOUD_LOG_OUTPUT_LEVEL=3
+CONFIG_NRF_CLOUD_SEND_DEVICE_STATUS=n
```
lincolnthomas, line 119: Does this mean by default it won't write these new things to the shadow?

plskeggs: The application is adding the device status, so we do not want the library to also add it.

It's a workaround until I do this task: https://nordicsemi.atlassian.net/browse/IRIS-6792.

samples/cellular/nrf_cloud_rest_cell_location/prj.conf
plskeggs: Changes to address Justin's comment here:
https://github.com/nrfconnect/sdk-nrf/pull/12034#discussion_r1294960739
did not build correctly without enabling this config to add the SIM info.

PR #12038, opened 2023-08-15
nrf_security: oberon: Add ECB to CTR_DRBG

mswarowsky:
The oberon CTR_DRBG algorithm uses the ECB_NO_PADDING internally there it needs to be selected.


SebastianBoe (approved):

subsys/nrf_security/src/drivers/nrf_oberon/Kconfig
```diff
@@ -115,6 +115,8 @@ config PSA_CRYPTO_DRIVER_ALG_CTR_DRBG_OBERON
 	default y
 	depends on PSA_CRYPTO_DRIVER_ALG_PRNG_OBERON && \
 		   PSA_WANT_ALG_CTR_DRBG
+	select PSA_WANT_ALG_ECB_NO_PADDING
```
vili-nordic, line 118: I think we want to use "depends on" instead of select for this?

vili-nordic (approved):

PR #12040, opened 2023-08-15
Bluetooth: Mesh: Develop/fixing light lc sm conditions

Balaklaka:
According to sections 6.2.3.5 of MshMDLd1.1r12, the value of the
Light LC Ambient LuxLevel state is set to 0 at power-up and when
the node is provisioned on the network.
According to sections 6.2.5.12 and 6.2.5.7 the Occupancy On should
not trigger any transition.

Signed-off-by: Ingar Kulbrandstad <ingar.kulbrandstad@nordicsemi.no>

### DIFF
--- a/subsys/bluetooth/mesh/light_ctrl_srv.c
+++ b/subsys/bluetooth/mesh/light_ctrl_srv.c
@@ -999,13 +999,29 @@ static int handle_sensor_status(struct bt_mesh_model *model, struct bt_mesh_msg_
 
 		/* Occupancy sensor */
 
-		/* OCC_MODE must be enabled for the occupancy sensors to be
-		 * able to turn on the light:
-		 */
-		if (srv->state == LIGHT_CTRL_STATE_STANDBY &&
-		    !atomic_test_bit(&srv->flags, FLAG_OCC_MODE) &&
-		    !(atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
-		      !atomic_test_bit(&srv->flags, FLAG_MANUAL))) {
+		if ((srv->state == LIGHT_CTRL_STATE_STANDBY &&
+			/* According to the Mesh Model Specification section
+			 * 6.2.5.6: When in the specifications STANDBY state,
+			 * and the Auto Occupancy condition is false, the
+			 * Occupancy On event should not be processed.
+			 */
+			((!atomic_test_bit(&srv->flags, FLAG_OCC_MODE) &&
+			  !atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
+			  !atomic_test_bit(&srv->flags, FLAG_MANUAL))
+			  ||
+			/* According to the Mesh Model Specification section
+			 * 6.2.5.12: When in the specifications FADE STANDBY
+			 * MANUAL state, the Occupancy On event should not be
+			 * processed.
+			 */
+			  (atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
+			   atomic_test_bit(&srv->flags, FLAG_MANUAL)))) ||
+		    /* According to the Mesh Model Specification section
+		     * 6.2.5.7: When in the specifications FADE ON state, the
+		     * Occupancy On event should not be processed.
+		     */
+		    (srv->state == LIGHT_CTRL_STATE_ON &&
+		     atomic_test_bit(&srv->flags, FLAG_TRANSITION))) {
 			continue;
 		}
 
--- a/subsys/bluetooth/mesh/light_ctrl_srv.c
+++ b/subsys/bluetooth/mesh/light_ctrl_srv.c
@@ -1622,6 +1622,7 @@ static int light_ctrl_srv_start(struct bt_mesh_model *model)
 
 	switch (srv->lightness->ponoff.on_power_up) {
 	case BT_MESH_ON_POWER_UP_OFF:
+	case BT_MESH_ON_POWER_UP_ON:
 		if (atomic_test_bit(&srv->flags,
 				    FLAG_CTRL_SRV_MANUALLY_ENABLED)) {
 			ctrl_enable(srv);
@@ -1637,17 +1638,6 @@ static int light_ctrl_srv_start(struct bt_mesh_model *model)
 		 * restarts, we'll restore to On even though we were off in the
 		 * previous power cycle, unless we store the Off state here.
 		 */
-		store(srv, FLAG_STORE_STATE);
-		break;
-	case BT_MESH_ON_POWER_UP_ON:
-		if (atomic_test_bit(&srv->flags,
-				    FLAG_CTRL_SRV_MANUALLY_ENABLED)) {
-			ctrl_enable(srv);
-		} else {
-			lightness_on_power_up(srv->lightness);
-			ctrl_disable(srv);
-			schedule_resume_timer(srv);
-		}
 		store(srv, FLAG_STORE_STATE);
 		break;
 	case BT_MESH_ON_POWER_UP_RESTORE:
--- a/include/bluetooth/mesh/light_ctrl_srv.h
+++ b/include/bluetooth/mesh/light_ctrl_srv.h
@@ -99,11 +99,18 @@ struct bt_mesh_light_ctrl_srv;
 
 /** Light Lightness Control Server state */
 enum bt_mesh_light_ctrl_srv_state {
-	/** Standby state */
+	/** Standby state, merges the Light LC State Machine states of OFF,
+	 * STANDBY, FADE_STANDBY_AUTO and FADE_STANDBY_MANUAL in the Mesh Model
+	 * Specification section 6.2.5.
+	 */
 	LIGHT_CTRL_STATE_STANDBY,
-	/** On state */
+	/** On state, merges the Light LC State Machine states of FADE_ON and
+	 * RUN in the Mesh Model Specification section 6.2.5.
+	 */
 	LIGHT_CTRL_STATE_ON,
-	/** Prolong state */
+	/** Prolong state, merges the Light LC State Machine states of FADE and
+	 * PROLONG in the Mesh Model Specification section 6.2.5.
+	 */
 	LIGHT_CTRL_STATE_PROLONG,
 
 	/** The number of states. */


subsys/bluetooth/mesh/light_ctrl_srv.c
```diff
@@ -1620,8 +1627,13 @@ static int light_ctrl_srv_start(struct bt_mesh_model *model)
 
 	atomic_set_bit(&srv->flags, FLAG_STARTED);
 
+	if (IS_ENABLED(CONFIG_BT_MESH_LIGHT_CTRL_SRV_REG) && srv->reg) {
+		srv->reg->measured = 0;
+	}
+
 	switch (srv->lightness->ponoff.on_power_up) {
 	case BT_MESH_ON_POWER_UP_OFF:
+	case BT_MESH_ON_POWER_UP_ON:
```
omkar3141, line 1636: :) :)

subsys/bluetooth/mesh/light_ctrl_srv.c
```diff
@@ -999,13 +999,20 @@ static int handle_sensor_status(struct bt_mesh_model *model, struct bt_mesh_msg_
...
-		      !atomic_test_bit(&srv->flags, FLAG_MANUAL))) {
+		if ((srv->state == LIGHT_CTRL_STATE_STANDBY &&
+			((!atomic_test_bit(&srv->flags, FLAG_OCC_MODE) &&
+			  !atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
+			  !atomic_test_bit(&srv->flags, FLAG_MANUAL))
+			  ||
+			  (atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
+			   atomic_test_bit(&srv->flags, FLAG_MANUAL)))) ||
+		    (srv->state == LIGHT_CTRL_STATE_ON &&
+		     atomic_test_bit(&srv->flags, FLAG_TRANSITION))) {
```
PavelVPV, line 1015: The Occupancy On event can not be triggered in Fade On and Fade Standby Manual states, which correspond to the following flags:
| Spec State | Implementation state | Flags |
| --- | --- | --- |
| Fade On | LIGHT_CTRL_STATE_ON | FLAG_TRANSITION |
| Fade Manual Auto | LIGHT_CTRL_STATE_STANDBY | FLAG_TRANSITION && FLAG_MANUAL |

`FLAG_OCC_MODE` flag corresponds to Auto Occupancy.
So perhaps it should be like this?
```suggestion
		if (!atomic_test_bit(&srv->flags, FLAG_OCC_MODE) ||
		    (srv->state == LIGHT_CTRL_STATE_STANDBY &&
		     atomic_test_bit(&srv->flags, FLAG_TRANSITION) && 
		     atomic_test_bit(&srv->flags, FLAG_MANUAL)) ||
		    (srv->state == LIGHT_CTRL_STATE_ON &&
		     atomic_test_bit(&srv->flags, FLAG_TRANSITION))) {
```

omkar3141 (changes requested):

subsys/bluetooth/mesh/light_ctrl_srv.c
PavelVPV: > Umm, may be. But, Fade Manual Auto? There is no such state.

Sorry, type, should have been `Fade Standby Manual` of course.

PavelVPV: From your suggestion:
`/* If in Run state stay in the same state (basically, do not process Occupancy). */`
seems not correct as Occupancy On event is present in the 6.5.2.8.

subsys/bluetooth/mesh/light_ctrl_srv.c
```diff
@@ -1620,8 +1627,13 @@ static int light_ctrl_srv_start(struct bt_mesh_model *model)
 
 	atomic_set_bit(&srv->flags, FLAG_STARTED);
 
+	if (IS_ENABLED(CONFIG_BT_MESH_LIGHT_CTRL_SRV_REG) && srv->reg) {
+		srv->reg->measured = 0;
```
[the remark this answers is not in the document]
PavelVPV: > If we change something with regulator startup, this will be easily missed.

This should be covered by a test.

subsys/bluetooth/mesh/light_ctrl_srv.c
Balaklaka: The Occupancy conditions is only valid when in the spec standby state (in our code that means (standby, transition false, manual false), so the if the occ_mode is false this will remove occupancy for spec state Fade Standby Auto. The second state will remove occupancy from spec run state, as code on state both represent Fade On and Run state. Fade On does not have occupancy event while run state has occupancy. The third state represent Fade Standby Auto, but if OCC_Mode is false this condition will be overrun. This is why I made the if case specific to represent the state machine in the spec. I will add better comment to each state to explain it better.

PavelVPV: Unless I'm missing something, Occupancy On event is still present in Standy, Fade, Prolong and Fade Standby Auto states. It is the Condition that was updated. But, only Auto Occupancy had a special case in Standby state, in other states there was a dash line in the Condition column meaning that all Occupancy On conditions could be matched. So now, since other Occupancy Conditions were removed, for the states there is no change really except that Occupancy On event can trigger transition only if Auto Occupancy is enabled.

omkar3141: > seems not correct as Occupancy On event is present in the 6.5.2.8.
Sorry, right. That comment and condition gets removed.

Balaklaka: There is 3 states Occupancy On event should not have any effect, when in Standby and Auto Occupancy is false, Fade On, Fade Standby Manual state. That is what I have added specifically in the cases. (And of course Off state, but that is in the top of the function.)

subsys/bluetooth/mesh/light_ctrl_srv.c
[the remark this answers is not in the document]
Balaklaka: This is to make sure that the regulator uses random values before it receive it's first Ambient light value. It can receive other values than Amb light level, and until it receive its first value the spec want to make sure that the value is 0. Is not that the purpose of that sentence in the spec.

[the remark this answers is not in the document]
PavelVPV: The ambient light value is only used together with the regulator. The regulator is off until the lc server receives first ambient light level report: https://github.com/nrfconnect/sdk-nrf/blob/e28de89573274076beae9d0afc558288383be87b/subsys/bluetooth/mesh/light_ctrl_srv.c#L987-L996

subsys/bluetooth/mesh/light_ctrl_srv.c
PavelVPV: I think this is a bit complicated. 

This IMO is difficult to understand:
```
((!atomic_test_bit(&srv->flags, FLAG_OCC_MODE) &&
  !atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
  !atomic_test_bit(&srv->flags, FLAG_MANUAL))
 ||
/* According to the Mesh Model Specification section
 * 6.2.5.12 specifications FADE STANDBY MANUAL state,
 * the Occupancy On event should not be processed.
 */
(atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
 atomic_test_bit(&srv->flags, FLAG_MANUAL)))) ||
```

This I think is more readable:
```
/* Auto Occupancy is disabled */
!atomic_test_bit(&srv->flags, FLAG_OCC_MODE) ||
/* Fade Standby Manual state */
(srv->state == LIGHT_CTRL_STATE_STANDBY &&
 atomic_test_bit(&srv->flags, FLAG_TRANSITION) && 
 atomic_test_bit(&srv->flags, FLAG_MANUAL)) ||
/* Fade On state */
(srv->state == LIGHT_CTRL_STATE_ON &&
 atomic_test_bit(&srv->flags, FLAG_TRANSITION))) {
```

subsys/bluetooth/mesh/light_ctrl_srv.c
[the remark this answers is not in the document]
Balaklaka: I also see that the regulator is started after the "measured" value is set. In one way I want to keep this as it is, to show that the regulator measured value is set to 0 when powered on and provisioned. I think this is more according to spec than. If you both agree that this should be removed, I have not problem removing this lines.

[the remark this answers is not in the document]
PavelVPV: We already have a test for this so I don't think this code is needed: https://github.com/nrfconnect/sdk-nrf/blob/e28de89573274076beae9d0afc558288383be87b/tests/subsys/bluetooth/mesh/light_ctrl/src/main.c#L1059-L1060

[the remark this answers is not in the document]
PavelVPV: Sorry, no, it is not this one, but anyway. We could write one more test to cover this, but I don't think this is needed. This code won't do anything anyway, so it is better to keep the code clean.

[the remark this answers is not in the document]
Balaklaka: What do you think @omkar3141? If you agree I can remove it.

subsys/bluetooth/mesh/light_ctrl_srv.c
```diff
@@ -999,13 +999,28 @@ static int handle_sensor_status(struct bt_mesh_model *model, struct bt_mesh_msg_
...
+			 * the Occupancy On event should not be processed.
+			 */
+			  (atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
+			   atomic_test_bit(&srv->flags, FLAG_MANUAL)))) ||
+		    /* According to the Mesh Model Specification section 6.2.5.7
+		     * specifications FADE ON state, the Occupancy On event
+		     * should not be processed.
+		     */
+		    (srv->state == LIGHT_CTRL_STATE_ON &&
+		     atomic_test_bit(&srv->flags, FLAG_TRANSITION))) {
```
omkar3141, line 1023: Looks fine now.

omkar3141 (approved):

PavelVPV (approved):

subsys/bluetooth/mesh/light_ctrl_srv.c
```diff
@@ -999,13 +999,28 @@ static int handle_sensor_status(struct bt_mesh_model *model, struct bt_mesh_msg_
...
-		/* OCC_MODE must be enabled for the occupancy sensors to be
-		 * able to turn on the light:
-		 */
-		if (srv->state == LIGHT_CTRL_STATE_STANDBY &&
-		    !atomic_test_bit(&srv->flags, FLAG_OCC_MODE) &&
-		    !(atomic_test_bit(&srv->flags, FLAG_TRANSITION) &&
-		      !atomic_test_bit(&srv->flags, FLAG_MANUAL))) {
+		if ((srv->state == LIGHT_CTRL_STATE_STANDBY &&
+			/* According to the Mesh Model Specification section
+			 * 6.2.5.6 specifications STANDBY state, and the Auto
```
ludvigsj, line 1004:
```suggestion
			 * 6.2.5.6 about the STANDBY state, if the Auto
```

ludvigsj: Same below

ludvigsj (changes requested):

subsys/bluetooth/mesh/light_ctrl_srv.c
Balaklaka: I do not agree in this change, because it is importent to reference to the specifications STANDBY state and not the STANDBY state in the code.

ludvigsj: The sentence as it stands makes no sense. What is the "and" related to? The "and" makes no sense. What are you trying to say? "According to section 6.2.5.6: when you are in the specifications STANDBY state and the Auto Occupancy is off, the event should not be processed"?

Balaklaka: Yes, you have understood it correctly.

ludvigsj: Then:
```suggestion
			 * 6.2.5.6: When in the specifications STANDBY state, and the Auto
```

Balaklaka: This is becoming nitpicking, but I will add "When in the".

ludvigsj (approved):

PR #12052, opened 2023-08-16
boards: arm: thingy91: Secure boot

gregersrygg:
This PR fixes some problems that appears when trying to build for thingy:91 with `CONFIG_SECURE_BOOT` enabled:
- ADP536X needed to be excluded from the secure bootloader. Created a new kconfig symbol `IS_BOOTLOADER` to make it easier to check for both mcuboot and the secure bootloader at the same time.
- The thingy:91 board CMakeLists.txt enforced a static partition layout that did not include partitions for the secure bootloader and an extra slot for updating mcuboot. Added new kconfig options to allow the user to select what partition layout to use.
- Not directly related, but also added a kconfig option (`CONFIG_THINGY91_PARTITION_MANAGER`) to not inject a thingy:91 board static configuration for those that want to use dynamic or an application specific static partition layout.

This does not solve every problem for using secure boot on thingy:91, but is a couple of steps in the right direction. Further steps are needed to get it working properly, but I want to reduce the scope of this PR and delegate some of the remaining work to other teams. Marked the new partitions layouts as experimental to make it clear that this is not production ready and under development, as described in [software maturity levels](https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/releases_and_maturity/software_maturity.html#software-maturity-levels).

Known issues:
- After updating mcuboot using nRF Programmer, the boot process halts after the immutable bootloader.
- Compiling with `CONFIG_THINGY91_PARTITION_MANAGER=y` and `CONFIG_SECURE_BOOT=n` will fail because the dynamic `app` partition is not aligned according to SPU flash region size.

CIA-604

### DIFF
--- a/boards/arm/thingy91_nrf9160/CMakeLists.txt
+++ b/boards/arm/thingy91_nrf9160/CMakeLists.txt
@@ -4,19 +4,10 @@
 #
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 
-if(CONFIG_BOARD_THINGY91_NRF9160 AND NOT DEFINED CONFIG_MCUBOOT)
-	zephyr_library()
-	zephyr_library_sources(adp5360_init.c)
-endif()
+zephyr_library()
+zephyr_library_sources_ifdef(CONFIG_ADP536X adp5360_init.c)
 
 if(CONFIG_BOARD_THINGY91_NRF9160_NS)
-	# If TF-M is used, the ADP5360 configuration must be done in non-secure
-	# as we can't instruct TF-M to run it.
-	if(CONFIG_BUILD_WITH_TFM)
-		zephyr_library()
-		zephyr_library_sources(adp5360_init.c)
-	endif()
-
 	# Use static partition layout to ensure the partition layout remains
 	# unchanged after DFU. This needs to be made globally available
 	# because it is used in other CMake files.
--- a/boards/arm/thingy91_nrf9160/CMakeLists.txt
+++ b/boards/arm/thingy91_nrf9160/CMakeLists.txt
@@ -7,15 +7,16 @@
 zephyr_library()
 zephyr_library_sources_ifdef(CONFIG_ADP536X adp5360_init.c)
 
-if(CONFIG_BOARD_THINGY91_NRF9160_NS)
-	# Use static partition layout to ensure the partition layout remains
-	# unchanged after DFU. This needs to be made globally available
-	# because it is used in other CMake files.
-	if(CONFIG_LWM2M_CARRIER)
-		set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static_lwm2m_carrier.yml CACHE INTERNAL "")
-	else()
-		set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static.yml CACHE INTERNAL "")
-	endif()
+if(CONFIG_THINGY91_STATIC_PARTITIONS_FACTORY)
+  set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static.yml CACHE INTERNAL "")
+endif()
+
+if(CONFIG_THINGY91_STATIC_PARTITIONS_SECURE_BOOT)
+  set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static_secure_boot.yml CACHE INTERNAL "")
+endif()
+
+if(CONFIG_THINGY91_STATIC_PARTITIONS_LWM2M_CARRIER)
+  set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static_lwm2m_carrier.yml CACHE INTERNAL "")
 endif()
 
 zephyr_library_sources(nrf52840_reset.c)
--- a/boards/arm/thingy91_nrf9160/Kconfig.board
+++ b/boards/arm/thingy91_nrf9160/Kconfig.board
@@ -12,4 +12,38 @@ config BOARD_THINGY91_NRF9160
 config BOARD_THINGY91_NRF9160_NS
 	bool "nRF9160 THINGY91 non-secure"
 
+if BOARD_THINGY91_NRF9160_NS
+
+choice
+	prompt "Thingy:91 partition layout"
+	default THINGY91_STATIC_PARTITIONS_SECURE_BOOT if SECURE_BOOT
+	default THINGY91_STATIC_PARTITIONS_LWM2M_CARRIER if LWM2M_CARRIER
+	default THINGY91_STATIC_PARTITIONS_FACTORY
+
+config THINGY91_STATIC_PARTITIONS_FACTORY
+	bool "Factory Thingy:91 partition layout"
+	help
+	   The default Thingy:91 partition layout used in the factory firmware. This ensures
+	   firmware updates are compatible with Thingy:91 when flashing firmware over USB or over
+	   the air.
+
+config THINGY91_STATIC_PARTITIONS_SECURE_BOOT
+	bool "Secure boot Thingy:91 partition layout [EXPERIMENTAL]"
+	depends on SECURE_BOOT
+	select EXPERIMENTAL
+	help
+	   Similar to the factory partition layout, but also has space for the Immutable Bootloader
+	   and two MCUboot slots. A debugger is needed to flash Thingy:91 the first time.
+	   This layout is still under development and should not be used in production.
+
+config THINGY91_STATIC_PARTITIONS_LWM2M_CARRIER
+	bool "LWM2M Carrier  partition layout"
+	depends on LWM2M_CARRIER
+	help
+	  Use a partition layout including a storage partition needed for the lwm2m carrier library.
+
+endchoice
+
+endif # BOARD_THINGY91_NRF9160_NS
+
 endif # SOC_NRF9160_SICA
--- a/boards/arm/thingy91_nrf9160/Kconfig.board
+++ b/boards/arm/thingy91_nrf9160/Kconfig.board
@@ -15,7 +15,7 @@ config BOARD_THINGY91_NRF9160_NS
 if BOARD_THINGY91_NRF9160_NS
 
 choice
-	prompt "Thingy:91 partition layout"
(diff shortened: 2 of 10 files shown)


boards/arm/thingy91_nrf9160/CMakeLists.txt
```diff
@@ -7,15 +7,16 @@
...
-	# Use static partition layout to ensure the partition layout remains
-	# unchanged after DFU. This needs to be made globally available
-	# because it is used in other CMake files.
-	if(CONFIG_LWM2M_CARRIER)
-		set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static_lwm2m_carrier.yml CACHE INTERNAL "")
-	else()
-		set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static.yml CACHE INTERNAL "")
-	endif()
+if(CONFIG_THINGY91_STATIC_PARTITIONS_FACTORY)
+	set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static.yml CACHE INTERNAL "")
```
tejlmand, line 11: we don't use tabs in CMake files, please convert to two spaces.
```suggestion
  set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static.yml CACHE INTERNAL "")
```

boards/arm/thingy91_nrf9160/CMakeLists.txt
```diff
@@ -7,15 +7,16 @@
...
+endif()
+
+if(CONFIG_THINGY91_STATIC_PARTITIONS_SECURE_BOOT)
+	set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static_secure_boot.yml CACHE INTERNAL "")
```
tejlmand, line 15:
```suggestion
  set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static_secure_boot.yml CACHE INTERNAL "")
```

boards/arm/thingy91_nrf9160/CMakeLists.txt
```diff
@@ -7,15 +7,16 @@
...
+endif()
+
+if(CONFIG_THINGY91_STATIC_PARTITIONS_LWM2M_CARRIER)
+	set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static_lwm2m_carrier.yml CACHE INTERNAL "")
```
tejlmand, line 19:
```suggestion
  set(PM_STATIC_YML_FILE ${CMAKE_CURRENT_LIST_DIR}/thingy91_pm_static_lwm2m_carrier.yml CACHE INTERNAL "")
```

boards/arm/thingy91_nrf9160/Kconfig.board
```diff
@@ -12,4 +12,36 @@ config BOARD_THINGY91_NRF9160
...
+	default THINGY91_STATIC_PARTITIONS_FACTORY
+
+config THINGY91_STATIC_PARTITIONS_FACTORY
+	bool "Factory Thingy:91 partition layout"
+	help
+	   The default Thingy:91 partition layout used in the factory firmware. This ensures
+	   firmware updates are compatible with Thingy:91 when flashing firmware over USB or over
+	   the air.
+
+config THINGY91_STATIC_PARTITIONS_SECURE_BOOT
```
tejlmand, line 30: question, should this has a `depends on SECURE_BOOT` ?
Cause otherwise the choice will be sticky after first CMake invocation, in such a way that if someone does an initial build and later uses menuconfig to enable / disable secure boot, then the partition layout choice will not change.

Thingy is a bit special, so the current behavior (stickiness) might be what you want.

boards/arm/thingy91_nrf9160/Kconfig.board
```diff
@@ -12,4 +12,36 @@ config BOARD_THINGY91_NRF9160
...
+	bool "Secure boot Thingy:91 partition layout [EXPERIMENTAL]"
+	select EXPERIMENTAL
+	help
+	   Similar to the factory partition layout, but also has space for the Immutable Bootloader
+	   and two MCUboot slots. A debugger is needed to flash Thingy:91 the first time.
+	   This layout is still under development and should not be used in production.
+
+config THINGY91_STATIC_PARTITIONS_LWM2M_CARRIER
```
tejlmand, line 38: similar question as the secuboot partition choice.

boards/arm/thingy91_nrf9160/Kconfig.board
```diff
@@ -40,6 +40,15 @@ config THINGY91_STATIC_PARTITIONS_LWM2M_CARRIER
 	help
 	  Use a partition layout including a storage partition needed for the lwm2m carrier library.
 
+config THINGY91_PARTITION_MANAGER
+	bool "Default partition manager logic [EXPERIMENTAL]"
```
tejlmand, line 44: Not sure default is the best wording here.

Maybe the word `dynamic` is better, as that is what is used in the PM guide: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/scripts/partition_manager/partition_manager.html

```suggestion
	bool "Dynamic partition manager logic [EXPERIMENTAL]"
```

boards/arm/thingy91_nrf9160/Kconfig.board
```diff
@@ -40,6 +40,15 @@ config THINGY91_STATIC_PARTITIONS_LWM2M_CARRIER
...
+	select EXPERIMENTAL
+	help
+	   Don't enforce static partition layout, but use the default Partition Manager logic to
```
tejlmand, line 47:
```suggestion
	   Don't enforce static partition layout, but use the dynamic Partition Manager logic to
```

tejlmand (changes requested):
some minor nits.

boards/arm/thingy91_nrf9160/Kconfig.board
gregersrygg: I struggled how to best formulate this. Dynamic is not precise, because it's still possible to define static partitions from the application if they want to. The PM_STATIC_YML_FILE define we use in the CMakeLists.txt has the top level priority in partition manager logic - which makes it impossible to define an application static partition. There seems to be no way to make a board static partition which is possible to override from the application. The THINGY91_PARTITION_MANAGER config is used to so it's possible to not force one of the board static partitions.

plskeggs (approved):

gregersrygg:
@tejlmand can you please have another look?

boards/arm/thingy91_nrf9160/Kconfig.board
tejlmand: I see what you mean, and upon second look I agree that `Dynamic` is not the best word either.

I still dislike the use `default` in this context.
As a user, it gives me the impression that I can go lookup somewhere to see what the default layout will be, which is not the case.
Here default doesn't refer to a given layout (as opposed to the choice options above).

When looking at this again, and the `choice` title:
https://github.com/nrfconnect/sdk-nrf/blob/154f3c80854f4416a0c8cef8e1d9307d9799c51b/boards/arm/thingy91_nrf9160/Kconfig.board#L18

How about changing that title to:
```
Fixed Thingy:91 partition layout
```
or 
```
Pre-defined Thingy:91 partition layout
```
Then the final choice could then be: `None`, and look as:
```
config THINGY91_PARTITION_MANAGER
	bool "None [EXPERIMENTAL]"
	select EXPERIMENTAL
	help
	   Don't use a pre-defined static partition layout, but use the default Partition Manager logic to
	   handle the partition layout. This allows the application to use dynamic layout or define
	   a custom static partition layout for the application. A debugger is needed to flash
	   Thingy:91 with a different partition layout.
```

And the menu, seen by the user will look:
```
Pre-defined Thingy:91 partition layout

  [X] "Factory Thingy:91 partition layout"
  [ ] "Secure boot Thingy:91 partition layout [EXPERIMENTAL]"
  [ ] "LWM2M Carrier partition layout"
  [ ] "None [EXPERIMENTAL]"
```

an alternative to `None` could be the word `auto`.

gregersrygg: Thanks! That sounds like a good way to solve it. I also renamed the kconfig option itself from `THINGY91_PARTITION_MANAGER` to `THINGY91_NO_PREDEFINED_LAYOUT`.

boards/arm/thingy91_nrf9160/Kconfig.board
gregersrygg: When looking at this again now, I don't think there's any point in mentioning Partition Manager here at all. So have updated the help text to: _«Disable pre-defined static partition layout. This allows the application to use dynamic layout or define a custom static partition layout for the application...»_

doc/nrf/device_guides/working_with_nrf/nrf91/thingy91.rst
```diff
@@ -634,3 +634,28 @@ Building and programming on the command line
...
+Configure the partition layout using one of the following configuration options:
+
+* :kconfig:option:`CONFIG_THINGY91_STATIC_PARTITIONS_FACTORY` - This option is the default Thingy:91 partition layout used in the factory firmware.
+  This ensures firmware updates are compatible with Thingy:91 when programming firmware through USB.
+* :kconfig:option:`CONFIG_THINGY91_STATIC_PARTITIONS_SECURE_BOOT` - This option is similar to the factory partition layout, but also has space for the immutable bootloader and two MCUboot slots.
+  A debugger is needed to program Thingy:91 for the first time.
+  This is an :ref:`experimental <software_maturity>` feature.
+* :kconfig:option:`CONFIG_THINGY91_STATIC_PARTITIONS_LWM2M_CARRIER` - This option uses a partition layout, including a storage partition needed for the :ref:`liblwm2m_carrier_readme` library.
+* :kconfig:option:`CONFIG_THINGY91_NO_PREDEFINED_LAYOUT` - Enabling this option disables Thingy:91 pre-defined static partitions.
+  This allows the application to use dynamic layout or define a custom static partition layout for the application.
```
divipillai, line 659:
```suggestion
  This allows the application to use a dynamic layout or define a custom static partition layout for the application.
```

tejlmand (approved):

divipillai (approved):

plskeggs:
Great solution @gregersrygg.

PR #12054, opened 2023-08-16
doc: Bluetooth: Mesh: Move SW maturity up to protocol

Balaklaka:
The Bluetooth mesh is now put as a feature under Bluetooth, but should be a seperate section under protocol to be more visible. Move the Bluetooth mesh software maturity to the protocol table.


omkar3141 (approved):

mia-ko (approved):
The move to protocols table seems ok to me, and the change from Bluetooth Mesh to Bluetooth mesh.
I'd also like to see a feature support section and table for Bluetooth mesh, so I'm hoping you are getting some input from @emidttun on what info should be going there for mesh.

alxelax (approved):

PR #12068, opened 2023-08-18
applications: serial_lte_modem: Use cached A-GPS data

junqingzou:
After a reset modem will send NRF_MODEM_GNSS_EVT_AGPS_REQ. If there is valid A-GPS data cached on modem side, app could inject current GPS time before starting GNSS, so that modem would not request new almanac and ephemerides, reducing data size for cloud downloading.

Sample log of cache data reused:
 AGPS_REQ.sv_mask_ephe = 0x00000000 instead of 0xffffffff
 AGPS_REQ.sv_mask_alm  = 0x00000000 instead of 0xffffffff
 AGPS_REQ.data_flags   = 0x00000033

Further note:
 A-GPS data is cached periodically or upon +CFUN=0.
 Cached data in RAM can be purged by #XGPSDEL=511.

### DIFF
--- a/applications/serial_lte_modem/src/gnss/slm_at_gnss.c
+++ b/applications/serial_lte_modem/src/gnss/slm_at_gnss.c
@@ -23,6 +23,16 @@ LOG_MODULE_REGISTER(slm_gnss, CONFIG_SLM_LOG_LEVEL);
 
 #define LOCATION_REPORT_MS 5000
 
+#define SEC_PER_MIN	(60UL)
+#define MIN_PER_HOUR	(60UL)
+#define SEC_PER_HOUR	(MIN_PER_HOUR * SEC_PER_MIN)
+#define HOURS_PER_DAY	(24UL)
+#define SEC_PER_DAY	(HOURS_PER_DAY * SEC_PER_HOUR)
+/* (6.1.1980 UTC - 1.1.1970 UTC) */
+#define GPS_TO_UNIX_UTC_OFFSET_SECONDS	(315964800UL)
+/* UTC/GPS time offset as of 1st of January 2017. */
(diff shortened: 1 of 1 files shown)


tokangas:
> Further note: A-GPS data is cached upon +CFUN=0.
Cached data can be purged by #XGPSDEL=511.

`nrf_modem_gnss_nv_data_delete()` clears the GNSS data from RAM. The changes gets written to flash in the same way as usual, i.e. either periodically or when `+CFUN=0` is given.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -613,6 +639,28 @@ int handle_at_agps(enum at_cmd_type cmd_type)
 				}
 			} /* else leave it to default or previously configured timeout */
 
+			/* inject current GPS time. If there is cached A-GPS data before expiry,
```
tokangas, line 642:
```suggestion
			/* Inject current GPS time. If there is cached A-GPS data before expiry,
```

tokangas (approved):

SeppoTakalo (approved):

junqingzou:
> > Further note: A-GPS data is cached upon +CFUN=0.
> > Cached data can be purged by #XGPSDEL=511.
> 
> `nrf_modem_gnss_nv_data_delete()` clears the GNSS data from RAM. The changes gets written to flash in the same way as usual, i.e. either periodically or when `+CFUN=0` is given.

Thanks for the clarification.
The behavior is, when cached data is "deleted", modem request alm/ephe for all 32 PRNs again.

MarkusLassila (approved):

PR #12071, opened 2023-08-18
net: lib: aws_fota: Determine protocol from job document

fnawratil:
For FOTA downloads using the aws_fota library, the protocol is now determined from the job document (protocol field or url). If no security tag is configured and the protocol is set to https, the job is marked as failed.

### DIFF
--- a/doc/nrf/libraries/networking/aws_fota.rst
+++ b/doc/nrf/libraries/networking/aws_fota.rst
@@ -7,14 +7,14 @@ AWS FOTA
    :local:
    :depth: 2
 
-The Amazon Web Services firmware over-the-air (AWS FOTA) library combines the :ref:`lib_aws_jobs` and :ref:`lib_fota_download` libraries into a library that can perform an over-the-air firmware update using HTTP and MQTT TLS.
+The Amazon Web Services firmware over-the-air (AWS FOTA) library combines the :ref:`lib_aws_jobs` and :ref:`lib_fota_download` libraries into a library that can perform an over-the-air firmware update using HTTP and HTTPS.
 
 It connects to the specified broker using the existing or given certificates and uses `TLS`_ for the MQTT connection.
 This means that the data sent in each MQTT message is encrypted.
 
 When an update is available, the library receives a notification that contains metadata about the update.
 This metadata contains the location of the uploaded firmware image.
-The library then uses HTTP to download the firmware image, replacing the current firmware with the downloaded firmware.
+The library then uses HTTP or HTTPS to download the firmware image, replacing the current firmware with the downloaded firmware.
 After the new firmware image is downloaded, the application must perform a reboot to run the new image and complete the FOTA process.
 The library supports the updating of delta images for both the application and modem firmware.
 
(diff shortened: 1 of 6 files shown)


doc/nrf/libraries/networking/aws_fota.rst
```diff
@@ -14,7 +14,7 @@ This means that the data sent in each MQTT message is encrypted.
 
 When an update is available, the library receives a notification that contains metadata about the update.
```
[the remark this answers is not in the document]
fnawratil: Added HTTPS and removed the MQTT TLS (since it's misleading) on line 10.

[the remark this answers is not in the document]
coderbyheart: Then line 13 no longer makes sense: https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/device_guides/working_with_nrf/nrf91/nrf9160_gs.html

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -60,6 +60,7 @@ static uint8_t get_topic[AWS_JOBS_TOPIC_MAX_LEN];
 
 /* Allocated buffers for keeping hostname, json payload and file_path. */
 static uint8_t payload_buf[CONFIG_AWS_FOTA_PAYLOAD_SIZE];
+static uint8_t protocol[10];
```
simensrostad, line 63:
```suggestion
static uint8_t protocol[sizeof("https")];
```

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -366,9 +369,17 @@ static int job_update_accepted(struct mqtt_client *const client, uint32_t payloa
 		LOG_DBG("Start downloading firmware from %s/%s",
 			(char *)hostname, (char *)file_path);
 
-#if defined(CONFIG_AWS_FOTA_DOWNLOAD_SECURITY_TAG)
-		sec_tag = CONFIG_AWS_FOTA_DOWNLOAD_SECURITY_TAG;
+		/* If the protocol is https set the sec_tag. This instructs the download client
+		 * to use https. Fail if the protocol is https but no sec tag is configured.
+		 */
+		if (strncmp(protocol, "https", 5) == 0) {
+#if defined(CONFIG_AWS_FOTA_DOWNLOAD_SECURITY_TAG) && CONFIG_AWS_FOTA_DOWNLOAD_SECURITY_TAG != -1
```
simensrostad, line 376:
```suggestion
if (IS_ENABLED(CONFIG_AWS_FOTA_DOWNLOAD_SECURITY_TAG) && (CONFIG_AWS_FOTA_DOWNLOAD_SECURITY_TAG != -1)
```

simensrostad: If `IS_ENABLED` can be used instead of the preprocessor statement it would be better IMO. Preprocessor statements is not particulary readable.

fnawratil: Using preprocessor statements is the way this is currently achieved, so I reused that part.
`IS_ENABLED` can - as far as I understand - not be used, since this is not a boolean value but the integer for the security tag?

divipillai (approved):

subsys/net/lib/aws_fota/src/aws_fota.c
```diff
@@ -60,6 +60,7 @@ static uint8_t get_topic[AWS_JOBS_TOPIC_MAX_LEN];
 
 /* Allocated buffers for keeping hostname, json payload and file_path. */
 static uint8_t payload_buf[CONFIG_AWS_FOTA_PAYLOAD_SIZE];
+static uint8_t protocol[sizeof("https://")];
```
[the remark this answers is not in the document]
fnawratil: Might make sense to use an existing type for that, since we already depend on e.g. `http_parser_url`. 
I'll look into that separately.

PR #12072, opened 2023-08-18
applications: nrf_desktop: Cleanup BLE scan module implementation

MarekPieta:
Cleanup BLE scan module implementation before introducing new features.

Jira: NCSDK-20843

### DIFF
--- a/applications/nrf_desktop/src/modules/ble_scan.c
+++ b/applications/nrf_desktop/src/modules/ble_scan.c
@@ -8,6 +8,7 @@
 
 #include <zephyr/bluetooth/bluetooth.h>
 #include <bluetooth/scan.h>
+#include <bluetooth/gatt_dm.h>
 #include <zephyr/settings/settings.h>
 
 #include <string.h>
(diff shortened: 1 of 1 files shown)


applications/nrf_desktop/src/modules/ble_scan.c
```diff
@@ -591,13 +576,6 @@ static bool handle_ble_peer_operation_event(const struct ble_peer_operation_even
 	case PEER_OPERATION_ERASED:
 		reset_subscribers();
 		store_subscribed_peers();
-		if (count_conn() == CONFIG_BT_MAX_CONN) {
-			if (IS_ENABLED(CONFIG_BT_SCAN_CONN_ATTEMPTS_FILTER)) {
-				bt_scan_conn_attempts_filter_clear();
-			}
-			peers_only = false;
-			break;
-		}
```
alstrzebonski: Is this no longer needed?

kapi-no: +1

kapi-no (approved):
Looking good (waiting for the answer for the @alstrzebonski question)

applications/nrf_desktop/src/modules/ble_scan.c
MarekPieta: There is no break, the fall-through causes executing the same code located below. The `scan_start` function would not start canning, because of internal conn count check.

kapi-no: well actually there is if the count_conn() is equal to CONFIG_BT_MAX_CONN. In that case, the scan_start is not called

MarekPieta: Yes, but if the `count_conn() == CONFIG_BT_MAX_CONN`, the `scan_start` would not start scanning anyway and behavior is actually unchanged.

kapi-no: okay, seems good to me. Thanks for explaining this simplification

I also appreciate the multi-commit structure of this PR which made it more review-friendly

alstrzebonski: Okay, thanks for the explanation.

PR #12074, opened 2023-08-18
applications: Matter Bridge: added BLE environmental sensor support

markaj-nordic:
* Added implementation of the BLE data provider which supports subscription to the humidity and temperature characteristics.
* implemented one provider to many Matter devices relation
* updated README

### DIFF
--- a/applications/matter_bridge/CMakeLists.txt
+++ b/applications/matter_bridge/CMakeLists.txt
@@ -44,7 +44,7 @@ target_sources(app PRIVATE
     src/main.cpp
     src/bridge_shell.cpp
     ${COMMON_ROOT}/src/bridge/bridge_manager.cpp
-    ${COMMON_ROOT}/src/bridge/bridged_device.cpp
+    ${COMMON_ROOT}/src/bridge/matter_bridged_device.cpp
     ${COMMON_ROOT}/src/bridge/bridge_storage_manager.cpp
     ${COMMON_ROOT}/src/bridge/bridged_device_data_provider.cpp
     src/zap-generated/IMClusterCommandHandler.cpp
@@ -62,6 +62,25 @@ if(CONFIG_BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE)
         src/bridged_device_types/ble_onoff_light_data_provider.cpp
     )
 endif() # CONFIG_BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE
+
+if(CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE)
+    target_sources(app PRIVATE
+        src/bridged_device_types/humidity_sensor.cpp
+    )
+endif() # CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE
+
+if(CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE)
+    target_sources(app PRIVATE
+        src/bridged_device_types/temperature_sensor.cpp
+    )
+endif() # CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE
+
+# Assume it makes no sense to support BLE environmental sensor without all Matter counterparts
+if(CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE AND CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE)
+    target_sources(app PRIVATE
+        src/bridged_device_types/ble_environmental_data_provider.cpp
+    )
+endif() # CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE AND CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE
 
 else()
 if(CONFIG_BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE)
--- a/applications/matter_bridge/Kconfig
+++ b/applications/matter_bridge/Kconfig
@@ -11,11 +11,11 @@ config BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE
 
 config BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE
 	bool "Support for Temperature Sensor bridged device"
-	default y if BRIDGED_DEVICE_SIMULATED
+	default y
 
 config BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE
 	bool "Support for Humidity Sensor bridged device"
-	default y if BRIDGED_DEVICE_SIMULATED
+	default y
 
 choice BRIDGED_DEVICE_IMPLEMENTATION
 	prompt "Bridged Device implementation"
@@ -53,13 +53,17 @@ config BT_SCAN_FILTER_ENABLE
 
 # Configure how many Bluetooth LE service UUIDs the Matter bridge can scan.
 config BT_SCAN_UUID_CNT
-	default 1
+	default 2
 
 config BT_GATT_CLIENT
 	default y
 
 config BT_GATT_DM
 	default y
+
+config BRIDGE_BLE_DEVICE_POLLING_INTERVAL
+	int "BLE humidity measurement readout polling interval in ms"
+	default 3000
 
 endif
 
--- a/applications/matter_bridge/README.rst
+++ b/applications/matter_bridge/README.rst
@@ -23,8 +23,12 @@ The application supports the following development kits:
 
 To test the Matter bridge application with the :ref:`Bluetooth LE bridged device <matter_bridge_app_bridged_support>`, you also need the following:
 
-* An additional development kit compatible with the :ref:`peripheral_lbs` sample.
-* A micro USB cable to connect the development kit to the PC.
+* An additional development kit compatible with one of the following Bluetooth LE samples:
+
+  * :ref:`peripheral_lbs`
+  * :ref:`peripheral_esp`
+
+* A micro-USB cable to connect the development kit to the PC.
 
 To commission the Matter bridge device and control it remotely through a Wi-Fi network, you also need a Matter controller device :ref:`configured on PC or smartphone <ug_matter_configuring>`.
 This requires additional hardware depending on your setup.
@@ -182,6 +186,7 @@ Adding a Bluetooth LE bridged device to the Matter bridge
      The argument is mandatory and accepts the following values:
 
      * ``256`` - On/Off Light.
+     * ``291`` - Environmental Sensor (a combination of Temperature Sensor and Humidity Sensor).
 
    * *<ble_device_index>* is the Bluetooth LE device index on the list returned by the ``scan`` command.
      The argument is mandatory and accepts only the values returned by the ``scan`` command.
@@ -329,8 +334,12 @@ After building the sample and programming it to your development kit, complete t
 
       .. group-tab:: Testing with Bluetooth LE bridged devices
 
-         a. Build and program the :ref:`peripheral_lbs` sample to an additional development kit compatible with the sample.
-         #. Connect the development kit that is running the :ref:`peripheral_lbs` sample to the PC.
+         a. Build and program the one of the following Bluetooth LE samples to an additional development kit compatible with the sample:
+
+            * :ref:`peripheral_lbs`
+            * :ref:`peripheral_esp`
+
+         #. Connect the development kit that is running the Bluetooth LE sample to the PC.
          #. Using the terminal emulator connected to the bridge, run the following :ref:`Matter CLI command <matter_bridge_cli>` to scan for available Bluetooth LE devices:
 
             .. code-block:: console
@@ -373,6 +382,8 @@ After building the sample and programming it to your development kit, complete t
                I: Adding OnOff Light bridged device
                I: Added device to dynamic endpoint 3 (index=0)
 
+            For the Environmental Sensor, two endpoints are created: one implements the Temperature Sensor, and the other implements the Humidity Sensor.
+
          #. Write down the value for the bridged device dynamic endpoint ID.
             This is going to be used in the next steps (*<bridged_device_endpoint_ID>*).
          #. Use the :doc:`CHIP Tool <matter:chip_tool_guide>` to read the value of an attribute from the bridged device endpoint.
@@ -382,6 +393,22 @@ After building the sample and programming it to your development kit, complete t
                :class: highlight
 
                ./chip-tool onoff read on-off *<bridge_node_ID>* *<bridged_device_endpoint_ID>*
+
+            In case of the Environmental Sensor, the current temperature and humidity measurements forwarded by the Bluetooth LE Environmental Sensor can be read as follows:
+
+              * temperature:
+
+                .. parsed-literal::
+                   :class: highlight
+
+                   ./chip-tool temperaturemeasurement read measured-value *<bridge_node_ID>* *<bridged_device_endpoint_ID>*
+
+              * humidity:
+
+                .. parsed-literal::
+                   :class: highlight
+
+                    ./chip-tool relativehumiditymeasurement read measured-value *<bridge_node_ID>* *<bridged_device_endpoint_ID>*
 
 Enabling remote control
 =======================
--- a/applications/matter_bridge/src/bridge_shell.cpp
+++ b/applications/matter_bridge/src/bridge_shell.cpp
@@ -21,7 +21,7 @@ LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
 static BridgedDeviceDataProvider *CreateSimulatedProvider(int deviceType)
 {
 	return BridgeFactory::GetSimulatedDataProviderFactory().Create(
-		static_cast<BridgedDevice::DeviceType>(deviceType), BridgeManager::HandleUpdate);
+		static_cast<MatterBridgedDevice::DeviceType>(deviceType), BridgeManager::HandleUpdate);
 }
 #endif /* CONFIG_BRIDGED_DEVICE_SIMULATED */
 
@@ -29,14 +29,14 @@ static BridgedDeviceDataProvider *CreateSimulatedProvider(int deviceType)
 
 static BridgedDeviceDataProvider *CreateBleProvider(int deviceType)
 {
-	return BridgeFactory::GetBleDataProviderFactory().Create(static_cast<BridgedDevice::DeviceType>(deviceType),
-								 BridgeManager::HandleUpdate);
+	return BridgeFactory::GetBleDataProviderFactory().Create(
+		static_cast<MatterBridgedDevice::DeviceType>(deviceType), BridgeManager::HandleUpdate);
 }
 
 struct BluetoothConnectionContext {
 	const struct shell *shell;
 	int deviceType;
-	char nodeLabel[BridgedDevice::kNodeLabelSize];
+	char nodeLabel[MatterBridgedDevice::kNodeLabelSize];
 	BLEBridgedDeviceProvider *provider;
 };
 #endif /* CONFIG_BRIDGED_DEVICE_BT */
@@ -44,21 +44,43 @@ struct BluetoothConnectionContext {
 static void AddDevice(const struct shell *shell, int deviceType, const char *nodeLabel,
 		      BridgedDeviceDataProvider *provider)
 {
-	VerifyOrReturn(provider != nullptr,
-		       shell_fprintf(shell, SHELL_INFO, "Cannot allocate data provider of given type\n"));
+	VerifyOrReturn(provider != nullptr, shell_fprintf(shell, SHELL_INFO, "No valid data provider!\n"));
 
-	auto *newBridgedDevice = BridgeFactory::GetBridgedDeviceFactory().Create(
-		static_cast<BridgedDevice::DeviceType>(deviceType), nodeLabel);
+	CHIP_ERROR err;
+	if (deviceType == MatterBridgedDevice::DeviceType::EnvironmentalSensor) {
+		/* Handle special case for environmental sensor */
+		auto *temperatureBridgedDevice = BridgeFactory::GetBridgedDeviceFactory().Create(
+			MatterBridgedDevice::DeviceType::TemperatureSensor, nodeLabel);
 
-	VerifyOrReturn(newBridgedDevice != nullptr, delete provider,
-		       shell_fprintf(shell, SHELL_INFO, "Cannot allocate Matter device of given type\n"));
+		VerifyOrReturn(temperatureBridgedDevice != nullptr, delete provider,
+			       shell_fprintf(shell, SHELL_INFO, "Cannot allocate Matter device of given type\n"));
 
-	CHIP_ERROR err = BridgeManager::Instance().AddBridgedDevices(newBridgedDevice, provider);
+		auto *humidityBridgedDevice = BridgeFactory::GetBridgedDeviceFactory().Create(
+			MatterBridgedDevice::DeviceType::HumiditySensor, nodeLabel);
+
+		VerifyOrReturn(humidityBridgedDevice != nullptr, delete provider, delete temperatureBridgedDevice,
+			       shell_fprintf(shell, SHELL_INFO, "Cannot allocate Matter device of given type\n"));
+
+		MatterBridgedDevice *newBridgedDevices[] = { temperatureBridgedDevice, humidityBridgedDevice };
+		err = BridgeManager::Instance().AddBridgedDevices(newBridgedDevices, provider,
+								  ARRAY_SIZE(newBridgedDevices));
+
+	} else {
+		auto *newBridgedDevice = BridgeFactory::GetBridgedDeviceFactory().Create(
+			static_cast<MatterBridgedDevice::DeviceType>(deviceType), nodeLabel);
+
+		MatterBridgedDevice *newBridgedDevices[] = { newBridgedDevice };
+		VerifyOrReturn(newBridgedDevice != nullptr, delete provider,
+			       shell_fprintf(shell, SHELL_INFO, "Cannot allocate Matter device of given type\n"));
+		err = BridgeManager::Instance().AddBridgedDevices(newBridgedDevices, provider,
+								  ARRAY_SIZE(newBridgedDevices));
+	}
+
 	if (err == CHIP_NO_ERROR) {
 		shell_fprintf(shell, SHELL_INFO, "Done\n");
 	} else if (err == CHIP_ERROR_INVALID_STRING_LENGTH) {
 		shell_fprintf(shell, SHELL_ERROR, "Error: too long node label (max %d)\n",
-			      BridgedDevice::kNodeLabelSize);
+			      MatterBridgedDevice::kNodeLabelSize);
 	} else if (err == CHIP_ERROR_NO_MEMORY) {
 		shell_fprintf(shell, SHELL_ERROR, "Error: no memory\n");
 	} else if (err == CHIP_ERROR_INVALID_ARGUMENT) {
@@ -119,6 +141,7 @@ static int AddBridgedDeviceHandler(const struct shell *shell, size_t argc, char 
 	BridgedDeviceDataProvider *provider = CreateBleProvider(contextPtr->deviceType);
 
 	if (!provider) {
+		shell_fprintf(shell, SHELL_INFO, "Cannot allocate data provider of given type\n");
 		return -ENOMEM;
 	}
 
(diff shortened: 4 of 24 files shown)


samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -140,13 +152,13 @@ CHIP_ERROR BridgeManager::HandleWrite(uint16_t index, ClusterId clusterId,
...
 
-	auto &device = Instance().mDevicesMap[index]->mDevice;
+	auto *device = Instance().mDevicesMap[index].value1;
 	CHIP_ERROR err = device->HandleWrite(clusterId, attributeMetadata->attributeId, buffer);
 
 	/* After updating Matter BridgedDevice state, forward request to the non-Matter device. */
 	if (err == CHIP_NO_ERROR) {
-		return Instance().mDevicesMap[index]->mProvider->UpdateState(clusterId, attributeMetadata->attributeId,
-									     buffer);
+		return Instance().mDevicesMap[index].value2->UpdateState(clusterId, attributeMetadata->attributeId,
```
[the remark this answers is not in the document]
markaj-nordic: I think, maybe we should concretize this map. I wanted to keep it generic for now to evaluate if it can be reused for `DeviceFactory` but maybe it's pointless. I will either do it, or add some aliases to make the code more readable.

samples/matter/common/src/bridge/bridged_device.h
```diff
@@ -43,6 +43,7 @@ class BridgedDevice {
 		OnOffLight = 0x0100,
 		TemperatureSensor = 0x0302,
 		HumiditySensor = 0x0307,
+		EnvironmentalSensor = 0x123
```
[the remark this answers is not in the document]
markaj-nordic: I agree it should be moved. Also, I hope we can remove it completely from the user interface once the KRKNWK-17561 is done.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -31,10 +31,12 @@ void BridgeManager::Init()
 				     false);
 }
 
-CHIP_ERROR BridgeManager::AddBridgedDevices(BridgedDevice *aDevice, BridgedDeviceDataProvider *aDataProvider)
+CHIP_ERROR BridgeManager::AddBridgedDevices(BridgedDevice *aDevices[], BridgedDeviceDataProvider *aDataProvider,
+					    uint8_t aDeviceListSize)
 {
+	VerifyOrReturnError(aDeviceListSize <= kMaxDevicesPerProvider, CHIP_ERROR_NO_MEMORY);
```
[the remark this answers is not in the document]
markaj-nordic: Maybe we don't, but we may end up with different number of DataProviders and MatterDevices, right?

applications/matter_bridge/README.rst
```diff
@@ -156,6 +156,7 @@ Adding a simulated bridged device to the Matter bridge
      The argument is mandatory and accepts the following values:
 
       * ``256`` - On/Off Light.
+      * ``291`` - Environmental Sensor (a composition of Temperature Sensor and Humidity Sensor).
```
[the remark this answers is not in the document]
markaj-nordic: This is a magic number just to distinguish the scenario where we have 1 BLE device - 2 Matter devices relation.
Yes, I think we should do something like that. But for that we would need to first recognize the BLE device by discovering its characteristics, or do you mean something else?

applications/matter_bridge/CMakeLists.txt
```diff
@@ -60,6 +60,20 @@ if(CONFIG_BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE)
...
+    target_sources(app PRIVATE
+        src/bridged_device_types/humidity_sensor.cpp
+        src/bridged_device_types/ble_environmental_data_provider.cpp
+    )
+endif() # CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE
+
+if(CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE)
+    target_sources(app PRIVATE
+        src/bridged_device_types/temperature_sensor.cpp
+        src/bridged_device_types/ble_environmental_data_provider.cpp
```
ArekBalysNordic, line 73: Is it ok to be added it twice when both temperature and humidity sensors are enabled?

applications/matter_bridge/README.rst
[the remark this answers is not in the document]
kkasperczyk-no: Yes, we would need to first do the GATT discovery.

ArekBalysNordic, line 159: Shouldn't it be moved below into the `Adding a Bluetooth LE bridged device to the Matter bridge` section?

applications/matter_bridge/Kconfig
```diff
@@ -53,7 +53,7 @@ config BT_SCAN_FILTER_ENABLE
 
 # Configure how many Bluetooth LE service UUIDs the Matter bridge can scan.
 config BT_SCAN_UUID_CNT
-	default 1
+	default 2
```
ArekBalysNordic, line 56: Why not 3 since we have temperature, humidity, and onoff uuids support? Unless the Environmental Sensor has one UUID, then it is correct.

applications/matter_bridge/src/bridged_device_factory.h
```diff
@@ -35,6 +35,12 @@
 #endif
 #endif
 
+#ifdef CONFIG_BRIDGED_DEVICE_BT
+#if defined(CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE) || defined(CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE)
```
ArekBalysNordic, line 39:
```suggestion
#if defined(CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE) && defined(CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE)
```
Can they be separated in your scenario of environmental device type?

applications/matter_bridge/src/bridged_device_factory.h
```diff
@@ -121,10 +127,16 @@ inline SimulatedDataProviderFactory &GetSimulatedDataProviderFactory()
...
 inline BleDataProviderFactory &GetBleDataProviderFactory()
 {
-	static BleDataProviderFactory sDeviceDataProvider{
+	static BleDataProviderFactory sDeviceDataProvider
+	{
 #ifdef CONFIG_BRIDGE_ONOFF_LIGHT_BRIDGED_DEVICE
 		{ DeviceType::OnOffLight,
 		  [](UpdateAttributeCallback clb) { return chip::Platform::New<BleOnOffLightDataProvider>(clb); } },
+#endif
+#if defined(CONFIG_BRIDGE_TEMPERATURE_SENSOR_BRIDGED_DEVICE) || defined(CONFIG_BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE)
```
ArekBalysNordic, line 136: The same question as the previous one.

samples/matter/common/src/bridge/bridged_device.h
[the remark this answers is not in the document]
kkasperczyk-no: Ok, I wasn't aware of such task. If we have a follow-up I think we can address it outside of this PR.

samples/matter/common/src/bridge/bridge_manager.cpp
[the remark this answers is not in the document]
kkasperczyk-no: Yes, and this is exactly what happens in this situation, but I think this is correct situation.

samples/matter/common/src/bridge/bridge_manager.cpp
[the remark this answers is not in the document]
ArekBalysNordic: I agree that `value1` and `value2` are so confusing here. I think it will be hard to understand by our customers. I know that also, that the map should be as generic as possible, but in this way, it can increase the difficulty to understand the code. Adding some aliases to the code sounds good.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -31,10 +31,12 @@ void BridgeManager::Init()
 				     false);
 }
 
-CHIP_ERROR BridgeManager::AddBridgedDevices(BridgedDevice *aDevice, BridgedDeviceDataProvider *aDataProvider)
+CHIP_ERROR BridgeManager::AddBridgedDevices(BridgedDevice *aDevices[], BridgedDeviceDataProvider *aDataProvider,
```
ArekBalysNordic, line 34:
```suggestion
CHIP_ERROR BridgeManager::AddBridgedDevices(BridgedDevice *Devices[], BridgedDeviceDataProvider *DataProvider,
```

NIT: I haven't spotted it recently, but I guess we should stick to the same convention in all methods and AFAIR we decided to not use `a` prefixes for methods' arguments.

applications/matter_bridge/src/bridged_device_types/ble_environmental_data_provider.cpp
```diff
@@ -0,0 +1,269 @@
...
+	BleEnvironmentalDataProvider *provider = reinterpret_cast<BleEnvironmentalDataProvider *>(context);
+
+	provider->NotifyUpdateState(Clusters::RelativeHumidityMeasurement::Id,
+				    Clusters::RelativeHumidityMeasurement::Attributes::MeasuredValue::Id,
+				    &provider->mHumidityValue, sizeof(provider->mHumidityValue));
+}
+
+void BleEnvironmentalDataProvider::StartHumidityTimer()
+{
+	static constexpr uint32_t kMeasurementsIntervalMs{ 3000 };
```
ArekBalysNordic, line 223: What about making this value a constexpr/kconfig somewhere else than in the method body?

applications/matter_bridge/Kconfig
kkasperczyk-no: I think it is correct, as ES is one service containing few different characteristics.

applications/matter_bridge/README.rst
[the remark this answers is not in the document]
markaj-nordic: Ok, so for now I will rename BridgedDevice -> MatterBridgedDevice.

samples/matter/common/src/bridge/bridged_device.h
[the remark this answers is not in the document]
markaj-nordic: Ok, so for now I will rename BridgedDevice -> MatterBridgedDevice.

samples/matter/common/src/bridge/bridge_manager.cpp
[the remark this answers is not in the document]
markaj-nordic: Ahh, sorry, I misunderstood the intention in the first place. Will fix it.

applications/matter_bridge/src/bridged_device_factory.h
markaj-nordic: We discussed it actually makes no sense to support them separately.

applications/matter_bridge/src/bridged_device_factory.h
markaj-nordic: ditto.

applications/matter_bridge/README.rst
markaj-nordic: And why would that be? This is a list of device types supported by the `add` command.

applications/matter_bridge/CMakeLists.txt
markaj-nordic: I guess it doesn't hurt, but we'd better avoid that anyway.

applications/matter_bridge/README.rst
ArekBalysNordic: But we have another section regarding BLE devices with its examples below and the Environmental Sensor is a BLE device, isn't it? Or, if it is supported as BLE and simulated device I think it should be mentioned in both sections.

ArekBalysNordic: image

markaj-nordic: Ahh ok, now I got you. Will fix it.

samples/matter/common/src/bridge/bridge_manager.cpp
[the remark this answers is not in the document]
markaj-nordic: Done, I have re-designed the implementation of storage of multiple providers as well as the bridged device removal.

applications/matter_bridge/README.rst
```diff
@@ -182,6 +182,7 @@ Adding a Bluetooth LE bridged device to the Matter bridge
      The argument is mandatory and accepts the following values:
 
      * ``256`` - On/Off Light.
+     * ``291`` - Environmental Sensor (a composition of Temperature Sensor and Humidity Sensor).
```
wiba-nordic, line 185:
```suggestion
     * ``291`` - Environmental Sensor (a combination of Temperature Sensor and Humidity Sensor).
```

applications/matter_bridge/README.rst
```diff
@@ -329,8 +330,10 @@ After building the sample and programming it to your development kit, complete t
 
       .. group-tab:: Testing with Bluetooth LE bridged devices
 
-         a. Build and program the :ref:`peripheral_lbs` sample to an additional development kit compatible with the sample.
-         #. Connect the development kit that is running the :ref:`peripheral_lbs` sample to the PC.
+         a. Build and program the one of the following Bluetooth LE sample to an additional development kit compatible with the sample:
```
wiba-nordic, line 333:
```suggestion
         a. Build and program the one of the following Bluetooth LE samples to an additional development kit compatible with the sample:
```

applications/matter_bridge/README.rst
```diff
@@ -329,8 +330,10 @@ After building the sample and programming it to your development kit, complete t
...
+            * :ref:`peripheral_lbs`
+            * :ref:`peripheral_esp`
```
wiba-nordic, line 335:
```suggestion

            * :ref:`peripheral_lbs`
            * :ref:`peripheral_esp`

```
Need empty lines both before and after indented lists.

applications/matter_bridge/README.rst
```diff
@@ -373,6 +376,8 @@ After building the sample and programming it to your development kit, complete t
                I: Adding OnOff Light bridged device
                I: Added device to dynamic endpoint 3 (index=0)
 
+            Note that in the case of Environmental Sensor, two endpoints are created: the first one implements the Temperature Sensor, and the second one implements Humidity Sensor.
```
wiba-nordic, line 379:
```suggestion
            For the Environmental Sensor, two endpoints are created: one implements the Temperature Sensor, and the other implements the Humidity Sensor.
```
I think this reads a bit better. Unless the order of the endpoints is very important, in which case it should "the first one" and "the second" instead of "one" and "the other", respectively.

applications/matter_bridge/README.rst
```diff
@@ -383,6 +388,21 @@ After building the sample and programming it to your development kit, complete t
...
+               :class: highlight
+
+               ./chip-tool temperaturemeasurement read measured-value *<bridge_node_ID>* *<bridged_device_endpoint_ID>*
+
+              * humidity:
+
+            .. parsed-literal::
+               :class: highlight
+
+               ./chip-tool relativehumiditymeasurement read measured-value *<bridge_node_ID>* *<bridged_device_endpoint_ID>*
```
wiba-nordic, line 404:
```suggestion

              * temperature:

                .. parsed-literal::
                   :class: highlight

                   ./chip-tool temperaturemeasurement read measured-value *<bridge_node_ID>* *<bridged_device_endpoint_ID>*

              * humidity:

                .. parsed-literal::
                   :class: highlight

                    ./chip-tool relativehumiditymeasurement read measured-value *<bridge_node_ID>* *<bridged_device_endpoint_ID>*
```
Empty line before the first bullet, and correct indentation for the parsed-literal entries.

wiba-nordic (changes requested):

applications/matter_bridge/README.rst
markaj-nordic: I agree, the order here is not important.

wiba-nordic (approved):
Docs are good.

applications/matter_bridge/src/bridged_device_types/ble_environmental_data_provider.cpp
```diff
@@ -0,0 +1,268 @@
...
+void BleEnvironmentalDataProvider::Init()
+{
+	/* Configure subscription for the temperature characteristic */
+	mGattTemperatureSubscribeParams.ccc_handle = mCccTemperatureHandle;
+	mGattTemperatureSubscribeParams.value_handle = mTemperatureCharacteristicHandle;
+	mGattTemperatureSubscribeParams.value = BT_GATT_CCC_NOTIFY;
+	mGattTemperatureSubscribeParams.notify = BleEnvironmentalDataProvider::GattTemperatureNotifyCallback;
+
+	/* Configure subscription for the humidity characteristic */
+	mGattHumiditySubscribeParams.ccc_handle = mCccHumidityHandle;
```
kkasperczyk-no, line 82: I wonder, why we want to keep `mGattHumiditySubscribeParams` in a class, when we know the subscription establishment will fail? Wouldn't be better to just configure read parameters and add a comment that subscriptions are not supported? If we would like to implement generic check if subscription is supported maybe we should add a helper method to the BLEConnectivityManager. In this class it doesn't make much sense to me, as no-one will be able to re-use it outside anyway, and here it looks like redundant check.

applications/matter_bridge/src/bridged_device_types/ble_environmental_data_provider.cpp
```diff
@@ -0,0 +1,268 @@
...
+{
+	if (mUpdateAttributeCallback) {
+		mUpdateAttributeCallback(*this, clusterId, attributeId, data, dataSize);
+	}
+}
+
+CHIP_ERROR BleEnvironmentalDataProvider::UpdateState(chip::ClusterId clusterId, chip::AttributeId attributeId,
+						     uint8_t *)
+{
+	LOG_INF("Updating state of the BleEnvironmentalDataProvider, cluster ID: %u, attribute ID: %u. Dropping, currently not supported.",
```
kkasperczyk-no, line 101: I think this log is not needed, it's a common situation that write may not supported for some device type.

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -21,44 +21,64 @@ LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
...
-	VerifyOrReturn(provider != nullptr,
-		       shell_fprintf(shell, SHELL_INFO, "Cannot allocate data provider of given type\n"));
+	VerifyOrReturn(provider != nullptr, shell_fprintf(shell, SHELL_INFO, "No valid data provider!\n"));
 
-	auto *newBridgedDevice = BridgeFactory::GetBridgedDeviceFactory().Create(
-		static_cast<BridgedDevice::DeviceType>(deviceType), nodeLabel);
+	CHIP_ERROR err;
+	if (deviceType == MatterBridgedDevice::DeviceType::EnvironmentalSensor) {
+		/* Handle special case for environmental sensor */
+		auto *newBridgedDevice1 = BridgeFactory::GetBridgedDeviceFactory().Create(
```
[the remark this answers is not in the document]
markaj-nordic: Sure, it's actually a leftover from the draft implementation.

applications/matter_bridge/Kconfig
```diff
@@ -53,14 +53,18 @@ config BT_SCAN_FILTER_ENABLE
...
 
 config BT_GATT_CLIENT
 	default y
 
 config BT_GATT_DM
 	default y
 
+config BRIDGE_BLE_DEVICE_HUMIDITY_POLLING_INTERVAL
```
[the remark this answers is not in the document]
markaj-nordic: Ok, I thought we may have wanted to have different polling time for different sensors, but I have nothing against generic implementation for all of those which do not support subscription.

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -21,44 +21,64 @@ LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
...
+			       shell_fprintf(shell, SHELL_INFO, "Cannot allocate Matter device of given type\n"));
+
+		auto *newBridgedDevice2 = BridgeFactory::GetBridgedDeviceFactory().Create(
+			MatterBridgedDevice::DeviceType::HumiditySensor, nodeLabel);
+
+		VerifyOrReturn(newBridgedDevice2 != nullptr, delete provider, delete newBridgedDevice1,
+			       shell_fprintf(shell, SHELL_INFO, "Cannot allocate Matter device of given type\n"));
+
+		MatterBridgedDevice *newBridgedDevices[] = { newBridgedDevice1, newBridgedDevice2 };
+		err = BridgeManager::Instance().AddBridgedDevices(newBridgedDevices, provider, 2);
```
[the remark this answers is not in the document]
markaj-nordic: Yep, but it would be `sizeof(newBridgedDevices)/sizeof(MatterBridgedDevice *)` (or Zephyr macro).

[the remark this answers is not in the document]
kkasperczyk-no: Ah, right, sorry, it's ok then.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -31,96 +31,141 @@ void BridgeManager::Init()
...
+	}
+}
 
-	Platform::UniquePtr<BridgedDevice> device(aDevice);
-	Platform::UniquePtr<BridgedDeviceDataProvider> provider(aDataProvider);
-	VerifyOrReturnError(device && provider, CHIP_ERROR_INTERNAL);
+CHIP_ERROR BridgeManager::AddDevices(MatterBridgedDevice *devices[], BridgedDeviceDataProvider *dataProvider,
+				     uint8_t deviceListSize)
+{
+	VerifyOrReturnError(devices && dataProvider, CHIP_ERROR_INTERNAL);
```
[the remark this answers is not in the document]
markaj-nordic: It's good for debugging but I'm fine with removing it. But to be consistent we should do the same in the simulated temperature and humidity data providers.

applications/matter_bridge/src/bridged_device_types/ble_environmental_data_provider.cpp
markaj-nordic: It's good for debugging but I'm fine with removing it. But to be consistent we should do the same in the simulated temperature and humidity data providers.

kkasperczyk-no: I didn't even know we have such logs there too and I thought this provider will do it in a different way. Then I leave the decision to you, I'm ok with both leaving it or removing everywhere.

applications/matter_bridge/src/bridged_device_types/ble_environmental_data_provider.cpp
```diff
@@ -0,0 +1,268 @@
...
+		clusterId, attributeId);
+	return CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE;
+}
+
+int BleEnvironmentalDataProvider::ParseDiscoveredData(bt_gatt_dm *discoveredData)
+{
+	ParseTemperatureCharacteristic(discoveredData);
+	ParseHumidityCharacteristic(discoveredData);
+
+	/* Parsed data is verified when subscribing */
```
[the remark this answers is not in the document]
markaj-nordic: But how the error is supposed to be propagated? I mean, does it make sense to have only support for one characteristic?

applications/matter_bridge/src/bridged_device_types/ble_environmental_data_provider.cpp
markaj-nordic: discussed f2f

samples/matter/common/src/bridge/bridge_manager.cpp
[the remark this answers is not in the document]
markaj-nordic: discussed f2f

wiba-nordic:
Almost forgot, but this should have a changelog entry.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -31,96 +31,141 @@ void BridgeManager::Init()
 				     false);
 }
 
-CHIP_ERROR BridgeManager::AddBridgedDevices(BridgedDevice *aDevice, BridgedDeviceDataProvider *aDataProvider)
+CHIP_ERROR BridgeManager::AddBridgedDevices(MatterBridgedDevice *devices[], BridgedDeviceDataProvider *dataProvider,
+					    uint8_t deviceListSize)
 {
-	aDataProvider->Init();
-	CHIP_ERROR err = AddDevices(aDevice, aDataProvider);
+	VerifyOrReturnError(deviceListSize <= kMaxDevicesPerProvider, CHIP_ERROR_NO_MEMORY);
```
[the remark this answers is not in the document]
kkasperczyk-no: I think we should get rid of it, as the customer may want to use more devices per provider and it will not be convenient to make him modify the core component. I think at some point we will just test it well and put the recommended max number of supported devices to the documentation, as a limitation. For now it may stay, but I would be happy to have at least TODO comment saying it should be removed once we will be confident of our solution.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -31,96 +31,149 @@ void BridgeManager::Init()
...
+				return SafelyRemoveDevice(index);
 			}
 		}
 		index++;
 	}
 	return CHIP_ERROR_NOT_FOUND;
 }
 
-CHIP_ERROR BridgeManager::AddDevices(BridgedDevice *aDevice, BridgedDeviceDataProvider *aDataProvider)
+CHIP_ERROR BridgeManager::SafelyRemoveDevice(uint8_t index)
```
ArekBalysNordic, line 60: Looking at the body of this function I can see it is used to remove duplications only, so maybe it is worth changing its name to RemoveDuplications or something like that.

samples/matter/common/src/bridge/bridge_manager.h
```diff
@@ -30,31 +31,66 @@ class BridgeManager {
...
+
+		MatterBridgedDevice *mDevice;
+		BridgedDeviceDataProvider *mProvider;
 	};
-	using DeviceMap = FiniteMap<DevicePair, kMaxBridgedDevices>;
 
-	CHIP_ERROR AddDevices(BridgedDevice *aDevice, BridgedDeviceDataProvider *aDataProvider);
+	static constexpr uint8_t kMaxBridgedDevices = CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT;
+	static constexpr uint8_t kMaxDataProviders = CONFIG_BRIDGE_MAX_BRIDGED_DEVICES_NUMBER;
+	static constexpr uint8_t kMaxDevicesPerProvider = 2;
```
ArekBalysNordic, line 86: What about making a kConfig for that?

markaj-nordic: I will be removed totally.

samples/matter/common/src/bridge/bridge_manager.cpp
markaj-nordic: Actually this method checks if there is another Matter device (e.g. humidity) associated with the provider which we want to remove together with the Matter device under `index` (e.g. temperature). If so, then we cannot release the memory of this provider as it will be still used, so the implementation somehow mimics the shared pointer - we release the resource only when it stops being owned by all entities. Therefore the `RemoveDuplications` doesn't sound right here.

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -205,7 +228,7 @@ SHELL_STATIC_SUBCMD_SET_CREATE(
 		add, NULL,
 		"Adds bridged device. \n"
 		"Usage: add <bridged_device_type> [node_label]\n"
-		"* bridged_device_type - the bridged device's type, e.g. 256 - OnOff Light, 770 - TemperatureSensor, 775 - HumiditySensor\n"
+		"* bridged_device_type - the bridged device's type, e.g. 256 - OnOff Light, 291 - EnvironmentalSensor, 770 - TemperatureSensor, 775 - HumiditySensor\n"
```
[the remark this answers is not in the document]
markaj-nordic: Right, will move it. Thanks for the catch.

samples/matter/common/src/bridge/bridge_manager.cpp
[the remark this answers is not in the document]
markaj-nordic: Will remove it and add some doxygen for `AddBridgedDevices` method. I guess we should add docs for all APIs but most probably we have a task for that.

markaj-nordic:
> Almost forgot, but this should have a changelog entry.

Done.

wiba-nordic (approved):

ArekBalysNordic (approved):

kkasperczyk-no (approved):

PR #12076, opened 2023-08-18
bluetooth: controller: kconfig: Add periodic adv event length kconfig

sean-madigan:
This can be used to set the periodic advertising event length via sdc_hci_cmd_vs_periodic_adv_event_length_set

### DIFF
--- a/subsys/bluetooth/controller/Kconfig
+++ b/subsys/bluetooth/controller/Kconfig
@@ -138,6 +138,19 @@ config BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT
 	  microseconds. The event length and the connection interval are the
 	  primary parameters for setting the throughput of a connection.
 
+config BT_CTLR_SDC_PERIODIC_ADV_EVENT_LEN_DEFAULT_OVERRIDE
+	bool "Override periodic adv event length default"
+
+config BT_CTLR_SDC_PERIODIC_ADV_EVENT_LEN_DEFAULT
+	int "Default periodic advertising event length [us]"
+	default 7500 if !BT_CTLR_SDC_PERIODIC_ADV_EVENT_LEN_DEFAULT_OVERRIDE
+	range 0 4000000
+	help
+	  The time set aside for periodic advertising each periodic advertising
+	  interval in microseconds. The event length is the primary parameter for
+	  how much data can be transmitted by the periodic advertiser without
+	  scheduling conflicts occurring.
+
 config BT_CTLR_SDC_TX_PACKET_COUNT
 	int "Number of Link Layer ACL TX buffers"
 	default 3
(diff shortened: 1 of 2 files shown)


subsys/bluetooth/controller/hci_driver.c
```diff
@@ -490,6 +490,14 @@ static int configure_supported_features(void)
 		if (err) {
 			return -ENOTSUP;
 		}
+
+		sdc_hci_cmd_vs_periodic_adv_event_length_set_t params = {
+			.event_length_us = CONFIG_BT_CTLR_SDC_PERIODIC_ADV_EVENT_LEN_DEFAULT
```
rugeGerritsen, line 495: Do we have to change something inside the lib to ensure this value is used after HCI reset?

sean-madigan: That's a good point

subsys/bluetooth/controller/Kconfig
```diff
@@ -138,6 +138,19 @@ config BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT
 	  microseconds. The event length and the connection interval are the
 	  primary parameters for setting the throughput of a connection.
 
+config BT_CTLR_SDC_PERIODIC_ADV_EVENT_LEN_DEFAULT_OVERRIDE
```
hermabe, line 141: What is the reason for having the `_OVERRIDE` config? Can we not just set the default for `BT_CTLR_SDC_PERIODIC_ADV_EVENT_LEN_DEFAULT` unconditionally?

rugeGerritsen: We want to be able override the default out of tree for certain configurations

hermabe: That should be possible anyway as long as the out-of-tree config definition is found before this one. Though maybe simpler to keep it for now.

rugeGerritsen: > That should be possible anyway as long as the out-of-tree config definition is found before this one.

Unfortunately this is not always the case

subsys/bluetooth/controller/hci_driver.c
```diff
@@ -490,6 +490,14 @@ static int configure_supported_features(void)
...
+		};
+		err = sdc_hci_cmd_vs_periodic_adv_event_length_set(&params);
```
seko-nordic, line 497: I'm not sure it's good to call HCI commands before sdc_enable. Can we move this call to hci_driver_open instead?

subsys/bluetooth/controller/hci_driver.c
```diff
@@ -965,6 +965,17 @@ static int hci_driver_open(void)
 		return err;
 	}
```
seko-nordic, line 967: can you also set normal event length here using VS cmd and remove corresponding call to sdc_cfg_set(SDC_CFG_TYPE_EVENT_LENGTH)
That would be a nice unification / cleanup

seko-nordic (approved):

PR #12084, opened 2023-08-21
lib: sms: Support for characters in hex format

trantanen:
Added support for giving data as binary but sending to network is still always occuring in 7bit GSM encoding.
This is required by certification to send certain special characters that are not available in ASCII and 0x00 which is '@' sign in 7bit GSM encoding.
Modem shell also updated to support new input data format.

Jira: MOSH-492


samples/cellular/modem_shell/src/sms/sms.c
```diff
@@ -99,27 +100,49 @@ int sms_unregister(void)
...
 
 	ret = sms_register();
 	if (ret != 0) {
 		return ret;
 	}
 
-	ret = sms_send_text(number, text);
+	if (type == SMS_DATA_TYPE_GSM7BIT) {
+		/* Process data to be sent if it's in hex format */
+		data_hex_len = strlen(data) / 2;
```
tokangas, line 130: If the data length is not divisible by 2, the buffer will be too small. That will cause `str_hex_to_bytes()` to not do the conversion and the sent data will be 0 bytes.

If input data always needs to be divisible by 2, it would be good to have check for it and print an error. Otherwise, the length should be calculated like in `hex2bin()`:
```
data_len = strlen(data);
data_hex_len = data_len / 2 + data_len % 2;
```
And one more thing, aren't the names `data_hex` and `data_hex_len` misleading? Input data is hex, but the output is binary, right?

include/modem/sms.h
```diff
@@ -207,6 +217,34 @@ void sms_unregister_listener(int handle);
  */
 int sms_send_text(const char *number, const char *text);
 
+/**
+ * @brief Send SMS message.
```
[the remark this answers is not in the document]
trantanen: I don't know if we should break the API. I've thought that having a clear function for sending text would be nice although I realize that the use cases for such an operation in IOT device is limited.

In any case, maybe `sms_send_data()` could still be `sms_send()` because it will be the function to send all kinds of data (if we ever implement support for normal 8bit data).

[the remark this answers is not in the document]
tokangas: Renaming `sms_send_data()` to `sms_send()` sounds good to me. You can keep the `sms_send_text()` if you like, it's not a deal breaker for me. 😄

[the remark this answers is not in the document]
trantanen: Done. I'll keep `sms_send_text()` although I'm not sure I'm just too used to it and it wouldn't really be needed.

samples/cellular/modem_shell/src/sms/sms.c
trantanen: Yep, I've seen empty messages when hex data is not divisible by 2. I'll add a check that the length is divisible by 2.

I tried how it would work with hex strings such as `616263646` and `hex2bin()` adds `0` in front of the data, which will change the content of the SMS message look completely different and hence confusing if you accidentally miss one number from the hex data.

`data_hex*` renamed to `data_bin*`.

include/modem/sms.h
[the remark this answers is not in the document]
trantanen: Clarifications done to function descriptions of `sms_send_text()` and `sms_send()`.

samples/cellular/modem_shell/src/sms/sms.c
tokangas: > Yep, I've seen empty messages when hex data is not divisible by 2. I'll add a check that the length is divisible by 2.

Is there a test for this, at least I didn't spot it?

tokangas (approved):

samples/cellular/modem_shell/src/sms/sms.c
trantanen: Not yet, I'll check the branch coverage data today and try to fill the gaps, including this one.

trantanen: Coverage data for the library unit tests still looks good. This change is in MOSH so no tests for that.

tokangas: > Coverage data for the library unit tests still looks good. This change is in MOSH so no tests for that.

Of course, my bad.

PR #12085, opened 2023-08-21
boards: shields: nrf7002ek: Fix SPI CS GPIOs overwrite

krish2718:
In case the original DTS has multi-element array for CS GPIOs, then the Wi-Fi shield will overwrite causing build failures. E.g., nRF8160DK 0.14.0 (default now).

As the default value needed for Wi-Fi is already set in original DTS remove the entry from the overlays.

### DIFF
--- a/boards/shields/nrf7002ek/nrf7002ek.overlay
+++ b/boards/shields/nrf7002ek/nrf7002ek.overlay
@@ -7,7 +7,6 @@
 
 &arduino_spi {
 	status = "okay";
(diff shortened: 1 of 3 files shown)


trantanen:
> In case the original DTS has multi-element array for CS GPIOs, then the Wi-Fi shield will overwrite causing build failures. E.g., nRF8160DK 0.14.0 (default now).

8190->9160 just to avoid confusion.

gregersrygg (approved):

trantanen (approved):

anangl (approved):

PR #12105, opened 2023-08-22
doc: add mention of nRF Cloud Multi-service sample to CoAP library

wiba-nordic:
Adds the nRF Cloud multi-service library to the Samples using the library section of the nRF Cloud CoAP library.


plskeggs (approved):
Thanks for catching this!

PR #12107, opened 2023-08-22
applications: nrf_desktop: Add state machine and forced scan state to BLE scan module

MarekPieta:
PR introduces:

- a state machine to BLE scan module
- forced scan state.
  In this state scanning is not interrupted by HID data transmission to host.
  The new state speeds up connection establishment e.g. after dongle boots.

Jira: NCSDK-20843


applications/nrf_desktop/src/modules/ble_scan.c
```diff
@@ -757,7 +777,7 @@ static bool handle_hid_report_event(const struct hid_report_event *event)
 		 * avoid negatively affecting HID report rate.
 		 */
 		scan_start_counter = 0;
-	} else if ((state == STATE_ACTIVE) || (state == STATE_DELAYED_ACTIVE)) {
+	} else if (state == STATE_ACTIVE) {
```
kapi-no, line 780: this change looks like a side fix, it would be better to move it to the commit that introduces the state machine.

kapi-no (commented):
Generally looks good to me, I left a few comments

applications/nrf_desktop/src/modules/ble_scan.c
```diff
@@ -544,8 +513,235 @@ static void scan_init(void)
...
+
+	case STATE_INACTIVE:
+		return "INACTIVE";
+
+	case STATE_OFF:
+		return "OFF";
+
+	case STATE_ERROR:
+		return "ERROR";
+
```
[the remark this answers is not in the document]
MarekPieta: Can we keep it as is?
CAF BLE advertising module uses switch to map states to strings too. A lot of code in the NCS uses switch for similar purpose too, I would keep it for consistency.

applications/nrf_desktop/src/modules/ble_scan.c
MarekPieta: Actually it's a proper update that is needed after `STATE_FORCED_ACTIVE` was introduced. The `STATE_DELAYED_ACTIVE` is going to become `STATE_FORCED_ACTIVE` so that we should not disable scanning on HID report when in `STATE_DELAYED_ACTIVE` state. Previously, the `STATE_DELAYED_ACTIVE` used to become `STATE_ACTIVE`, so proper behavior here was to stop scanning on HID report event.

kapi-no (approved):

applications/nrf_desktop/src/modules/ble_scan.c
kapi-no: I see, thanks

applications/nrf_desktop/src/modules/ble_scan.c
```diff
@@ -544,8 +516,252 @@ static void scan_init(void)
...
+{
+	__ASSERT_NO_MSG(!is_module_init(prev_state) || is_module_init(new_state));
+
+	if (!is_module_init(prev_state)) {
+		if (is_module_init(new_state) && (new_state != STATE_ERROR)) {
+			scan_init();
+		} else {
+			/* Do not perform actions before initialized. */
+			return 0;
+		}
```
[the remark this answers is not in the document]
kapi-no: You forgot to update the comment: ```/* Do not perform actions before initialized or in the error state. */```

alstrzebonski (approved):
LGTM

PR #12115, opened 2023-08-23
applications: serial_lte_modem: Fix #XHTTPCRSP's documentation

tomi-font:
The advertised syntax was incorrect and contradictory to examples appearing on the same page.

It now reflects the actual syntax used by the SLM.

### DIFF
--- a/applications/serial_lte_modem/doc/HTTPC_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/HTTPC_AT_commands.rst
@@ -45,9 +45,9 @@ Response syntax
 
 ::
 
-   #XHTTPCCON=<state>
+   #XHTTPCCON: <state>
 
-* The ``<state>`` value can return one of the following:
+* ``<state>`` is one of the following:
 
   * ``0`` - Disconnected
   * ``1`` - Connected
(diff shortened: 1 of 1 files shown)


applications/serial_lte_modem/doc/HTTPC_AT_commands.rst
```diff
@@ -245,13 +245,11 @@ Syntax
...
-   #XHTTPCRSP=<byte_received>,<state><CR><LF><response>
+   <response><CR><LF>#XHTTPCRSP:<received_byte_count>,<state>
 
-* The ``<byte_received>`` is an integer.
+* ``<response>`` is the raw data of the HTTP response, including headers and body.
+* ``<received_byte_count>`` is an integer.
   It represents the length of a partially received HTTP response.
-* The ``<state>`` value can return one of the following:
-
+* ``<state>`` is one of the following:
```
wiba-nordic, line 253: Needs an empty line after this line.

Is there a reason for the wording change?

wiba-nordic (changes requested):

applications/serial_lte_modem/doc/HTTPC_AT_commands.rst
wiba-nordic: The same wording is on Line 156, is the difference that there it's a response, and this is part of a notification?

wiba-nordic: If you could change the other line as well, then. It's preferable to have the same wording when describing equivalent things, especially on the same page.

wiba-nordic (approved):

rlubos (approved):

PR #12116, opened 2023-08-23
applications: asset_tracker_v2: Fix Wi-Fi+GNSS location priority

trantanen:
ATv2 is ignoring the result of the cloud location methods (Wi-Fi and cellular) and simply indicating to Location library that location is unknown. Hence, Location library will fallback to next method.

This is fine when you have location method priority order as GNSS, Wi-Fi and Cellular and there is nothing to fallback after WiFi and cellular. However, when the priorities are Wi-Fi, GNSS and Cellular, you don't want to fallback to GNSS if Wi-Fi positioning
returns a valid location.

Implemented support for waiting for the location from the cloud and acting based on the outcome. If the cloud backend doesn't support returning the location back to the device, Location library will be still indicated that the location is unknown and it'll do fallback to GNSS.

Jira: NCSDK-22143


applications/asset_tracker_v2/src/events/cloud_module_event.h
```diff
@@ -115,13 +127,15 @@ struct cloud_module_event {
 	enum cloud_module_event_type type;
 
 	union {
-		/** Variable that contains a new configuration received from the cloud service. */
+		/** New configuration received from the cloud service. */
 		struct cloud_data_cfg config;
-		/** Variable that contains data that was attempted to be sent. Could be used
+		/** Cloud location data received from the cloud service. */
+		struct location_data cloud_location;
```
trantanen, line 133: This requires including `modem/location.h` which requires `CONFIG_LOCATION=y`. In `native_posix` configuration this won't be set as location module is turned off.

I guess we have two options. To flag this parameter and its usage with `CONFIG_LOCATION` or `CONFIG_LOCATION_MODULE`. Or create an own `cloud_location_data` structure with similar information to `location_data`.

@simensrostad Opinions?

applications/asset_tracker_v2/src/cloud/nrf_cloud_integration.c
```diff
@@ -440,6 +449,16 @@ int cloud_wrap_cloud_location_send(char *buf, size_t len, bool ack, uint32_t id)
...
 }
 
+bool cloud_wrap_cloud_location_response_wait(void)
+{
+	/* If GNSS is the first priority, then we can ignore the location response from the cloud */
+#if defined(CONFIG_LOCATION_REQUEST_DEFAULT_METHOD_FIRST_GNSS)
+	return false;
+#else
+	return true;
+#endif
```
trantanen, line 459: I've currently done this function so that cloud implementation can decide whether the response is being waited for or not. AWS and Azure integrations will just return false here always.

I've intentionally made here in nRF Cloud integration the trick to return false or true depending on `CONFIG_LOCATION_REQUEST_DEFAULT_METHOD_FIRST_GNSS` to provoke discussion. Originally I had here just `return true`. Flagging on lines 239-245 is tightly related to this.

@simensrostad @tokangas Comments are welcome.

applications/asset_tracker_v2/src/modules/data_module.c
```diff
@@ -855,10 +855,15 @@ static void data_send_work_fn(struct k_work *work)
...
 {
 	if (!k_work_delayable_is_pending(&data_send_work)) {
-		/* If the data_send_work is not pending it means that the module has already
-		 * triggered an data encode/send.
-		 */
-		LOG_DBG("Data already encoded and sent, abort.");
+		if (data_type == APP_DATA_LOCATION) {
+			/* We may get location related data when a fallback to next method occurs */
+			data_send_work_fn(NULL);
+		} else {
```
trantanen, line 861: Allowing location data sending after the first set of data is sent is for the fallbacks. I guess this might be a bug in current implementation too. This adds data transfer between samplings so @simensrostad please comment.

applications/asset_tracker_v2/src/cloud/nrf_cloud_integration.c
trantanen: Yes, this is the most "correct" implementation. However, things done in this PR to handle location response somewhere and ignore in other places complicates the implementation and doesn't make it easier for someone else to come and maintain the code :) It would be easier if we would ignore or handle it always irrespective of the used cloud backend and location method priority list ordering.

applications/asset_tracker_v2/src/cloud/cloud_codec/cloud_codec.h
```diff
@@ -319,6 +320,22 @@ int cloud_codec_encode_cloud_location(
...
+ *
+ * @param[in] input String buffer with encoded cloud location data.
+ * @param[in] input_len Length of input.
+ * @param[out] location Where to store the decoded cloud location data.
+ *
+ * @retval 0 on success.
+ * @retval -EFAULT if cloud returned with error to the location request.
+ * @retval -ENOMEM if codec couldn't allocate memory.
+ * @retval -EINVAL if the data is invalid.
+ * @retval -ENOTSUP if the function is not supported by the encoding backend.
```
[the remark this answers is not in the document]
trantanen: I'll use just backend

applications/asset_tracker_v2/src/cloud/cloud_codec/cloud_codec.h
```diff
@@ -319,6 +320,22 @@ int cloud_codec_encode_cloud_location(
...
+ * @brief Decode received cloud location data.
+ *
+ * @param[in] input String buffer with encoded cloud location data.
+ * @param[in] input_len Length of input.
+ * @param[out] location Where to store the decoded cloud location data.
+ *
+ * @retval 0 on success.
+ * @retval -EFAULT if cloud returned with error to the location request.
+ * @retval -ENOMEM if codec couldn't allocate memory.
+ * @retval -EINVAL if the data is invalid.
```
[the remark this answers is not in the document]
trantanen: I guess this is a leftover from the development. I'll remove it.

applications/asset_tracker_v2/src/cloud/nrf_cloud_integration.c
trantanen: Even with this implementation for WiFi+GNSS+Cellular priority order, we will wait for the location unnecessarily for the last Cellular but I'm not sure that's something we need to optimize.

applications/asset_tracker_v2/src/cloud/cloud_codec/cloud_codec.h
```diff
@@ -319,6 +320,21 @@ int cloud_codec_encode_cloud_location(
...
+ * @brief Decode received cloud location data.
+ *
+ * @param[in] input String buffer with encoded cloud location data.
+ * @param[in] input_len Length of input.
+ * @param[out] location Where to store the decoded cloud location data.
+ *
+ * @retval 0 on success.
+ * @retval -EFAULT if cloud returned with error to the location request.
+ * @retval -ENOMEM if codec couldn't allocate memory.
+ * @retval -ENOTSUP if the function is not supported by the backend.
```
tokangas, line 333: Maybe I'm nitpicking, but I'm still wondering about the return values. The relevant ones for the implementation are `-EFAULT` and `-ENOTSUP`, because these are handled differently in `cloud_location_data_handle()`. These are documented in the API, so no problems there. However, the nRF Cloud implementation can also return at least `-EBADMSG`, which is not documented here.

The returned error values can be somewhat implementation specific, so maybe we could only list the relevant `-EFAULT` and `-NOTSUP` here, and allow other negative errnos to be returned as needed by the implementation. The error value is printed by `cloud_location_data_handle()`, so it may add value to return it as it is, instead of converting the error to a more generic value.

The same may be true for multiple `cloud_codec_` functions, but I'm not asking you to change everything. 😄

applications/asset_tracker_v2/src/cloud/nrf_cloud_integration.c
```diff
@@ -440,6 +449,16 @@ int cloud_wrap_cloud_location_send(char *buf, size_t len, bool ack, uint32_t id)
 	return 0;
 }
 
+bool cloud_wrap_cloud_location_response_wait(void)
+{
+	/* If GNSS is the first priority, then we can ignore the location response from the cloud */
+#if defined(CONFIG_LOCATION_REQUEST_DEFAULT_METHOD_FIRST_GNSS)
```
simensrostad, line 455: Could use `IS_ENABLED`

applications/asset_tracker_v2/src/events/cloud_module_event.h
simensrostad: I'm not sure what I prefer. I guess flagging is fine as long as it doesn't spawm more flagging across modules to accommadate this.

applications/asset_tracker_v2/src/modules/cloud_module.c
```diff
@@ -818,6 +854,17 @@ static void on_sub_state_cloud_connected(struct cloud_msg_data *msg)
 				CLOUD_LOCATION,
 				QOS_FLAG_RELIABILITY_ACK_REQUIRED,
 				true);
+
+		/* Check if the configured cloud service will return the resolved location back
+		 * to the device. If it does not, indicate that location result is unknown.
+		 */
+		if (!cloud_wrap_cloud_location_response_wait()) {
```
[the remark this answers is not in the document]
simensrostad: suggestion, a bit more explicit IMO

applications/asset_tracker_v2/src/modules/cloud_module.c
```diff
@@ -324,6 +324,38 @@ static void pgps_data_handle(const uint8_t *buf, const size_t len)
...
 
+static void cloud_location_data_handle(uint8_t *buf, const size_t len)
+{
+	int err;
+	struct location_data location;
+	struct cloud_module_event *cloud_module_event;
+
+	err = cloud_codec_decode_cloud_location(buf, len, &location);
+	if (err == 0) {
+		cloud_module_event = new_cloud_module_event();
```
simensrostad, line 335: Some reaping code, perhaps something like this would do good:

```
void submit_cloud_module_event(enum cloud_module_event_type type, int error) {
    struct cloud_module_event *cloud_module_event = new_cloud_module_event();
    __ASSERT(cloud_module_event, "Not enough heap left to allocate event");
    
    cloud_module_event->type = type;
    if (type == CLOUD_EVT_CLOUD_LOCATION_RECEIVED) {
        cloud_module_event->data.cloud_location = location;
    } else if (type == CLOUD_EVT_ERROR) {
        cloud_module_event->data.error = error;
    }
    
    APP_EVENT_SUBMIT(cloud_module_event);
}

err = cloud_codec_decode_cloud_location(buf, len, &location);
if (err == 0) {
    submit_cloud_module_event(CLOUD_EVT_CLOUD_LOCATION_RECEIVED, 0);
} else if (err == -EFAULT) {
    submit_cloud_module_event(CLOUD_EVT_CLOUD_LOCATION_ERROR, err);
} else if (err == -ENOTSUP) {
    submit_cloud_module_event(CLOUD_EVT_CLOUD_LOCATION_UNKNOWN, err);
} else {
    LOG_ERR("Decoding of cloud location response, error: %d", err);
    SEND_ERROR(cloud, CLOUD_EVT_ERROR, err);
}
```

applications/asset_tracker_v2/src/modules/data_module.c
simensrostad: What do you mean by data transfer between samplings? Could you elaborate a bit on that

trantanen: I mean there is an additional cloud communication somewhere in the middle of the sampling interval.
Generally the idea in ATv2 is to start sampling, collect various data, send all the data to cloud, and after the sampling interval, do the same thing again. In the middle of the sampling interval, ATv2 is not supposed to send data to cloud.

With this change, we are adding an additional data transfer in the middle of the sampling interval because we first send some location related data where location resolution fails. Then we do fallback in Location library and to get the resolution from this location retrieval into the cloud, we need to transmit that set of data.

As I said, there probably should be something like this irrespective as right now I'm not sure location fallbacks are handled properly. whenever GNSS is not the first priority method.

applications/asset_tracker_v2/src/cloud/cloud_codec/aws_iot/aws_iot_codec.c
```diff
@@ -75,26 +75,24 @@ int cloud_codec_encode_cloud_location(
 		return -ENOMEM;
 	}
 
-	if (!cloud_location->neighbor_cells_valid) {
```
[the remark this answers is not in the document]
trantanen: At the moment it won't be called without ncells or APs. I can add the following if we want but that's dead code from real use point of view. I haven't checked if some unit tests test it though.
`if (!cloud_location->neighbor_cells_valid && !cloud_location->wifi_access_points_valid)`

applications/asset_tracker_v2/src/modules/cloud_module.c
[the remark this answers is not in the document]
trantanen: The reason my naming is weird is that this is not only about whether cloud service/implementation supports location response back to device, but it's also about whether that's preferred solution so some cloud implementation in device side could indicate it doesn't support this although it does, if ignoring the location response was preferred. I know this is not a real issue and I can change the naming, although it's not 100% accurate.

applications/asset_tracker_v2/src/events/cloud_module_event.h
trantanen: I've done the flagging to get native_posix compilation through so please check the instances of `#if defined(CONFIG_LOCATION_MODULE)` in this PR.

applications/asset_tracker_v2/src/modules/data_module.c
simensrostad: I see, I guess this is fine and a penalty we have to deal with when not being able to resolve the first method.

trantanen:
DNM because I'm seeing some reset occasionally when handling cloud location procedure so I need to figure out the root cause and if it's caused by these changes.

plskeggs (approved):

applications/asset_tracker_v2/src/cloud/cloud_codec/cloud_codec.h
trantanen: I'll add @return statement there:
`@return 0 on success or negative error value on failure.`

There's also a separate `@retval 0` line but I don't want to break the consistency within the file with this `@return` so leaving that also.

trantanen:
> DNM because I'm seeing some reset occasionally when handling cloud location procedure so I need to figure out the root cause and if it's caused by these changes.

I think I found this issue and looks promising so far. I'll be still running longer tests but the content of the PR is ready from my perspective.

tokangas (approved):

applications/asset_tracker_v2/doc/location_module.rst
```diff
@@ -88,6 +88,7 @@ Wi-Fi positioning has the following limitations:
 
    The Wi-Fi configuration uses both flash and SRAM extensively.
    You can configure the number of scan results with the :kconfig:option:`CONFIG_LOCATION_METHOD_WIFI_SCANNING_RESULTS_MAX_CNT` Kconfig option to reduce SRAM consumption.
+   Align :kconfig:option:`CONFIG_WIFI_MGMT_SCAN_MAX_BSS_CNT` Kconfig option with the :kconfig:option:`CONFIG_LOCATION_METHOD_WIFI_SCANNING_RESULTS_MAX_CNT` Kconfig option.
```
peknis, line 91:
```suggestion
   Align the :kconfig:option:`CONFIG_WIFI_MGMT_SCAN_MAX_BSS_CNT` Kconfig option with :kconfig:option:`CONFIG_LOCATION_METHOD_WIFI_SCANNING_RESULTS_MAX_CNT`.
```

applications/asset_tracker_v2/doc/location_module.rst
```diff
@@ -100,7 +101,17 @@ All incoming events from other modules are handled in the context of the Applica
 
 The :ref:`lib_location` library handles cellular and Wi-Fi positioning together when the location request method list has them next to each other.
 This means that LTE neighbor cell measurements and Wi-Fi scanning results are combined into the same :c:enum:`LOCATION_EVT_CLOUD_LOCATION_EXT_REQUEST` event.
-The location module responds to the :ref:`lib_location` library with unknown location resolution, because it does not request the location back from cloud service.
+
+The location module sends the cellular and Wi-Fi positioning request forward with :c:enum:`LOCATION_MODULE_EVT_CLOUD_LOCATION_DATA_READY` event,
+and waits for an event from the cloud module on the result of the positioning.
+This result is forwarded to the :ref:`lib_location` library.
+It is up to the cloud integration to select whether positioning result from the cloud service is requested or not.
```
peknis, line 108:
```suggestion
Depending on the cloud integration, positioning result from the cloud service might be requested or not.
```

applications/asset_tracker_v2/doc/location_module.rst
```diff
@@ -100,7 +101,17 @@ All incoming events from other modules are handled in the context of the Applica
...
+The result must be indicated with one of the following events:
+
+* :c:enum:`CLOUD_EVT_CLOUD_LOCATION_RECEIVED`: Location has been resolved successfully. The :ref:`lib_location` library will not continue to use other positioning methods but the location request completes quickly.
+* :c:enum:`CLOUD_EVT_CLOUD_LOCATION_ERROR`: Location resolution failed. The :ref:`lib_location` library will do a fallback to the next positioning method in the priority list.
+* :c:enum:`CLOUD_EVT_CLOUD_LOCATION_UNKNOWN`: Location resolution result is unknown, that is, cloud integration is not capable of or willing to resolve it.
+  The :ref:`lib_location` library will do a fallback to the next positioning method similarly to when :c:enum:`CLOUD_EVT_CLOUD_LOCATION_ERROR` event is received.
```
peknis, line 114:
```suggestion
* :c:enum:`CLOUD_EVT_CLOUD_LOCATION_RECEIVED`: Location has been resolved successfully.
  The :ref:`lib_location` library does not continue to use other positioning methods and the location request completes quickly.
* :c:enum:`CLOUD_EVT_CLOUD_LOCATION_ERROR`: Location resolution failed.
  The :ref:`lib_location` library performs a fallback to the next positioning method in the priority list.
* :c:enum:`CLOUD_EVT_CLOUD_LOCATION_UNKNOWN`: Location resolution result is unknown, that is, cloud integration cannot resolve it.
  The :ref:`lib_location` library performs a fallback to the next positioning method.
```

applications/asset_tracker_v2/src/cloud/cloud_codec/cloud_codec.h
```diff
@@ -319,6 +322,25 @@ int cloud_codec_encode_cloud_location(
...
 
+/* Foward declaration */
+struct location_data;
+
+/**
+ * @brief Decode received cloud location data.
+ *
+ * @param[in] input String buffer with encoded cloud location data.
+ * @param[in] input_len Length of input.
+ * @param[out] location Where to store the decoded cloud location data.
```
peknis, line 333:
```suggestion
 * @param[out] location Storage for the decoded cloud location data.
```

applications/asset_tracker_v2/src/cloud/cloud_codec/cloud_codec.h
```diff
@@ -319,6 +322,25 @@ int cloud_codec_encode_cloud_location(
...
+ *
+ * @retval 0 on success.
+ * @retval -EFAULT if cloud returned with error to the location request.
+ * @retval -ENOMEM if codec couldn't allocate memory.
```
peknis, line 337:
```suggestion
 * @retval -ENOMEM if codec could not allocate memory.
```

applications/asset_tracker_v2/src/cloud/cloud_wrapper.h
```diff
@@ -210,6 +214,14 @@ int cloud_wrap_ui_send(char *buf, size_t len, bool ack, uint32_t id,
  */
 int cloud_wrap_cloud_location_send(char *buf, size_t len, bool ack, uint32_t id);
 
+/**
+ * @brief Indicates whether cloud implementation can and is willing to return the resolved
```
peknis, line 218: Dunno whether this is a matter of willingness.
I'd rather leave it out like I suggested in the doc part.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -179,6 +179,7 @@ nRF9160: Asset Tracker v2
   * Default value of the Kconfig option :kconfig:option:`CONFIG_DATA_ACTIVE_TIMEOUT_SECONDS` is changed to 300 seconds.
   * Enabled link time optimization to reduce the flash size of the application.
     You can disable this using the Kconfig option :kconfig:option:`CONFIG_ASSET_TRACKER_V2_LTO`.
+  * Possibility for the cloud integration to request the location back to the device for WiFi/cellular positioning.
```
peknis, line 182:
```suggestion
  * Possibility for the cloud integration to request the location back to the device for Wi-Fi or cellular positioning.
```

applications/asset_tracker_v2/src/cloud/cloud_wrapper.h
trantanen: Well, it's not purely about the capability. Even if the cloud would be capable of providing the location back to the device in its APIs, it may well not do so due to other factors such as configured location method priority order, etc.
Whether "willigness" is a correct term or not is another question but it's not just about capability.

peknis: How about something like "... can and is configured to return..."?

trantanen: I'll use it.

trantanen: Used also the same phrase in location_module.rst.

peknis: I still see the word willing...?

trantanen: For some unknown reason, my `git push` didn't work properly although locally everything showed that they are aligned with the remote branch. After some tricks, it should be fine now.

peknis (approved):

PR #12122, opened 2023-08-23
drivers: wifi: Optimize the memory consumed by logs

rado17:
Limit the information displayed in logging (function, line etc.). Disable printing of debug message by default.

### DIFF
--- a/drivers/wifi/nrf700x/Kconfig
+++ b/drivers/wifi/nrf700x/Kconfig
@@ -70,6 +70,22 @@ config NRF700X_REG_DOMAIN
 config NET_MGMT_EVENT_STACK_SIZE
 	default 2048 if !WPA_SUPP
 
+config NRF700X_LOG_VERBOSE
+	bool "Maintains the verbosity of information in logs"
+
+module = WIFI_NRF700X
+module-dep = LOG
+module-str = Log level for Wi-Fi nRF700x driver
+module-help = Sets log level for Wi-Fi nRF700x driver
+source "subsys/logging/Kconfig.template.log_config"
+
+config NRF700X_LOG_LEVEL
+	int "Sets the log level of Wi-Fi driver"
+	default 0 if WIFI_NRF700X_LOG_LEVEL_DBG # MSG_EXCESSIVE
+	default 3 if WIFI_NRF700X_LOG_LEVEL_INF # MSG_INFO
+	default 5 if WIFI_NRF700X_LOG_LEVEL_ERR # MSG_ERROR
+	default 6
+
 config NRF700X_ON_QSPI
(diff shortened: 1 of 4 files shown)


VivekUppunda (approved):

drivers/wifi/nrf700x/Kconfig
```diff
@@ -70,6 +70,10 @@ config NRF700X_REG_DOMAIN
 config NET_MGMT_EVENT_STACK_SIZE
 	default 2048 if !WPA_SUPP
 
+config NRF700X_LOG_VERBOSE
+	bool "Maintains the verbosity of information in logs"
+	default n
```
krish2718, line 75: not needed, default is `default n`

modules/hostap/Kconfig
```diff
@@ -170,4 +170,5 @@ config LOG_BUFFER_SIZE
 
 config WPA_SUPP_NO_DEBUG
 	bool "Disable printing of debug messages, saves code size significantly"
+	default y
```
krish2718, line 173: This should be a separate commit.

krish2718: And I don't think default should be no debugs, this hinders customer support.

krish2718 (changes requested):

modules/hostap/Kconfig
rado17: In that case, shall we enable this specifically in all matter related samples for now?

krish2718: Yes, each sample can decide individually.

rado17: Done

drivers/wifi/nrf700x/Kconfig
rado17: Removed.

sachinthegreen (approved):

rado17:
> For Matter we have `WPA_SUPP_NO_DEBUG` already set by default:
> 
> https://github.com/nrfconnect/sdk-connectedhomeip/blob/master/config/nrfconnect/chip-module/Kconfig.defaults#L250

Will drop this commit.

LuDuda (approved):

LuDuda:
Thank you for this improvement! :+1:

As discussed F2F - introduction of `CONFIG_NRF700X_LOG_VERBOSE` is a good improvement for debug variants. Still however, we don't have control over log level in the Wi-Fi OSAL layer. E.g. even if LOGS are completely off (release build), log strings will occupy the flash memory. It would be perfect to have Zephyr based logging approach there.

LuDuda (approved):
Awesome!

modules/hostap/Kconfig
rado17: Right. This was inadvertantly included. Removed now.

krish2718 (approved):

PR #12128, opened 2023-08-24
tests: bluetooth: iso: Add ACL+BIS test scenario

rick1082:
Add ACL test feature for testing BIS with ACL scenario. OCT-2483


tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,476 @@
...
+
+	LOG_WRN("Connection not found");
+
+	return -EINVAL;
+}
+
+static void work_dummy_data_send(struct k_work *work)
+{
+	int ret;
+	char dummy_string[CONFIG_BT_MAX_CONN][sizeof(uint32_t)] = {0};
```
alexsven, line 66: Should this size be configurable as well? We might want to test sending more than 4 bytes

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	k_work_schedule(&remote_peer[channel_index].dummy_data_send_work, K_MSEC(100));
+
+	if (all_peers_connected() == false) {
+		LOG_INF("Not all devices connected");
+		start_scan();
+	} else {
+		LOG_INF("All devices connected");
+		ret = bt_le_scan_stop();
```
alexsven, line 118: As far as I can see, scan is already stopped when device is found, so this seems unnecessary.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,476 @@
...
+static bool all_peers_connected(void)
+{
+	for (int i = 0; i < ARRAY_SIZE(remote_peer); i++) {
+		if (remote_peer[i].conn == NULL) {
+			return false;
+		}
+	}
+
+	return true;
+}
```
alexsven, line 279: This can be moved to the top of the file, that way you don't need to declare it on line 42

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,476 @@
...
+	if (all_peers_connected() == false) {
+		LOG_INF("Not all devices connected");
+		start_scan();
+	} else {
+		LOG_INF("All devices connected");
+		ret = bt_le_scan_stop();
+		if (ret && ret != -EALREADY) {
+			LOG_WRN("Stop scan failed: %d", ret);
+		}
+	}
```
alexsven, line 122:
```suggestion
	if (all_peers_connected()) {
		LOG_INF("All devices connected");
	} else {
		LOG_INF("Not all devices connected");
		start_scan();
	}
```

alexsven (changes requested):

tests/bluetooth/iso/modules/iso_broadcast_sink.c
```diff
@@ -198,9 +198,7 @@ static void iso_recv(struct bt_iso_chan *chan, const struct bt_iso_recv_info *in
 	static uint32_t counts_fail;
 	static uint32_t counts_success;
 
-	if (buf->len == sizeof(count)) {
-		count = sys_get_le32(buf->data);
-	}
+	count = sys_get_le32(buf->data);
```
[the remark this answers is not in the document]
rick1082: Probably no, since we could send a packet which length is 120 bytes, but only first 4 bytes are used for counting.

tests/bluetooth/iso/modules/peripheral.c
```diff
@@ -0,0 +1,138 @@
...
+
+static struct bt_conn_cb conn_callbacks = {.connected = connected_cb,
+					   .disconnected = disconnected_cb};
+
+int peripheral_init(void)
+{
+	int ret;
+
+	LOG_INF("%s", __func__);
+	bt_nus_init(&nus_cb);
```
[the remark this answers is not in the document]
rick1082: In here I involve NUS (Nordic UART Service) for sending dummy data from central to peripheral.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,476 @@
...
+		LOG_ERR("Channel index not found");
+		return;
+	}
+	acl_send_count[channel_index]++;
+	sys_put_le32(acl_send_count[channel_index], dummy_string[channel_index]);
+
+	ret = bt_nus_client_send(&nus_client[channel_index], dummy_string[channel_index],
+				 sizeof(dummy_string[0]));
+	if (ret) {
+		LOG_DBG("failed to send, ret = %d", ret);
```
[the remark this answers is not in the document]
rick1082: Might not be error, since this might occur sometimes because of ACL has hard retransmission mechanism, which keeps retransmit a packet until timeout. So we could see this warning pop up if a packet cannot be sent to peer in a connection interval and we try to send a new packet.

tests/bluetooth/iso/modules/central.c
rick1082: This is for ACL dummy data, let me figure out a way to make it.

tests/bluetooth/iso/modules/central.c
rick1082: Let me make a test.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	k_work_reschedule(&remote_peer[channel_index].dummy_data_send_work, K_MSEC(100));
```
koffes, line 96: Can this be hard coded to 100 ? It will depend on the connection interval.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,476 @@
...
+
+		ret = bt_le_scan_stop();
+		if (ret && ret != -EALREADY) {
+			LOG_WRN("Stop scan failed: %d", ret);
+		}
+
+		ret = bt_conn_le_create(addr, BT_CONN_LE_CREATE_CONN, &conn_param, &conn);
+		if (ret) {
+			LOG_ERR("Could not init connection");
+			start_scan();
```
koffes, line 210:
```suggestion
			scan_start();
```
Verb at the end, like what was done with bt_le_scan_stop

tests/bluetooth/iso/modules/central.c
rick1082: Yes, scan will stop once if the MAX_CONN devices are connected.

koffes (changes requested):

tests/bluetooth/iso/modules/iso_broadcast_sink.c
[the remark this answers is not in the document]
koffes: I don't quite see what this change does.

tests/bluetooth/iso/modules/iso_broadcast_src.c
```diff
@@ -65,8 +65,8 @@ static struct bt_iso_chan_ops iso_ops = {
 };
 
 static struct bt_iso_chan_io_qos iso_tx_qos = {
-	.sdu = sizeof(uint32_t), /* bytes */
-	.rtn = 1,
+	.sdu = 120,
```
koffes, line 68: The default values should be Kconfigs.

tests/bluetooth/iso/modules/central.c
rick1082: I will make it configurable. But a packet will be retransmitted until timeout in ACL link, so we might not able to set it the same as connection interval, because there's a chance that we cannot send it to peer device in one interval, and the application will push more data to host, then we will get error code from host shows that the last packet hasn't sent to peer device frequently.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	ret = bt_gatt_dm_start(conn, BT_UUID_NUS_SERVICE, &discovery_cb,
+			       &nus_client[channel_index]);
+	return ret;
+}
+
+static uint8_t ble_data_received(struct bt_nus_client *nus, const uint8_t *data, uint16_t len)
+{
+	LOG_HEXDUMP_INF(data, len, "NUS received:");
```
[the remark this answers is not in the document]
rick1082: No, in the current code only central sends packet to peripheral periodically for mimic the scenario from the customer.

tests/bluetooth/iso/modules/iso_broadcast_src.c
rick1082: Should we set all the QoS default setting from the KConfig?

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+		LOG_ERR("ACL connection to %s failed, error %d", addr, err);
+		bt_conn_unref(conn);
+		scan_start();
+
+		return;
+	}
+
+	ret = gatt_discover(conn);
+	if (ret) {
+		LOG_ERR("could not start the discovery procedure, ret = %d", ret);
```
alexsven, line 297:
```suggestion
		LOG_ERR("Failed to start the discovery procedure, ret = %d", ret);
```

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+
+	LOG_INF("Disconnected: %s (reason 0x%02x)", addr, reason);
+	bt_conn_unref(conn);
+
+	ret = channel_index_get(conn, &channel_index);
+	if (ret) {
+		LOG_WRN("Unknown connection");
+	} else {
+		remote_peer[channel_index].conn = NULL;
+	}
```
alexsven, line 320: Add blank line after }

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+	int ret;
+
+	ret = bt_le_scan_start(BT_LE_SCAN_PASSIVE, on_device_found);
+	if (ret && ret != -EALREADY) {
+		LOG_WRN("Scanning failed to start: %d", ret);
+		return;
+	}
+
+	LOG_INF("Scanning successfully started");
+}
```
alexsven, line 278:
```suggestion
static int scan_start(void)
{
	int ret;

	ret = bt_le_scan_start(BT_LE_SCAN_PASSIVE, on_device_found);
	if (ret && ret != -EALREADY) {
		LOG_WRN("Scanning failed to start: %d", ret);
		return ret;
	}

	LOG_INF("Scanning successfully started");
	return 0;
}
```

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+
+static struct bt_conn_cb conn_callbacks = {
+	.connected = connected_cb,
+	.disconnected = disconnected_cb,
+};
+
+int central_start(void)
+{
+	scan_start();
+	running = true;
```
alexsven, line 333:
```suggestion
	int ret;

	ret = scan_start();
	if (ret) {
	        return ret;
	}
	running = true;
```

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+
+	if (running) {
+		shell_error(shell, "Arguments can not be changed while running");
+		return -EACCES;
+	}
+
+	return arg_val;
+}
+
+static int set_param(const struct shell *shell, size_t argc, char **argv)
```
alexsven, line 417:
```suggestion
static int param_set(const struct shell *shell, size_t argc, char **argv)
```

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+
+	return 0;
+}
+
+SHELL_STATIC_SUBCMD_SET_CREATE(central_cmd,
+			       SHELL_CMD(init, NULL, "Init central NUS.", central_init),
+			       SHELL_CMD(start, NULL, "Start central NUS.", central_start),
+			       SHELL_CMD(stop, NULL, "Stop central NUS.", central_stop),
+			       SHELL_CMD(cfg, NULL, "Print config.", central_print_cfg),
+			       SHELL_CMD_ARG(set, NULL, "set", set_param, 3, 0),
```
alexsven, line 481:
```suggestion
			       SHELL_CMD_ARG(set, NULL, "set", param_set, 3, 0),
```

tests/bluetooth/iso/modules/peripheral.c
```diff
@@ -0,0 +1,139 @@
...
+}
+
+static struct bt_nus_cb nus_cb = {
+	.received = bt_receive_cb,
+};
+
+static struct bt_conn_cb conn_callbacks = {.connected = connected_cb,
+					   .disconnected = disconnected_cb};
+
+int peripheral_init(void)
```
alexsven, line 110: Init  function should be below start

alexsven (changes requested):

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+#include <stdlib.h>
+#include <ctype.h>
+#include <getopt.h>
+#include <unistd.h>
```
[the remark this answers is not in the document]
rick1082: Those are for command line interface, let me do a further check.

tests/bluetooth/iso/modules/central.c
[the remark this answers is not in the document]
rick1082: Since central and peripheral are terms for ACL, probably we can use it without prefix to make the name shorter.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+	ret = nus_client_init();
+	bt_conn_cb_register(&conn_callbacks);
+	for (int i = 0; i < ARRAY_SIZE(remote_peer); i++) {
+		k_work_init_delayable(&remote_peer[i].dummy_data_send_work, work_dummy_data_send);
+	}
+
+	return ret;
+}
+
+int central_print_cfg(const struct shell *shell, size_t argc, char **argv)
```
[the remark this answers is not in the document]
rick1082: Ah, was wondering if I should expose that into header and call in main.c, but seems we don't need to expose it, let me fix it in peripheral and central.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+	}
+
+	nus[channel_index] = context;
+
+	ret = bt_nus_handles_assign(dm, nus[channel_index]);
+	if (ret) {
+		LOG_ERR("NUS handle assigne failed, ret = %d", ret);
+	}
+	bt_nus_subscribe_receive(nus[channel_index]);
+	ret = bt_gatt_dm_data_release(dm);
```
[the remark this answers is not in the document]
rick1082: May I know if there's a tool can detect this ?

tests/bluetooth/iso/modules/central.c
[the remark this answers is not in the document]
rick1082: These four are needed for shell command.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,484 @@
...
+	int ret;
+	struct bt_conn *conn;
+
+	if (all_peers_connected()) {
+		LOG_DBG("All peripherals connected");
+		return 0;
+	}
+
+	if ((data_len == REMOTE_DEVICE_NAME_PEER_LEN) &&
+	    (strncmp(REMOTE_DEVICE_NAME_PEER, data, REMOTE_DEVICE_NAME_PEER_LEN) == 0)) {
```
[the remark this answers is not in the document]
rick1082: Hmmm, I got the point, but probably postive logic would be better than negative one in here since it easy to read.

tests/bluetooth/iso/modules/central.c
rick1082: Did a test, but found that the shell module won't print error log if the helper function is return anything :(
But I will update the API.

alexsven (approved):

tests/bluetooth/iso/modules/peripheral.c
```diff
@@ -0,0 +1,139 @@
...
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+#include <zephyr/shell/shell.h>
+#include <zephyr/bluetooth/bluetooth.h>
+#include <zephyr/bluetooth/conn.h>
+#include <zephyr/sys/byteorder.h>
+#include <bluetooth/services/nus.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(peripheral, CONFIG_ISO_TEST_LOG_LEVEL);
```
koffes, line 13: We may want to rename here. Strange to see ISO in a ACL file.

koffes (approved):

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,490 @@
...
+	struct k_work_delayable dummy_data_send_work;
+};
+static struct peer_device remote_peer[CONFIG_BT_MAX_CONN];
+static struct bt_nus_client *nus[CONFIG_BT_MAX_CONN];
+static struct bt_nus_client nus_client[CONFIG_BT_MAX_CONN];
+static struct bt_le_conn_param conn_param = {
+	.interval_min = 32, /* Set default ACL interval min to 40 ms */
+	.interval_max = 32, /* Set default ACL interval max to 40 ms */
+	.latency = 0,	    /* Set default slave latency to 0 */
+	.timeout = 40	    /* Set default supervision timeout to 4000 ms*/
```
koffes, line 39: This is very long. Can we shorten?

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,490 @@
...
+		LOG_ERR("NUS handle assigne failed, ret = %d", ret);
+	}
+	bt_nus_subscribe_receive(nus[channel_index]);
+
+	ret = bt_gatt_dm_data_release(dm);
+	if (ret) {
+		LOG_ERR("Release service discovery data failed, ret = %d", ret);
+	}
+
+	k_work_schedule(&remote_peer[channel_index].dummy_data_send_work, K_MSEC(100));
```
koffes, line 124: We need to discuss this delay.

tests/bluetooth/iso/modules/central.c
```diff
@@ -0,0 +1,490 @@
...
+		uint8_t len = net_buf_simple_pull_u8(p_ad);
+		uint8_t type;
+
+		/* Check for early termination */
+		if (len == 0) {
+			return;
+		}
+
+		if (len > p_ad->len) {
+			LOG_WRN("AD malformed");
```
koffes, line 243: LOG_ERR?

koffes (changes requested):

tests/bluetooth/iso/modules/central.c
rick1082: This is the time between "service discover finished" and "start sending dummy data over ACL", we can have a quick discuss on Monday.

tests/bluetooth/iso/modules/central.c
koffes: The default is still at 4000 ms. Let's discuss today.

koffes (approved):

tests/bluetooth/iso/modules/central.c
rick1082: Ah... my bad, I was wondering that the naming is too long.
I will change that to 2000ms, like what we have in audio application.

rick1082: The current audio application is using 1s, will change to that.

PR #12137, opened 2023-08-25
Applications: nrf5340_audio: Thread renaming

koffes:
Renamed threads for clarity


alexsven (approved):

erikrobstad (commented):
Is there a reason for power_meas using k_thread_create() while button_handler is using K_THREAD_DEFINE()?
In other words: Why are we doing it in different ways?

rick1082 (approved):

PR #12008, opened 2023-08-10
wifi: Scan extensions

sachinthegreen:
Add support for Wi-Fi scan extensions.

PR #12143, opened 2023-08-28
samples: nrf_cloud_multi_service: Wi-Fi scan config

junqingzou:
CONFIG_NRF700X_SCAN_LIMIT has been removed in PR #12008. 
Replace with CONFIG_WIFI_MGMT_SCAN_MAX_BSS_CNT.

### DIFF
--- a/samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
+++ b/samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
@@ -98,7 +98,7 @@ CONFIG_LOCATION_METHOD_WIFI=y
 # We set this to 30 in this overlay (as opposed to 60 in the wifi-scan-only overlay) so that
 # CONFIG_HEAP_MEM_POOL_SIZE can be small enough to leave at least 24kB of unused RAM in the final
 # build, which is required by WPA_supp
-CONFIG_NRF700X_SCAN_LIMIT=30
+CONFIG_WIFI_MGMT_SCAN_MAX_BSS_CNT=30
 CONFIG_LOCATION_METHOD_WIFI_SCANNING_RESULTS_MAX_CNT=30
 
 ## Enable shell and WIFI_CREDENTIALS shell
@@ -120,7 +120,7 @@ CONFIG_NET_RX_STACK_SIZE=2048
 CONFIG_POSIX_MAX_FDS=16
 CONFIG_NET_SOCKETS_POLL_MAX=8
 # Heap allocation should be changed when CONFIG_LOCATION_METHOD_WIFI_SCANNING_RESULTS_MAX_CNT
-# and CONFIG_NRF700X_SCAN_LIMIT (which should be the same value) are changed. With the limit at 30,
+# and CONFIG_WIFI_MGMT_SCAN_MAX_BSS_CNT (which should be the same value) are changed. With the limit at 30,
 # this could be set as low as 130000. Add 256 bytes for each additional scanning result, assuming
 # sane SSID lengths. We set the heap much higher in case of long SSIDs.
 CONFIG_HEAP_MEM_POOL_SIZE=130000


jayteemo (approved):

plskeggs (approved):

sachinthegreen (approved):

samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
```diff
@@ -124,3 +126,6 @@ CONFIG_NET_SOCKETS_POLL_MAX=8
 # this could be set as low as 130000. Add 256 bytes for each additional scanning result, assuming
 # sane SSID lengths. We set the heap much higher in case of long SSIDs.
 CONFIG_HEAP_MEM_POOL_SIZE=130000
+
+## Configure device provision certificate
+CONFIG_NRF_CLOUD_PROVISION_CERTIFICATES=y
```
glarsennordic, line 131: This is not required by CMakeLists.txt -- you mention it is needed for a test, can you elaborate on that? I think this should be added to the config of that test, rather than in the nrf7002 overlay

samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
```diff
@@ -61,7 +61,9 @@ CONFIG_WIFI_MGMT_EXT=y
 CONFIG_WIFI_CREDENTIALS_BACKEND_SETTINGS=y
 
 ## Use compile-time client ID for nRF Cloud
-CONFIG_NRF_CLOUD_CLIENT_ID_SRC_COMPILE_TIME=y
+CONFIG_NRF_CLOUD_CLIENT_ID_SRC_HW_ID=y
+CONFIG_HW_ID_LIBRARY=y
+CONFIG_HW_ID_LIBRARY_SOURCE_NET_MAC=y
```
glarsennordic, line 66: Similarly, you could enable this in your test, rather than in the overlay. I would rather not prescribe a specific means of getting ID in this overlay, it's meant to be compatible with whatever the user configures themselves

glarsennordic (changes requested):

samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
glarsennordic: Okay. As we discussed over there, requiring `-DCONFIG_NRF_CLOUD_PROVISION_CERTIFICATES` to be enabled  manually was a deliberate decision, but this is not a hill I want to die on

samples/cellular/nrf_cloud_multi_service/src/cloud_connection.c
```diff
@@ -489,6 +490,14 @@ static int setup_cloud(void)
 		return err;
 	}
 
+	err = nrf_cloud_client_id_get(client_id, sizeof(client_id));
+	if (err) {
+		LOG_ERR("nRF Cloud client_id is not initialized, error: %d", err);
+		return err;
+	} else {
+		LOG_INF("nRF Cloud client_id is %s", client_id);
+	}
```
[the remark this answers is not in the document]
glarsennordic: In that case I think we should add this as a feature to the library itself rather than include it here

But sure, let's include it here and I will write a ticket to make printing the client ID as an INFO and remove this later

glarsennordic (approved):

samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
jayteemo: i didn't mean to approve.  i think this should be removed?

jayteemo (changes requested):

samples/cellular/nrf_cloud_multi_service/src/cloud_connection.c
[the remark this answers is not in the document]
glarsennordic: Ticket exists now.

IRIS-6931

samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
junqingzou: Sorry I'm completely not aware of the doc.

`west build --board your_board -p always -- -DCONFIG_NRF_CLOUD_PROVISION_CERTIFICATES=y -DCONFIG_NRF_CLOUD_CLIENT_ID_SRC_COMPILE_TIME=y -DCONFIG_NRF_CLOUD_CLIENT_ID="698d4c11-0ccc-4f04-89cd-6882724e3f6f"`

In my test, I only added one `-DOVERLAY_CONFIG=overlay_nrf7002ek_wifi_no_lte.conf`

junqingzou: > This is not required by CMakeLists.txt -- you mention it is needed for a test, can you elaborate on that? I think this should be added to the config of that test, rather than in the nrf7002 overlay

I mean this line:
`# Include certs files if enabled.
zephyr_include_directories_ifdef(CONFIG_NRF_CLOUD_PROVISION_CERTIFICATES certs)`

As far as I know this is the only way to supply the credentials to lib_nrf_cloud for Wi-Fi devices as the TF-M PS storage is not ready yet.

junqingzou:
Thanks for all the comments, I think that I'll remove the second commit as of now. 

For the long term, if device credentials could be saved somewhere for lib_nrf_cloud to fetch, I think it makes sense to use Wi-Fi MAC address as nRF Cloud client_id, which makes one build of the sample to run on any devices.

Furthermore, this sample works greatly on nRF7002DK (nRF5340 + nRF7002) so maybe the document could be updated to reflect that.

jayteemo (approved):

PR #12148, opened 2023-08-28
samples: openthread: increase the default RX stack size for logging

maciejbaczmanski:
For nrf5340 in cli samples with logging enabled, `nrf5_rx` usage can go up to 808 bytes, resulting with overflow.
This commit increases the default RX stack size if `overlay-logging.conf` is enabled.

### DIFF
--- a/samples/openthread/cli/overlay-logging.conf
+++ b/samples/openthread/cli/overlay-logging.conf
@@ -5,6 +5,9 @@
 # Enable Zephyr logging
 CONFIG_LOG=y
 
+# Increase the default RX stack size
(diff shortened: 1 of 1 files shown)


canisLupus1313 (approved):

edmont:
Doesn't this apply to the other samples?

maciejbaczmanski:
> Doesn't this apply to the other samples?

I've checked coap samples and usage doesn't exceed 652 bytes there.
Coprocessor sample doesn't support nrf5340

edmont (approved):

PR #12162, opened 2023-08-29
sample: lwm2m_client: Rd client start and disconnect fix

jheiskan81:
Replased Client state machine to use a proper function which release proper lock and trig a new state.


jheiskan81:
@rlubos Please review

rlubos (approved):

PR #12170, opened 2023-08-29
samples: openthread: reduce number of OpenThread message buffers

LuDuda:
This commit reduces the default value of OpenThread message buffers depending on the used libraries (it does free from 4k to 8k of RAM).

Additionally, unused SSL Server and SSL Cookie support is disabled when not using Master variant - this gives almost 5k of flash.


maciejbaczmanski (approved):
Stress tests on Thread CI passed:
https://jenkins-ncs.nordicsemi.no/job/latest/job/night/job/test-fw-nrfconnect-thread/job/debugging%252Fmab6%252Freduced_ot_msg_bufs_tests/2/

rlubos (approved):

PR #12175, opened 2023-08-30
applications: asset_tracker_v2: Use new Zephyr overlay arguments

gregersrygg:
Replace overlay arguments OVERLAY_CONFIG and DTC_OVERLAY_FILE with the new Zephyr overlay arguments EXTRA_CONF_FILE and EXTRA_DTC_OVERLAY_FILE in documentation and sample.yaml.

This fixes a problem with some overlays overriding the board overlay for the nRF9160 DK 0.14.0 board in the application folder.

### DIFF
--- a/scripts/quarantine.yaml
+++ b/scripts/quarantine.yaml
@@ -8,13 +8,6 @@
 #    - all
 
 - scenarios:
-    - applications.asset_tracker_v2.nrf7002ek_wifi-debug
-    - applications.asset_tracker_v2.nrf7002ek_wifi-release
-  platforms:
-    - all
-  comment: "Temporary disable till the issue is fixed"
-
-- scenarios:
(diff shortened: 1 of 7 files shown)


simensrostad (approved):

scripts/quarantine.yaml
```diff
@@ -7,13 +7,6 @@
 #  platforms:
 #    - all
 
-- scenarios:
```
jtguggedal, line 10: Was this intentional? 
If so, I think it should be a separate commit

gregersrygg: Taking the tests out of quarantine is intentional. This PR fixes the last issue that caused these tests to fail. I can move it to a separate commit.

divipillai (approved):

PR #12016, opened 2023-08-11
modules/zcbor/tests: Fix command line parameter order

de-nordic:
Correct parameter order to zcbor cddl parser.

PR #12193, opened 2023-08-31
tools: upgrade all python packages & add 'pip-requirements' action

jaci-nordic:
- This is recommit of #11994 which was accidently merged without reviews complete. 

**pip-requirements GH Action:**
- The workflow merges the requirements from nrf, mcuboot and zephyr
- Checks if the requirements-fixed and new generated file match
- Packages that where only in the current requirements-fixed are added to requirements-ci.txt

**python upgrades**
-  ZCBOR commit was pulled in from #12016 to fix twister test. 
- Following packages will be added to requirements-fixed.txt with this change

    Deprecated
    Markdown
    PyGithub
    PyJWT
    PyNaCl
    azure-core
    azure-storage-blob
    capstone
    dill
    editdistance
    exceptiongroup
    gitlint-core
    graphviz
    grpcio
    importlib-resources
    isodate
    jsonschema-specifications
    lark
    libusb-package
    m2r2
    mistune
    msgpack
    pathspec
    pkgutil_resolve_name
    platformdirs
    pypng
    referencing
    rpds-py
    sphinx-copybutton
    sphinx-markdown-tables
    sphinx-ncs-theme
    sphinx-notfound-page
    sphinx-togglebutton
    sphinxcontrib-jquery
    sphinxcontrib-plantuml


tests/modules/lib/zcbor/decode/CMakeLists.txt
```diff
@@ -12,8 +12,8 @@ project(test_zcbor_decode)
 find_program(ZCBOR zcbor REQUIRED)
 
 set(zcbor_args
-  -c ${CMAKE_CURRENT_LIST_DIR}/../cddl/test.cddl
   code
+  -c ${CMAKE_CURRENT_LIST_DIR}/../cddl/test.cddl
```
gmarull, line 16: please split zcbor changes into their own commit

thst-nordic: done

thst-nordic (approved):
bot editing of the requirements-fixed.txt file will come in a future PR. 
We have it working, but not ready for this week.

scripts/requirements-ci.txt
```diff
@@ -0,0 +1,15 @@
+aenum
+bitarray
+chardet
+gitlint==0.18.0
```
karhama, line 4: Could you please add gitpython (it is needed for zephyr test_plan.py)

gmarull (approved):

scripts/fix-requirements.sh
thst-nordic, line 1: i think we should delete this file, and use 'generate-requirements-fixed.sh' which is already there

thst-nordic: done

karhama (approved):

.github/workflows/pip-requirements.yml
```diff
@@ -0,0 +1,81 @@
...
+          OLD=ncs/nrf/scripts/requirements-fixed.txt
+          echo -e "$NEW\t\t\t\t\t"$OLD
+          diff -y $NEW $OLD
+
+      - name: Post summary
+        if: ${{ !cancelled() }}
+        run: |
+          if [[ "failure" == "${{ steps.diff.outcome }}" ]]; then
+            echo -e 'New requirements-fixed.txt file differs from the current file.
+            Please add the newly created file to the PR.' >> $GITHUB_STEP_SUMMARY
```
tejlmand, line 66: i'm aligned with @gmarull on this, ref: https://github.com/nrfconnect/sdk-nrf/pull/11994#discussion_r1304385526

The way pip packages are installed, ref: https://github.com/nrfconnect/sdk-nrf/blob/7408162be1c682b8ddcd22cebd10c14f594d835b/.github/workflows/pip-requirements.yml#L38

means that a new pip package released, which are completely independent of NCS, can cause the  requirements-fixed.txt to change.

A change completely unrelated to the PR being tested, for example, I create a PR which changes foo -> bar in main.c can result in a failed PR, because a new pip package has been released.

Fixing such failure should not be the responsibility of a PR owner (unless that PR owner has made changes to a `requirements-*` file.

If `requirements-fixed.txt` changes so frequently, as mentioned here: https://github.com/nrfconnect/sdk-nrf/pull/11994#discussion_r1306016974, then I propose to setup a daily cron job which generates `requirements-fixed.txt` and opens a PR whenever it changes.

tejlmand (changes requested):
Not approving a PR which may cause PRs to fail for reason that are completely unrelated to the changes in the PR itself.

If the requirements-fixed.txt PR goes in, then we can suddenly have 10 PRs that fails for the same reason, and which author of those 10 should fix the failure ?
Or even worse, all 10 authors waste time to investigate and fix the same error.

Please use a daily cron job, or similar for handling this issue.
Current proposed solution will create a lot of noise for everyone else.

tejlmand:
> If the requirements-fixed.txt PR goes in, then we can suddenly have 10 PRs that fails for the same reason, and which author of those 10 should fix the failure ?
> Or even worse, all 10 authors waste time to investigate and fix the same error.

Seems I've overlooked this line:
https://github.com/nrfconnect/sdk-nrf/blob/e76e113d767bfc5a8daf52851bf79c55416be447/.github/workflows/pip-requirements.yml#L9-L10

so no concerns wrt. when this workflow runs.

Please reply to other comments, as they are still valid.

.github/workflows/pip-requirements.yml
tejlmand: Fixed, see this comment: https://github.com/nrfconnect/sdk-nrf/pull/12193#issuecomment-1704855970
> Seems I've overlooked this line:
> https://github.com/nrfconnect/sdk-nrf/blob/e76e113d767bfc5a8daf52851bf79c55416be447/.github/workflows/pip-requirements.yml#L9-L10

scripts/requirements-build.txt
```diff
@@ -6,4 +6,4 @@ imagesize>=1.2.0
 intelhex
 protobuf
 pylint
-zcbor>=0.7.0
+zcbor==0.6.0
```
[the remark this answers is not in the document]
tejlmand: Ref, the commit to which this comment applies: https://github.com/nrfconnect/sdk-nrf/pull/12193/commits/76ba675c6e4f21196979d7b498181ba144d41f2c

scripts/compile-requirements.sh
```diff
@@ -0,0 +1,32 @@
...
+
+source ~/.local/bin/virtualenvwrapper.sh
+[[ $? != 0 ]] && echo "error sourcing virtualenvwrapper" && exit 1
+
+
+rmvirtualenv pip-fixed-venv > /dev/null 2>&1
+mkvirtualenv pip-fixed-venv  > /dev/null 2>&1
+workon pip-fixed-venv  > /dev/null 2>&1
+pip3 install pip-tools > /dev/null 2>&1
+
```
thst-nordic, line 21: add here: 
`pip3 install setuptools --upgrade `

scripts/requirements-build.txt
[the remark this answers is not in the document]
thst-nordic: @oyvindronningstad I have confirmed that the tests are working on zcbor==0.6 and failing on 0.7
can you comment on why so we can add this information to the commit log, or add a fix commit to this PR quickly?

scripts/requirements.txt
```diff
@@ -1,2 +1,4 @@
 -r requirements-base.txt
 -r requirements-build.txt
+-r requirements-ci.txt
```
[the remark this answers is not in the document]
thst-nordic: @jaci-nordic removed it from requirements.txt and added to to the pip-compile command in compile-requirements.txt

scripts/requirements-build.txt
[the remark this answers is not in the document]
oyvindronningstad: This patch should fix it:
```
diff --git a/tests/modules/lib/zcbor/decode/src/main.c b/tests/modules/lib/zcbor/decode/src/main.c
index 52c9e6f7cb..66d412d343 100644
--- a/tests/modules/lib/zcbor/decode/src/main.c
+++ b/tests/modules/lib/zcbor/decode/src/main.c
@@ -13,7 +13,7 @@ ZTEST(lib_zcbor_test1, test_decode)
        uint32_t payload_len;
        uint8_t time[] = {1, 2, 3, 4, 5, 6, 7, 8};
 
-       ZCBOR_STATE_D(states, 4, payload, sizeof(payload), 1);
+       ZCBOR_STATE_E(states, 4, payload, sizeof(payload), 1);
        struct Test test;
        bool res;
        int int_res;
```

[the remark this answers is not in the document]
thst-nordic: Thank you. I added it to PR, and removed lock on 0.6 from requirements-build.txt.

scripts/requirements.txt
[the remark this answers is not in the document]
tejlmand: could you please fixup this commit: https://github.com/nrfconnect/sdk-nrf/pull/12193/commits/7851b12da6f944c7fbeb9d749f66f788f25fc2b5
to avoid adding the lines in the first place.

It's annoying when reviewing PRs or doing a blame later, to have lines added in one commit, which are then removed again in the next commit.

In this case, those lines:
https://github.com/nrfconnect/sdk-nrf/blob/7851b12da6f944c7fbeb9d749f66f788f25fc2b5/scripts/requirements.txt#L3-L4
are added in this commit: https://github.com/nrfconnect/sdk-nrf/pull/12193/commits/7851b12da6f944c7fbeb9d749f66f788f25fc2b5
and removed in this commit: https://github.com/nrfconnect/sdk-nrf/pull/12193/commits/0176f8f4850d169ea72017d23702480c6845179d

scripts/requirements-fixed.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.8
```
tejlmand, line 2: This file was added in previous commit, using pip-compile with Python 3.10.

Now, in this commit, in same pr, we are suddenly running pip-compile with Python 3.8.

Why has the Python version used for creating `requirement-fixed.txt` suddenly changed ?

And could this change impact the outcome of pip-compile ?

I would like only a single commit in this PR to have the requirements-fixed.txt to be updated, cause I believe that some of the changes to requirements-fixed.txt between commit  https://github.com/nrfconnect/sdk-nrf/pull/12193/commits/7851b12da6f944c7fbeb9d749f66f788f25fc2b5 and https://github.com/nrfconnect/sdk-nrf/pull/12193/commits/0176f8f4850d169ea72017d23702480c6845179d

Some of the reason the second commit suddenly has changes may be because of:
- zbor is no longer having a fixed revision
- Some pip-packages may have been updated remotely, so that a new version is now fetched, compared to when PR was initially opened.
-  Switch from python 3.10 to python 3.8

oyvindronningstad (approved):

tejlmand (changes requested):
Please settle on whether Python 3.8 or 3.10 should be used when running pip-compile to create `requirements-fixed.txt`.

Also please have a single commit which updates `requirements-fixed.txt`.

Some of the changes between the two commits, are very likely to be because pip-compile has been invoked at two different points in time, plus the fact that in one commit `zbor` is fixed at `0.6.0`, and not in later commit.

To approve this PR, I would like a single pip-compile to have been executed.
And that it has been settled whether Python 3.8 or 3.10 should be used.
Cause if we are not specific as to how / which Python to use, then there is a high risk that this check will fail: https://github.com/nrfconnect/sdk-nrf/blob/0176f8f4850d169ea72017d23702480c6845179d/.github/workflows/pip-requirements.yml#L56-L59
when users try to generate an updated `requirements-fixed.txt`, which won't match the file generated by ci.

tejlmand (approved):
looking better.

Still not fond of a PR which adds lines in one commit, just to remove the exact same lines in later commit, as mentioned here https://github.com/nrfconnect/sdk-nrf/pull/12193#discussion_r1319522298

The requirements-fixed now has fewer changes between the two commits, which is good.
Be the fact that there are changes makes me worried that in future PR authors can face a failed CI (github workflow), even if they add a commit with the output from the action, simply because that when they add such commit, then the next run of the action will have another package which may have updated from a `x.y.z` to `x.y.z+1`.

The fact that so far, in this PR, it have not been possible to only have one commit with changes to `requirements-fixed.txt`, increases my worries that this action will cause a lot of noise to those impacted by it.

But besides my concerns, then the state of the PR as it is now is approved.

scripts/requirements-fixed.txt
jaci-nordic: python set to 3.8
zbor vesrion is not fixed
commits related with python packages squashed to one commit
rebased on top of main branch
we will work with autocommit future on separate PR

tejlmand (approved):
Really looking great now, thanks for the extra effort of improving even further 👍

PR #12215, opened 2023-09-01
lib: lte_link_control: remove invalid system mode

tokangas:
Removed `LTE_LC_SYSTEM_MODE_NONE` from the LTE link control API because it is not a valid system mode.

modem_shell used `LTE_LC_SYSTEM_MODE_NONE`, so it had to be updated as well. Changed modem_shell to use the modem system mode by default. Used system mode can be changed using the "link sysmode" command.


peknis (approved):

samples/cellular/modem_shell/prj.conf
```diff
@@ -96,8 +96,8 @@ CONFIG_SMS=y
 
 # LTE link control
 CONFIG_LTE_LINK_CONTROL=y
-CONFIG_LTE_NETWORK_MODE_LTE_M_GPS=y
-CONFIG_LTE_MODE_PREFERENCE_AUTO=y
+# Use modem system mode setting by default.
+CONFIG_LTE_NETWORK_DEFAULT=y
```
trantanen, line 100: This is something that should be mentioned in release notes

jtguggedal (approved):

samples/cellular/modem_shell/prj.conf
tokangas: Yes, you are right. I will ask from the FT team if this change is ok for them. If it is, I'll update the changelog.

tokangas: The change shouldn't affect the FT team, because they always explicitly set the system mode. I'll update the changelog.

trantanen (approved):

jhirsi (approved):

PR #12218, opened 2023-09-01
lib: lte_link_control: fix CE level enum naming

tokangas:
The number of repetitions in the CE level enum names were incorrect. Renamed the values to only have the CE level number, the number of repetitions is mentioned in the description.

### DIFF
--- a/include/modem/lte_lc.h
+++ b/include/modem/lte_lc.h
@@ -619,19 +619,19 @@ enum lte_lc_tau_triggered {
 /** CE level. */
 enum lte_lc_ce_level {
 	/** No repetitions or a small number of repetitions. */
-	LTE_LC_CE_LEVEL_0_NO_REPETITION		= 0,
+	LTE_LC_CE_LEVEL_0	= 0,
 
 	/** Medium number of repetitions. */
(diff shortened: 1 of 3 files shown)


include/modem/lte_lc.h
```diff
@@ -621,19 +621,19 @@ enum lte_lc_tau_triggered {
 /** CE level. */
 enum lte_lc_ce_level {
 	/** No repetitions or a small number of repetitions. */
-	LTE_LC_CE_LEVEL_0_NO_REPETITION		= 0,
+	LTE_LC_CE_LEVEL_0		= 0,
```
trantanen, line 624: You could align the numbers little closer to the name now that all the names are very short.

trantanen (approved):

include/modem/lte_lc.h
tokangas: The values are aligned with tabs and I though that one less tab would not be enough for `LTE_LC_CE_LEVEL_UNKNOWN`, but it seems that it is. I'll remove one tab from all.

jhirsi (approved):

peknis (approved):

jtguggedal (approved):

PR #12221, opened 2023-09-01
applications: serial_lte_modem: add #XNRFCLOUDPOS AT command

tomi-font:
It replaces #XCELLPOS and #XWIFIPOS.

Changes not described in the changelog entries include:
- Fixed the device location sending to nRF Cloud in the case where the log level is DBG.
- Fixed compilation warnings when CONFIG_SLM_NRF_CLOUD=y.
- Fixed some references in the documentation.


applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -67,21 +69,17 @@ static enum {
...
-#if defined(CONFIG_SLM_NRF_CLOUD)
-extern bool nrf_cloud_ready;
-extern bool nrf_cloud_location_signify;
-#endif
 extern struct k_work_q slm_work_q;
 extern struct at_param_list at_param_list;
 extern uint8_t at_buf[SLM_AT_MAX_CMD_LEN];
 
-static int64_t utc_to_gps_sec(const int64_t utc_sec)
+static inline int64_t utc_to_gps_sec(const int64_t utc_sec)
```
junqingzou, line 76: This fucntion is only need under CONFIG_NRF_CLOUD_AGPS.
Please move to right above handle_at_agps().

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -67,21 +69,17 @@ static enum {
...
 {
 	return (utc_sec - GPS_TO_UNIX_UTC_OFFSET_SECONDS) + GPS_TO_UTC_LEAP_SECONDS;
 }
 
-static void gps_sec_to_day_time(int64_t gps_sec, uint16_t *gps_day, uint32_t *gps_time_of_day)
+static inline void
+gps_sec_to_day_time(int64_t gps_sec, uint16_t *gps_day, uint32_t *gps_time_of_day)
 {
```
junqingzou, line 83: This fucntion is only need under CONFIG_NRF_CLOUD_AGPS.
Please move to right above handle_at_agps().

applications/serial_lte_modem/src/nrfcloud/slm_at_nrfcloud.h
```diff
@@ -7,11 +7,31 @@
...
  * @brief Vendor-specific AT command for nRF Cloud service.
  * @{
  */
+#if defined(CONFIG_SLM_NRF_CLOUD)
+
+#include <stdbool.h>
+#include <modem/at_cmd_parser.h>
+
+/* Whether the connection to nRF Cloud is ready. */
+extern bool g_nrf_cloud_ready;
```
junqingzou, line 21: I think that "extern" is used most commonly in C source instead of header.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -394,29 +396,25 @@ static void fix_rep_wk(struct k_work *work)
 	}
 
 #if defined(CONFIG_SLM_NRF_CLOUD)
-	if (IS_ENABLED(CONFIG_SLM_LOG_LEVEL_DBG)) {
```
junqingzou, line 397: This change is not aligned with gnss_startup(). Original design is, under CONFIG_SLM_LOG_LEVEL_DBG, SLM enable ALL NMEA masks so all NMEA messages could be logged (which is for customer to debug GPS issue before a fix). While in the other case, if not CONFIG_SLM_LOG_LEVEL_DBG, SLM only enable GGA that is only needed to be signified to nRF Cloud for map drawing. With current change, SLM will send lots of unnecessary NMEA messages to nRF Cloud for location signification.

applications/serial_lte_modem/doc/NRFCLOUD_AT_commands.rst
```diff
@@ -67,23 +68,22 @@ When the ``<signify>`` parameter is not specified, it does not signify the locat
...
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 ::
 
-   #XNRFCLOUD: <ready>,<signify>
+   #XNRFCLOUD: <ready>,<send_location>
 
-* The ``<ready>`` value indicates whether the nRF Cloud connection is ready or not.
-* The ``<signify>`` value indicates whether the location info will be signified to nRF Cloud or not.
+Refer to the read command.
```
wiba-nordic, line 86: We can actually re-use content from the read command, so I'd suggest using it here:
```suggestion
.. include:: NRFCLOUD_AT_commands.rst
   :start-after: #XNRFCLOUD: <ready>,<send_location>,<sec_tag>,<device_id>
   :end-before: * The ``<sec_tag>`` value indicates the ``sec_tag`` used for accessing nRF Cloud.
```
This _should_ get just the two definitions needed.

applications/serial_lte_modem/doc/GNSS_AT_commands.rst
```diff
@@ -69,6 +69,7 @@ Unsolicited notification
 
    #XGPS: <NMEA message>
 
+This is emitted when a fix is achieved and location sending to :ref:`nRF Cloud <SLM_AT_NRFCLOUD>` is enabled.
```
wiba-nordic, line 72: I don't think we generally use the term "emitted" unless referring to light from LEDs. Is there another term that could be used?

wiba-nordic (changes requested):

applications/serial_lte_modem/src/nrfcloud/slm_at_nrfcloud.c
```diff
@@ -24,31 +25,28 @@ LOG_MODULE_REGISTER(slm_nrfcloud, CONFIG_SLM_LOG_LEVEL);
...
+/* Cellular positioning services .*/
+enum slm_nrfcloud_cellpos {
+	CELLPOS_NONE, /* Keep first (== 0). */
+	CELLPOS_SINGLE_CELL,
+	CELLPOS_MULTI_CELL,
+
+	CELLPOS_COUNT /* Keep last. */
+};
+/* Whether cellular positioning is requested and if so, which. */
+static enum slm_nrfcloud_cellpos s_cell_pos = CELLPOS_NONE;
```
MarkusLassila, line 42: Prefixing static variables with s_ does not seem to be well established in applications or samples.

applications/serial_lte_modem/src/nrfcloud/slm_at_nrfcloud.h
SeppoTakalo: https://en.cppreference.com/w/c/language/storage_duration

Usage of `extern` in C file is a sign of bad code design and a source of bugs that compiler cannot detect. You might accidentally violate [DCL40-C. Do not create incompatible declarations of the same function or object](https://wiki.sei.cmu.edu/confluence/display/c/DCL40-C.+Do+not+create+incompatible+declarations+of+the+same+function+or+object).

So if a variable must be shared between C files, this is the correct way. Declare it in the header as `extern`. Define it once in C file.

junqingzou: Thanks for the knowledge

applications/serial_lte_modem/doc/NRFCLOUD_AT_commands.rst
wiba-nordic: You can also add tags, though they can sometimes have problems when used inside lists. The tags should be decently specific and need to have an empty line both before and after the tag. 
```
.. tag_name_start

content to be re-used

.. tag_name_end
```

wiba-nordic: Oh, you don't need to have the `..`, but otherwise yes. It's basically adding RST comments to the file, and just referencing the contents of those instead of a content line.

applications/serial_lte_modem/doc/NRFCLOUD_AT_commands.rst
```diff
@@ -183,200 +183,89 @@ Example
...
 
-::
+The ``<MAC x>`` parameter is a string.
+It indicates the MAC address of a Wi-Fi access point and must be formatted as ``%02x:%02x:%02x:%02x:%02x:%02x`` (:c:macro:`WIFI_MAC_ADDR_TEMPLATE`).
 
-   #XCELLPOS=?
+The ``<RSSI x>`` parameter is an optional integer.
+It indicates the signal strength of a Wi-Fi access point in dBm, between ``-128`` and ``0``.
+If provided, it must follow the MAC address parameter of the access point.
+Providing the RSSI parameters helps improve the accuracy of the Wi-Fi location.
```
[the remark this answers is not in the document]
MarkusLassila: My idea was that if I would try this command, it would be faster for me to remember/see the SSID:s of the WLAN's that are currently in the area. This might be useful for occasional use, but has problems with WLAN:s which have same SSID:s.

However, now that I look at the original command, it needed both ssid and mac, so it is OK to remove SSID.

applications/serial_lte_modem/doc/NRFCLOUD_AT_commands.rst
```diff
@@ -183,200 +183,89 @@ Example
...
 
   OK
 
-nRF Cloud cellular location
-===========================
+nRF Cloud location
+==================
 
-The ``#XCELLPOS`` command runs the nRF Cloud cellular location service for location information.
+The ``#XNRFCLOUDPOS`` command sends a request to nRF Cloud to determine the device's location using information from the cellular network and/or Wi-Fi access points.
```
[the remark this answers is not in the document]
wiba-nordic: In general, the use of a slash in running text is imprecise (it can mean both "or" and "and or", depending on context). Used in "and/or", the meaning is clear, but it then becomes the only place where we could use it. For consistency, we try to not use it.

applications/serial_lte_modem/doc/GNSS_AT_commands.rst
wiba-nordic: Would "transmitted" work as an alternative? I'm not saying "emitted" can't be used, it's just the first time I've seen it in this context in our documentation.

applications/serial_lte_modem/src/nrfcloud/slm_at_nrfcloud.c
```diff
@@ -162,140 +166,119 @@ static void ncell_meas_mon(const char *notify)
...
 	if (err) {
-		LOG_ERR("Failed to request Wi-Fi location, error: %d", err);
-		rsp_send("\r\n#XWIFIPOS: %d\r\n", err);
-		run_type = RUN_TYPE_NONE;
-	} else {
-		LOG_INF("nRF Cloud Wi-Fi location requested");
+		rsp_send("\r\n#XNRFCLOUDPOS: %d\r\n", err);
+	}
+	if (s_wifi_pos) {
+		k_free(s_wifi_data.ap_info);
```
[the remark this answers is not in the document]
MarkusLassila: Ah, yes. My original comment is misleading. I think that I had something else in here when I was investigating this and left "this is ok" notification instead. s_wifi_pos is indeed necessary.

applications/serial_lte_modem/doc/NRFCLOUD_AT_commands.rst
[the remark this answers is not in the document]
wiba-nordic: You could add ", or both" to the end of that particular sentence to keep the meaning without adding too many words.

applications/serial_lte_modem/doc/GNSS_AT_commands.rst
```diff
@@ -118,6 +116,9 @@ Syntax
 Response syntax
 ~~~~~~~~~~~~~~~
 
+..
+   gps_read_response_start
+
 ::
 
    #XGPS: <gnss_service>,<gnss_status>
```
wiba-nordic, line 124: The includes are failing without an error in the build process, and the end result is just adding this to the build:
image

There's two ways to fix this:
```suggestion
.. gps_read_response_start

.. code-block::

   #XGPS: <gnss_service>,<gnss_status>
```
or
```suggestion

::

   #XGPS: <gnss_service>,<gnss_status>
   

.. gps_read_response_start
```

wiba-nordic: One more hitch, it seems. The content to be included needs to appear in the file before the command to include it. Otherwise the command is, basically, finding itself and causing some of instances of the artifacting.

wiba-nordic (approved):
Now it looks good.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
MarkusLassila: @tomi-font @junqingzou : I must admit that I am not fully following the conversation, but this is keeping me from approving this PR. Have you reached a conclusion whether the current design is sufficient for our use-cases?

MarkusLassila (approved):

junqingzou (approved):

SeppoTakalo (approved):

PR #12232, opened 2023-09-04
NCSIDB-1084 ESB fast rampup static configuration

too1:
- Fixes an issue mentioned in NCSIDB-1084 which made fast ramp up mode unreliable. 
- Made changes to the library to make the fast ramp up option a Kconfig parameter rather than a dynamic init parameter. This in order to more easily set static performance parameters based on the choice of ramp up mode. 
- Optimized the RX_ACK_TIMEOUT and RETRANSMIT_DELAY_MIN parameters to increase performance of the library, and reduce time spent in RX mode when no ACK is received (ACK timeout), as requested in NCSIDB-444.


lemrey (approved):
Looks good, but it is missing the changelog entry.

subsys/esb/Kconfig
```diff
@@ -45,6 +45,16 @@ config ESB_PIPE_COUNT
 	  accidental use of additional pipes, but it's not a problem leaving
 	  this at 8 even if fewer pipes are used.
 
+config ESB_FAST_RAMP_UP
+	bool "Ramp up time configuration for the radio"
+	default n
```
anangl, line 50: `default n` is redundant.

subsys/esb/esb.c
```diff
@@ -1436,8 +1440,9 @@ int esb_init(const struct esb_config *config)
 
 	disable_event.event.generic.event = esb_ppi_radio_disabled_get();
 
-	nrf_radio_modecnf0_set(NRF_RADIO, esb_cfg.use_fast_ramp_up,
-		nrf_radio_modecnf0_dtx_get(NRF_RADIO));
+#if CONFIG_ESB_FAST_RAMP_UP
+	nrf_radio_modecnf0_set(NRF_RADIO, true, nrf_radio_modecnf0_dtx_get(NRF_RADIO));
+#endif
```
anangl, line 1463: Is it okay that now this function is not called when the fast ramp up is not enabled?
Perhaps it would be better to use:
```
	nrf_radio_modecnf0_set(NRF_RADIO, IS_ENABLED(CONFIG_ESB_FAST_RAMP_UP),
		nrf_radio_modecnf0_dtx_get(NRF_RADIO));
```

subsys/esb/esb.c
```diff
@@ -33,32 +33,50 @@ LOG_MODULE_REGISTER(esb, CONFIG_ESB_LOG_LEVEL);
...
 /* Radio Tx ramp-up time in microseconds. */
 #define TX_RAMP_UP_TIME_US 129
 
 /* Radio Rx ramp-up time in microseconds. */
 #define RX_RAMP_UP_TIME_US 124
+
+/* Minimum retransmit time
+ * Tested for 1M and 2M radio modes only
+ */
+#define RETRANSMIT_DELAY_MIN 180
```
anangl, line 79: Wasn't this value supposed to be 360?

PR #12236, opened 2023-09-05
lib: nrf_modem_lib: Dynamic modem trace control

junqingzou:
Currently modem trace is controlled by static configuration. 
Add one configuration to enable deferred trace output: 
`NRF_MODEM_LIB_TRACE_DEFERRED`
Also add two new APIs to provide trace output control: 
`int nrf_modem_lib_trace_start(void)`
`int nrf_modem_lib_trace_stop(void)`

Main purpose is to enable applications to start/stop modem trace dynamically or only when necessary, so as to stay in low power when modem trace is not required.


junqingzou:
Replaced by solution in #12261

PR #12238, opened 2023-09-05
lib: rest_client: Fix memory leak during timeout

trantanen:
This issue was pointed out by Coverity.

### DIFF
--- a/subsys/net/lib/rest_client/src/rest_client.c
+++ b/subsys/net/lib/rest_client/src/rest_client.c
@@ -197,7 +197,8 @@ static int rest_client_sckt_connect(int *const fd,
 		/* Check if timeout has already elapsed */
 		if (time_used >= *timeout_ms) {
 			LOG_WRN("Timeout occurred during DNS query");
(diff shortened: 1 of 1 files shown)


subsys/net/lib/rest_client/src/rest_client.c
```diff
@@ -197,7 +197,8 @@ static int rest_client_sckt_connect(int *const fd,
 		/* Check if timeout has already elapsed */
 		if (time_used >= *timeout_ms) {
 			LOG_WRN("Timeout occurred during DNS query");
-			return -ETIMEDOUT;
+			ret = -ETIMEDOUT;
+			goto clean_up;
```
tokangas, line 201: There's another timeout check around line 245, that should probably be handled in the same way.

trantanen: Thanks, added.

tokangas (approved):

jhirsi (approved):

PR #12246, opened 2023-09-05
net: lib: nrf_cloud: add mutex around modem info struct

jayteemo:
IRIS-6938

### DIFF
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
@@ -31,6 +31,7 @@ bool initialized;
 static const char *application_version;
 
 #if defined(CONFIG_MODEM_INFO)
+static K_MUTEX_DEFINE(modem_inf_mutex);
 static struct modem_param_info modem_inf;
 static bool modem_inf_initd;
 static int init_modem_info(void);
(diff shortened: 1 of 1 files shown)


plskeggs (approved):
The error goes away now?

jayteemo:
> The error goes away now?

Yes, but feel free to test with this change to confirm.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -1650,13 +1659,17 @@ int nrf_cloud_modem_info_json_encode(const struct nrf_cloud_modem_info *const mo
...
-			LOG_ERR("modem_info_params_get() failed: %d", err);
+			LOG_ERR("get_modem_info() failed: %d", err);
 			goto cleanup;
 		}
+		(void)k_mutex_lock(&modem_inf_mutex, K_FOREVER);
 		mpi = &modem_inf;
 	}
 
 	err = encode_modem_info_json_object(mpi, tmp, mod_inf->application_version);
+	if (mpi == &modem_inf) {
```
glarsennordic, line 1670: can `mod_inf->mpi` definitely never already be equal to `&modem_inf`?

Maybe a dedicated boolean that gets set true prior to the lock would be a good idea just to sidestep even having to answer that question

jayteemo: > would be a good idea just to sidestep even having to answer that question

🤔 
ok... flag added, discussion side-stepped.

glarsennordic (approved):

PR #12253, opened 2023-09-06
net: lib: download_client: Allow larger HTTP fragment sizes

jtguggedal:
* The user should be able to set the HTTP fragment size they wish to optimize for their use-case. The option to use preset sizes is kept as is, with the addition of 8192 kB and 16384 kB.

* HTTP fragment size is decoupled from TLS fragment size, so range requests are now only made if explicitly enabled in Kconfig.

### DIFF
--- a/subsys/net/lib/download_client/Kconfig
+++ b/subsys/net/lib/download_client/Kconfig
@@ -11,25 +11,30 @@ if DOWNLOAD_CLIENT
 
 config DOWNLOAD_CLIENT_BUF_SIZE
 	int "Buffer size (static)"
-	range 256 4096
-	default 2048
+	range 256 18432
+	default 2048 if NRF_MODEM_LIB
+	default 4096
 	help
 	  Size of the internal buffer used for incoming and
 	  outgoing packets. It must be large enough to
 	  acommodate for the largest between the HTTP fragment
-	  and CoAP block. In case of CoAP, the CoAP header
+	  and CoAP block plus respective headers. In case of CoAP, the CoAP header
 	  length of 20 bytes should be taken into account.
+	  For HTTP, the header size can not be known in advance in a generic way, and has to be
+	  accomodated for case-by-case.
 
 config DOWNLOAD_CLIENT_HTTP_FRAG_SIZE
-	int
+	int "HTTP fragment size"
 	default 256 if DOWNLOAD_CLIENT_HTTP_FRAG_SIZE_256
 	default 512 if DOWNLOAD_CLIENT_HTTP_FRAG_SIZE_512
 	default 1024 if DOWNLOAD_CLIENT_HTTP_FRAG_SIZE_1024
 	default 2048 if DOWNLOAD_CLIENT_HTTP_FRAG_SIZE_2048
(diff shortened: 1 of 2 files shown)


lemrey (commented):
I'm not sure there is any need for this.. we only use `range` requests when the protocol is HTTPS, otherwise, when using HTTP you can just disable `CONFIG_DOWNLOAD_CLIENT_RANGE_REQUESTS` and the client will send only one `HTTP GET` request and the server will send all the bytes back in one response.

What difference does it make for the application if the bytes are received 2k at a time or 16k at a time? It makes no difference in terms of TCP/IP traffic..

jtguggedal:
> I'm not sure there is any need for this.. we only use `range` requests when the protocol is HTTPS, otherwise, when using HTTP you can just disable `CONFIG_DOWNLOAD_CLIENT_RANGE_REQUESTS` and the client will send only one `HTTP GET` request and the server will send all the bytes back in one response.
> 
> What difference does it make for the application if the bytes are received 2k at a time or 16k at a time? It makes no difference in terms of TCP/IP traffic..

This is need for non-nRF91 devices to increase throughput when downloading over TLS. Range requests and reduces TLS fragment sizes reduce the throughput significantly.

subsys/net/lib/download_client/Kconfig
```diff
@@ -11,25 +11,30 @@ if DOWNLOAD_CLIENT
...
+	default 4096
 	help
 	  Size of the internal buffer used for incoming and
 	  outgoing packets. It must be large enough to
 	  acommodate for the largest between the HTTP fragment
-	  and CoAP block. In case of CoAP, the CoAP header
+	  and CoAP block plus respective headers. In case of CoAP, the CoAP header
 	  length of 20 bytes should be taken into account.
+	  For HTTP, the header size can not be known in advance in a generic way, and has to be
+	  accomodated for case-by-case.
```
simensrostad, line 24:
```suggestion
	  accommodated for case-by-case.
```

simensrostad (approved):
:+1:

lemrey:
Then it should be sufficient to update the check that determines whether to use range request
from:
```
if (client->proto == IPPROTO_TLS_1_2
	   || IS_ENABLED(CONFIG_DOWNLOAD_CLIENT_RANGE_REQUESTS))
```
to:
```
if ((IS_ENABLED(CONFIG_NRF_MODEM_LIB) && (client->proto == IPPROTO_TLS_1_2)
	   || IS_ENABLED(CONFIG_DOWNLOAD_CLIENT_RANGE_REQUESTS))
```

If we don't `range` requests, then the size of the buffer doesn't matter.

lemrey:
We are planning some improvements to this library that would overcome several of its shortcomings, but we are going in a different direction where instead the fragment size as implemented today is going to be removed; so I am not sure we should proceed with this.

jtguggedal:
> We are planning some improvements to this library that would overcome several of its shortcomings, but we are going in a different direction where instead the fragment size as implemented today is going to be removed; so I am not sure we should proceed with this.

Alright, I'd be interested in taking a look at that. Just need to make sure that this works for Wi-Fi too, where we don't have the same limitations as for nRF91.

PR #12254, opened 2023-09-06
memfault: Update version to 1.2.1 and various improvements

jtguggedal:
* Update Memfault SDK version to 1.2.1 to pull in the latest changes.

* Use CONFIG_MEMFAULT_PLATFORM_EXTRA_CONFIG_FILE to patch in
the extra configuration file `memfault_platform_config_extra.h`
when needed. This allows users to have their own
`memfault_platform_config.h` files without getting a build error.

* Add the current modem firmware version as metric.


modules/memfault-firmware-sdk/CMakeLists.txt
```diff
@@ -33,3 +33,7 @@ zephyr_library_sources_ifdef(
 zephyr_include_directories_ifdef(
         CONFIG_MEMFAULT_METRICS_EXTRA_DEFS_FILE
         config)
+
+zephyr_include_directories_ifdef(
+        CONFIG_MEMFAULT_PLATFORM_EXTRA_CONFIG_FILE
+        config)
```
KajaKoren, line 39: The config folder gets added if you enable CONFIG_MEMFAULT_METRICS_EXTRA_DEFS_FILE, even though CONFIG_MEMFAULT_PLATFORM_EXTRA_CONFIG_FILE is disabled. Can this cause side effects?

KajaKoren (approved):

modules/memfault-firmware-sdk/CMakeLists.txt
gminn: No side effects! `memfault_platform_config_extra.h` will only get used if `CONFIG_MEMFAULT_PLATFORM_EXTRA_CONFIG_FILE` is defined, which is a result of [this](https://github.com/memfault/memfault-firmware-sdk/blob/master/ports/zephyr/config/memfault_zephyr_platform_config.h#L82) check in the Zephyr platform config.

jori-nordic (approved):

samples/bluetooth/peripheral_mds/sample.yaml
```diff
@@ -7,6 +7,8 @@ tests:
     extra_configs:
       - CONFIG_MEMFAULT_NCS_PROJECT_KEY="dummy-key"
       - CONFIG_MEMFAULT_NCS_DEVICE_ID="dummy-device-id"
+      # Needed to avoid warning from Memfault SDK
+      - CONFIG_LOG_PRINTK=n
```
KAGA164, line 11: Is it ok to disable this config only in this place?

KAGA164 (approved):

samples/bluetooth/peripheral_mds/sample.yaml
gminn: We just shipped a fix in v1.3.3 of the Memfault SDK (see Changlog [here](https://github.com/memfault/memfault-firmware-sdk/blob/master/CHANGES.md#chart_with_upwards_trend-improvements)) addressing the warning, so this extra config can be removed as long as the sdk version is bumped to 1.3.3.

jtguggedal: Thanks, I have updated the version to 1.3.3 and removed the workaround

PR #12262, opened 2023-09-07
Add missing cell RSRP value for GNSS assistance

tokangas:
Added missing cell RSRP value for GNSS assistance to Location library and modem_shell sample.

Changed Location library to use `%NCELLMEAS` instead of `%XMONITOR` to get the cell information, because `%XMONITOR` may return stale cell information when modem is in PSM. `%XMONITOR` was also used as a fallback for MFWs which do not support `%NCELLMEAS`, but that is no longer needed with the MFWs supported by the next SDK release.


trantanen (approved):
I created NCSDK-23454 so we'll re-evaluate what search types we should use in cellular positioning so we can approve this PR.

jhirsi (approved):

PR #12265, opened 2023-09-07
samples: bluetooth: encrypt throughput sample

martintv:
By the same logic we used when we decided to encrypt the llpm sample in https://github.com/nrfconnect/sdk-nrf/pull/11720
we should also encrypt throughput sample. We should demonstrate the throughput we can get, with encryption on.

And conenct with 7.5ms interval to speed up encryption. The sample will switch to the interval selected by console (or 400ms if nothing is selected) before doing the throgputmeasurment.

Currently failing sample nightly test run, so that needs to be looked into before mergeing. https://jenkins-ncs.nordicsemi.no/blue/organizations/jenkins/latest%2Fsub%2Ftest-fw-nrfconnect-ble_samples/detail/master/9307/pipeline/

Disconnect with BT_HCI_ERR_INSTANT_PASSED

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -288,6 +288,11 @@ Bluetooth samples
 
   * Updated by disabling the :kconfig:option:`CONFIG_BT_SETTINGS_CCC_LAZY_LOADING` Kconfig option as a workaround fix for the `Zephyr issue #61033`_.
 
+* :ref:`ble_throughput` sample:
+
+  * Enabled encryption in the sample.
+    The measured throughput is calculated over the encrypted data, which is how most of the Bluetooth products use this protocol.
+
 Bluetooth mesh samples
 ----------------------
 
--- a/samples/bluetooth/throughput/src/main.c
+++ b/samples/bluetooth/throughput/src/main.c
@@ -29,8 +29,8 @@
 
 #define DEVICE_NAME	CONFIG_BT_DEVICE_NAME
 #define DEVICE_NAME_LEN (sizeof(DEVICE_NAME) - 1)
-#define INTERVAL_MIN	0x140	/* 320 units, 400 ms */
-#define INTERVAL_MAX	0x140	/* 320 units, 400 ms */
+#define INTERVAL_MIN	0x6	/* 6 units, 7.5 ms, only used to setup connection */
+#define INTERVAL_MAX	0x6	/* 6 units, 7.5 ms, only used to setup connection */
 
 #define THROUGHPUT_CONFIG_TIMEOUT K_SECONDS(20)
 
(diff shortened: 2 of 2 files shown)


alwa-nordic (approved):

KAGA164 (commented):
Could security be enabled through Kconfig? The default option can be that encryption is always on.

martintv:
> Could security be enabled through Kconfig? The default option can be that encryption is always on.

we could do that, but we didnt for the llpm sample. I think we should just enable it, without kconfig. In the spirit that security shouldn't be optional. if anyone wants to turn it off for some reason, its easy to remove the call to bt_conn_set_security, but we shouldn't encourage that imo... 

but I don't have so strong feeling here, I could add Kconfig, but we shuold do the same for llpm sample then

grochu (commented):
Please document the update in the changelog: https://github.com/nrfconnect/sdk-nrf/blob/main/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst

martintv:
> Please document the update in the changelog: https://github.com/nrfconnect/sdk-nrf/blob/main/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst

had a look, its confusing, Im 100% sure I would do it wrong, can whoever was going to review it do it themselfs? techwriter probably?

samples/bluetooth/throughput/src/main.c
```diff
@@ -29,8 +29,8 @@
 
 #define DEVICE_NAME	CONFIG_BT_DEVICE_NAME
 #define DEVICE_NAME_LEN (sizeof(DEVICE_NAME) - 1)
-#define INTERVAL_MIN	0x140	/* 320 units, 400 ms */
-#define INTERVAL_MAX	0x140	/* 320 units, 400 ms */
+#define INTERVAL_MIN	0x6	/* 6 units, 7.5 ms, only used to setup connection */
+#define INTERVAL_MAX	0x6	/* 6 units, 7.5 ms, only used to setup connection */
```
jori-nordic, line 33: is this to get the data length update instant faster?

jori-nordic (approved):

grochu:
> had a look, its confusing, Im 100% sure I would do it wrong, can whoever was going to review it do it themselfs? techwriter probably?

I'd propose this:
":ref:`ble_throughput` sample:

  * Enabled encryption in the sample - the measured throughput is calculated over the encrypted data, which is how most of the Bluetooth products use this protocol."
@peknis could you please check this statement? Adding you as reviewer of this PR. Thanks
@martintv the text will depend on whether you decide to add a Kconfig option there.

samples/bluetooth/throughput/src/main.c
martintv: yes, and encrypting the link faster on initial connection.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -288,6 +288,10 @@ Bluetooth samples
 
   * Updated by disabling the :kconfig:option:`CONFIG_BT_SETTINGS_CCC_LAZY_LOADING` Kconfig option as a workaround fix for the `Zephyr issue #61033`_.
 
+* :ref:`ble_throughput` sample:
+
+  * Enabled encryption in the sample - the measured throughput is calculated over the encrypted data, which is how most of the Bluetooth products use this protocol.
```
peknis, line 293:
```suggestion
  * Enabled encryption in the sample.
    The measured throughput is calculated over the encrypted data, which is how most of the Bluetooth products use this protocol.
```

peknis (changes requested):

peknis (approved):

PR #12275, opened 2023-09-07
lib: lte_link_control: Cleanup Kconfig file

tokangas:
Fixed help text indentation and did some other small improvements for the Kconfig file.

### DIFF
--- a/lib/lte_link_control/Kconfig
+++ b/lib/lte_link_control/Kconfig
@@ -5,7 +5,7 @@
 #
 
 menuconfig LTE_LINK_CONTROL
-	bool "nRF91 LTE Link control library"
+	bool "LTE link control library"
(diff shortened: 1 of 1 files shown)


lib/lte_link_control/Kconfig
```diff
@@ -5,7 +5,7 @@
 #
 
 menuconfig LTE_LINK_CONTROL
-	bool "nRF91 LTE Link control library"
+	bool "nRF91 LTE link control library"
```
trantanen, line 8: I'm thinking if we should or shouldn't have nRF91 mentioned here. I don't know if other libraries mention it...

trantanen (approved):

jhirsi (approved):

PR #12281, opened 2023-09-08
applications: serial_lte_modem: Rename AT#XUDPCLI's <cid> to <use_dtls_cid>

tomi-font:
Avoid name clash with the PDP context ID in the socket AT commands.

### DIFF
--- a/applications/serial_lte_modem/src/slm_at_udp_proxy.c
+++ b/applications/serial_lte_modem/src/slm_at_udp_proxy.c
@@ -487,11 +487,7 @@ static bool socket_is_in_use(void)
 	return true;
 }
 
-/**@brief handle AT#XUDPSVR commands
- *  AT#XUDPSVR=<op>[,<port>]
- *  AT#XUDPSVR? READ command not supported
- *  AT#XUDPSVR=?
- */
+/* Handles AT#XUDPSVR commands. */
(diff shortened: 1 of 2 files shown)


applications/serial_lte_modem/src/slm_at_udp_proxy.c
```diff
@@ -487,11 +487,7 @@ static bool socket_is_in_use(void)
 	return true;
 }
 
-/**@brief handle AT#XUDPSVR commands
- *  AT#XUDPSVR=<op>[,<port>]
- *  AT#XUDPSVR? READ command not supported
- *  AT#XUDPSVR=?
- */
+/* Handles AT#XUDPSVR commands. */
 int handle_at_udp_server(enum at_cmd_type cmd_type)
```
junqingzou, line 491: Why remove? It explains the parameters to be parsed.

junqingzou: Same code style for all AT command handlers, not just UDP client

MarkusLassila: Shorter description works for me. It is quite easy to see in the code which operations are supported.

junqingzou (approved):

MarkusLassila (approved):

PR #12285, opened 2023-09-08
nrf_security: Organization improvements

Vge0rge:
Generic improvements of nrf_security files. We move and rename some files so they match their functionality better.

See the commit messages for each change. 

Ref: NCSDK-21866


vili-nordic (approved):
looks alright to me

mswarowsky (approved):
Lgtm,

We should fix the ITS to be encrypted to not rely on name changes

joerchan (approved):

PR #12292, opened 2023-09-11
applications: nrf_desktop: Ensure proper data width

pdunaj:
Avoid possible overflow.

### DIFF
--- a/applications/nrf_desktop/src/hw_interface/motion_buttons.c
+++ b/applications/nrf_desktop/src/hw_interface/motion_buttons.c
@@ -108,7 +108,7 @@ static int16_t ts_diff_to_motion(int64_t diff, int32_t *reminder)
 	return CLAMP(res, INT16_MIN, INT16_MAX);
 }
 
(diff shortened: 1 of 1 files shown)


applications/nrf_desktop/src/hw_interface/motion_buttons.c
```diff
@@ -108,7 +108,7 @@ static int16_t ts_diff_to_motion(int64_t diff, int32_t *reminder)
 	return CLAMP(res, INT16_MIN, INT16_MAX);
 }
 
-static int64_t motion_to_ts_diff(int16_t m)
+static int64_t motion_to_ts_diff(int64_t m)
```
MarekPieta, line 111: The function is actually never called with value of `m` that could trigger an overflow. The `motion_event` also uses `int16_t` to represent motion. Change may not be necessary with the current implementation.
Still, would not block if you prefer to use bigger type here.

MarekPieta (approved):

nvlsianpu (approved):

PR #12296, opened 2023-09-11
memfault: add connectivity layer to sample

KajaKoren:
Use common connectivity layer for LTE and Wi-Fi.
Remove network_lte.c and network_wifi.c


samples/debug/memfault/src/main.c
```diff
@@ -6,22 +6,37 @@
...
+#define L4_EVENT_MASK	      (NET_EVENT_L4_CONNECTED | NET_EVENT_L4_DISCONNECTED)
+#define CONN_LAYER_EVENT_MASK (NET_EVENT_CONN_IF_FATAL_ERROR)
+
 static K_SEM_DEFINE(nw_connected_sem, 0, 1);
 
+/* Macro called upon a fatal error, reboots the device. */
+#define FATAL_ERROR()                                                                              \
+	LOG_ERR("Fatal error! Rebooting the device.");                                             \
+	LOG_PANIC();                                                                               \
+	IF_ENABLED(CONFIG_REBOOT, (sys_reboot(0)))
```
jtguggedal, line 34: I suggest that we don't reboot directly in this sample, as the fatal errors will not cause coredumps and therefore not be reported to Memfault. I think we should use `__ASSERT(...)` insted.

samples/debug/memfault/src/main.c
```diff
@@ -68,33 +83,57 @@ static void button_handler(uint32_t button_states, uint32_t has_changed)
...
 static void on_connect(void)
 {
 #if IS_ENABLED(MEMFAULT_NCS_LTE_METRICS)
 	uint32_t time_to_lte_connection;
-
 	/* Retrieve the LTE time to connect metric. */
-	memfault_metrics_heartbeat_timer_read(
-		MEMFAULT_METRICS_KEY(Ncs_LteTimeToConnect),
-		&time_to_lte_connection);
+	memfault_mek_sem_take(&nw_connected_sem, K_FOREVER);
```
jtguggedal, line 134: I think there's a typo in this line which will cause a compilation fail

jtguggedal (commented):
Good stuff! These are very welcome improvements. I left a few minor comments to consider, in addition to removing the files that are no longer needed

samples/debug/memfault/src/main.c
KajaKoren: seems like I accidentally managed to add a k_sem_take in the middle of the function call, not sure when and how that happend, but it hasn't seem to affect the building of the sample.

simensrostad (commented):
Nice work! Left a few comments, mostly nits.

samples/debug/memfault/src/main.c
```diff
@@ -68,33 +77,57 @@ static void button_handler(uint32_t button_states, uint32_t has_changed)
...
+	network_disconnected();
+}
+
+static void l4_event_handler(struct net_mgmt_event_callback *cb, uint32_t event,
+			     struct net_if *iface)
+{
+	switch (event) {
+	case NET_EVENT_L4_CONNECTED:
+		LOG_INF("Network connectivity established");
+		on_net_event_l4_connected();
```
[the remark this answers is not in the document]
KajaKoren: yes, that make sense There is quite a  bit that can be removed.

jtguggedal (approved):
Looks good!

samples/debug/memfault/README.rst
```diff
@@ -234,18 +234,6 @@ The sample requires the Memfault SDK, which is part of |NCS|'s West manifest, an
...
-
-* :ref:`Trusted Firmware-M <ug_tfm>`
-
-For Wi-Fi, the sample also uses modules found in the following locations in the |NCS| folder structure:
-
-* :file:`modules/lib/hostap`
-* :file:`modules/mbedtls`
+* :ref:`net_if_interface`
+* :ref:`net_mgmt_interface`
+* :ref:`ug_memfault`
```
divipillai, line 239:
```suggestion
It uses the following Zephyr libraries:

* :ref:`net_if_interface`
* :ref:`net_mgmt_interface`
```

samples/debug/memfault/README.rst
```diff
@@ -234,18 +234,6 @@ The sample requires the Memfault SDK, which is part of |NCS|'s West manifest, an
 This sample uses the following |NCS| libraries and drivers:
 
 * :ref:`dk_buttons_and_leds_readme`
```
divipillai, line 236: * :ref:`ug_memfault`

balaji-nordic (approved):
Good stuff :)

samples/debug/memfault/README.rst
```diff
@@ -234,18 +234,6 @@ The sample requires the Memfault SDK, which is part of |NCS|'s West manifest, an
...
-* :ref:`lte_lc_readme`
-* :ref:`nrf_security`
-
-It uses the following `sdk-nrfxlib`_ libraries:
-
-* :ref:`nrfxlib:nrf_modem`
```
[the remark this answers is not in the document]
KajaKoren: I just included the top level libraries, and followed how it was done in https://github.com/nrfconnect/sdk-nrf/blob/8cc62418f3dc4007319fbb90c166ec2bc2cebb3a/samples/net/aws_iot/README.rst?plain=1#L240C13-L240C22, so I was thinking it wasn't necessary to have it listed.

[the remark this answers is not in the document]
divipillai: We list all the relevant libraries, drivers, and secure firmware component here : https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/templates/sample_README.html#dependencies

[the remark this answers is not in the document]
KajaKoren: Ok, thanks! I think it should be following the example now :)

divipillai (approved):

PR #12297, opened 2023-09-11
applications: Improved handling states in Matter Bridge

kkasperczyk-no:
Introduced several improvements for handling different application states:
* Added using IsReachable attribute to decide about propagating read/write requests and return failure response in case device is offline.
* Fixed fault caused by calling connection unref in BLEConnectivityManager on connection coming from a Matter core service.
* Disabled single connection mode, as it resulted in stopping Matter BLE service advertising after any BLE bridged device was connected.

### DIFF
--- a/applications/matter_bridge/src/bridged_device_types/temperature_sensor.cpp
+++ b/applications/matter_bridge/src/bridged_device_types/temperature_sensor.cpp
@@ -98,50 +98,58 @@ CHIP_ERROR TemperatureSensorDevice::HandleReadTemperatureMeasurement(AttributeId
 CHIP_ERROR TemperatureSensorDevice::HandleAttributeChange(chip::ClusterId clusterId, chip::AttributeId attributeId,
 							  void *data, size_t dataSize)
 {
-	if (clusterId != Clusters::TemperatureMeasurement::Id || !data) {
+	CHIP_ERROR err = CHIP_NO_ERROR;
+	if (!data) {
 		return CHIP_ERROR_INVALID_ARGUMENT;
 	}
 
-	CHIP_ERROR err;
+	switch (clusterId) {
+	case Clusters::BridgedDeviceBasicInformation::Id:
+		return HandleWriteDeviceBasicInformation(clusterId, attributeId, data, dataSize);
+	case Clusters::TemperatureMeasurement::Id: {
+		switch (attributeId) {
+		case Clusters::TemperatureMeasurement::Attributes::MeasuredValue::Id: {
+			int16_t value;
 
-	switch (attributeId) {
-	case Clusters::TemperatureMeasurement::Attributes::MeasuredValue::Id: {
-		int16_t value;
+			err = CopyAttribute(data, dataSize, &value, sizeof(value));
 
-		err = CopyAttribute(data, dataSize, &value, sizeof(value));
+			if (err != CHIP_NO_ERROR) {
+				return err;
+			}
 
-		if (err != CHIP_NO_ERROR) {
-			return err;
+			SetMeasuredValue(value);
+
+			break;
 		}
+		case Clusters::TemperatureMeasurement::Attributes::MinMeasuredValue::Id: {
+			int16_t value;
 
-		SetMeasuredValue(value);
+			err = CopyAttribute(data, dataSize, &value, sizeof(value));
 
-		break;
-	}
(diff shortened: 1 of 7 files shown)


samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -411,10 +413,8 @@ CHIP_ERROR BLEConnectivityManager::RemoveBLEProvider(bt_addr_le_t address)
 		return CHIP_ERROR_INTERNAL;
 	}
 
-	/* Release the connection immediately in case of disconnection failed, as callback will not be called. */
-	if (bt_conn_disconnect(provider->GetBLEBridgedDevice().mConn, BT_HCI_ERR_REMOTE_USER_TERM_CONN) != 0) {
-		bt_conn_unref(provider->GetBLEBridgedDevice().mConn);
-	}
+	bt_conn_disconnect(provider->GetBLEBridgedDevice().mConn, BT_HCI_ERR_REMOTE_USER_TERM_CONN);
```
ArekBalysNordic, line 416: What will happen if this function returns an Error? Shouldn't we return at least an error code?

ArekBalysNordic (approved):
One question, overall lgtm.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: The `RemoveBLEProvider` method is called in one of two scenarios, adding a bridged device corresponding to this provider failed, so we have to delete it or the provider destructor was called, so it will be dead soon anyway.
Returning an error would suggest that the removal process failed, which is not true. I don't know how we should handle this error, as if disconnect failed it mean we were not able to close the connection gently, and we will do that brutally in the next step and unref the connection.

ArekBalysNordic: ok, I see, so let's leave it as it is :)

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -113,9 +113,11 @@ void BLEConnectivityManager::DisconnectionHandler(bt_conn *conn, uint8_t reason)
 	/* Try to find provider matching the connection */
 	BLEBridgedDeviceProvider *provider = Instance().FindBLEProvider(*bt_conn_get_dst(conn));
 
-	bt_conn_unref(conn);
-
 	if (provider) {
+		if (conn) {
```
Damian-Nordic, line 117: `conn` is never null per API contract, right? BTW, is it correct to unref the connection in the disconnection handler? Per documentation, when the handler is called, there might be only one reference to the connection object left. So by unref you effectively destroy the object. Or did you `bt_conn_ref` this connection object somewhere?

applications/matter_bridge/src/bridged_device_types/temperature_sensor.cpp
```diff
@@ -98,54 +98,65 @@ CHIP_ERROR TemperatureSensorDevice::HandleReadTemperatureMeasurement(AttributeId
...
-	if (clusterId != Clusters::TemperatureMeasurement::Id || !data) {
+	CHIP_ERROR err = CHIP_NO_ERROR;
+	if (!data) {
 		return CHIP_ERROR_INVALID_ARGUMENT;
 	}
 
-	CHIP_ERROR err;
+	switch (clusterId) {
+	case Clusters::BridgedDeviceBasicInformation::Id:
+		err = HandleWriteDeviceBasicInformation(clusterId, attributeId, data, dataSize);
```
Damian-Nordic, line 108: nit
```suggestion
		return HandleWriteDeviceBasicInformation(clusterId, attributeId, data, dataSize);
```
And similarly below. You don't seem to need a single exit from the function so the function would be easier to understand (I wouldn't need to scroll down to the switch case end :)).

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: Right, removed `conn` check. As discussed F2F calling unref seems to be necessary by the `bt_conn_le_create` method API. I'm still not 100% sure if we do that correctly, however it's the only way that functionally works :) If we do not invoke unref we are not able to create new connection after it was closed (connect->disconnect->connect again scenario).

Damian-Nordic (approved):

PR #12300, opened 2023-09-11
applications: asset_tracker_v2: fix garbage return

balaji-nordic:
cloud_codec_encode_cloud_location was returning garbage value in some cases. This is now fixed by setting the init value of err to -ENODATA.

Found as violation of MISRA C:2004, 9.1 by sonarcloud.

[Link to bug](https://sonarcloud.io/project/issues?resolved=false&types=BUG&sinceLeakPeriod=true&pullRequest=163&id=balaji-nordic_sdk-nrf )

### DIFF
--- a/applications/asset_tracker_v2/src/cloud/cloud_codec/aws_iot/aws_iot_codec.c
+++ b/applications/asset_tracker_v2/src/cloud/cloud_codec/aws_iot/aws_iot_codec.c
@@ -62,7 +62,7 @@ int cloud_codec_encode_cloud_location(
 	struct cloud_codec_data *output,
 	struct cloud_data_cloud_location *cloud_location)
 {
-	int err;
+	int err = -ENODATA;
 	char *buffer;
 
 	__ASSERT_NO_MSG(output != NULL);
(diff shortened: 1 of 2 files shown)


applications/asset_tracker_v2/src/cloud/cloud_codec/aws_iot/aws_iot_codec.c
```diff
@@ -62,7 +62,7 @@ int cloud_codec_encode_cloud_location(
 	struct cloud_codec_data *output,
 	struct cloud_data_cloud_location *cloud_location)
 {
-	int err;
+	int err = -ENODATA;
```
trantanen, line 65: Because this function is not called when there are no neighbor cells or WiFi access points, I had removed the branch returning `-ENODATA`. I'm thinking if we should add it back as static analysis tool seem to complain about it. The main question is whether `cJSON_PrintUnformatted()` works with "empty" cJSON object. I'd expect it does but maybe the function is a little vulnerable after I removed `-ENODATA` branch.

jtguggedal (approved):

simensrostad (approved):

applications/asset_tracker_v2/src/cloud/cloud_codec/aws_iot/aws_iot_codec.c
balaji-nordic: I see. Would you recommend getting this in and then reworking the code in a later point of time?

trantanen: Yep, you can get this merged.

trantanen (approved):

PR #12304, opened 2023-09-12
applications: matter: Improve BLE reconnection mechanism for Bridge

ArekBalysNordic:
Now the reconnection mechanism works independently of the normal scan procedure.
When the connection to the BLE device is lost, the reconnection mechanism starts working, but when a user wants to perform scanning for the new devices, the mechanism stops working while the new scan is in progress, and after that goes back to searching for the lost device.
The scanned devices list is now independent of the reconnection mechanism, so it is not lost during the reconnection scan in progress.
Also, connection to a new BLE device is possible while the reconnection mechanism is working.

### DIFF
--- a/samples/matter/common/src/bridge/ble_bridged_device.h
+++ b/samples/matter/common/src/bridge/ble_bridged_device.h
@@ -20,6 +20,8 @@ struct BLEBridgedDevice {
 	bt_addr_le_t mAddr;
 	BLEConnectivityManager::DeviceConnectedCallback mFirstConnectionCallback;
 	void *mFirstConnectionCallbackContext;
+	bool mInitiallyConnected = false; /* Indicates whether the first connection has been established and the device
+					     has been successfully added to the Bridge. */
 	bt_conn *mConn;
 	BLEBridgedDeviceProvider *mProvider;
 };
@@ -27,10 +29,7 @@ struct BLEBridgedDevice {
 class BLEBridgedDeviceProvider : public BridgedDeviceDataProvider {
 public:
 	explicit BLEBridgedDeviceProvider(UpdateAttributeCallback callback) : BridgedDeviceDataProvider(callback) {}
-	~BLEBridgedDeviceProvider()
-	{
-		BLEConnectivityManager::Instance().RemoveBLEProvider(GetBtAddress());
-	}
+	~BLEBridgedDeviceProvider() { BLEConnectivityManager::Instance().RemoveBLEProvider(GetBtAddress()); }
 
 	virtual bt_uuid *GetServiceUuid() = 0;
 	virtual int ParseDiscoveredData(bt_gatt_dm *discoveredData) = 0;
@@ -46,6 +45,28 @@ public:
 
 	BLEBridgedDevice &GetBLEBridgedDevice() { return mDevice; }
 	void SetConnectionObject(bt_conn *conn) { mDevice.mConn = conn; }
+	void RemoveConnectionObject() { mDevice.mConn = nullptr; }
+
+	/**
+	 * @brief Check if the bridged device has been initially connected.
+	 *
+	 * The bridged device should call the @ref mFirstConnectionCallback only when it is the first connection.
+	 * This method returns indicates whether the callback has been called and the device is already initially
+	 * connected.
+	 *
+	 * @return true if the bridged device has been initially connected.
+	 * @return false if this is the first connection to the bridged device.
+	 */
+	bool IsInitiallyConnected() { return mDevice.mInitiallyConnected; }
+
+	/**
+	 * @brief Confirm that the @ref mFirstConnectionCallback has been already called .
+	 *
+	 * This method informs the bridged device that the @ref mFirstConnectionCallback has been called
+	 * and connection has been established successfully.
+	 *
+	 */
+	void ConfirmInitialConnection() { mDevice.mInitiallyConnected = true; }
 
 	bt_addr_le_t GetBtAddress() { return mDevice.mAddr; }
 
--- a/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
+++ b/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
@@ -33,6 +33,19 @@ static struct bt_conn_le_create_param *create_param = BT_CONN_LE_CREATE_CONN;
 void BLEConnectivityManager::FilterMatch(bt_scan_device_info *device_info, bt_scan_filter_match *filter_match,
 					 bool connectable)
 {
+	if (!filter_match) {
+		LOG_ERR("Filter match error");
+		return;
+	}
+
+	if (filter_match->addr.match) {
+		/* Result comes from the recovery mechanism */
+		Instance().mRecovery.mScannedDevice.mAddr = *filter_match->addr.addr;
+		Instance().mRecovery.mScannedDevice.mConnParam = *device_info->conn_param;
+		Instance().mRecovery.mAddressMatched = true;
+		return;
+	}
+
 	auto scannedDevices = Instance().mScannedDevices;
 	auto scannedDevicesCounter = Instance().mScannedDevicesCounter;
 	/* Limit the number of devices that can be scanned. */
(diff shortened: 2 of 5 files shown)


samples/matter/common/src/bridge/ble_bridged_device.h
```diff
@@ -46,8 +44,11 @@ class BLEBridgedDeviceProvider : public BridgedDeviceDataProvider {
 
 	BLEBridgedDevice &GetBLEBridgedDevice() { return mDevice; }
 	void SetConnectionObject(bt_conn *conn) { mDevice.mConn = conn; }
+	void RemoveConnectionObject() { mDevice.mConn = nullptr; }
+	bool IsInitialized() { return mDevice.mIsInitialized; }
+	void InitializationDone() { mDevice.mIsInitialized = true; }
 
-	bt_addr_le_t GetBtAddress() { return mDevice.mAddr; }
+	bt_addr_le_t *GetBtAddress() { return &mDevice.mAddr; }
```
kkasperczyk-no, line 51: Why do we want to return member address instead of value? It potentially allows to access and modify the `mDevice` member by external parties. I think it would be safer to allow only provider to modify its `mDevice`.

samples/matter/common/src/bridge/ble_bridged_device.h
```diff
@@ -46,8 +44,11 @@ class BLEBridgedDeviceProvider : public BridgedDeviceDataProvider {
 
 	BLEBridgedDevice &GetBLEBridgedDevice() { return mDevice; }
 	void SetConnectionObject(bt_conn *conn) { mDevice.mConn = conn; }
+	void RemoveConnectionObject() { mDevice.mConn = nullptr; }
+	bool IsInitialized() { return mDevice.mIsInitialized; }
```
kkasperczyk-no, line 48: I generally agree with the concept of marking first connection using additional flag, however I think naming it initialization is quite confusing. Provider is initialized before we request creating the BLE connection. We also have `InitializeBridgedDevice` method, so I would expect device is initialized after it was called, but rest of the API suggest something opposite. Maybe we could just name precisely to inform about the purpose e.g. `mInitiallyConnected`, `ConfirmInitialConnection`. Adding some docs to do not leave any doubts about the usage could be useful too.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -198,7 +215,20 @@ CHIP_ERROR BLEConnectivityManager::Init(bt_uuid **serviceUuids, uint8_t serviceU
 	bt_scan_init(&scan_init);
 	bt_scan_cb_register(&scan_cb);
 
-	int err;
+	CHIP_ERROR err = PrepareFiltersForUuid();
+	if (CHIP_NO_ERROR != err) {
+		return err;
+	}
+
+	return CHIP_NO_ERROR;
```
kkasperczyk-no, line 223: It could be just `return PrepareFiltersForUuid();` right?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -198,7 +215,20 @@ CHIP_ERROR BLEConnectivityManager::Init(bt_uuid **serviceUuids, uint8_t serviceU
...
+}
+
+CHIP_ERROR BLEConnectivityManager::PrepareFiltersForUuid()
```
kkasperczyk-no, line 226: Super minor, but I think it should be `PrepareFilterForUuid` and `PrepareFilterForAddress` (singular instead of plural)

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -247,38 +302,30 @@ CHIP_ERROR BLEConnectivityManager::Scan(ScanDoneCallback callback, void *context
...
+							reinterpret_cast<BLEBridgedDeviceProvider *>(context));
+					},
+					reinterpret_cast<intptr_t>(provider));
+				return;
 			}
 		}
-	} else if (Instance().mRecovery.IsNeeded()) {
-		Instance().mRecovery.StartTimer();
+		/* check if device still need to be found (for example it could be removed from the database) */
+		if (Instance().FindBLEProvider(*provider->GetBtAddress())) {
```
kkasperczyk-no, line 321: I'm not sure I understand this check. Maybe there is something I'm missing. I can see it like this:
1. `BLEBridgedDeviceProvider *provider = Instance().mRecovery.GetProvider();` - get provider address
2. `if (provider) {` - it can be true (valid) or false (provider doesn't exist, e.g. it was removed)
3. Try to reconnect
4. `if (Instance().FindBLEProvider(*provider->GetBtAddress())) {` - try to get pointer to provider using a pointer to the same provider object?

ArekBalysNordic: The provider got from the `Instance().mRecovery.GetProvider();` method is an instance in the ring buffer, so it possibly can not be null when we have removed the provider that existed in the bridged devices map. We should check whether the provider in the map still exists before putting it into the ring buffer again, so the 4th bullet is correct. I will add some more comments there.

ArekBalysNordic: But, after rethinking, the checking of the `Instance().FindBLEProvider(provider->GetBtAddress()` should be done instead of point no. 2.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -33,6 +33,21 @@ static struct bt_conn_le_create_param *create_param = BT_CONN_LE_CREATE_CONN;
...
 					 bool connectable)
 {
+	if (!filter_match) {
+		LOG_ERR("Filter match error");
+		return;
+	}
+
+	if (filter_match->addr.match) {
+		/* Result comes from the recovery mechanism */
+		memcpy(&Instance().mRecovery.mScannedDevice.mAddr, filter_match->addr.addr,
```
Damian-Nordic, line 43: nit: it's a matter of preferece, but you could probably just assign the variables: `Instance().mRecovery.mScannedDevice.mAddr = filter_match->addr.addr`.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -105,10 +121,11 @@ void BLEConnectivityManager::ConnectionHandler(bt_conn *conn, uint8_t conn_err)
 
 void BLEConnectivityManager::DisconnectionHandler(bt_conn *conn, uint8_t reason)
 {
-	char addr[BT_ADDR_LE_STR_LEN];
-	bt_addr_le_to_str(bt_conn_get_dst(conn), addr, sizeof(addr));
+	if (!conn) {
```
Damian-Nordic, line 124: Wouldn't a nullptr indicate a bug? Or it can happen in normal circumstances?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -219,9 +246,35 @@ CHIP_ERROR BLEConnectivityManager::Init(bt_uuid **serviceUuids, uint8_t serviceU
...
+	return CHIP_NO_ERROR;
+}
+
 CHIP_ERROR BLEConnectivityManager::Scan(ScanDoneCallback callback, void *context, uint32_t scanTimeoutMs)
 {
-	if (mScanActive) {
+	if (callback != ReScanCallback) {
+		mRecovery.CancelTimer();
+		StopScan();
+		PrepareFilterForUuid();
```
Damian-Nordic, line 274: Should the error be checked?

Damian-Nordic (approved):

kkasperczyk-no (approved):

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -459,6 +505,13 @@ bt_le_conn_param *BLEConnectivityManager::GetScannedDeviceConnParams(bt_addr_le_
 	return nullptr;
 }
 
+char *BLEConnectivityManager::LogBTAddress(const bt_addr_le_t *addr)
+{
+	static char addrStr[BT_ADDR_LE_STR_LEN];
```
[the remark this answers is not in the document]
ArekBalysNordic: You're right. Currently, we don't use it in multiple threads, but someday we forget about it and use :) I think this method is not necessary here and I will remove it.

Damian-Nordic (approved):

PR #12305, opened 2023-09-12
sample: cellular: coap_client: Use network manager

fnawratil:
The coap_client sample now uses the network manager instead of directly interacting with the nRF91 modem. Thus the sample now also supports the nRF70 DK. Additionally printk statements have been converted to use the log system.

The sample can be moved to `samples/net` in an upcoming PR.


samples/cellular/coap_client/prj.conf
```diff
@@ -7,26 +7,25 @@
 CONFIG_HEAP_MEM_POOL_SIZE=4096
 CONFIG_NEWLIB_LIBC=y
 
+CONFIG_LOG=y
+CONFIG_LOG_BACKEND_UART=y
+CONFIG_LOG_MODE_IMMEDIATE=y
```
simensrostad, line 12: Any particular reason why you switched to immediate logging?

samples/cellular/coap_client/boards/nrf7002dk_nrf5340_cpuapp.conf
```diff
@@ -0,0 +1,45 @@
...
+
+# Configuration file for nRF7002 DK
+# This file is merged with prj.conf in the application folder, and options
+# set here will take precedence if they are present in both files.
+
+# General
+CONFIG_POSIX_CLOCK=y
+CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=4096
+CONFIG_LOG_BUFFER_SIZE=4096
+CONFIG_HEAP_MEM_POOL_SIZE=153600
```
lemrey, line 15: Why do we need 153kb of heap?

samples/cellular/coap_client/sample.yaml
```diff
@@ -3,7 +3,13 @@ sample:
 tests:
   sample.cellular.coap_client:
     build_only: true
+    build_on_all: true
     integration_platforms:
       - nrf9160dk_nrf9160_ns
```
simensrostad, line 8: Maybe add an entry for the 9161 while you are at it

samples/cellular/coap_client/src/main.c
```diff
@@ -220,41 +237,52 @@ static int wait(int timeout)
 	return 0;
 }
 
-int main(void)
+/**
+ * @brief Waits for the establishment of a network connection.
+ */
+static void wait_for_network(void)
+{
+	k_mutex_lock(&network_connected_lock, K_FOREVER);
```
simensrostad, line 245: Have you considered making is_connected an atomic variable?
https://docs.zephyrproject.org/latest/kernel/services/other/atomic.html#defining-an-atomic-variable
Removing the need to protect the variable with the mutex

then you should be able to get a way with using a simple sema? This might reduce the complexity a bit.

simensrostad (approved):
Lookin good. Just a few considerations

samples/cellular/coap_client/sample.yaml
fnawratil: Since I have no possibility to test that, this should be done in a separate PR, as discussed.

samples/cellular/coap_client/boards/nrf7002dk_nrf5340_cpuapp.conf
fnawratil: This seems to be necessary for the WiFi stack. Not sure about the exact size requirements, I've taken this value from `aws_iot` sample.

samples/cellular/coap_client/src/main.c
fnawratil: Keeping current solution, as discussed in person.

samples/cellular/coap_client/boards/nrf7002dk_nrf5340_cpuapp.conf
simensrostad: that is indeed due to extended heap usage of the wifi stack

samples/net/coap_client/src/main.c
```diff
@@ -0,0 +1,395 @@
...
+
+	/* Bring all network interfaces up.
+	 * Wi-Fi or LTE depending on the board that the sample was built for.
+	 */
+	LOG_INF("Bringing network interface up and connecting to the network");
+	err = conn_mgr_all_if_up(true);
+	if (err) {
+		LOG_ERR("conn_mgr_all_if_up, error: %d", err);
+		return err;
+	}
```
lemrey, line 386: Could we please add an explicit call to `conn_mgr_connect` ?

samples/net/coap_client/src/main.c
```diff
@@ -0,0 +1,395 @@
...
+
+		received = recv(sock, coap_buf, sizeof(coap_buf), MSG_DONTWAIT);
+		if (received < 0) {
+			if (errno == EAGAIN || errno == EWOULDBLOCK) {
+				LOG_ERR("socket EAGAIN");
+				continue;
+			} else {
+				LOG_ERR("Socket error...");
+				continue;
+			}
```
lemrey, line 311: I don't think we should handle `EAGAIN` differently than any other errors at this point, given that we checked the socket had POLLIN beforehand. I would remove one branch and print the `errno`.

samples/net/coap_client/src/main.c
KajaKoren: Thanks, should be fixed now!

samples/net/coap_client/src/main.c
KajaKoren: Also fixed!

KajaKoren (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -226,6 +226,13 @@ Multicore samples
 Networking samples
 ------------------
 
+* :ref:`_nrf_coap_client_sample` sample:
+
+  * Added support for Wi-Fi and LTE connectivity through the connection manager API.
+  * Updated by moving the sample from :file:`cellular/coap_client` folder to :file:`net/coap_client`.
+    The documentation is now found in the :ref:`networking_samples` section.
+
+
```
divipillai, line 235:
```suggestion
```
extra line.

divipillai (approved):

PR #12307, opened 2023-09-12
Deprecate AT%XRAI functionality

tokangas:
Deprecated legacy RAI functionality in LTE link control.

Updated the UDP sample to use `AT%RAI` instead. Refactored the sample a bit.


doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -625,6 +629,8 @@ Modem libraries
 
   * Updated:
 
+    * :c:func:`lte_lc_rai_req`, :c:func:`lte_lc_rai_param_set` and :kconfig:option:`CONFIG_LTE_RAI_REQ_VALUE` are now deprecated.
```
peknis, line 632:
```suggestion
    * The functions :c:func:`lte_lc_rai_req` and :c:func:`lte_lc_rai_param_set` and the Kconfig option :kconfig:option:`CONFIG_LTE_RAI_REQ_VALUE` are now deprecated.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -434,6 +434,10 @@ Cellular samples (renamed from nRF9160 samples)
     * The sample to use the :ref:`lib_caf` library instead of the :ref:`dk_buttons_and_leds_readme` library.
     * Displaying an error message when the sample fails to send an alert to nRF Cloud.
 
+* :ref:`udp` sample:
+
+   * Updated the sample to use ``AT%RAI`` for Release Assistance Indication (RAI) functionality.
```
peknis, line 439:
```suggestion
   * Updated the sample to use the AT command ``AT%RAI`` for Release Assistance Indication (RAI) functionality.
```

samples/cellular/udp/README.rst
```diff
@@ -64,45 +64,35 @@ CONFIG_UDP_DATA_UPLOAD_SIZE_BYTES - UDP data upload size configuration
...
-
 .. _CONFIG_UDP_RAI_ENABLE:
 
 CONFIG_UDP_RAI_ENABLE - RAI configuration
    This configuration option, if set, allows the sample to request RAI for transmitted messages.
 
 .. note::
-   PSM, eDRX and RAI value or timers are set using the configurable options for the :ref:`lte_lc_readme` library.
-
+   PSM and eDRX are configured using options for the :ref:`lte_lc_readme` library.
```
peknis, line 83:
```suggestion
   To configure PSM and eDRX, use the options from the :ref:`lte_lc_readme` library.
```

peknis (changes requested):

peknis (approved):

tokangas:
Declaring `lte_lc_rai_req()` and `lte_lc_rai_param_set()` as deprecated has an unwanted side effect in tests. When the LTE LC API is mocked, the mocks for these functions cause compilation warnings and those warnings are treated as errors. I'll see what could be done to work around that.

simensrostad (approved):

trantanen:
> Declaring `lte_lc_rai_req()` and `lte_lc_rai_param_set()` as deprecated has an unwanted side effect in tests. When the LTE LC API is mocked, the mocks for these functions cause compilation warnings and those warnings are treated as errors. I'll see what could be done to work around that.

Remove all uses of the deprecated functions from the samples and tests. That's what I was instructed to do when I deprecated multicell_location library.

trantanen:
> > Declaring `lte_lc_rai_req()` and `lte_lc_rai_param_set()` as deprecated has an unwanted side effect in tests. When the LTE LC API is mocked, the mocks for these functions cause compilation warnings and those warnings are treated as errors. I'll see what could be done to work around that.
> 
> Remove all uses of the deprecated functions from the samples and tests. That's what I was instructed to do when I deprecated multicell_location library.

Äh, I guess this is not enough as the generated mock will still have it. I guess there were some configs to tell whether specific function is mocked or not. I don't know if some of these could be used.

include/modem/lte_lc.h
```diff
@@ -1457,26 +1457,34 @@ int lte_lc_edrx_req(bool enable);
  *
  * For reference see 3GPP 24.301 Ch. 9.9.4.25.
  *
+ * @deprecated Use AT%RAI and socket options NRF_SO_RAI_* instead.
```
trantanen, line 1460: `%` sign not visible in generated documentation.

And then thinking aloud.
I'm still thinking if we should have a function for RAI... We should check if there are similar examples. Maybe it's enough that the use of AT%RAI is mentioned in the NRF_SO_RAI_* descriptions properly.

trantanen (approved):

lemrey:
You can use `FUNC_EXCLUDE` in `cmock_handle`, like so:
```
cmock_handle(myheader.h FUNC_EXCLUDE "foo")
```

include/modem/lte_lc.h
tokangas: I'll check the percentage sign.

@trantanen We could have a function for `AT%RAI`, one benefit would be that there could be a brief documentation how the feature is used. However, logical name for the function would be `lte_lc_rai_req()`, which currently uses `AT%XRAI` and is being deprecated. The function would have to be named differently or then we could just change the functionality, but not sure if that's ok.

@lemrey @simensrostad @jtguggedal  Do you have any insight, could we just update `lte_lc_rai_req()` to use the new RAI functionality? I'm just worried that developers could miss the change and not notice that they need to take the socket options into use. `lte_lc_rai_param_set()` would still be deprecated.

tokangas:
> You can use `FUNC_EXCLUDE` in `cmock_handle`, like so:
> 
> ```
> cmock_handle(myheader.h FUNC_EXCLUDE "foo")
> ```

Great. thanks!

include/modem/lte_lc.h
trantanen: If we think there are customers using RAI functionality with the existing `lte_lc_rai_req()` function, I'd be a bit reluctant in reusing the same function for the `AT%RAI` command.

tokangas: @trantanen I have a proposal for this. I don't think there's need to change `AT%RAI` value at runtime, so the logical thing to do is to add a new Kconfig option for it. The name can be `CONFIG_LTE_RAI_REQ`, which is unused and in line with other similar options. Both `lte_lc_rai_req()` and `lte_lc_rai_param_set()` can be deprecated and still use `AT%XRAI`.

trantanen: Without checking all details, sounds like a plan.

lemrey: I agree about the Kconfig option. If needed to be toggled at runtime, it's a one API call anyway; we could have a Kconfig option that sets it automatically. One question is whether the `AT%RAI` setting preserved on `CFUN` changes or is it reset?

tokangas: @lemrey `AT%RAI` setting is stored into modem NVM. I implemented this so, that `AT%RAI` command is always sent at boot time. So, depending on `CONFIG_LTE_RAI_REQ`, RAI is always explicitly enabled or disabled at boot. Also PSM and eDRX work like this.

tokangas:
I did some rework based on the discussion in the PR.

Kconfig options for PSM and eDRX in the UDP sample were restored, because those are used by the Online Power Profiler tool.

A new Kconfig option `CONFIG_LTE_RAI_REQ` was added in LTE link control. This can be used to enable RAI using the `AT%RAI` command. There's no runtime control for it.

SeppoTakalo (approved):

peknis (approved):

trantanen (approved):

MirkoCovizzi (approved):

PR #12310, opened 2023-09-12
manifest: update Find My revision

kapi-no:
Updated Find My revision to include the increase in the MCUboot partition size.


kapi-no:
We did not want to merge this Find My PR, as it turned out it was unnecessary. Anyway, the manifest will be bumped in the following PR that is maintained by @MarekPieta:

https://github.com/nrfconnect/sdk-nrf/pull/12329

PR #12312, opened 2023-09-12
lib: location: Fix possible buffer overflow

balaji-nordic:
- The sprintk operation was not caring for the terminating null
character and a comma.
- The strncat operation used to append a comma to the element string
was not caring for the terminating null character.

These are now fixed.

Found by clang.

### DIFF
--- a/lib/location/cloud_service/cloud_service_here_rest.c
+++ b/lib/location/cloud_service/cloud_service_here_rest.c
@@ -273,7 +273,7 @@ static int cloud_service_here_rest_pos_req_encode_cellular(
 	for (size_t i = 0; i < cell_data->ncells_count; i++) {
 		char element[NEIGHBOR_ELEMENT_SIZE];
 
-		len = snprintk(element, sizeof(element) - 1 /* ',' */,
+		len = snprintk(element, sizeof(element) - 2, /* Care for terminating '\0' and ',' */
 			       HTTP_REQUEST_BODY_NEIGHBOR_ELEMENT,
 			       cell_data->neighbor_cells[i].earfcn,
 			       cell_data->neighbor_cells[i].phys_cell_id,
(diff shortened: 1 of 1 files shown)


lib/location/cloud_service/cloud_service_here_rest.c
```diff
@@ -295,7 +295,7 @@ static int cloud_service_here_rest_pos_req_encode_cellular(
 
 		/* Append ',' if this is not the last element in the array */
 		if (i < (cell_data->ncells_count - 1)) {
-			strncat(element, ",", sizeof(element) - strlen(element));
+			strncat(element, ",", sizeof(element) - strlen(element) - 1);
```
trantanen, line 298: Comment related to `,` on line 276 should probably be removed as that doesn't really cover the comma but NUL termination, right?

trantanen (approved):

lib/location/cloud_service/cloud_service_here_rest.c
tokangas: Hmm, the `count` parameter for `strncat()` tells how many characters should be appended from `src`, not the maximum length of `dest`. So, the `count` parameter should be one.

balaji-nordic: @trantanen  Yes I believe thats the intention of the code too. If there is no opposition, I will fix the comment in this PR. 

@tokangas  The `n` in `strncat()` represents a maximum number of characters to be appended. It serves as a guard to prevent buffer overflow. I believe the intention here is to not append the `,` if the `element` is full (the `n` will evaluate to `0` in that case). So if the `n` is set to `1`, then it will result in an overflow when `element` is full.

balaji-nordic: Irrespective of that, I believe that the `strncat()` operation should be tolerant to code changes in other places in the function.

tokangas: Yeah, I guess it should be `snprintk(element, sizeof(element) - 2` instead, so that there should always be room for both comma and null. And `strncat()` is fine for me as it is now.

tokangas (approved):

lib/location/cloud_service/cloud_service_here_rest.c
trantanen: I guess this is fine although not 100% accurate because the last comma is not needed for the last element.
It's anyway code that is rarely if ever used in real end user application.

PR #12314, opened 2023-09-13
applications: serial_lte_modem: fix AT#XNRFCLOUDPOS

tomi-font:
A local variable named identically to a static variable caused the command to try to send location requests with no cellular nor Wi-Fi data.

Prefix all the static variables in that file with `nrfcloud_` to discern them better.


VTPeltoketo (approved):

MarkusLassila (approved):
Sorry, forgot to approve this yesterday. We should think more on the naming, but this is fine for now.

PR #12316, opened 2023-09-13
modules: nrfxlib: nrf_802154: RTC0 used for HW Task

ahasztag:
Instead of using an RTC1 channel for hardware tasks triggering used by 802.15.4, an RTC0 channel is used. This frees an RTC1 CC channel, allowing it to be used by Zephyr for other purposes.


ahasztag:
I assume that `cc_` functions (for example `cc_bind_to_ppi`) should be added a `hw_task_rtc_` prefix, probably it would be more readable. What do you think?

ahasztag:
Note that although RTC0 is used for scheduling the HW task, RTC1 is still used to perform all the timing checks (if we are not too late and so on). This can be done, as timers are synchronized, and allows to change less code (and not to add more functions which are missing for RTC0)

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -104,34 +117,134 @@ static bool hw_task_state_set(enum hw_task_state_type expected_state,
...
+		}
+	}
+
+	return 0;
+}
+#endif
+
+static int hw_task_rtc_timer_set(uint64_t fire_lpticks)
+{
+#if defined(CONFIG_SOC_NRF5340_CPUNET)
```
ankuns, line 231: Missing equivalent of too distant time check using 64-bit arith.
reference https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/nrf_rtc_timer.c#L347

Missing equivalent of attempt to set in the past (far past) using 64-bit arith
reference https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/timer/nrf_rtc_timer.c#L361

Currently attempt to set `hw_task_rtc_timer_set(z_nrf_rtc_timer_read() - (RTC_COUNTER_SPAN - 10))` would result in a RTC scheduled 10 ticks in future, while fully 64-bit value is clearly in the past.

Naive fix would be to read `z_nrf_rtc_timer_read()` once more, perform checks and then if conditions are ok call `zephyr_rtc_ticks_to_hw_task_rtc_counter` and `hw_task_rtc_counter_compare_set`.

Better would be to change prototype of `zephyr_rtc_ticks_to_hw_task_rtc_counter`
```c
static bool zephyr_rtc_ticks_to_hw_task_rtc_counter(uint64_t ticks, uint32_t * cc_value)
```
that would validate internally `zephyr_rtc_counter_now` variable against `ticks` parameter
return true only if ticks are in future but not further than RTC_COUNTER_HALF_SPAN

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -270,6 +385,7 @@ nrf_802154_sl_lptimer_platform_result_t nrf_802154_platform_sl_lptimer_hw_task_p
 		return NRF_802154_SL_LPTIMER_PLATFORM_NO_RESOURCES;
 	}
 
+#if !defined(CONFIG_SOC_NRF5340_CPUNET)
```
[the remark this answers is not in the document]
ahasztag: Ok @e-rk confirmed it can be probably moved, so I will add a separate commit with it.

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -104,34 +120,139 @@ static bool hw_task_state_set(enum hw_task_state_type expected_state,
...
 
 	return *(volatile uint32_t *)event_address;
 }
 
+static void hw_task_rtc_timer_abort(void)
+{
+#if defined(CONFIG_SOC_NRF5340_CPUNET)
+	/* No interrupt disabling/clearing is needed, as HW task does not use interrupts*/
+	nrf_rtc_event_clear(HW_TASK_RTC, NRF_RTC_CHANNEL_EVENT_ADDR(m_hw_task.chan));
+	nrf_rtc_event_disable(HW_TASK_RTC, NRF_RTC_CHANNEL_INT_MASK(m_hw_task.chan));
```
ankuns, line 160: nit: first disable, then clear.

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -104,34 +120,139 @@ static bool hw_task_state_set(enum hw_task_state_type expected_state,
...
+		barrier_isync_fence_full();
+		hw_task_rtc_counter_now_2 = nrf_rtc_counter_get(HW_TASK_RTC);
+
+	} while(hw_task_rtc_counter_now != hw_task_rtc_counter_now_2);
+
+	/* This function can only be called in close proximity of ticks */
+	assert(ticks >= zephyr_rtc_counter_now ||  zephyr_rtc_counter_now - ticks < RTC_COUNTER_HALF_SPAN);
+	assert(ticks <= zephyr_rtc_counter_now ||  ticks - zephyr_rtc_counter_now < RTC_COUNTER_HALF_SPAN);
+
+	uint32_t diff = (uint32_t) (hw_task_rtc_counter_now - zephyr_rtc_counter_now) & RTC_COUNTER_MAX;
```
ankuns, line 193: again you could use `rtc_counter_diff` if you had one

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -104,34 +120,139 @@ static bool hw_task_state_set(enum hw_task_state_type expected_state,
...
+		barrier_dmem_fence_full();
+		barrier_isync_fence_full();
+		hw_task_rtc_counter_now_2 = nrf_rtc_counter_get(HW_TASK_RTC);
+
+	} while(hw_task_rtc_counter_now != hw_task_rtc_counter_now_2);
+
+	/* This function can only be called in close proximity of ticks */
+	assert(ticks >= zephyr_rtc_counter_now ||  zephyr_rtc_counter_now - ticks < RTC_COUNTER_HALF_SPAN);
+	assert(ticks <= zephyr_rtc_counter_now ||  ticks - zephyr_rtc_counter_now < RTC_COUNTER_HALF_SPAN);
+
```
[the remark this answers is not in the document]
ahasztag: Although in general I agree that adding `rtc_counter_diff` and using this in several places is the right way to go, however using `uint32_t` values misses the point of these asserts. They are to insure that currently in close proximity of the scheduled HW task time - that we are not to far away both in the future or in the past. Changing to 32 bit value does not ensure this (not that both `zephyr_rtc_counter_now` and `ticks` are 64 bits variables.

Also note that the suggested assertion only checks that `ticks` is too far in the future. It does not check if we are too far in the past.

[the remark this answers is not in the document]
ahasztag: So in other words the role of this assert is to make sure the absolute value of the difference between 64-bit values `ticks` and `zephyr_rtc_counter_now` is smaller then `RTC_COUNTER_HALF_SPAN` - which is the requirement of using this API.

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -104,34 +120,139 @@ static bool hw_task_state_set(enum hw_task_state_type expected_state,
...
+	nrf_rtc_event_clear(HW_TASK_RTC, NRF_RTC_CHANNEL_EVENT_ADDR(m_hw_task.chan));
+
+	nrf_rtc_cc_set(HW_TASK_RTC, m_hw_task.chan, cc_value);
+
+	nrf_rtc_event_enable(HW_TASK_RTC, NRF_RTC_CHANNEL_INT_MASK(m_hw_task.chan));
+
+	current_counter_value = nrf_rtc_counter_get(HW_TASK_RTC);
+
+	if (((cc_value - (current_counter_value + RTC_MIN_CYCLES_FROM_NOW)) & RTC_COUNTER_MAX)
+		> (RTC_COUNTER_HALF_SPAN - RTC_MIN_CYCLES_FROM_NOW))
```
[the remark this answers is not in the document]
ahasztag: Done

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
ahasztag: Fixed

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
ahasztag: Done, although in this case it is not as pretty, because % had to be used to convert `zephyr_rtc_counter_now` to a 32bit value.

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -11,12 +11,25 @@
...
+#include <zephyr/sys/barrier.h>
+#include <hal/nrf_rtc.h>
 #include <helpers/nrfx_gppi.h>
 
 #include "platform/nrf_802154_clock.h"
 #include "nrf_802154_sl_utils.h"
 
 #define RTC_CHAN_INVALID (-1)
+#if defined(CONFIG_SOC_NRF5340_CPUNET)
+#define HW_TASK_RTC	     NRF_RTC0
```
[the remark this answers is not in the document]
e-rk: I think dependencies like this should be handled through kconfig. You can and `depends on MPSL` to srcx equivalent of this file: https://github.com/nrfconnect/sdk-nrfxlib/blob/main/nrf_802154/zephyr/Kconfig.nrfxlib#L48

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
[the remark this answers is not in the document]
ankuns: You are right, I missed the 64-bit checks here.

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -104,34 +117,134 @@ static bool hw_task_state_set(enum hw_task_state_type expected_state,
...
+	return z_nrf_rtc_timer_exact_set(m_hw_task.chan, fire_lpticks, NULL, NULL);
+#endif
+}
+
 void nrf_802154_platform_sl_lp_timer_init(void)
 {
 	m_in_critical_section = false;
 	m_hw_task.state = HW_TASK_STATE_IDLE;
+#if defined(CONFIG_SOC_NRF5340_CPUNET)
+	m_hw_task.chan = HW_TASK_RTC_CHAN;
```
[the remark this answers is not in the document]
e-rk: A agree with @ahasztag. I don't think there is too much value in adding more `#ifdefs` to save 4 bytes at most.

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
e-rk: Agreed with @ahasztag. The HW tasks are always scheduled shortly before radio operation. Having half RTC span of margin is more than enough.

e-rk (commented):
Looking very solid. I just have one nitpick.

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -98,40 +114,150 @@ static void timer_start_at(struct timer_desc *timer, uint64_t target_time)
...
+	 * If the two reads are different, the procedure is repeated.
+	 * This is to make sure that no RTC tick occured between reading RTC0 and
+	 * RTC1 and so the calculated difference between the timers is correct.
+	 */
+	do
+	{
+		hw_task_rtc_counter_now = nrf_rtc_counter_get(HW_TASK_RTC);
+		zephyr_rtc_counter_now = z_nrf_rtc_timer_read();
+		barrier_dmem_fence_full();
+		barrier_isync_fence_full();
```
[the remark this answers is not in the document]
ahasztag: True, I will delete it.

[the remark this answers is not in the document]
ahasztag: Done

ankuns:
We don't use user CC channel from nrf_rtc_timer.c anymore for `CONFIG_SOC_NRF5340_CPUNET`.
Please add (possibly as a separate commit) setting of `CONFIG_NRF_RTC_TIMER_USER_CHAN_COUNT` to 2 for `CONFIG_SOC_NRF5340_CPUNET`.

Please note that ocurrences are present not only in sdk-nrf but also in sdk-zephyr (upstream zephyr), please take into account internal test repo.

ahasztag:
> We don't use user CC channel from nrf_rtc_timer.c anymore for `CONFIG_SOC_NRF5340_CPUNET`. Please add (possibly as a separate commit) setting of `[CONFIG_NRF_RTC_TIMER_USER_CHAN_COUNT](https://github.com/nrfconnect/sdk-nrf/pull/12316)` to 2 for `CONFIG_SOC_NRF5340_CPUNET`.
> 
> Please note that ocurrences are present not only in sdk-nrf but also in sdk-zephyr (upstream zephyr), please take into account internal test repo.

Updated it for sdk-nrf, I have also created an appropriate task for other repositories.

samples/peripheral/802154_phy_test/prj.conf
```diff
@@ -26,7 +26,7 @@ CONFIG_NRFX_TIMER2=y
 CONFIG_NRF_802154_RADIO_DRIVER=y
 
 # Allow sharing the RTC between IEEE 802.15.4 and Zephyr
-CONFIG_NRF_RTC_TIMER_USER_CHAN_COUNT=3
+CONFIG_NRF_RTC_TIMER_USER_CHAN_COUNT=2
```
ankuns, line 29: This is incorrect for platforms other than `CONFIG_SOC_NRF5340_CPUNET`
This is correct only for `CONFIG_SOC_NRF5340_CPUNET`

samples/peripheral/802154_sniffer/prj.conf
```diff
@@ -5,7 +5,7 @@ CONFIG_IEEE802154_RAW_MODE=y
 CONFIG_NET_PKT_TIMESTAMP=y
 
 # Allow sharing the RTC between IEEE 802.15.4 and Zephyr
-CONFIG_NRF_RTC_TIMER_USER_CHAN_COUNT=3
+CONFIG_NRF_RTC_TIMER_USER_CHAN_COUNT=2
```
ankuns, line 8: Again valid for `CONFIG_SOC_NRF5340_CPUNET` only, not in general.

ankuns (changes requested):
Having different requirements on CC channels for each platform brings us to trouble.

samples/peripheral/802154_phy_test/prj.conf
ahasztag: Fixed

samples/peripheral/802154_sniffer/prj.conf
ahasztag: True, reverted.

However the `nrf5340dk_nrf5340_cpunet` board is not supported for this sample, see https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/peripheral/802154_sniffer/README.html

So no changes should be introduced here.

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -98,40 +114,150 @@ static void timer_start_at(struct timer_desc *timer, uint64_t target_time)
 	nrf_802154_sl_mcu_critical_exit(state);
 }
 
+static inline uint32_t rtc_counter_diff(uint32_t a, uint32_t b)
+{
+	return (a - b) & RTC_COUNTER_MAX;
```
ankuns, line 119: `RTC_COUNTER_MAX` is provided only under `#if defined(CONFIG_SOC_NRF5340_CPUNET)`, doesn't build

modules/nrfxlib/nrf_802154/sl/platform/nrf_802154_platform_sl_lptimer_zephyr.c
```diff
@@ -98,40 +114,150 @@ static void timer_start_at(struct timer_desc *timer, uint64_t target_time)
...
+}
```
ankuns, line 120: RTC_COUNTER_MAX is provided only under #if defined(CONFIG_SOC_NRF5340_CPUNET), doesn't build
```suggestion
#if defined(CONFIG_SOC_NRF5340_CPUNET)
static inline uint32_t rtc_counter_diff(uint32_t a, uint32_t b)
{
	return (a - b) & RTC_COUNTER_MAX;
}
#endif
```

ankuns (approved):

PR #12317, opened 2023-09-13
nrf_security: Add missing PSA want key types

joerchan:
The following key types are always enabled in the Oberon PSA core.
Set them as enabled in the PSA configuration header to reflect this.
    
NCSDK-22424

Update zephyr to include missing PSA want key types.
These may be needed to include the TF-M key module which depends on
the PSA_HAS_KEY_TYPE_SUPPORT configuration which cannot be set by the
application directly.


mswarowsky (approved):
lgtm

subsys/nrf_security/configs/psa_crypto_config.h.template
```diff
@@ -75,16 +75,21 @@
 #cmakedefine PSA_WANT_ECC_SECP_R1_384                           @PSA_WANT_ECC_SECP_R1_384@
 #cmakedefine PSA_WANT_ECC_SECP_R1_521                           @PSA_WANT_ECC_SECP_R1_521@
 
-#cmakedefine PSA_WANT_KEY_TYPE_DERIVE                           @PSA_WANT_KEY_TYPE_DERIVE@
-#cmakedefine PSA_WANT_KEY_TYPE_HMAC                             @PSA_WANT_KEY_TYPE_HMAC@
+/* The following key types are not configurable in the Oberon PSA implementation */
```
SebastianBoe, line 78: How are we enforcing that we don't get a mismatch between KConfig and these C defines?

de-nordic:
Need rebase

SebastianBoe (changes requested):
changes requested

SebastianBoe (approved):

joerchan:
Dropped manifest commit, changes already included in another manifest update to sdk-zephyr.

subsys/nrf_security/configs/psa_crypto_config.h.template
joerchan: offline discussion:
Decided to forward the Kconfig instead. Having these features accepted without the PSA_WANT symbol set to enabled is acceptable.

PR #12327, opened 2023-09-14
bluetooth: controller: Fix command complete

Tschet1:
When HCI commands for legacy and extended commands are being used, a Command Disallowed error should be produced. In this case, the command complete event was missing the other (command specific) parameters in the event.

Added them.

### DIFF
--- a/subsys/bluetooth/controller/hci_internal.c
+++ b/subsys/bluetooth/controller/hci_internal.c
@@ -22,6 +22,12 @@
 				+ sizeof(struct bt_hci_evt_cmd_complete) \
 				+ sizeof(struct bt_hci_evt_cc_status))
 
+enum type_of_adv_cmd {
+	ADV_COMMAND_TYPE_NONE,
+	ADV_COMMAND_TYPE_LEGACY,
+	ADV_COMMAND_TYPE_EXTENDED,
+};
+
 static struct
 {
 	bool occurred; /**< Set in only one execution context */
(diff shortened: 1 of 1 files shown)


subsys/bluetooth/controller/hci_internal.c
```diff
@@ -1319,21 +1438,21 @@ int hci_internal_cmd_put(uint8_t *cmd_in)
...
+		 * up in this case.
 		 */
 		if (command_generates_command_complete_event(opcode)) {
-			uint8_t param_length = sizeof(struct bt_hci_evt_cmd_complete)
-					     + sizeof(struct bt_hci_evt_cc_status);
-			(void)encode_command_complete_header(cmd_complete_or_status.raw_event,
-							     opcode,
-							     param_length,
-							     BT_HCI_ERR_CMD_DISALLOWED);
+			generate_status_disallowed_command_complete_event(cmd_in, opcode);
```
minkrg, line 1450: I might be missing something, but the opcodes in `command_generates_command_complete_event ` are less than the ones in the newly added function `generate_status_disallowed_command_complete_event `
Shouldn't they be the same?

de-nordic:
@rugeGerritsen @minkrg Please review.

rugeGerritsen (changes requested):
Discussed offline. Will see if we can combine with something in `is_host_using_legacy_and_extended_commands()` to reduce code size

subsys/bluetooth/controller/hci_internal.c
```diff
@@ -55,51 +61,100 @@ static bool command_generates_command_complete_event(uint16_t hci_opcode)
...
 #else
 	/* A host is not allowed to use both legacy and extended HCI commands.
 	 * See Core v5.1, Vol2, Part E, 3.1.1 Legacy and extended advertising
 	 */
-	static enum {
-		ADV_COMMAND_TYPE_NONE,
-		ADV_COMMAND_TYPE_LEGACY,
-		ADV_COMMAND_TYPE_EXTENDED,
-	} type_of_adv_cmd_used_since_reset;
+	static enum type_of_adv_cmd type_of_adv_cmd_used_since_reset;
```
rugeGerritsen, line 114: You could consider explicitly initializing `type_of_adv_cmd_used_since_reset` to `ADV_COMMAND_TYPE_NONE`

rugeGerritsen (approved):

minkrg (approved):

PR #12334, opened 2023-09-15
applications: Improved UI for handling BT devices in Matter bridge

kkasperczyk-no:
Introduced several changes that should improve UX, while adding BLE devices to Matter bridge:
* Added printing of UUID to the scanned devices list
* Removed device type argument from matter_bridge add command, as the device type is not automatically mapped based on used BLE service
* Moved bridged_device_factory and bridged_devices_creator for simulated and ble devices into separate files to improve the code readability.
* Removed using optionals from CreateDevice method for BLE devices, as it was confusing. Instead created a two separate methods for creating devices based on Matter device type and based on BLE service.


applications/matter_bridge/README.rst
```diff
@@ -140,11 +140,12 @@ Getting a list of Bluetooth LE devices available to be added
...
-      I: Scan result:
-      I: -----------------------------
-      I: | Index |      Address      |
-      I: -----------------------------
-      I: | 0     | e6:11:40:96:a0:18 |
+      Scan result:
+      ---------------------------------------------------------------------
+      | Index |      Address      |                   UUID
+      ---------------------------------------------------------------------
+      | 0     | e6:11:40:96:a0:18 | 0x181a (Environmental Sensing Service)
```
doublemis1, line 147: Does the UUID directly map to a BLE device?
so the (Environmental Sensing Service) is not needed if the `0x181a` map to only that device

applications/matter_bridge/README.rst
```diff
@@ -346,15 +341,16 @@ After building the sample and programming it to your development kit, complete t
 
                uart:~$ matter_bridge scan
 
-            The terminal output is similar to the following one:
+            The terminal output is similar to the following one (in case of using a single Bluetooth LE device only one entry will be available):
```
wiba-nordic, line 344:
```suggestion
            The terminal output is similar to the following one:
```
Since we're already saying it's similar, it is not necessary to say anything about the number. If you feel it is necessary, I'd use the following instead:
```suggestion
            The terminal output is similar to the following one, with an entry for each connected Bluetooth LE device:
```

applications/matter_bridge/README.rst
```diff
@@ -363,24 +359,22 @@ After building the sample and programming it to your development kit, complete t
...
                :class: highlight
 
-               uart:~$ matter_bridge add *<device_type>* *<bluetooth_device_index>*
+               uart:~$ matter_bridge add *<bluetooth_device_index>*
 
-            The *<device_type>* is a Matter device type that will be used to represent a new bridged device in the Matter Data Model.
-            See the description in :ref:`matter_bridge_cli` for the list of supported values.
             The *<bluetooth_device_index>* is a Bluetooth LE device index that was scanned in the previous step.
-            For example, if you want to add a new Bluetooth LE bridged device with index `0` that is exposed as an On/Off Light endpoint, use the following command:
+            For example, if you want to add a new Bluetooth LE bridged device with index `1`, use the following command:
```
wiba-nordic, line 365:
```suggestion
            For example, if you want to add a new Bluetooth LE bridged device with index ``1``, use the following command:
```
Just a formatting difference since we are talking about a specific value that is going to be input.

wiba-nordic (changes requested):
A couple of small nits from me on the docs.

Damian-Nordic (approved):

applications/matter_bridge/README.rst
kkasperczyk-no: I'm not sure I understand your question. The UUID describes BLE profile used by the particular device, but it is not unique for each and every device (it will be the same for all BLE devices using e.g LBS). Regarding the log output I added text description of the profile in brackets, so user would not have to look for the meaning of hex number of UUID.

doublemis1: ok, thank you.

wiba-nordic (approved):

wiba-nordic:
Removed the `changelog-entry-required` label since this is a change to a feature that already has an entry for being added.

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
```diff
@@ -0,0 +1,428 @@
...
+			return CHIP_ERROR_BUFFER_TOO_SMALL;
+		}
+		deviceType[0] = MatterBridgedDevice::DeviceType::OnOffLight;
+		count = 1;
+	} break;
+	case BleBridgedDeviceFactory::ServiceUuid::EnvironmentalSensorService: {
+		if (maxCount < 2) {
+			return CHIP_ERROR_BUFFER_TOO_SMALL;
+		}
+		deviceType[0] = MatterBridgedDevice::DeviceType::TemperatureSensor;
```
ArekBalysNordic, line 51: NIT: If we use this method to get the Matter Device Type, how we will know that the Temperature Sensor is at index 0 and the Humidity Sensor is at index 1 without looking at the code of this method?

ArekBalysNordic (approved):

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
kkasperczyk-no: And why we would like to know it? It's an internal method to help mapping one type to another, so we would know what device types should be created and assigned to endpoints. I don't think the order of indexes really matter.

PR #12338, opened 2023-09-15
mpsl: coex: disable granted interrupt on callback deregister

ahasztag:
This commit makes calling the .p_register_callback (called by MPSL) disable the "granted" interrupt if NULL pointer is passed as an argument (deregistering callback). The interrupt is enabled if a non-NULL pointer is passed.

If all the protocols deregister their callback when they no longer use the radio (after ending their timeslot), then this will disable the interrupt during sleep periods. The processor will no longer be waken up on (possibly spurious) changes of the GRANT pin.

### DIFF
--- a/subsys/mpsl/cx/nrf700x/mpsl_cx_nrf700x.c
+++ b/subsys/mpsl/cx/nrf700x/mpsl_cx_nrf700x.c
@@ -26,6 +26,7 @@
 #include <zephyr/devicetree.h>
 
 #include "hal/nrf_gpio.h"
+#include <nrfx_gpiote.h>
 
 /*
  * Typical part of device tree describing coex (sample port and pin numbers).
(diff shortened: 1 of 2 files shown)


subsys/mpsl/cx/nrf700x/mpsl_cx_nrf700x.c
```diff
@@ -219,9 +219,19 @@ static uint32_t req_grant_delay_get(void)
...
+	int32_t ret = 0;
 	callback = cb;
 
-	return 0;
+	if (cb != NULL) {
+		ret = gpio_pin_interrupt_configure_dt(&grant_spec,
+			GPIO_INT_ENABLE | GPIO_INT_EDGE | GPIO_INT_EDGE_BOTH);
+	}
+	else
+	{
```
ankuns, line 230: compliance check fail, please try:
```suggestion
	} else {
```
and in other places as well

ankuns (approved):

martintv (approved):

subsys/mpsl/cx/thread/mpsl_cx_thread.c
```diff
@@ -165,6 +170,12 @@ static int32_t register_callback(mpsl_cx_cb_t cb)
 {
 	callback = cb;
 
+	if (cb != NULL) {
+		nrfx_gpiote_trigger_enable(grant_abs_pin, true);
+	} else {
+		nrfx_gpiote_trigger_disable(grant_abs_pin);
+	}
```
ankuns, line 177: It's OK, just leaving a comment to document.
The reason why `gpio_pin_interrupt_configure_dt` is not called here but lower level api is uset is that the `register_callback` function is used in time critical procedures that cannot tolerate so time consuming call like `gpio_pin_interrupt_configure_dt`.

ankuns (approved):

PR #12347, opened 2023-09-18
samples: cellular: modem_shell: Rework UART suspend handling

tokangas:
Reworked how UARTs are put into low power state.

The shell UART is used in asynchronous mode is suspended when needed using the Zephyr PM API. The DT configuration is now used to find the correct UART.

Asynchronous UARTs are configured into low power mode by default. For example, modem UART tracing uses the UART in asynchronous mode. In low power mode, the UART driver disables the device when it's not active.

MOSH-485


moksanen (approved):
Based on a couple of current measurement test runs in CI, this seems to solve the issues when snippet is used to activate UART traces, and "uart disable" is called in modem_shell.

trantanen (approved):

PR #12350, opened 2023-09-18
wifi: Add hostap as zephyr module

krish2718:


sachinthegreen (approved):

krish2718:
This is only for Zephyr modules, but in NCS `hostap` is a nRF module, so, this isn't applicable and only needed for upstreaming `hostap` but that is done using a new repo, so, abandoning this PR.

PR #12355, opened 2023-09-19
lib: location: Change NCELLMEAS type from ext_comp to ext_light

trantanen:
For some reason, we decided to use NCELLMEAS search_type 2 (extended complete) for normal neighbor cells.
At the same time we used search type 4 (extended light) for GCI search.
Because "extended complete" can take a lot of time for normal and GCI cells, we are changing normal neighbor search also to use "extended light" search type.

Jira: NCSDK-23454


tokangas (approved):
The commit and PR descriptions say that "extended complete" search is used for GCI search, that should be "extended light". Otherwise looks good to me.

jhirsi (approved):

PR #12356, opened 2023-09-19
samples: matter: Fixed handling bridged device reachable attribute

kkasperczyk-no:
The current implementation returns failure if the device is not reachable and read request is received. It applied also to the reachable attribute itself.

Moved handling attributes common to all bridged devices to the bridge_manager, before validating the reachable state.


ArekBalysNordic (approved):

doublemis1:
Can we add any information in documentation about this behavior,
if not reachable then you can get x, y, z values, the rest will send report with Failure(0x01) status

kkasperczyk-no:
> Can we add any information to documentation about this behavior, if not reachable then you can get x, y, z values, the rest will send report with Failure(0x01) status

Yeah, I planned to keep it in mind and add in another task about adding documentation guide. I updated the KRKNWK-17270 task to do not forget about it. Is it ok for you?

doublemis1 (approved):

PR #12359, opened 2023-09-19
applications: serial_lte_modem: support A-GPS and P-GPS combined

tomi-font:
See the changelog entries for more details.

### DIFF
--- a/applications/serial_lte_modem/doc/GNSS_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/GNSS_AT_commands.rst
@@ -7,53 +7,101 @@ GNSS AT commands
    :local:
    :depth: 2
 
-The following commands list contains GNSS-related AT commands.
+The following list contains GNSS-related AT commands.
 
-GNSS
-====
+Control GNSS
+============
 
 The ``#XGPS`` command controls the GNSS.
 
 Set command
 -----------
 
-The set command allows you to start and stop the GNSS.
+The set command allows you to start and stop the GNSS module.
+
+Both the :ref:`lib_nrf_cloud_agps` service and the :ref:`lib_nrf_cloud_pgps` service can be used with the module, either together or independently of each other.
+Using them reduces the time it takes the GNSS module to estimate its position.
+
+To use either of them, the device must be connected to nRF Cloud (using the :ref:`#XNRFCLOUD <SLM_AT_NRFCLOUD>` AT command) when starting the GNSS module.
+In addition, the following Kconfig options must be enabled:
+
+   * :kconfig:option:`CONFIG_NRF_CLOUD_AGPS <CONFIG_NRF_CLOUD_AGPS>` to use A-GNSS.
+   * :kconfig:option:`CONFIG_NRF_CLOUD_PGPS <CONFIG_NRF_CLOUD_PGPS>` to use P-GPS.
+
+If both assistive services were enabled during compilation, you cannot choose to use only one of them at run time.
 
 Syntax
 ~~~~~~
 
 ::
 
-   #XGPS=<op>[,<interval>[,<timeout>]]
+   #XGPS=<op>,<cloud_assistance>,<interval>[,<timeout>]
 
-The ``<op>`` parameter accepts the following integer values:
+The ``<op>`` parameter can have the following integer values:
 
-* ``0`` - Stop GNSS
-* ``1`` - Start GNSS
+* ``0`` - Stop the GNSS module.
+  In this case, no other parameter is allowed.
+* ``1`` - Start the GNSS module.
 
-The ``<interval>`` parameter represents the GNSS fix interval in seconds.
-It must be set when starting the GNSS.
-It accepts the following integer values:
+``<cloud_assistance>`` is an integer that indicates whether to use the nRF Cloud assistive services that were enabled during compilation.
+It is ``0`` for disabled or ``1`` for enabled.
+
+``<interval>`` is an integer that indicates the GNSS fix interval in seconds.
+It can have one of the following values:
 
 * ``0`` - Single-fix navigation mode.
 * ``1`` - Continuous navigation mode.
-  The fix interval is set to 1 second
+  The fix interval is set to 1 second.
 * Ranging from ``10`` to ``65535`` - Periodic navigation mode.
   The fix interval is set to the specified value.
 
-In periodic navigation mode, the ``<timeout>`` parameter controls the maximum time in seconds that the GNSS receiver is allowed to run while trying to produce a valid PVT estimate.
-In continuous navigation mode, this parameter does not have any effect.
-It accepts the following integer values:
+In continuous navigation mode, the ``<timeout>`` parameter must be omitted.
+In single-fix and periodic navigation modes, the ``<timeout>`` parameter indicates the maximum time in seconds that the GNSS receiver is allowed to run while trying to produce a valid Position, Velocity, and Time (PVT) estimate.
+It can be one of the following:
 
 * ``0`` - The GNSS receiver runs indefinitely until a valid PVT estimate is produced.
-* Any positive integer lower than the ``<interval>`` value - The GNSS receiver is turned off after the specified time is up, even if a valid PVT estimate was not produced.
+* Any positive integer - The GNSS receiver is turned off after the specified time is up, even if a valid PVT estimate was not produced.
+* Omitted - In single-fix or periodic navigation mode, the timeout defaults to 60 seconds.
 
-When not specified, it defaults to a timeout value of 60 seconds.
+.. note::
+
+   When ``<cloud_assistance>`` is disabled, no request is made to nRF Cloud for assistance data.
+   However, if it has been previously enabled and used, such data may remain locally and will be used if still valid.
+
+.. note::
+
+   In periodic navigation mode, the ``<interval>`` and ``<timeout>`` parameters are temporarily ignored during the first fix (for up to 60 seconds) and when the GNSS module determines it needs to download ephemerides or almanacs from the broadcast.
+
+.. note::
+
+   Make sure that the GNSS antenna is configured properly.
+   This can be achieved two ways:
+
+   * Using the :kconfig:option:`CONFIG_MODEM_ANTENNA_AT_MAGPIO` and :kconfig:option:`CONFIG_MODEM_ANTENNA_AT_COEX0` Kconfig options.
+   * By issuing the ``%XMAGPIO`` and ``%XCOEX0`` AT commands manually at run time.
+
+.. tip::
+
+   When the LTE link is enabled, make sure to have either Power Saving Mode (PSM) or extended Discontinuous Reception (eDRX) enabled to give the GNSS receiver the time it needs to acquire fixes.
+
+.. tip::
+
+   Enable the :kconfig:option:`CONFIG_SLM_LOG_LEVEL_DBG` Kconfig option if you have trouble acquiring fixes.
+   It makes the application print NMEA and PVT data when trying to acquire fixes, which can be of help when solving the issue.
+
+.. note::
+
+   See the documentation for the :ref:`lib_nrf_cloud_agps` and :ref:`lib_nrf_cloud_pgps` libraries for information on how to best configure and use A-GNSS and P-GPS, respectively.
+
+.. note::
+
+   When using P-GPS, make sure that the value of the :kconfig:option:`CONFIG_SLM_PGPS_INJECT_FIX_DATA` Kconfig option matches your use case.
+   It is enabled by default but should be disabled if the device is expected to move distances longer than a few dozen kilometers between fix attempts.
+
+As an alternative to GNSS-based positioning, see :ref:`#XNRFCLOUDPOS <SLM_AT_NRFCLOUDPOS>` for cellular and Wi-Fi positioning.
 
 Unsolicited notification
 ~~~~~~~~~~~~~~~~~~~~~~~~
-
-.. gps_pos_notif_start
 
 .. code-block::
 
@@ -67,18 +115,17 @@ Unsolicited notification
 * The ``<heading>`` value represents the heading of the movement of the user in degrees.
 * The ``<datetime>`` value represents the UTC date-time.
 
-.. gps_pos_notif_end
-
 .. gps_status_notif_start
 
 .. code-block::
 
    #XGPS: <gnss_service>,<gnss_status>
 
-* The ``<gnss_service>`` value is an integer.
-  When it returns the value of ``1``, it means that GNSS is supported in ``%XSYSTEMMODE`` and activated in ``+CFUN``.
+* The ``<gnss_service>`` parameter is an integer.
+  When it has the value ``1``, it means that GNSS is supported in ``%XSYSTEMMODE`` and activated in ``+CFUN``.
 
-* The ``<gnss_status>`` value is an integer.
+* The ``<gnss_status>`` parameter is an integer.
+  It can have the following values:
 
   * ``0`` - GNSS is stopped.
   * ``1`` - GNSS is started.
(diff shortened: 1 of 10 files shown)


SeppoTakalo (approved):

applications/serial_lte_modem/doc/NRFCLOUD_AT_commands.rst
```diff
@@ -198,8 +200,8 @@ The request uses information from the cellular network, Wi-Fi access points, or
 .. note::
    To use ``#XNRFCLOUDPOS``, the following preconditions apply:
 
-   * You must define :ref:`CONFIG_SLM_NRF_CLOUD <CONFIG_SLM_NRF_CLOUD>` and :kconfig:option:`CONFIG_NRF_CLOUD_LOCATION <CONFIG_NRF_CLOUD_LOCATION>`.
-   * You must be connected to nRF Cloud (using ``#XNRFCLOUD``).
+   * The device must be connected to nRF Cloud using :ref:`#XNRFCLOUD <SLM_AT_NRFCLOUD>`.
+   * :kconfig:option:`CONFIG_NRF_CLOUD_LOCATION <CONFIG_NRF_CLOUD_LOCATION>` must be enabled.
```
wiba-nordic, line 204:
```suggestion
   * The :kconfig:option:`CONFIG_NRF_CLOUD_LOCATION <CONFIG_NRF_CLOUD_LOCATION>` Kconfig option must be enabled.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -210,6 +210,7 @@ Serial LTE modem
   * An ``auto_connect`` operation in the ``#XCARRIER`` carrier command.
     The operation controls automatic registration of UE to LTE network.
   * ``#XNRFCLOUDPOS`` AT command to send location requests to nRF Cloud using cellular or Wi-Fi positioning, or both.
+  * Support to use A-GPS and P-GPS at the same time.
```
wiba-nordic, line 213:
```suggestion
  * Support for using A-GPS and P-GPS at the same time.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -228,6 +229,8 @@ Serial LTE modem
   * Operations to read or erase the MCUboot secondary slot from the ``#XFOTA`` AT command because the application update process overwrites the slot in any case.
   * AT commands ``#XCELLPOS`` and ``#XWIFIPOS``.
     They are replaced by the ``#XNRFCLOUDPOS`` command that allows to combine cellular and Wi-Fi data to determine the device location.
+  * AT commands ``#XAGPS`` and ``#XPGPS``.
```
wiba-nordic, line 232:
```suggestion
  * The AT commands ``#XAGPS`` and ``#XPGPS``.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -228,6 +229,8 @@ Serial LTE modem
...
+    Their functionality is merged into the ``#XGPS`` AT command that now allows to use A-GPS and P-GPS at the same time.
```
wiba-nordic, line 233:
```suggestion
    Their functionality is merged into the ``#XGPS`` AT command that now allows using A-GPS and P-GPS at the same time.
```

wiba-nordic (changes requested):

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -488,247 +593,106 @@ int handle_at_gps(enum at_cmd_type cmd_type)
...
-			/* no scheduled downloads for peoridical tracking */
-			if (interval >= 10) {
-				err = nrf_modem_gnss_use_case_set(
-						NRF_MODEM_GNSS_USE_CASE_MULTIPLE_HOT_START |
-						NRF_MODEM_GNSS_USE_CASE_SCHED_DOWNLOAD_DISABLE);
-				if (err) {
-					LOG_ERR("Failed to set use case, error: %d", err);
-					return err;
-				}
+			if (gnss_pgps_enabled && !IS_ENABLED(CONFIG_NRF_CLOUD_PGPS)) {
```
MarkusLassila, line 626: I am wondering whether we need to support the different options in runtime as the PGPS and AGPS availability is decided in the compile time. It would be more straightforward to follow the same compile time flagging and offer only the operation that is flagged in.

I am also not sure if the operation that we have for example, with AGPS + PGPS, but only requested for AGPS is equivalent to compilation with AGPS + requested AGPS. I am sure that SLM codes are equivalent, but in other components there are differences.

applications/serial_lte_modem/doc/GNSS_AT_commands.rst
```diff
@@ -7,54 +7,97 @@ GNSS AT commands
...
 The ``#XGPS`` command controls the GNSS.
 
 Set command
 -----------
 
-The set command allows you to start and stop the GNSS.
+The set command allows to start and stop the GNSS module.
+
+It can be used with or without :ref:`lib_nrf_cloud_agps` and :ref:`lib_nrf_cloud_pgps`.
+Both assistive services can also be used independently of each other.
```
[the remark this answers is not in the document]
wiba-nordic: Alright, I checked the libraries' own documentation, and they do refer to "nRF Cloud A-GPS service" and "nRF Cloud P-GPS service".
```suggestion
Both the :ref:`lib_nrf_cloud_agps` service and the :ref:`lib_nrf_cloud_pgps` service can be used with the module, either together or independently of each other.
```

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
MarkusLassila: > By testing and observing the answer is no. Especially in the case you mention. When both A-GPS and P-GPS are compiled in, A-GPS starts relying on P-GPS for processing some of the data. That's why I put some countermeasures in that code, to try to not inject the data from one service (A-GPS/P-GPS) if it is not requested in the AT command. But it may be a foolish endeavor in the end. At least one that is pretty solid and good to keep I think is allowing to not use any of the two services at run time even though one or both have been compiled in. By ignoring events and not triggering any related action it's possible (and the case here) to have the libraries not do anything.

Having countermeasures to hide some of the existing functionality might be problematic in case there are real differences in some scenarios, or if the GNSS libraries change functionality, so that there would be.

Having a possibility to ignore both, would be pretty much similar functionality that we had previously, where you could send GPS, AGPS and PGPS requests. I am pretty certain that neither AGPS or PGPS worked correctly if both were active, so that possibility did not really exist.

MarkusLassila: @VTPeltoketo, @SeppoTakalo, @junqingzou: What are your takes on this?

wiba-nordic (approved):
Looks good so I'm approving, though leaving a comment that's more me musing out loud.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
VTPeltoketo: I would say that if you have both AGPS and PGPS compiled, you can activate only both at the same time. Still, I'd like got get opinion from some GNSS-expert.

VTPeltoketo: So, according to my understanding we have three options when AGPS and PGPS are activated at compiler time:
- GNSS is disabled
- GNSS is enabled without assistance
- GNSS is enabled with assistance when both AGPS and PGPS are activated

applications/serial_lte_modem/doc/GNSS_AT_commands.rst
```diff
@@ -7,54 +7,103 @@ GNSS AT commands
...
+Both the :ref:`lib_nrf_cloud_agps` service and the :ref:`lib_nrf_cloud_pgps` service can be used with the module, either together or independently of each other.
+Using them reduces the time it takes the GNSS module to estimate its position.
+
+To use either of them, the device must be connected to nRF Cloud (using the :ref:`#XNRFCLOUD <SLM_AT_NRFCLOUD>` AT command) when starting the GNSS module.
+In addition, the following Kconfig options must be enabled:
+
+   * :kconfig:option:`CONFIG_NRF_CLOUD_AGPS <CONFIG_NRF_CLOUD_AGPS>` to use A-GPS.
+   * :kconfig:option:`CONFIG_NRF_CLOUD_PGPS <CONFIG_NRF_CLOUD_PGPS>` to use P-GPS.
+
+When using either assistive service, the choice of which are used is made at compile time only.
```
wiba-nordic, line 31: Would this work:
```suggestion
Assistive services cannot be enabled after compile time.
```
Alternatively, if the Kconfig options are what must be enabled at compile time, then it makes sense to move this information to the sentence that introduces them, so using this for Line 26 and removing Line 31:
```
In addition, the following Kconfig options must be enabled at compile time:
```

applications/serial_lte_modem/doc/GNSS_AT_commands.rst
```diff
@@ -7,54 +7,103 @@ GNSS AT commands
...
-It accepts the following integer values:
+In continuous navigation mode, the ``<timeout>`` parameter must not be given.
+In single-fix and periodic navigation modes, the ``<timeout>`` parameter indicates the maximum time in seconds that the GNSS receiver is allowed to run while trying to produce a valid PVT estimate.
+It can have one of the following integer values:
 
 * ``0`` - The GNSS receiver runs indefinitely until a valid PVT estimate is produced.
-* Any positive integer lower than the ``<interval>`` value - The GNSS receiver is turned off after the specified time is up, even if a valid PVT estimate was not produced.
+* Any positive integer - The GNSS receiver is turned off after the specified time is up, even if a valid PVT estimate was not produced.
+
+When not specified in single-fix or periodic navigation modes, it defaults to a timeout value of 60 seconds.
```
wiba-nordic, line 65:
```suggestion
  When not specified in single-fix or periodic navigation modes, it defaults to a timeout value of 60 seconds.
```
Indenting to make it part of the bullet.

wiba-nordic (changes requested):
Fresh look gave a couple of points I'd like changed.

applications/serial_lte_modem/doc/GNSS_AT_commands.rst
wiba-nordic: Hmm, in that case, I might move this to a negative declaration like so:
```suggestion
If both assistive services are enabled at compile time, you cannot choose to use only one of them at run time.
```

applications/serial_lte_modem/doc/GNSS_AT_commands.rst
wiba-nordic: Discussed on Teams, and landed on this suggestion:
```suggestion
* No value - In single-fix or periodic navigation modes, the timeout defaults to 60 seconds.
```

wiba-nordic (approved):

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -230,41 +355,46 @@ static void pgps_event_handler(struct nrf_cloud_pgps_event *event)
...
+				LOG_WRN("Failed to request new P-GPS predictions (%d).", err);
+			}
 		}
 	} break;
 	/* All P-GPS predictions are available. */
 	case PGPS_EVT_READY:
 		LOG_INF("PGPS_EVT_READY");
 		break;
-	default:
+	case PGPS_EVT_REQUEST:
```
MarkusLassila, line 388: No default in switch-case always makes me suspicious that there should be something. Could we keep the default: break?

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -380,79 +511,62 @@ static void fix_rep_wk(struct k_work *work)
 	}
 
 #if defined(CONFIG_SLM_NRF_CLOUD)
+
```
MarkusLassila, line 514: Additional line change.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -488,247 +602,103 @@ int handle_at_gps(enum at_cmd_type cmd_type)
...
-			err = nrf_modem_gnss_fix_interval_set(interval);
-			if (err) {
-				LOG_ERR("Failed to set fix interval, error: %d", err);
-				return err;
-			}
-			if (at_params_unsigned_short_get(&slm_at_param_list, 3, &timeout) == 0) {
+				/* Make sure to configure the timeout in all cases.
+				 * An old value might otherwise remain from a previous GPS run.
+				 */
 				err = nrf_modem_gnss_fix_retry_set(timeout);
```
MarkusLassila, line 677: Above comment seems to bit misleading as this is not set if the timeout is not given. If it possible to have this from previous GPS run, then this might be problematic in case we first run with timeout and after that run continous.

MarkusLassila (commented):
Looks good, some clarification requests.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
MarkusLassila: It tends to increase readability of the code. In this case, I am reading the code and I am thinking, what will happen when a new event is added. Default: break, would indicate that we would ignore it. Logging unknown value would be even better.

That said, I did not know that compiler would generate an error in this case. Moreover, I did not even notice (and it is not obvious from the code) that the event->type is an enum, which would generate the said error.

So readability would have been the point. In this case though, there is also benefit for omitting the default as well, as that forces addition of new values.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
MarkusLassila: I don't think that this is official notation for our code. I am not against having this or others like this, but I do not want to force people to follow this approach.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
MarkusLassila: Can be resolved.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
MarkusLassila: Can be resolved.

MarkusLassila (approved):

wiba-nordic:
My local documentation build of this branch worked without issue, so trying to rerun it here.

SeppoTakalo:
Please rebase.
There seem to be one conflict, and also documentation build was apparently broken yesterday, so it might be fixed once you rebase.

SeppoTakalo (approved):

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -92,63 +116,180 @@ static bool is_gnss_activated(void)
...
-	} else if (run_type == RUN_TYPE_PGPS) {
-		rsp_send("\r\n#XPGPS: 1,%d\r\n", status);
+	/* A notification is sent when the GNSS status is updated.
+	 * However this gets called in interrupt context (from gnss_event_handler()),
+	 * from which UART messages cannot be sent. This task is thus delegated
+	 * to a worker, to which the the new GNSS status is passed.
+	 * A FIFO rather than a single variable is used to not miss GNSS status changes
+	 * in case they happen twice (or more) before the worker had the time to process them.
+	 */
+	const int put_ret = k_fifo_alloc_put(&gnss_status_fifo, (void *)status);
```
MarkusLassila, line 134: This allocates dynamic memory. Using stack would avoid dynamic memory usage.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -92,63 +116,180 @@ static bool is_gnss_activated(void)
...
+
+	if (put_ret == 0) {
+		k_work_submit_to_queue(&slm_work_q, &gnss_status_notifier);
```
MarkusLassila, line 137: This does not actually launch another worker if the item is already being processed, ran into this with UART implementation. Not likely to happen, but if we want to get all the notifications, you should unload them all in gnss_status_notifier_wk.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
MarkusLassila: Fair enough. I think that there still might be way to avoid dynamic memory usage, but I guess that it does not matter too much in this case.

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
MarkusLassila: If I recall correctly, this failed in scenario where there was something running, a second workqueue submit was scheduled and the third one failed. I did not observe any failures when the queue was emptied in while loop while processing UART, but that does not mean that there would not be scenarios that are problematic.

Anyway, I do not think that these notifications come often enough that they would really be lost with your approach either.

MarkusLassila (approved):

VTPeltoketo (approved):

wiba-nordic (changes requested):
Just a little fix needed. Not something that was introduced in this PR, but might as well fix it when we're touching related lines.

wiba-nordic (approved):

PR #12364, opened 2023-09-19
lib: location: Handle NCELLMEAS results after cellular timeout

trantanen:
Location library used to throw NCELLMEAS results into trash if a timeout occured. When timeout occurs, we send NCELLMEASSTOP. After this modem sends NCELLMEAS notification with already found cells.
We are changing the semantics of cellular timeout to only guard NCELLMEAS operations. Connection and data transfer with the
cloud to resolve the location is not taken into account for cellular timeout. Overall location_request() timeout can still interrupt cloud data transfer.

Fixing also a bug in lte_lc when NCELLMEASSTOP is called and already found cells are used.

Asset Tracker's default cellular timeout modified from 11s to 30s because GCI search support can take advantage of longer than 11s timeout.

Jira: NCSDK-21489

Signed-off-by: Tommi Rantanen <tommi.rantanen@nordicsemi.no>


lib/location/method_cloud_location.c
```diff
@@ -44,34 +44,20 @@ static void method_cloud_location_positioning_work_fn(struct k_work *work)
...
-	if (wifi_config != NULL && cell_config != NULL) {
-		used_timeout_ms = MIN(cell_config->timeout, wifi_config->timeout);
-	} else if (cell_config != NULL) {
-		used_timeout_ms = cell_config->timeout;
-	} else {
-		__ASSERT_NO_MSG(wifi_config != NULL);
-		used_timeout_ms = wifi_config->timeout;
-	}
 
-	location_core_timer_start(used_timeout_ms);
```
trantanen, line 64: Removing this means that now when `scan_cellular` or `scan_wifi` times out, they won't be indicated as timeout event to the application because we try to check if scan returned something and if it didn't, the error code is `-ENODATA`. This is not optimal but passing time out info from `scan_*` modules would be a bit clumsy.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -677,7 +677,13 @@ Modem libraries
 * :ref:`lib_location` library:
 
   * Added support for accessing nRF Cloud services using CoAP through the :ref:`lib_nrf_cloud_coap` library.
-  * Updated the neighbor cell search to use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
+
+  * Updated:
+
+    * The neighbor cell search use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
```
peknis, line 683:
```suggestion
    * The neighbor cell search to use GCI search depending on the :c:member:`location_cellular_config.cell_count` value.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -677,7 +677,13 @@ Modem libraries
...
+    * The semantics of cellular and Wi-Fi timeouts only apply to neighbor cell measurement and Wi-Fi scan, respectively.
+      Earlier these timeouts applied to the upcoming cloud connection to send the data to the cloud for position resolution.
```
peknis, line 685:
```suggestion
      Earlier, these timeouts applied to the cloud connection to send the data to the cloud for position resolution.
```
Not sure I understood this correctly.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -677,7 +677,13 @@ Modem libraries
 * :ref:`lib_location` library:
 
   * Added support for accessing nRF Cloud services using CoAP through the :ref:`lib_nrf_cloud_coap` library.
-  * Updated the neighbor cell search to use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
+
+  * Updated:
+
+    * The neighbor cell search use GCI search depending on :c:member:`location_cellular_config.cell_count` value.
+    * The semantics of cellular and Wi-Fi timeouts only apply to neighbor cell measurement and Wi-Fi scan, respectively.
```
peknis, line 684:
```suggestion
    * The semantics of cellular and Wi-Fi timeouts to only apply to neighbor cell measurement and Wi-Fi scan, respectively.
```

peknis (changes requested):

lib/location/method_cloud_location.c
```diff
@@ -149,21 +135,18 @@ static void method_cloud_location_positioning_work_fn(struct k_work *work)
 		location_result.latitude = location.latitude;
 		location_result.longitude = location.longitude;
 		location_result.accuracy = location.accuracy;
-		if (running) {
-			running = false;
-			location_core_event_cb(&location_result);
-		}
+		location_core_event_cb(&location_result);
```
tokangas, line 138: Just a thought and it's up to you if you do it or not, but this could be moved to the end of the function, where `location_core_event_cb_timeout()` and `location_core_event_cb_error()` are called. Then all event callback calls would be in one place, like clearing of the `running` flag is now.

lib/location/scan_cellular.c
```diff
@@ -24,11 +24,28 @@ static struct lte_lc_cells_info scan_cellular_info = {
...
+/** Handler for timeout. */
+static void scan_cellular_timeout_work_fn(struct k_work *work);
+
+/** Work item for timeout handler. */
+K_WORK_DELAYABLE_DEFINE(scan_cellular_timeout_work, scan_cellular_timeout_work_fn);
+
+/**
+ * Handler for backup timeout, which ensures we won't be waiting for LTE_LC_EVT_NEIGHBOR_CELL_MEAS
+ * event forever after lte_lc_neighbor_cell_measurement_cancel() in case it is never sent by
+ * lower layers due to a bug.
```
tokangas, line 42: I would rephrase the ending to something like "in case it would never be sent". The current comment could be understood so, that there is a bug, because of which the event is sometimes not sent. Which I don't think is the case?

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
trantanen: We need some wording to tell that the timeouts used to apply to both neighbor cell search and the cloud connection or data transfer that was used to send the neighbor data to the cloud.

By removing "upcoming", this is either not the case or at least very unclear.

peknis: Roger that. I had a bit of hard time digesting this.

lib/location/scan_cellular.c
```diff
@@ -24,11 +24,28 @@ static struct lte_lc_cells_info scan_cellular_info = {
 };
 
 static bool running;
+static bool timeout_occurred;
```
tokangas, line 27: Not sure if it makes any difference, but this could be `volatile`. It's being set and checked from different threads.

tokangas (approved):

lib/location/method_cloud_location.c
trantanen: `location_core_event_cb()` shouldn't be sent when running is checked on line 72.
Also, it's kind of nice to call this right after the `location_result` is set on lines 135-137.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
trantanen: Comma added and "also" too. Please check.

lib/location/method_cloud_location.c
tokangas: > `location_core_event_cb()` shouldn't be sent when running is checked on line 72.

Ok, I didn't spot that because it wasn't visible in the diff by default.

peknis (approved):

jhirsi (approved):

maxd-nordic (approved):

PR #12372, opened 2023-09-20
applications: matter: Fix the wrongly handled status on Matter Bridge

ArekBalysNordic:
Currently, the Matter Bridge cannot read the current LED state from the LBS characteristic. After losing the connection between the Bridge and the BLE LBS device we lost information about the current LED state.

To avoid losing the LED state synchronization, when the Matter bridge retrieves the connection to a Bluetooth LE LBS device we should set the current value of LED on the BLE device. We should focus on two scenarios:

1. If The Bridge has not been rebooted and the connection is retrieved, then the state of the LED that is kept in RAM should be set on the LBS device.
2. The Bridge has been rebooted and it restored all connections to BLE devices, and then we should force the initial LED state on the BLE device.

Another change is regarding ESP subscriptions. We cannot subscribe to the GATT in the Init method because we do not have
an active connection. We should subscribe after checking all GATT descriptors and unsubscribe when the lost connection is detected and stop the timer.

### DIFF
--- a/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
+++ b/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
@@ -158,12 +158,16 @@ void BLEConnectivityManager::DiscoveryCompletedHandler(bt_gatt_dm *dm, void *con
 	discoveryResult = true;
 exit:
 	if (provider) {
+		VerifyOrReturn(provider->ParseDiscoveredData(dm) == 0,
+			       LOG_ERR("Cannot parse the GATT discovered data."));
+
 		if (!provider->IsInitiallyConnected()) {
 			/* Provider is not initalized it so we need to call the first connection callback */
 			provider->GetBLEBridgedDevice().mFirstConnectionCallback(
-				dm, discoveryResult, provider->GetBLEBridgedDevice().mFirstConnectionCallbackContext);
+				discoveryResult, provider->GetBLEBridgedDevice().mFirstConnectionCallbackContext);
 			provider->ConfirmInitialConnection();
 		}
+
 		VerifyOrReturn(CHIP_NO_ERROR == provider->NotifyReachableStatusChange(true),
 			       LOG_WRN("The device has not been notified about the status change."));
 	}
@@ -179,7 +183,7 @@ void BLEConnectivityManager::DiscoveryNotFound(bt_conn *conn, void *context)
 	if (provider) {
 		if (!provider->IsInitiallyConnected()) {
 			provider->GetBLEBridgedDevice().mFirstConnectionCallback(
-				nullptr, false, provider->GetBLEBridgedDevice().mFirstConnectionCallbackContext);
+				false, provider->GetBLEBridgedDevice().mFirstConnectionCallbackContext);
 		} else {
 			Instance().mRecovery.NotifyProviderToRecover(provider);
 		}
@@ -194,7 +198,7 @@ void BLEConnectivityManager::DiscoveryError(bt_conn *conn, int err, void *contex
 
 	if (!provider->IsInitiallyConnected()) {
 		provider->GetBLEBridgedDevice().mFirstConnectionCallback(
-			nullptr, false, provider->GetBLEBridgedDevice().mFirstConnectionCallbackContext);
+			false, provider->GetBLEBridgedDevice().mFirstConnectionCallbackContext);
 	}
 }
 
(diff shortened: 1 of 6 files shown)


ArekBalysNordic:
> Should also cover the update of the led state if the state has been changed not by the bridge e.g. the LBS will change the light state manually?

Unfortunately not, because LBS sample does not support reading from the Bluetooth LE characteristic and we don't know about the state change. In other BT light bulb implementations it probably will be possible, but with this sample it isn't. So, we can control the state of LBS sample directly from the Bridge only and to keep synchronization of LED state we need to force the state after each connection to the LBS decvice.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -163,6 +163,11 @@ void BLEConnectivityManager::DiscoveryCompletedHandler(bt_gatt_dm *dm, void *con
 			provider->GetBLEBridgedDevice().mFirstConnectionCallback(
 				dm, discoveryResult, provider->GetBLEBridgedDevice().mFirstConnectionCallbackContext);
 			provider->ConfirmInitialConnection();
+		} else {
+			/* Provider has been already initialized and we need to parse the received data to attach the
+			 * gatt handle after Bridge reboot */
+			VerifyOrReturn(provider->ParseDiscoveredData(dm) == 0,
```
jens6151, line 169: Besides loosing the GATT data in case of a reconnect, the subscriptions also get lost. 
Subscribe is only called in the providers Init function (not yet for onoff, but e.g. for environmental data provider).
So would this be the correct place to also call provider->Init()?

ArekBalysNordic: Actually, `provider->Init()` will be called in the other place, to be more precise, here: https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/common/src/bridge/bridge_manager.cpp#L287 so calling it in `DiscoveryCompletedHandler` is redundant. A different situation is with `ParseDiscoveredData` because this method is called within the `mFirstConnectionCallback` callback, but when the device is restoring from the memory (after the Bridge reboot) it is not called because we have a provider assigned as `InitiallyConnected`.

kkasperczyk-no:
> Should also cover the update of the led state if the state has been changed not by the bridge e.g. the LBS will change the light state manually?

Beside of what Arek wrote, the LBS is not able to change the state manually. This sample supports only changing the state using BLE GATT.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no, line 169: I'm wondering if calling `ParseDiscoveredData` in two places (here and inside the `mFirstConnectionCallback`) is a good idea. Maybe we should remove calling it from the callback and call it here, unconditionally, before the `if (!provider->IsInitiallyConnected())` check.
It would make implementation cleaner and we would not push the responsibility for calling it to the user's application.

ArekBalysNordic: Ok, I'm ok with it. Actually, I was wondering about that, and your comment assured me that do it in that way. :)

ArekBalysNordic: Moreover, my comment there is redundant when we do it as you wrote because it seems natural that in DiscoevryCompleteHandeler we parse the discovered data.

jens6151: You are right. I can see that there is a call stack from AppTask::RestoreBridgedDevices() to the given location. In that case, Subscribe() will be called before getting the log "app: Connected: 62:19:96:A4:E0:DF (random)" resulting in "Invalid connection object". (means before being connected)
The second call stack to reach the given location is via BluetoothDeviceConnected(...) which is only reached from the shell command add.

You mentioned that ParseDiscoveredData will not be called when restoring. Actually ParseDiscoveredData is required to fill in the mGattPostureSubscribeParams. Without these, Subscription will also fail and the polling fallback is also missing the correct mPostureCharacteristicHandle.

Actually the the issue with the Subscribe is not really related to this merge request. Just thought this is an opportunity for feedback as there is no "Issues" tab on this repository.

ArekBalysNordic: Hmm, ok you're right! :) We've not noticed that and focused on the LBS sample, but yes, a subscription should be called after the connection establishment, not in the init() method. Thank you for the input.

kkasperczyk-no (approved):

doublemis1 (approved):
LGTM

PR #12343, opened 2023-09-17
doc improvement: ESB sample page

kittydepa:
This is a recreation of previous PR #10739

__________________________________________________

[NCSDK-13353](https://nordicsemi.atlassian.net/browse/NCSDK-13353)

Split ESB samples into two and reorder based on the samples template.
Moved ESB samples one heading higher in the "Samples" section.

Minor change to ESB Protocols page to one heading for consistency with other Protocol docs.

[NCSDK-13353]: https://nordicsemi.atlassian.net/browse/NCSDK-13353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #12378, opened 2023-09-20
samples: esb: fix esb_ptx path

magnev:
Renames esb/esp_ptx to the correct name esb/esb_ptx


lemrey (approved):
Looks good, thanks. If it had the correct path in NCS `v2.4.0`, we do not need a changelog entry, so you can go ahead and remove the label if that's the case.

magnev:
The path to the samples were changed in this PR: https://github.com/nrfconnect/sdk-nrf/pull/12343

stephen-nordic (approved):

PR #12380, opened 2023-09-20
samples: matter: Align cluster changes in all Matter samples

ArekBalysNordic:
According to Matter Specs, for Matter 1.1.0.1 we need to set the Software Diagnostic Feature Map value to "0" and set the Basic Information Cluster Revision value to "1" in all our samples.


kkasperczyk-no:
Shouldn't we align weather station and bridge as well?

applications/matter_bridge/src/bridge.zap
```diff
@@ -3207,10 +3207,10 @@
               "side": "server",
               "type": "bitmap32",
               "included": 1,
-              "storageOption": "RAM",
+              "storageOption": "External",
```
kkasperczyk-no, line 3210: Why did you change it from RAM to External? I'm not saying it's wrong, I just see FeatureMap is stored in RAM for all clusters, so I wonder why you decided to change it only for this one.

kkasperczyk-no (approved):

applications/matter_bridge/src/bridge.zap
ArekBalysNordic: Recently, we have done it in this way for DoorLock sample, and AFAIR it worked properly so I aligned all samples to that change. @doublemis1 What do you think?

doublemis1: I am not aware about these solution, what suits best.
However, I agree with Arek that it should be aligned across all samples/clusters/endpoints

doublemis1 (approved):

PR #12397, opened 2023-09-21
manifest: net: Add copying of missing flags inside clone_pkt_attributes()

MarGasiorek:
PR includes changes in source code and tests update.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -61,7 +61,7 @@ manifest:
     # https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/modules.html
     - name: zephyr
       repo-path: sdk-zephyr
-      revision: 0c7fcf3c32a712f5003def83a3607404850dc9e2
+      revision: pull/1312/head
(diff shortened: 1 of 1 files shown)


rlubos:
Twister fails due to FLASH overflow in LwM2M interop test app. It seems that we either need to cherry-pick https://github.com/zephyrproject-rtos/zephyr/commit/0a104185fef22caf1019a78b2d2b2f74301771ad as well (not sure if it'd apply cleanly), or just wait for the upmerge to complete and drop this one.

MarGasiorek:
> Twister fails due to FLASH overflow in LwM2M interop test app. It seems that we either need to cherry-pick [zephyrproject-rtos/zephyr@0a10418](https://github.com/zephyrproject-rtos/zephyr/commit/0a104185fef22caf1019a78b2d2b2f74301771ad) as well (not sure if it'd apply cleanly), or just wait for the upmerge to complete and drop this one.

When we can expect upmerge completed?

PR #12401, opened 2023-09-21
tfm: huk: Using full buffer to derive key

mswarowsky:
Also the first bytes nrf_cc3xx_platform_kmu_shadow_key_derive are different if output_size changes.
Therefore even on CC310 devices we have to use the previously used 32 bytes and just use the beginning of the buffer.

This ensures that still the same key gets derived after switching to the new buildin key framework.


SebastianBoe (approved):
This was a regression from 7e00649bb6a2c72044813276ee3389886e50ec38. hw_unique_key_derive_key should be given the size of buf.

de-nordic (approved):

PR #12402, opened 2023-09-21
applications: nrf5340_audio: Add user-specific BT name

MariusVaardal:
Add user-specific BT broadcast name option in buildprog.py.

### DIFF
--- a/applications/nrf5340_audio/tools/buildprog/buildprog.py
+++ b/applications/nrf5340_audio/tools/buildprog/buildprog.py
@@ -16,6 +16,7 @@ import json
 import subprocess
 import re
 import glob
+import getpass
 from colorama import Fore, Style
 from prettytable import PrettyTable
 from nrf5340_audio_dk_devices import (
@@ -44,6 +45,8 @@ TARGET_DEV_GATEWAY_FOLDER = NRF5340_AUDIO_FOLDER / "build/dev_gateway"
 
 TARGET_RELEASE_FOLDER = "build_release"
 TARGET_DEBUG_FOLDER = "build_debug"
+
+MAX_USER_NAME_LEN = 248 - len('\0')
 
 
 def __print_add_color(status):
@@ -108,6 +111,20 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
 
         if options.nrf21540:
             device_flag += " -DSHIELD=nrf21540ek_fwd"
+
+        if options.custom_bt_name is not None and options.user_bt_name:
+            raise Exception(
+                "User BT name option is invalid when custom BT name is set")
+
+        if options.custom_bt_name is not None:
+            custom_bt_name = "_".join(options.custom_bt_name)[
+                :MAX_USER_NAME_LEN].upper()
+            device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\"" + custom_bt_name + "\\\""
+
+        if options.user_bt_name:
+            user_specific_bt_name = (
+                "AUDIO_DEV_" + getpass.getuser())[:MAX_USER_NAME_LEN].upper()
+            device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\"" + user_specific_bt_name + "\\\""
 
         if os.name == 'nt':
             release_flag = release_flag.replace('\\', '/')
@@ -292,14 +309,30 @@ def __main():
         choices=["external", "internal"],
         default='',
         help="MCUBOOT with external, internal flash",
-        )
+    )
     parser.add_argument(
         "--nrf21540",
         action="store_true",
         dest="nrf21540",
         default=False,
         help="Set when using nRF21540 for extra TX power",
-        )
+    )
+    parser.add_argument(
+        "-cn",
+        "--custom_bt_name",
+        nargs='*',
+        dest="custom_bt_name",
+        default=None,
+        help="Use custom Bluetooth device name.",
+    )
+    parser.add_argument(
+        "-u",
+        "--user_bt_name",
+        action="store_true",
+        dest="user_bt_name",
+        default=False,
+        help="Set to generate a user specific Bluetooth device name. Note that this will put the computer user name on air in clear text.",
+    )
     options = parser.parse_args(args=sys.argv[1:])
 
     # Post processing for Enums
@@ -352,7 +385,8 @@ def __main():
     # Reboot step start
 
     if options.only_reboot == SelectFlags.TBD:
-        program_threads_run(device_list, options.mcuboot, sequential=options.sequential_prog)
+        program_threads_run(device_list, options.mcuboot,
+                            sequential=options.sequential_prog)
         __finish(device_list)
 
     # Reboot step finished
@@ -393,7 +427,8 @@ def __main():
         for dev in device_list:
             if dev.snr_connected:
                 __populate_hex_paths(dev, options)
-        program_threads_run(device_list, options.mcuboot, sequential=options.sequential_prog)
+        program_threads_run(device_list, options.mcuboot,
+                            sequential=options.sequential_prog)
 
     # Program step finished
 


applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -109,6 +110,10 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
         if options.nrf21540:
             device_flag += " -DSHIELD=nrf21540ek_fwd"
 
+        if options.user_specific_name:
+            user_specific_bt_name = "NRF5340_AUDIO_" + getpass.getuser().upper()
```
koffes, line 114: We need to truncate if the user name is too long.

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -109,6 +110,10 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
...
+            device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\""+ user_specific_bt_name + "\\\""
```
koffes, line 115: Should we add the broadcast name here as well? @alexsven @erikrobstad ?

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -300,6 +305,14 @@ def __main():
         default=False,
         help="Set when using nRF21540 for extra TX power",
         )
+    parser.add_argument(
+        "-u"
+        "--user_bt_name",
+        action="store_true",
+        dest="user_specific_name",
+        default=False,
+        help="Set to true for user specific BT broadcast name",
```
koffes, line 314:
```suggestion
        help="Set to true for user specific BT broadcast name. Note that this will put the user name on air in clear text",
```

koffes (changes requested):

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -300,6 +305,14 @@ def __main():
         default=False,
         help="Set when using nRF21540 for extra TX power",
         )
+    parser.add_argument(
+        "-u"
+        "--user_bt_name",
+        action="store_true",
+        dest="user_specific_name",
```
andvib, line 312:
```suggestion
        dest="user_bt_name",
```
Unless there's a specific reason I think it's better to have the flag and the variable have the same naming

applications/nrf5340_audio/tools/buildprog/buildprog.py
andvib, line 314:
```suggestion
        help="Set to generate a user specific BT broadcast name",
```

andvib (changes requested):
Just two minor changes

applications/nrf5340_audio/tools/buildprog/buildprog.py
alexsven, line 314: This is not the broadcast name though, this is the advertised name, or 'friendly name'. Since it is hard coded to be NRF5340_AUDIO_[username or something similar], that should be stated. There could be a situation where another company takes our sample and creates a prj.conf specific for them setting CONFIG_BT_DEVICE_NAME=[project_product]. Then they figure out they want their developers to have different BT_DEVICE_NAMEs and uses this, suddenly all their kits will be 'NRF5340_AUDIO_....'

applications/nrf5340_audio/tools/buildprog/buildprog.py
alexsven, line 114: As stated in my other comment this makes sense for us here, developing inhouse, but it might not make sense for anyone else, as this will overwrite their specific BT_DEVICE_NAME with NRF5340_AUDIO_....

erikrobstad (changes requested):

applications/nrf5340_audio/tools/buildprog/buildprog.py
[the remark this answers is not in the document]
MariusVaardal: For some reason. I had to do this for it to work. I need the command to look like this:  -DCONFIG_BT_DEVICE_NAME=\"NRF5340_AUDIO_USER_NAME\"

applications/nrf5340_audio/tools/buildprog/buildprog.py
MariusVaardal: What is too long? The compiler compilains at total length of 248. I use a max length of 50 for now

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -109,6 +110,16 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
         if options.nrf21540:
             device_flag += " -DSHIELD=nrf21540ek_fwd"
 
+        if options.user_bt_name:
+            MAX_USER_NAME_LEN = 248
+            user_specific_bt_name = ("AUDIO_DEV_" + getpass.getuser())[:MAX_USER_NAME_LEN].upper()
+            device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\"" + user_specific_bt_name + "\\\""
+
+        if options.custom_bt_name != None:
+            MAX_USER_NAME_LEN = 248
```
MariusVaardal, line 119: I know I have defined this variable twice. I thought this was better than to have only one declaration that is far from where it is used. But I can change this if people want

applications/nrf5340_audio/tools/buildprog/buildprog.py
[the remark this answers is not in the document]
erikrobstad: I see now that this comment is invalid, since it is not the broadcast name that is set here.

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -109,6 +110,16 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
         if options.nrf21540:
             device_flag += " -DSHIELD=nrf21540ek_fwd"
 
+        if options.user_bt_name:
+            MAX_USER_NAME_LEN = 248
+            user_specific_bt_name = ("AUDIO_DEV_" + getpass.getuser())[:MAX_USER_NAME_LEN].upper()
+            device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\"" + user_specific_bt_name + "\\\""
+
+        if options.custom_bt_name is not None:
+            MAX_USER_NAME_LEN = 248
```
erikrobstad, line 119: I suggest putting an "if !options.user_bt_name:" check here and only use custom_bt_name if user_bt_name is not set

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -300,6 +311,23 @@ def __main():
...
+    parser.add_argument(
+        "-u",
+        "--user_bt_name",
+        action="store_true",
+        dest="user_bt_name",
+        default=False,
+        help="Set to generate a user specific Bluetooth device name. Note that this will put the user name on air in clear text.",
+        )
+    parser.add_argument(
+        # should only be used when --user_bt_name is not used
```
erikrobstad, line 323: Remove this comment

erikrobstad (changes requested):

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -16,6 +16,7 @@
 import subprocess
 import re
 import glob
+import getpass
```
koffes, line 19: This is a new python tool, so this needs to be added to the requirements list.
It should probably go into /nrf/scripts/requirements-fixed.txt but please check with Thomas Stilwell or Sebastian Wezel.

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -109,6 +110,16 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
         if options.nrf21540:
             device_flag += " -DSHIELD=nrf21540ek_fwd"
 
+        if options.user_bt_name:
+            MAX_USER_NAME_LEN = 248
+            user_specific_bt_name = ("AUDIO_DEV_" + getpass.getuser())[:MAX_USER_NAME_LEN].upper()
```
koffes, line 115: Pitfall: If the max name len is 248, then you need to deduct "AUDIO_DEV_" (10) from `:MAX_USER_NAME_LEN`

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -109,6 +110,16 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
         if options.nrf21540:
             device_flag += " -DSHIELD=nrf21540ek_fwd"
 
+        if options.user_bt_name:
```
koffes, line 113: There needs to be a guard here. What happens if a user supplies both --user_bt_name" and "--custom_bt_name"?
The simplest way to do this is to generate precedence by checking `if options.custom_bt_name is not None:` first and then do an `else if` for `options.user_bt_name:`

koffes (changes requested):

applications/nrf5340_audio/tools/buildprog/buildprog.py
andvib: I think it's better to define this once at the top of the file. Variables in capital letters are expected to be global constants, so people will know where to look for it.

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -300,6 +311,23 @@ def __main():
...
+        action="store_true",
+        dest="user_bt_name",
+        default=False,
+        help="Set to generate a user specific Bluetooth device name. Note that this will put the user name on air in clear text.",
+        )
+    parser.add_argument(
+        # should only be used when --user_bt_name is not used
+        "-cn",
+        "--custom_bt_name",
+        nargs='*',
```
andvib, line 326: Why do we need to allow for multiple arguments? Shouldn't this argument be given only once?

applications/nrf5340_audio/tools/buildprog/buildprog.py
andvib: I think we should just raise an error and exit if both arguments are given, as giving both is for sure a usage error.

andvib (changes requested):

applications/nrf5340_audio/tools/buildprog/buildprog.py
MariusVaardal: It is not necessary. But now you can write your name with spaces between first and last name and it will set the BT dev name to "MARIUS_VÅRDAL", for example

applications/nrf5340_audio/tools/buildprog/buildprog.py
MariusVaardal: I dont understand this. In my head, it is the user_specific_bt_name that has to have a max length of 248. You might not have noticed that the [:MAX_USER_NAME_LEN] is applied to ("AUDIO_DEV_" + getpass.getuser()). Otherwise, I dont understand what you mean.

applications/nrf5340_audio/tools/buildprog/buildprog.py
MariusVaardal: I will send a message:)

MariusVaardal: I just checked, and it it a part of Python Standard Library, so I guess I dont need to (?)

MariusVaardal: It is a part of Python Standard Library, so I dont need to

andvib (approved):

applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -109,6 +112,17 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
...
+            raise Exception("User BT name option is invalid when custom BT name is set")
+
+        if options.custom_bt_name is not None:
+            custom_bt_name = "_".join(options.custom_bt_name)[:MAX_USER_NAME_LEN].upper()
+            device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\"" + custom_bt_name + "\\\""
+
+        if options.user_bt_name:
+            user_specific_bt_name = ("AUDIO_DEV_" + getpass.getuser())[:MAX_USER_NAME_LEN].upper()
+            device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\"" + user_specific_bt_name + "\\\""
+
```
alexsven, line 125:
```suggestion
        if options.custom_bt_name is not None 
            if options.user_bt_name:
                raise Exception("User BT name option is invalid when custom BT name is set")
            
            custom_bt_name = "_".join(options.custom_bt_name)[:MAX_USER_NAME_LEN].upper()
            device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\"" + custom_bt_name + "\\\""
        else:
            if options.user_bt_name:
                user_specific_bt_name = ("AUDIO_DEV_" + getpass.getuser())[:MAX_USER_NAME_LEN].upper()
                device_flag += " -DCONFIG_BT_DEVICE_NAME=\\\"" + user_specific_bt_name + "\\\""

```

alexsven: Just to make it clearer that custom_name and user_name can't be set at the same time

alexsven (approved):

applications/nrf5340_audio/tools/buildprog/buildprog.py
MariusVaardal: Now you are making me choose between you and Andreas. I think it should be pretty obvious that you can't both set your own name and have the program generate one for you. I will just leave it as is because it is more aesthetic i.m.o.

erikrobstad (approved):

koffes (approved):

PR #12409, opened 2023-09-22
pcd: Add read network core application version command

sigvartmh:
Added PCD command for reading network core application core version. This command reads the version number stored in the firmware info metadata of the network core application image stored on the network core.

Making a draft PR to gather feedback on how this is implemented. I could only find a nice way to implement this with redesigning and re-implementing the whole PCD protocol, which is a bigger task I have in the backlog.

The reason for adding this command is to enable downgrade prevention of the network core during multi-image updates in MCUBoot.

Edit: 

Ref. NCSDK-21379

(!) Note: If this one comes in, this also needs to be taken: https://github.com/nrfconnect/sdk-nrf/pull/12660


include/dfu/pcd.h
```diff
@@ -54,9 +54,20 @@ extern "C" {
 enum pcd_status {
 	PCD_STATUS_COPY = 0,
 	PCD_STATUS_COPY_DONE = 1,
-	PCD_STATUS_COPY_FAILED = 2
+	PCD_STATUS_COPY_FAILED = 2,
+	PCD_STATUS_READ_VERSION = 3,
+	PCD_STATUS_READ_VERSION_DONE = 1,
+	PCD_STATUS_READ_VERSION_FAILED = 2,
```
sigvartmh, line 60: I have a wage memory that I've read somewhere that this is a bad idea but I can't remember where. I do find it more readable to have duplicate enum values with regards to the code.

nordicjm:
> The reason for adding this command is to enable downgrade prevention of the network core during multi-image updates in MCUBoot.

This is already the case as far as I understand it, the image will be transferred to the network core then binned, so whilst MCUboot will see the transfer as successful, it will not actually change the image, best way to test this is take a thingy53, load a 2.4 bluetooth sample, then DFU a 2.3 zigbee application, and it should either fail to boot or fail to use network core functions

sigvartmh:
> > The reason for adding this command is to enable downgrade prevention of the network core during multi-image updates in MCUBoot.
> 
> This is already the case as far as I understand it, the image will be transferred to the network core then binned, so whilst MCUboot will see the transfer as successful, it will not actually change the image, best way to test this is take a thingy53, load a 2.4 bluetooth sample, then DFU a 2.3 zigbee application, and it should either fail to boot or fail to use network core functions

So when a transfer is done, this check is applied:
https://github.com/nrfconnect/sdk-nrf/blob/b1eba9a982ce3ca708e6cad9d6b21f9f7171a286/include/bl_validation.h#L34
See-> https://github.com/nrfconnect/sdk-nrf/blob/b1eba9a982ce3ca708e6cad9d6b21f9f7171a286/samples/nrf5340/netboot/src/main.c#L49C11-L49C31

So, you are correct in saying that there is some inherent rollback protection in b0n.  I'll have to check if we use monotonic in b0n, but I'm pretty sure we are.

Jira: NCSDK-21379

subsys/pcd/src/pcd.c
```diff
@@ -183,10 +208,11 @@ static int network_core_pcd_cmdset(const void *src_addr, size_t len, bool wait)
 		k_busy_wait(1 * USEC_PER_SEC);
 
 		err = pcd_fw_copy_status_get();
-	} while (err == PCD_STATUS_COPY);
+	} while (err == command_status);
 
-	if (err == PCD_STATUS_COPY_FAILED) {
+	if (err == PCD_STATUS_FAILED) {
 		LOG_ERR("Network core update failed");
+		network_core_pcd_tidy();
```
sigvartmh, line 215: I'm a little bit uncertain if this will break things. It resets the network core back to non-secure if either the network core update fails or the version number reading failed.

subsys/pcd/src/pcd.c
```diff
@@ -183,10 +208,11 @@ static int network_core_pcd_cmdset(const void *src_addr, size_t len, bool wait)
 		k_busy_wait(1 * USEC_PER_SEC);
 
 		err = pcd_fw_copy_status_get();
-	} while (err == PCD_STATUS_COPY);
+	} while (err == command_status);
```
hakonfam, line 211: while we are here, could we rename `err` to something more precise? `err` makes the code harder to read.

subsys/pcd/src/pcd.c
hakonfam: But why do we need to add it now?

modules/mcuboot/hooks/nrf53_hooks.c
```diff
@@ -19,9 +19,51 @@
...
+ * @retval -1           If ver1 is less than ver2.
+ * @retval 0            If the image version numbers are equal.
+ * @retval 1            If ver1 is greater than ver2.
+ */
+int boot_version_cmp_net(const struct flash_area *fap, struct image_header *hdr)
+{
+	int err;
+	const struct fw_info *firmware_info;
+	uint32_t version = 0;
+	uint32_t read_flash_page[4096];
```
hakonfam, line 35: Should this buffer size be configurable? Or at least come from a read-only kconfig option?

modules/mcuboot/hooks/nrf53_hooks.c
```diff
@@ -19,9 +19,51 @@
 #define NET_CORE_VIRTUAL_PRIMARY_SLOT 3
 
 #include <dfu/pcd.h>
+#include <fw_info_bare.h>
 
-int boot_read_image_header_hook(int img_index, int slot,
-		struct image_header *img_head)
+/**
+ * Compare image version numbers
```
hakonfam, line 25: Should the docs be in some header file and in doxygen format?

modules/mcuboot/hooks/nrf53_hooks.c
```diff
@@ -19,9 +19,51 @@
...
+ * @retval -1           If ver1 is less than ver2.
+ * @retval 0            If the image version numbers are equal.
+ * @retval 1            If ver1 is greater than ver2.
+ */
+int boot_version_cmp_net(const struct flash_area *fap, struct image_header *hdr)
+{
+	int err;
+	const struct fw_info *firmware_info;
+	uint32_t version = 0;
```
hakonfam, line 34: Don't we need a double word? Ref. the doc of the pcd version read function

modules/mcuboot/hooks/nrf53_hooks.c
```diff
@@ -19,9 +19,51 @@
...
+	uint32_t read_flash_page[4096];
+
+	err = pcd_network_core_app_version(&version);
+	if (err != 0) {
+		BOOT_LOG_ERR("Failure fetching network core application version");
+		return -EFAULT;
+	}
+
+	err = flash_area_read(fap, hdr->ih_load_addr, &read_flash_page, 4096 * sizeof(uint32_t));
```
hakonfam, line 43: we need to use a define for 4096 here, why not ARRAY_SIZE directly?

hakonfam (dismissed):

modules/mcuboot/hooks/nrf53_hooks.c
sigvartmh: DWORD (32 bits/4 bytes)

modules/mcuboot/hooks/nrf53_hooks.c
sigvartmh: Forgot about the array_size macro

samples/nrf5340/netboot/src/main.c
```diff
@@ -37,10 +37,24 @@ int main(void)
 		goto failure;
 	}
 
+
```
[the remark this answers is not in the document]
sigvartmh: nope

include/dfu/pcd.h
```diff
@@ -123,6 +132,14 @@ const void *pcd_cmd_data_ptr_get(void);
  */
 int pcd_fw_copy(const struct device *fdev);
 
+/** @brief Read network core image version.
+ *
+ * Finds and sets the version number in the data buffer provided by PCD CMD structure.
+ *
+ * @retval non-negative integer on success, negative on failure.
+ */
+int pcd_find_fw_version(void);
```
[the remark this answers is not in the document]
sigvartmh: Hmm, find_and_load_fw_version? It does two things. First checks if the firmware info is there then grabs the version then loads it into RAM and signals.

modules/mcuboot/hooks/nrf53_hooks.c
sigvartmh: I wasn't able to reply to the above comment:

> But why do we need to add it now?

Because you could end up reading the version and it failed, try to start the application. This would fail as the network core would be in secure while the application expects non-secure.

> Should this buffer size be configurable? Or at least come from a read-only kconfig option?

I'll look into adding it as a read-only konfig. This buffer just needs to be large enough to find the firmware version in the network core.

modules/mcuboot/hooks/nrf53_hooks.c
sigvartmh: I'm not 100% set on the location of this function. I couldn't find a generic way to include this. Maybe I would start comparing it against he major version of the mcuboot header.

sigvartmh:
@hakonfam Tried to address some of your comments so please re-review

subsys/pcd/src/pcd.c
sigvartmh: In my mind: "Because you could end up reading the version and it failed, try to start the application. This would fail as the network core would be in secure while the application expects non-secure." this was the logic. I'll go through the code once more to see if this is true.

modules/mcuboot/hooks/nrf53_hooks.c
hakonfam: I have never heard that term, and googling it seems to indicate that its a Microsoft/windows specific term, should we instead state that its a word? or a 32bit or 4 byte value?

modules/mcuboot/hooks/nrf53_hooks.c
hakonfam: there is no need to pull it out simply to be able to add doxygen docs to these IMO

modules/mcuboot/hooks/nrf53_hooks.c
sigvartmh: Agreed lets change it's obscure and misleading to call it double word.

modules/mcuboot/hooks/nrf53_hooks.c
sigvartmh: Sure will do.

include/dfu/pcd.h
```diff
@@ -53,10 +53,19 @@ extern "C" {
...
-	PCD_STATUS_COPY_DONE = 1,
-	PCD_STATUS_COPY_FAILED = 2
+	PCD_STATUS_DONE = 1,
+	PCD_STATUS_FAILED = 2,
+	PCD_STATUS_READ_VERSION = 3,
 };
 
+/** @brief Sets up the PCD command structure and point the data buffer to version
+ *
+ * @param version Pointer to a doubleword sized buffer in RAM.
```
nordicjm, line 63: Can we not use things like doubleword and to highlight why to avoid them is word on armv7 is 32 bits, so "doubleword" here would be 64-bits

nordicjm (dismissed):

modules/mcuboot/hooks/nrf53_hooks.c
```diff
@@ -19,9 +19,52 @@
...
+ * @retval 1            If ver1 is greater than ver2.
+ */
+int boot_version_cmp_net(const struct flash_area *fap, struct image_header *hdr)
+{
+	int err;
+	const struct fw_info *firmware_info;
+	uint32_t version = 0;
+	uint32_t read_flash_page[CONFIG_PCD_VERSION_PAGE_BUF_SIZE];
+
+	err = pcd_network_core_app_version((uint8_t *)&version, 4);
```
nordicjm, line 37: `sizeof(version)`

modules/mcuboot/hooks/nrf53_hooks.c
```diff
@@ -19,9 +19,52 @@
...
+	}
+
+	err = flash_area_read(fap, hdr->ih_load_addr, &read_flash_page,
+				ARRAY_SIZE(read_flash_page));
+	if (err != 0) {
+		BOOT_LOG_ERR("Failed reading flash area");
+		return -EFAULT;
+	}
+
+	firmware_info = fw_info_find((uint32_t) &read_flash_page);
```
nordicjm, line 50: no space before &

modules/mcuboot/hooks/nrf53_hooks.c
```diff
@@ -19,9 +19,52 @@
...
 #include <dfu/pcd.h>
+#include <fw_info_bare.h>
 
-int boot_read_image_header_hook(int img_index, int slot,
-		struct image_header *img_head)
+/**
+ * Compare image version numbers
+ * @retval -1           If ver1 is less than ver2.
+ * @retval 0            If the image version numbers are equal.
+ * @retval 1            If ver1 is greater than ver2.
```
nordicjm, line 28: No mention that an errno can also be returned in case of error

samples/nrf5340/netboot/src/main.c
```diff
@@ -39,8 +39,22 @@ int main(void)
...
+		err = pcd_find_fw_version();
+		if (err < 0) {
+			printk("Unable to find valid firmware version %d\n\r", err);
+			pcd_fw_copy_invalidate();
+		}
+
+		/* Success, waiting to be rebooted */
+		while (1)
+			;
+		CODE_UNREACHABLE;
```
nordicjm, line 54: If the version is checked, why would you then want to have to reboot the processor to load a firmware update, surely you'd want to upload it right away after, then reboot?

include/dfu/pcd.h
sigvartmh: Agreed above

modules/mcuboot/hooks/nrf53_hooks.c
sigvartmh: Removed

samples/nrf5340/netboot/src/main.c
sigvartmh: Fair enough. However, the PCD CMD implementation is a bit weird, and I wanted to keep that behaviour the same in this patch. I'm planning and working on reworking the protocol from the ground up. This is a stop-gap to help out projects that needs this functionality before it is ready.

modules/mcuboot/hooks/nrf53_hooks.c
sigvartmh: Fixed.

sigvartmh:
@nordicjm @hakonfam Tried to address your comments please re-review.

sigvartmh:
Same question for this. It would be nice for those waiting to get it in 2.5.0

modules/mcuboot/hooks/nrf53_hooks.c
```diff
@@ -19,9 +19,56 @@
 #define NET_CORE_VIRTUAL_PRIMARY_SLOT 3
 
 #include <dfu/pcd.h>
+#if defined(CONFIG_PCD_APP) && defined(CONFIG_NRF53_MULTI_IMAGE_UPDATE)
+#include <fw_info_bare.h>
 
-int boot_read_image_header_hook(int img_index, int slot,
-		struct image_header *img_head)
+/** @brief Compare image version numbers of network core and network core update.
+ *
```
nordicjm, line 26: I'm not seeing any changes to `boot_serial_uploaded_hook` here so someone can completely bypass this just by doing serial recovery? Or is that expected, as an emergency recovery?

sigvartmh: I would assume that's expected as an emergency recovery but I'll ask the team needing this functionality.

nordicjm: No serial recovery does not work like that, if you have downgrade protection enabled, you can load an older image via serial recovery but it will not boot, it needs to be the same or newer than the existing image

sigvartmh: > No serial recovery does not work like that, if you have downgrade protection enabled, you can load an older image via serial recovery but it will not boot, it needs to be the same or newer than the existing image 

Isn't this statement true only if you have enabled HW downgrade prevention? I can't find any references to downgrade prevention here https://github.com/nrfconnect/sdk-mcuboot/blob/main/boot/boot_serial/src/boot_serial.c 

and MCUBoot only stores the version number in the MCUBoot header AFAIK. So if you use software downgrade prevention Serial Recovery does boot the new image.

nordicjm: If you use software protection yes, it is hardware protection that enforces it, but isn't that kind of the point on the nrf5340, because it has hardware protection? Seems odd to have hardware protection on one core and no protection on the other

sigvartmh: That's how nRF53 hardware is, at least, from my point of view. 

Hardware protection in nRF53 is done using security counters based on OTP. These are in the application core stored in UICR see https://infocenter.nordicsemi.com/topic/ps_nrf5340/chapters/uicr/doc/uicr.html?cp=4_0_0_4_3_2

They are, however, not present in the network core, meaning you cannot have hardware protection in the network core.

https://infocenter.nordicsemi.com/topic/ps_nrf5340/chapters/uicr.network/doc/uicr.network.html?cp=4_0_0_5_3_1

hakonfam (approved):

subsys/pcd/src/pcd.c
```diff
@@ -75,6 +80,24 @@ const void *pcd_cmd_data_ptr_get(void)
...
+	const struct fw_info *firmware_info;
+
+	if (cmd->magic != PCD_CMD_MAGIC_READ_VERSION) {
+		return -EFAULT;
+	}
+
+	firmware_info = fw_info_find(PM_APP_ADDRESS);
+
+	if (firmware_info != NULL) {
+		memcpy((void *)cmd->data, &firmware_info->version, cmd->len);
```
de-nordic, line 94: Are we sure that the size of data to copy is taken from the destination buffer?

sigvartmh: Pretty much yes as the command which sets `PCD_CMD_MAGIC_READ_VERSION` also sets `len` and `data` but if we want to ensure that the API can't be misused I can add another check.

de-nordic (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -1077,6 +1076,16 @@ DFU libraries
 
 * Added a new DFU SMP target for the image update to an external MCU by using the MCUmgr SMP Client.
 
+Peripheral CPU DFU (PCD) library
+--------------------------------
+
+* Added a new function for reading peripherals CPU version number.
```
peknis, line 1082:
```suggestion
* :ref:`subsys_pcd` library:

  * Added a new function for reading the CPU version number of a peripheral.
```
Do you want to mention the exact function?

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -1077,6 +1076,16 @@ DFU libraries
 
 * Added a new DFU SMP target for the image update to an external MCU by using the MCUmgr SMP Client.
```
peknis, line 1077: Which library exactly is this referring to?

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -1077,6 +1076,16 @@ DFU libraries
...
 
+Peripheral CPU DFU (PCD) library
+--------------------------------
+
+* Added a new function for reading peripherals CPU version number.
+
+Netboot
```
peknis, line 1084: Which library do you want to refer to?

include/dfu/pcd.h
```diff
@@ -53,10 +53,19 @@ extern "C" {
...
 enum pcd_status {
 	PCD_STATUS_COPY = 0,
-	PCD_STATUS_COPY_DONE = 1,
-	PCD_STATUS_COPY_FAILED = 2
+	PCD_STATUS_DONE = 1,
+	PCD_STATUS_FAILED = 2,
+	PCD_STATUS_READ_VERSION = 3,
 };
 
+/** @brief Sets up the PCD command structure and point the data buffer to version
```
peknis, line 61:
```suggestion
/** @brief Set up the PCD command structure and point the data buffer to version.
```

peknis (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -453,8 +453,7 @@ Cellular samples (renamed from nRF9160 samples)
   * Updated credentials for the HTTPS connection.
 
 * :ref:`http_modem_delta_update_sample` sample:
-
-  * Updated credentials for the HTTPS connection.
+* Updated credentials for the HTTPS connection.
```
peknis, line 456:
```suggestion
  * Updated credentials for the HTTPS connection.
```

peknis (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
sigvartmh: Sure we could refer to `int pcd_find_fw_version(void);`

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
sigvartmh: I want to refer to https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/nrf5340/netboot/README.html

peknis: Well, in that case, the entry is in wrong place. It should be under Samples -> nRF4340 samples, something like:

* :ref:`nc_bootloader` sample:

  * Write your change here

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
peknis: So, tag it like this: :c:func:`int pcd_find_fw_version(void);`

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -1077,6 +1079,9 @@ DFU libraries
 
 * Added a new DFU SMP target for the image update to an external MCU by using the MCUmgr SMP Client.
```
peknis, line 1080: Is this referring to a specific DFU library, or just a common thing within the DFU libraries?

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -1108,7 +1113,7 @@ The code for integrating MCUboot into |NCS| is located in the :file:`ncs/nrf/mod
 
 The following list summarizes both the main changes inherited from upstream MCUboot and the main changes applied to the |NCS| specific additions:
 
-|no_changes_yet_note|
+* Version check for network core when downgrade prevention is enabled.
```
peknis, line 1116:
```suggestion
* Added a version check for network core when downgrade prevention is enabled.
```

peknis (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
sigvartmh: I did not add `* Added a new DFU SMP target for the image update to an external MCU by using the MCUmgr SMP Client."` this part if you are referring to that? or are you referring to something else?

sigvartmh: I've added this part as suggested
```

* :ref:`subsys_pcd` library:

  * Added function :c:func:`pcd_network_core_app_version` for reading peripheral CPU application version number.
  
 ```
 
 Should it be moved back to the PCD subsection? You suggested to remove it in a previous comment so I removed it.

peknis: Yep, that is what I mean. I need to find out who wrote it then.

nordicjm: That would be @jheiskan81

sigvartmh: last edited 82c0e739c7

peknis: No worries @sigvartmh, no need to move anything anywhere. We are good. Just need to squeeze Juha Heiskanen about the earlier entry under DFU libraries.

wiba-nordic (approved):

sigvartmh:
Sidewalk test was marked as unstable, re-running.

cvinayak:
@sigvartmh update SHA

annakielar (approved):

cvinayak:
Merging bypassing branch protections as CI has passed before with the PR.

PR #12413, opened 2023-09-22
doc: bluetooth: mesh: Split out mesh samples

mia-ko:
The list of Bluetooth samples, including the Bluetooth mesh samples, has gotten longer. Bluetooth mesh samples drown a little and are not easily spotted.
Splitting out mesh samples onto a separate page.

### DIFF
--- a/samples/bluetooth/mesh/ble_peripheral_lbs_coex/README.rst
+++ b/samples/bluetooth/mesh/ble_peripheral_lbs_coex/README.rst
@@ -1,18 +1,18 @@
 .. _bluetooth_ble_peripheral_lbs_coex:
 
-Bluetooth: Mesh and peripheral coexistence
-##########################################
+Bluetooth mesh: Coexistence with other LE services
+##################################################
 
 .. contents::
    :local:
    :depth: 2
 
-This sample demonstrates how to combine Bluetooth® mesh and Bluetooth Low Energy features in a single application.
+This sample demonstrates how to combine Bluetooth® mesh and another Bluetooth Low Energy (LE) service in a single application.
 
 Requirements
 ************
 
-The mesh and peripheral coexistence sample supports the following development kits:
+This sample supports the following development kits:
 
 .. table-from-sample-yaml::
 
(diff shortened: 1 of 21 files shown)


Balaklaka:
The Bluetooth mesh samples are now more hidden than it was before. Before it was several entries showing "Bluetooth mesh:" and you could see them by that there was several entries. Now it is just one entry at the bottom of the list, and that it not showing at all. This should not be a subpage of the Bluetooth samples. It should be move up one level so it is on the same level as Bluetooth, then it will be more similar as how it is under "Protocols".

Balaklaka (changes requested):
The Bluetooth mesh samples are now more hidden than it was before. Before it was several entries showing "Bluetooth mesh:" and you could see them by that there was several entries. Now it is just one entry at the bottom of the list, and that it not showing at all. This should not be a subpage of the Bluetooth samples. It should be move up one level so it is on the same level as Bluetooth, then it will be more similar as how it is under "Protocols".

mia-ko:
> I think the sample names should be changed: example "Bluetooth mesh: Chat".

As we discussed, I agree.
We only had doubts about the Bluetooth mesh: Peripheral coexistence, if it was ok to call it that and not Bluetooth mesh: Mesh and peripheral coexistence. Right?

samples/bluetooth/mesh/ble_peripheral_lbs_coex/README.rst
```diff
@@ -1,7 +1,7 @@
 .. _bluetooth_ble_peripheral_lbs_coex:
 
-Bluetooth: Mesh and peripheral coexistence
-##########################################
+Bluetooth mesh: Peripheral coexistence
```
omkar3141, line 3: Now I would rename this as: 
```suggestion
Bluetooth mesh: Coexistence with other LE services
```

The thing is "Peripheral" role already exist in all mesh nodes. All mesh nodes, with the default configuration are already, GATT peripherals by definition. However, all that magic happens inside the stack.

What this sample is showing, how to add "another" BLE service to the existing mesh node.

omkar3141 (changes requested):

samples/bluetooth/mesh/ble_peripheral_lbs_coex/README.rst
mia-ko: The way you explain it, it makes sense to me :) That said, could you then have a look at the Overview section of this sample and see if more text should be tweaked? Because now I am not sure myself.

Balaklaka (approved):

ludvigsj (approved):

omkar3141 (approved):

PR #12418, opened 2023-09-25
wifi: hostap: Avoid using internal supplicant APIs from other modules

rado17:


maxd-nordic:
thanks for working on this! it will be very useful.

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
```diff
@@ -24,6 +36,69 @@ LOG_MODULE_REGISTER(wifi_mgmt_ext, CONFIG_WIFI_MGMT_EXT_LOG_LEVEL);
...
+	params->ssid = ssid;
+	params->ssid_length = creds->header.ssid_len;
+
+	/* PSK (optional) */
+	if (creds->password_len > 0) {
+		psk = (char *)k_malloc(creds->password_len + 1);
+		if (!psk) {
+			LOG_ERR("Failed to allocate memory for PSK\n");
+			ret = -ENOMEM;
+		goto no_mem_out;
```
sachinthegreen, line 64: Indentation

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
```diff
@@ -24,6 +36,69 @@ LOG_MODULE_REGISTER(wifi_mgmt_ext, CONFIG_WIFI_MGMT_EXT_LOG_LEVEL);
...
 
+static int __stored_creds_to_params(struct wifi_credentials_personal *creds,
+				struct wifi_connect_req_params *params)
+{
+	char *ssid = NULL;
+	char *psk = NULL;
+	int ret;
+
+	/* SSID */
+	ssid = (char *) k_malloc(creds->header.ssid_len + 1);
```
sachinthegreen, line 47: No space between `(char *)`  and `k_malloc`

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
```diff
@@ -156,25 +213,59 @@ static int add_static_network_config(void)
...
-	return add_network_from_credentials_struct_personal(&creds);
+	return add_network_from_credentials_struct_personal(&creds, iface);
 #else
 	return 0;
 #endif /* defined(CONFIG_WIFI_CREDENTIALS_STATIC) */
 }
 
+static void wifi_mgmt_ext_connect_result(struct net_mgmt_event_callback *cb)
+{
+	const struct wifi_status *status = (const struct wifi_status *) cb->info;
```
sachinthegreen, line 224: No space between `(const struct wifi_status *)` and  `cb->info`

sachinthegreen (changes requested):

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
rado17: Done

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
rado17: Done

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
rado17: Done

modules/hostap/src/supp_api.c
```diff
@@ -239,6 +239,18 @@ int z_wpa_supplicant_connect(const struct device *dev,
 	_wpa_cli_cmd_v("set_network %d scan_ssid 1", resp.network_id);
 	_wpa_cli_cmd_v("set_network %d key_mgmt NONE", resp.network_id);
 	_wpa_cli_cmd_v("set_network %d ieee80211w 0", resp.network_id);
+	if ((params->band & BIT(WIFI_FREQ_BAND_2_4_GHZ)) == BIT(WIFI_FREQ_BAND_2_4_GHZ)) {
```
krish2718, line 242: This capability of setting channels, should be in a separate commit.

krish2718: And have you tried skipping this altogether and let WPA supplicant derive these from the driver's capability? Its not ideal to duplicate these esp. without the regulatory domain information.

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
```diff
@@ -24,6 +24,60 @@ LOG_MODULE_REGISTER(wifi_mgmt_ext, CONFIG_WIFI_MGMT_EXT_LOG_LEVEL);
...
+		return -ENOMEM;
+	}
+	memset(ssid, 0, creds->header.ssid_len + 1);
+	strncpy(ssid, creds->header.ssid, creds->header.ssid_len);
+	params->ssid = ssid;
+	params->ssid_length = creds->header.ssid_len;
+
+	/* PSK (optional) */
+	if (creds->password_len > 0) {
+		psk = (char *)k_malloc(creds->password_len + 1);
```
krish2718, line 60: use `k_zalloc` to avoid extra memset.

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
```diff
@@ -24,6 +24,60 @@ LOG_MODULE_REGISTER(wifi_mgmt_ext, CONFIG_WIFI_MGMT_EXT_LOG_LEVEL);
...
+		if (!psk) {
+			LOG_ERR("Failed to allocate memory for PSK\n");
+			return -ENOMEM;
+		}
+		memset(psk, 0, creds->password_len + 1);
+		strncpy(psk, creds->password, creds->password_len);
```
krish2718, line 67: no check for truncation and missing NULL termination?

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
```diff
@@ -24,6 +24,60 @@ LOG_MODULE_REGISTER(wifi_mgmt_ext, CONFIG_WIFI_MGMT_EXT_LOG_LEVEL);
...
+	/* Defaults */
+	params->security = WIFI_SECURITY_TYPE_PSK;
+	params->mfp = WIFI_MFP_OPTIONAL;
+	params->channel = WIFI_CHANNEL_ANY;
+	params->timeout = CONNECTION_TIMEOUT;
+
+	/* Security type (optional) */
+	if (creds->header.type <= WIFI_SECURITY_TYPE_MAX) {
+		params->security = creds->header.type;
+	} else {
```
krish2718, line 81: The current default is overwritten in both branches, so, if you make this as default then `else` branch can be removed.

krish2718 (changes requested):

modules/hostap/src/supp_api.c
rado17: Should we then remove band configuration support from stored wifi credentials as well?

krish2718: Well, we can still keep it and implement, but we need to dynamically query the supported channels from the driver and use that rather than a static list.

sachinthegreen (approved):

kapbh (approved):

modules/hostap/src/supp_api.c
rado17: Done

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
rado17: Looks like k_zalloc is not available in zephyr. Keeping this as is.

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
rado17: Moved to snprintf which takes care of adding the null termination and added checks for truncation.

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
rado17: Done

modules/hostap/src/supp_api.c
```diff
@@ -242,6 +247,67 @@ int z_wpa_supplicant_connect(const struct device *dev,
...
+				retval = snprintf(chan_list + offset, 6, " %d",
+						     mode->channels[i].freq);
+				offset += retval;
+			}
+
+			printf("chanlist_24: %s\n", chan_list);
+			_wpa_cli_cmd_v("set_network %d freq_list%s", resp.network_id, chan_list);
+			k_free(chan_list);
+		}
+	}
```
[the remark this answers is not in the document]
rado17: Done.

modules/hostap/src/supp_api.c
```diff
@@ -144,6 +144,73 @@ static void supp_shell_connect_status(struct k_work *work)
...
+
+			retval = 0;
+			offset = 0;
+			for (i = 0; i < mode->num_channels; i++) {
+				retval = snprintf(_chan_list + offset, 6, " %d",
+						  mode->channels[i].freq);
+				offset += retval;
+			}
+			*chan_list = _chan_list;
+		}
```
krish2718, line 180: This logic is common for 2.4/5, so, can be refactored?

rado17: Done

modules/hostap/src/supp_api.c
```diff
@@ -144,6 +144,52 @@ static void supp_shell_connect_status(struct k_work *work)
...
+	struct hostapd_hw_modes *mode = NULL;
+	int i, j;
+	int offset, retval;
+	int size;
+	char *_chan_list;
+
+	if ((band & BIT(WIFI_FREQ_BAND_2_4_GHZ)) == BIT(WIFI_FREQ_BAND_2_4_GHZ)) {
+		mode = NULL;
+		for (j = 0; j < ((wpa_s)->hw).num_modes; j++) {
+			if (((wpa_s)->hw).modes[j].mode != HOSTAPD_MODE_IEEE80211A) {
```
krish2718, line 158: This assumes dual band only, what about 6GHz?

modules/hostap/src/supp_api.c
```diff
@@ -144,6 +144,52 @@ static void supp_shell_connect_status(struct k_work *work)
...
+			}
+		}
+	} else if ((band & BIT(WIFI_FREQ_BAND_5_GHZ)) == BIT(WIFI_FREQ_BAND_5_GHZ)) {
+		mode = NULL;
+		for (j = 0; j < ((wpa_s)->hw).num_modes; j++) {
+			if (((wpa_s)->hw).modes[j].mode == HOSTAPD_MODE_IEEE80211A) {
+				mode = &((wpa_s)->hw).modes[j];
+				break;
+			}
+		}
```
krish2718, line 170: Even this can be refactored if you can get mode i.e., `HOSTAPD_MODE_*` from `WIFI_FREQ_BAND_*`.

modules/hostap/src/supp_api.c
```diff
@@ -144,6 +144,52 @@ static void supp_shell_connect_status(struct k_work *work)
...
+	}
+
+	if (mode) {
```
krish2718, line 173: Return early to save indent.

```suggestion
	if (!mode) {
	    return 0;
	}
```

subsys/net/lib/wifi_credentials/Kconfig
```diff
@@ -40,7 +40,7 @@ endchoice
 
 config WIFI_CREDENTIALS_MAX_ENTRIES
 	int "Number of supported WiFi credentials"
-	default 2
+	default 1
```
krish2718, line 43: > Hence reducing the max entries to 1 to avoid these 2 issues.

This sounds like a workaround than actual fix, what if the user configures this to 2?

krish2718:
> attempt for connection is initiated with the subsequent credentials in the stored list.

This doesn't seem right, second credentials should only be used if first one isn't successful.

> Under DISCONNECTED event processing, l2_conn module
tries to read the flags which again is done under conn_binding lock
which causes a deadlock.

This sounds like a bug in the code, we should fix the locking.

modules/hostap/src/supp_api.c
rado17: Added checks for 6GHz

modules/hostap/src/supp_api.c
rado17: Done

modules/hostap/src/supp_api.c
rado17: Made changes to identify hardware mode based on band.

modules/hostap/src/supp_api.c
```diff
@@ -145,6 +145,53 @@ static void supp_shell_connect_status(struct k_work *work)
...
+	int i;
+	int offset, retval;
+	int size;
+	char *_chan_list;
+
+	mode = get_mode_by_band(wpa_s, band);
+	if (!mode) {
+		return -EINVAL;
+	}
+	size = ((mode->num_channels) * 6) + 1;
```
krish2718, line 177: 6?

modules/hostap/src/supp_api.c
```diff
@@ -145,6 +145,53 @@ static void supp_shell_connect_status(struct k_work *work)
 	k_mutex_unlock(&wpa_supplicant_mutex);
 }
 
+static struct hostapd_hw_modes *get_mode_by_band(struct wpa_supplicant *wpa_s, uint8_t band)
+{
+	enum hostapd_hw_mode hw_mode;
+	bool is_6ghz = (band & BIT(WIFI_FREQ_BAND_6_GHZ))?true:false;
```
krish2718, line 151: Ternary needs spaces around

modules/hostap/src/supp_api.c
```diff
@@ -145,6 +145,53 @@ static void supp_shell_connect_status(struct k_work *work)
...
+	int offset, retval;
+	int size;
+	char *_chan_list;
+
+	mode = get_mode_by_band(wpa_s, band);
+	if (!mode) {
+		return -EINVAL;
+	}
+	size = ((mode->num_channels) * 6) + 1;
+	_chan_list = (char *)k_malloc(size);
```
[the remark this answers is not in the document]
rado17: Removed the casting

modules/hostap/src/supp_api.c
rado17: Done

modules/hostap/src/supp_api.c
```diff
@@ -145,6 +145,53 @@ static void supp_shell_connect_status(struct k_work *work)
 	k_mutex_unlock(&wpa_supplicant_mutex);
 }
 
+static struct hostapd_hw_modes *get_mode_by_band(struct wpa_supplicant *wpa_s, uint8_t band)
+{
+	enum hostapd_hw_mode hw_mode;
+	bool is_6ghz = (band & BIT(WIFI_FREQ_BAND_6_GHZ)) ? true:false;
```
krish2718, line 151: `(band & BIT(WIFI_FREQ_BAND_6_GHZ)) ? true : false;`

modules/hostap/src/supp_api.c
```diff
@@ -145,6 +145,53 @@ static void supp_shell_connect_status(struct k_work *work)
...
+
+	if ((band & BIT(WIFI_FREQ_BAND_2_4_GHZ))) {
```
krish2718, line 153: Just realized that the Wi-Fi management bands are not bitmaps, just normal enum values.

modules/hostap/src/supp_api.c
```diff
@@ -145,6 +145,53 @@ static void supp_shell_connect_status(struct k_work *work)
...
+	int i;
+	int offset, retval;
+	int size;
+	char *_chan_list;
+
+	mode = get_mode_by_band(wpa_s, band);
+	if (!mode) {
+		return -EINVAL;
+	}
+	size = ((mode->num_channels) * CHAN_NUM_LEN) + 1;
```
krish2718, line 177: a newline before

modules/hostap/src/supp_api.h
```diff
@@ -12,6 +12,7 @@
 
 #define MAX_SSID_LEN 32
 #define MAC_ADDR_LEN 6
+#define CHAN_NUM_LEN 6
```
krish2718, line 15: `x.abc` + NULL terminator?

rado17: abcd+" "+NULL terminator

modules/hostap/src/supp_api.c
rado17: Yes. Used BIT macro here to differentiate the case where no band is assigned (band initialized to 0 in configuration API) and the case where band is set to 2.4GHz.

modules/hostap/src/supp_api.c
rado17: Done

modules/hostap/src/supp_api.c
rado17: Done.

modules/hostap/src/supp_api.c
krish2718: But the input is not modified `params->band` is a normal value not BIT()?

modules/hostap/src/supp_api.h
krish2718: May be adding a comment might help including `space is needed to concatenate`

modules/hostap/src/supp_api.c
rado17: We are setting BIT() for param band in wifi_ext_mgmt module:

```
if (creds->header.flags & WIFI_CREDENTIALS_FLAG_2_4GHz) {
		params->band |= BIT(WIFI_FREQ_BAND_2_4_GHZ);
	}

	if (creds->header.flags & WIFI_CREDENTIALS_FLAG_5GHz) {
		params->band |= BIT(WIFI_FREQ_BAND_5_GHZ);
	}
```

modules/hostap/src/supp_api.h
rado17: Will do

modules/hostap/src/supp_api.c
krish2718: That is wrong, as per docs `params->band` is just an enum, so, any driver will try to process it as enum and might fail. Though I don't see this param used currently.

rado17: Do you see any issue if I change the enum definition i.e., set 2,4GHz to 1?

krish2718: Nope, its fine. But that would mean an upstream change :)

rado17: I would still suggest that we use the BIT() approach, since having 0 as a valid value for a field in structure seems problematic. Specially in this case, since, for a structure instantiated to 0, the default band becomes 2.4GHz-only. 
Also, this avoids the upstream change.

krish2718: Sure, in that case then please use an intermediate variable rather than modify `params->band`, in fact `params` should be treated as `const` (missed this in the original implementation).

rado17: As discussed, avoided usage of `BIT()` for `params->band`. Instead setting band to `WIFI_FREQ_BAND_UNKNOWN` in case band is not configure.

modules/hostap/src/supp_api.c
```diff
@@ -145,6 +145,54 @@ static void supp_shell_connect_status(struct k_work *work)
 	k_mutex_unlock(&wpa_supplicant_mutex);
 }
 
+static struct hostapd_hw_modes *get_mode_by_band(struct wpa_supplicant *wpa_s, uint8_t band)
+{
+	enum hostapd_hw_mode hw_mode;
+	bool is_6ghz = (band & BIT(WIFI_FREQ_BAND_6_GHZ)) ? true : false;
```
krish2718, line 151: This still has BIT() .

modules/hostap/src/supp_api.h
krish2718: Don't see the comment?

rado17: Added now.

modules/hostap/src/supp_api.c
rado17: Corrected this.

PR #12419, opened 2023-09-25
lib: lte_link_control: add name to LTE priority symbol

JordanYates:
Add a name to the LTE mode priority symbol so that the default can be updated by other kconfig files.
`LTE_MODE_PREFERENCE` would be preferred to align with the choice options, but this is already taken by the final integer symbol.

### DIFF
--- a/lib/lte_link_control/Kconfig
+++ b/lib/lte_link_control/Kconfig
@@ -177,7 +177,7 @@ config LTE_NETWORK_MODE_LTE_M_NBIOT_GPS
 
 endchoice
 
-choice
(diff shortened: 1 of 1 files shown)


JordanYates:
Sadly I am not in a position to sign a CLA on behalf on an entire federal government organisation

trantanen:
We'll add a symbol for this choice symbol.

carlescufi:
> We'll add a name for this choice symbol.

Thanks @trantanen!

tokangas:
@JordanYates I created PR #12426 which includes this change. I did use a different name to try to keep the naming consistent.

PR #12421, opened 2023-09-25
bluetooth: Fix assert when handling illegal response data cmd

olivier-le-sage:
Fixes an assert that could happen if the LE set periodic advertising response data was issued more than once without fetching the command complete event.

### DIFF
--- a/subsys/bluetooth/controller/hci_internal.c
+++ b/subsys/bluetooth/controller/hci_internal.c
@@ -34,6 +34,10 @@ static struct
 	uint8_t raw_event[BT_BUF_EVT_RX_SIZE];
 } cmd_complete_or_status;
 
+#if defined(CONFIG_BT_CTLR_SDC_PAWR_SYNC)
+static bool padv_response_data_cmd_pending;
+#endif
+
 static hci_internal_user_cmd_handler_t user_cmd_handler;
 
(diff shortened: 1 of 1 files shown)


olivier-le-sage:
test https://projecttools.nordicsemi.no/bitbucket/projects/NCS-TEST/repos/test-fw-nrfconnect-ble/pull-requests/4912/overview

subsys/bluetooth/controller/hci_internal.c
```diff
@@ -1358,6 +1360,7 @@ int hci_internal_cmd_put(uint8_t *cmd_in)
 		 */
 
 		cmd_complete_or_status.occurred = false;
+		padv_response_data_cmd_pending = true;
 	}
```
seko-nordic, line 1364: may be we should put that entire `if (opcode == SDC_HCI_OPCODE_CMD_LE_SET_PERIODIC_ADV_RESPONSE_DATA` under corresponding KConfig ifdef. Currently it eats few cycles for every event we generate, even for an application that has nothing to do with PAwR.

seko-nordic (approved):

rugeGerritsen (changes requested):
I would be happy if you could add some ifdefs. That saves power

rugeGerritsen (approved):

PR #12434, opened 2023-09-26
applications: serial_lte_modem: make UART message sending multithread-safe

tomi-font:
And detect (and block) attempts to send UART messages from ISRs.

### DIFF
--- a/applications/serial_lte_modem/src/slm_at_host.c
+++ b/applications/serial_lte_modem/src/slm_at_host.c
@@ -43,8 +43,6 @@ struct at_param_list slm_at_param_list;
 uint8_t slm_at_buf[SLM_AT_MAX_CMD_LEN];
 uint8_t slm_data_buf[SLM_MAX_MESSAGE_SIZE];
 
-static char rsp_buf[SLM_AT_MAX_RSP_LEN];
-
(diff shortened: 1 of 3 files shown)


SeppoTakalo (approved):

applications/serial_lte_modem/src/slm_at_host.c
```diff
@@ -653,13 +651,28 @@ void rsp_send_error(void)
 
 void rsp_send(const char *fmt, ...)
 {
+	static K_MUTEX_DEFINE(mutex_rsp_buf);
+	static char rsp_buf[SLM_AT_MAX_RSP_LEN];
+
+	if (k_is_in_isr()) {
```
MarkusLassila, line 657: This check should be in rsp_send_ok, rsp_send_error and data_send as well.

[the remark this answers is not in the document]
MarkusLassila: Assert is also a possibility, but it does not convey information what is wrong. I liked the message option, but assert is fine as well.

MarkusLassila: Or it could be in slm_uart_tx_write.

MarkusLassila (approved):

VTPeltoketo (approved):

PR #12439, opened 2023-09-26
drivers: wifi: Update RPU patch to rev#5322b40bc14

karun2796:
[SHEL-3941]: Firmware change for  association failure to the new ssid  in linux case.


karun2796:
Closing PR. Changes merged to ncs/main with another PR

PR #12441, opened 2023-09-27
application: serial_lte_modem: improve MQTT for long KeepAlive

junqingzou:
Detect disconnection as early as possible, namely socket poll error, MQTT input error, MQTT ping error, MQTT ping response timeout, etc. Abort MQTT connection once disconnection is detected and block all following MQTT operation until re-connection.

Set CONFIG_MQTT_CLEAN_SESSION=y to clean session on every connect.

Correct documentation mistakes.

JIRA: NCSDK-23744


applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -133,7 +133,12 @@ Response syntax
 
 ::
 
-   #XMQTTCON: <client_id>,<url>,<port>[,<sec_tag>]]
+   #XMQTTCON: <status>,<client_id>,<url>,<port>[,<sec_tag>],<keepalive>]
+
+* The ``<status>`` value is an integer.
```
wiba-nordic, line 138:
```suggestion
* The ``<status>`` value is an integer.
  It can accept one of the following values:
```

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -133,7 +133,12 @@ Response syntax
...
+
+    * ``0`` - MQTT disconnected.
+    * ``1`` - MQTT connected.
```
wiba-nordic, line 141:
```suggestion
    * ``0`` - MQTT is not connected.
    * ``1`` - MQTT is connected.
```
This presumes that the read command is actively used to gain this information. If this is instead something that gets automatically reported when MQTT connects or disconnects, then the original, active wording is correct.

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -143,20 +148,46 @@ Response syntax
   It indicates the MQTT broker port.
 * The ``<sec_tag>`` value is an integer.
   It indicates the credential of the security tag used for establishing a secure connection.
+* The ``<keepalive>`` value is an integer.
+  It indicates the MQTT KeepAlive value (in seconds) used in current MQTT connection.
```
wiba-nordic, line 152:
```suggestion
  It indicates the MQTT Keep Alive value (in seconds) used in the current MQTT connection.
```
I believe the feature name is "Keep Alive" with two words.

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -143,20 +148,46 @@ Response syntax
...
    AT#XMQTTCON?
-   #XMQTTCON: "MyMQTT-Client-ID","","","mqtt.server.com",1883
+   #XMQTTCON: 1,"MyMQTT-Client-ID","","","mqtt.server.com",1883,60
    OK
 
 Test command
 ------------
 
-The test command is not supported.
+The test command tests the existence of the command and provides information about the type of its subparameters.
```
wiba-nordic, line 166:
```suggestion
The test command tests the existence of the ``#XMQTTCON`` command and provides information about its subparameters.
```
Subparameters, or just parameters?

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -143,20 +148,46 @@ Response syntax
...
+::
+
+   #XMQTTCON=?
+
+Response syntax
+~~~~~~~~~~~~~~~
+
+::
+
+   #XMQTTCON: (0,1,2),<client_id>,<username>,<password>,<url>,<port>,<sec_tag>
```
wiba-nordic, line 180: These should be explained. If they're the same as for the Set command, you could use an `.. include::` like was recently added for the GNSS AT commands.

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -143,20 +148,46 @@ Response syntax
...
+
+Examples
+~~~~~~~~
+
+::
+
+   AT#XMQTTCON=?
+   #XMQTTCON: (0,1,2),<client_id>,<username>,<password>,<url>,<port>,<sec_tag>
```
wiba-nordic, line 188: This looks like the syntax, not an example response.

wiba-nordic (changes requested):
In addition to the other points, this should have a changelog entry since it changes some of the command syntax.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -313,6 +344,23 @@ static void client_init(void)
 	client.tx_buf = tx_buffer;
 	client.tx_buf_size = sizeof(tx_buffer);
 
+	/* MQTT Will configuration */
+	struct mqtt_topic will_topic;
```
MarkusLassila, line 348: Nothing guarantees that these memory areas exist after you exit this function. These should be static or alternatively defined in previous level, and brought as parameters, if only used in mqtt_connect.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -370,7 +418,7 @@ static int do_mqtt_disconnect(void)
 	int err;
 
 	if (!ctx.connected) {
-		return -ENOTCONN;
+		return -EAGAIN;
```
MarkusLassila, line 421: EAGAIN indicates that the operation would succeed at later point of time, which is not the case in here, if we have to close the connection and reopen it. Same for others.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -217,41 +220,69 @@ static void mqtt_thread_fn(void *arg1, void *arg2, void *arg3)
...
-			}
-		}
+		/* MQTT v3.1.1: Note that a Server is permitted to disconnect a Client that it
+		 * determines to be inactive or non-responsive at any time, regardless of the
+		 * Keep Alive value provided by that Client.
+		 */
 		if ((fds.revents & POLLERR) == POLLERR) {
 			LOG_ERR("POLLERR");
-			mqtt_abort(&client);
+			err = -ENETDOWN;
```
MarkusLassila, line 238: This should likely be -EIO.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -217,41 +220,69 @@ static void mqtt_thread_fn(void *arg1, void *arg2, void *arg3)
...
 			break;
 		}
 		if ((fds.revents & POLLHUP) == POLLHUP) {
 			LOG_ERR("POLLHUP");
-			mqtt_abort(&client);
+			err = -ECONNRESET;
```
MarkusLassila, line 243: POLLHUP indicates gracefull termination, err should be 0.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -217,41 +220,69 @@ static void mqtt_thread_fn(void *arg1, void *arg2, void *arg3)
...
 			break;
 		}
 		if ((fds.revents & POLLNVAL) == POLLNVAL) {
 			LOG_ERR("POLLNVAL");
-			mqtt_abort(&client);
+			err = -ENOTCONN;
```
MarkusLassila, line 248: POLLNVAL indicates that socket has been closed, which is rather unexpected. Could use ENOENT, but no strong opinion on this.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -217,41 +220,69 @@ static void mqtt_thread_fn(void *arg1, void *arg2, void *arg3)
...
 			break;
 		}
+		if ((fds.revents & POLLIN) == POLLIN) {
```
MarkusLassila, line 251: All of the above errors skip this, which is problematic, because there can also be data left to be read when connection is closed. It might be better to still read the data in mqtt_input, which might even allow for returning of a more exact correct error code, if it sets errno.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -217,41 +220,69 @@ static void mqtt_thread_fn(void *arg1, void *arg2, void *arg3)
...
+			err = mqtt_input(&client);
+			if (err != 0) {
+				LOG_ERR("ERROR: mqtt_input %d", err);
+				break;
+			}
+			/* MQTT v3.1.1: If a Client does not receive a PINGRESP Packet within a
+			 *  reasonable amount of time after it has sent a PINGREQ, it SHOULD close
+			 *  the Network Connection to the Server.
+			 */
+			if (client.unacked_ping > 0) {
```
MarkusLassila, line 261: How does this work for scenario where we send the ping request, but at the same time the server sends us anoher message? Wont we then hit this and connect erronously?

MarkusLassila (changes requested):

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
junqingzou: The read command is just to report, not get input

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
junqingzou: Yes

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
junqingzou: The response should be
#XMQTTCON: (list of op),<client_id>,<username>,<password>,<url>,<port>,<sec_tag>

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
wiba-nordic: What I was more looking for, is this command used by the user? Or is it something that gets automatically run?

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: Yes they should be static

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: cannot be 0 as need to do abort if unexpected down below.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: The spec says client could disconnect if no response for mqtt ping

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: We can do mqtt_input() before poll abnormal checks. If mqtt_input() returns error we break from there.

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
junqingzou: Copy/paste from other SLM commands

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -51,6 +53,7 @@ static uint16_t mqtt_broker_port;
 static char mqtt_clientid[MQTT_MAX_CID_LEN + 1];
 static char mqtt_username[SLM_MAX_USERNAME + 1];
 static char mqtt_password[SLM_MAX_PASSWORD + 1];
+static uint8_t mqtt_will_topic[MQTT_MAX_CID_LEN + 5];
```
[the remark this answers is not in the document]
junqingzou: It should be static as well as it pointed by will_topic.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: "err" only for logging before abort

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
junqingzou: Yes it's for user to call
AT#XMQTTCON=....   SET comannd
AT#XMQTTCON?       READ command
AT#XMQTTCON=?    TEST command

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: In libc errno EAGAIN means no more context, i.e. wrong context in my understanding.
Anyway any errno will do as we only want to return ERROR for the AT command.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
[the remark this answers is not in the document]
junqingzou: As the will topic is one-time usage during `mqtt_connect`, let's remove this buffer and re-use `pub_topic` instead.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -313,6 +346,20 @@ static void client_init(void)
 	client.tx_buf = tx_buffer;
 	client.tx_buf_size = sizeof(tx_buffer);
 
+	/* MQTT Will configuration */
+	strcpy((char *)pub_topic, mqtt_clientid);
```
MarkusLassila, line 350: I think that this overwrites unwanted memory. pub_topic is not a memory area. It works by having a pointer to memory area.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
MarkusLassila: Generally, EAGAIN and it's sister code EWOULDBLOCK are used when something would fail, but could succeed in other conditions. I think that in this case, if the user needs to connect again, it would be better to use previous -ENOTCONN or other error that indicates that we are not connected/have disconnected.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
MarkusLassila: I understand where you are coming from, but the error is misleading in this case. I think that even -ENOTCONN would be better. Best would be to return 0.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
MarkusLassila: This works well for a scenario where we ping and after some time receive a PUB-message, but if we receive a PUB-message when we have just send the ping, we will hit the POLLIN branch and now have unacked ping. Then we will disconnect erroneously.

This should have some kind of time treshold or other mechanism to prevent above scenario.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: Original MQTT poll behavior was copied from TCP/UDP client

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: `static uint8_t pub_topic[MQTT_MAX_TOPIC_LEN];`
It's not used until MQTTPUB.

MarkusLassila: This was my bad, somehow got mqtt_topic and pub_topic confused.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
MarkusLassila: Yes, the same behavior is in TCP server, and it is wrong there as well.

junqingzou: I don't think there is a need to discuss further what error code to return in this PR as not serving the main purpose.
But it can be done in a new task, to review all the poll threads and other scenarios on the return errno in SLM-wide

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
MarkusLassila: Discussed this with @junqingzou, we currently have a similar implementation in nrf_cloud. And while the implementation does not work well in this particular case, it is not critical as the connection can terminate for multitude of reasons and there should either be user re-trial or re-trial logic in case of usage as SW component.

If this becomes problematic, the correct way to fix this, would be to note that  we have sent ping req, lower the poll timeout and check whether we have received ping resp during the timeout. It is not straightforward solution as there can again be other packets coming through this time. 

It is possible that this cannot be improved without a separate timer or having time as a concept in here.

But I think that we can go with this for now. A quick fix in possible problem case is to set the unacked_ping > 1.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
MarkusLassila: Can be done as a separate change.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
MarkusLassila: @junqingzou: Please change this to ENOTCONN or other value that indicates not connected / disconnected.

junqingzou:
DNM: solution not verified by customer

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
wiba-nordic: Just realized that this one is slightly wrong. Should be:
```suggestion
* The ``<status>`` value is an integer.
  It can have one of the following values:
```

junqingzou:
Removed DNM.
Investigation on customer issue shows it's related to carrier network behavior.
This PR tries with minimal code change to help detecting MQTT connection loss in time and blocks improper operation after it. The Will publish mechanism could not be verified with MQTT brokers so it's removed, too.

wiba-nordic (approved):

wiba-nordic (commented):
Ah, just realized that the changelog entry is still required. Please add that.

junqingzou:
> Ah, just realized that the changelog entry is still required. Please add that.

Only code optimization and no AT interface changes, can we save updating release note?

wiba-nordic:
Isn't `<value>` new for `#XMQTTCON`?

junqingzou:
> Isn't `<value>` new for `#XMQTTCON`?

Not "new" in code, but doc was mistaken and corrected in this PR

wiba-nordic:
Ah, then we don't need the changelog entry.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -217,41 +217,70 @@ static void mqtt_thread_fn(void *arg1, void *arg2, void *arg3)
...
+
+		/* poll timeout or revent, send KEEPALIVE */
+		err = mqtt_live(&client);
+		if (err != 0 && err != -EAGAIN) {
+			LOG_ERR("ERROR: mqtt_live %d", err);
 			break;
 		}
 	}
 
+	if (ctx.connected && err != 0) {
```
MarkusLassila, line 274: Shouldn't this terminate the connection if ctx.connected is true at this point? It would be simpler and would allow the POLLHUP to return 0.

applications/serial_lte_modem/prj.conf
```diff
@@ -94,6 +94,10 @@ CONFIG_MODEM_INFO_ADD_DATE_TIME=n
 # UUID and JWT
 CONFIG_MODEM_JWT=y
 
+# MQTT client
+#CONFIG_MQTT_KEEPALIVE=60
```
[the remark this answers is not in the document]
junqingzou: lib_mqtt default 60s, this is for customer to quickly change keepalive

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: This is to protect `#XMQTTCON=0`, without it there will be two URC, `: 1,0`  and `: 1,-113`

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -133,7 +133,13 @@ Response syntax
 
 ::
 
-   #XMQTTCON: <client_id>,<url>,<port>[,<sec_tag>]]
+   #XMQTTCON: <status>,<client_id>,<url>,<port>[,<sec_tag>]
```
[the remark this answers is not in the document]
junqingzou: Another round of doc review 🤦‍♂️

MarkusLassila (approved):

wiba-nordic (approved):

PR #12445, opened 2023-09-27
doc: nrf21540: fix typo in shield name

greg-fer:
Fixed -DSHIELD=nrf21540ek_fwd to -DSHIELD=nrf21540_ek_fwd. Reported by a customer.

### DIFF
--- a/applications/nrf5340_audio/README.rst
+++ b/applications/nrf5340_audio/README.rst
@@ -396,12 +396,12 @@ Adding FEM support
 You can add support for the nRF21540 front-end module (FEM) to this application by using one of the following options, depending on how you decide to build the application:
 
 * If you opt for :ref:`nrf53_audio_app_building_script`, add the ``--nrf21540`` to the script's building command.
-* If you opt for :ref:`nrf53_audio_app_building_standard`, add the ``-DSHIELD=nrf21540ek_fwd`` to the ``west build`` command.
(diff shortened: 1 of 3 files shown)


erikrobstad (approved):

koffes (commented):
Has this changed recently? If so, we need to change our build script as well. Any comments @rick1082 ?

rick1082 (approved):

greg-fer:
> Has this changed recently? If so, we need to change our build script as well. Any comments @rick1082 ?

No clue. @jciupis , would you know?

erikrobstad:
That is a good point @koffes.
We have been able to build for nRF21540 using our script before, so it has probably changed at some point

rick1082:
Hello all, thanks for the remind!
We might need to reconsider this PR, since the name just updated from this commit:
https://github.com/nrfconnect/sdk-nrf/commit/18c27de5eb3d43ef4583210fe69d50b8b657abf6

I saw the issue report from the customer, I would assume there's a mismatch between the SDK version they are using now and the doc.

erikrobstad (changes requested):
We can not merge this change.

greg-fer:
Clarified with @rick1082 and this is going against the recent changes from https://github.com/nrfconnect/sdk-nrf/commit/18c27de5eb3d43ef4583210fe69d50b8b657abf6

PR #12446, opened 2023-09-27
drivers: wifi: Remove bss ie and len setting in auth info

ajayparida:
Zephyr case firmware is not expecting the bss ie field. As bss ie field is valid only for cfg80211 offload case. The value ibss ie length for Zephyr case set to 0.


sachinthegreen (approved):

rado17 (approved):

drivers/wifi/nrf700x/zephyr/src/zephyr_wpa_supp_if.c
```diff
@@ -711,12 +711,14 @@ int nrf_wifi_wpa_supp_authenticate(void *if_priv, struct wpa_driver_auth_params
...
 
 	if (params->ie) {
-		memcpy(auth_info.bss_ie.ie, params->ie, params->ie_len);
+		if (params->ie_len > NRF_WIFI_MAX_IE_LEN) {
+			memcpy(&auth_info.ie.ie, params->ie, NRF_WIFI_MAX_IE_LEN);
+			auth_info.ie.ie_len = NRF_WIFI_MAX_IE_LEN;
+		} else {
+			memcpy(&auth_info.ie.ie, params->ie, params->ie_len);
+			auth_info.ie.ie_len = params->ie_len;
+		}
```
sachinthegreen, line 720: Why not use a `len` variable and set it to min of `NRF_WIFI_MAX_IE_LEN` and `params->ie_len` ?

sachinthegreen (changes requested):

sachinthegreen (approved):

PR #12447, opened 2023-09-27
Bluetooth: define BT_COMPANY_ID_NORDIC

jori-nordic:
Also set BT_COMPANY_ID to BT_COMPANY_ID_NORDIC.

Making a kconfig option makes it also usable outside sdk-nrf.


tokangas (approved):
Ok for `modem_shell`.

alwa-nordic (approved):

KAGA164 (approved):
LGTM. I would consider changing  CONFIG_BT_COMPANY_ID_NORDIC->CONFIG_BT_COMPANY_ID in C language related files

samples/bluetooth/mesh/chat/include/chat_cli.h
```diff
@@ -22,31 +22,28 @@ extern "C" {
...
 /* .. include_startingpoint_chat_cli_rst_1 */
-/** Company ID of the Bluetooth Mesh Chat Client model. */
-#define BT_MESH_CHAT_CLI_VENDOR_COMPANY_ID    0x0059
-
 /** Model ID of the Bluetooth Mesh Chat Client model. */
 #define BT_MESH_CHAT_CLI_VENDOR_MODEL_ID      0x000A
 
 /** Non-private message opcode. */
 #define BT_MESH_CHAT_CLI_OP_MESSAGE BT_MESH_MODEL_OP_3(0x0A, \
-				       BT_MESH_CHAT_CLI_VENDOR_COMPANY_ID)
```
omkar3141, line 33: I believe this should not be changed. We want to allow people to update/expand this model and use it for their own purpose and in that case, substitute this number with their own company ID. Applies to all changes in this file.

include/bluetooth/mesh/vnd/dm_common.h
```diff
@@ -21,7 +21,6 @@
 extern "C" {
 #endif
 
-#define BT_MESH_VENDOR_COMPANY_ID 0x0059
```
omkar3141, line 24: I believe this should not be changed. We want to allow people to update/expand this model and use it for their own purpose and in that case, substitute this number with their own company ID. Applies to all changes in this file.

subsys/bluetooth/mesh/shell/shell_dm_cli.c
```diff
@@ -35,7 +35,7 @@ static int cmd_cfg(const struct shell *shell, size_t argc, char *argv[])
 		return err;
 	}
 
-	if (!mod && !shell_vnd_model_first_get(BT_MESH_VENDOR_COMPANY_ID,
```
omkar3141, line 38: I believe this should not be changed. We want to allow people to update/expand this model and use it for their own purpose and in that case, substitute this number with their own company ID. Applies to all changes in this file.

omkar3141 (changes requested):

samples/bluetooth/mesh/chat/include/chat_cli.h
```diff
@@ -23,7 +23,7 @@ extern "C" {
 
 /* .. include_startingpoint_chat_cli_rst_1 */
 /** Company ID of the Bluetooth Mesh Chat Client model. */
-#define BT_MESH_CHAT_CLI_VENDOR_COMPANY_ID    0x0059
+#define BT_MESH_CHAT_CLI_VENDOR_COMPANY_ID    CONFIG_BT_COMPANY_ID
```
omkar3141, line 26:
```suggestion
#define BT_MESH_CHAT_CLI_VENDOR_COMPANY_ID    CONFIG_BT_COMPANY_ID_NORDIC
```

samples/bluetooth/nrf_dm/src/main.c
```diff
@@ -316,7 +315,7 @@ static int bt_sync_init(void)
 
 	printk("DM Bluetooth LE Synchronization initialization\n");
 
-	mfg_data.company_code = sys_cpu_to_le16(COMPANY_CODE);
+	mfg_data.company_code = sys_cpu_to_le16(CONFIG_BT_COMPANY_ID);
```
omkar3141, line 318: Isn't this Nordic Specific feature? I think this should be:
```suggestion
	mfg_data.company_code = sys_cpu_to_le16(BT_COMPANY_ID_NORDIC);
```

omkar3141 (changes requested):

samples/bluetooth/nrf_dm/src/main.c
jori-nordic: I don't know mesh that well :p

omkar3141 (approved):

PR #12452, opened 2023-09-27
bl_storage: Fix nv_counters bug with FIH

Vge0rge:
The NRF non-volatile counters function boot_nv_security_counter_get is called by MCUBOOT using the macro FIH_CALL(). When the fault injection protection is enabled this macro increases a counter in MCUBOOT which is expected to be the same before and after the call to the function. The counter is increased by FIH_CALL() and is expected to get decreased by the FIH_RET() macro. This means that all the functions which are called by FIH_CALL() need to return with FIH_RET(), otherwise the fault injection protection will stop the execution.

This change fixes a bug with our implementation of nv_counters which didn't return FIH_RET(). This had as a consequence that when the hardware downgrade prevention and the fault injection protection was enabled in MCUBOOT:
CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION=y
CONFIG_BOOT_FIH_PROFILE_LOW=y

MCUBOOT could not boot anymore.

Ref: NCSDK-23761

### DIFF
--- a/doc/nrf/releases_and_maturity/known_issues.rst
+++ b/doc/nrf/releases_and_maturity/known_issues.rst
@@ -2442,6 +2442,16 @@ Bootloader
 
 The issues in this section are related to :ref:`Bootloaders <app_bootloaders>`.
 
+.. rst-class:: v2-4-0 v2-4-1 v2-4-2
+
+NCSDK-23761: MCUboot fails to boot when both the :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` and :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` Kconfig options are enabled.
+  The MCUboot downgrade prevention mechanism relies on platform-specific implementation of hardware security counters.
+  The |NCS| implementation of the hardware security counters is not compatible with the fault injection hardening code (FIH) of MCUboot.
+  In the |NCS|, the Kconfig option :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` of MCUboot is enabled by default for TF-M builds.
+  You can enable the MCUboot Kconfig option :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` in the |NCS| using the option :kconfig:option:`CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION`.
+
+  **Workaround:** To fix the issue, disable either the :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` or :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` Kconfig option in MCUboot.
+
 .. rst-class:: v2-4-2 v2-4-1 v2-4-0 v2-3-0
 
 SHEL-1352: Incorrect base address used in the OTP TX trim coefficients
--- a/subsys/bootloader/bl_storage/nrf_nv_counters.c
+++ b/subsys/bootloader/bl_storage/nrf_nv_counters.c
@@ -12,6 +12,7 @@
 #include <stdint.h>
 #include <bl_storage.h>
 #include "bootutil/fault_injection_hardening.h"
+#include "bootutil/bootutil_public.h"
 
 fih_int boot_nv_security_counter_init(void)
 {
@@ -24,29 +25,29 @@ fih_int boot_nv_security_counter_get(uint32_t image_id, fih_int *security_cnt)
 	uint16_t cur_sec_cnt;
 
 	if (security_cnt == NULL) {
-		return FIH_FAILURE;
+		FIH_RET(FIH_FAILURE);
 	}
 
 	/* We support only one counter in MCUBOOT */
 	if (image_id > 0) {
-		return FIH_FAILURE;
+		FIH_RET(FIH_FAILURE);
 	}
 
 	err = get_monotonic_counter(BL_MONOTONIC_COUNTERS_DESC_MCUBOOT_ID0, &cur_sec_cnt);
 	if (err != 0) {
-		return FIH_FAILURE;
+		FIH_RET(FIH_FAILURE);
 	}
 
-	*security_cnt = cur_sec_cnt;
+	*security_cnt = fih_int_encode(cur_sec_cnt);
 
-	return FIH_SUCCESS;
+	FIH_RET(FIH_SUCCESS);
 }
 
 int32_t boot_nv_security_counter_update(uint32_t image_id, uint32_t img_security_cnt)
 {
 	/* We only support 16 bits image counters */
 	if ((img_security_cnt & 0xFFFF0000) > 0) {
-		return -1;
+		return -BOOT_EBADARGS;
 	}
 
 	/* MCUBoot permits calling this function with img_security_cnt
@@ -56,16 +57,20 @@ int32_t boot_nv_security_counter_update(uint32_t image_id, uint32_t img_security
 	 */
 	fih_int security_cnt;
 
-	fih_int fih_err = boot_nv_security_counter_get(image_id, &security_cnt);
+	fih_int fih_err;
+	int err;
 
-	if (fih_err != FIH_SUCCESS) {
-		return fih_err;
+	FIH_CALL(boot_nv_security_counter_get, fih_err, image_id, &security_cnt);
+	if (FIH_NOT_EQ(fih_err, FIH_SUCCESS)) {
+		return -BOOT_EBADSTATUS;
 	}
 
-	if (security_cnt == (uint16_t)img_security_cnt) {
-		return FIH_SUCCESS;
+	if (fih_int_decode(security_cnt) == img_security_cnt) {
+		return 0;
 	}
 
-	return set_monotonic_counter(BL_MONOTONIC_COUNTERS_DESC_MCUBOOT_ID0,
-				     (uint16_t)img_security_cnt);
+	err = set_monotonic_counter(BL_MONOTONIC_COUNTERS_DESC_MCUBOOT_ID0,
+				    (uint16_t)img_security_cnt);
+
+	return err == 0 ? 0 : -BOOT_EBADSTATUS;
 }


subsys/bootloader/bl_storage/nrf_nv_counters.c
```diff
@@ -24,22 +24,22 @@ fih_int boot_nv_security_counter_get(uint32_t image_id, fih_int *security_cnt)
 	uint16_t cur_sec_cnt;
 
 	if (security_cnt == NULL) {
-		return FIH_FAILURE;
+		FIH_RET(FIH_FAILURE);
```
hakonfam, line 27: Is this file used for both B0/NSIB and mcuboot?

doc/nrf/releases_and_maturity/known_issues.rst
```diff
@@ -2481,6 +2481,16 @@ Bootloader
 
 The issues in this section are related to :ref:`Bootloaders <app_bootloaders>`.
 
+.. rst-class:: v2-4-0 v2-4-1 v2-4-2
+
+NCSDK-23761: MCUboot fails to boot when both :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` and :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` options are enabled.
+  The MCUboot downgrade prevention mechanism relies on platform specific implementation of hardware security counters.
```
melwee01, line 2487:
```suggestion
  The MCUboot downgrade prevention mechanism relies on platform-specific implementation of hardware security counters.
```

doc/nrf/releases_and_maturity/known_issues.rst
```diff
@@ -2481,6 +2481,16 @@ Bootloader
...
+  The |NCS| implementation of the hardware security counters is not compatible with the fault injection hardening code (FIH) of MCUboot.
+  In |NCS| the :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` of MCUboot is enabled by default for TF-M builds.
```
melwee01, line 2489:
```suggestion
  In the |NCS|, the Kconfig option :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` of MCUboot is enabled by default for TF-M builds.
```

doc/nrf/releases_and_maturity/known_issues.rst
```diff
@@ -2481,6 +2481,16 @@ Bootloader
 
 The issues in this section are related to :ref:`Bootloaders <app_bootloaders>`.
 
+.. rst-class:: v2-4-0 v2-4-1 v2-4-2
+
+NCSDK-23761: MCUboot fails to boot when both :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` and :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` options are enabled.
```
melwee01, line 2486:
```suggestion
NCSDK-23761: MCUboot fails to boot when both the :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` and :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` Kconfig options are enabled.
```

doc/nrf/releases_and_maturity/known_issues.rst
```diff
@@ -2481,6 +2481,16 @@ Bootloader
...
+  The MCUboot downgrade prevention mechanism relies on platform specific implementation of hardware security counters.
+  The |NCS| implementation of the hardware security counters is not compatible with the fault injection hardening code (FIH) of MCUboot.
+  In |NCS| the :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` of MCUboot is enabled by default for TF-M builds.
+  The :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` of MCUboot can be enabled in |NCS| with the option :kconfig:option:`CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION`.
```
melwee01, line 2490:
```suggestion
 You can enable the MCUboot Kconfig option :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION`
 in the |NCS| using the option :kconfig:option:`CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION`.
```

doc/nrf/releases_and_maturity/known_issues.rst
```diff
@@ -2481,6 +2481,16 @@ Bootloader
...
+
+  **Workaround:** To fix the issue, disable :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` or :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` in MCUboot.
```
melwee01, line 2492:
```suggestion
  **Workaround:** To fix the issue, disable either the :kconfig:option:`CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION` or :kconfig:option:`CONFIG_BOOT_FIH_PROFILE_LOW` Kconfig option in MCUboot.
```

subsys/bootloader/bl_storage/nrf_nv_counters.c
Vge0rge: no, this is the implementation used only by MCUboot. B0 has access to the counters through this:
https://github.com/nrfconnect/sdk-nrf/blob/343ba1e878ace99992a299cf58966d536bec5c36/subsys/bootloader/bl_storage/bl_storage.c#L241

melwee01 (approved):

mswarowsky (approved):
Good catch

subsys/bootloader/bl_storage/nrf_nv_counters.c
```diff
@@ -24,22 +24,22 @@ fih_int boot_nv_security_counter_get(uint32_t image_id, fih_int *security_cnt)
...
 	}
 
 	err = get_monotonic_counter(BL_MONOTONIC_COUNTERS_DESC_MCUBOOT_ID0, &cur_sec_cnt);
 	if (err != 0) {
-		return FIH_FAILURE;
+		FIH_RET(FIH_FAILURE);
 	}
 
 	*security_cnt = cur_sec_cnt;
```
vili-nordic, line 40: I think we need to use some fih-macros and functions to encode fih_int here

subsys/bootloader/bl_storage/nrf_nv_counters.c
```diff
@@ -56,14 +56,15 @@ int32_t boot_nv_security_counter_update(uint32_t image_id, uint32_t img_security
...
-	if (fih_err != FIH_SUCCESS) {
-		return fih_err;
+	FIH_CALL(boot_nv_security_counter_get, fih_err, image_id, &security_cnt);
+	if (FIH_NOT_EQ(fih_err, FIH_SUCCESS)) {
+		return -1;
 	}
 
 	if (security_cnt == (uint16_t)img_security_cnt) {
-		return FIH_SUCCESS;
+		return 0;
```
joerchan, line 67: In the realm of mixing error codes from different domains.
From what I can tell this function returns BOOT_E<code> error codes, while the `set_monotonic_counter` function uses zephyr errno error codes.
If the codes carry meaning they can now be changed. 

So I think it would be more clean to also check and convert that error code as well.

subsys/bootloader/bl_storage/nrf_nv_counters.c
```diff
@@ -56,14 +56,15 @@ int32_t boot_nv_security_counter_update(uint32_t image_id, uint32_t img_security
...
+	fih_int fih_err;
 
-	if (fih_err != FIH_SUCCESS) {
-		return fih_err;
+	FIH_CALL(boot_nv_security_counter_get, fih_err, image_id, &security_cnt);
+	if (FIH_NOT_EQ(fih_err, FIH_SUCCESS)) {
+		return -1;
 	}
 
 	if (security_cnt == (uint16_t)img_security_cnt) {
```
vili-nordic, line 66: incorrect handling of fih_int I believe. (fih_int can be compiled to be a struct, so need to use some of their magic fih-macros or functions to compare)

subsys/bootloader/bl_storage/nrf_nv_counters.c
```diff
@@ -56,14 +56,15 @@ int32_t boot_nv_security_counter_update(uint32_t image_id, uint32_t img_security
...
 	fih_int security_cnt;
 
-	fih_int fih_err = boot_nv_security_counter_get(image_id, &security_cnt);
+	fih_int fih_err;
 
-	if (fih_err != FIH_SUCCESS) {
-		return fih_err;
+	FIH_CALL(boot_nv_security_counter_get, fih_err, image_id, &security_cnt);
+	if (FIH_NOT_EQ(fih_err, FIH_SUCCESS)) {
+		return -1;
```
joerchan, line 63: Appears to be BOOT_EFLASH, not sure if this is the right one.

subsys/bootloader/bl_storage/nrf_nv_counters.c
vili-nordic: Probably only issue in HIGH_PROFILE, not sure if we support or that, or want to have code that's broken for that case anyways

subsys/bootloader/bl_storage/nrf_nv_counters.c
Vge0rge: You are right, I will update this

subsys/bootloader/bl_storage/nrf_nv_counters.c
Vge0rge: How can you tell that? Could you explain please? I only see that it says "0 on success; nonzero on failure." in the signagure.

subsys/bootloader/bl_storage/nrf_nv_counters.c
Vge0rge: I see no  reason to not use the macros I think. Even if we don't  support it (so we don't test it) it's a better practice to use the macros that they provide to us I think.

vili-nordic (approved):
lgtm

subsys/bootloader/bl_storage/nrf_nv_counters.c
joerchan: In the bootloader code we usually return BOOT_E<code> error code, while the function that you're calling here returns E<code> error codes.
One returns 'int' the other 'uint32_t', so there is also that.
So the value that they return are in different domains, meaning that the code cannot be correctly interpreted, other than 0 being success.

joerchan (approved):

PR #12454, opened 2023-09-27
samples: peripheral_uart: Fix hard-coded timeout value

pideu-mh:
Calls to uart_rx_enable() require a timeout argument, for which UART_WAIT_FOR_RX is used in 2 of 3 calls in this application.

The initial init call, however, uses a hardcoded value of 50 (milliseconds) instead of UART_WAIT_FOR_RX which is clearly an oversight.

### DIFF
--- a/samples/bluetooth/peripheral_uart/src/main.c
+++ b/samples/bluetooth/peripheral_uart/src/main.c
@@ -322,7 +322,7 @@ static int uart_init(void)
 		return err;
 	}
 
-	err = uart_rx_enable(uart, rx->data, sizeof(rx->data), 50);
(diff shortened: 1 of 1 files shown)


KAGA164:
Thanks for your contribution @pideu-mh. I check the reason why CI failed:

 6: UC4 Commit message body line exceeds max length (75>72): "The initial init call, however, uses a hardcoded value of 50 (milliseconds)"
Error: Process completed with exit code 1.

Could you format your commit message according to above requirement?

KAGA164:
One more thing. The PR target is wrong. This should go to ncs main branch. @pideu-mh Could you change PR target?

carlescufi:
@pideu-mh can you please retarget this PR against the `main` branch and then we'll take care of backporting to v2.4?

PR #12471, opened 2023-09-29
Sysbuild additions B

nordicjm:


nordicjm:
> Don't know what's going on. I only removed myself and @VTPeltoketo, but actually many got removed. And when I try to add them back manually nothing happens.

Github supports a maximum of 15 reviewers per PR, though using the API (i.e. bots) can actually add more, but from a web browser the 15 limit applies

glarsennordic:
When are we planning to remove child-image boardfiles such as `nrf/samples/cellular/nrf_cloud_multi_service/child_image/*`?

nordicjm:
> When are we planning to remove child-image boardfiles such as `nrf/samples/cellular/nrf_cloud_multi_service/child_image/*`?

Not for a few releases

SebastianBoe (approved):

pdunaj:
FYI @nvlsianpu , @kapi-no

samples/cellular/nrf_cloud_multi_service/sysbuild/mcuboot/boards/nrf9160dk_nrf9160_0_14_0.overlay
```diff
@@ -0,0 +1,16 @@
+/*
```
plskeggs, line 1: Note that the nrf_cloud_multi_service sample now has an overlay in `child_image/mcuboot/boards/nrf9161dk_nrf9161_ns.overlay` -- perhaps you need one here too?

plskeggs (approved):

samples/cellular/nrf_cloud_multi_service/sysbuild/mcuboot/boards/nrf9160dk_nrf9160_0_14_0.overlay
nordicjm: 9161 overlays are not being added for now

tejlmand (approved):

samples/cellular/nrf_cloud_multi_service/sysbuild/mcuboot/prj.conf
glarsennordic, line 1: This file doesn't match what's in child_img/mcuboot/nrf9160dk_nrf9160.conf

For instance, adding FPROTECT, and disabling SPI

Is that expected? Can you expound upon it a bit?

nordicjm: Overlays are no longer supported, sysbuild works by changing the application configuration directory which means the desired prj.conf (which includes the additional configurations from the overlay) must be placed there

kapi-no:
FYI:
@alstrzebonski (Fast Pair sample)
@MarekPieta (nRF Desktop app)
(I cannot add you to the reviewer's list as it is saturated)

applications/nrf_desktop/sysbuild/mcuboot/prj.conf
kapi-no, line 1: nRF Desktop has multiple configuration targets, some of which use unique MCUboot configurations.
For instance, see the following configuration:
``nrf/applications/nrf_desktop/configuration/nrf52840gmouse_nrf52840/child_image/mcuboot/prj_fast_pair.conf``
that enables the Direct-XIP for nRF Desktop configuration with Fast Pair support (nRF52840 Gaming mouse target)

I believe we need to cover all configuration variants if we want to support sysbuild simultaneously with the legacy build system solution.

This comment also applies to other child images (HCI RPMsg, B0 bootloader)

applications/nrf_desktop/sysbuild/mcuboot/app.overlay
kapi-no, line 1: We define partition YAML files for each target configuration variant in the legacy build system. I think we need to port the configuration for sysbuild that preserves the partition location

kapi-no (commented):
Leaving some initial comments/thoughts regarding the nRF Desktop changes

applications/nrf_desktop/sysbuild/mcuboot/prj.conf
nordicjm: This is not supported, there is a general image configuration folder e.g. `sysbuild/mcuboot` which will apply to any mcuboot image configured as part of that image, it does not support having different configuration files on a per-board basis, though a folder can be created the sample.yml can be updated to manually specify the file

applications/nrf_desktop/sysbuild/mcuboot/app.overlay
nordicjm: The default pm_static.yml for a board in the configuration folder will work, to use a different one you just need to set PM_STATIC_YML_FILE to the file

kapi-no: I see. This part seems to work the same as in the case of legacy build system

samples/cellular/nrf_cloud_multi_service/sysbuild/mcuboot/prj.conf
glarsennordic: Ah, got it, I understand now, thanks. LGTM.

glarsennordic (approved):

applications/nrf_desktop/sysbuild/mcuboot/app.overlay
nordicjm: It is present: https://github.com/nrfconnect/sdk-mcuboot/blob/main/boot/zephyr/app.overlay

nordicjm:
@tomchy @Vge0rge @joerchan @mswarowsky @edmont @jayteemo @lemrey @ludvigsj @SeppoTakalo @nordic-auko @coderbyheart @rakons @57300 please take a look

applications/nrf_desktop/sysbuild/mcuboot/prj.conf
kapi-no: I think we need to provide such a feature before we remove support for the legacy subsystem, as it diminishes the UX when working with nRF Desktop application in NCS. As I understand it right now, even if we define child image configurations for each target, then the user needs to manually specify the child image configuration for the chosen target in the build command

nordicjm: No, if there is a folder in the sysbuild folder, it will be used automatically, no additional parameters are needed when configuring an application

carlescufi:
@kapi-no @MarekPieta could you please review?
@grochu can you please review the Bluetooth bits?

muraliThokala (approved):

krish2718:
For Wi-Fi CI failure please rebase the fix has been merged to NCS already.

kapi-no (approved):

samples/bluetooth/direct_test_mode/sysbuild/remote_shell/boards/nrf5340dk_nrf5340_cpuapp.overlay
```diff
@@ -0,0 +1,26 @@
+/*
```
KAGA164, line 1: It looks like image for application core is not built when using sysbuild

samples/bluetooth/peripheral_mds/sysbuild/hci_rpmsg/prj.conf
```diff
@@ -0,0 +1,30 @@
...
+
+CONFIG_HEAP_MEM_POOL_SIZE=8192
+
+CONFIG_MAIN_STACK_SIZE=512
+CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=512
+
+CONFIG_BT=y
+CONFIG_BT_HCI_RAW=y
+CONFIG_BT_HCI_RAW_RESERVE=1
+#CONFIG_BT_MAX_CONN=16
```
KAGA164, line 12: Why is it commented here?

KAGA164: I think this line can be removed or set to desired value 2 instead of having line 30

samples/bluetooth/throughput/sysbuild/hci_rpmsg/prj.conf
```diff
@@ -0,0 +1,32 @@
+CONFIG_IPC_SERVICE=y
```
KAGA164, line 1: Wouldn't be better to create just configuration overlay file instead of introducing full configuration file in every sample that needs to modify other core image?

samples/bluetooth/direct_test_mode/sysbuild/remote_shell/boards/nrf5340dk_nrf5340_cpuapp.overlay
nordicjm: Yes, this does not work NCSDK-23805

samples/bluetooth/peripheral_mds/sysbuild/hci_rpmsg/prj.conf
nordicjm: It is commented as that part of the file is the copy from the upstream file, they can probably be merged better at some point

samples/bluetooth/throughput/sysbuild/hci_rpmsg/prj.conf
nordicjm: A feature of sysbuild is not having the underlying image silently change your configuration which is something that happens with child/parent image

KAGA164: Seems like there is option to overlay Konfig fragment using sysbuild:

> In the folder of the main application, create a Kconfig fragment or a devicetree overlay under a sysbuild folder, where the name of the file is <image>.conf or <image>.overlay, for example if your main application includes my_sample then create a sysbuild/my_sample.conf file or a devicetree overlay sysbuild/my_sample.overlay.

Having multiple copy of the same file will be a nightmare in maintenance.

nordicjm: This is not supported when using board overlay, hence why the files are copied because the APPLICATION_CONFIG_DIR becomes that `sysbuild/app` folder

> Having multiple copy of the same file will be a nightmare in maintenance.

Having configurations change without the user being aware is a nightmare in updating modules. But yes, maintenance burden will be much higher

MarkusLassila (approved):

canisLupus1313 (approved):

samples/wifi/sr_coex/sysbuild/hci_rpmsg/prj.conf
```diff
@@ -0,0 +1,32 @@
+CONFIG_IPC_SERVICE=y
```
tomchy, line 1: Missing license headers?

nordicjm: These files originate from zephyr, so yes the headers do not apply

PR #12494, opened 2023-10-03
samples: wifi: sr_coex: Wi-Fi and Bluetooth LE throughput tests

muraliThokala:
Support for all combinations of Wi-Fi throughput and Bluetooth LE throughput. That is, support for Wi-Fi UDP and TCP transport protocols in client and server roles, Bluetooth LE in central and peripheral roles.


krish2718 (commented):
Can you please divide this in to smaller commits? Its a big PR across modules to review.

samples/wifi/sr_coex/test/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+#
```
krish2718, line 1: Please move all tests to the `tests` directory, samples are for demonstration not for testing, see https://docs.zephyrproject.org/latest/samples/sample_definition_and_criteria.html#samples-must-not-be-used-as-a-testcase.

And then convert each `overlay-wifi*` to a test case, I would also recommend removing the test code from the sample and convert from compile time tests to run time tests (no changes to the configuration).

samples/wifi/sr_coex/prj.conf
```diff
@@ -35,11 +35,11 @@ CONFIG_NET_L2_ETHERNET=y
 CONFIG_NET_CONFIG_SETTINGS=y
 CONFIG_NET_CONFIG_INIT_TIMEOUT=0
 
-CONFIG_NET_SOCKETS_POLL_MAX=6
+#CONFIG_NET_SOCKETS_POLL_MAX=6
 
 # Memories
 CONFIG_MAIN_STACK_SIZE=4096
-CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=2048
+#CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=2048
```
krish2718, line 42: Please remove commented lines

drivers/wifi/nrf700x/zephyr/src/zephyr_coex.c
```diff
@@ -182,9 +307,81 @@ int nrf_wifi_coex_config_pta(enum nrf_wifi_pta_wlan_op_band wlan_band, bool sepa
...
+		params.configbuf[index] = (start_offset << COEX_REG_CFG_OFFSET_SHIFT) |
+			(*(config_buffer_ptr+index));
+		start_offset++;
+	}
+
+	cmd_len = (COEX_CONFIG_FIXED_PARAMS + num_reg_to_config) * sizeof(uint32_t);
+
+	status = nrf_wifi_fmac_conf_btcoex(rpu_ctx->rpu_ctx,
+					   (void *)(&params), cmd_len);
+
```
krish2718, line 360: no newlines before error checks, its the same logical group.

drivers/wifi/nrf700x/zephyr/inc/zephyr_coex.h
```diff
@@ -23,49 +23,54 @@ enum nrf_wifi_pta_wlan_op_band {
...
 /**
  * @function   nrf_wifi_coex_config_pta(enum nrf_wifi_pta_wlan_op_band wlan_band,
- *             bool separate_antennas)
+ *             bool antenna_mode, bool ble_role, bool wlan_role)
  *
  * @brief      Function used to configure PTA tables of coexistence hardware.
  *
  * @param[in]  enum nrf_wifi_pta_wlan_op_band wlan_band
- * @param[in]  separate_antennas
+ * @param[in]  antenna_mode
```
krish2718, line 31: Using a boolean type and calling it `antenna_mode` doesn't make sense, `mode` implies a `enum`, IMHO, original naming is better for a bool unless you want to make it an `enum`.

drivers/wifi/nrf700x/zephyr/inc/zephyr_coex_struct.h
```diff
@@ -16,66 +16,108 @@
...
 	(NUM_ELEMENTS_EXCL_CONFIGBUF*sizeof(uint32_t)))>>2)
-/* Number of elements in coex_sr_traffic_info other than sr_traffic_info[] */
+/* Number of elements in nrf_wifi_coex_sr_traffic_info other than sr_traffic_info[] */
 #define NUM_ELEMENTS_EXCL_SRINFOBUF 1
 /* Each SR Traffic Info is of type uint32_t */
 #define MAX_SR_TRAFFIC_BUF_SIZE 32
 
+/* Indicates the type of priority window i.e., WLAN or SR */
+enum {
+	NRF_WIFI_WIFI_WINDOW = 0,
```
krish2718, line 31: IMHO, if you are not expecting more things to be added here, you can just convert them to boolean and call `is_X`. And even if you think more will be added, these shouldn't be anonymous enums and should aptly describe the grouping.

Same for others.

krish2718 (changes requested):

PR #12502, opened 2023-10-03
openthread: kconfig: Add experimental PSA Crypto support

maciejbaczmanski:
Introduce to PSA Crypto API in OpenThread.

- Enabled PSA internal storage
- Enabled OPENTHREAD_CRYPTO_PSA
- Enabled all PSA_WANT configs which are required by OpenThread
- Enabled configs needed for DTLS/TLS functionality
- Aligned defines needed for OT DTLS functionality
- Fixed unused variable during `otPlatCryptoInit`


subsys/net/openthread/Kconfig
```diff
@@ -19,12 +19,47 @@ config NORDIC_SECURITY_PROMPTLESS
...
 endchoice
 
+config OPENTHREAD_NRF_SECURITY_PSA
+	bool
+	select EXPERIMENTAL
+	select NRF_SECURITY
+	select MBEDTLS_ENABLE_HEAP
+	select MBEDTLS_PSA_CRYPTO_STORAGE_C if !BUILD_WITH_TFM
+	select PSA_NATIVE_ITS if !BUILD_WITH_TFM
+	select PSA_WANT_ALG_RSA_OAEP
```
Damian-Nordic, line 39: Does OpenThread use RSA? I noticed that flash increase can be reduced fro 31kB to 14kB if we do not enable RSA :)

subsys/net/openthread/Kconfig
```diff
@@ -19,12 +19,47 @@ config NORDIC_SECURITY_PROMPTLESS
...
+	select PSA_WANT_ALG_HKDF
+	select PSA_WANT_ALG_CCM
```
Damian-Nordic, line 41: Is CCM needed? Note that OpenThread implements CCM itself and only AES_ECB_NO_PADDING is required from the crypto layer. HKDF is also implemented in OpenThread with HMAC.

subsys/net/openthread/Kconfig
```diff
@@ -19,12 +19,47 @@ config NORDIC_SECURITY_PROMPTLESS
...
+	select PSA_WANT_ALG_CMAC
+	select PSA_WANT_ALG_HMAC
+	select PSA_WANT_ALG_ECB_NO_PADDING
+	select PSA_WANT_ALG_SHA_256
+	select PSA_WANT_ALG_JPAKE
```
Damian-Nordic, line 46: This may be a candidate for a followup task, but it may be helpful to enable TLS/JPAKE bits only if the commissioner/joiner is enabled to save flash on devices that don't need it.

subsys/net/openthread/Kconfig.defconfig
```diff
@@ -82,6 +82,9 @@ choice OPENTHREAD_SECURITY
 	default OPENTHREAD_NRF_SECURITY_CHOICE if SOC_FAMILY_NRF
 endchoice
 
+config MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG
+	default n if OPENTHREAD_NRF_SECURITY_PSA
```
Damian-Nordic, line 86: Could you explain why this is needed?

Damian-Nordic (commented):
Some comments that may help optimize the flash usage, but note that I'm only not super-familiar with Thread commissioning (especially, commercial commissioning), so maybe some algorithms that I mentioned are needed somewhere indeed :)

subsys/net/openthread/Kconfig
maciejbaczmanski: Yes, we've discussed that earlier and decided to proceed with this approach for the experimental support as there are some dependencies in openthread that require ecjpake support whenever DTLS is used (so we cannot simply make it dependent only on commissioner/joiner, because COAPS uses DTLS too). We will work on these dependencies as next step.

subsys/net/openthread/Kconfig
maciejbaczmanski: These configurations are mainly based on sample provided by crypto team:
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/crypto/psa_tls/prj.conf
Also from config in openthread:
https://github.com/openthread/openthread/blob/main/third_party/mbedtls/mbedtls-config.h

In both files CCM support is defined. Also PSA TLS sample enables HKDF. 
@frkv could you confirm if configurations here are correct and needed?

subsys/net/openthread/Kconfig
maciejbaczmanski: I cannot see the RSA define in OpenThread config: https://github.com/openthread/openthread/blob/main/third_party/mbedtls/mbedtls-config.h

@ArekBalysNordic do you think we need RSA support to be enabled here?

subsys/net/openthread/Kconfig
Damian-Nordic: Sample from the crypto team doesn't really tell us anything about OT requirements, but the reference mbedTLS config indeed defines MBEDTLS_TLS_ECJPAKE_WITH_AES_128_CCM_8, so looks like AES CCM may be required for TLS, so let's keep it :).

subsys/net/openthread/Kconfig
ArekBalysNordic: No, we don't need it if OpenThread doesn't use it. Maybe it was a redundant config in my PR, I didn't optimize it recently, just ran it. :)

subsys/net/openthread/Kconfig.defconfig
ArekBalysNordic: As far as I remember, I added it to my recent PR regarding PSA crypto, because I wanted to use the internal implementation of the RNG when the PSA api is used, but I don't remember why it wasn't setting automatically... Maybe now we can remove those lines.

subsys/net/openthread/Kconfig.defconfig
```diff
@@ -152,7 +155,8 @@ endif # !(OPENTHREAD_BORDER_AGENT || OPENTHREAD_COMMISSIONER || OPENTHREAD_COAPS
 
 config MBEDTLS_HEAP_SIZE
 	int
-	default 12440
+	default 15360 if OPENTHREAD_NRF_SECURITY_PSA
```
ArekBalysNordic, line 158: Note here, that is an experimental value. I didn't do any reliable measurements, but for 12440 it didn't work for me.  We can do some measurements using watermarks, or at least add a note here that the number may be possibly reduced in the future.

maciejbaczmanski: added a note. We'll work on that as next step

subsys/net/openthread/Kconfig
maciejbaczmanski: I've removed this part and tested locally- commissioning/joining and COAPS look good. I'll schedule full nightly suite with PSA enabled too.

subsys/net/openthread/Kconfig.defconfig
maciejbaczmanski: I've removed this part and tested locally- commissioning/joining and COAPS look good. I'll schedule full nightly suite with PSA enabled too.

maciejbaczmanski: I've checked build files once more and seems like external RNG is enabled by default (here: 
https://github.com/nrfconnect/sdk-nrf/blob/3003f42ee1a14ac6ca6f3cb6990ea2fcc05bb6a7/subsys/nrf_security/Kconfig.psa#L22 )
Do we want to keep it that way then? Or should we switch to internal RNG as in your recent PR?

Damian-Nordic: @joerchan What is the recommendation of the crypto team with regards to MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG when PSA crypto API is used? I don't quite understand what it changes under the hood.

Damian-Nordic (approved):

subsys/net/openthread/Kconfig
```diff
@@ -19,12 +19,46 @@ config NORDIC_SECURITY_PROMPTLESS
...
+	select PSA_WANT_ALG_TLS12_ECJPAKE_TO_PMS
+	select PSA_WANT_ALG_DETERMINISTIC_ECDSA if OPENTHREAD_ECDSA
+	select PSA_WANT_ALG_ECDSA if OPENTHREAD_ECDSA
+	select PSA_WANT_ALG_ECDH if OPENTHREAD_ECDSA
+	select PSA_WANT_ECC_SECP_R1_256
+	select PSA_WANT_GENERATE_RANDOM
+	select PSA_WANT_KEY_TYPE_AES
+	select PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY
```
joerchan, line 54: Selecting this is optional, KEY_PAIR will include PUBLIC_KEY on its own.

subsys/net/openthread/Kconfig
```diff
@@ -19,12 +19,46 @@ config NORDIC_SECURITY_PROMPTLESS
...
+	select PSA_WANT_KEY_TYPE_ECC_KEY_PAIR
```
joerchan, line 55: Is the ECC curve and KEY only needed with OPENTHREAD_ECDSA enabled maybe?

subsys/net/openthread/Kconfig.defconfig
joerchan: @Damian-Nordic This is an old configuration, from when we had MBEDTLS as the PSA Core.

It used to be a setting that would direct the RNG to an external source i.e. cc3xx. This is now replaced by regular PSA crypto drivers.

We have removed this configuration, so please remove this here as well.

joerchan (commented):
Remove MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG .

subsys/net/openthread/Kconfig.defconfig
Damian-Nordic: Thanks for the clarification!

subsys/net/openthread/Kconfig
```diff
@@ -19,12 +19,46 @@ config NORDIC_SECURITY_PROMPTLESS
...
+	select PSA_NATIVE_ITS if !BUILD_WITH_TFM
+	select PSA_WANT_ALG_HKDF
+	select PSA_WANT_ALG_CCM
+	select PSA_WANT_ALG_CMAC
+	select PSA_WANT_ALG_HMAC
+	select PSA_WANT_ALG_ECB_NO_PADDING
+	select PSA_WANT_ALG_SHA_256
+	select PSA_WANT_ALG_JPAKE
+	select PSA_WANT_ALG_TLS12_ECJPAKE_TO_PMS
+	select PSA_WANT_ALG_DETERMINISTIC_ECDSA if OPENTHREAD_ECDSA
```
joerchan, line 48: You could also do it like this:

```
config OPENTHREAD_NRF_SECURITY_PSA
    bool
    select PSA_<basic_features>

config OPENTHREAD_NRF_SECURITY_PSA
     select PSA_WANT_<ecdsa_features>
     depends on OPENTHREAD_ECDSA
```

subsys/net/openthread/Kconfig
maciejbaczmanski: If I remove direct selection of `PSA_WANT_KEY_TYPE_ECC_PUBLIC_KEY` and make `PSA_WANT_KEY_TYPE_ECC_KEY_PAIR` dependend on `OPENTHREAD_ECDSA`, I'd need to make the same dependency for `PSA_WANT_ECC_SECP_R1_256` . Then I'll get build errors:
```
   77 | #error "MBEDTLS_DHM_C defined, but not all prerequisites"
      |  ^~~~~
/home/mbaczmanski/ncs/nrf/ext/oberon/psa/core/include/mbedtls/check_config.h:148:2: error: #error "MBEDTLS_ECP_C defined, but not all prerequisites"
  148 | #error "MBEDTLS_ECP_C defined, but not all prerequisites"
      |  ^~~~~
/home/mbaczmanski/ncs/nrf/ext/oberon/psa/core/include/mbedtls/check_config.h:327:2: error: #error "MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED defined, but not all prerequisites"
  327 | #error "MBEDTLS_KEY_EXCHANGE_ECJPAKE_ENABLED defined, but not all prerequisites"
      |  ^~~~~
/home/mbaczmanski/ncs/nrf/ext/oberon/psa/core/include/mbedtls/check_config.h:989:2: error: #error "MBEDTLS_X509_USE_C defined, but not all prerequisites"
  989 | #error "MBEDTLS_X509_USE_C defined, but not all prerequisites"
```
which would force me to disable joiner and coaps. We have added a note to next step of PSA crypto API support to take a deeper look into all dependencies. OPENTHREAD_ECDSA is onef them as it isn't directly correlated with commissioner/joiner/COAPS configs, but joiner is eabled in all library configurations for now, so it needs to be always selected. I think we could keep this one the way it is for experimental support and optimize it later?

subsys/net/openthread/Kconfig
maciejbaczmanski: I agree, but as above- the dependency between OPENTHREAD_ECDSA and functionalities like Joiner which requires ECJPAKE are not directly dependent now. I think we should fix that first and then optimize this part accordingly?

subsys/net/openthread/Kconfig.defconfig
maciejbaczmanski: removed, thanks

subsys/net/openthread/Kconfig
maciejbaczmanski: removed

Damian-Nordic (approved):

subsys/net/openthread/Kconfig
joerchan: Ok, if you get check_config errors then lets go with what works :) It can be a future improvement.

subsys/net/openthread/Kconfig
joerchan: Fair enough, it's only a suggestion anyway.

joerchan (approved):

west.yml
```diff
@@ -61,7 +61,7 @@ manifest:
     # https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/modules.html
     - name: zephyr
       repo-path: sdk-zephyr
-      revision: 6b30e02121a9852c98ba0c7600e041da24b2c1fd
+      revision: pull/1320/head
```
rlubos, line 64: `9f513ec4ec1520912f65e7d7d938e373b5e04bcd`

rlubos (approved):

PR #12506, opened 2023-10-03
Bluetooth: Mesh: Fix dim to dark behavior

ludvigsj:
Fixes issues with Move Set and Delta Set handlers not
handling dimming starting from 0 correctly.

 * Clamp targets to 0 when dimming down from 0, instead of clamping to
   RANGE_MIN, to avoid a message requesting a dim down causing the light
   to actually turn on.
 * Take the gap between 0 and RANGE_MIN into account when computing the
   transition time for a move set before this is sent to the application
   callback, to avoid a slow transition for dims starting at 0
 * Update the samples and tester to correctly transition the
   lightness state when dimming up from zero when a RANGE_MIN is set.
 * Update the documentation for the callback to mention this
   requirement.
 * Small bug fixes
    * Range converted to actual in delta set to match with the delta and
      target_actual
    * Do not compute transition in move set if target == current

### DIFF
--- a/include/bluetooth/mesh/lightness_srv.h
+++ b/include/bluetooth/mesh/lightness_srv.h
@@ -75,6 +75,10 @@ struct bt_mesh_lightness_srv_handlers {
 	 * @ref bt_mesh_model_transition_time returns a nonzero value, the application is
 	 * responsible for publishing a value of the Light state at the end of the transition.
 	 *
+	 * When performing a non-instantaneous state change, the Light state at any intermediate
+	 * point must be clamped to the current lightness range of the server using
+	 * @ref bt_mesh_lightness_clamp.
+	 *
 	 * @note This handler is mandatory.
 	 *
 	 * @param[in] srv Server to set the Light state of.
@@ -89,6 +93,10 @@ struct bt_mesh_lightness_srv_handlers {
 				struct bt_mesh_lightness_status *rsp);
 
 	/** @brief Get the current Light state.
+	 *
+	 * When in the middle of a non-instantaneous state change, the Light value filled in
+	 * @c rsp must be clamped to the current lightness range of the server using
+	 * @ref bt_mesh_lightness_clamp.
 	 *
 	 * @note This handler is mandatory.
 	 *
@@ -189,6 +197,18 @@ struct bt_mesh_lightness_srv {
 	struct bt_mesh_light_ctrl_srv *ctrl;
 #endif
 };
+
+/** @brief Clamp lightness to lightness range.
+ *
+ *  @return @c lightness clamped to the lightness range set on the
+ *  @ref bt_mesh_lightness_srv pointed to by @c srv if @c lightness > 0, zero
+ *  otherwise.
+ */
+static inline uint16_t bt_mesh_lightness_clamp(const struct bt_mesh_lightness_srv *srv,
+					       uint16_t lightness)
+{
+	return lightness == 0 ? 0 : CLAMP(lightness, srv->range.min, srv->range.max);
+}
 
 /** @brief Publish the current Light state.
  *
--- a/subsys/bluetooth/mesh/lightness_srv.c
+++ b/subsys/bluetooth/mesh/lightness_srv.c
@@ -588,11 +588,18 @@ static void lvl_delta_set(struct bt_mesh_lvl_srv *lvl_srv,
 		srv->delta_start = to_actual(status.current);
 	}
 
-	/* Clamp the value to the lightness range to prevent it be moved back to zero in binding
-	 * with Generic Level state.
-	 */
-	target_actual = CLAMP(srv->delta_start + delta_set->delta,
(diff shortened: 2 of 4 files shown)


subsys/bluetooth/mesh/lightness_srv.c
```diff
@@ -588,11 +588,18 @@ static void lvl_delta_set(struct bt_mesh_lvl_srv *lvl_srv,
 		srv->delta_start = to_actual(status.current);
 	}
 
-	/* Clamp the value to the lightness range to prevent it be moved back to zero in binding
-	 * with Generic Level state.
-	 */
-	target_actual = CLAMP(srv->delta_start + delta_set->delta,
-			      srv->range.min, srv->range.max);
+	if (srv->delta_start == 0 && delta_set->delta < 0) {
```
PavelVPV, line 591: Shouldn't it be `<=`? If `delta_start` and `delta` are both zero, this will set lightness to range min.

tests/bluetooth/tester/src/model_handler.c
```diff
@@ -440,6 +440,10 @@ static void light_set(struct bt_mesh_lightness_srv *srv,
 		l_ctx->remaining = 0;
 	}
 
+	if (l_ctx->target > 0 && l_ctx->current < l_ctx->srv.range.min) {
+		l_ctx->current = l_ctx->srv.range.min;
+	}
```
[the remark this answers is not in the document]
ludvigsj: Something like `bt_mesh_lightness_get_start_value(current, target)`?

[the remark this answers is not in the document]
PavelVPV: + server pointer to get range min.

PavelVPV (approved):

include/bluetooth/mesh/lightness_srv.h
```diff
@@ -75,6 +75,11 @@ struct bt_mesh_lightness_srv_handlers {
 	 * @ref bt_mesh_model_transition_time returns a nonzero value, the application is
 	 * responsible for publishing a value of the Light state at the end of the transition.
 	 *
+	 * When performing a non-instantaneous state change, the applicaton must call
+	 * @ref bt_mesh_lightness_start_value. The change must move instantaneously to the start
+	 * value returned by this function and start gradually changing from that point over the
+	 * time specified by the transition.
```
mia-ko, line 81:
```suggestion
	 * When performing a non-instantaneous state change, the applicaton must call
	 * @ref bt_mesh_lightness_start_value. The change must move instantaneously
	 * to the start value returned by this function, and then gradually change from
	 * that value over time specified by the transition.
```

include/bluetooth/mesh/lightness_srv.h
```diff
@@ -190,6 +195,20 @@ struct bt_mesh_lightness_srv {
 #endif
 };
 
+/** @brief Get the starting point for a non-instantaneous change in lightness.
+ *
+ * When performing a non-instantaneous change of the lightness state, the change
+ * should immediately jump to this starting value and then start gradually changing.
```
mia-ko, line 201:
```suggestion
 * should immediately jump to this starting value and then gradually start changing.
```

subsys/bluetooth/mesh/lightness_srv.c
```diff
@@ -588,11 +588,18 @@ static void lvl_delta_set(struct bt_mesh_lvl_srv *lvl_srv,
...
 	}
 
-	/* Clamp the value to the lightness range to prevent it be moved back to zero in binding
-	 * with Generic Level state.
-	 */
-	target_actual = CLAMP(srv->delta_start + delta_set->delta,
-			      srv->range.min, srv->range.max);
+	if (srv->delta_start == 0 && delta_set->delta <= 0) {
+		/* Do not clamp to range when dimming down from zero, to avoid
+		 * a dim down to cause the light to turn on.
```
mia-ko, line 593:
```suggestion
		/* Do not clamp to range when dimming down from zero to avoid
		 * the light to turn on when dimmed down.
```

subsys/bluetooth/mesh/lightness_srv.c
```diff
@@ -588,11 +588,18 @@ static void lvl_delta_set(struct bt_mesh_lvl_srv *lvl_srv,
...
+		 */
+		target_actual = 0;
+	} else {
+		/* Clamp the value to the lightness range to prevent it be moved
+		 * back to zero in binding with Generic Level state.
```
mia-ko, line 598:
```suggestion
		/* Clamp the value to the lightness range to prevent that it moves
		 * back to zero in binding with Generic Level state.
```

ludvigsj:
There is an ongoing discussion for specs about the expected behavior here, will leave this PR DNM until discussion about specs is done

mia-ko (changes requested):

mia-ko (approved):

PavelVPV (approved):

PR #12520, opened 2023-10-04
samples: cellular: modem_shell: ppp: restarting logic changed

jhirsi:
After Zephyr upmerge PPP changes, there is no ppp carrier off event anymore when lcp is disconnected, 
i.e. when ppp disconnect is initiated from PC. Thus, MOSH started to use ppp dead/running events instead to trigger the need for PPP restart.
Jira: MOSH-475

### DIFF
--- a/samples/cellular/modem_shell/src/ppp/ppp_ctrl.c
+++ b/samples/cellular/modem_shell/src/ppp/ppp_ctrl.c
@@ -65,7 +65,8 @@ static struct k_work ppp_ctrl_restart_work;
 
 static bool ppp_up;
 
-static volatile bool ppp_closing;
+static volatile bool ppp_restarting;
+static volatile bool ppp_prot_running;
 
 /******************************************************************************/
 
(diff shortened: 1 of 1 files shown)


moksanen (approved):
This fixes PPP in modem_shell. modem_shell CI tests for PPP are passing after this change.

trantanen (approved):

samples/cellular/modem_shell/src/ppp/ppp_ctrl.c
```diff
@@ -173,26 +205,33 @@ static void ppp_ctrl_net_mgmt_event_handler(struct net_mgmt_event_callback *cb,
...
 		mosh_print("PPP carrier ON");
 		return;
 	}
-
 	if (mgmt_event == NET_EVENT_PPP_CARRIER_OFF) {
 		mosh_print("PPP carrier OFF");
+		return;
+	}
 
+	if (mgmt_event == NET_EVENT_PPP_PHASE_DEAD && ppp_prot_running) {
```
tokangas, line 226: This function could maybe be refactored as a switch-case. Or at least this should be moved inside the `if` on line 213, which already checks if the event is `NET_EVENT_PPP_PHASE_DEAD`.

jhirsi: Added a switch-case.

tokangas (approved):

PR #12523, opened 2023-10-04
cmake: create store / restore functions for NCS prefixed variables

tejlmand:
Introduce two new functions in NCS sysbuild code:
- store_ncs_vars
- restore_ncs_vars

the store_ncs_vars() pushes all CMake variables prefixed with NCS_` which exists in the top-level sdk-nrf sysbuild CMake file to a property and upon calls to sysbuild hooks, those NCS variables are restored.

This allows CMake code in sdk-nrf to share CMake variables between sysbuild configure stages.

### DIFF
--- a/sysbuild/CMakeLists.txt
+++ b/sysbuild/CMakeLists.txt
@@ -4,12 +4,30 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
+function(store_ncs_vars)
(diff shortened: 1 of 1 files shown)


nordicjm (approved):
Fixes secure boot, many thanks!

57300 (approved):
Lukewarm +1. Let's see how this will be applied 👍

PR #12525, opened 2023-10-04
samples: matter: remove diagnostic clusters from template ZAP

ArekBalysNordic:
We do not need to have Thread, Wi-Fi, and Software diagnostic clusters in the template sample, so we can remove it from the ZAP file.


doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -509,6 +509,10 @@ Matter samples
 
   * Added the :ref:`Matter thermostat <matter_thermostat_sample>` sample.
 
+* :ref:`matter_template_sample`:
+
+  * Removed Thread, Wi-Fi, and software diagnostic clusters support from the ZAP file.
```
wiba-nordic, line 514:
```suggestion
  * Removed support for the Thread, Wi-Fi, and software diagnostics clusters from the ZAP file.
```

wiba-nordic (changes requested):

samples/matter/template/src/template.zap
```diff
@@ -1,5 +1,5 @@
 {
-  "featureLevel": 96,
+  "featureLevel": 98,
```
Damian-Nordic, line 2: Make sure that you have used the ZAP Tool version that we currently recommend. Otherwise, people will struggle to open this file with the recommended version (I did that in the past ;)).

Damian-Nordic (approved):
Approving but the feature version bump is concerning if we do not bump the recommended ZAP version.

samples/matter/template/src/template.zap
ArekBalysNordic: Oh, right, I have two ZAP Tools installed and probably I used the wrong one. Thanks for spotting that!

LuDuda (approved):

samples/matter/template/src/template.zap
ArekBalysNordic: Fixed.

wiba-nordic (approved):

doublemis1 (approved):
LGTM!

PR #12530, opened 2023-10-04
applications: Matter Bridge: make adding new devices more flexible

markaj-nordic:
This patch allows to add different Matter devices tied to the same data provider
in multiple steps - up till now it has to be added in a single call.

### DIFF
--- a/samples/matter/common/src/bridge/bridge_manager.cpp
+++ b/samples/matter/common/src/bridge/bridge_manager.cpp
@@ -118,12 +118,16 @@ CHIP_ERROR BridgeManager::RemoveBridgedDevice(uint16_t endpoint, uint8_t &device
 CHIP_ERROR BridgeManager::SafelyRemoveDevice(uint8_t index)
 {
 	uint16_t duplicatedItemKeys[kMaxBridgedDevices];
-	uint8_t size;
 	bool removeProvider = true;
-
 	auto &devicePair = mDevicesMap[index];
-	if (mDevicesMap.ValueDuplicated(devicePair, duplicatedItemKeys, size)) {
-		for (uint8_t i = 0; i < size; i++) {
+
+	uint8_t duplicatesNumber = mDevicesMap.GetDuplicatesCount(devicePair, duplicatedItemKeys);
+	/* There must be at least 2 duplicates in the map to determine the real duplicate,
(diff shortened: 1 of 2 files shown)


LuDuda (approved):

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -176,6 +178,18 @@ CHIP_ERROR BridgeManager::AddSingleDevice(MatterBridgedDevice *device, BridgedDe
 {
 	uint8_t index{ 0 };
 	CHIP_ERROR err;
+	uint16_t duplicatedItemKeys[kMaxBridgedDevices];
+	BridgedDevicePair pair(device, dataProvider);
+
+	/* Check if the current provider is already contained in the map - if so, there is at least one duplicate */
+	bool isNewProvider = (Instance().mDevicesMap.GetDuplicates(pair, duplicatedItemKeys) == 0);
```
kkasperczyk-no, line 185: Nit: do we need to allocate a variable for it? Maybe we could just move this check to the if statement two lines below.

samples/matter/common/src/bridge/util/bridge_util.h
```diff
@@ -125,21 +125,20 @@ template <typename T, std::size_t N> struct FiniteMap {
 	}
 	std::size_t FreeSlots() { return N - mElementsCount; }
 
-	bool ValueDuplicated(const T &value, uint16_t *key, uint8_t &numberOfDuplicates)
+	uint8_t GetDuplicates(const T &value, uint16_t *key)
```
kkasperczyk-no, line 128: How about `GetDuplicatesCount or Number`? Currently the name suggest we would get e.g. list of duplicated objects.

kkasperczyk-no (approved):
Left two minor comments, but LGTM overall.

samples/matter/common/src/bridge/bridge_manager.cpp
markaj-nordic: I wanted to achieve more self-documented code :) I'd like to keep it that way if you don't mind. Depending on the returned number of duplicates we may have different logic in the application, so I guess explicitness, even if redundant, does not hurt.

PR #12533, opened 2023-10-05
drivers: mpsl: permit non-sync flash operations from fault exceptions

vidarbe:
This change allows the core dump module to store coredumps to flash when a fault exception has been raised.

Ref. NCSIDB-1033

### DIFF
--- a/drivers/mpsl/flash_sync/flash_sync_mpsl.c
+++ b/drivers/mpsl/flash_sync/flash_sync_mpsl.c
@@ -140,9 +140,16 @@ void nrf_flash_sync_set_context(uint32_t duration)
 	_context.request_length_us = duration;
 }
 
+static bool is_in_fault_isr(void)
+{
+	int32_t irqn = __get_IPSR() - 16;
+
+	return (irqn >= HardFault_IRQn && irqn <= UsageFault_IRQn);
+}
+
 bool nrf_flash_sync_is_required(void)
 {
-	return mpsl_is_initialized();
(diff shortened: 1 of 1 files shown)


rugeGerritsen (approved):

drivers/mpsl/flash_sync/flash_sync_mpsl.c
```diff
@@ -140,9 +140,17 @@ void nrf_flash_sync_set_context(uint32_t duration)
 	_context.request_length_us = duration;
 }
 
+static bool is_in_fault_isr(void)
+{
+	uint32_t isr;
+
+	isr = __get_IPSR();
+	return (isr >= 3 && isr <= 6);
```
gmarull, line 148: please do not hardcode numbers here, define 3 and 6, or do something like

```
int32_t irqn = __get_IPSR() - 16;

return (irqn >= HardFault_IRQn && irqn <= UsageFault_IRQn);
```

gmarull (changes requested):

drivers/mpsl/flash_sync/flash_sync_mpsl.c
de-nordic: Magic numbers.

vidarbe: Fixed

gmarull (approved):

de-nordic (approved):

PR #12536, opened 2023-10-05
applications: asset_tracker_v2: link cloud docs to TechDocs

coderbyheart:


peknis:
We want to wait on making updates like this until just before we close down the other platforms. NCS as a whole will need to make many similar changes.

We are not redirecting anything at the moment, as in no link changes to point to the new site from the current docs. It is just the banner that takes the user to the new site.

coderbyheart:
Ah, got it!

PR #12538, opened 2023-10-05
samples: matter: bridge sample with smp added

HubertHugo:
Bridge debug sample with smp_dfu added.

### DIFF
--- a/applications/matter_bridge/sample.yaml
+++ b/applications/matter_bridge/sample.yaml
@@ -2,7 +2,6 @@ sample:
   description: Matter Bridge
   name: Matter Bridge application
 tests:
-  # Excluded in quarantine.yaml to limit resources usage in integration builds
   applications.matter_bridge.release:
     build_only: true
     extra_args: CONF_FILE=prj_release.conf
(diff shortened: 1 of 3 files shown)


applications/matter_bridge/sample.yaml
```diff
@@ -26,3 +26,9 @@ tests:
     integration_platforms:
       - nrf7002dk_nrf5340_cpuapp
     platform_allow: nrf7002dk_nrf5340_cpuapp
+  applications.matter_bridge.smp_dfu.br_ble:
```
doublemis1, line 29: can you check in your code using ctf_tm -c spec -r  build, if this is correct name of the sample?

HubertHugo: I checked, it works fine

jnsgsr, line 29: * "debug" configuration will be redundant, I think it will be enough to have only "smp_dfu" similar to what we have for door lock
* in qurantine-integration remove 'applications.matter_bridge.release.br_ble' and for the comment in line 5 mark that excluded configuration is `applications.matter_bridge.release` that is put line there like here https://github.com/nrfconnect/sdk-nrf/blob/main/applications/matter_weather_station/sample.yaml#L14

jnsgsr (changes requested):

jnsgsr (approved):

doublemis1 (approved):

PR #12552, opened 2023-10-05
ppi_trace: add ppi_trace_dppi_ch_trace function

ankuns:
For current `ppi_trace_config` it is impossible to trace events for which PUBLISH register is overwritten in run-time after `ppi_trace_config` is called. This function allows also to trace just one event and does not allow to trace many-to-one topologies which are possible with use of DPPIs.

To overcome above limitations `ppi_trace_dppi_ch_trace` is added. In contrast to `ppi_trace_config` the function `ppi_trace_dppi_ch_trace` focuses on tracing rather the DPPI channel leaving the preparation of event network publishing to it to a user.

### DIFF
--- a/include/debug/ppi_trace.h
+++ b/include/debug/ppi_trace.h
@@ -6,6 +6,8 @@
 #ifndef __PPI_TRACE_H
 #define __PPI_TRACE_H
 
+#include <stdint.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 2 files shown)


jciupis (approved):

subsys/debug/ppi_trace/ppi_trace.c
```diff
@@ -196,6 +196,35 @@ void *ppi_trace_pair_config(uint32_t pin, uint32_t start_evt, uint32_t stop_evt)
...
+{
+#ifdef DPPI_PRESENT
+	uint32_t task;
+	int gpiote_ch;
+	nrf_gpiote_task_t task_id;
+
+	gpiote_ch = gpiote_channel_alloc(pin);
+	if (gpiote_ch < 0) {
+		LOG_ERR("Failed to allocate GPIOTE channel.");
+		return false;
```
nordic-krch, line 209: It would be better to use error codes `-ENOMEM` here and `-ENOTSUP` if DPPI is not present.

include/debug/ppi_trace.h
```diff
@@ -45,6 +48,24 @@ void *ppi_trace_config(uint32_t pin, uint32_t evt);
...
+ *
+ * @note Supported only on platforms equipped with DPPI.
+ *
+ * @param pin		Pin to use for tracing.
+ * @param dppi_ch	DPPI channel number to be traced on the pin.
+ *
+ * @retval true		The configuration succeeded.
+ * @return false	The configuration failed.
+ */
+bool ppi_trace_dppi_config(uint32_t pin, uint32_t dppi_ch);
```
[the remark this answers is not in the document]
ankuns: `ppi_trace_disable` do not free any resources just disable ppi. For `ppi_trace_dppi_config` we cannot do this way because we could damage functionality of the dppi network the dppi channel is part of. 
I'll rename the function to `ppi_trace_dppi_ch_trace`

subsys/debug/ppi_trace/ppi_trace.c
ankuns: done

nordic-krch (approved):

PR #12553, opened 2023-10-05
net: lib: rest_client: update rest_client_close_connection() log msgs

jayteemo:
Change keep alive message to `LOG_DBG`.
Add new `LOG_DBG` for when the socket is closed.
Set `used_socket_is_alive flag` to false since the `connect_socket` value is set to -1 after `close()` is called.

### DIFF
--- a/subsys/net/lib/rest_client/src/rest_client.c
+++ b/subsys/net/lib/rest_client/src/rest_client.c
@@ -273,11 +273,14 @@ static void rest_client_close_connection(struct rest_client_req_context *const r
 		ret = close(req_ctx->connect_socket);
 		if (ret) {
 			LOG_WRN("Failed to close socket, error: %d", errno);
+		} else {
+			LOG_DBG("Socket with id: %d was closed", req_ctx->connect_socket);
 		}
+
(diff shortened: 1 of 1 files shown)


subsys/net/lib/rest_client/src/rest_client.c
```diff
@@ -273,11 +273,14 @@ static void rest_client_close_connection(struct rest_client_req_context *const r
 		ret = close(req_ctx->connect_socket);
 		if (ret) {
 			LOG_WRN("Failed to close socket, error: %d", errno);
+		} else {
+			LOG_DBG("Socket with id: %d was closed", req_ctx->connect_socket);
 		}
+		resp_ctx->used_socket_is_alive = false;
```
tokangas, line 279: Most of the `resp_ctx` values are already initialized in `rest_client_do_api_call()`, only `http_status_code` and `used_socket_is_alive` are left uninitialized. I would move this line there and also set `http_status_code` to zero.

tokangas (approved):

trantanen (approved):

jhirsi (approved):

PR #12561, opened 2023-10-06
bl_storage: Remove temporary definition used for TF-M attestation

oyvindronningstad:


sigvartmh (approved):

mswarowsky (approved):

subsys/bootloader/bl_storage/bl_storage.c
```diff
@@ -72,6 +72,9 @@ uint32_t num_public_keys_read(void)
 	return nrfx_nvmc_uicr_word_read(&BL_STORAGE->num_public_keys);
 }
 
+
+#ifdef CONFIG_SECURE_BOOT_CRYPTO
```
joerchan, line 76: I don't understand this part about conditional compilation exlusion of these functions.
They appear to all be about modifying or reading what is stored, and no crypto appears to be involved.

oyvindronningstad: They rely on definitions (`SB_*`) only available when bl_crypto is enabled, see e.g. line 90.

oyvindronningstad:
@mswarowsky can you take another look, needed some fixes to get CI to pass.

subsys/bootloader/bl_storage/bl_storage.c
joerchan: I can only see SB_PUBLIC_KEY_HASH_LEN, which does not appear to have any conditionals on it.
It's also used in the definition of `struct bl_storage_data`, which is the struct definition for BL_STORAGE.

Am I overlooking something?

mswarowsky (approved):

subsys/bootloader/bl_storage/bl_storage.c
joerchan: I removed these, seems to work just fine.

joerchan (approved):

PR #12570, opened 2023-10-09
application: serial_lte_modem: remove #XSLMUART AT command

tomi-font:
UART is now configured based only on the device tree.


SeppoTakalo (approved):

wiba-nordic (changes requested):

applications/serial_lte_modem/src/slm_uart_handler.c
```diff
@@ -329,15 +331,26 @@ int slm_uart_power_off(void)
...
-	LOG_DBG("Set UART baudrate to: %d", slm_uart.baudrate);
+	err = uart_config_get(uart_dev, &cfg);
+	if (err) {
+		LOG_ERR("uart_config_get: %d", err);
+		return err;
+	}
 
-	err = uart_configure(uart_dev, &slm_uart);
-	if (err != 0) {
+	err = uart_configure(uart_dev, &cfg);
```
MarkusLassila, line 346: I don't think that you need to call this as the baudrate is already set. You just need to get it with uart_config_get to show the relevant information in traces. I think that this method could be removed and put back to slm_uart_handler_init, but can be kept as a separate method as well, as long as it is renamed to indicate that we are not changing anything.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -247,6 +247,7 @@ Serial LTE modem
     They are replaced by the ``#XNRFCLOUDPOS`` command that allows to combine cellular and Wi-Fi data to determine the device location.
   * The AT commands ``#XAGPS`` and ``#XPGPS``.
```
MarkusLassila, line 248: There is another mention about removing a part of #XSLMUART command. I think that it could be removed. We ended up removing everything.

MarkusLassila:
Remove from slm_testing.rst as well.

MarkusLassila (changes requested):

wiba-nordic (approved):

MarkusLassila (approved):

PR #12575, opened 2023-10-09
subsys: net: lib: nrf_cloud: Reuse socket for A-GNSS

tokangas:
A-GNSS data download usually requires more than one HTTP request. Now the same TCP/TLS connection is used for all HTTP requests.


trantanen (approved):

jhirsi (approved):

plskeggs (approved):

jayteemo (approved):
nice fix 😅

glarsennordic (approved):

PR #12578, opened 2023-10-09
net_core_monitor: Make it easy to use

maje-emb:
The `__weak ncm_net_core_event_handler` function is used to notify the application about the network core staus.
This repleace the periodic call of the
`ncm_net_status_check` function.
This makes it easier to use and configure.

### DIFF
--- a/doc/nrf/libraries/others/network_core_monitor.rst
+++ b/doc/nrf/libraries/others/network_core_monitor.rst
@@ -27,12 +27,8 @@ The library uses two general-purpose registers of the IPC peripheral in the appl
 
 The ``GPMEM[0]`` register is divided into two 16-bit parts.
 
-The ``COUNTER`` value is incremented by the network core every :kconfig:option:`CONFIG_NCM_APP_FEEDING_INTERVAL_MSEC`.
-The application core periodically calls the :c:func:`ncm_net_status_check` function to check the status of the network core.
+The ``COUNTER`` value is incremented by the network core every :kconfig:option:`CONFIG_NCM_FEEDING_INTERVAL_MSEC`.
 If the network core is suspended, the counter value is not updated.
-The :c:func:`ncm_net_status_check` function returns error code ``-EBUSY``.
-
-To correctly detect the network core status, the :c:func:`ncm_net_status_check` function should be called less frequently than the value set in the :kconfig:option:`CONFIG_NCM_APP_FEEDING_INTERVAL_MSEC` Kconfig option.
 
 The ``FLAGS`` field has the ``Reset`` flag as the bit 0.
 
@@ -43,13 +39,23 @@ During the startup of the network core, the reset bit is set and the cause of th
 This value is rewritten from the network core's ``RESET.RESETREAS`` register.
 For a detailed description of the bits in this register, see the `RESETREAS`_ section for nRF5340.
 
-The :c:func:`ncm_net_status_check` function returns the following error codes:
+The :c:func:`ncm_net_core_event_handler` function may be implemented by your application.
+On the application core, the network core monitor checks the values of IPC registers written by the network core every :kconfig:option:`CONFIG_NCM_FEEDING_INTERVAL_MSEC`.
+If the network core malfunctions and fails to increment the ``COUNTER`` value, the :c:func:`ncm_net_core_event_handler` function is called on the application core.
+This function is also called when the network core is restarted.
+The network core monitor provides a ``__weak`` implementation of that function in the :file:`nrf/subsys/net_core_monitor/app_core.c` file.
 
-* ``-EBUSY`` if the network core is suspended
-* ``-EFAULT`` if a network core reset has occurred
+The following events are supported and also listed in the :file:`nrf/include/net_core_monitor.h` file:
 
-The function takes a pointer to a variable of type uint32_t as a parameter.
-When a network core reset is detected, the cause of the reset is stored in this pointer.
+* :c:macro:`NCM_EVT_NET_CORE_RESET`
+
+   * Event triggered when a network core reset occurs.
+   * The ``reset_reas`` variable holds the reason for the reset.
+     It is rewritten from the ``RESET.RESETREAS`` register.
+
+* :c:macro:`NCM_EVT_NET_CORE_FREEZE`
+
+   * Event triggered when the network core is not responding.
 
 Configuration
 *************
@@ -68,38 +74,28 @@ Usage
 
 To enable the Network core monitor library, set the :kconfig:option:`CONFIG_NET_CORE_MONITOR` Kconfig option.
 
-On the application core, periodically call the :c:func:`ncm_net_status_check` function to monitor the state of the network core.
+The :c:func:`ncm_net_core_event_handler` function can be used to notify the application about the status of the network core.
+To define the user action for the event, you must override the ``__weak`` function definition of :c:func:`ncm_net_core_event_handler`.
+Otherwise, the ``__weak``` definition is called and it prints information about the event.
+
 See the following usage example.
 
 .. code-block::
 
    #include "net_core_monitor.h"
    ...
-   static void print_reset(uint32_t reset_reas)
+   /* This is the override for the __weak handler. */
+   void ncm_net_core_event_handler(enum ncm_event_type event, uint32_t reset_reas)
    {
-      if (reset_reas & NRF_RESET_RESETREAS_RESETPIN_MASK) {
-         printk("Reset by pin-reset\n");
-      } else if (reset_reas & NRF_RESET_RESETREAS_DOG0_MASK) {
-         printk("Reset by application watchdog timer 0 \n");
-      } else if (reset_reas & NRF_RESET_RESETREAS_SREQ_MASK) {
-         printk("Reset by soft-reset\n");
-      } else if (reset_reas) {
-         printk("Reset by a different source (0x%08X)\n", reset_reas);
-      }
-   }
-
-   int main(void)
-   {
-      uint32_t reset_reas;
-      ...
-      for (;;) {
-         ret = ncm_net_status_check(&reset_reas);
-         if (ret == -EBUSY) {
-            /* do something*/
-         } else if (ret == -EFAULT) {
-            print_reset(reset_reas);
-         }
-         k_sleep(K_MSEC(1000));
+      switch (event) {
+      case NCM_EVT_NET_CORE_RESET:
+         printk("The network core reset.\n");
+         /* do something */
+         break;
+      case NCM_EVT_NET_CORE_FREEZE:
+         printk("The network core is not responding.\n");
+         /* do something */
+         break;
       }
    }
 
--- a/include/net_core_monitor.h
+++ b/include/net_core_monitor.h
@@ -18,21 +18,29 @@
 extern "C" {
 #endif
 
-/** @brief Check the status of the network core.
+/** @brief Network core monitor event types, used to signal the application. */
+enum ncm_event_type {
+	/** Event triggered when a network core reset occurs.
+	 *  The ``reset_reas`` variable holds the reason for the reset.
+	 *  It is rewritten from the RESET.RESETREAS register.
+	 */
+	NCM_EVT_NET_CORE_RESET,
+
+	/** Event triggered when the network core is not responding. */
+	NCM_EVT_NET_CORE_FREEZE
+};
+
+/** @brief Event handler that is called by the Network core monitor library when an event occurs.
  *
- *  The function should be called less frequently than
- *  @kconfig{CONFIG_NCM_APP_FEEDING_INTERVAL_MSEC} to correctly detect the network core status.
+ * @note This function should be defined by the application.
+ *       Otherwise, `__weak` definition will called and it prints information about the event.
  *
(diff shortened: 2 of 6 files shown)


doc/nrf/libraries/others/network_core_monitor.rst
```diff
@@ -43,13 +39,20 @@ During the startup of the network core, the reset bit is set and the cause of th
 This value is rewritten from the network core's ``RESET.RESETREAS`` register.
 For a detailed description of the bits in this register, see the `RESETREAS`_ section for nRF5340.
 
-The :c:func:`ncm_net_status_check` function returns the following error codes:
+The :c:func:`ncm_net_core_event_handler` function may be implemented by your application.
+A ``__weak`` implementation can be found in the :file:`nrf/subsys/net_core_monitor/app_core.c`.
+
+The following events are supported and are also listed in :file:`nrf/include/net_core_monitor.h`.
```
peknis, line 45:
```suggestion
The following events are supported and also listed in the :file:`nrf/include/net_core_monitor.h` file:
```

doc/nrf/libraries/others/network_core_monitor.rst
```diff
@@ -43,13 +39,20 @@ During the startup of the network core, the reset bit is set and the cause of th
...
-* ``-EFAULT`` if a network core reset has occurred
+   * Event triggered when a network core reset occurs.
+   * The ``reset_reas`` variable holds the reason for the reset.
+     It is rewritten from the ``RESET.RESETREAS`` register.
 
-The function takes a pointer to a variable of type uint32_t as a parameter.
-When a network core reset is detected, the cause of the reset is stored in this pointer.
+* :c:macro:`NCM_FREEZ_EVENT`
+
+   * Event triggered when a network core is not responding.
```
peknis, line 55:
```suggestion
   * Event triggered when the network core is not responding.
```

doc/nrf/libraries/others/network_core_monitor.rst
```diff
@@ -68,38 +71,27 @@ Usage
 
 To enable the Network core monitor library, set the :kconfig:option:`CONFIG_NET_CORE_MONITOR` Kconfig option.
 
-On the application core, periodically call the :c:func:`ncm_net_status_check` function to monitor the state of the network core.
+The :c:func:`ncm_net_core_event_handler` function can be used to notify the application about the status of the network core.
+To define the user action for the event, you must to override the ``__weak`` function definition of the :c:func:`ncm_net_core_event_handler`.
+Otherwise, the ``__weak``` definition is called, which prints information about the event.
```
peknis, line 76:
```suggestion
Otherwise, the ``__weak``` definition is called and it prints information about the event.
```

include/net_core_monitor.h
```diff
@@ -18,21 +18,29 @@
...
- *  @kconfig{CONFIG_NCM_APP_FEEDING_INTERVAL_MSEC} to correctly detect the network core status.
+/** @brief Network core monitor event types, used to signal the application. */
+enum ncm_event_type {
+	/** Event triggered when a network core reset occurs.
+	 *  The @param reset_reas variable holds the reason for the reset.
+	 *  It is rewritten from the RESET.RESETREAS register.
+	 */
+	NCM_EVT_NET_CORE_RESET,
+
+	/** Event triggered when a network core is not responding. */
```
peknis, line 29:
```suggestion
	/** Event triggered when the network core is not responding. */
```

include/net_core_monitor.h
```diff
@@ -18,21 +18,29 @@
...
+	NCM_EVT_NET_CORE_FREEZE
+};
+
+/** @brief Event handler that is called by the Network Core Monitor library when an event occurs.
  *
- *  @param[out]  reset_reas  Reason for network core reset.
- *                           When the -EFAULT code is returned, the variable is set.
- *                           This value is rewritten from the network core's RESET.RESETREAS
- *                           register.
+ * @note This function should be defined bye the application.
```
peknis, line 35:
```suggestion
 * @note This function should be defined by the application.
```

include/net_core_monitor.h
```diff
@@ -18,21 +18,29 @@
...
+ *       Otherwise, `__weak` definition will called, that prints information about the event.
```
peknis, line 36:
```suggestion
 *       Otherwise, `__weak` definition will called and it prints information about the event.
```

doc/nrf/libraries/others/network_core_monitor.rst
```diff
@@ -68,38 +71,27 @@ Usage
 
 To enable the Network core monitor library, set the :kconfig:option:`CONFIG_NET_CORE_MONITOR` Kconfig option.
 
-On the application core, periodically call the :c:func:`ncm_net_status_check` function to monitor the state of the network core.
+The :c:func:`ncm_net_core_event_handler` function can be used to notify the application about the status of the network core.
+To define the user action for the event, you must to override the ``__weak`` function definition of the :c:func:`ncm_net_core_event_handler`.
```
peknis, line 75:
```suggestion
To define the user action for the event, you must override the ``__weak`` function definition of the :c:func:`ncm_net_core_event_handler`.
```

include/net_core_monitor.h
```diff
@@ -18,21 +18,29 @@
...
+	 *  The @param reset_reas variable holds the reason for the reset.
+	 *  It is rewritten from the RESET.RESETREAS register.
+	 */
+	NCM_EVT_NET_CORE_RESET,
+
+	/** Event triggered when a network core is not responding. */
+	NCM_EVT_NET_CORE_FREEZE
+};
+
+/** @brief Event handler that is called by the Network Core Monitor library when an event occurs.
```
peknis, line 33:
```suggestion
/** @brief Event handler that is called by the Network core monitor library when an event occurs.
```

peknis (changes requested):

doc/nrf/libraries/others/network_core_monitor.rst
```diff
@@ -43,13 +39,20 @@ During the startup of the network core, the reset bit is set and the cause of th
 This value is rewritten from the network core's ``RESET.RESETREAS`` register.
 For a detailed description of the bits in this register, see the `RESETREAS`_ section for nRF5340.
 
-The :c:func:`ncm_net_status_check` function returns the following error codes:
+The :c:func:`ncm_net_core_event_handler` function may be implemented by your application.
+A ``__weak`` implementation can be found in the :file:`nrf/subsys/net_core_monitor/app_core.c`.
+
+The following events are supported and are also listed in :file:`nrf/include/net_core_monitor.h`.
+
+* :c:macro:`NCM_RESET_EVENT`
```
jciupis, line 47: These names should be taken from `net_core_monitor.h`, right? They differ now

subsys/net_core_monitor/app_core.c
```diff
@@ -17,7 +20,12 @@ LOG_MODULE_REGISTER(net_core_monitor, CONFIG_NET_CORE_MONITOR_LOG_LEVEL);
 #define IPC_MEM_CNT_IDX  0
 #define IPC_MEM_REAS_IDX 1
 
-int ncm_net_status_check(uint32_t * const reset_reas)
+#define NET_CORE_CHECK_INTERVAL_MSEC (CONFIG_NCM_FEEDING_INTERVAL_MSEC + 500)
```
jciupis, line 23: I think this is misleading. `CONFIG_NCM_FEEDING_INTERVAL_MSEC` affects power consumption on both cores so we shouldn't hardcode any offset here.

My suggestion is to change the definition of the symbol in Kconfig so that is has different default depending on the core. Also, the requirement for the interval on the application core being not lesser than the internal on the network core should be documented. Something like:
```
config NCM_FEEDING_INTERVAL_MSEC
	default 500 if SOC_NRF5340_CPUNET
        default 1000 if SOC_NRF5340_CPUAPP
	int "Feeding interval in milliseconds"
        help
          The value of this parameter on the application core must be greater
          than the value on the network core. Otherwise the network core monitor
          will report false positive network code malfunctions.
```

or similar.

doc/nrf/libraries/others/network_core_monitor.rst
maje-emb: You are right, these names should be taken from `net_core_monitor.h` file.

subsys/net_core_monitor/app_core.c
maje-emb: Good point. I have applied your suggestions.

peknis (approved):

doc/nrf/libraries/others/network_core_monitor.rst
```diff
@@ -43,13 +39,20 @@ During the startup of the network core, the reset bit is set and the cause of th
 This value is rewritten from the network core's ``RESET.RESETREAS`` register.
 For a detailed description of the bits in this register, see the `RESETREAS`_ section for nRF5340.
 
-The :c:func:`ncm_net_status_check` function returns the following error codes:
+The :c:func:`ncm_net_core_event_handler` function may be implemented by your application.
```
jciupis, line 42: IMO it would be good to explain _why_ the user could want to implement this function. For example:
> On the application core, the network core monitor checks the values of IPC registers written by the network core every :kconfig:option:`CONFIG_NCM_FEEDING_INTERVAL_MSEC`. In case the network core malfunctions and fails to increment the ``COUNTER`` value as documented above, :c:func:`ncm_net_core_event_handler` is called on the application core. The network core monitor provides a ``__weak`` implementation of that function...

Without it it's unclear what this function does and why anyone should want to implement it.

jciupis (commented):
One more comment regarding the documentation changes. The code looks good to me

doc/nrf/libraries/others/network_core_monitor.rst
peknis: - In case -> if
- leave out "as documented above"
- The :c:func:... function is called

maje-emb: Expanded the description with:

> On the application core, the network core monitor checks the values of IPC registers written by the network core every :kconfig:option:`CONFIG_NCM_FEEDING_INTERVAL_MSEC`.
If the network core malfunctions and fails to increment the ``COUNTER`` value, the :c:func:`ncm_net_core_event_handler` function is called on the application core.
This function is also called when the network core is restarted.
The network core monitor provides a ``__weak`` implementation of that function in the :file:`nrf/subsys/net_core_monitor/app_core.c` file.

The usage section contains a sentence why users may want to implement this function.

jciupis (approved):

subsys/net_core_monitor/app_core.c
```diff
@@ -49,3 +57,49 @@ int ncm_net_status_check(uint32_t * const reset_reas)
...
+		LOG_DBG("The network core reset.");
+		if (reset_reas & NRF_RESET_RESETREAS_RESETPIN_MASK) {
+			LOG_DBG("Reset by pin-reset.");
+		} else if (reset_reas & NRF_RESET_RESETREAS_DOG0_MASK) {
+			LOG_DBG("Reset by application watchdog timer 0.");
+		} else if (reset_reas & NRF_RESET_RESETREAS_SREQ_MASK) {
+			LOG_DBG("Reset by soft-reset");
+		} else if (reset_reas) {
+			LOG_DBG("Reset by a different source (0x%08X).", reset_reas);
+		}
```
KAGA164, line 90: Please add empty else statement with comment inside

subsys/net_core_monitor/app_core.c
```diff
@@ -49,3 +57,49 @@ int ncm_net_status_check(uint32_t * const reset_reas)
...
+	int ret;
+	uint32_t reset_reas;
+
+	ret = ncm_net_status_check(&reset_reas);
+
+	if (ret == -EBUSY) {
+		ncm_net_core_event_handler(NCM_EVT_NET_CORE_FREEZE, 0);
+	} else if (ret == -EFAULT) {
+		ncm_net_core_event_handler(NCM_EVT_NET_CORE_RESET, reset_reas);
+	}
```
KAGA164, line 72: Please add empty else statement with comment inside

doc/nrf/libraries/others/network_core_monitor.rst
```diff
@@ -68,38 +74,28 @@ Usage
 
 To enable the Network core monitor library, set the :kconfig:option:`CONFIG_NET_CORE_MONITOR` Kconfig option.
 
-On the application core, periodically call the :c:func:`ncm_net_status_check` function to monitor the state of the network core.
+The :c:func:`ncm_net_core_event_handler` function can be used to notify the application about the status of the network core.
+To define the user action for the event, you must override the ``__weak`` function definition of the :c:func:`ncm_net_core_event_handler`.
```
KAGA164, line 78: @peknis  ?
```suggestion
To define the user action for the event, you must override the ``__weak`` function definition of the :c:func:`ncm_net_core_event_handler` function.
```

peknis: Maybe this is a bit of an overkill, now that we already talk about a function definition. Maybe you can leave it without the definite article "the" in front of the function name and the word "function" at the end.

maje-emb: Thanks for the clarification.

KAGA164 (approved):

PR #12581, opened 2023-10-09
drivers: wifi: Fix throughput degradation in BLE

krish2718:
The anomaly 43 workaround introduces extra delay due to waiting for QSPI ready and causes BLE throughput drops as Wi-Fi driver is waiting and halting the APP.

But this anomaly is only for XIP, so, can be disabled when using QSPI for Wi-Fi comms.

Fixes SHEL-1633.

### DIFF
--- a/drivers/wifi/nrf700x/CMakeLists.txt
+++ b/drivers/wifi/nrf700x/CMakeLists.txt
@@ -115,3 +115,10 @@ if (CONFIG_NRF_WIFI_PATCHES_EXT_FLASH)
 # Run patches from the external flash (XIP). No need to copy.
 zephyr_code_relocate(FILES zephyr/src/zephyr_fw_load.c LOCATION EXTFLASH_RODATA NOCOPY)
 endif()
+
+# These are XIP related anomalies and aren't applicable for nRF7002 and cause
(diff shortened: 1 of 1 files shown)


sachinthegreen (approved):

muraliThokala (approved):

drivers/wifi/nrf700x/CMakeLists.txt
```diff
@@ -115,3 +115,12 @@ if (CONFIG_NRF_WIFI_PATCHES_EXT_FLASH)
 # Run patches from the external flash (XIP). No need to copy.
 zephyr_code_relocate(FILES zephyr/src/zephyr_fw_load.c LOCATION EXTFLASH_RODATA NOCOPY)
 endif()
+
+if (CONFIG_NRF700X_ON_QSPI)
+# These are XIP related anomalies and aren't applicable for nRF7002 and cause
+# throughput issues.
+zephyr_compile_definitions(
```
anangl, line 122: Wouldn't `zephyr_compile_definitions_ifdef(CONFIG_NRF700X_ON_QSPI` suit better here?

anangl (approved):

PR #12586, opened 2023-10-09
lib: modem_slm: handle missing null-terminator

tomi-font:
This fixes the case where the UART buffer is not null-terminated.


SeppoTakalo (approved):

junqingzou (approved):
Pls forward to customer to test

MarkusLassila (approved):

PR #12594, opened 2023-10-10
samples: wifi: twt: Added checks for handling multiple errors.

chiranjeevi2776:
1. Update the Flow ID from the TWT response for teardown
2. Run the traffic when the TWT response is 'accept'; otherwise, do not run the traffic.
3. Added checks for TWT teardown in case of errors during traffic.
4. Print client report when server report not received.

### DIFF
--- a/samples/wifi/twt/src/main.c
+++ b/samples/wifi/twt/src/main.c
@@ -55,7 +55,7 @@ static struct {
 	};
 } context;
 
-static bool twt_supported, twt_resp_received;
+static bool twt_supported, twt_resp_received, twt_resp_accept;
 static uint32_t twt_flow_id = 1;
 
 static bool wait_for_twt_resp_received(void)
@@ -65,7 +65,7 @@ static bool wait_for_twt_resp_received(void)
 	for (i = 0; i < timeout_polls; i++) {
 		k_sleep(K_MSEC(STATUS_POLLING_MS));
 		if (twt_resp_received) {
-			return true;
+			return twt_resp_accept;
 		}
 	}
 
@@ -253,6 +253,10 @@ static void handle_wifi_twt_event(struct net_mgmt_event_callback *cb)
 	if (resp->resp_status == WIFI_TWT_RESP_RECEIVED) {
 		LOG_INF("TWT response: %s",
 		      wifi_twt_setup_cmd_txt(resp->setup_cmd));
+
+		if (resp->setup_cmd == WIFI_TWT_SETUP_CMD_ACCEPT) {
+			twt_resp_accept = true;
+		}
 
 		twt_resp_received = true;
 		twt_flow_id = resp->flow_id;
--- a/samples/wifi/twt/src/main.c
+++ b/samples/wifi/twt/src/main.c
@@ -56,6 +56,7 @@ static struct {
 } context;
 
 static bool twt_supported, twt_resp_received;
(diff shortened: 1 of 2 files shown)


krish2718 (changes requested):
The numbered list in the commit is an indication that this commit needs to be split, each item can be a commit. 

nits, please use `imperative` form (Added->Add) and no full-stop at the end for the title.

samples/wifi/twt/src/main.c
```diff
@@ -64,7 +65,12 @@ static bool wait_for_twt_resp_received(void)
 	for (i = 0; i < timeout_polls; i++) {
 		k_sleep(K_MSEC(STATUS_POLLING_MS));
 		if (twt_resp_received) {
-			return true;
+			/* check for twt resp as accept, if not return failure */
```
krish2718, line 68: What is AP says `suggest` shouldn't we demonstrate how to handle that case to our customers?

samples/wifi/twt/src/main.c
```diff
@@ -55,7 +55,8 @@ static struct {
 	};
 } context;
 
-static bool twt_supported, twt_resp_received;
+static bool twt_supported, twt_resp_received, twt_resp_accept;
+uint32_t twt_flow_id = 1;
```
krish2718, line 59: This should also be `static`

samples/wifi/twt/src/main.c
```diff
@@ -64,7 +65,12 @@ static bool wait_for_twt_resp_received(void)
 	for (i = 0; i < timeout_polls; i++) {
 		k_sleep(K_MSEC(STATUS_POLLING_MS));
 		if (twt_resp_received) {
-			return true;
+			/* check for twt resp as accept, if not return failure */
+			if (twt_resp_accept) {
+				return true;
+			} else {
+				return false;
+			}
```
krish2718, line 73: Just doing `return twt_resp_accept` should do.

samples/wifi/twt/src/main.c
```diff
@@ -249,9 +256,15 @@ static void handle_wifi_twt_event(struct net_mgmt_event_callback *cb)
 	}
 
 	if (resp->resp_status == WIFI_TWT_RESP_RECEIVED) {
-		twt_resp_received = 1;
 		LOG_INF("TWT response: %s",
 		      wifi_twt_setup_cmd_txt(resp->setup_cmd));
+
+		if (resp->setup_cmd == WIFI_TWT_SETUP_CMD_ACCEPT)
```
krish2718, line 262: NCS uses braces even for single statement

samples/wifi/twt/src/main.c
```diff
@@ -249,9 +256,15 @@ static void handle_wifi_twt_event(struct net_mgmt_event_callback *cb)
...
+			twt_resp_accept = true;
+
+		twt_resp_received = 1;
```
krish2718, line 265: `true` might be better

samples/wifi/twt/src/main.c
```diff
@@ -443,12 +456,14 @@ int main(void)
 
 			ret = traffic_gen_start(&tg_config);
 			if (ret < 0) {
+				teardown_twt();
```
krish2718, line 459: why not use `goto` here as you have used above?

krish2718 (changes requested):

PR #12599, opened 2023-10-10
doc: wifi: update memory footprint table

richabp:
Updated the memory footprint table with the latest values.


ioannisg:
@krish2718  I wonder if we should just wait for this until the scan-only patch PR is merged (assuming this would lower the ROM numbers)

krish2718:
> @krish2718 I wonder if we should just wait for this until the scan-only patch PR is merged (assuming this would lower the ROM numbers)

Its merged today :).

krish2718:
@richabp can you please rebase this PR and re-run the CI as well?

ioannisg:
> > @krish2718 I wonder if we should just wait for this until the scan-only patch PR is merged (assuming this would lower the ROM numbers)
> 
> Its merged today :).

So confirming the scan-only ROM numbers are updated with the scan-only patch size?

krish2718:
> > > @krish2718 I wonder if we should just wait for this until the scan-only patch PR is merged (assuming this would lower the ROM numbers)
> > 
> > 
> > Its merged today :).
> 
> So confirming the scan-only ROM numbers are updated with the scan-only patch size?

Yup, the diff shows 76->36, 40K reduction.

krish2718 (approved):

visinor (approved):

kittydepa (approved):

PR #12604, opened 2023-10-10
Bluetooth: Mesh: NLC samples CDP2 support

Balaklaka:
Adding CDP2 support to the NLC samples. Also change the samples to use Mesh-v1.1 as CDP2 is dependent on Mesh-v1.1.


doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -352,10 +352,12 @@ Bluetooth mesh samples
 
   * Fixed an issue where the sample could return an invalid Light Lightness Status message if the transition time was evaluated to zero.
   * Removed support for the configuration with :ref:`CMSE enabled <app_boards_spe_nspe_cpuapp_ns>` for :ref:`zephyr:thingy53_nrf5340`.
+  * Added :kconfig:option:`CONFIG_BT_MESH_COMP_PAGE_2` to add Composition Data Page 2 support, Composition Data Page 2 is dependent on Bluetooth mesh 1.1 support.
```
mia-ko, line 355: Move added to before fixed, at the top of the list.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -352,10 +352,12 @@ Bluetooth mesh samples
...
 
 * :ref:`bluetooth_mesh_light_dim` sample:
 
   * Removed support for the configuration with :ref:`CMSE enabled <app_boards_spe_nspe_cpuapp_ns>` for :ref:`zephyr:thingy53_nrf5340`.
+  * Added :kconfig:option:`CONFIG_BT_MESH_COMP_PAGE_2` to add Composition Data Page 2 support, Composition Data Page 2 is dependent on Bluetooth mesh 1.1 support.
```
mia-ko, line 360: See previous comment. Also move this bullet point to top of the list.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -366,6 +368,7 @@ Bluetooth mesh samples
   * Added a getter for the :c:var:`bt_mesh_sensor_rel_runtime_in_a_dev_op_temp_range` sensor.
   * Removed support for the configuration with :ref:`CMSE enabled <app_boards_spe_nspe_cpuapp_ns>` for :ref:`zephyr:thingy53_nrf5340`.
   * Fixed an issue where the :c:var:`bt_mesh_sensor_time_since_presence_detected` sensor could report an invalid value when the time delta would exceed the range of the characteristic.
+  * Added :kconfig:option:`CONFIG_BT_MESH_COMP_PAGE_2` to add Composition Data Page 2 support, Composition Data Page 2 is dependent on Bluetooth mesh 1.1 support.
```
mia-ko, line 371: See previous comments. Move to the top of the list.

mia-ko (changes requested):

ludvigsj (approved):

samples/bluetooth/mesh/light_ctrl/prj.conf
```diff
@@ -52,6 +52,8 @@ CONFIG_BT_MESH_ADV_EXT_GATT_SEPARATE=y
 CONFIG_BT_MESH_DK_PROV=y
 CONFIG_BT_MESH_NLC_PERF_CONF=y
 CONFIG_BT_MESH_NLC_PERF_LIGHTNESS_CTRL=y
+CONFIG_BT_MESH_V1d1=y
+CONFIG_BT_MESH_COMP_PAGE_2=y
```
omkar3141, line 56: I would suggest to move this inside and use `BT_MESH_NLC_PERF_CONF` to derive the `y` value. That way customers have one less item to configure in "prj.conf". Something like this:

nrf\subsys\bluetooth\mesh\Kconfig
```
config BT_MESH_NLC_PERF_CONF
   ...
   select BT_MESH_COMP_PAGE_2
   ...
```

omkar3141 (changes requested):

samples/bluetooth/mesh/light_ctrl/src/model_handler.c
```diff
@@ -21,6 +21,37 @@ struct lightness_ctx {
 	uint32_t rem_time;
 };
 
+#if IS_ENABLED(CONFIG_BT_MESH_COMP_PAGE_2)
+static const uint8_t cmp2_elem_offset1[2] = { 0, 1 };
+static const uint8_t cmp2_elem_offset2[1] = { 0 };
+
+static const struct bt_mesh_comp2_record comp_rec[2] = {
+	{
+	.id = 0x1601, /* Basic Lightness Controller NLC Profile 1.0 */
```
omkar3141, line 30: Please define macros for these values, as they are assigned numbers. May be in `models.h`.

omkar3141 (changes requested):

samples/bluetooth/mesh/light_ctrl/src/model_handler.c
Balaklaka: We do not have any defined macros for these model ids.

omkar3141: I'm suggesting to define new ones in `models.h`.

samples/bluetooth/mesh/light_ctrl/prj.conf
Balaklaka: I was thinking the same, the reason I did not do it was to show that we now enable Mesh-v1.1. Will move the settings.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -718,6 +748,12 @@ static const struct bt_mesh_comp comp = {
 
 const struct bt_mesh_comp *model_handler_init(void)
 {
+#if IS_ENABLED(CONFIG_BT_MESH_COMP_PAGE_2)
```
omkar3141, line 751:
```suggestion
#if IS_ENABLED(CONFIG_BT_MESH_NLC_PERF_CONF)
```

samples/bluetooth/mesh/light_dimmer/src/model_handler.c
```diff
@@ -275,6 +305,12 @@ static const struct bt_mesh_comp comp = {
 
 const struct bt_mesh_comp *model_handler_init(void)
 {
+#if IS_ENABLED(CONFIG_BT_MESH_COMP_PAGE_2)
```
omkar3141, line 308:
```suggestion
#if IS_ENABLED(CONFIG_BT_MESH_NLC_PERF_CONF)
```

samples/bluetooth/mesh/light_ctrl/src/model_handler.c
```diff
@@ -244,6 +275,12 @@ void model_handler_start(void)
 {
 	int err;
 
+#if IS_ENABLED(CONFIG_BT_MESH_COMP_PAGE_2)
```
omkar3141, line 278: Also I suggest this:

```suggestion
#if IS_ENABLED(CONFIG_BT_MESH_NLC_PERF_CONF)
```

Since, this is required if NLC profiles are supported, it seems more logical to use profile specific KConfig option (which would in-turn, enable COMP_PAGE_2 option).

omkar3141 (changes requested):

samples/bluetooth/mesh/sensor_server/src/model_handler.c
Balaklaka: But at the same time, if someone want to use CONFIG_BT_MESH_COMP_PAGE_2 without CONFIG_BT_MESH_NLC_PERF_CONF. This will then be left out. That means the user have to use NLC not only comp page 2. That is why I would like to keep it as is.

samples/bluetooth/mesh/light_dimmer/src/model_handler.c
Balaklaka: But at the same time, if someone want to use CONFIG_BT_MESH_COMP_PAGE_2 without CONFIG_BT_MESH_NLC_PERF_CONF. This will then be left out. That means the user have to use NLC not only comp page 2. That is why I would like to keep it as is.

samples/bluetooth/mesh/light_ctrl/src/model_handler.c
Balaklaka: But at the same time, if someone want to use CONFIG_BT_MESH_COMP_PAGE_2 without CONFIG_BT_MESH_NLC_PERF_CONF. This will then be left out. That means the user have to use NLC not only comp page 2. That is why I would like to keep it as is.

ludvigsj: I thought so at first as well. But, without NLC_PERF_CONF they wouldn't be profile compliant, and then the _should_ have to define the CDP2 contents manually (because the code here would incorrectly claim NLC compliance otherwise)

ludvigsj: No, they will have to change the code then. They can't have this content that is inside the #if (specifying the NLC profiles) in the CDP2. Because then the device would claim to be NLC compliant (when other nodes ask for CDP2), but wouldn't actually be compliant. So then it's fine if this is #if-ed out, because they either way need to set up their own CDP contents.

ludvigsj: Also, the only purpose of CDP2 (at least currently) is to claim that you support one or more NLC profiles. It makes no sense to use CDP2 but not turn on NLC compliance.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -350,11 +350,15 @@ Bluetooth mesh samples
 
 * :ref:`bluetooth_mesh_light_lc` sample:
 
+  * Added the Kconfig option :kconfig:option:`CONFIG_BT_MESH_COMP_PAGE_2` for Composition Data Page 2 support.
```
omkar3141, line 353: I think these release notes need to change as per the current changes. Also applies to the similar lines in this file.
```suggestion
  * Added Composition Data Page 2 support.
```

omkar3141 (changes requested):

omkar3141 (approved):

PR #12605, opened 2023-10-10
samples/cellular: update default libc

lemrey:
These samples needed to explicitly select `newlibc` when `minimal` libc was the default upstream.
Now that `picolibc` is the default, let samples uses the default C library where possible.

In some places I have removed float support in `printf` and `scanf` too, as it did not seem to be necessary.


lemrey:
I kept out `cellular/download` and `cellular/https_client` since they are being moved, and `cellular/http_update` because it is being updated.

SeppoTakalo (approved):

trantanen (approved):

samples/cellular/azure_fota/prj.conf
```diff
@@ -67,8 +67,7 @@ CONFIG_NET_NATIVE=n
 CONFIG_NRF_MODEM_LIB=y
 
 # newlibc
```
tokangas, line 69: This comment should be removed or updated.

tokangas (approved):

juhaylinen (approved):

samples/cellular/lwm2m_carrier/prj.conf
```diff
@@ -7,7 +7,6 @@
 CONFIG_LWM2M_CARRIER=y
 
 # The library requires newlibc
```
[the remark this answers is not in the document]
lemrey: thanks, updated

samples/cellular/azure_fota/prj.conf
lemrey: Thanks, I have updated that

rlubos (approved):

PR #12609, opened 2023-10-10
lib: modem_slm: fix IPv4 connection to an IPv6-enabled host

tomi-font:
The hostname used to be resolved to an IPv6 address (if found) even if the device had only an IPv4 address, which would make the connection to any IPv6-enabled host fail.

### DIFF
--- a/subsys/net/lib/ftp_client/src/ftp_client.c
+++ b/subsys/net/lib/ftp_client/src/ftp_client.c
@@ -408,45 +408,28 @@ static void keepalive_timeout(struct k_timer *dummy)
 	}
 }
 
-static int host_lookup(const char *hostname, int family, struct sockaddr *sa)
-{
-	int err;
-	struct addrinfo *ai;
-	struct addrinfo hints = {
-		.ai_family = family,
-		.ai_socktype = SOCK_STREAM
-	};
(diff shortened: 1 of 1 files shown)


subsys/net/lib/ftp_client/src/ftp_client.c
```diff
@@ -408,45 +408,28 @@ static void keepalive_timeout(struct k_timer *dummy)
...
 		LOG_ERR("FTP already connected");
 		return -EINVAL;
 	}
+	int ret;
+	struct addrinfo *ai;
 
-	/* Attempt IPv6 resolution, fallback to IPv4 if failed */
-	ret = host_lookup(hostname, AF_INET6, (struct sockaddr *)(&client.remote6));
+	/* Resolve the hostname in the preferred IP version .*/
+	ret = getaddrinfo(hostname, NULL, NULL, &ai);
```
MarkusLassila, line 421: Just to make sure, does this return IPv4 address if we only have IPv4 address for the device, but the server has both IPv6 and IPv4? And same for IPv6?

MarkusLassila (approved):

junqingzou (approved):

PR #12626, opened 2023-10-11
ci: Update pip requirements-fixed compilation

DatGizmo:
* The workflow now runs in pull_request_target context
* The workflow now posts a comment informing the author that the PR resulted in changes to the pip-requirements.txt
* If the user reacts to the comment with the :eyes: emote and reruns the check, the check will add the changes to the PR

Github actions used by this workflow:
https://github.com/NordicBuilder/action-script-diff 
https://github.com/NordicBuilder/action-comment

Example execution in sdk-nrf-testing:
https://github.com/nrfconnect/sdk-nrf-testing/pull/195

PR #12631, opened 2023-10-11
ci: Update requirements-fixed.txt format

DatGizmo:
* The compile script output format was changed, updating requirements-fixed.txt to the new format

This PR requires https://github.com/nrfconnect/sdk-nrf/pull/12626 to be merged, and might need a rebase.


thst-nordic (approved):

karhama (approved):
lgtm

scripts/requirements-fixed.txt
```diff
@@ -2,361 +2,132 @@
 # This file is autogenerated by pip-compile with Python 3.8
 # by the following command:
 #
-#    pip-compile --allow-unsafe --output-file=requirements-fixed.txt --strip-extras ./bootloader/mcuboot/scripts/requirements.txt ./nrf/scripts/requirements-ci.txt ./nrf/scripts/requirements-extra.txt ./nrf/scripts/requirements.txt ./zephyr/scripts/requirements.txt
+#    pip-compile --allow-unsafe --annotation-style=line --output-file=/home/runner/work/sdk-nrf-testing/sdk-nrf-testing/ncs/nrf/scripts/requirements-fixed.txt --strip-extras ./bootloader/mcuboot/scripts/requirements.txt ./nrf/scripts/requirements-ci.txt ./nrf/scripts/requirements-extra.txt ./nrf/scripts/requirements.txt ./zephyr/scripts/requirements.txt
```
tejlmand: can we avoid invoking `pip-compile` with the full path to the out file ?
Ref: `--output-file=/home/runner/work/sdk-nrf-testing/sdk-nrf-testing/ncs/nrf/scripts/requirements-fixed.txt`

Especially because it's unfortunate to simply have changes to a file simply because a github workflow uses a different path.

DatGizmo: I don't think the path will change in the future.
This change is only highlighted here as I used a locally generated file to trigger the pipeline.
Will check if it is possible to drop the full path and only use relative format

thst-nordic: if it's ok with Torsten, I propose we merge this as is (since the path is generated in a file that isn't in this PR) 

and then look at printing a relative path when we 'backport/upmerge' this to sdk-nrf-next.

thst-nordic:
this is still needed right?  take out of draft?

PR #12633, opened 2023-10-11
net: lib: download_client: Send error event on all failures

SeppoTakalo:
In connection phase, various invalid conditions did not produce similar error event like connection would. Refactor the connection/DNS resolve phase so that all failures produce error event before CLOSED event.
This way fota_download library does not incorrectly translate CLOSED event as a indication of succesfull download.

This is a follow up of https://github.com/nrfconnect/sdk-nrf/pull/12555 which fixed the connection phase. Few more cases need to be dealt the same way.


jheiskan81 (approved):

SeppoTakalo:
Working on test case, I found one more case that was not handled.
Marking as Do not merge, until I have test cases for these corner cases.

SeppoTakalo:
Fixed one possible buffer overrun.
Fixed an issue where DISCONNECT event might have been send twice, first on `connect_client()` and second time from main event loop. With proper timing, this broke some tests.

Modified the module tests so that all events are pushed into a pipe, so we can check correct order of all the events, without missing one.

SeppoTakalo:
Changelog is already modified in https://github.com/nrfconnect/sdk-nrf/pull/12555

juhaylinen (approved):

rlubos (approved):

PR #12635, opened 2023-10-11
wifi: nRF70 enable verbose logging

krish2718:
Should be merged after https://github.com/nrfconnect/sdk-nrf/pull/12634 to avoid Matter failures as this increases memory.


sachinthegreen (approved):

drivers/wifi/nrf700x/osal/os_if/inc/osal_api.h
```diff
@@ -16,7 +16,7 @@
 #include "osal_structs.h"
 
 #ifndef CONFIG_NRF700X_LOG_VERBOSE
-#define __func__ ""
+#define __func__ "<snipped>"
```
rado17, line 19: Wouldn't this add to the binary size?

rado17 (approved):

PR #12640, opened 2023-10-11
Applications: nrf5340_audio: Post upmerge cleanup

koffes:
Minor changes after the upmerge


applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
```diff
@@ -64,7 +64,7 @@ static void periodic_adv_sync(const struct bt_le_scan_recv_info *info, uint32_t
 	struct bt_le_per_adv_sync_param param;
 
 	bt_le_scan_cb_unregister(&scan_callback);
-	cb_registered = false;
+	scan_cb_registered = false;
```
koffes, line 67: Since only the scan callback and not the sync cb is set to false here, I had reservations against combining them.

alexsven: I don't think we should combine them since we unregister the scan_cb but not the sync_cb

alexsven (approved):

erikrobstad (approved):

PR #12641, opened 2023-10-11
doc: update on cellular UDP sample

divipillai:
update on cellular UDP sample

### DIFF
--- a/samples/cellular/udp/README.rst
+++ b/samples/cellular/udp/README.rst
@@ -7,7 +7,7 @@ Cellular: UDP
    :local:
    :depth: 2
 
-The UDP sample demonstrates the sequential transmission of UDP packets to a predetermined server identified by an IP address and a port.
+The Cellular: UDP sample demonstrates the sequential transmission of UDP packets to a predetermined server identified by an IP address and a port.
 The sample uses the :ref:`nrfxlib:nrf_modem` and :ref:`lte_lc_readme` library.
 
 Requirements
@@ -32,7 +32,7 @@ These parameters can be set through the sample configuration file :file:`prj.con
 You can configure the frequency with which the packets are transmitted and the size of the UDP payload through the Kconfig system.
 In addition to setting of the above options, you can also set the various LTE parameters that are related to current consumption for adding low power behavior to the device.
 
-The UDP sample can be used to characterize the current consumption of the nRF91 Series SiP.
+You can use this sample to characterize the current consumption of the nRF91 Series SiP.
 This is due to the simple UDP/IP behavior demonstrated by the sample, which makes it suitable for current measurement.
 
 .. note::


peknis:
@gmarull, the mysterious image compliance failure is also here inspite of the adjusted .checkpatch.conf (https://github.com/nrfconnect/sdk-nrf/pull/12638/files). Please check what is going on.

samples/cellular/udp/README.rst
```diff
@@ -7,7 +7,7 @@ Cellular: UDP
    :local:
    :depth: 2
 
-The UDP sample demonstrates the sequential transmission of UDP packets to a predetermined server identified by an IP address and a port.
+The Cellular: UDP sample demonstrates the sequential transmission of UDP packets to a predetermined server identified by an IP address and a port.
```
peknis, line 10:
```suggestion
This sample demonstrates the sequential transmission of UDP packets to a predetermined server identified by an IP address and a port.
```

samples/cellular/udp/README.rst
```diff
@@ -32,7 +32,7 @@ These parameters can be set through the sample configuration file :file:`prj.con
 You can configure the frequency with which the packets are transmitted and the size of the UDP payload through the Kconfig system.
 In addition to setting of the above options, you can also set the various LTE parameters that are related to current consumption for adding low power behavior to the device.
 
-The UDP sample can be used to characterize the current consumption of the nRF91 Series SiP.
+The Cellular: UDP sample can be used to characterize the current consumption of the nRF91 Series SiP.
```
peknis, line 35:
```suggestion
You can use this sample to characterize the current consumption of the nRF91 Series SiP.
```

peknis (changes requested):

samples/cellular/udp/README.rst
divipillai: The reason to explicitly mention the sample name as cellular: UDP is, we have a similar sample in the networking folder which has Wi-Fi support. https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/net/udp/README.html.

So, I think we could keep this as is?

samples/cellular/udp/README.rst
divipillai: may be : _You can use  Cellular: UDP sample ...._ ?

samples/cellular/udp/README.rst
peknis: I'd say we can still safely say "This sample", as the reader is possibly on this page when reading.

divipillai: Discussed offline, and decided to keep the sample name at the first instance.

peknis (approved):

PR #12647, opened 2023-10-11
samples: crypto: Add EdDSA sample

joerchan:
Add sample that shows EdDSA message signing and verifying using the PureEdDSA algorithm using Twisted Edwards curve.


samples/crypto/eddsa/src/main.c
```diff
@@ -0,0 +1,247 @@
...
+
+	/* Configure the key attributes */
+	psa_set_key_usage_flags(&key_attributes, PSA_KEY_USAGE_SIGN_MESSAGE);
+	psa_set_key_lifetime(&key_attributes, PSA_KEY_LIFETIME_VOLATILE);
+	psa_set_key_algorithm(&key_attributes, PSA_ALG_PURE_EDDSA);
+	psa_set_key_type(&key_attributes, PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_TWISTED_EDWARDS));
+	psa_set_key_bits(&key_attributes, 255);
+
+	/* Generate a random keypair. The keypair is not exposed to the application,
+	 * we can use it to signing/verification the key handle.
```
SebastianBoe, line 105:
```suggestion
	 * we can use it to sign/verify the key handle.
```

samples/crypto/eddsa/src/main.c
```diff
@@ -0,0 +1,247 @@
...
+	/* Export the public key */
+	status = psa_export_public_key(m_key_pair_handle, m_pub_key, sizeof(m_pub_key), &m_pub_key_len);
+	if (status != PSA_SUCCESS) {
+		LOG_INF("psa_export_public_key failed! (Error: %d)", status);
+		return APP_ERROR;
+	}
+
+	PRINT_HEX("Public-key", m_pub_key, m_pub_key_len);
+
+	/* After the key handle is acquired the attributes are not needed */
```
SebastianBoe, line 122:
```suggestion
	/* After the key handle is acquired the attributes are not needed. This may free up resources. */
```

samples/crypto/eddsa/README.rst
```diff
@@ -0,0 +1,55 @@
...
+
+Overview
+********
+
+The sample performs the following operations:
+
+1. Initialization:
+
+   a. The Platform Security Architecture (PSA) API is initialized.
+   #. A random Elliptic Curve Cryptography (ECC) key pair is generated and imported into the PSA crypto keystore.
```
SebastianBoe, line 29: This is misleading. It is only the private key of the key pair that is imported into the keystore at this point no?

The way this (and the below) is written makes it sounds like there are 3 keys in the store. But it's just one pair right?

SebastianBoe (changes requested):

samples/crypto/eddsa/src/main.c
```diff
@@ -0,0 +1,247 @@
...
+	return APP_SUCCESS;
+}
+
+int sign_message(void)
+{
+	psa_status_t status;
+
+	LOG_INF("Signing a message using EDDSA...");
+
+	/* Sign the hash */
```
vili-nordic, line 159: sign the message?

samples/crypto/eddsa/src/main.c
```diff
@@ -0,0 +1,247 @@
...
+	status = psa_sign_message(m_key_pair_handle,
+				  PSA_ALG_PURE_EDDSA,
+				  m_plain_text,
+				  sizeof(m_plain_text),
+				  m_signature,
+				  sizeof(m_signature),
+				  &m_signature_len);
+	if (status != PSA_SUCCESS) {
+		LOG_INF("psa_sign_hash failed! (Error: %d)", status);
```
vili-nordic, line 168: psa_sign_message

samples/crypto/eddsa/src/main.c
```diff
@@ -0,0 +1,247 @@
...
+		return APP_ERROR;
+	}
+
+	LOG_INF("Signed the message successful!");
```
vili-nordic, line 172: broken english?
->successfully

samples/crypto/eddsa/src/main.c
```diff
@@ -0,0 +1,247 @@
...
+
+	/* Verify the signature of the hash */
+	status = psa_verify_message(m_pub_key_handle,
+				    PSA_ALG_PURE_EDDSA,
+				    m_plain_text,
+				    sizeof(m_plain_text),
+				    m_signature,
+				    m_signature_len);
+	if (status != PSA_SUCCESS) {
+		LOG_INF("psa_verify_hash failed! (Error: %d)", status);
```
vili-nordic, line 193: message

samples/crypto/eddsa/src/main.c
vili-nordic: should probably say "message" instead of "key handle" ?

melwee01 (approved):

samples/crypto/eddsa/src/main.c
```diff
@@ -0,0 +1,247 @@
...
+	LOG_INF("Generating random EDDSA keypair...");
+
+	/* Configure the key attributes */
+	psa_key_attributes_t key_attributes = PSA_KEY_ATTRIBUTES_INIT;
+
+	/* Configure the key attributes */
+	psa_set_key_usage_flags(&key_attributes, PSA_KEY_USAGE_SIGN_MESSAGE);
+	psa_set_key_lifetime(&key_attributes, PSA_KEY_LIFETIME_VOLATILE);
+	psa_set_key_algorithm(&key_attributes, PSA_ALG_PURE_EDDSA);
+	psa_set_key_type(&key_attributes, PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_TWISTED_EDWARDS));
```
mswarowsky, line 101: Seems like we need a clang-format on this file to get rid of the long line warnings

samples/crypto/eddsa/README.rst
SebastianBoe: I thought the public key is derived/calculated from the private key?

How would you describe a situation where both a private key and it's public key is imported into the PSA crypto keystore?

mswarowsky: > I thought the public key is derived/calculated from the private key?
> 
Well on key generation you create them as a pair, but given a private key you can derive the public key.


> How would you describe a situation where both a private key and it's public key is imported into the PSA crypto keystore?

That means importing a  key pair.

SebastianBoe: That looks much better.

It is more clear that at the end we will have a private key and it's corresponding public key physically stored in the key store.

joerchan: So, the sample is actually trying to demonstrate being both sides of the usage-case in one single main file.
Having two entries in the keystore is not needed to sign and verify the message.
The sample is showing how to generate the key, and then give it to the other "side".
And how the other "side" takes this key and then use it.

samples/crypto/eddsa/src/main.c
joerchan: It should, also the usage has not been set for verify. So removing that part.

vili-nordic (approved):

samples/crypto/eddsa/boards/nrf9160dk_nrf9160_ns.overlay
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2020 Nordic Semiconductor ASA
```
Vge0rge, line 2: The nits of nits but don't we need to update the year of this?

Vge0rge (approved):

mswarowsky (approved):

samples/crypto/ecdsa/README.rst
```diff
@@ -26,8 +26,9 @@ The sample performs the following operations:
 1. Initialization:
 
    a. The Platform Security Architecture (PSA) API is initialized.
-   #. A random Elliptic Curve Cryptography (ECC) key pair is generated and imported into the PSA crypto keystore.
-   #. The public key of the ECDSA key pair is imported into the PSA crypto keystore.
+   # A random Elliptic Curve Cryptography (ECC) key pair is generated in the PSA crypto keystore for signing usage.
+   # The public key of the ECC key pair is exported to the application.
+   # The exported public key is imported in the PSA crypto key storage for verify usage.
```
mia-ko, line 31:
```suggestion
   #. A random Elliptic Curve Cryptography (ECC) key pair is generated in the PSA crypto keystore for signing purposes.
   #. The public key of the ECC key pair is exported to the application.
   #. The exported public key is imported into the PSA crypto key storage to be used for verification.
```

samples/crypto/eddsa/README.rst
```diff
@@ -0,0 +1,56 @@
...
+********
+
+The sample performs the following operations:
+
+1. Initialization:
+
+   a. The Platform Security Architecture (PSA) API is initialized.
+   # A random Elliptic Curve Cryptography (ECC) key pair is generated in the PSA crypto keystore for signing usage.
+   # The public key of the ECC key pair is exported to the application.
+   # The exported public key is imported in the PSA crypto key storage for verify usage.
```
mia-ko, line 31: Check the previous comment

mia-ko (changes requested):
Remember to add entry in the changelog.
Would also consider full name for samples in the first line for both samples (line 10 in both rst files).
The Elliptic Curve Digital Signature Algorithm (ECDSA) sample...
The Edwards-curve Digital Signature Algorithm (EdDSA) sample...

samples/crypto/ecdsa/README.rst
joerchan: usage is the PSA specification word used.
https://armmbed.github.io/mbed-crypto/html/api/keys/policy.html#key-usage-flags

mia-ko: Glancing through it quickly, it seems usage is mentioned when talking about usage flags.
Should these sentences rather mention usage flags instead then?
E.g.: A random Elliptic Curve Cryptography (ECC) key pair is generated in the PSA crypto keystore with the signing usage flag.
Or verfication usage flag for that later sentence.
But if not mentioning the usage flags, it does not seem to me like usage is needed in the sentence.

mia-ko (approved):

PR #12650, opened 2023-10-11
manifest: zephyr: Pull mesh fixes for OOB upload

PavelVPV:
Pull mesh fixes for OOB upload.

PR #12652, opened 2023-10-11

anangl:
Pull in a few upstream fixes that should be included in NCS 2.5.0.


anangl:
No longer needed. SHA will be updated to a newer revision by #12650.

PR #12664, opened 2023-10-12
workflow: Increase the speed of compliance workflow by doing narrow west update

balaji-nordic:
Change compliance workflow to do a narrow west update so that the
workflow runs faster.

It now takes roughly 1.5 to 2.5 minutes to complete instead of about 3.5 to 5.5 minutes

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -50,8 +50,7 @@ jobs:
         # debug
         git log  --pretty=oneline | head -n 10
         west init -l .
-        west update
-        west zephyr-export
+        west update --narrow -o=--depth=1
 
     - name: Run CODEOWNERS test
       id: codeowners


thst-nordic (approved):

.github/workflows/compliance.yml
```diff
@@ -50,8 +50,7 @@ jobs:
         # debug
         git log  --pretty=oneline | head -n 10
         west init -l .
-        west update
-        west zephyr-export
```
thst-nordic, line 54: what is this used for?  not required anymore?

balaji-nordic: I dont know why we did `west zephyr-export` inside the compliance workflow. It is only used to register current zephyr installation as a cmake configuration package. And we dont 'build' any samples in compliance workflow. Essentially rendering this useless.

simensrostad (approved):

.github/workflows/compliance.yml
carlescufi: Agreed

carlescufi (approved):

.github/workflows/compliance.yml
```diff
@@ -50,8 +50,7 @@ jobs:
...
+        west update --narrow -o=--depth=1
```
carlescufi, line 53: I'm afraid this will make the oss-history check fail, because it looks for a merge-base between the current manifest-rev and upstream. This requires having the history.

carlescufi (changes requested):

.github/workflows/compliance.yml
balaji-nordic: I see. Is the oss-history check part of compliance? Isn't that done in a separate workflow? namely the "OSS history check" which passed for this PR?

carlescufi: Ah sorry, that one has its own cloning of the repos.

carlescufi (approved):

PR #12665, opened 2023-10-12
net: lib: nrf_cloud: Remove QZSS from A-GNSS request with CoAP

tokangas:
nRF Cloud does not yet support QZSS GNSS assistance over CoAP. QZSS data needed to be removed from the assistance request, otherwise the cloud returned an error.

`CONFIG_COAP_EXTENDED_OPTIONS_LEN_VALUE` needed to be increased for modem_shell, otherwise CoAP registration always failed.


trantanen (approved):

jhirsi (approved):

jayteemo (approved):
I asked the coap backend team what was missing for qzss support and am waiting on the response.

tokangas:
> I asked the coap backend team what was missing for qzss support and am waiting on the response.

Thanks. Not sure if changes would be needed on the device side or not to make QZSS work with CoAP. The request is successfully sent and the response would be processed using the same function as with REST or MQTT, so it might be that no device side changes are necessary.

PR #12667, opened 2023-10-12
Applications: nrf5340_audio: DFU doc added

koffes:
Added note of prints during DFU

### DIFF
--- a/applications/nrf5340_audio/README.rst
+++ b/applications/nrf5340_audio/README.rst
@@ -350,6 +350,11 @@ For example:
 
 The first part of these names is based on :kconfig:option:`CONFIG_BT_DEVICE_NAME`.
 
+.. note::
+   When performing DFU for the nRF5340 Audio application, there will be one or more error prints related to opening flash area ID 1.
+   This is due to restrictions in the DFU system, and the error print is expected.
+   The DFU process should still complete successfully.
+
 .. _nrf53_audio_app_building:
 
 Building and running


rick1082 (approved):

applications/nrf5340_audio/README.rst
```diff
@@ -352,6 +352,11 @@ The first part of these names is based on :kconfig:option:`CONFIG_BT_DEVICE_NAME
 
 .. _nrf53_audio_app_building:
 
+Note
```
erikrobstad, line 355: This note should be above ".. _nrf53_audio_app_building:"

applications/nrf5340_audio/README.rst
```diff
@@ -352,6 +352,11 @@ The first part of these names is based on :kconfig:option:`CONFIG_BT_DEVICE_NAME
...
+----
+Due to restrictions in the DFU system, when performing an audio DFU, there will be prints: "Failed to open flash area...".
```
erikrobstad, line 357: I suggest rewriting to something like this:

When performing DFU for the nRF5340 Audio application, there will be one error print related to opening flash area ID 1.
This is due to restrictions in the DFU system, and the error print is expected. The DFU process should still complete successfully.

erikrobstad (changes requested):

applications/nrf5340_audio/README.rst
Frodevan: The error print can happen multiple times if the DFU tool used tries to do certain actions several times, so maybe something like  "one or more error prints may be logged" instead of "there will be one error print"

applications/nrf5340_audio/README.rst
koffes: Good catch.

applications/nrf5340_audio/README.rst
koffes: @Frodevan what about now?

cvinayak:
Dont we need  a tech writer approval here?

PR #12668, opened 2023-10-12
Bluetooth: mesh: Add model correspond relation

HaavardRei:
Adds correspond relationship to models as specified in MshMDL1.1. Requires the KCONFIG CONFIG_BT_MESH_COMP_PAGE_1
to register correspondence.

Adds an init callback for Generic Location Setup Server to register model extension and correspondence to Generic Location Server.

Adds a field to the 'bt_mesh_light_temp_srv' struct in 'light_temp_srv.h' which contains the address of the corresponding Light CTL server (used to register correspondence).

Signed-off-by: Håvard Reierstad <haavard.reierstad@nordicsemi.no>

### DIFF
--- a/subsys/bluetooth/mesh/gen_loc_srv.c
+++ b/subsys/bluetooth/mesh/gen_loc_srv.c
@@ -259,6 +259,24 @@ const struct bt_mesh_model_cb _bt_mesh_loc_srv_cb = {
 	.reset = bt_mesh_loc_srv_reset,
 };
 
+static int bt_mesh_loc_setup_srv_init(struct bt_mesh_model *model)
+{
+	struct bt_mesh_loc_srv *srv = model->user_data;
+#if defined(CONFIG_BT_MESH_COMP_PAGE_1)
+	int err = bt_mesh_model_correspond(model, srv->model);
+
+	if (err) {
+		return err;
+	}
+#endif
+
+	return bt_mesh_model_extend(model, srv->model);
+}
+
+const struct bt_mesh_model_cb _bt_mesh_loc_setup_srv_cb = {
+	.init = bt_mesh_loc_setup_srv_init,
+};
+
 int bt_mesh_loc_srv_global_pub(struct bt_mesh_loc_srv *srv,
 			       struct bt_mesh_msg_ctx *ctx,
 			       const struct bt_mesh_loc_global *global)
--- a/subsys/bluetooth/mesh/light_hue_srv.c
+++ b/subsys/bluetooth/mesh/light_hue_srv.c
@@ -448,6 +448,7 @@ static int hue_srv_pub_update(struct bt_mesh_model *model)
 static int hue_srv_init(struct bt_mesh_model *model)
 {
 	struct bt_mesh_light_hue_srv *srv = model->user_data;
+	int err;
 
 	srv->model = model;
 	srv->pub.update = hue_srv_pub_update;
@@ -459,14 +460,23 @@ static int hue_srv_init(struct bt_mesh_model *model)
 	srv->emds_entry.entry.id = EMDS_MODEL_ID(model);
 	srv->emds_entry.entry.data = (uint8_t *)&srv->transient;
 	srv->emds_entry.entry.len = sizeof(srv->transient);
-	int err = emds_entry_add(&srv->emds_entry);
+	err = emds_entry_add(&srv->emds_entry);
 
 	if (err) {
 		return err;
 	}
 #endif
 
-	return bt_mesh_model_extend(model, srv->lvl.model);
+#if defined(CONFIG_BT_MESH_COMP_PAGE_1)
+	err = bt_mesh_model_correspond(srv->hsl->model, model);
+
+	if (err) {
+		return err;
+	}
+#endif
+
+	err = bt_mesh_model_extend(model, srv->lvl.model);
+	return err;
 }
 
 static int hue_srv_settings_set(struct bt_mesh_model *model, const char *name,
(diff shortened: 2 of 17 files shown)


subsys/bluetooth/mesh/gen_loc_srv.c
```diff
@@ -259,6 +259,24 @@ const struct bt_mesh_model_cb _bt_mesh_loc_srv_cb = {
 	.reset = bt_mesh_loc_srv_reset,
 };
 
+static int bt_mesh_loc_setup_srv_init(struct bt_mesh_model *model)
+{
+	struct bt_mesh_loc_srv *srv = model->user_data;
+#if IS_ENABLED(CONFIG_BT_MESH_COMP_PAGE_1)
```
PavelVPV, line 265: Please avoid using `IS_ENABLED` with `#if`:
https://github.com/zephyrproject-rtos/zephyr/blob/e6197fed5fc9001023fb3d90ec41cbee51594c0c/include/zephyr/sys/util_macro.h#L117-L118

subsys/bluetooth/mesh/light_hue_srv.c
```diff
@@ -459,10 +459,18 @@ static int hue_srv_init(struct bt_mesh_model *model)
 	srv->emds_entry.entry.id = EMDS_MODEL_ID(model);
 	srv->emds_entry.entry.data = (uint8_t *)&srv->transient;
 	srv->emds_entry.entry.len = sizeof(srv->transient);
-	int err = emds_entry_add(&srv->emds_entry);
+	int err_emds = emds_entry_add(&srv->emds_entry);
```
PavelVPV, line 462: The same variable can be used for error checking. I think you can use `(void)err;` to avoid warning.

subsys/bluetooth/mesh/gen_loc_srv.c
```diff
@@ -259,6 +259,24 @@ const struct bt_mesh_model_cb _bt_mesh_loc_srv_cb = {
...
+	int err = bt_mesh_model_correspond(model, srv->model);
+
+	if (err) {
+		return err;
+	}
+#endif
+
+	return bt_mesh_model_extend(model, srv->model);
```
PavelVPV, line 273: Good catch!

subsys/bluetooth/mesh/gen_loc_srv.c
HaavardRei: Right, so would be better to use `#if (is_defined(CONFIG_BT_MESH_COMP_PAGE_1) && CONFIG_BT_MESH_COMP_PAGE_1==1))`?

subsys/bluetooth/mesh/light_hue_srv.c
HaavardRei: The problem was that the err related to emds is inside of a #if IS_ENABLED, which means that it might not be declared and thus reusable in the #if statement related to correspondence. I tried moving the err declaration outside of the #ifs, which results in a warning in the case none of the #ifs are true.

subsys/bluetooth/mesh/gen_loc_srv.c
PavelVPV: `#if defined(CONFIG_BT_MESH_COMP_PAGE_1)` should work.

subsys/bluetooth/mesh/light_hue_srv.c
Andrewpini: Not sure if this is an ideal fix to this issue, but what if you change the lats return statement of this function to:

```
err = bt_mesh_model_extend(model, srv->lvl.model);
return err;
```

and declare `int err` at the top of the function? I think that would let you use a single err variable at least.

subsys/bluetooth/mesh/gen_loc_srv.c
Andrewpini: What about
```
if (IS_ENABLED(CONFIG_BT_MESH_COMP_PAGE_1)) {
...
}
```
? Always thought it was the cleaner option.

subsys/bluetooth/mesh/light_hue_srv.c
HaavardRei: Hmm, the problem is only present in `light_sat_srv.c`, `light_hue_srv.c` and `light_temp_srv.c`, so I think Anders' suggestion looks more clean and should be used in these cases. Thoughts @PavelVPV ?

PavelVPV: Sounds good

subsys/bluetooth/mesh/gen_loc_srv.c
HaavardRei: Should these be changed as well to avoid using `IS_ENABLED`? https://github.com/nrfconnect/sdk-nrf/blob/baf6159e6140031d8bc3085b3ccc65a529cf9e13/subsys/bluetooth/mesh/light_sat_srv.c#L389

HaavardRei: I went for the "#if defined()" approach, just changing those related to CDP1.

subsys/bluetooth/mesh/light_hue_srv.c
HaavardRei: Fixed now!

subsys/bluetooth/mesh/gen_loc_srv.c
PavelVPV: > Should these be changed as well to avoid using `IS_ENABLED`?
> 
> https://github.com/nrfconnect/sdk-nrf/blob/baf6159e6140031d8bc3085b3ccc65a529cf9e13/subsys/bluetooth/mesh/light_sat_srv.c#L389

If you want, you can fix this. In a separate commit or separate PR.

PavelVPV (approved):

Andrewpini (approved):

cvinayak:
@akredalen waiting for your response...

PR #12670, opened 2023-10-12
applications: serial_lte_modem: fix FTP data reception

tomi-font:
Fix a too early return of ftp_client lib functions. They now return only when the data transfer is complete.

Additionally, feed the received data to UART in chunks to allow receiving data bigger than the size of the UART buffer (2048 bytes) *if* the size of the ring buffer (ftp_data_buf) is increased, because it is currently the limiting factor.

NCSDK-20213

### DIFF
--- a/applications/serial_lte_modem/doc/FTP_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/FTP_AT_commands.rst
@@ -49,6 +49,12 @@ The ``<cmd>`` command is a string, and can be used as follows:
 The values of the parameters depend on the command string used.
 When using the ``put``, ``uput`` and ``mput`` commands, if the ``<data>`` attribute is not specified, SLM enters ``slm_data_mode``.
 
+.. note::
+   The maximum size of the data that can be received when using the commands ``ls`` and ``get`` is limited by the size of the internal FTP buffer, which is 2048 bytes.
+   The size can be adjusted by modifying the declaration of ``ftp_data_buf`` in :file:`src/ftp_c/slm_at_ftp.c`.
+
+   If more data than can be handled is received after issuing an ``ls`` or ``get`` command, the data will be truncated to the maximum supported size.
+
 Response syntax
 ~~~~~~~~~~~~~~~
 
--- a/applications/serial_lte_modem/src/ftp_c/slm_at_ftp.c
+++ b/applications/serial_lte_modem/src/ftp_c/slm_at_ftp.c
@@ -111,7 +111,8 @@ void ftp_ctrl_callback(const uint8_t *msg, uint16_t len)
 	char code_str[4];  /* Proprietary code 900 ~ 999 */
 	int code;
 
-	strncpy(code_str, msg, 3);
+	strncpy(code_str, msg, sizeof(code_str) - 1);
+	code_str[sizeof(code_str) - 1] = '\0';
 	code = atoi(code_str);
 	if (FTP_PROPRIETARY(code)) {
 		switch (code) {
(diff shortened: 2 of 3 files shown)


applications/serial_lte_modem/src/ftp_c/slm_at_ftp.c
```diff
@@ -157,26 +158,22 @@ void ftp_ctrl_callback(const uint8_t *msg, uint16_t len)
...
 
-static int ftp_data_send(void)
+static void ftp_data_send(void)
 {
-	uint32_t sz_send = 0;
+	uint32_t sz_send;
 
-	if (ring_buf_is_empty(&ftp_data_buf) == 0) {
-		sz_send = ring_buf_get(&ftp_data_buf, slm_data_buf, sizeof(slm_data_buf));
+	while (sz_send = ring_buf_get(&ftp_data_buf, slm_data_buf, sizeof(slm_data_buf))) {
```
MarkusLassila, line 174: Twister fails with warning, which is interpreted as error:

15:43:53  ../../../../nrf/applications/serial_lte_modem/src/ftp_c/slm_at_ftp.c:174:16: error: suggest parentheses around assignment used as truth value [-Werror=parentheses]
15:43:53    174 |         while (sz_send = ring_buf_get(&ftp_data_buf, slm_data_buf, sizeof(slm_data_buf))) {
15:43:53        |

subsys/net/lib/ftp_client/src/ftp_client.c
```diff
@@ -739,13 +751,7 @@ int ftp_get(const char *file)
 		return -EIO;
 	}
 
-	/* Set up data connection */
-	k_work_submit_to_queue(&ftp_work_q, &data_task_param.work);
-
-	ret = poll_data_task_done();
-	if (ret == FTP_CODE_226) {
```
MarkusLassila, line 746: ret can be something else as well, should this still be removed?

subsys/net/lib/ftp_client/src/ftp_client.c
```diff
@@ -346,26 +348,31 @@ static void do_ftp_recv_data(const char *pasv_msg)
...
 {
-	int ret = -ENODATA;
-	int wait_time = 0;
+	int ret = 0;
+	unsigned int wait_time_ms = 0;
 
-	do {
-		ret = do_ftp_recv_ctrl(true, FTP_CODE_226);
-		if (ret == FTP_CODE_226) {
+	while (ret >= 0 || ret == -ETIMEDOUT) {
```
MarkusLassila, line 354: Just a note that this loop had a lot going on. I would have perhaps considered separating the data task running modification to it's own while loop, even if it would add some extra code, it would likely be easier to follow.

SeppoTakalo (approved):

MarkusLassila (approved):

applications/serial_lte_modem/doc/FTP_AT_commands.rst
```diff
@@ -49,6 +49,12 @@ The ``<cmd>`` command is a string, and can be used as follows:
 The values of the parameters depend on the command string used.
 When using the ``put``, ``uput`` and ``mput`` commands, if the ``<data>`` attribute is not specified, SLM enters ``slm_data_mode``.
 
+.. note::
+   The maximum size of the data that can be received when using the commands "ls" and "get" is limited by the size of the internal FTP buffer, which is 2048 bytes.
```
wiba-nordic, line 53:
```suggestion
   The maximum size of the data that can be received when using the commands ``ls`` and ``get`` is limited by the size of the internal FTP buffer, which is 2048 bytes by default.
```

applications/serial_lte_modem/doc/FTP_AT_commands.rst
```diff
@@ -49,6 +49,12 @@ The ``<cmd>`` command is a string, and can be used as follows:
...
+   It is ``ftp_data_buf``, defined in :c:file:`src/ftp_c/slm_at_ftp.c`.
```
wiba-nordic, line 54:
```suggestion
   The size of the buffer can be changed using ``ftp_data_buf``, defined in :c:file:`src/ftp_c/slm_at_ftp.c`.
```

applications/serial_lte_modem/doc/FTP_AT_commands.rst
```diff
@@ -49,6 +49,12 @@ The ``<cmd>`` command is a string, and can be used as follows:
...
+
+   If more data than can be handled is received after issuing an "ls" or "get" command, it will be truncated to the maximum supported size.
```
wiba-nordic, line 56:
```suggestion
   If more data than can be handled is received after issuing an ``ls`` or ``get`` command, the data will be truncated to the maximum supported size.
```

wiba-nordic (changes requested):

MarkusLassila (approved):

wiba-nordic (approved):

cvinayak:
do you have a issue number or do we need one for this bugfix?

cvinayak:
> > do you have a issue number or do we need one for this bugfix?
> 
> NCSDK-20213 Is an issue number required?

Helps me to be sure when merging PRs during the feature freeze week.

PR #12677, opened 2023-10-12
net: lib: nrf_cloud: Remove benign shadow error

plskeggs:
An error was logged by mistake when the initial non-delta shadow update is processed.

Jira: IRIS-7314

### DIFF
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
@@ -381,18 +381,15 @@ static int cc_rx_data_handler(const struct nct_evt *nct_evt)
 
 	err = nrf_cloud_device_control_update(&nct_evt->param.cc->data, &msg.data,
 					      &status);
-	if (!err && msg.data.ptr) {
-		if ((status == NRF_CLOUD_CTRL_REPLY) || (status == NRF_CLOUD_CTRL_REJECT)) {
-			LOG_DBG("Confirming shadow: %s",
-				(const char *)msg.data.ptr);
-			err = nct_cc_send(&msg);
-			nrf_cloud_free((void *)msg.data.ptr);
-			if (err) {
-				LOG_ERR("nct_cc_send failed %d", err);
-			}
-		}
-	} else {
+	if (err) {
 		LOG_ERR("Error updating device control: %d", err);
+	} else if (msg.data.ptr) {
+		LOG_DBG("Confirming shadow: %s", (const char *)msg.data.ptr);
+		err = nct_cc_send(&msg);
+		nrf_cloud_free((void *)msg.data.ptr);
+		if (err) {
+			LOG_ERR("nct_cc_send failed %d", err);
+		}
 	}
 
 	struct nrf_cloud_evt cloud_evt = {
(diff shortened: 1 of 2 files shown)


subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
```diff
@@ -381,18 +381,17 @@ static int cc_rx_data_handler(const struct nct_evt *nct_evt)
...
-			err = nct_cc_send(&msg);
-			nrf_cloud_free((void *)msg.data.ptr);
-			if (err) {
-				LOG_ERR("nct_cc_send failed %d", err);
-			}
-		}
-	} else {
+	if (err) {
 		LOG_ERR("Error updating device control: %d", err);
+	} else if (msg.data.ptr &&
```
jayteemo, line 386: `} else if ((status == NRF_CLOUD_CTRL_REPLY) || (status == NRF_CLOUD_CTRL_REJECT)) {`
is probably sufficient?

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
jayteemo, line 1: why enable DBG by default?

jayteemo (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
glarsennordic: Maybe the other way around, check for msg.data.ptr and don't check status? It seems to me there is a message we need to forward to cloud (and eventually free) if and only if msg.data.ptr is not NULL. But possibly I am misunderstanding.

glarsennordic (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
jayteemo: the status values indicate if the data should be sent to the cloud.
the ptr should only be valid for those 2 cases.
i was just trying to simplify the `if` a bit... but it really isn't worth messing with.

subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
```diff
@@ -381,18 +381,17 @@ static int cc_rx_data_handler(const struct nct_evt *nct_evt)
...
+		   ((status == NRF_CLOUD_CTRL_REPLY) || (status == NRF_CLOUD_CTRL_REJECT))) {
+		LOG_DBG("Confirming shadow: %s",
```
jayteemo, line 388:
```suggestion
		LOG_DBG("Confirming shadow: %s", (const char *)msg.data.ptr);
```

subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
glarsennordic: IMO it should just be a NULL check against msg.data.ptr -- yes, that can only ever happen if `(status == NRF_CLOUD_CTRL_REPLY) || (status == NRF_CLOUD_CTRL_REJECT)`, but its up to `nrf_cloud_device_control_update`, not this function, to determine whether a message should be sent, right? I think it makes sense for this function to just check if there is a message waiting to be sent. 

I agree it's not a big deal, but I do think its worth at least bringing to Pete's attention, so I am unresolving so he can see it.

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
plskeggs: Gah.  I made this change for unrelated reasons and committed it by mistake.  I'll remove.

subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
plskeggs: OK, guys, yes, it could be simpler.  I did it in 10 minutes before going to have fun.  I'll pick one of your options and use it when I remove the mistaken commit of overlay_coap.conf.

tokangas (approved):

PR #12678, opened 2023-10-13
application: serial_lte_modem: Add #XMQTTCFG command

junqingzou:
Add a new command to configure MQTT client before connecting to a broker, enabling dynamic tuning of Keep Alive time, etc.

JIRA ticket: NCSDK-23815

### DIFF
--- a/applications/serial_lte_modem/doc/MQTT_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/MQTT_AT_commands.rst
@@ -48,6 +48,112 @@ Syntax
   * *Negative value* - Failure.
     It is the error code indicating the reason for the failure.
 
+MQTT configure #XMQTTCFG
+========================
+
+The ``#XMQTTCFG`` command allows you to configure the MQTT client before connecting to a broker.
+
+Set command
+-----------
+
+The set command allows you to configure the MQTT client.
+
+Syntax
+~~~~~~
+
+::
+
+   AT#XMQTTCFG=<client_id>[,<keep_alive>[,<clean_session>]]
+
+* The ``<client_id>`` parameter is a string.
+  It indicates the MQTT Client ID.
+  If this command is not issued, SLM uses the default value of ``slm_default_client_id``.
+* The ``<keep_alive>`` parameter is an integer.
+  It indicates the maximum Keep Alive time in seconds for MQTT.
+  The default Kepp Alive time is 60 seconds.
+* The ``<clean_session>`` parameter is an integer.
+  It can have one of the following values:
+
+    * ``0`` - Connect to a MQTT broker using a persistent session.
+    * ``1`` - Connect to a MQTT broker using a clean session.
+
+  The default is using a persistent session.
+
+Examples
+~~~~~~~~
+
+::
+
+   AT#XMQTTCFG="MyMQTT-Client-ID",300,1
+   OK
+
+Read command
+------------
+
+The read command shows MQTT client configuration information.
+
+Syntax
+~~~~~~
+
+::
+
+   AT#XMQTTCFG?
+
+Response syntax
+~~~~~~~~~~~~~~~
+
+::
+
+   #XMQTTCFG: <client_id>,<keep_alive>,<clean_session>
+
+* The ``<client_id>`` parameter is a string.
+  It indicates the MQTT Client ID.
+* The ``<keep_alive>`` parameter is an integer.
+  It indicates the maximum Keep Alive time in seconds for MQTT.
+* The ``<clean_session>`` parameter is an integer.
+  It can have one of the following values:
+
+    * ``0`` - Connect to a MQTT broker using a persistent session.
+    * ``1`` - Connect to a MQTT broker using a clean session.
+
+Examples
+~~~~~~~~
+
+::
+
+   AT#XMQTTCFG?
+   #XMQTTCFG: "MyMQTT-Client-ID",60,0
+   OK
+
+Test command
+------------
+
+The test command tests the existence of the command and provides information about the type of its subparameters.
+
+Syntax
+~~~~~~
+
+::
+
+   #XMQTTCFG=?
+
+Response syntax
+~~~~~~~~~~~~~~~
+
+::
+
+   #XMQTTCFG: <client_id>,<keep_alive>,<clean_session>
+
+Examples
+~~~~~~~~
+
+::
+
+   AT#XMQTTCFG=?
+   #XMQTTCFG: <client_id>,,<keep_alive>,<clean_session>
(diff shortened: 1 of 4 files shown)


junqingzou:
DNM: not for NCS v2.5.0, no review required as of now

wiba-nordic (changes requested):
Should also add a changelog entry for adding the new command.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -459,6 +473,55 @@ static int do_mqtt_subscribe(uint16_t op,
 	return err;
 }
 
+/* Handles AT#XMQTTCFG commands. */
+int handle_at_mqtt_config(enum at_cmd_type cmd_type)
```
MarkusLassila, line 477: @junqingzou: Since we have moved from dynamic configuration to respecting the static configuration in UART as well as with GNSS (well, not fully), I have to ask, what is the value for all customers in allowing dynamic configuration for CONFIG_MQTT_KEEPALIVE and CONFIG_MQTT_CLEAN_SESSION?

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
```diff
@@ -459,6 +473,55 @@ static int do_mqtt_subscribe(uint16_t op,
...
+
+	default:
+		break;
+	}
+
+	return err;
+}
+
 /* Handles AT#XMQTTCON commands. */
 int handle_at_mqtt_connect(enum at_cmd_type cmd_type)
```
MarkusLassila, line 526: What was the reasoning for not having the parameters in the MQTTCON?

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: It's out of the scope of the PR but I could explain a bit. It was actually what I used in testing against the Chinese MQTT broker to find the best time of Keep Alive that would not lose TCP connection in China Mobile network. Each carrier network has different firewall and NAT TCP binding so one  static Keep Alive will not work in all networks. Furthermore, MQTT broker has so-called max keep-alive parameter, which is not fetchable in MQTT 3.1.x but can in MQTT 5.x. Dynamic MQTT keep-alive on the client side allows adjusting for both needs of above.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
junqingzou: ?

MarkusLassila: Keepalive and clean session parameters could also have been added to MQTTCON, and be in effect for a particular connection. Instead they are set in separate AT-command and are in effect until the device is reset.

junqingzou: In my opinion #XMQTTCON should only ask for parameter as in MQTT URI scheme.
But if you think that is better, I can close this PR for a new one.

applications/serial_lte_modem/src/mqtt_c/slm_at_mqtt.c
MarkusLassila: Wen't through the usage of CONFIG_MQTT_KEEPALIVE and CONFIG_MQTT_CLEAN_SESSION and beyond using hard coded values in nrf_cloud, there did not seem to be any reason why this could not be allowed.

junqingzou: Zephyr MQTT "hardcode" [here](https://github.com/nrfconnect/sdk-zephyr/blob/main/subsys/net/lib/mqtt/mqtt.c#L179) and expect app to modify after init().

MarkusLassila (changes requested):

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -48,6 +48,112 @@ Syntax
   * *Negative value* - Failure.
```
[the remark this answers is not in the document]
junqingzou: The Will message was left over from previous PR#12441, removed there and forgot to remove here.
Was testing with the two PR combined.

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -48,6 +48,112 @@ Syntax
...
     It is the error code indicating the reason for the failure.
 
+MQTT configure #XMQTTCFG
+========================
+
+The ``#XMQTTCON`` command allows you to configure the MQTT client before connecting to a broker.
```
wiba-nordic, line 54:
```suggestion
The ``#XMQTTCFG`` command allows you to configure the MQTT client before connecting to a broker.
```

wiba-nordic (changes requested):

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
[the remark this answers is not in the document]
MarkusLassila: @junqingzou: Please combine the commits so that there is no commit that adds will message and a second one that immediately removes it.

MarkusLassila (changes requested):

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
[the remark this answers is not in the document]
junqingzou: Sorry, I missed that.
Should have not used old code base.

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
junqingzou: Great catch 🤦‍♂️

applications/serial_lte_modem/doc/MQTT_AT_commands.rst
```diff
@@ -106,7 +210,7 @@ Examples
 
 ::
 
-   AT#XMQTTCON=1,"MyMQTT-Client-ID","","","mqtt.server.com",1883
+   AT#XMQTTCON=1,"","","mqtt.server.com",1883
```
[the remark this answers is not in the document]
junqingzou: There is a line in `AT#XMQTTCFG`

wiba-nordic (approved):

MarkusLassila (approved):

PR #12679, opened 2023-10-13
doc: nrf: Added information about nRF7002 Amazon FFS certification

kkasperczyk-no:
The nRF7002 DK passed Amazon FFS certification for Matter, so the table containing platforms should be updated.


wiba-nordic (approved):

LuDuda (approved):
LGTM :+1:

PR #12689, opened 2023-10-13
Enable support in Zigbee Fota for three upgradeable images

sebastiandraus:


sigvartmh:
I'm a bit curious about this. We don't have a lot of handling of >2 images in partition manager so how does this system work?

sebastiandraus:
> I'm a bit curious about this. We don't have a lot of handling of >2 images in partition manager so how does this system work?

This is related to https://github.com/NordicPlayground/ncs-nrf5340-xip-app/pull/10, when XiP is enabled application uses three images and these changes are required to have ZIgbee FOTA working with 3 images (like when XiP is being used)

tomchy (approved):

subsys/zigbee/lib/zigbee_fota/src/dfu_multi_target.c
```diff
@@ -54,6 +57,15 @@ static struct dfu_image_writer dfu_multi_target_writer_1 = {
 };
 #endif
 
+#if CONFIG_UPDATEABLE_IMAGE_NUMBER > 2
+/** DFU target writer for image 1 (NET core). */
```
LuDuda, line 61:
```suggestion
/** DFU target writer for image 2 (external memory). */
```

LuDuda (approved):

subsys/zigbee/lib/zigbee_fota/src/dfu_multi_target.c
sebastiandraus: Small change in comments pushed, please verify :)

nordicjm: Above seems unfixed

nordicjm (commented):
Will +1 once third commit is merged into second

nordicjm (approved):

sigvartmh (approved):

sebastiandraus:
Force-pushed to trigger CI

PR #12692, opened 2023-10-13
samples: cellular: http_update: Fix Carrier Library integration

nordic-mare:
The carrier library integration was broken as the link controller would never connect with carrier library enabled


eivindj-nordic (approved):
Thanks!

kacperradoszewski (approved):

nordic-hani (approved):

PR #12694, opened 2023-10-13

anangl:
Pull in a fix in the new way of configuring the NFC and nRESET pins via the uicr node in devicetree.


gmarull (approved):

anangl:
No longer needed. Manifest will be updated accordingly by #12700.

PR #12707, opened 2023-10-16
tfm: Kconfig: Enable TF-M logging by default also with minimal conf

SebastianBoe:
Stop enabling TFM_LOG_LEVEL_SILENCE from the TFM_PROFILE_TYPE_MINIMAL option.

Without logging it becomes more difficult to debug failures, both in the non-secure image (secure faults), and in TF-M.

I assume we disabled logging at some point due to the 32kB resource usage limit. But my testing shows that we have space for this now, presumably due to some optimization work that we have done.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -464,7 +464,10 @@ zcbor
 Trusted Firmware-M
 ==================
 
-|no_changes_yet_note|
+* The minimal TF-M build profile no longer silences TF-M logs by default.
+
+  .. note::
+     This can be a breaking change if the UART instance used by TF-M is already in use, for example by modem trace with a UART backend.
 
 cJSON
 =====
--- a/modules/trusted-firmware-m/Kconfig
+++ b/modules/trusted-firmware-m/Kconfig
@@ -82,7 +82,6 @@ choice TFM_PROFILE_TYPE
 config TFM_PROFILE_TYPE_MINIMAL
 	bool "Use minimal TF-M build"
 	select PSA_DEFAULT_OFF
-	imply TFM_LOG_LEVEL_SILENCE
 	help
 	  Use minimal TF-M build. This will make the TF-M image fit within 32kB.
 	  No configuration of the size of the partition nor the features of the


joerchan:
> I assume we disabled logging at some point due to the 32kB resource usage limit. But my testing shows that we have space for this now, presumably due to some optimization work that we have done.

Originally it was excluded, but added as an option later.
Changing the default to be included makes sense. Although it can have an impact on projects that use uart1 for other use-cases.
An example of this is Modem Trace.

This should be added to release notes at the very least for that reason.

mswarowsky (approved):
With the requested change for a change log entry, from Joakim, fine for me

SebastianBoe:
> Changing the default to be included makes sense. Although it can have an impact on projects that use uart1 for other use-cases.
> An example of this is Modem Trace.
> 
> 

I was hoping CI would demonstrate this, but it did not.

I see that there are not many plans being run.

I will touch the manifest and see if I can get this to fail.

SebastianBoe:
touching the manifest doesn't work anymore :( Trying with a github label

lemrey (approved):
Thanks

joerchan (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -451,7 +451,11 @@ zcbor
 Trusted Firmware-M
 ==================
 
-|no_changes_yet_note|
+The minimal TF-M profile no longer silences TF-M logs by default.
+
+.. note::
+   This can be a breaking change if the UART instance used by TF-M is
+   already in use elsewhere, e.g. in modem trace with a UART backend.
```
melwee01, line 458:
```suggestion
   This can be a breaking change if the UART instance used by TF-M is
   already in use elsewhere, for example, in modem trace with a UART backend.
```

melwee01 (changes requested):
As a general rule, no Latin abbreviations (etc., e.g., i.e.)

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
mia-ko, line 458:
```suggestion
* The minimal TF-M build profile no longer silences TF-M logs by default.

  .. note::
     This can be a breaking change if the UART instance used by TF-M already is in use, for example by modem trace with a UART backend.
```

mia-ko (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
SebastianBoe: I wasn't able to apply this as it conflicted with Mel's feedback. But I recreated it to the best of my ability.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -451,7 +451,10 @@ zcbor
 Trusted Firmware-M
 ==================
 
-|no_changes_yet_note|
+* The minimal TF-M build profile no longer silences TF-M logs by default.
+
+  .. note::
+     This can be a breaking change if the UART instance used by TF-M already is in use, for example by modem trace with a UART backend.
```
melwee01, line 457:
```suggestion
     This can be a breaking change if the UART instance used by TF-M is already in use, for example by modem trace with a UART backend.
```

mia-ko (approved):

melwee01 (approved):

PR #12764, opened 2023-10-18
Bluetooth: mesh: composition data fixes

HaavardRei:
Adds the (already implemented) scheduler setup server callback to the `BT_MESH_MODEL_SCHEDULER_SRV` macro.

Reorders property models in Bluetooth tester to register composition data correctly.


ludvigsj:
If a user creates an application with these in the wrong order, will something go wrong? If so, is this documented anywhere?

HaavardRei:
@ludvigsj The problem lies with how the model extension (Admin extending User) is registered, and is documented here: https://github.com/zephyrproject-rtos/zephyr/blob/6076d54d079ca6a3f40f191f826e9bd0b620a0ec/include/zephyr/bluetooth/mesh/access.h#L1073-L1075

The User Property Server needs to be initialized before the Admin Server is due to the initialization of the Admin Server calling `bt_mesh_model_extend`

ludvigsj (approved):

alxelax (approved):

PR #12767, opened 2023-10-18
nrf_security: Always enabled MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER

mswarowsky:
The cryptocell binaries are build with this option enabled.  As this option changes the size of psa_core_key_attributes_t  it also should also be enabled when the core is build, else the some of the struct members will be misaligned.

Ref: NCSDK-23764

### DIFF
--- a/subsys/nrf_security/CMakeLists.txt
+++ b/subsys/nrf_security/CMakeLists.txt
@@ -19,12 +19,6 @@ set(mbedtls_target    mbedtls)
 # Populate ARM_MBEDTLS_PATH with the value of ZEPHYR_MBEDTLS_MODULE_DIR
 set(ARM_MBEDTLS_PATH ${ZEPHYR_MBEDTLS_MODULE_DIR})
 
-# MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER must be disabled for Zephyr
-# builds or when MBEDTLS_USE_PSA_CRYPTO is enabled (e.g. for TLS/DTLS
-# and x.509 support) Note: This configuration is internal and may be
-# removed with a new mbed TLS version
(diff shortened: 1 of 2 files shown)


subsys/nrf_security/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ set(ARM_MBEDTLS_PATH ${ZEPHYR_MBEDTLS_MODULE_DIR})
 # builds or when MBEDTLS_USE_PSA_CRYPTO is enabled (e.g. for TLS/DTLS
 # and x.509 support) Note: This configuration is internal and may be
 # removed with a new mbed TLS version
-set(CONFIG_MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER     False)
+set(CONFIG_MBEDTLS_PSA_CRYPTO_KEY_ID_ENCODES_OWNER     True)
```
joerchan, line 26: Comment above is not reflecting the change in this line.

However I don't like modifying Kconfig in cmake, so I suggest instead to remove this completely and set the Kconfig symbol to y instead.
https://github.com/nrfconnect/sdk-nrf/blob/main/subsys/nrf_security/Kconfig.psa#L17

joerchan (changes requested):

subsys/nrf_security/CMakeLists.txt
mswarowsky: Okay sounds fair, then I remove this block completely and just set the default to y

joerchan (approved):

frkv (approved):
LGTM

joerchan (changes requested):
Changed my mind, let's no do this.

PR #12770, opened 2023-10-18
samples: matter: Disabled Wi-Fi low power mode in Matter samples

kkasperczyk-no:
Disabled Wi-Fi LPM due to instabilities observed in automated tests.


kkasperczyk-no:
DNM for now, the PR is opened to prove the fix is going to stabilize automated test results.

PR #12775, opened 2023-10-18
samples: wifi: scan: Keep doing scan

krish2718:
Doing a single scan isn't that useful, so, add a configuration to take scan interval and keep doing scan.

### DIFF
--- a/samples/wifi/scan/Kconfig
+++ b/samples/wifi/scan/Kconfig
@@ -16,6 +16,12 @@ config WIFI_MAC_ADDRESS
 		use the default MAC address (locally administrative). This is only used
 		when the nRF7002 OTP is not programmed with a MAC address.
 
+config WIFI_SCAN_INTERVAL_S
+	int "Scan interval (seconds)"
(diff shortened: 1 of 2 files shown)


samples/wifi/scan/Kconfig
```diff
@@ -16,6 +16,12 @@ config WIFI_MAC_ADDRESS
 		use the default MAC address (locally administrative). This is only used
 		when the nRF7002 OTP is not programmed with a MAC address.
 
+config WIFI_SCAN_INTERVAL
```
ioannisg, line 19: you could embed the time unit in the Kconfig name, but that's nit picking

ioannisg (approved):

samples/wifi/scan/Kconfig
krish2718: No, its a good point, will do, thanks.

D-Triveni (approved):

bama-nordic (approved):

PR #12781, opened 2023-10-19
Fix debug build

krish2718:


VivekUppunda (approved):

rado17 (approved):

sachinthegreen (approved):

drivers/wifi/nrf700x/osal/os_if/src/osal.c
```diff
@@ -200,7 +200,7 @@ void nrf_wifi_osal_spinlock_irq_rel(struct nrf_wifi_osal_priv *opriv,
 }
 
 
-#ifdef CONFIG_WIFI_NRF700X_LOG_LEVEL_DBG
+#if CONFIG_WIFI_NRF700X_LOG_LEVEL >= CONFIG_WIFI_NRF700X_LOG_LEVEL_DBG
```
Damian-Nordic, line 203: Is it correct? Isn't CONFIG_WIFI_NRF700X_LOG_LEVEL_DBG a boolean like in other modules? I think you should rather compare the log level to `LOG_LEVEL_XXX` constants from `log.h`.

krish2718: Oops, you are right its a boolean choice, but I don't think we have a per-module logging levels?

krish2718: Except for using the number directly (`4` in this case), I don't see a way here, WDYT?

krish2718: ok, thanks, it should.

Damian-Nordic (approved):

Damian-Nordic:
@krish2718 It seems that you missed including `logging/log.h` in `osal_api.h` and that causes that by default all logs are enabled because C pre-processor evaluates conditions like:
```
#if 1 >= MISSING_CONFIG
```
as:
```
#if 1 >= 0
```
That causes a substantial flash growth in Matter samples.

krish2718:
> @krish2718 It seems that you missed including `logging/log.h` in `osal_api.h` and that causes that by default all logs are enabled because C pre-processor evaluates conditions like:
> 
> ```
> #if 1 >= MISSING_CONFIG
> ```
> 
> as:
> 
> ```
> #if 1 >= 0
> ```
> 
> That causes a substantial flash growth in Matter samples. I can see that including `logging/log.h` causes other build failures due to the assert macro name collission, so how about replacing the `LOG_LEVEL_XXX` macros with the corresponding numbers as we're so late in the release process?

Oops missed that, that is a OSAL file so cannot include Zephyr header, but I just tested redefining the macros and it does save huge memory, will raise a PR, thanks.

PR #12786, opened 2023-10-19
ci: add dockerfile and workflow action for releasing image

karhama:
Dockerfile and Github workflow action which builds and releases Docker image. Docker image will have Nordic toolchain bundle inside it.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,31 @@
+name: Build and Push Docker Image
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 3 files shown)


thst-nordic (approved):

.github/workflows/docker.yml
```diff
@@ -0,0 +1,31 @@
+name: Build and Push Docker Image
+
+on:
+  push:
+    tags:
+      - 'v*'
```
jangalda-nsc, line 6: should it be triggered only when new tag is pushed?

jangalda-nsc (approved):

.github/workflows/docker.yml
karhama: This is what we decided to do for now. So with this setup there won't be docker images between releases.

tejlmand (approved):

PR #12796, opened 2023-10-19
drivers: wifi: Fix inclusion of macros

krish2718:
Using Zephyr macros without Zephyr header files make all the code enabled for compilation costing memory, as we cannot use Zephyr macros in OSAL, define a new ones matching to Zephyr.

### DIFF
--- a/drivers/wifi/nrf700x/osal/os_if/inc/osal_api.h
+++ b/drivers/wifi/nrf700x/osal/os_if/inc/osal_api.h
@@ -15,6 +15,11 @@
 #include <stdarg.h>
 #include "osal_structs.h"
 
+/* Have to match zephyr/include/zephyr/logging/log_core.h */
+#define NRF_WIFI_LOG_LEVEL_ERR  1U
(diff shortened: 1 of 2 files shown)


drivers/wifi/nrf700x/osal/os_if/inc/osal_api.h
```diff
@@ -318,7 +323,7 @@ void nrf_wifi_osal_spinlock_irq_rel(struct nrf_wifi_osal_priv *opriv,
 				     unsigned long *flags);
 
 
-#if CONFIG_WIFI_NRF700X_LOG_LEVEL >= LOG_LEVEL_DBG
+#if CONFIG_WIFI_NRF700X_LOG_LEVEL >= NRF_WIFI_LOG_LEVEL_DBG
```
Damian-Nordic, line 326: `osal.c` should also be updated, right?

Damian-Nordic (approved):

rado17 (approved):

PR #12803, opened 2023-10-19
applications: nrf5340_audio: Check if local json file exists.

MariusVaardal:
If there is a json file specified by the
AUDIO_KIT_IDS environmental variable, use this as
an alternative to the json file in the buildprog folder when building.

### DIFF
--- a/applications/nrf5340_audio/tools/buildprog/buildprog.py
+++ b/applications/nrf5340_audio/tools/buildprog/buildprog.py
@@ -35,7 +35,11 @@ from pathlib import Path
 BUILDPROG_FOLDER = Path(__file__).resolve().parent
 NRF5340_AUDIO_FOLDER = (BUILDPROG_FOLDER / "../..").resolve()
 NRF_FOLDER = (BUILDPROG_FOLDER / "../../../..").resolve()
-USER_CONFIG = BUILDPROG_FOLDER / "nrf5340_audio_dk_devices.json"
+if os.getenv("AUDIO_KIT_SERIAL_NUMBERS_JSON") is None:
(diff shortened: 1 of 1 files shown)


applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -35,7 +35,7 @@
 BUILDPROG_FOLDER = Path(__file__).resolve().parent
 NRF5340_AUDIO_FOLDER = (BUILDPROG_FOLDER / "../..").resolve()
 NRF_FOLDER = (BUILDPROG_FOLDER / "../../../..").resolve()
-USER_CONFIG = BUILDPROG_FOLDER / "nrf5340_audio_dk_devices.json"
+USER_CONFIG = BUILDPROG_FOLDER / "nrf5340_audio_dk_devices.json" if os.getenv("AUDIO_KIT_IDS") is None else Path(os.getenv("AUDIO_KIT_IDS"))
```
koffes, line 38: Suggest this is made more readable by:
```suggestion
 if os.getenv("AUDIO_KIT_IDS") is None:
    USER_CONFIG = BUILDPROG_FOLDER / "nrf5340_audio_dk_devices.json" 
 else: 
     USER_CONFIG = Path(os.getenv("AUDIO_KIT_IDS"))
```

koffes: Also, mind the naming. It is a tad strange that "AUDIO_KIT_IDS are set to "USER_CONFIG".
I think your name is better.  Suggest that both AUDIO_KIT_IDS and USER_CONFIG is renamed to something like AUDIO_KIT_SNRS or AUDIO_KIT_SERIAL_NUMBERS_JSON

koffes (changes requested):

koffes (approved):

andvib (approved):

PR #12805, opened 2023-10-19
drivers: wifi: Support for nrf52840 QSPI transport

Rallare:
nRF52840 NRF_QSPI does not have support for encrypted QSPI interface. Thus we need to add checks to ensure that QSPI_ENC features are not compiled in when selecting nRF52840 and QSPI transport.


krish2718 (commented):
`qspi_enable_encryption` and `qspi_cmd_encryption` are protected with `CONFIG_SOC_SERIES_NRF53X`, so, you need to add 52840 support there as well.

Rallare:
> `qspi_enable_encryption` and `qspi_cmd_encryption` are protected with `CONFIG_SOC_SERIES_NRF53X`, so, you need to add 52840 support there as well.

For nRF52840, I believe it is fine that these report ENOTSUP, as the hardware does not support it. Are you thinking for future compatibility, that `CONFIG_SOC_SERIES_NRF53X` needs to be replaced with `(NRF_QSPI_HAS_XIP_ENC || NRF_QSPI_HAS_DMA_ENC)` to showcase actual hardware support?

krish2718:
> > `qspi_enable_encryption` and `qspi_cmd_encryption` are protected with `CONFIG_SOC_SERIES_NRF53X`, so, you need to add 52840 support there as well.
> 
> For nRF52840, I believe it is fine that these report ENOTSUP, as the hardware does not support it. Are you thinking for future compatibility, that `CONFIG_SOC_SERIES_NRF53X` needs to be replaced with `(NRF_QSPI_HAS_XIP_ENC || NRF_QSPI_HAS_DMA_ENC)` to showcase actual hardware support?

NVM, Re-reading my statement it is wrong, the APIs need 53 series protection.

krish2718 (approved):

sachinthegreen (approved):

PR #12809, opened 2023-10-20
samples: Fix peripheral samples hardfault on first boot

dchat-nordic:
Fix hardfault on some peripheral samples first boot. The settings subsystem requires more stack when setting up the nvs partition on first boot.

Jira:
NCSDK-22558
NCSDK-22559
NCSDK-22560


KAGA164 (approved):

rakons (approved):

samples/bluetooth/peripheral_cgms/prj.conf
```diff
@@ -24,3 +24,6 @@ CONFIG_BT_CGMS_MAX_MEASUREMENT_RECORD=100
 CONFIG_LOG=y
 CONFIG_BT_CGMS_LOG_LEVEL_INF=y
 CONFIG_SFLOAT=y
+
+# This example requires more stack
```
rakons, line 28: A question - are we sure this is applicable for all the targets? I mean - stack usage may differ between platforms probably?

dchat-nordic: The stack overflow appeared on nRF52DK, but the overall stack size requirement increased with the upstream rebase (which pulled the subsys/settings rework). In case of nRF52DK it was enough to push the stack usage over the size limit.

PR #12812, opened 2023-10-20
pcd: Refactor test, rename functions

sigvartmh:
Refactored the test for reading out the firmware version number as it was wrong trying to read the version into flash. Now it tests both where the buffer provided is located and that it is in RAM and we test the version reading function correctly.

Also refactored PCD function naming to align more with the rest at is no longer only does `fw_copy`.

Ref. NCSDK-24216
Ref. NCSDK-13947


hakonfam:
Should we re-factor the test immediately instead of removing it? If not we should capture that this test needs to be re-implemented in a separate task IMO.

katgiadla (approved):

subsys/pcd/Kconfig
```diff
@@ -34,6 +34,7 @@ if PCD_APP || PCD_NET
 
 config PCD_READ_NETCORE_APP_VERSION
 	bool "Include PCD command to read network core application version"
+	default y if PCD_NET
```
nordicjm, line 37: I have no idea what adding this has to do with the commit details which are:
```
 test: pcd: Removed invalid test

The test for reading out the firmware version number is wrong and the
test needs to be re-factored.
```

subsys/pcd/src/pcd.c
```diff
@@ -229,7 +231,9 @@ static int network_core_pcd_cmdset(uint32_t cmd, const void *src_addr, size_t le
 #ifdef CONFIG_PCD_READ_NETCORE_APP_VERSION
 int pcd_network_core_app_version(uint8_t *buf, size_t len)
 {
-	if (buf == NULL || len < 4) {
+	if (buf == NULL || len < 4
+	   || ((uintptr_t)buf < (uintptr_t)&_image_ram_start)
+	   || ((uintptr_t)buf > (uintptr_t)&_image_ram_end)) {
```
nordicjm, line 236: Where should this be placed, because this will exclude the 0x2000 bytes at the start of RAM which are set up when PCD is enabled?

samples/nrf5340/netboot/src/main.c
```diff
@@ -39,14 +39,13 @@ int main(void)
...
-	uint8_t status = pcd_fw_copy_status_get();
-
+	uint8_t status = pcd_cmd_status_get();
 #ifdef CONFIG_PCD_READ_NETCORE_APP_VERSION
 	if (status == PCD_STATUS_READ_VERSION) {
 		err = pcd_find_fw_version();
 		if (err < 0) {
 			printk("Unable to find valid firmware version %d\n\r", err);
-			pcd_fw_copy_invalidate();
+			goto failure;
```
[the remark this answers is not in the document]
sigvartmh: I don't want it to call `pcd_done`. I know this changes the behaviour a bit but calling done here would indicate that the command completed successfully which is not true. Meaning that if you read out the data at this point it will just be garbage.

subsys/pcd/src/pcd.c
sigvartmh: This buffer can be anywhere in RAM as long as it's in RAM and not in FLASH. 

The idea here is that you will put the buffer in the stack as a variable `uint32_t`  or `uint8_t arr[4]`or a static allocated RAM buffer/variable and both those I would assume is limited by Zephyr's build system to be within `_image_ram_start` and `_image_ram_end`.

I could limit it with `PM_PCD_SRAM_ADDRESS`  and `PM_RPMSG_NRF53_SRAM_END_ADDRESS` but it feels a bit wrong as the linker wouldn't place any of the above mentioned outside the range of `_image_ram_start` and `_image_ram_end`.

samples/nrf5340/netboot/src/main.c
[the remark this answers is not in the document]
sigvartmh: Thinking about this a bit more you are correct and it should invalidate and proceede

lemrey:
Perhaps we want to deprecate the old APIs instead of just removing them?

sigvartmh:
> Perhaps we want to deprecate the old APIs instead of just removing them?

I'm pretty certain that we are the only consumer of this library as it's only used in B0N(NSIB on the network core) and MCUBoot and I don't think the customers are touching this so I don't see the reason for deprecating them instead of removing.

PerMac:
@sigvartmh the issue is still not fixed

PR #12815, opened 2023-10-20
lib: location: Remove A-GNSS request fallback

tokangas:
Remove fallback to `NRF_MODEM_GNSS_EVT_AGNSS_REQ` event for determining GNSS assistance data need.
`nrf_modem_gnss_agnss_expiry_get()` can be used with all supported modem firmware versions.

Simplified GNSS assistance data handling.

Cleaned up error logging and some comments.

### DIFF
--- a/lib/location/method_gnss.c
+++ b/lib/location/method_gnss.c
@@ -68,9 +68,9 @@ BUILD_ASSERT(
 #define PGPS_EXPIRY_THRESHOLD 0
 /* A-GNSS minimum number of expired ephemerides to request all ephemerides. */
 #define AGNSS_EPHE_MIN_COUNT 3
-/* P-GPS minimum number of expired ephemerides to trigger injection of predictions.
- * With P-GPS, predictions are not available for all satellites, especially when
- * predictions are made further into the future, so it is natural that predictions are
+/* P-GPS minimum number of expired ephemerides to trigger injection of a prediction.
+ * With P-GPS, ephemerides are not available for all satellites, especially when
+ * predictions are made further into the future, so it is natural that ephemerides are
  * missing for some of the satellites.
  */
 #define PGPS_EPHE_MIN_COUNT 12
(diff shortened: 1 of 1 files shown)


lib/location/method_gnss.c
```diff
@@ -1175,38 +1142,18 @@ static void method_gnss_agnss_expiry_process(const struct nrf_modem_gnss_agnss_e
 	}
 }
 
-/* Queries assistance data need from GNSS.
- *
- * To maintain backward compatibility with older modem firmware versions, two different methods are
- * supported. If the nrf_modem_gnss_agnss_expiry_get() API is not supported (pre-v1.3.2 modem
```
trantanen, line 1181: I cannot add comment to old line 1258 but there are similar comments about if this is not supported etc.

lib/location/method_gnss.c
```diff
@@ -1175,38 +1142,18 @@ static void method_gnss_agnss_expiry_process(const struct nrf_modem_gnss_agnss_e
...
- * firmware), GNSS is briefly started to trigger the NRF_MODEM_GNSS_EVT_AGNSS_REQ event from GNSS
- * in case assistance data is currently needed.
- */
+/* Queries assistance data need from GNSS. */
 static void method_gnss_agnss_req_work_fn(struct k_work *work)
```
trantanen, line 1146: I'm thinking if this still needs to be running as a separate worker... Not sure but now that start and stop are not used, are all functions simply synchronous. Things would simplify and we might get rid of a sleep on line 1264.

Maybe function name could be something like `method_gnss_agnss_data_needed()` etc.

tokangas: I think the extra sleep could be removed without making the function synchronous as well, but you are right, this can now be simplified further by making the A-GNSS data need check synchronous. I'll change it.

tokangas:
I renamed `method_gnss_positioning_work_fn()` to `method_gnss_start_work_fn()` to align the work item and work handler naming, and moved a couple of functions to a different place. Because of this the diff looks worse than the changes really are.

trantanen (approved):
Good changes to make the code easier to read

jhirsi (approved):

PR #12825, opened 2023-10-23
bootloader: bl_boot: Fix disabling of UARTs

anangl:
This is a follow-up to commit 56d117d442c9148c75c621359e22283ffd7caef0 ("bootloader: bl_boot: update UART Kconfigs").

- add missing `CONFIG_` prefixes to `HAS_HW_NRF_*` symbols
- remove invalid entries that refer to non-existing peripherals (UART1 and UART2)
- move `#ifdef CONFIG_UART_NRFX` to the function, where it is actually needed (`CONFIG_UART_NRFX_UART` and `CONFIG_UART_NRFX_UARTE` imply `CONFIG_UART_NRFX`, as they cannot be defined otherwise, so there is no need for that additional `#ifdef` in inclusions).

### DIFF
--- a/subsys/bootloader/bl_boot/bl_boot.c
+++ b/subsys/bootloader/bl_boot/bl_boot.c
@@ -10,32 +10,29 @@
 #include <fw_info.h>
 #include <fprotect.h>
 #include <hal/nrf_clock.h>
-#ifdef CONFIG_UART_NRFX
 #ifdef CONFIG_UART_NRFX_UART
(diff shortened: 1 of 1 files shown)


LuDuda (approved):
This seems to fix the blocker issue with power consumption on nRF5340. Please include it in the release.

ArekBalysNordic (approved):

de-nordic (approved):

PR #12827, opened 2023-10-23
lib: at_host: improve format

MirkoCovizzi:
This commit improves the code format.

### DIFF
--- a/lib/at_host/at_host.c
+++ b/lib/at_host/at_host.c
@@ -67,13 +67,13 @@ static void response_handler(const char *response)
 
 static void cmd_send(struct k_work *work)
 {
-	int               err;
+	int err;
 
 	ARG_UNUSED(work);
 
-    /* Sending through string format rather than raw buffer in case
-     * the buffer contains characters that need to be escaped
-     */
+	/* Sending through string format rather than raw buffer in case
+	 * the buffer contains characters that need to be escaped
+	 */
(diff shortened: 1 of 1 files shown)


lib/at_host/at_host.c
```diff
@@ -93,8 +93,8 @@ static void uart_rx_handler(uint8_t character)
 	/* Handle control characters */
 	switch (character) {
 	case 0x08: /* Backspace. */
-		/* Fall through. */
-	case 0x7F: /* DEL character */
+	case 0x7F: /* Fall through. */
+		/* DEL character */
```
eivindj-nordic, line 97: When there is nothing else in the case, I don't think the fall through comment is needed. Also, the comment of which character it is should be on the same line as the case, alternatively the comment below should say `Backspace and DEL character`.

```suggestion
	case 0x7F: /* DEL character */
```

lib/at_host/at_host.c
```diff
@@ -156,13 +157,15 @@ static void uart_rx_handler(uint8_t character)
 		if (*c > ' ') {
 			break;
 		} else if (*c == '\0') {
-			return; /* Drop command, if it has no such character */
+			/* Drop command, if it has no such character */
```
eivindj-nordic, line 160:
```suggestion
			/* Drop command if it has no string termination character */
```

rlubos (approved):

eivindj-nordic (approved):

lemrey (approved):

PR #12832, opened 2023-10-23
modules: hostap: fixed wpa_cli_cmd error codes propagation

markaj-nordic:
_wpa_cli_cmd_v macro lands in the `out` section in case of an error, yet it does not set the triggered error code, so it is not passed to the caller or higher layers. This gets deceptive, especially in apps that cannot afford supplicant logging.

### DIFF
--- a/modules/hostap/src/supp_api.c
+++ b/modules/hostap/src/supp_api.c
@@ -43,7 +43,8 @@ enum status_thread_state {
 
 #define _wpa_cli_cmd_v(cmd, ...) \
 	do { \
-		if (z_wpa_cli_cmd_v(cmd, ##__VA_ARGS__) < 0) { \
+		ret = z_wpa_cli_cmd_v(cmd, ##__VA_ARGS__); \
+		if (ret < 0) { \
 			wpa_printf(MSG_ERROR, "Failed to execute wpa_cli command: %s", cmd); \
 			goto out; \
 		} \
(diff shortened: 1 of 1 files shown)


modules/hostap/src/supp_api.c
```diff
@@ -41,10 +41,12 @@ enum status_thread_state {
 
 #define DISCONNECT_TIMEOUT_MS 5000
 
-#define _wpa_cli_cmd_v(cmd, ...) \
+#define _wpa_cli_cmd_v(cmd, status, ...) \
 	do { \
-		if (z_wpa_cli_cmd_v(cmd, ##__VA_ARGS__) < 0) { \
+		int ec = z_wpa_cli_cmd_v(cmd, ##__VA_ARGS__); \
```
krish2718, line 46: You can directly use `ret` here, as this is a macro.

markaj-nordic: Hmm sure, but the supp_shell_connect_status uses different naming (`status` instead of `ret`), so I wanted to keep the macro generic.

krish2718: We can change that, in general the conditional macros rely on a common variable for error checking and don't define it themself, makes the code readable.

markaj-nordic: Ok, will fix that, thanks.

krish2718 (approved):

rado17 (approved):

PR #12657, opened 2023-10-11
doc: update known issues for NCS v2.5.0

divipillai:
Added v2.5.0 tag to known issues.

----

- [x] Asset Tracker v2 done by @simensrostad
- [x] Serial LTE Modem issue checked by @MarkusLassila 
- [x]  LTE Sensor Gateway issue checked by @lemrey (NCSDK-20026)
- [x] Low Power UART sample issue on nRF5340 DK by @nordic-krch and @anangl
- [x] Thingy:53 issue checked by @MarekPieta
- [x] Thread issues checked by @bpienk
- [x] Zigbee issues checked by @sebastiandraus
- [x] Matter issues checked by @kkasperczyk-no
- [x] HomeKit issues checked by @mariuszpos
- [x] nRF Desktop, CAF, nRF Machine Learning checked by @MarekPieta
- [x] nRF Audio (including Controller) checked by @erikrobstad 
- [x] CAF issues checked by @MarekPieta
- [x] NCS modem libraries and Networking libraries issues checked by @simensrostad  
- [x] Sidewalk issues checked by @annwoj
- [x] BLE issue checked by @grochu
- [x] BT Mesh issue checked by @Balaklaka
- [x] Build system issue checked by @osaether
- [x] DFU and FOTA issues checked by @osaether
- [x] NFC issues checked by @grochu
- [x] SPM issues checked by @osaether
- [x] Crypto issues checked by @frkv and @joerchan 
- [x] MPSL issues checked by @laxiLang
- [x] Radio driver issues checked by @ankuns
- [x] SD Controller issues checked by @laxiLang
- [x] TF-M done by @SebastianBoe
- [x] Radio test sample issue checked by @KAGA164 and @grochu
- [x] nRFxlib Modem library issues checked by @lemrey 
- [x] Zephyr revisions checked by @cvinayak 
- [x]  Mcuboot revisions checked by @de-nordic 
- [x] Zephyr LwM2M issues checked by @SeppoTakalo

PR #12846, opened 2023-10-24
modules/mcuboot/nrf53_hooks: fix boot_image_check_hook()

nvlsianpu:
boot_image_check_hook() implementation shall return fih_ret value type. The function should return FIH_BOOT_HOOK_REGULAR instead of BOOT_HOOK_REGULAR.

If the Fault Injection Hardening (FIH) is enabled then fih_int type is different than fih_ret type. Also BOOT_HOOK_REGULAR and FIH_BOOT_HOOK_REGULAR are not the same things.

ref. NCSDK-24203

Consequence of this is that when FIH is enabled, the standard boot_image_check() call is never performed by the mcuboot (for real images) and subsequent checks gets BOOT_HOOK_REGULAR as its results. In mcuboot loader.c this prematurely cause identification of invalid image.

### DIFF
--- a/modules/mcuboot/hooks/nrf53_hooks.c
+++ b/modules/mcuboot/hooks/nrf53_hooks.c
@@ -88,13 +88,13 @@ int boot_read_image_header_hook(int img_index, int slot, struct image_header *im
 	return BOOT_HOOK_REGULAR;
 }
 
-fih_int boot_image_check_hook(int img_index, int slot)
+fih_ret boot_image_check_hook(int img_index, int slot)
 {
 	if (img_index == 1 && slot == 0) {
 		FIH_RET(FIH_SUCCESS);
 	}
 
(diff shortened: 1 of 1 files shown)


de-nordic (approved):

rakons (approved):

nordicjm (approved):

nvlsianpu:
PR Rebased - CI issues were related to platform which are not affected by this patch, so probably it must have been fixed aside.

jumi-nordic (approved):
Tested locally - fixed

cvinayak:
@nvlsianpu please get changelog update and/or known issue update requested in https://github.com/nrfconnect/sdk-nrf/pull/12657

rakons:
Removed from the 2.5.0 scope - agreed to have known issues update: https://github.com/nrfconnect/sdk-nrf/pull/12657#pullrequestreview-1697164948

PR #12847, opened 2023-10-24
wifi: Fix scan timeout crash

krish2718:


markaj-nordic (approved):
Thanks! The crash is gone. I see some different issues now, but it needs more investigation.

drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
```diff
@@ -205,9 +204,7 @@ void nrf_wifi_scan_timeout_work(struct k_work *work)
 
 #ifdef CONFIG_NET_L2_WIFI_MGMT
 	if (disp_scan_cb) {
-		memset(&res, 0x0, sizeof(res));
-
-		disp_scan_cb(vif_ctx_zep->zep_net_if_ctx, -ETIMEDOUT, &res);
+		disp_scan_cb(vif_ctx_zep->zep_net_if_ctx, -ETIMEDOUT, NULL);
```
rado17: `In case of scan timeout, driver should indicate scan done with status as set to timed out, but instead an empty scan result is being sent.`

krish2718: Yes, the fix here is sending `NULL` instead of empty result which means scan is done.

rado17 (approved):

drivers/wifi/nrf700x/zephyr/src/zephyr_fmac_main.c
krish2718: Reworded the commit message.

krish2718: > The change here seems to be removal of the structure that was being declared. The status was still being set to -ETIMEDOUT Right?

There is no change in status, yes removed intermediate structs.

PR #12848, opened 2023-10-25
samples: cellular: modem_shell: Improving -h/--help handling

trantanen:
Harmonizing the way -h and --help options are handled. Starting to use CONFIG_SHELL_HELP_OPT_PARSE=n so Zephyr shell implementation doesn't "steal" those options.

We still have commands made completely with Zephyr shell macros, and those that use getopt, and that won't change.

Jira: MOSH-210


samples/cellular/modem_shell/src/location/location_shell.c
```diff
@@ -444,7 +446,7 @@ int location_shell(const struct shell *shell, size_t argc, char **argv)
 
 	gnss_location_to_cloud = false;
 
-	while ((opt = getopt_long(argc, argv, "m:t:", long_options, &long_index)) != -1) {
+	while ((opt = getopt_long(argc, argv, "m:t:h", long_options, &long_index)) != -1) {
```
tokangas, line 449: `location get` doesn't give an error if there are parameters without a hyphen. For example `location get foobar` starts location retrieval. Would there be an easy way to check that and return an error?

samples/cellular/modem_shell/src/ping/icmp_ping_shell.c
```diff
@@ -165,7 +165,11 @@ int icmp_ping_shell_th(const struct shell *shell, size_t argc, char **argv,
 			}
 			break;
 			}
-		case 'h': /* help */
+
+		case '?':
+			mosh_error("Unknown option. See usage:");
```
tokangas, line 170: Because of the `if (argc < 3)` check above, this isn't printed with invalid parameters unless there's at least two of them. Maybe that check could be removed, it seems that the presence of mandatory parameter `-d` is anyway checked separately.

samples/cellular/modem_shell/src/rest/rest_shell.c
```diff
@@ -240,11 +239,13 @@ int rest_shell(const struct shell *shell, size_t argc, char **argv)
 		case REST_SHELL_OPT_PRINT_RESP_HEADERS:
 			print_headers_from_resp = true;
 			break;
+
 		case '?':
+			mosh_error("Unknown option. See usage:");
```
tokangas, line 244: I'm not getting this error neither with `rest --foobar` nor `rest foobar`.

samples/cellular/modem_shell/src/sms/sms_shell.c
```diff
@@ -94,7 +95,9 @@ int sms_shell(const struct shell *shell, size_t argc, char **argv)
 	} else if (!strcmp(argv[1], "recv")) {
 		command = SMS_CMD_RECV;
 	} else {
-		mosh_error("Unsupported command=%s\n", argv[1]);
+		if (strcmp(argv[1], "-h") != 0 && strcmp(argv[1], "--help") != 0) {
+			mosh_error("Unsupported command=%s\n", argv[1]);
```
tokangas, line 99: `Unsupported command=%s` is used in a few places, while those where shell subcommands are used print `%s: subcommand not found`. I think it would be nice to align those.

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -174,15 +174,16 @@ static struct option long_options[] = {
...
+	{ 0,                0,                 0, 0   }
 };
 
 static void sock_print_usage(void)
 {
 	mosh_print_no_format(sock_usage_str);
 }
 
-int sock_shell(const struct shell *shell, size_t argc, char **argv)
+static int sock_shell(const struct shell *shell, size_t argc, char **argv)
```
tokangas, line 186: IMHO, the sock command has too long help, it's 100 lines long. I think this command should be split into subcommand with separate help texts, but not in this PR.

tokangas (commented):
Nice improvement.

trantanen:
As a general comment, I have an upcoming PR (Jira: MOSH-522) which will make all subcommands into Zephyr shell command macros (at, location, sms, sock).

samples/cellular/modem_shell/src/sock/sock_shell.c
trantanen: I've been doing this into similar form as link module's subcommands: https://github.com/nrfconnect/sdk-nrf/pull/12848#issuecomment-1780530644

samples/cellular/modem_shell/src/sms/sms_shell.c
trantanen: I didn't change these because these lines will be removed in an upcoming PR: https://github.com/nrfconnect/sdk-nrf/pull/12848#issuecomment-1780530644

samples/cellular/modem_shell/src/location/location_shell.c
trantanen: This is same with all commands using getopt and has always been. I haven't tried to add errors for all cases to keep things simple because the same error handling has to be copied into all places (currently 30 places in my working version). Something like this after all getopt_long while loops should do the trick:
```
	if (optind < argc) {
		mosh_error("Unknown argument '%s'", argv[optind]);
	}
```
Which would print as follows for `sms send -n 1234567890 ohhoh -m moikka jaaha`:
`Unknown argument 'ohhoh'`

samples/cellular/modem_shell/src/ping/icmp_ping_shell.c
trantanen: Yep, removed. I need to check all these `if (argc <` comparisons.

samples/cellular/modem_shell/src/rest/rest_shell.c
trantanen: Removing:
```
	if (argc < 3) {
		show_usage = true;
		goto end;
	}
```

samples/cellular/modem_shell/src/link/link_shell.c
```diff
@@ -724,10 +724,12 @@ static int link_shell_connect(const struct shell *shell, size_t argc, char **arg
 		case 'P': /* auth password */
 			password = optarg;
 			break;
+
 		case '?':
+			mosh_error("Unknown option. See usage:");
 			goto show_usage;
+		case 'h':
```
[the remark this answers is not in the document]
trantanen: `'h'` and `'h'` wasn't in the `short_options` or  `long_options`. It's added now.

trantanen:
Now that 2.5.0 is out, let's try to get this in so I can make the PR on my other changes related to making all subcommands into Zephyr shell command macros (at, location, sms, sock). We can fix any other help related command line handling there for other modules too.

tokangas (approved):

jhirsi (approved):

PR #12872, opened 2023-10-26
Raw tx pkt sample

D-Triveni:


richabp (changes requested):
Some initial comments.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+Configuration options
+=====================
+
+The following sample-specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`) :
+
+.. options-from-kconfig::
+
+**Configuration options for connected mode:**
+
+To configure the sample in connected mode, you must configure the following Wi-Fi credentials in the :file:`prj.conf` file:
```
richabp, line 38:
```suggestion
To configure the sample in connected mode, you must configure the following Wi-Fi® credentials in the :file:`prj.conf` file:
```

richabp (changes requested):

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+     west build -b nrf7002dk_nrf5340_cpuapp -- -DCONFIG_RAW_TX_PACKET_SAMPLE_IDLE_MODE=y -DCONFIG_RAW_TX_PACKET_SAMPLE_PACKET_TRANSMISSION_MODE=0
+
+* Fixed number of packet transmission with the station-idle mode:
+
+  .. code-block:: console
+
+     west build -b nrf7002dk_nrf5340_cpuapp -- -DCONFIG_RAW_TX_PACKET_SAMPLE_IDLE_MODE=y -DCONFIG_RAW_TX_PACKET_SAMPLE_PACKET_TRANSMISSION_MODE=1 -DCONFIG_RAW_TX_PACKET_SAMPLE_FIXED_NUM_PACKETS=<number of packets to be sent>
+
+Change the build target as given below for the nRF7002 EK.
```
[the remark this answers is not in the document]
D-Triveni: No, this can't be done for this example. All the above build commands can be executed with 'EK' by simply replacing the 'target_board' name.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+|config|
+
+Configuration options
+=====================
+
+The following sample-specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`) :
+
+.. options-from-kconfig::
+
+**Configuration options for connected mode:**
```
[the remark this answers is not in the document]
kittydepa: @D-Triveni I think this suggestion got missed. Please implement when you get the chance.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+Configuration options for Connected mode
+
+To configure the sample in connected mode, you must configure the following Wi-Fi® credentials in the :file:`prj.conf` file:
+
+* ``Service Set Identifier (SSID)``: Defines the name of your Wi-Fi network.
+* ``Key management``: Specifies the key management method, such as WPA2, WPA3, or other encryption protocols.
+* ``Password``: Sets the Pre-shared Key (PSK) of your Wi-Fi network.
+
+.. note::
+   You can also use ``menuconfig`` to enable ``Key management`` option.
```
kittydepa, line 45:
```suggestion
   You can also use ``menuconfig`` to enable the ``Key management`` option.
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+Configuration
+*************
+
+|config|
+
+Configuration options
+=====================
+
+The following sample-specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`) :
```
kittydepa, line 32:
```suggestion
The following sample-specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`):
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+* ``Service Set Identifier (SSID)``: Defines the name of your Wi-Fi network.
+* ``Key management``: Specifies the key management method, such as WPA2, WPA3, or other encryption protocols.
+* ``Password``: Sets the Pre-shared Key (PSK) of your Wi-Fi network.
+
+.. note::
+   You can also use ``menuconfig`` to enable ``Key management`` option.
+
+See :ref:`zephyr:menuconfig` in the Zephyr documentation for instructions on how to run ``menuconfig``.
+
+Configuration options for Idle mode
```
kittydepa, line 49:
```suggestion
Configuration options for Idle mode
----------------------------------------
```

An similar suggestion for line 36

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+Configuration options for Raw TX Packet header
+
+* ``Rate value``: Specifies the data transmission rate.
+* ``Rate flags``: Specifies the data transmission mode.
+* ``Queue number``: Specifies the transmission queue to which raw TX packets are assigned for sending.
+
+Additionally, you must configure the following option in the :file:`prj.conf` file: to define the time delay between raw TX packets:
+
+``Inter-frame delay``: Sets the time duration between raw TX packets.
```
kittydepa, line 63:
```suggestion
* ``Inter-frame delay``: Sets the time duration between raw TX packets.
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+IP addressing
+*************
```
kittydepa, line 66:
```suggestion
*************

```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+Requirements
+************
+
+The sample supports the following development kits:
+
+.. table-from-sample-yaml::
+
+Overview
+********
```
kittydepa, line 20:
```suggestion
********

```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+* Continuous transmission with the Station-connected mode.
+* Fixed number of packet transmission with the Station-connected mode.
+* Continuous transmission with the Station-idle mode.
+* Fixed number of packet transmission with the Station-idle mode.
+
+To build for the nRF7002 DK, use the ``nrf7002dk_nrf5340_cpuapp`` build target.
+The following are examples of the CLI commands:
+
+* Continuous transmission with the station-connected mode:
```
kittydepa, line 96:
```suggestion
* Continuous transmission with the Station-connected mode:
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+  .. code-block:: console
+
+     west build -b nrf7002dk_nrf5340_cpuapp -- -DCONFIG_RAW_TX_PACKET_SAMPLE_CONNECTION_MODE=y -DCONFIG_RAW_TX_PACKET_SAMPLE_PACKET_TRANSMISSION_MODE=0
+
+* Fixed number of packet transmission with the station-connected mode:
```
kittydepa, line 102:
```suggestion
* Fixed number of packet transmission with the Station-connected mode:
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+  .. code-block:: console
+
+     west build -b nrf7002dk_nrf5340_cpuapp -- -DCONFIG_RAW_TX_PACKET_SAMPLE_CONNECTION_MODE=y -DCONFIG_RAW_TX_PACKET_SAMPLE_PACKET_TRANSMISSION_MODE=1 -DCONFIG_RAW_TX_PACKET_SAMPLE_FIXED_NUM_PACKETS=<number of packets to be sent>
+
+* Continuous transmission with the station-idle mode:
```
kittydepa, line 108:
```suggestion
* Continuous transmission with the Station-idle mode:
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+  .. code-block:: console
+
+     west build -b nrf7002dk_nrf5340_cpuapp -- -DCONFIG_RAW_TX_PACKET_SAMPLE_IDLE_MODE=y -DCONFIG_RAW_TX_PACKET_SAMPLE_PACKET_TRANSMISSION_MODE=0
+
+* Fixed number of packet transmission with the station-idle mode:
```
kittydepa, line 114:
```suggestion
* Fixed number of packet transmission with the Station-idle mode:
```

kittydepa (changes requested):

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,485 @@
...
+	/* Wi-Fi access category mapping for packet */
+	unsigned char queue;
+	/* reserved parameter for driver */
+	unsigned char reserved;
+};
+
+/* SSID: NRF_RAW_TX_PACKET_APP
+ * Transmitter Address: a0:69:60:e3:52:15
+ */
+char beacon_frame[] = {
```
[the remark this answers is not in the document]
krish2718: For making it easier to edit, you can do something like this:

```
static const struct test_beacon_frame {
         u16 fc;
         u16 duration:
         char a1[6];

         ..
         payload[];
} = {       0x8000, 0x0000, ...};
```

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,485 @@
...
+	sa.sll_family = AF_PACKET;
+	sa.sll_ifindex = net_if_get_by_iface(iface);
+
+	/* Bind the socket */
+	ret = bind(sockfd, (struct sockaddr *)&sa, sizeof(struct sockaddr_ll));
+	if (ret < 0) {
+		LOG_ERR("Error: Unable to bind socket to the network interface:%d", errno);
+		close(sockfd);
+		return;
+	}
```
krish2718, line 374: This can be a helper function to `setup_raw_socket()` and outside `send_raw_tx` as it can be called many times.

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -0,0 +1,124 @@
...
+	  The valid channel range is 1 to 233.
+
+config RAW_TX_PACKET_SAMPLE_RATE_VALUE
+	int "Rate value"
+	default 0
+	help
+	  Specify the Wi-Fi data rate value for raw TX packet transmission.
+	  The value represents the data rate in Mbps.
+	  Legacy: 1, 2, 55, 11, 6, 9, 12, 18, 24, 36, 48, 54
+	  Non-legacy: (MCS index): 0-7
```
krish2718, line 78:
```suggestion
	  MCS index: 0-7
```

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -0,0 +1,124 @@
...
+
+config RAW_TX_PACKET_SAMPLE_CHANNEL
+	int "Channel for Station-Idle mode"
+	range 1 233
+	help
+	  Specify the Wi-Fi channel for transmitting raw TX packets in Station-Idle mode.
+	  The valid channel range is 1 to 233.
+
+config RAW_TX_PACKET_SAMPLE_RATE_VALUE
+	int "Rate value"
```
krish2718, line 72:
```suggestion
	int "MCS Index or Data Rate value"
```

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,485 @@
...
+		free(buffer);
+		return;
+	}
+	LOG_INF("Sending %d number of raw tx packets", num_frames);
+#else
+	num_frames = -1;
+	LOG_INF("Sending raw tx packets continuously");
+#endif
+	while (1) {
+		if (transmission_mode == 1 && num_frames == 0) {
```
krish2718, line 412: Use defines instead of magic number for mode.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,485 @@
...
+			break;
+		}
+
+		ret = sendto(sockfd, buffer, buf_length, 0,
+				(struct sockaddr *)&sa, sizeof(sa));
+		if (ret < 0) {
+			LOG_ERR("Unable to send beacon frame");
```
krish2718, line 419:
```suggestion
			LOG_ERR("Unable to send beacon frame: %s", strerror(errno));
```

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,485 @@
...
+			close(sockfd);
+			free(buffer);
+			return;
+		}
+
+		if (transmission_mode == 1 && num_frames > 0) {
+			num_frames--;
+		}
+
+		k_sleep(K_MSEC(CONFIG_RAW_TX_PACKET_SAMPLE_INTER_FRAME_DELAY_MS));
```
krish2718, line 429: `k_msleep`?

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,485 @@
...
+
+	/* Raw Tx Packet header */
+	raw_tx_pkt.magic_number = 0x12345678;
+	raw_tx_pkt.data_rate = CONFIG_RAW_TX_PACKET_SAMPLE_RATE_VALUE;
+	raw_tx_pkt.packet_length = sizeof(beacon_frame);
+	raw_tx_pkt.tx_mode = CONFIG_RAW_TX_PACKET_SAMPLE_RATE_FLAGS;
+	raw_tx_pkt.queue = CONFIG_RAW_TX_PACKET_SAMPLE_QUEUE_NUM;
+	raw_tx_pkt.reserved = 0;
+
+	buffer = malloc(sizeof(struct nrf_wifi_fmac_rawpkt_info) + sizeof(beacon_frame));
```
krish2718, line 384:
```suggestion
	test_frame = malloc(sizeof(struct nrf_wifi_fmac_rawpkt_info) + sizeof(beacon_frame));
```

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,485 @@
...
+		if ((channel_info.channel < WIFI_CHANNEL_MIN) ||
+			   (channel_info.channel > WIFI_CHANNEL_MAX)) {
+			LOG_ERR("Invalid channel number. Range is (1-233)");
+			return;
+		}
+	}
+
+	ret = net_mgmt(NET_REQUEST_WIFI_CHANNEL, iface,
+			&channel_info, sizeof(channel_info));
+
```
krish2718, line 301: Remove empty lines for NULL checks.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+Configuration options for Idle mode
+
+To configure the sample in Idle mode, you must configure the following parameter in the :file:`prj.conf` file:
+
+* ``Channel``: Specifies the Wi-Fi channel to be used for communication on the wireless network.
+
+Configuration options for Raw TX Packet header
+
+* ``Rate value``: Specifies the data transmission rate.
```
krish2718, line 57:
```suggestion
* ``MCS index or data Rate value``: Specifies the data transmission PHY rate.
```

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -0,0 +1,124 @@
...
+	bool "Open Security"
+	help
+	  Enable for Open Security
+
+config RAW_TX_PACKET_SAMPLE_KEY_MGMT_WPA2
+	bool "WPA2 Security"
+	help
+	  Enable for WPA2 Security
+
+config RAW_TX_PACKET_SAMPLE_KEY_MGMT_WPA2_256
```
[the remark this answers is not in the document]
D-Triveni: @krish2718 - @bama-nordic asked to prioritize #12907  as low.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
+.. _wifi_raw_tx_packet_sample:
+
+Wi-Fi: Raw TX Packet
```
richabp, line 3: Perhaps "packet" throughout the doc?

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+|config|
+
+Configuration options
+=====================
+
+The following sample-specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`) :
+
+.. options-from-kconfig::
+
+Configuration options for Connected mode
```
richabp, line 36:
```suggestion
Configuration options for Connected mode
----------------------------------------
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+To configure the sample in connected mode, you must configure the following Wi-Fi® credentials in the :file:`prj.conf` file:
```
richabp, line 38:
```suggestion
To configure the sample in Connected mode, you must configure the following Wi-Fi® credentials in the :file:`prj.conf` file:
```

samples/wifi/raw_tx_packet/README.rst
richabp: This got missed.

samples/wifi/raw_tx_packet/README.rst
richabp, line 49:
```suggestion
Configuration options for Idle mode
-----------------------------------
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+See :ref:`zephyr:menuconfig` in the Zephyr documentation for instructions on how to run ``menuconfig``.
+
+Configuration options for Idle mode
+
+To configure the sample in Idle mode, you must configure the following parameter in the :file:`prj.conf` file:
+
+* ``Channel``: Specifies the Wi-Fi channel to be used for communication on the wireless network.
+
+Configuration options for Raw TX Packet header
```
richabp, line 55:
```suggestion
Configuration options for Raw TX Packet header
----------------------------------------------
```

samples/wifi/raw_tx_packet/README.rst
richabp: This got missed.

samples/wifi/raw_tx_packet/README.rst
richabp: This got missed and at lines 108, 114 as well.

richabp (changes requested):
Please add a Changelog entry for this sample.

D-Triveni:
> Please add a Changelog entry for this sample.

Added.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,161 @@
...
+
+.. options-from-kconfig::
+
+Configuration options for Connected mode
+----------------------------------------
+
+To configure the sample in Connected mode, you must configure the following Wi-Fi® credentials in the :file:`prj.conf` file:
+
+* ``Service Set Identifier (SSID)``: Defines the name of your Wi-Fi network.
+* ``Key management``: Specifies the key management method, such as WPA2, WPA3, or other encryption protocols.
```
richabp, line 43:
```suggestion
* ``Key management``: Specifies the key management method, such as WPA2™, WPA3™, or other encryption protocols.
```

richabp (changes requested):

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,161 @@
...
+************
+
+The sample supports the following development kits:
+
+.. table-from-sample-yaml::
+
+Overview
+********
+
+The sample demonstrates how to transmit raw TX packets in both Station-connected and Station-idle modes of operation.
```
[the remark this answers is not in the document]
ioannisg: Maybe good to mention that these different options will be Kconfig-based configurations, not runtime.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,515 @@
...
+	while (1) {
+		wifi_connect();
+
+		while (!context.connect_result) {
+			cmd_wifi_status();
+			k_sleep(K_MSEC(STATUS_POLLING_MS));
+		}
+
+		if (context.connected) {
+			wifi_send_raw_tx_packet();
```
krish2718, line 503: I know we are following this model in all samples, but this makes it tough to read, I propose below which separates the actual operation from connection.

```
try_wifi_connect()
{
  while (1) {
      ....
      ....
  }
}

status = try_wifi_connect();
if (status != SUCCESS) 
    return -1;

wifi_send_raw_tx_packet();
```

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,515 @@
...
+	}
+	LOG_INF("Sending %d number of raw tx packets", *num_tx_pkts);
+#else
+	*num_tx_pkts = -1;
+	LOG_INF("Sending raw tx packets continuously");
+#endif
+	return 0;
+}
+
+static void wifi_send_raw_tx_packet(void)
```
krish2718, line 407: This sends more than one packets, so, should be `wifi_send_raw_tx_packets` and then you can have a helper `wifi_send_raw_tx_packet`.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,515 @@
...
+	memcpy(test_frame + sizeof(struct nrf_wifi_fmac_rawpkt_info),
+				&test_beacon_frame, sizeof(test_beacon_frame));
+
+	transmission_mode = CONFIG_RAW_TX_PACKET_SAMPLE_PACKET_TRANSMISSION_MODE;
+	while (1) {
+		if (transmission_mode == FIXED_MODE_TRANSMISSION && num_pkts == 0) {
+			break;
+		}
+
+		ret = sendto(sockfd, test_frame, buf_length, 0,
```
krish2718, line 446: Sending same packet makes it tough to debug, can you please increase `seq_ctrl` per-packet(make sure to handle wrap around at 4095)?

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,515 @@
...
+	uint16_t duration;
+	uint8_t da[6];
+	uint8_t sa[6];
+	uint8_t bssid[6];
+	uint16_t seq_ctrl;
+	uint8_t payload[256];
+};
+
+static const struct beacon test_beacon_frame = {
+	.frame_control = htons(0x8000),
```
krish2718, line 88: nit. In general capital letters are used in HEX.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,515 @@
...
+	unsigned short packet_length;
+	/* Mode describing if packet is VHT, HT, HE or Legacy */
+	unsigned char tx_mode;
+	/* Wi-Fi access category mapping for packet */
+	unsigned char queue;
+	/* reserved parameter for driver */
+	unsigned char reserved;
+};
+
+/* SSID: NRF_RAW_TX_PACKET_APP
```
krish2718, line 74: Should this comment be moved to HEX array?

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,515 @@
...
+			uint8_t connected	: 1;
+			uint8_t connect_result	: 1;
+			uint8_t disconnect_requested	: 1;
+			uint8_t _unused		: 5;
+		};
+		uint8_t all;
+	};
+} context;
+
+struct nrf_wifi_fmac_rawpkt_info {
```
krish2718, line 59: Why is this re-defined here? We should directly use the driver public header file.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,163 @@
...
+
+The sample supports the following development kits:
+
+.. table-from-sample-yaml::
+
+Overview
+********
+
+The sample demonstrates how to transmit raw TX packets in both connected Station and non-connected Station modes of operation.
+The sample provides the option to transmit raw TX packets either continuously or in a fixed number of packets.
```
VivekUppunda, line 23: can this be made a bulleted point such as 

The sample provides the option to transmit raw TX packets as described below:
1) Continuous packet transmission. 
2) Fixed number of packet transmit.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,163 @@
...
+The options to select connected Station/non-connected Station mode and continuous/fixed packets are build-time configurations.
```
VivekUppunda, line 24:
```suggestion
The options to select connected Station/non-connected Station mode and continuous/fixed packet transmissions are build-time configurations.
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,163 @@
...
+
+Configuration
+*************
+
+|config|
+
+Configuration options
+=====================
+
+The following sample-specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`):
```
VivekUppunda, line 35:
```suggestion
The following sample specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`):
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,163 @@
...
+Building and running
+********************
+
+.. |sample path| replace:: :file:`samples/wifi/raw_tx_packet`
+
+.. include:: /includes/build_and_run_ns.txt
+
+The sample can be built for the following configurations:
+
+* Continuous transmission with the connected Station mode.
```
VivekUppunda, line 95: Should this be 

Continuous raw 802.11 packet transmission in connected Station mode.

Similarly for the below sentences. change "with the" to in.
Also, should it be referred that the mode is Station with TX Injection mode

samples/wifi/raw_tx_packet/src/main.c
VivekUppunda: yes - if you can refer raw_tx_pkt_header from fmac_structs.h - kindly refer that structure

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,515 @@
...
+
+	/* MFP (optional) */
+	params->mfp = WIFI_MFP_OPTIONAL;
+
+	return 0;
+}
+
+static int wifi_connect(void)
+{
+	struct net_if *iface = net_if_get_default();
```
VivekUppunda, line 270: I am aware that this coding style is used in almost all the samples and other zephyr network samples as well. It is a belief that net_if_get_default will always return an iface. But, the function can return NULL. I would always prefer this line be broken up into two parts and code be written as below. One's code should always be able to catch any unintended results in our applications that we write

struct net_if *iface = NULL;
iface = net_if_get_default();
if (iface == NULL) {
     LOG_ERR("iface is NULL");
}

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,515 @@
...
+	/* Raw Tx Packet header */
+	raw_tx_pkt->magic_number = 0x12345678;
+	raw_tx_pkt->data_rate = CONFIG_RAW_TX_PACKET_SAMPLE_RATE_VALUE;
+	raw_tx_pkt->packet_length = sizeof(test_beacon_frame);
+	raw_tx_pkt->tx_mode = CONFIG_RAW_TX_PACKET_SAMPLE_RATE_FLAGS;
+	raw_tx_pkt->queue = CONFIG_RAW_TX_PACKET_SAMPLE_QUEUE_NUM;
+	raw_tx_pkt->reserved = 0;
+}
+
+static int print_pkt_count_info(int *num_tx_pkts)
```
VivekUppunda, line 391: Is this printing sent packet count or getting the number of packets to send? I believe this name can be changed to get_ pkt_transmit_count - no??

samples/wifi/raw_tx_packet/src/main.c
krish2718: We should all transition to `net_if_get_first_wifi()` and of course do a NULL check, just in case.

samples/wifi/raw_tx_packet/README.rst
D-Triveni: hyphen should be there.

ioannisg:
One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?

samples/wifi/raw_tx_packet/README.rst
D-Triveni: Rephrased the sentence.

krish2718:
> One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?

I thought that is what is already being done in this PR, no?

VivekUppunda (approved):

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,166 @@
+.. _wifi_raw_tx_packet_sample:
+
+Wi-Fi: Raw TX packet
+####################
+
+.. contents::
+   :local:
+   :depth: 2
+
+The Raw TX packet sample demonstrates how to transmit raw TX packets.
```
ioannisg, line 10:
```suggestion
The Raw TX packet sample demonstrates how to transmit raw IEEE 802.11 packets.
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,166 @@
...
+
+The sample supports the following development kits:
+
+.. table-from-sample-yaml::
+
+Overview
+********
+
+The sample demonstrates how to transmit raw TX packets in both connected Station and non-connected Station modes of operation.
+The sample provides the option to select the either of the modes of transmission:
```
ioannisg, line 23:
```suggestion
The sample provides the option to select the traffic pattern between the following modes:
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,166 @@
...
+
+* Continuous packet transmission.
+* Fixed number of packet transmission.
```
ioannisg, line 26:
```suggestion
* Fixed number of transmitted packets.
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,166 @@
...
+|config|
+
+Configuration options
+=====================
+
+The following sample-specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`):
+
+.. options-from-kconfig::
+
+Configuration options for connected Station mode
```
ioannisg, line 42: Could we add some indentation here, for ease of reading? That is to ensure that the reader understand what configuration is needed in each case:

- Operating mode
  - Configuration options for connected Station mode
  - Configuration options for non-connected Station mode
- Configuration options for Raw TX packet header
- Misc (or, Additional configurations)

ioannisg:
> > One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?
> 
> I thought that is what is already being done in this PR, no?

@krish2718 @D-Triveni you're right - I browsed the code and I see that you're basically fabricating a beacon frame to be sent as the raw TX packet. I think we should make this more visible in the sample documentation. We should add a sentence to clarify that "the sample uses the raw packet mode to fabricate and broadcast 802.11 beacon frames; as a result the nRF70 Series device can be identified as a Wi-Fi beaconing device.

krish2718:
> > > One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?
> > 
> > 
> > I thought that is what is already being done in this PR, no?
> 
> @krish2718 @D-Triveni you're right - I browsed the code and I see that you're basically fabricating a beacon frame to be sent as the raw TX packet. I think we should make this more visible in the sample documentation. We should add a sentence to clarify that "the sample uses the raw packet mode to fabricate and broadcast 802.11 beacon frames; as a result the nRF70 Series device can be identified as a Wi-Fi beaconing device.

In that case we should rename this sample itself to a specific usecase, i.e., `raw_tx_beacon` or just `beacon`.

D-Triveni:
> > > One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?
> > 
> > 
> > I thought that is what is already being done in this PR, no?
> 
> @krish2718 @D-Triveni you're right - I browsed the code and I see that you're basically fabricating a beacon frame to be sent as the raw TX packet. I think we should make this more visible in the sample documentation. We should add a sentence to clarify that "the sample uses the raw packet mode to fabricate and broadcast 802.11 beacon frames; as a result the nRF70 Series device can be identified as a Wi-Fi beaconing device.

sure @ioannisg , will update the document.

ioannisg:
> > > > One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?
> > > 
> > > 
> > > I thought that is what is already being done in this PR, no?
> > 
> > 
> > @krish2718 @D-Triveni you're right - I browsed the code and I see that you're basically fabricating a beacon frame to be sent as the raw TX packet. I think we should make this more visible in the sample documentation. We should add a sentence to clarify that "the sample uses the raw packet mode to fabricate and broadcast 802.11 beacon frames; as a result the nRF70 Series device can be identified as a Wi-Fi beaconing device.
> 
> In that case we should rename this sample itself to a specific usecase, i.e., `raw_tx_beacon` or just `beacon`.

No, this is not necessary IMHO. We want this sample to be seen as a generic sample for raw TX. Another reason for a generic name is because this corresponds directly to the stack-feature name (Raw TX injection/packet transmission). But, since we _do_ create and send out beacons in this example, it is nice to tell user about this in the readme. Some will want to try to sniff these beacons and confirm they can see the SSID in their scan list...

D-Triveni:
> > > > One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?
> > > 
> > > 
> > > I thought that is what is already being done in this PR, no?
> > 
> > 
> > @krish2718 @D-Triveni you're right - I browsed the code and I see that you're basically fabricating a beacon frame to be sent as the raw TX packet. I think we should make this more visible in the sample documentation. We should add a sentence to clarify that "the sample uses the raw packet mode to fabricate and broadcast 802.11 beacon frames; as a result the nRF70 Series device can be identified as a Wi-Fi beaconing device.
> 
> In that case we should rename this sample itself to a specific usecase, i.e., `raw_tx_beacon` or just `beacon`.

I suggest using `raw_tx_beacon`.

krish2718:
> > > > > One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?
> > > > 
> > > > 
> > > > I thought that is what is already being done in this PR, no?
> > > 
> > > 
> > > @krish2718 @D-Triveni you're right - I browsed the code and I see that you're basically fabricating a beacon frame to be sent as the raw TX packet. I think we should make this more visible in the sample documentation. We should add a sentence to clarify that "the sample uses the raw packet mode to fabricate and broadcast 802.11 beacon frames; as a result the nRF70 Series device can be identified as a Wi-Fi beaconing device.
> > 
> > 
> > In that case we should rename this sample itself to a specific usecase, i.e., `raw_tx_beacon` or just `beacon`.
> 
> No, this is not necessary IMHO. We want this sample to be seen as a generic sample for raw TX. Another reason for a generic name is because this corresponds directly to the stack-feature name (Raw TX injection/packet transmission). But, since we _do_ create and send out beacons in this example, it is nice to tell user about this in the readme. Some will want to try to sniff these beacons and confirm they can see the SSID in their scan list...

Sure, but the sample only supports sending `beacon` frame with various parameters, so, not a generic sample i.e., doesn't take arbitrary payload as input, so, we might just call it `beaconing` to make it abundantly clear.

ioannisg:
> > > > > > One high level comment from my side would be to make this example more tailored to a beacon-only use case, that is, the raw TX API should be used to construct and send beacons. Would that be interesting @D-Triveni @krish2718 @VivekUppunda ?
> > > > > 
> > > > > 
> > > > > I thought that is what is already being done in this PR, no?
> > > > 
> > > > 
> > > > @krish2718 @D-Triveni you're right - I browsed the code and I see that you're basically fabricating a beacon frame to be sent as the raw TX packet. I think we should make this more visible in the sample documentation. We should add a sentence to clarify that "the sample uses the raw packet mode to fabricate and broadcast 802.11 beacon frames; as a result the nRF70 Series device can be identified as a Wi-Fi beaconing device.
> > > 
> > > 
> > > In that case we should rename this sample itself to a specific usecase, i.e., `raw_tx_beacon` or just `beacon`.
> > 
> > 
> > No, this is not necessary IMHO. We want this sample to be seen as a generic sample for raw TX. Another reason for a generic name is because this corresponds directly to the stack-feature name (Raw TX injection/packet transmission). But, since we _do_ create and send out beacons in this example, it is nice to tell user about this in the readme. Some will want to try to sniff these beacons and confirm they can see the SSID in their scan list...
> 
> Sure, but the sample only supports sending `beacon` frame with various configurable parameters, so, not a generic sample i.e., doesn't take arbitrary payload as input, so, we might just call it `beaconing` to make it abundantly clear.

It does, but, as said, I want developers to easily associate the sample with the feature. If we call it "beaconing sample", they won't immediatly know it is the starting point for the raw mode (i.e. they would need to think first) :)

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -0,0 +1,124 @@
...
+
+choice RAW_TX_PACKET_SAMPLE_OPERATION_MODE
+	prompt "Choose Wi-Fi operation mode"
+
+config RAW_TX_PACKET_SAMPLE_CONNECTION_MODE
+	bool "Connected Station mode"
+	help
+	  Enable the option to select Connected Station mode.
+
+config RAW_TX_PACKET_SAMPLE_IDLE_MODE
```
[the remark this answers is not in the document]
D-Triveni: As we are using ``non-connected Station`` mode in the document, can I use the same here?

bama-nordic (approved):

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,561 @@
...
+#include <zephyr/sys/printk.h>
+#include <zephyr/init.h>
+
+#include <zephyr/net/net_if.h>
+#include <zephyr/net/wifi_mgmt.h>
+#include <zephyr/net/net_event.h>
+#include <zephyr/net/socket.h>
+#include <zephyr/drivers/gpio.h>
+
+#include <fmac_structs.h>
```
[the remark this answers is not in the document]
D-Triveni: @krish2718 It is part of ``nrfxlib``. I guess it's not required to set the path in `` CMakelists.txt`` as it's being set in ``nrf/drivers/wifi/nrf700x/CMakeLists.txt``.

samples/wifi/raw_tx_packet/Kconfig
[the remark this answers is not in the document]
krish2718: Sure, just follow a single term, IDLE is wrong.

samples/wifi/raw_tx_packet/src/main.c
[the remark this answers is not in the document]
krish2718: No, each `CMakeLists.txt` sets the include path for that module/target, `nrf/drivers/wifi/nrf700x/CMakeLists.txt` uses `zephyr_include_directories` which sets it for `zephyr_interface` (See https://github.com/zephyrproject-rtos/zephyr/blob/main/CMakeLists.txt#L100) that can be used globally by all Zephyr libraries, not samples.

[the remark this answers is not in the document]
D-Triveni: ok, thanks.  Will resolve this.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,177 @@
...
+
+  * Configuration options for connected Station mode
+  * Configuration options for non-connected Station mode
+
+* Configuration options for Raw TX packet header
+
+Configuration options for connected Station mode
+------------------------------------------------
+
+To configure the sample in connected Station mode, you must configure the following Wi-Fi® credentials in the :file:`prj.conf` file:
```
richabp, line 54:
```suggestion
To configure the sample in connected Station mode, you must configure the following Wi-Fi credentials in the :file:`prj.conf` file:
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,177 @@
...
+************
+
+The sample supports the following development kits:
+
+.. table-from-sample-yaml::
+
+Overview
+********
+The sample generates and broadcasts 802.11 beacon frames as raw TX packets.
+As a consequence, the nRF70 Series device can be identified as a Wi-Fi beaconing device.
```
richabp, line 22:
```suggestion
As a consequence, the nRF70 Series device can be identified as a Wi-Fi® beaconing device.
```

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,177 @@
...
+
+|config|
+
+Configuration options
+=====================
+
+The following sample-specific Kconfig options are used in this sample (located in :file:`samples/wifi/raw_tx_packet/Kconfig`):
+
+.. options-from-kconfig::
+
```
richabp, line 43: We can add a line here like, "Additionally, the following configuration options are available for the sample:"
Then we can move the content under these subheadings to avoid redundancy.

richabp (changes requested):

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,34 @@
...
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(raw_tx_packet, CONFIG_LOG_DEFAULT_LEVEL);
+
+#include <nrfx_clock.h>
+#include <zephyr/kernel.h>
+
+#include "net_private.h"
+
+int try_wifi_connect(void);
```
krish2718, line 28: please move to a header file.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,34 @@
...
+
+int main(void)
+{
+	int status;
+
+	LOG_INF("Starting %s with CPU frequency: %d MHz", CONFIG_BOARD, SystemCoreClock/MHZ(1));
```
krish2718, line 288: I guess this isn't relevant to this sample

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -0,0 +1,34 @@
...
+	k_sleep(K_SECONDS(1));
+
+	status = try_wifi_connect();
+	if (status < 0) {
+		return status;
+	}
```
krish2718, line 299: `return try_wifi_connect();`

samples/wifi/raw_tx_packet/src/wifi_connection.c
```diff
@@ -0,0 +1,261 @@
...
+		wifi_connect();
+
+		while (!context.connect_result) {
+			cmd_wifi_status();
+			k_sleep(K_MSEC(STATUS_POLLING_MS));
+		}
+
+		if (context.connected) {
+			break;
+		} else if (!context.connect_result) {
```
krish2718, line 254: It now has a enum for `connect_result`, so, log the result. (We should do this for existing samples as well).

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -47,4 +47,16 @@ config RAW_TX_PACKET_SAMPLE_PASSWORD
...
+	int "Set Wi-Fi STA mode"
+	default 1
+	help
+	  Set the device operating mode as Station mode.
+
+config RAW_TX_PACKET_SAMPLE_TX_INJECTION_MODE
+	int "Set Wi-Fi TX injection mode"
+	default 4
+	help
+	  Set the device operating mode as TX injection mode.
```
krish2718, line 60: We should convert these too `bool` and handle the bitmap logic in the code.

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -59,4 +59,18 @@ config RAW_TX_PACKET_SAMPLE_TX_INJECTION_MODE
 	help
 	  Set the device operating mode as TX injection mode.
 
+choice RAW_TX_PACKET_SAMPLE_OPERATION_MODE
+	prompt "Choose Wi-Fi operation mode"
```
krish2718, line 63: default missing?

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -44,17 +44,19 @@ static void wifi_set_mode(void)
 
 int main(void)
 {
-	int status;
-
 	LOG_INF("Starting %s with CPU frequency: %d MHz", CONFIG_BOARD, SystemCoreClock/MHZ(1));
 	k_sleep(K_SECONDS(1));
 
 	wifi_set_mode();
```
krish2718, line 291: shouldn't the mode also change as per the Kconfig?

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -81,4 +81,38 @@ config RAW_TX_PACKET_SAMPLE_CHANNEL
 	  Specify the Wi-Fi channel for transmitting raw TX packets in non-connected Station mode.
 	  The valid channel range is 1 to 233.
 
+config RAW_TX_PACKET_SAMPLE_RATE_VALUE
+	int "MCS index or Data rate value"
+	default 0
+	help
+	  Specify the Wi-Fi data rate value for raw TX packet transmission.
+	  The value represents the data rate in Mbps.
```
krish2718, line 89:
```
The value represents either the data rate in Mbps or Modulating and coding (MCS) index.
Note: 5.5Mbps is treated specially as 55 to avoid floating point usage.
```

samples/wifi/raw_tx_packet/CMakeLists.txt
```diff
@@ -10,6 +10,10 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 project(nrf_wifi_raw_tx_packet)
 
 target_include_directories(app PUBLIC ${ZEPHYR_BASE}/subsys/net/ip)
+target_include_directories(
+	app
+	PRIVATE
+	${ZEPHYR_BASE}/../nrfxlib/nrf_wifi/fw_if/umac_if/inc/default/)
```
krish2718, line 16: Please limit till `nrfxlib`, and use `<nrf_wifi/fw_if/umac_if/inc/default/fmac_structs.h>`

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -13,11 +13,58 @@ LOG_MODULE_REGISTER(raw_tx_packet, CONFIG_LOG_DEFAULT_LEVEL);
...
 int try_wifi_connect(void);
 
+struct beacon {
+	uint16_t frame_control;
+	uint16_t duration;
+	uint8_t da[6];
+	uint8_t sa[6];
+	uint8_t bssid[6];
+	uint16_t seq_ctrl;
+	uint8_t payload[256];
```
krish2718, line 37: please use a define

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -76,6 +123,97 @@ static void wifi_set_mode(void)
...
+	}
+
+	sa->sll_family = AF_PACKET;
+	sa->sll_ifindex = net_if_get_by_iface(iface);
+
+	/* Bind the socket */
+	ret = bind(*sockfd, (struct sockaddr *)sa, sizeof(struct sockaddr_ll));
+	if (ret < 0) {
+		LOG_ERR("Error: Unable to bind socket to the network interface:%s",
+								strerror(errno));
```
krish2718, line 156: indent should match opening bracket.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -13,11 +13,58 @@ LOG_MODULE_REGISTER(raw_tx_packet, CONFIG_LOG_DEFAULT_LEVEL);
...
+	uint8_t sa[6];
+	uint8_t bssid[6];
+	uint16_t seq_ctrl;
+	uint8_t payload[256];
+} __packed;
+
+static struct beacon test_beacon_frame = {
+	.frame_control = htons(0X8000),
+	.duration = 0X0000,
+	.da = {0Xff, 0Xff, 0Xff, 0Xff, 0Xff, 0Xff},
```
krish2718, line 43: capital letters

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -13,11 +13,58 @@ LOG_MODULE_REGISTER(raw_tx_packet, CONFIG_LOG_DEFAULT_LEVEL);
...
+	uint8_t da[6];
+	uint8_t sa[6];
+	uint8_t bssid[6];
+	uint16_t seq_ctrl;
+	uint8_t payload[256];
+} __packed;
+
+static struct beacon test_beacon_frame = {
+	.frame_control = htons(0X8000),
+	.duration = 0X0000,
```
krish2718, line 42: Will this be filled by MAC? AFAIK, it has to be filled by application based on the rate. Without this we might get interference from others as their NAV isn't updated.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -76,6 +123,97 @@ static void wifi_set_mode(void)
...
+	test_frame = malloc(sizeof(struct raw_tx_pkt_header) + sizeof(test_beacon_frame));
+	if (!test_frame) {
+		LOG_ERR("Malloc failed for send buffer %d", errno);
+		return;
+	}
+
+	buf_length = sizeof(struct raw_tx_pkt_header) + sizeof(test_beacon_frame);
+	memcpy(test_frame, &packet, sizeof(struct raw_tx_pkt_header));
+	memcpy(test_frame + sizeof(struct raw_tx_pkt_header),
+				&test_beacon_frame, sizeof(test_beacon_frame));
```
krish2718, line 244: indent to match open brace

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -115,4 +115,29 @@ config RAW_TX_PACKET_SAMPLE_QUEUE_NUM
...
+	prompt "Select transmission mode"
+	default RAW_TX_PACKET_SAMPLE_TRANSMISSION_MODE_CONTINUOUS
+
+config RAW_TX_PACKET_SAMPLE_TRANSMISSION_MODE_CONTINUOUS
+	bool "Continuous transmission mode"
+
+config RAW_TX_PACKET_SAMPLE_TRANSMISSION_MODE_FIXED
+	bool "Fixed transmission mode"
+
+config RAW_TX_PACKET_SAMPLE_FIXED_NUM_PACKETS
```
krish2718, line 128: move this out of the choice and then protect with `if RAW_TX_PACKET_SAMPLE_TRANSMISSION_MODE_FIXED` instead of depends on.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -198,15 +238,44 @@ static void wifi_send_raw_tx_packets(void)
...
-		return;
+		ret = wifi_send_raw_tx_pkt(sockfd, test_frame, buf_length, &sa);
+		if (ret < 0) {
+			LOG_ERR("Unable to send beacon frame: %s", strerror(errno));
+			close(sockfd);
+			free(test_frame);
+			return;
+		}
+	} else {
+		while (1) {
```
krish2718, line 254: if you convert the code to use a for loop you can re-use the same code for both modes:

```
unsigned int num_pkts;

num_pkts = MAX_UINT if CONTINUOUS else num_pkts
for (i = 0; i < num_pkts; i++);
```
Of course this limits to MAX_UINT packets, but I guess that should be fine.

samples/wifi/raw_tx_packet/src/main.c
```diff
@@ -198,15 +238,44 @@ static void wifi_send_raw_tx_packets(void)
...
+			if (transmission_mode == FIXED_MODE_TRANSMISSION && num_pkts == 0) {
+				break;
+			}
+
+			ret = sendto(sockfd, test_frame, buf_length, 0,
+					(struct sockaddr *)&sa, sizeof(sa));
+			if (ret < 0) {
+				LOG_ERR("Unable to send beacon frame: %s", strerror(errno));
+				close(sockfd);
+				free(test_frame);
```
krish2718, line 267: Shouldn't we continue sending even if a packet fails? We are bound to see some intermittent failures.
And we should add a summary once the transmission is done for non-continuous packets:

`Sending num_pkts  with num_pkts_fails`  or something in that lines.

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,173 @@
...
+Overview
+********
+The sample generates and broadcasts 802.11 beacon frames as raw TX packets.
+As a consequence, the nRF70 Series device can be identified as a Wi-Fi® beaconing device.
+
+The sample demonstrates how to transmit raw TX packets in both connected Station and non-connected Station modes of operation.
+The sample provides the option to select the traffic pattern between the following modes:
+
+* Continuous packet transmission.
```
krish2718, line 27: Refer Kconfig options `:ref:kconfig:option ...`

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,173 @@
...
+By using the following configuration options, you can configure the sample for different operational modes, including specific raw TX packet header settings:
+
+Configuration options for operating modes
+-----------------------------------------
+
+  * For connected Station mode
+
+    To configure the sample in connected Station mode, you must configure the following Wi-Fi credentials in the :file:`prj.conf` file:
+
+    * ``Service Set Identifier (SSID)``: Defines the name of your Wi-Fi network.
```
krish2718, line 53: please use `:ref:kconfig:option ...`

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,173 @@
...
+  * ``Queue number``: Specifies the transmission queue to which raw TX packets are assigned for sending.
+
+Additionally, you must configure the ``Inter-frame delay`` parameter in the :file:`prj.conf` file to define the time delay between raw TX packets.
+
+This sets the time duration between raw TX packets.
+
+IP addressing
+*************
+
+The sample uses DHCP to obtain an IP address for the Wi-Fi interface.
```
krish2718, line 84: please add a note that this is only for connected mode.

krish2718 (changes requested):

samples/wifi/raw_tx_packet/src/main.c
D-Triveni: We don't ahve specific header for the sample. Is that okay if I write a header file only for this declaration?

samples/wifi/raw_tx_packet/src/main.c
D-Triveni: In success case(once connected), we need to continue sending packets. Returning from try_wifi_connect doesn't work that way, right?

samples/wifi/raw_tx_packet/src/wifi_connection.c
D-Triveni: I didn't find enum definition for ``connect_result``. Can you point me to that file?

D-Triveni: I didn't find enum definition for ``connect_result``. Can you point me to that file?

samples/wifi/raw_tx_packet/src/main.c
D-Triveni: We can have 2 Kconfigs : STA_MODE_ONLY and STA_TX_INJEXTION_MODE.
And if ``STA_TX_INJEXTION_MODE`` is enabled, we can call ``wifi_set_mode`` function.

Is that what you are expecting?

samples/wifi/raw_tx_packet/src/main.c
D-Triveni: In this case, I do not think it is necessary to update duration field as we are sending out beacon frames.

samples/wifi/raw_tx_packet/src/main.c
D-Triveni: As in the above comment, it is decided to send MAX_UINT packets, adding ``Sending num_pkts  with num_pkts_fails`` print in both the cases.

samples/wifi/raw_tx_packet/src/main.c
krish2718: Sure, it can act as a placeholder.

samples/wifi/raw_tx_packet/src/main.c
krish2718: In this commit we aren't doing anything, anyways, its fine as we change it later.

samples/wifi/raw_tx_packet/src/wifi_connection.c
krish2718: https://github.com/nrfconnect/sdk-zephyr/blob/main/include/zephyr/net/wifi_mgmt.h#L339

samples/wifi/raw_tx_packet/src/main.c
krish2718: IIUC, Set mode should be called always with appropriate bits set. no?

samples/wifi/raw_tx_packet/src/main.c
krish2718: For beacon its fine, but the sample should demonstrate how to do this for others as well in case customers want to send data frames. Anyways, we can take this up later.

samples/wifi/raw_tx_packet/README.rst
richabp: We could also use :ref: to link to the Kconfig options which are mentioned in the previous section "Configuration options".
For example: see this: https://developer.nordicsemi.com/nRF_Connect_SDK_dev/doc/PR-12872/nrf/samples/wifi/sr_coex/README.html#configuration-files

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -0,0 +1,146 @@
...
+	help
+	  Specify the Wi-Fi channel for transmitting raw TX packets in non-connected Station mode.
+	  The valid channel range is 1 to 233.
+
+config RAW_TX_PACKET_SAMPLE_RATE_VALUE
+	int "MCS index or Data rate value"
+	default 0
+	help
+	  Specify the Wi-Fi data rate value for raw TX packet transmission.
+	  The value represents either the data rate in Mbps or Modulating and coding (MCS) index.
```
krish2718, line 90:
```suggestion
	  The value represents either the data rate in Mbps or Modulation and coding (MCS) index.
```

krish2718 (approved):

kittydepa (approved):

samples/wifi/raw_tx_packet/README.rst
```diff
@@ -0,0 +1,180 @@
...
+
+* For connected Station mode
+
+  To configure the sample in connected Station mode, you must configure the following Wi-Fi credentials in the :file:`prj.conf` file:
+
+  * :kconfig:option:`CONFIG_RAW_TX_PACKET_SAMPLE_SSID`: Sets the name of your Wi-Fi network.
+  * :kconfig:option:`CONFIG_RAW_TX_PACKET_SAMPLE_KEY_MGMT_NONE`: Selects open security.
+  * :kconfig:option:`CONFIG_RAW_TX_PACKET_SAMPLE_KEY_MGMT_WPA2`: Selects PSK security.
+  * :kconfig:option:`CONFIG_RAW_TX_PACKET_SAMPLE_KEY_MGMT_WPA2_256`: Selects PSK-256 security.
+  * :kconfig:option:`CONFIG_RAW_TX_PACKET_SAMPLE_KEY_MGMT_WPA3`: Selects SAE security(default).
```
richabp, line 57:
```suggestion
  * :kconfig:option:`CONFIG_RAW_TX_PACKET_SAMPLE_KEY_MGMT_WPA3`: Selects SAE security (default).
```

richabp (changes requested):

richabp (approved):

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -0,0 +1,146 @@
...
+menu "Nordic Raw Tx Packet sample"
+
+config CONNECTION_TIMEOUT
+	int "Time to wait for a station to connect"
+	default 30
+
+config RAW_TX_PACKET_SAMPLE_SSID
+	string "SSID"
+	help
+	  Specify the SSID to connect
```
ioannisg, line 18: @D-Triveni  is this really the SSID to connect to?

ioannisg: If so, maybe add some more help text, i.e. that it is needed for the Connected Station mode, etc.

ioannisg (approved):
Thanks. I left a minor comment about a help-text for a Kconfig project option, but otherwise looks good

samples/wifi/raw_tx_packet/Kconfig
D-Triveni: Yes @ioannisg , we use this SSID to connect to an AP. Will add more details in help text.

rlubos (approved):

PR #12873, opened 2023-10-26
treewide: remove HomeKit from NCS

mariuszpos:
With the introduction of Matter, all HomeKit customers are recommended to use Matter for new designs of smart home products. As a result, HomeKit Accessory Development Kit has been deprecated and removed from NCS.

Depends on:
- [x] https://github.com/nrfconnect/sdk-nrfxlib/pull/1102
- [x] https://github.com/nrfconnect/sdk-zephyr/pull/1368

### DIFF
--- a/doc/_scripts/software_maturity/software_maturity_features.yaml
+++ b/doc/_scripts/software_maturity/software_maturity_features.yaml
@@ -5,7 +5,6 @@ top_table:
   Zigbee: ZIGBEE
   Thread: NET_L2_OPENTHREAD
   LTE: LTE_LINK_CONTROL
-  HomeKit: HOMEKIT
   Matter: CHIP
   Wi-Fi:
     rule: WIFI && WIFI_NRF700X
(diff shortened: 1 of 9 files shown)


rlubos (commented):
It should also be removed from the [release tagging script](https://github.com/nrfconnect/sdk-nrf/blob/main/scripts/tag_west_repos.sh#L47).

jangalda-nsc:
Should be also removed from https://projecttools.nordicsemi.no/bitbucket/projects/NCS-TEST/repos/test-ci-nrfconnect-cfg/browse
Please ping me tomorrow, I can prepare PR

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -5,7 +5,6 @@ top_table:
   Zigbee: ZIGBEE
   Thread: NET_L2_OPENTHREAD
   LTE: LTE_LINK_CONTROL
-  HomeKit: HOMEKIT
```
mariusz-nordicsemi, line 8: Hi @gautesl 
This doesn't seem to remove HomeKit entry from the top table. Should the `:remove-rows:` option be used to achieve that? https://github.com/nrfconnect/sdk-nrf/blob/main/doc/_extensions/software_maturity_table.py#L33-L37
Or is there some other way?

gautesl: That should not be necessary, this change will affect the report that is used when building the sml part of the docs, so you will not be able to see this change before after the PR is merged

mariusz-nordicsemi (approved):

annwoj (approved):

rlubos (approved):

PR #12874, opened 2023-10-26
net: fota_download: Only retry on ECONNRESET

SeppoTakalo:
Only the ECONNRESET is documented in Download Client as a possibility to rettry.

* Fix up the documentation regarding other error codes.
* Fix the FOTA download so that we don't assume retry.
* Add testcases for download client errors.

### DIFF
--- a/include/net/download_client.h
+++ b/include/net/download_client.h
@@ -58,11 +58,16 @@ enum download_client_evt_id {
 	 * returning zero from the callback will let the library attempt
 	 * to reconnect to the server and download the last fragment again.
 	 * Otherwise, the application may return any non-zero value
-	 * to stop the download.
+	 * to stop the download. On any other error code than ECONNRESET, the client
+	 * will not attempt to reconnect and ignores the return value.
 	 *
-	 * In case the download is stopped, the application should manually
-	 * disconnect (@ref download_client_disconnect) to clean up the
-	 * network socket as necessary before re-attempting the download.
+	 * In case the download is stopped, and it was started using @ref download_client_get,
+	 * the download client automatically closes the connection. The application should wait for
+	 * DOWNLOAD_CLIENT_EVT_CLOSED before attempting another download.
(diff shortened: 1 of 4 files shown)


jheiskan81 (approved):

include/net/download_client.h
```diff
@@ -58,11 +58,16 @@ enum download_client_evt_id {
...
+	 * to stop the download. On any other error code than ECONNRESET, client
+	 * will not attempt to reconnect and ignores the return value.
 	 *
-	 * In case the download is stopped, the application should manually
-	 * disconnect (@ref download_client_disconnect) to clean up the
-	 * network socket as necessary before re-attempting the download.
+	 * In case the download is stopped, and it was started using @ref download_client_get,
+	 * then download client automatically closes the connection. Application should wait for
+	 * DOWNLOAD_CLIENT_EVT_CLOSED before attempting another download.
+	 * If download is stopped, and it was sterted using @ref download_client_start
```
umapraseeda, line 67:
```suggestion
	 * If download is stopped, and it was started using @ref download_client_start
```

subsys/net/lib/download_client/src/download_client.c
```diff
@@ -415,8 +415,14 @@ static int client_connect(struct download_client *dl)
 
 	err = connect(dl->fd, &dl->remote_addr, addrlen);
 	if (err) {
-		LOG_ERR("Unable to connect, errno %d", errno);
 		err = -errno;
+		LOG_ERR("Unable to connect, errno %d", -err);
+		/* We need to ensure that we don't return ECONNRESET as it has a special meaning
+		 * in the download client API
```
umapraseeda, line 421:
```suggestion
		/* Make sure that ECONNRESET is not returned as it has a special meaning
		 * in the download client API
```

peknis (changes requested):

umapraseeda (approved):

PR #12885, opened 2023-10-27
net: lib: lwm2m_client_utils: Update RAI

juhaylinen:
Sending of dummy packet after setting socket option SO_RAI_LAST is not needed anymore. Modem lib handles this.
Add dummy parameter to socket commands.

### DIFF
--- a/subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_rai.c
+++ b/subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_rai.c
@@ -22,7 +22,7 @@ static int set_rel14feat(void)
 	int ret;
 
 	ret = nrf_modem_at_printf("AT%%REL14FEAT=0,1,0,0,0");
-	if (ret) {
+	if (ret < 0) {
 		LOG_ERR("AT command failed, error code: %d", ret);
 		return -EFAULT;
 	}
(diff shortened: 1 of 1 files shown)


subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_rai.c
```diff
@@ -122,8 +122,10 @@ int lwm2m_rai_no_data(void)
 		}
 
 		if (ctx->sock_fd >= 0) {
+			int dummy;
```
SeppoTakalo, line 125:
```suggestion
			int dummy = 1;
```

Even when we don't use the value, initialize it, so at least we don't get warnings of using uninitialized variables, if static analyzers see this.

subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_rai.c
```diff
@@ -150,22 +151,16 @@ int lwm2m_rai_last(void)
...
```
SeppoTakalo, line 154:
```suggestion
			int dummy = 1;
```

SeppoTakalo (changes requested):

SeppoTakalo (approved):

samples/cellular/lwm2m_client/prj.conf
```diff
@@ -153,3 +153,5 @@ CONFIG_LTE_LC_TAU_PRE_WARNING_NOTIFICATIONS=y
 # Optimize powersaving by using tickless mode in LwM2M engine
 CONFIG_NET_SOCKETPAIR=y
 CONFIG_LWM2M_TICKLESS=y
+CONFIG_LWM2M_CLIENT_UTILS_RAI=y
+CONFIG_NRF_MODEM_LIB_TRACE=y
```
jheiskan81, line 157: Why this is enabled?

juhaylinen: Oops, I'll revert this change.

jheiskan81 (approved):

PR #12891, opened 2023-10-27
sample: aws_iot: Dummy credential format

junqingzou:
Change the dummy credential file format according to credentials as generated by AWS IoT, to avoid mistake in copying certificate and private key.

### DIFF
--- a/samples/net/aws_iot/certs/ca-cert.pem
+++ b/samples/net/aws_iot/certs/ca-cert.pem
@@ -1,3 +1,3 @@
-"-----BEGIN CA CERTIFICATE-----\n"
-"-----CERTIFICATE-----\n"
-"-----END CA CERTIFICATE-----\n"
+"-----BEGIN CERTIFICATE-----\n"
+"-----certificate-----\n"
+"-----END CERTIFICATE-----\n"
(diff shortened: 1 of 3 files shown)


simensrostad (approved):

samples/net/aws_iot/certs/ca-cert.pem
```diff
@@ -1,3 +1,3 @@
-"-----BEGIN CA CERTIFICATE-----\n"
-"-----CERTIFICATE-----\n"
-"-----END CA CERTIFICATE-----\n"
+"-----BEGIN CERTIFICATE-----\n"
+"-----certificate-----\n"
```
jtguggedal, line 2: Maybe we could make this even more explicit by saying some like `Replace this with the AWS IoT root CA` and similar for the other entries

coderbyheart (commented):
Why not use an invalid certificate instead so the example data has the right format?

junqingzou:
> Why not use an invalid certificate instead so the example data has the right format?

During my test I only replaced the body so it took me time to find out the header/trailer problem.
Pls propose "invalid cert and key" :)

samples/net/aws_iot/certs/ca-cert.pem
junqingzou: Please suggest changes

junqingzou:
Hi @jtguggedal and @coderbyheart shall we move this simple PR forward?

samples/net/aws_iot/certs/ca-cert.pem
jtguggedal: I was thinking of just replacing `-----certificate-----\n` with `Replace this with the AWS IoT root CA`, but don't really have strong opinions. I'll approve as is.

jtguggedal (approved):

PR #12896, opened 2023-10-27
samples: cellular: modem_shell: Subcommands into Zephyr macros

trantanen:
Changing modules which parse subcommands manually into using Zephyr shell command macros.

Jira: MOSH-522

test-sdk-nrf: sdk-nrf-pr-12896

### DIFF
--- a/samples/cellular/modem_shell/src/link/link_shell.c
+++ b/samples/cellular/modem_shell/src/link/link_shell.c
@@ -64,14 +64,14 @@ enum link_shell_common_options {
 	LINK_COMMON_RESET
 };
 
-/******************************************************************************/
 static const char link_settings_usage_str[] =
 	"Usage: link settings --read | --reset | --mreset_all | --mreset_user\n"
 	"Options:\n"
 	"  -r, --read,        Read and print current persistent settings\n"
 	"      --reset,       Reset all persistent settings to defaults\n"
 	"      --mreset_all,  Reset all modem settings to defaults\n"
-	"      --mreset_user, Reset modem user configurable settings to defaults\n";
+	"      --mreset_user, Reset modem user configurable settings to defaults\n"
+	"  -h, --help,        Shows this help information";
 
 static const char link_defcont_usage_str[] =
 	"Usage: link defcont --enable [options] | --disable | --read\n"
@@ -81,7 +81,8 @@ static const char link_defcont_usage_str[] =
 	"  -e, --enable,       Enable custom config for default PDP context\n"
 	"  -a, --apn, [str]    Set default Access Point Name\n"
 	"  -f, --family, [str] Address family: 'ipv4v6' (default), 'ipv4', 'ipv6',\n"
-	"                      'non-ip'\n";
+	"                      'non-ip'\n"
+	"  -h, --help,         Shows this help information";
 
 static const char link_defcontauth_usage_str[] =
 	"Usage: link defcontauth --enable [options] | --disable | --read\n"
@@ -91,8 +92,8 @@ static const char link_defcontauth_usage_str[] =
 	"  -e, --enable,       Enable custom config for default PDP context\n"
 	"  -U, --uname, [str]  Username\n"
 	"  -P, --pword, [str]  Password\n"
-	"  -A, --prot, [int]   Authentication protocol (Default: 0 (None), 1 (PAP),\n"
-	"                      2 (CHAP)\n";
+	"  -A, --prot, [int]   Authentication protocol (Default: 0 (None), 1 (PAP), 2 (CHAP)\n"
+	"  -h, --help,         Shows this help information";
 
 static const char link_connect_usage_str[] =
 	"Usage: link connect --apn <apn str> [--family <pdn family str>] [auth options]\n"
@@ -102,12 +103,14 @@ static const char link_connect_usage_str[] =
 	"Optional authentication options:\n"
 	"  -U, --uname, [str]  Username\n"
 	"  -P, --pword, [str]  Password\n"
-	"  -A, --prot,  [int]  Authentication protocol: 0 (None), 1 (PAP), 2 (CHAP)\n";
+	"  -A, --prot,  [int]  Authentication protocol: 0 (None), 1 (PAP), 2 (CHAP)\n"
+	"  -h, --help,         Shows this help information";
 
 static const char link_disconnect_usage_str[] =
 	"Usage: link disconnect -I <cid>\n"
 	"Options:\n"
-	"  -I, --cid, [int]    Use this option to disconnect specific PDN CID\n";
+	"  -I, --cid, [int]    Use this option to disconnect specific PDN CID\n"
+	"  -h, --help,         Shows this help information";
 
 static const char link_sysmode_usage_str[] =
 	"Usage: link sysmode [options] | --read | --reset\n"
@@ -121,11 +124,12 @@ static const char link_sysmode_usage_str[] =
 	"  -M, --ltem_gnss,        Set LTE-M + GNSS system mode\n"
 	"  -N, --nbiot_gnss,       Set NB-IoT + GNSS system mode\n"
 	"      --ltem_nbiot_gnss,  Set LTE-M + NB-IoT + GNSS system mode\n"
+	"  -h, --help,             Shows this help information\n"
 	"\n"
 	"Additional LTE mode preference that can be optionally given\n"
 	"and might make an impact with multimode system modes in modem,\n"
 	" i.e. with --ltem_nbiot or --ltem_nbiot_gnss\n"
-	"      --pref_auto,            auto, selected by modem (set as default if not\n"
+	"      --pref_auto,            Auto, selected by modem (set as default if not\n"
 	"                              given)\n"
 	"      --pref_ltem,            LTE-M is preferred over PLMN selection\n"
 	"      --pref_nbiot,           NB-IoT is preferred over PLMN selection\n"
@@ -148,8 +152,8 @@ static const char link_funmode_usage_str[] =
 	"      --gnsson,            Activates GNSS without changing LTE.\n"
 	"      --uiccoff,           Deactivates UICC.\n"
 	"      --uiccon,            Activates UICC.\n"
-	"      --flightmode_uiccon, Sets the device to flight mode without shutting down\n"
-	"                           UICC.\n";
+	"      --flightmode_uiccon, Sets the device to flight mode without shutting down UICC.\n"
+	"  -h, --help,              Shows this help information";
 
 static const char link_normal_mode_at_usage_str[] =
 	"Usage: link nmodeat --read | --mem<1-3>\n"
@@ -158,7 +162,8 @@ static const char link_normal_mode_at_usage_str[] =
 	"      --mem[1-3],   Set at cmd to given memory slot,\n"
 	"                    Example: \"link nmodeat --mem1 \"at%xbandlock=2,\\\"100\\\"\"\"\n"
 	"                    To clear the given memslot by given the empty string:\n"
-	"                    \"link nmodeat --mem2 \"\"\"\n";
+	"                    \"link nmodeat --mem2 \"\"\"\n"
+	"  -h, --help,       Shows this help information";
 
 static const char link_normal_mode_auto_usage_str[] =
 	"Usage: link nmodeauto --read | --enable | --enable_no_rel14 | --disable\n"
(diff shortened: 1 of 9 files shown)


trantanen:
PPP still not converted.

samples/cellular/modem_shell/src/ping/icmp_ping_shell.c
```diff
@@ -80,8 +80,9 @@ int icmp_ping_shell_th(const struct shell *shell, size_t argc, char **argv,
 	}
 #endif
 
-	/* Start from the 1st argument */
+	optreset = 1;
 	optind = 1;
+	int opt;
```
tokangas, line 85: While not directly related to this PR, I'm wondering if the parameters for the `ping` command should be changed. Currently, ping takes one "mandatory option" and "optional options":
```
Usage: ping [optional options] -d destination

mandatory options:
  -d, --destination, [str] Name or IP address
optional options:
  -t, --timeout, [int]     Ping timeout in milliseconds
...
```
I've been always wondering why the destination address is given using option `-d`. It's not an option, because it's not optional. The command could work in the same way as in Linux and Windows:
```
Usage
  ping [options] <destination>
```
```
Usage: ping [-t] [-a] [-n count] [-l size] [-f] [-i TTL] [-v TOS]
            [-r count] [-s count] [[-j host-list] | [-k host-list]]
            [-w timeout] [-R] [-S srcaddr] [-c compartment] [-p]
            [-4] [-6] target_name
```

samples/cellular/modem_shell/src/ppp/ppp_shell.c
```diff
@@ -72,14 +70,13 @@ static struct option long_options[] = { { "baudrate", required_argument, 0, 'b'
 					{ "help", no_argument, 0, 'h' },
 					{ 0, 0, 0, 0 } };
 
-static void ppp_shell_print_usage(struct ppp_shell_cmd_args_t *ppp_cmd_args)
+static void ppp_shell_print_usage(enum ppp_shell_command command)
```
tokangas, line 73: This could also be removed.

samples/cellular/modem_shell/src/sms/sms_shell.c
```diff
@@ -161,34 +136,85 @@ static int sms_shell(const struct shell *shell, size_t argc, char **argv)
...
-		err = sms_send_msg(arg_number, arg_message, arg_message_type);
-		break;
-	case SMS_CMD_RECV:
-		err = sms_recv(arg_receive_start);
-		break;
-	default:
-		mosh_error("Internal error. Unknown sms command=%d", command);
-		err = -EINVAL;
-		break;
+	err = sms_send_msg(arg_number, arg_message, arg_message_type);
```
tokangas, line 139: `sms send` without options prints "Number not given". Wouldn't it be better to just print the help instead? Now user has to type "sms send -h/--help" to see the command syntax.

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -27,36 +27,21 @@
...
-	"            globally enabled with 'link rai' command.\n"
-	"  list:     List open sockets. No options available.\n"
-	"\n"
-	"General options:\n"
+static const char sock_connect_usage_str[] =
+	"Usage: sock connect -i <socket id> -a <address> -p <port>\n"
+	"       [-f <family>] [-t <type>] [-b <port>] [-I <cid>]\n"
+	"       [-S] [-T <sec_tag>] [-c] [-V <level>] [-H <hostname>]\n"
+	"Options:\n"
 	"  -i, --id, [int]           Socket id. Use 'sock list' command to see open\n"
```
tokangas, line 43: I think `-i` is not used with `sock connect`.

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -100,8 +96,29 @@ static const char sock_usage_str[] =
...
+	"                            data with given length, summary of data throughput\n"
+	"                            is printed. Should be used with -r option.\n"
+	"  -B, --blocking, [int]     Blocking (1) or non-blocking (0) mode. This only\n"
+	"                            accounts when -r is given. Default value is 0.\n"
+	"  -P, --print_format, [str] Set receive data print format: 'str' (default) or\n"
+	"                            'hex'";
+
+static const char sock_rai_usage_str[] =
+	"Usage: sock rai -i <socket id> [--rai_last] [--rai_no_data] [--rai_one_resp]\n"
+	"       [--rai_ongoing] [--rai_wait_more]\n"
```
tokangas, line 118: Because only one RAI option can be given at a time, maybe the syntax should be like this:
```suggestion
	"Usage: sock rai -i <socket id> [--rai_last | --rai_no_data | --rai_one_resp |\n"
	"       --rai_ongoing | --rai_wait_more]\n"
```
At least similar syntax has been used for `sock send`.

tokangas (commented):
Great improvement, especially the `sock` command looks much nicer now. And tab works nicely with the subcommands.

samples/cellular/modem_shell/src/location/location_shell.c
```diff
@@ -149,18 +140,14 @@ static struct option long_options[] = {
 static bool gnss_location_to_cloud;
 static enum nrf_cloud_gnss_type gnss_location_to_cloud_format;
 
-/******************************************************************************/
-
-static void location_shell_print_usage(const struct shell *shell,
-				       enum location_shell_command command)
+static void location_shell_print_usage(enum location_shell_command command)
```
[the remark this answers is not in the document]
trantanen: That's been on my mind and I kept it for consistency plus extendability. I was kind of waiting for a comment on this to remove it :)
So I'll remove it.

samples/cellular/modem_shell/src/ping/icmp_ping_shell.c
trantanen: I don't know or remember but changing this makes the parsing a bit more complex. It's not something getopt_long returns then but you'll take the argument after the options.

Generally, I'm neutral on this comment as I'm not changing it in the context of this PR.

@jhirsi, thoughts?

trantanen: Plus, it's a change in test scripts.

samples/cellular/modem_shell/src/ppp/ppp_shell.c
```diff
@@ -151,41 +148,44 @@ static void ppp_shell_uart_conf_print(struct uart_config *ppp_uart_config)
...
+	return 0;
+}
+
+static int cmd_ppp_uartconf(const struct shell *shell, size_t argc, char **argv)
 {
 	int ret = 0;
 	bool uartconf_option_given = false;
+	struct uart_config ppp_uart_config;
 
 	ppp_shell_cmd_defaults_set(&ppp_cmd_args);
```
trantanen, line 178: I'm tempted to remove `struct ppp_shell_cmd_args_t` and put all variables into this function as local arguments.
Mainly because this structure is not passed to any other function so it's not really needed.
@jhirsi, what do you think?

samples/cellular/modem_shell/src/sock/sock_shell.c
trantanen: True. Removed here, usage and getopt loop.

samples/cellular/modem_shell/src/sock/sock_shell.c
trantanen: Hmmm, the code allows to give one or many of these options.

samples/cellular/modem_shell/src/sms/sms_shell.c
```diff
@@ -161,34 +136,85 @@ static int sms_shell(const struct shell *shell, size_t argc, char **argv)
...
+		}
 	}
 
+	err = sms_recv(arg_receive_start);
+
 	return err;
 
 show_usage:
 	/* Reset getopt for another users */
 	optreset = 1;
```
trantanen, line 183: I'm removing `optreset = 1` from the end of the functions because it's not everywhere and it's in the beginning for all functions.

samples/cellular/modem_shell/src/sms/sms_shell.c
trantanen: `ping` and `rest` will print failure and help. In linux world, I guess an error print with the plain command is more common although many commands has functionality without params but ping in Unix and Windows is an example.

We tend to print the help in many places when there are no arguments probably mainly because that's the concept with Zephyr shell macros.

So I guess the question is do we print error and help or just help?

samples/cellular/modem_shell/src/ping/icmp_ping_shell.c
trantanen: This change would be something like this after getopt loop (plus error handling from `-d`):

```
	if (optind < argc) {
		printf("argv[%d]=%s\n", optind, argv[optind]);
		strcpy(ping_args.target_name, argv[optind]);
	}

```

tokangas: Yes, it would make parsing the command line more complicated. And would require changes in tests. If we don't want to go that way, I suggest we at least update the help text. For example:
```
Usage: ping [options] -d destination

  -d, --destination, [str] Name or IP address
options:
  -t, --timeout, [int]     Ping timeout in milliseconds
...
```

samples/cellular/modem_shell/src/sms/sms_shell.c
tokangas: My point is that `sms send` prints ONLY the error, not the help. `ping` and `rest` commands without parameters print an error, but also the help. I'm fine with printing the error and the help with `sms send`.

samples/cellular/modem_shell/src/sock/sock_shell.c
tokangas: That should probably be corrected. The socket uses the last value set for the next send call, except no_data which gets applied immediately.

samples/cellular/modem_shell/src/ping/icmp_ping_shell.c
trantanen: Help changed.

samples/cellular/modem_shell/src/sms/sms_shell.c
trantanen: It prints now just the help.

samples/cellular/modem_shell/src/link/link_shell.c
```diff
@@ -336,7 +336,7 @@ static const char link_propripsm_usage_str[] =
 
 /******************************************************************************/
 
-/* Following are not having short options: */
+/* Following are not having short options */
```
tokangas, line 339:
```suggestion
/* The following do not have short options */
```

samples/cellular/modem_shell/src/location/location_shell.c
```diff
@@ -102,8 +94,6 @@ static const char location_get_usage_str[] =
 	"  --cloud_resp_disabled,      Do not wait for location response from cloud.\n"
 	"                              Valid if CONFIG_LOCATION_SERVICE_EXTERNAL is set.\n";
 
-/******************************************************************************/
-
 /* Following are not having short options */
```
tokangas, line 97:
```suggestion
/* The following do not have short options */
```

samples/cellular/modem_shell/src/rest/rest_shell.c
```diff
@@ -47,12 +47,12 @@ static const char rest_shell_cmd_usage_str[] =
 	"      --print_headers, Print HTTP headers of the response\n"
 	"  -i, --sckt_id,       Use existing socket id (default: create a new connection)\n";
 
-/* Following are not having short options: */
+/* Following are not having short options */
```
tokangas, line 50:
```suggestion
/* The following do not have short options */
```

samples/cellular/modem_shell/src/startup_cmd/startup_cmd_shell.c
```diff
@@ -35,14 +35,14 @@ static const char startup_cmd_usage_str[] =
 
 /******************************************************************************/
 
-/* Following are not having short options: */
+/* Following are not having short options */
```
tokangas, line 38:
```suggestion
/* The following do not have short options */
```

tokangas (commented):
Nitpicking, but since you already touched these...

samples/cellular/modem_shell/src/sms/sms_shell.c
trantanen: I've now actually change `ping` and `rest` commands also not to print errors but just usage if no options are given.

samples/cellular/modem_shell/src/sock/sock_shell.c
trantanen: Jira ticket created for this: MOSH-527

tokangas (approved):

jhirsi (approved):

samples/cellular/modem_shell/src/ppp/ppp_shell.c
trantanen: This is now done.

PR #12899, opened 2023-10-27
nrf_security: Remove enabling of legacy Mbed TLS features in PSA configuration

joerchan:
This PR is needed to enable PSA configurations for Pure EdDSA and Twisted Edwards curve.
This issue was discovered in #12647.

The issue is that enabling Legacy Mbed TLS crypto APIs based on enabled PSA crypto APIs leads to a check_config error because mbedtls does not support EdDSA and Twisted Edwards curve.
Enabling ECP requires at least one supported curve to be enabled.

The build error happens because the MBEDTLS_PSA_ACCEL_KEY_TYPE_ECC_* configurations are no longer
enabled, and when the PSA_WANT_KEY_TYPE_ECC_* configurations are enabled we enable MBEDTLS_ECP_C.

In order to fix this remove everything that is strictly not needed.

Damage assessment:
Which Legacy Mbed TLS configurations are needed when not enabling legacy Mbed TLS.

Report:
- MD needs at least one legacy hash algorithm, stop enabling it by default. The TLS stack requires MD so enable it together with that configuration
- TLS Key exchange algorithms still require legacy Mbed TLS crypto and we have added PSA_WANT as a valid dependency to enable them.
- OpenThread PSA crypto still has a dependency on Legacy Mbed TLS. Reported as KRKNWK-18015
- Enabling ECP requires enabling at least one ECC curve to be enabled.


frkv (approved):
LGTM!

PR #12902, opened 2023-10-30
manifest: Update nrfxlib revision

tokangas:
Updated nrfxlib revision to pull GNSS interface documentation updates.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -141,7 +141,7 @@ manifest:
     - name: nrfxlib
       repo-path: sdk-nrfxlib
       path: nrfxlib
-      revision: c230f914828d2ecf44a9f28922b40804216f0408
+      revision: 89d19c9b364ddd6cb54c1926649769246d1138a4
     - name: trusted-firmware-m
       repo-path: sdk-trusted-firmware-m
       path: modules/tee/tf-m/trusted-firmware-m


doc/nrf/documentation/glossary.rst
```diff
@@ -623,6 +623,16 @@ Glossary
...
 
+   Scheduled downloads
+      A mechanism to maintain GNSS data availability in periodic navigation mode.
+      If GNSS only runs long enough to calculate the first usable fix, it is probably never able to download for example almanacs or ionospheric correction data from the satellite broadcast.
+      To ensure GNSS data availability and to improve accuracy in periodic navigation mode, GNSS performs scheduled downloads.
+      When GNSS determines it needs to download ephemerides or almanacs from the broadcast, the fix interval and fix retry parameters are temporarily ignored and GNSS runs continuously for a longer period of time.
+      GNSS performs scheduled downloads until it has downloaded the data it needs.
+      After the downloads, normal operation is resumed.
+      When `A-GNSS`_ is used, scheduled downloads are not necessary and you can disable those using the :ref:`nrfxlib:nrf_modem_gnss_api`.
+      Scheduled downloads are only performed in periodic navigation mode.
```
peknis, line 634:
```suggestion
      A mechanism to maintain GNSS data availability in periodic navigation mode.
      
      If GNSS only runs long enough to calculate the first usable fix, it is probably never able to download for example almanacs or ionospheric correction data from the satellite broadcast.
      To ensure GNSS data availability and to improve accuracy in periodic navigation mode, GNSS performs scheduled downloads.
      When GNSS determines it needs to download ephemerides or almanacs from the broadcast, the fix interval and fix retry parameters are temporarily ignored and GNSS runs continuously for a longer period of time.
      GNSS performs scheduled downloads until it has downloaded the data it needs.
      After the downloads, normal operation is resumed.
      When `A-GNSS`_ is used, scheduled downloads are not necessary and you can disable them using the :ref:`nrfxlib:nrf_modem_gnss_api`.
      Scheduled downloads are only performed in periodic navigation mode.
```

peknis (changes requested):

peknis (approved):

trantanen (approved):

tokangas:
I'll create a separate PR to add scheduled downloads into glossary. That needs to be merged first to make sure there's no doc build break.

hiltunent (approved):

PR #12917, opened 2023-10-31
Zephyr migration guide and downstream merge-bases

gautesl:
WIP

This PR introduces a mechanism to find merge-bases of downstream repos in nrfconnect and insert their commit SHA into the documentation, so they do not have to be updated on every release and upmerge.

It also finds the most recent Zephyr migration guides from the two latest releases and generates a migration guide based on their differences.

### DIFF
--- a/doc/_extensions/ncs_migration_guide_extraction.py
+++ b/doc/_extensions/ncs_migration_guide_extraction.py
@@ -1,0 +1,490 @@
+"""
+Copyright (c) 2023 Nordic Semiconductor ASA
+
+SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+This extension adds four custom roles:
+    :merge-base:`<docset>`
+    :merge-base-short:`<docset>`
+    :previous-version-merge-base:`<docset>`
+    :previous-version-merge-base-short:`<docset>`
+
+These roles will be replaced with a long or short SHA for the docset in question.
+For a docset to be used with these roles, its upstream URL and main branch must
+specified with the config option `fetch_upstream_repos`. This option takes a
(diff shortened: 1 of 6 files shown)


doc/_extensions/ncs_migration_guide_extraction.py
```diff
@@ -0,0 +1,490 @@
...
+                    for title_sections in parsed_guide[change_type].values()
+                    for section in title_sections
+                ]
+        return flattened
+
+    def add_difference(base, new, add_to_base=False) -> Dict[str, Dict[str, List[str]]]:
+        target = base if add_to_base else {}
+        flattened = flatten(base)
+
+        for change_type in ("required", "recommended"):
```
gmarull, line 231: Aren't we going to remove "required"/"recommended" sections upstream?

In general, I'm very skeptical about this change. We already build Zephyr docs, and all the migration guides will already be there. If the problem is that we never pick a released Zephyr... well... let's just discuss whether this makes sense in a first place instead of building more tooling on top.

doc/nrf/releases_and_maturity/migration/zephyr_migration_guide.rst
```diff
@@ -0,0 +1,10 @@
+.. _ncs_zephyr_migration:
+
+Migration notes for SDK Zephyr
+##############################
+
+.. contents::
+   :local:
+   :depth: 2
+
+.. zephyr-migration-guide::
```
gmarull, line 10: as said: this is duplicating information that is already in the Zephyr docset (will just be split between 2 guides). I think it's easier to explain this _problem_ to developers than developing a new extension.

PR #12925, opened 2023-10-31
DFU target smp & MCUmgr SMP client unit tests

jheiskan81:
Added unit test for validate DFU Target SMP API. 
Added unit test for validate mcumgr_smp_client API.
Added image list count to 0 at dfu_target_smp_client_init()

### DIFF
--- a/tests/subsys/dfu/dfu_target/smp/CMakeLists.txt
+++ b/tests/subsys/dfu/dfu_target/smp/CMakeLists.txt
@@ -1,0 +1,38 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+cmake_minimum_required(VERSION 3.20.0)
+
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(dfu_target_smp_test)
+
+FILE(GLOB app_sources src/*.c)
+target_sources(app PRIVATE ${app_sources})
+
+target_sources(app
+  PRIVATE
+  ${ZEPHYR_NRF_MODULE_DIR}/subsys/dfu/dfu_target/src/dfu_target_smp.c
+  )
+
+set(includes
+"src/"
+)
+
+target_include_directories(app
+  PRIVATE
+  ${ZEPHYR_NRF_MODULE_DIR}/subsys/dfu/include
+  ${includes}
+  )
+
+# Mandatory stubbed flags for building test setup
+target_compile_options(app
+  PRIVATE
+  -DCONFIG_DFU_TARGET_LOG_LEVEL=2
+  -DCONFIG_DFU_TARGET_SMP=1
+  -DCONFIG_DFU_TARGET_SMP_IMAGE_LIST_SIZE=2
+  )
+
+  zephyr_library_link_libraries(MCUBOOT_BOOTUTIL)
--- a/tests/subsys/dfu/dfu_target/smp/src/os_gr_stub.c
+++ b/tests/subsys/dfu/dfu_target/smp/src/os_gr_stub.c
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/net/buf.h>
+#include <string.h>
+#include <stdio.h>
+#include <zephyr/mgmt/mcumgr/smp/smp_client.h>
+#include <zephyr/mgmt/mcumgr/mgmt/mgmt.h>
+#include <zephyr/mgmt/mcumgr/grp/os_mgmt/os_mgmt.h>
+#include <zephyr/mgmt/mcumgr/grp/os_mgmt/os_mgmt_client.h>
+#include <zephyr/sys/byteorder.h>
+
+#include <zcbor_common.h>
+#include <zcbor_encode.h>
+#include <zcbor_decode.h>
+#include <mgmt/mcumgr/util/zcbor_bulk.h>
+#include <mgmt/mcumgr/transport/smp_internal.h>
+#include "smp_stub.h"
+
+static const char *echo_ptr;
+
+void os_stub_init(const char *echo_str)
+{
+	echo_ptr = echo_str;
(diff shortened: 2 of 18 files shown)


tests/subsys/dfu/dfu_target/smp/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+target_include_directories(app
+  PRIVATE
+  ${ZEPHYR_NRF_MODULE_DIR}/subsys/dfu/include
+  ${includes}
+  )
+
+target_compile_options(app
+  PRIVATE
+  -DCONFIG_IMG_BLOCK_BUF_SIZE=4096
```
sigvartmh, line 32: is there a reason why these are provided through Cmake and not through the YAML?

jheiskan81: These need to define here that build is not failing. Enabling CONFIG_TARGET_SMP=y was causing a the other problem. We have done similar for other Unit test also.

sigvartmh: then if it fails for a specific platform you'll have to just filter the allow list.

sigvartmh (approved):

tests/subsys/dfu/dfu_target/smp/testcase.yaml
```diff
@@ -0,0 +1,4 @@
+tests:
+  dfu.dfu_target.smp:
+    platform_allow: native_posix
```
gopiotr, line 3: Please add `integration_platforms` option:
```yaml
integration_platforms:
  - native_posix
```

tests/subsys/net/lib/mcumgr_smp_client/testcase.yaml
```diff
@@ -0,0 +1,4 @@
+tests:
+  subsys.net.lib.mcumgr_smp_client.unittest:
+    platform_allow: native_posix
```
gopiotr, line 3: Please add `integration_platforms` option:
```yaml
integration_platforms:
  - native_posix
```

gopiotr (commented):
Please add to this PR commit with entry into CODEOWNERS file (https://github.com/nrfconnect/sdk-nrf/blob/main/CODEOWNERS) with path to test `/tests/subsys/net/lib/mcumgr_smp_client/` and person/people responsible for this test.

jheiskan81:
@gopiotr Thanks for review. I have updated PR based on your comments.

jheiskan81:
@gopiotr can you review my changes.

gopiotr:
> @gopiotr can you review my changes.

@jheiskan81 - thank you for changes which I asked for. It looks okay for me. I was asked for review in this PR due to the gap in CODEOWNERS file and I try to "guard" this file content (at least in a `tests` area). But I have no knowledge about `dfu` and `mcumgr` and I can't say if tests which you made are okay.

So, I will not block this PR, but it will be good if someone else with more competence could take look at this. 

@de-nordic @hakonfam please review this or pass it on to someone else.

tests/subsys/net/lib/mcumgr_smp_client/src/main.c
```diff
@@ -0,0 +1,259 @@
...
+{
+	res_buf->status = MGMT_ERR_EOK;
+	res_buf->image_list_length = 2;
+	res_buf->image_list = image_list;
+	return 0;
+}
+
+/* Stubbed Functions end */
+
+ZTEST(mcumgr_smp_client_test, test_init)
```
gopiotr, line 127: `Ztest` framework executes tests in alphanumerical order. At first glance it seems to me that this `init` test should be executed at the beginning (but I didn't analyze it deeper, so maybe I'm wrong). When I run those tests locally I can see that test execution order is as below:
```
DEBUG   - OUTPUT:  - PASS - [mcumgr_smp_client_test.test_download] duration = 0.000 seconds
DEBUG   - OUTPUT:  - PASS - [mcumgr_smp_client_test.test_erase_and_confirm] duration = 0.000 seconds
DEBUG   - OUTPUT:  - PASS - [mcumgr_smp_client_test.test_init] duration = 0.000 seconds
DEBUG   - OUTPUT:  - PASS - [mcumgr_smp_client_test.test_reset] duration = 0.000 seconds
DEBUG   - OUTPUT:  - PASS - [mcumgr_smp_client_test.test_schedule_image] duration = 0.000 seconds
```
`test_init` test is executed as a third test. Is it intentionally? If not, maybe you should consider changing this `test_init` test as a `ztest_suite_setup_t` function to guarantee that those actions will be executed always at the beginning of test suite.

jheiskan81: I can add init to ZTEST_SUITE() pre action function so Then it will be done properly. Good point. I can change that easily.

jheiskan81:
@gopiotr I move test_init inside ztest_suite_setup_t .

SeppoTakalo (approved):

carlescufi:
Waiting for @nordicjm to take a look.

tests/subsys/dfu/dfu_target/smp/src/os_gr_stub.c
```diff
@@ -0,0 +1,98 @@
...
+	nb = smp_response_buf_allocation();
+	if (nb) {
+		nb->len = 0;
+	}
+}
+
+static void os_echo_response(int status, struct zcbor_string *echo_data)
+{
+	struct net_buf *nb;
+	zcbor_state_t zse[3 + 2];
```
nordicjm, line 43: define?

tests/subsys/dfu/dfu_target/smp/src/os_gr_stub.c
```diff
@@ -0,0 +1,98 @@
...
+		smp_client_response_buf_clean();
+	} else {
+		nb->len = zse->payload - nb->data;
+	}
+}
+
+void os_echo_verify(struct net_buf *nb)
+{
+	/* Parse CBOR data: hash and confirm */
+	zcbor_state_t zsd[3 + 2];
```
nordicjm, line 75: As above

nordicjm (commented):
minor nits.

nordicjm (approved):

jheiskan81:
rebased

PR #12926, opened 2023-10-31
nrf_security: Fix broken montgomery and twisted edwards key management

joerchan:
Fix broken configuration for montgomery and twisted edwards curve key management. The configurations for the ECC curve family was missing.


mswarowsky (approved):
Good catch

Vge0rge (approved):

PR #12927, opened 2023-10-31
lib: lte_link_control: Allow more missing params for CEREG

trantanen:
The following CEREG syntax has been seen:
`+CEREG: 4,,,,,,"11100000","11100000"`

This was not understood by lte_lc because it required TAC and Cell Id when TAU and active time was present in order to succeed. AT spec says those are optional:
`+CEREG: <stat>[,[<tac>],[<ci>],[<AcT>][,[<cause_type>],[<reject_cause>][,[<Active-Time>],[<Periodic-TAU>]]]]`

Jira: NCSDK-24297


lib/lte_link_control/lte_lc.c
```diff
@@ -235,15 +235,16 @@ static void at_handler_cereg(const char *response)
...
 		memcpy(&prev_cell, &cell, sizeof(struct lte_lc_cell));
 		memcpy(&evt.cell, &cell, sizeof(struct lte_lc_cell));
 		event_handler_list_dispatch(&evt);
 	}
 
-	if (lte_mode != prev_lte_mode) {
+	if (lte_mode != prev_lte_mode && lte_mode != LTE_LC_LTE_MODE_NONE) {
 		prev_lte_mode = lte_mode;
 		evt.type = LTE_LC_EVT_LTE_MODE_UPDATE;
 		evt.lte_mode = lte_mode;
```
trantanen, line 250: The trace has been wrong because `lte_mode` could be `LTE_LC_LTE_MODE_NONE`. So depending on how we leave above condition, this might need to be fixed too.

lib/lte_link_control/lte_lc.c
```diff
@@ -235,15 +235,16 @@ static void at_handler_cereg(const char *response)
...
+	    (cell.tac != prev_cell.tac && cell.tac != LTE_LC_CELL_TAC_INVALID)) {
 		evt.type = LTE_LC_EVT_CELL_UPDATE;
 
 		memcpy(&prev_cell, &cell, sizeof(struct lte_lc_cell));
 		memcpy(&evt.cell, &cell, sizeof(struct lte_lc_cell));
 		event_handler_list_dispatch(&evt);
 	}
 
-	if (lte_mode != prev_lte_mode) {
+	if (lte_mode != prev_lte_mode && lte_mode != LTE_LC_LTE_MODE_NONE) {
```
trantanen, line 247: Do we want to send events when we don't get a valid value for `cell.id`, `cell.tac` or `lte_mode`?
I made the change to provoke discussion.
I know LTE_LC_EVT_CELL_UPDATE would still be sent if either `cell.id` or `cell.tac` is valid.

Comments @jtguggedal?

jtguggedal: Good question.. I guess it could trigger a few unnecessary events while searching for a cell if events are dispatched even if both are invalid. The question is if there's really a use-case for an application that needs to receive events if both are invalid. Can't really see one

tokangas: Shouldn't the `LTE_LC_EVT_CELL_UPDATE` event be sent also to inform that the current cell has been lost?

`lte_mode` may be a different case. I'm not sure, though. For example, when modem drops from the network, I guess `LTE_LC_LTE_MODE_NONE` would be appropriate.

The point being that these are not invalid values, but rather indicating that the modem is not camped on a cell or not registered using any technology.

trantanen: I don't have strong opinion on this which is why I'm asking for opinions :)
Cell lost information is coming through `LTE_LC_EVT_NW_REG_STATUS` so I'm not sure if `LTE_LC_EVT_CELL_UPDATE` or `LTE_LC_EVT_CELL_UPDATE` are needed. Applications that are not monitoring `LTE_LC_EVT_NW_REG_STATUS` would maybe like that all relevant events are sent.

trantanen: @jtguggedal, Opinions on @tokangas's comment?

tokangas: I think we should aim to keep the behavior unchanged. I don't think it's enough that the same information is available through some other event, because such change would break application which rely on the current behavior. That said, I'm not totally sure what is the current behavior and are you changing that or not. 😃

tokangas: As discussed offline, `LTE_LC_EVT_CELL_UPDATE` with cell ID and TAC set to -1 is sent by the current implementation for example when a `+CEREG=0` notification is received. I think the behavior should be the same also with a `+CEREG: 4,,,,,,"11100000","11100000"` notification.

Also, in the same case a `LTE_LC_EVT_NW_REG_STATUS` notification is sent with `lte_mode` set to `LTE_LC_LTE_MODE_NONE`.

jtguggedal: Sounds reasonable.
I'm not sure how `+CEREG: 4,,,,,,"11100000","11100000"` is even possible, as it indicates unknown registration status and at the same time known parameters given by the network (which we don't know if we're registered with anymore), but that's a separate discussion...

lib/lte_link_control/lte_lc.c
trantanen: @maxd-nordic, please check that below change to memfault traces is fine.

lib/lte_link_control/lte_lc.c
trantanen: `4` is unknown status. It's not "not registered", which is `0`. We are not on a cell but we are registered. There are special conditions when this can be sent.

lib/lte_link_control/lte_lc.c
jtguggedal: This is fine, Memfault ignores LTE mode changes in the current code

jtguggedal (approved):

lib/lte_link_control/lte_lc.c
jtguggedal: Yeah, I'm probably just intepreting the docs on CEREG wrongly. When the status is set to "Unknown", I interpret that as that the modem does not know whether it's still registered or not, and in that case I find it confusing that it would know the PSM configuration.

tokangas: I find it also confusing that usually `+CEREG=4` notification does not have TAU and Active-Time parameters, but in some special case it does. Unless there's a very good reason why, I think it would have been better to always send the notification without those parameters.

tokangas (approved):

PR #12935, opened 2023-11-01
doc: wifi: update TWT

kittydepa:
Update TWT sample with Kconfig option needed to be able to run the traffic server

### DIFF
--- a/samples/wifi/twt/README.rst
+++ b/samples/wifi/twt/README.rst
@@ -31,13 +31,17 @@ When the traffic generator module is used with this sample, the traffic generato
 The server source is available in the :file:`scripts/traffic_gen_server` file.
 This is a `python3` based server, so you must make sure that `python3` is installed on your PC.
 
-To start the server, run the following command:
+To start the server:
 
-.. code-block:: console
+1. Set the :kconfig:option:`CONFIG_TRAFFIC_GEN_REMOTE_IPV4_ADDR` Kconfig option to the IPv4 address of the host running the traffic generator server.
 
-   python3 traffic_gen_server.py
+2. Run the following command:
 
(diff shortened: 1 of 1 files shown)


samples/wifi/twt/README.rst
```diff
@@ -31,13 +31,17 @@ When the traffic generator module is used with this sample, the traffic generato
 The server source is available in the :file:`scripts/traffic_gen_server` file.
 This is a `python3` based server, so you must make sure that `python3` is installed on your PC.
 
-To start the server, run the following command:
+To start the server:
 
-.. code-block:: console
+1. Set the :kconfig:option:`CONFIG_TRAFFIC_GEN_REMOTE_IPV4_ADDR` Kconfig option to the IPv4 address of the host running the traffic generator server.
```
krish2718, line 36: Should we clarify that this is for the client side i.e., for Zephyr build? or is it implied?

bama-nordic (approved):

samples/wifi/twt/README.rst
richabp, line 36:
```suggestion
1. Set the :ref:`CONFIG_TRAFFIC_GEN_REMOTE_IPV4_ADDR` Kconfig option to the IPv4 address of the host running the traffic generator server.
```

richabp (changes requested):
Do we need  a changelog entry for this as this Kconfig option is already available in the sample, the new instruction is about setting it to the IPv4 address of the host.

richabp (approved):

samples/wifi/twt/README.rst
ioannisg: I think this is implied well @kittydepa

ioannisg (approved):

PR #12936, opened 2023-11-01
SLM: FOTA fixes

tomi-font:
See commit messages for more information.


include/net/download_client.h
```diff
@@ -286,6 +288,18 @@ int download_client_start(struct download_client *client, const char *file,
...
+ * @brief Retrieve the number of bytes downloaded so far.
+ *
+ * The progress is only available after the download has begun.
+ *
+ * @param[in]  client	Client instance.
+ * @param[out] progress	Number of bytes downloaded so far.
+ *
+ * @retval int Zero on success, a negative error code otherwise.
+ */
+int download_client_downloaded_size_get(struct download_client *client, size_t *size);
```
wiba-nordic, line 301: Doc build error seems to be related to the definitions not matching up with the parameters.

wiba-nordic (approved):

sigvartmh (approved):

subsys/net/lib/fota_download/src/fota_download.c
```diff
@@ -106,9 +106,12 @@ static void stopped(void)
 static void dfu_target_callback_handler(enum dfu_target_evt_id evt)
 {
 	switch (evt) {
-	case DFU_TARGET_EVT_TIMEOUT:
+	case DFU_TARGET_EVT_ERASE_PENDING:
 		send_evt(FOTA_DOWNLOAD_EVT_ERASE_PENDING);
 		break;
+	case DFU_TARGET_EVT_TIMEOUT:
+		send_evt(FOTA_DOWNLOAD_EVT_ERASE_TIMEOUT);
```
MarkusLassila, line 113: Does this event need to be handled in other places as well?

MarkusLassila (changes requested):

subsys/net/lib/fota_download/src/fota_download.c
MarkusLassila: Yes. Some of the places that process FOTA_DOWNLOAD_EVT_ERASE_PENDING treat non-mapped events as unknown and report errors on those.

SeppoTakalo:
Is this now changing the meaning on  `FOTA_DOWNLOAD_EVT_ERASE_PENDING`?

I'm not sure from the documentation that what is the purpose of that event. Is the user supposed to react to that or not.

SeppoTakalo (approved):

subsys/net/lib/fota_download/src/fota_download.c
MarkusLassila: @tomi-font: This should be handled.

de-nordic: @MarkusLassila Can you re-review?

MarkusLassila: Apologies for delay. It seems that I am not receiving all the notifications from these.

@tomi-font
This will write an error to the log, if the event is received. It would be better to just process it, or at least add processing so that the people responsible for these could review this.

https://github.com/nrfconnect/sdk-nrf/blob/7c10920b971b1ef3285f8b86703c5430de8556d9/samples/cellular/modem_shell/src/fota/fota.c#L81C4-L105

And if I am not mistaken, the same happens with these:
https://github.com/nrfconnect/sdk-nrf/blob/7c10920b971b1ef3285f8b86703c5430de8556d9/subsys/net/lib/aws_fota/src/aws_fota.c#L733C1-L800C1

https://github.com/nrfconnect/sdk-nrf/blob/7c10920b971b1ef3285f8b86703c5430de8556d9/subsys/net/lib/azure_fota/azure_fota.c#L228C1-L268C3

wiba-nordic (approved):

MarkusLassila (approved):

simensrostad (approved):
AWS and Azure stuff :+1:

samples/cellular/modem_shell/src/fota/fota.c
```diff
@@ -88,6 +88,9 @@ static void fota_download_callback(const struct fota_download_evt *evt)
 		mosh_print("FOTA: Download finished");
 		modem_update_apply();
 		break;
+	case FOTA_DOWNLOAD_EVT_ERASE_TIMEOUT:
+		mosh_print("FOTA: Erasing reached timeout");
+		/* Fall through, erasing continues. */
 	case FOTA_DOWNLOAD_EVT_ERASE_PENDING:
 		mosh_print("FOTA: Still erasing...");
 		break;
```
trantanen, line 96: @tokangas Did you check this and is this fine?

jayteemo (commented):
✅ nrf_cloud and samples

samples/cellular/modem_shell/src/fota/fota.c
tokangas: @trantanen Yes, I did check it. As far as I understand, the timeout event gets triggered periodically, but erasing still continues. The Kconfig description for the timeout says that the LTE link could be disconnected to allow the erasing of the flash to happen more quickly, but with modem_shell I think we can leave that to the user.

PR #12937, opened 2023-11-01
Add crypto-related configuration options

Vge0rge:
Adds a configuration option to set the TF-M MM-IOVEC through Kconfig.

Also adds a configuration to set MBEDTLS_PSA_KEY_SLOT_COUNT through Kconfig.

### DIFF
--- a/modules/trusted-firmware-m/Kconfig
+++ b/modules/trusted-firmware-m/Kconfig
@@ -356,4 +356,12 @@ config TFM_NRF_PROVISIONING
 	  provisioning lifecycle state in order to boot. See provisioning
 	  guide for more details.
 
+config TFM_PSA_FRAMEWORK_HAS_MM_IOVEC
+	bool
+	prompt "TF-M MM-IOVEC"
+	depends on TFM_ISOLATION_LEVEL = 1
+	help
+	  Memory-mapped iovecs provide direct mapping of client input and output vectors into
+	  the Secure Partition.
+
 endif # BUILD_WITH_TFM
--- a/subsys/nrf_security/Kconfig.psa
+++ b/subsys/nrf_security/Kconfig.psa
@@ -74,6 +74,16 @@ config MBEDTLS_USE_PSA_CRYPTO
 	help
 	  Corresponds to MBEDTLS_USE_PSA_CRYPTO setting in mbed TLS config file.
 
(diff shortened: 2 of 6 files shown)


mswarowsky (approved):
lgtm, assuming the typo gets fixed

modules/trusted-firmware-m/Kconfig
```diff
@@ -466,4 +466,11 @@ config TFM_NRF_PROVISIONING
 	  provisioning lifecycle state in order to boot. See provisioning
 	  guide for more details.
 
+config TFM_PSA_FRAMEWORK_HAS_MM_IOVEC
+	bool
+	prompt "TF-M MM-IOVEC" if TFM_SFN
```
joerchan, line 471: I couldn't find a lot of documentation for this configuration, but from what I can tell it is not limited to SFN, but to isolation level 1.

There is a configuration check that when this is enabled the isolation level is not larger than 1.

Also, depends on seems better than prompt if checking.

subsys/nrf_security/Kconfig.psa
```diff
@@ -74,6 +74,15 @@ config MBEDTLS_USE_PSA_CRYPTO
 	help
 	  Corresponds to MBEDTLS_USE_PSA_CRYPTO setting in mbed TLS config file.
 
+config MBEDTLS_PSA_KEY_SLOT_COUNT
+	int "Number of PSA key slots available"
+	default 32
```
joerchan, line 79: Maybe we should give this a range, to disallow 0 as in input.

subsys/nrf_security/Kconfig.psa
```diff
@@ -74,6 +74,15 @@ config MBEDTLS_USE_PSA_CRYPTO
...
+	help
+	  Describes the number of avaiable PSA key slots. A key slot is used in PSA
```
joerchan, line 81: available.

modules/trusted-firmware-m/Kconfig
Vge0rge: I can update to isolation level 1, sure.

subsys/nrf_security/Kconfig.psa
Vge0rge: I added 1-32 since 32 is the default in MBedTLS.

joerchan (approved):

PR #12940, opened 2023-11-02
drivers: wifi: Provide option to select frequency

kapbh:
Provide a Kconfig option to select frequency band
2.4GHz, 5GHz or both(2.4 and 5 GHz).

### DIFF
--- a/drivers/wifi/nrf700x/Kconfig
+++ b/drivers/wifi/nrf700x/Kconfig
@@ -401,3 +401,22 @@ config NRF_WIFI_BEAMFORMING
 config WIFI_NRF700X_SCAN_TIMEOUT_S
 	int "Scan timeout in seconds"
 	default 30
+
+config NRF_WIFI_2G_BAND
+	bool "Set operation band to 2.4GHz"
+	default y if BOARD_NRF7001
+
+config NRF_WIFI_5G_BAND
+	bool "Set operation band to 5GHz"
+	depends on !BOARD_NRF7001
+
+config NRF_WIFI_OP_BAND
+	int "Options to set operation band"
+	default 1 if NRF_WIFI_2G_BAND
+	default 2 if NRF_WIFI_5G_BAND
+	default 3
+	help
+	  Set this option to select frequency band
+	  1 - 2.4GHz
+	  2 - 5GHz
+	  3 - All ( 2.4GHz and 5GHz )
--- a/drivers/wifi/nrf700x/src/fmac_main.c
+++ b/drivers/wifi/nrf700x/src/fmac_main.c
@@ -483,11 +483,7 @@ enum nrf_wifi_status nrf_wifi_fmac_dev_add_zep(struct nrf_wifi_drv_priv_zep *drv
 	enum nrf_wifi_status status = NRF_WIFI_STATUS_FAIL;
 	struct nrf_wifi_ctx_zep *rpu_ctx_zep = NULL;
 	void *rpu_ctx = NULL;
-#if defined(CONFIG_BOARD_NRF7001)
-	enum op_band op_band = BAND_24G;
-#else /* CONFIG_BOARD_NRF7001 */
-	enum op_band op_band = BAND_ALL;
-#endif /* CONFIG_BOARD_NRF7001 */
+	enum op_band op_band = CONFIG_NRF_WIFI_OP_BAND;
 #ifdef CONFIG_NRF_WIFI_LOW_POWER
 	int sleep_type = -1;
 


drivers/wifi/nrf700x/Kconfig
```diff
@@ -374,3 +374,21 @@ config NRF_WIFI_BEAMFORMING
...
+config NRF_WIFI_OP_BAND
+    int ""
+    default 1 if NRF_WIFI_2G_BAND
+	default 2 if NRF_WIFI_5G_BAND
+	default 0
+	help
+	  Set this option to select frequency band
+	  1 - 2.4GHz
+	  2 - 5GHz
+	  0 - 2.4GHz and 5GHz Both
```
VivekUppunda, line 394: you would probably need 6 GHz band also

kapbh: Yes probably. will add that option also. Thanks for suggestion.

drivers/wifi/nrf700x/Kconfig
```diff
@@ -374,3 +374,26 @@ config NRF_WIFI_BEAMFORMING
 config WIFI_NRF700X_SCAN_TIMEOUT_S
 	int "Scan timeout in seconds"
 	default 30
+
+config NRF_WIFI_2G_BAND
```
krish2718, line 378: Current code only chooses one of these, so it could have been a choice. But I  think we should support multiple selections.

drivers/wifi/nrf700x/Kconfig
```diff
@@ -374,3 +374,26 @@ config NRF_WIFI_BEAMFORMING
...
+	bool ""
```
krish2718, line 379: Why are these options hidden?

drivers/wifi/nrf700x/src/fmac_main.c
```diff
@@ -483,11 +483,7 @@ enum nrf_wifi_status nrf_wifi_fmac_dev_add_zep(struct nrf_wifi_drv_priv_zep *drv
 	enum nrf_wifi_status status = NRF_WIFI_STATUS_FAIL;
 	struct nrf_wifi_ctx_zep *rpu_ctx_zep = NULL;
 	void *rpu_ctx = NULL;
-#if defined(CONFIG_BOARD_NRF7001)
-	enum op_band op_band = BAND_24G;
-#else /* CONFIG_BOARD_NRF7001 */
-	enum op_band op_band = BAND_ALL;
-#endif /* CONFIG_BOARD_NRF7001 */
+	enum op_band op_band = CONFIG_NRF_WIFI_OP_BAND;
```
krish2718, line 486: Is the Kconfig and enum matching?

kapbh: Yes, Current enum order is matching with Kconfig. There are 2 options available with enum, BAND_ALL and BAND_24G. Other Kconfig are for further enhancement.

drivers/wifi/nrf700x/Kconfig
kapbh: The band will be set based on boards. nrf7001 only supports 2.4GHz so directly only 2.4GHz band will be set.

krish2718: We already have that, I thought the purpose of Kconfig option is to choose independent of SoC, no? i.e., even for 7002 we can choose the band.

drivers/wifi/nrf700x/Kconfig
kapbh: Changed the NRF_WIFI_OP_BAND option in bit. It will support multiple selections.

drivers/wifi/nrf700x/Kconfig
```diff
@@ -401,3 +401,26 @@ config NRF_WIFI_BEAMFORMING
...
+	default 1 if NRF_WIFI_2G_BAND
+	default 2 if NRF_WIFI_5G_BAND
+	default 4 if NRF_WIFI_6G_BAND
+	default 7
+	help
+	  Set this option to select frequency band
+	  1 - 2.4GHz
+	  2 - 5GHz
+	  4 - 6GHz
+	  7 - 2.4GHz and 5GHz both
```
rado17, line 426:
```suggestion
	  7 - All ( 2.4GHz, 5GHz and 6GHz)
```

rado17 (changes requested):

rado17 (approved):

drivers/wifi/nrf700x/Kconfig
```diff
@@ -401,3 +401,26 @@ config NRF_WIFI_BEAMFORMING
 config WIFI_NRF700X_SCAN_TIMEOUT_S
 	int "Scan timeout in seconds"
 	default 30
+
+config NRF_WIFI_2G_BAND
+	bool ""
+	default y if BOARD_NRF7001
+
+config NRF_WIFI_5G_BAND
```
krish2718, line 409: Why the empty help text? I guess this is the user facing option, no?

krish2718 (changes requested):

drivers/wifi/nrf700x/Kconfig
```diff
@@ -401,3 +401,26 @@ config NRF_WIFI_BEAMFORMING
...
+	default 1 if NRF_WIFI_2G_BAND
+	default 2 if NRF_WIFI_5G_BAND
+	default 4 if NRF_WIFI_6G_BAND
+	default 7
+	help
+	  Set this option to select frequency band
+	  1 - 2.4GHz
+	  2 - 5GHz
+	  4 - 6GHz
+	  7 - All ( 2.4GHz, 5GHz and 6GHz)
```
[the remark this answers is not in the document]
kapbh: Okay.

drivers/wifi/nrf700x/Kconfig
kapbh: Yes, Adding help text.

krga2022 (approved):

PR #12945, opened 2023-11-02
bluetooth: controller: add vs command set_power_control_request_params

knutel-nordic:
This HCI command will replace the existing commands vs_set_auto_power_control_request_param and
vs_set_power_control_apr_handling. The two existing commands are interdependent.


knutel-nordic:
This will be added to the changelog when the two existing commands are removed.

rugeGerritsen (approved):

KyraLengfeld (approved):

PR #12946, opened 2023-11-02
identity_key: Make identity key abstraction independent of HW backend

joerchan:
Add identity_key_free function which calls the
nrf_cc3xx_platform_identity_key_free function.
This makes the identity_key library a complete abstraction over the
nrf_cc3xx_platform library.

Make the identity_key_usage sample independent of the HW backend for
the identity key.
Remove the call for nrf_cc3xx_platform_init as this is handled by
existing initialization code, which also addresses initialization
without entropy.
Use identity_key library function for clearing identity key from
memory.

### DIFF
--- a/samples/keys/identity_key_usage/src/main.c
+++ b/samples/keys/identity_key_usage/src/main.c
@@ -29,12 +29,6 @@ int main(void)
 	uint8_t key[IDENTITY_KEY_SIZE_BYTES];
 	size_t olen;
 
-	err = nrf_cc3xx_platform_init();
-	if (err != NRF_CC3XX_PLATFORM_SUCCESS) {
-		LOG_INF("nrf_cc3xx_platform_init returned error: %d", err);
-		return 0;
-	}
-
 	LOG_INF("Initializing PSA crypto.");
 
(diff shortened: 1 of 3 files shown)


samples/keys/identity_key_usage/src/main.c
```diff
@@ -29,12 +29,6 @@ int main(void)
 	uint8_t key[IDENTITY_KEY_SIZE_BYTES];
 	size_t olen;
 
-	err = nrf_cc3xx_platform_init();
```
mswarowsky, line 32: Why can this removed here without showing up somewhere else?

joerchan: It's here: https://github.com/nrfconnect/sdk-nrf/blob/main/drivers/hw_cc310/hw_cc310.c#L27

mswarowsky: Ohh ok so it was there for nothing, thanks

mswarowsky (approved):

SebastianBoe (approved):

PR #12947, opened 2023-11-02
doc: nsib: edits to doc structure

greg-fer:
Moved some sections of the sample doc to more appropriate area. Added two links.
Fixed an incorrect link.
NCSIDB-1053.


samples/bootloader/README.rst
```diff
@@ -185,9 +185,10 @@ Building and running
 .. include:: /includes/build_and_run.txt
 
 .. caution::
-   The NSIB should be included as a child image in a multi-image build, rather than being built stand-alone.
+   The NSIB should be included as a child image in a :ref:`multi-image build <ug_multi_image>`, rather than being built stand-alone.
    While it is technically possible to build the NSIB by itself and merge it into other application images, this process is not supported.
    To reduce the development time and potential issues with this route, let the existing |NCS| infrastructure for multi-image builds handle the integration.
+   See :ref:`ug_multi_image_defining` for more information.
```
hellesvik-nordic, line 191: This documentation is intended for people who define their own child images. Users of NSIB will not be adding it as a multi image, but instead NSIB will automatically be included if they set CONFIG_SECURE_BOOT. I suggest that we do not mention this link, but instead say. "NSIB will automatically be added as a child image if CONFIG_SECURE_BOOT is set in an application.". I think that is the information users will need in practice.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -336,7 +336,7 @@ Hardware-based downgrade prevention
 
 Hardware-based downgrade prevention is implemented with a counter value stored in the *provision* partition of the non-volatile storage.
 
-See the :ref:`bootloader_monotonic_counter` section for how to enable and configure this feature.
+See the :ref:`ug_fw_update_image_versions` section for how to enable and configure this feature.
```
hellesvik-nordic, line 339: Is not this a subsection of the Configuring image versions section?

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -259,51 +259,60 @@ Recompile with ``priv_c.pem`` and the incremented firmware version to correctly
...
-=============
+Configuring image version with MCUboot
+======================================
 
-MCUboot supports two different implementations of downgrade protection:
+To assign a semantic version number to your application when you have opted for :ref:`ug_bootloader_adding_immutable_mcuboot` or :ref:`ug_bootloader_adding_upgradable_mcuboot`, add the version string to the :kconfig:option:`CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION` option for the application:
 
-* Software-based prevention, where the current version can be encoded into the currently active image.
-* Hardware-based prevention, where the current version can be encoded into a non-volatile storage partition outside of any image.
+.. code-block:: console
```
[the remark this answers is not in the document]
greg-fer: Added as a comment to https://nordicsemi.atlassian.net/browse/NCSIDB-964

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -331,16 +340,56 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
...
+.. bootloader_monotonic_counter_nsib_end
+
+.. _ug_fw_update_image_versions_mcuboot:
+
+Using MCUboot
+=============
+
+For MCUboot, the configuration options for implementing the anti-rollback protection with monotonic counter are the following:
+
+* :kconfig:option:`CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION`
```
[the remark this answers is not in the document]
greg-fer: @sigvartmh , would you know?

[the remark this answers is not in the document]
sigvartmh: I've not touched this configuration, so I'm unsure. I would preface it with that it probably does not work with (mcuboot image slot > 2) network core upgrades, as the last time someone tried this, it did not work.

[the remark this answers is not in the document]
MarekPieta: So that the rolling counter is updated next to the version and both need to be maintained by the user?

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -331,16 +353,58 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
    The optional label or build number specified after the ``+`` character is ignored when evaluating the version.
    An existing application image with version ``0.1.2+3`` can be overwritten by an uploaded image with ``0.1.2+2``, but not by one with ``0.1.1+2``.
```
hellesvik-nordic, line 354:
```suggestion
   For example, an existing application image with version ``0.1.2+3`` can be overwritten by an uploaded image with ``0.1.2+2``, but not by one with ``0.1.1+2``.
```

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -331,16 +353,58 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
...
+
+Hardware-based downgrade protection
+===================================
+
+.. bootloader_monotonic_counter_start
+
+You can implement hardware-based downgrade protection using a non-volatile monotonic counter.
+
+Counter updates are written to slots in the *Provision* area, with each new counter update occupying a new slot.
+For this reason, the number of counter updates, and therefore firmware version updates, is limited.
```
hellesvik-nordic, line 367: As a reader, I find myself wondering what this "Provision area" is. Is it part of non-volatile memory? How can I find out where it is? Do we have documentation referring to this area somewhere else?
```suggestion
For this reason, the number of counter updates, and therefore firmware version updates, is limited. The *Provision* is a partition in non-volatile memory, and its location can be found using :ref:`Partition Manager <partition_manager_report>`.
```

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -331,16 +353,58 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
...
+
+.. _ug_fw_update_hw_downgrade_nsib:
 
-Hardware-based downgrade prevention is implemented with a counter value stored in the *provision* partition of the non-volatile storage.
+Using |NSIB|
+============
+
+.. bootloader_monotonic_counter_nsib_start
+
+For |NSIB|, the configuration options for implementing the anti-rollback protection with monotonic counter are the following:
```
hellesvik-nordic, line 387: I think we should be clear that these options are intended for the application and not for the child image. The same for the below configurations for MCUboot.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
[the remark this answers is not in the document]
sigvartmh: The way it work(I think) I've not looked at this part of MCUBoot in a while is that once you update with a firmware version which is greater than your current version and is proven working the rolling counter is incremented. This means you will not be able to rollback to the previous version.

I also think this only works with XiP and it rotates between the slots primary/secondary depending on the rolling counter. So the user doesn't need to maintain the rolling counter only semantic versioning of MCUBoot.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
greg-fer: Moved.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -329,18 +351,63 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
...
+.. bootloader_monotonic_counter_nsib_end
+
+To set options for child images, see the :ref:`ug_multi_image_variables` section.
+
+.. _ug_fw_update_hw_downgrade_mcuboot:
+
+Using MCUboot
+=============
+
+For MCUboot, the configuration options for implementing the anti-rollback protection with monotonic counter in your application are the following:
```
hellesvik-nordic, line 400: Tricky thing is that the feature is for MCUboot child image, but the configuration is to be set for the application parent image. Can I suggest that we rephrase this:
```suggestion
To enable anti-rollback protection with monotonic counter for MCUboot, set the following configurations in the application (parent image):
```

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -329,18 +351,63 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
...
+See :ref:`ug_bootloader_adding_presigned_variants` for details.
+
+.. _ug_fw_update_hw_downgrade_nsib:
+
+Using |NSIB|
+============
+
+.. bootloader_monotonic_counter_nsib_start
+
+For |NSIB|, the configuration options for implementing the anti-rollback protection with monotonic counter in your application are the following:
```
hellesvik-nordic, line 386: Tricky thing is that the feature is for NSIB child image, but the configuration is to be set for the application parent image. Can I suggest that we rephrase this:
```suggestion
To enable anti-rollback protection with monotonic counter for |NSIB|, set the following configurations in the application (parent image):
```

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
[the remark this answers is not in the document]
MarekPieta: Thanks for clarification. Could you add these few sentences to documentation, please? (we could do it in a separate PR)

greg-fer:
@sigvartmh , @MarekPieta , @hellesvik-nordic , can you review the PR once again?

MarekPieta (commented):
btw. There seem bot be other sections in bootloader sample that should eventually be moved to generic documentation (e.g. `Provisioning` ). Do we have plans to move them?

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
hellesvik-nordic: This is not "using the partition manager". However, a user would use the "parititon manager report" instead

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -329,18 +351,63 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
...
+The *Provision* is a partition in non-volatile memory, and its location can be found using :ref:`partition_manager`.
+
+Using a counter is optional and can be configured for the application using configuration options.
+You can also configure the supported number of updates, but the number is limited by the size of the *Provision* area and how much of that area is taken up by other features, like public key hashes.
+In addition, you can configure what firmware version of the image you want to boot.
+
+.. bootloader_monotonic_counter_end
+
+Special handling is needed when updating the S1 variant of an image.
+See :ref:`ug_bootloader_adding_presigned_variants` for details.
```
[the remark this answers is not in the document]
greg-fer: @sigvartmh , can you check Marek's comment?

Looks like https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/config_and_build/bootloaders_and_dfu/bootloader_adding.html#generating-pre-signed-variants is wrongly referring to the original image as application image here: "this child image is an exact duplicate of the original image"
image

According to Marek, the "original image" is in reality (when testing this scenario) the MCUboot image.

Please confirm and we might need an additional issue to fix this.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
greg-fer: Fixed.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -259,51 +259,73 @@ Recompile with ``priv_c.pem`` and the incremented firmware version to correctly
...
+As part of this process, the bootloader compares the new firmware image version with the version of the currently booted firmware.
+The bootloader rejects the update if the new image version is lower than the version of the currently booted firmware.
 
-To assign a semantic version number to your application, pass the version string into the :kconfig:option:`CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION` option for the application:
+The implementation is related to the :ref:`ug_fw_update_image_versions` and also depends on the type of bootloader used:
 
-.. code-block:: console
+* Software-based downgrade protection is supported only by MCUboot.
+  In this kind of protection, the current firmware version is encoded into the currently active image.
+* Hardware-based downgrade protection is supported by both MCUboot and |NSIB|.
```
sigvartmh, line 317: As far as I know it's not supported when doing multi-image 4 slots with network core in MCUBoot I don't know if we need to mention this here.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -329,18 +351,63 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
 .. note::
 
    The optional label or build number specified after the ``+`` character is ignored when evaluating the version.
-   An existing application image with version ``0.1.2+3`` can be overwritten by an uploaded image with ``0.1.2+2``, but not by one with ``0.1.1+2``.
+   For example, an existing application image with version ``0.1.2+3`` can be overwritten by an uploaded image with ``0.1.2+2``, but not by one with ``0.1.1+2``.
```
sigvartmh, line 354: Also I think build number(what  comes after the + ) doesn't count when you do this comparison see: https://github.com/nrfconnect/sdk-mcuboot/blob/adab597a0eb0eb9c030a7b797748a49ca89988c2/boot/bootutil/src/loader.c#L685 
You need to enable a special option for that.

Don't know if this is important to mention

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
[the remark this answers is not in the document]
sigvartmh: FYI: @MarekPieta @greg-fer I see that it works for swaps too. The user doesn't need to maintain the security counter themselves if you don't provide a value as it would be calculated based on the version number see:

https://github.com/nrfconnect/sdk-mcuboot/blob/adab597a0eb0eb9c030a7b797748a49ca89988c2/scripts/imgtool/image.py#L185

If this option is set CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION_COUNTER_VALUE 
this value is provided:  set(imgtool_security_counter --security-counter ${CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION_COUNTER_VALUE})

Maybe just keep it as auto and recommend the users to have versioning 0.0.1 on their first software and only increase revision. Sounds a bit weird as you wouldn't keep a semantic versioning scheme with this.

https://github.com/nrfconnect/sdk-mcuboot/blob/adab597a0eb0eb9c030a7b797748a49ca89988c2/scripts/imgtool/main.py#L256C4-L256C4

Related MCUBoot code:
https://github.com/nrfconnect/sdk-mcuboot/blob/adab597a0eb0eb9c030a7b797748a49ca89988c2/boot/bootutil/src/loader.c#L923

So my bad seems that the counter value is being used just set by the script automatically.

sigvartmh (commented):
Looks good to me just added some NIT and a comment.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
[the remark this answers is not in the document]
greg-fer: This will be handled in https://nordicsemi.atlassian.net/browse/NCSIDB-1121

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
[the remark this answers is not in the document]
greg-fer: Does this mean we have to change the CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION option to be enabled by default ("maybe just keep it as auto")?

Also, if we need to update the information about image version, we need to do this in the "Configuring image version with MCUboot" section.

@MarekPieta , @sigvartmh ?

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
greg-fer: So if I enable CONFIG_MCUBOOT_VERSION_CMP_USE_BUILD_NUMBER, ``0.1.2+2`` won't overwrite ``0.1.2+3``?

hellesvik-nordic (approved):

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
MarekPieta: We could I think, I suggest to handle it in a follow-up task. What you think @greg-fer ?

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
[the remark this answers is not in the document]
MarekPieta: > Does this mean we have to change the CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION option to be enabled by default ("maybe just keep it as auto")?
> 
> Also, if we need to update the information about image version, we need to do this in the "Configuring image version with MCUboot" section.
> 
> @MarekPieta , @sigvartmh ?

Enabling MCUboot downgrade prevention by default would change current default behavior and may be a bit confusing for end users (if we opt for it, we would need to clearly inform about the change in a visible place I think).

Updating doc may be a good idea (either in this or subsequent PR).

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
MarekPieta: > So if I enable CONFIG_MCUBOOT_VERSION_CMP_USE_BUILD_NUMBER, `0.1.2+2` won't overwrite `0.1.2+3`?

IIRC yes. We may mention it in doc.

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
```diff
@@ -329,18 +352,63 @@ If this image has, in order of precedence, a *major*, *minor*, or *revision* val
...
+.. bootloader_monotonic_counter_end
+
+.. _ug_fw_update_hw_downgrade_nsib:
+
+Downgrade protection using |NSIB|
+=================================
+
+.. bootloader_monotonic_counter_nsib_start
+
+To enable anti-rollback protection with monotonic counter for |NSIB|, set the following configurations in the application (parent image):
```
MarekPieta, line 384: I am not sure if this does not need to be set for both bootloader and application. @sigvartmh, could you confirm, please?

MarekPieta (approved):

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
sigvartmh: For this setting it is enough to just set it in the application(parent image).

sigvartmh (approved):

doc/nrf/config_and_build/bootloaders_and_dfu/fw_update.rst
MarekPieta: Thanks for the clarification. In that case I think we are ready to merge this one and further improvements can be added in separate PRs.

PR #12948, opened 2023-11-02
doc: Change securely generating credentials link

divipillai:
Change the securely generated credentials link
to match with the cloud document, as now we have
support for nRF9161.

NCSDK-23335

### DIFF
--- a/doc/nrf/links.txt
+++ b/doc/nrf/links.txt
@@ -1026,7 +1026,7 @@
 .. _`nRF Cloud FOTA`: https://docs.nrfcloud.com/Devices/FirmwareUpdate/FOTAOverview.html
 .. _`nRF Cloud MQTT FOTA`: https://docs.nrfcloud.com/Devices/FirmwareUpdate/FOTAOverview.html#mqtt-job-execution-notifications
 .. _`nRF Cloud Getting Started FOTA documentation`: https://docs.nrfcloud.com/Devices/FirmwareUpdate/FOTATutorial.html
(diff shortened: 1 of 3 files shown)


doc/nrf/links.txt
```diff
@@ -1018,7 +1018,7 @@
 .. _`nRF Cloud FOTA`: https://docs.nrfcloud.com/Devices/FirmwareUpdate/FOTAOverview.html
 .. _`nRF Cloud MQTT FOTA`: https://docs.nrfcloud.com/Devices/FirmwareUpdate/FOTAOverview.html#mqtt-job-execution-notifications
 .. _`nRF Cloud Getting Started FOTA documentation`: https://docs.nrfcloud.com/Devices/FirmwareUpdate/FOTATutorial.html
-.. _`Securely generating credentials on the nRF9160`: https://docs.nrfcloud.com/Devices/Security/Credentials.html
+.. _`Securely generating credentials`: https://docs.nrfcloud.com/Devices/Security/Credentials.html
```
wiba-nordic, line 1021: The exact new name of the page is "Securely generating credentials for nRF91 Series devices". Do we want to match it exactly? I don't think it's necessary, but thought I'd ask just in case.

divipillai: My initial thought was to provide the full name, then I changed and just kept the term to be more generic. I don't have a strong opinion on this. Please let me know if I need to change.

wiba-nordic: Keeping it generic does make it less likely we would need to adjust link text at a later date, so I think it's fine as is.

wiba-nordic (approved):

peknis (approved):

PR #12951, opened 2023-11-02
applications: nrf5340_audio: Ignore invalid duplicated SDUs

alexsven:
- If a stream is started before data has been sent to the controller, the receiver will receive NULL packets with timestamp=0. These should not print a duplicated sdu_ref warning

### DIFF
--- a/applications/nrf5340_audio/src/audio/audio_datapath.c
+++ b/applications/nrf5340_audio/src/audio/audio_datapath.c
@@ -820,7 +820,7 @@ void audio_datapath_stream_out(const uint8_t *buf, size_t size, uint32_t sdu_ref
 		LOG_ERR("buf is NULL");
 	}
 
-	if (sdu_ref_us == ctrl_blk.previous_sdu_ref_us) {
+	if (sdu_ref_us == ctrl_blk.previous_sdu_ref_us && sdu_ref_us != 0) {
 		LOG_WRN("Duplicate sdu_ref_us (%d) - Dropping audio frame", sdu_ref_us);
 		return;
 	}


koffes (approved):

applications/nrf5340_audio/src/audio/audio_datapath.c
```diff
@@ -820,7 +820,7 @@ void audio_datapath_stream_out(const uint8_t *buf, size_t size, uint32_t sdu_ref
 		LOG_ERR("buf is NULL");
 	}
 
-	if (sdu_ref_us == ctrl_blk.previous_sdu_ref_us) {
+	if (sdu_ref_us == ctrl_blk.previous_sdu_ref_us && sdu_ref_us != 0) {
```
erikrobstad, line 823: Hmmm, should we return if sdu_ref_us == 0?

erikrobstad: Or maybe another solution:
If the receiver receive NULL packages in this scenario, does that mean buf == NULL?
If so, we could return here (after LOG_ERR("buf is NULL");)?

erikrobstad (changes requested):

applications/nrf5340_audio/src/audio/audio_datapath.c
alexsven: What happens is that all the SDUs comes in with bad_frame set, so the PLC will kick in. I believe that is the correct behaviour in case this happens mid-stream

rick1082 (approved):

applications/nrf5340_audio/src/audio/audio_datapath.c
erikrobstad: So the bad_frame flag is also set before audio is started over USB?
In the scenario where stream(s) are started, but no data is received over USB yet

alexsven: Correct

erikrobstad (approved):

PR #12953, opened 2023-11-02
wifi: MbedTLS builtin support

krish2718:
Allow WPA supplicant crypto to work with builtin MbedTLS to avoid NCS dependencies, useful for customers using non-nRF SoC's and also for upstream Zephyr WPA supplicant.


ehughes:
Hello.  

I wanted to mention that I just verified this configuration.   I have an nRF7002-EK wired to a custom board based upon the NXP i.MXRT685.

image

 
nrf7002

rado17 (approved):

VivekUppunda (approved):

PR #12954, opened 2023-11-02
wifi: Fix L2AC regression

krish2718:


modules/hostap/src/supp_api.c
```diff
@@ -344,13 +364,22 @@ int z_wpa_supplicant_connect(const struct device *dev,
...
+		if (params->band != WIFI_FREQ_BAND_UNKNOWN) {
+			freq = wpa_supp_band_chan_compat(wpa_s, params->band, params->channel);
+			if (freq < 0) {
+				_wpa_cli_cmd_v("remove_network %d", resp.network_id);
+				goto out;
+			}
+		} else {
+			freq = chan_to_freq(params->channel);
+			if (freq < 0) {
+				ret = -1;
```
rado17, line 378: Since we are returning error in this case, I think we can remove the network as is being done in case of valid band.

rado17 (changes requested):

modules/hostap/src/supp_api.c
krish2718: Moved the cleanup to goto to handle multiple exits.

modules/hostap/src/supp_api.c
```diff
@@ -194,6 +194,26 @@ static int wpa_supp_supported_channels(struct wpa_supplicant *wpa_s, uint8_t ban
...
+		wpa_printf(MSG_ERROR, "Unsupported or invalid band: %d", band);
+		return -EINVAL;
+	}
+
+	for (i = 0; i < mode->num_channels; i++) {
+		if (mode->channels[i].freq == channel) {
+			return mode->channels[i].freq;
+		}
+	}
+
```
sachinthegreen, line 213: Is it better to have a error log here to indicate band-channel compatibility check failure.

sachinthegreen (approved):

modules/hostap/src/supp_api.c
krish2718: Not only log, we need to propagate these error to the Application.

rlubos:
@krish2718 Please check the compliance failure
@rado17 Please revisit

rado17 (approved):

PR #12961, opened 2023-11-03
bluetooth: Add VS cmd for the connection event trigger feature

olivier-le-sage:
Connection event triggers are an old feature from SoftDevice which we've decided to port to NCS.

### DIFF
--- a/subsys/bluetooth/controller/Kconfig
+++ b/subsys/bluetooth/controller/Kconfig
@@ -345,6 +345,16 @@ config BT_CTLR_LE_POWER_CONTROL
 	  Enable support for LE Power Control feature that defined in the Bluetooth
 	  Core specification, Version 5.3 | Vol 6, Part B, Section 4.6.31.
 
+config BT_CTLR_SDC_EVENT_TRIGGER
+	bool "Event Trigger"
+	help
+	  Enable support for the Event Trigger feature.
+	  The event trigger is a proprietary SDC feature which can be used
+	  to perform tasks in synchronization with radio activity.
+	  When enabled, this feature will trigger a (D)PPI task at the start of radio events.
+	  When used for connections, the connection event trigger can be configured to trigger
+	  every N connection events starting from a given connection event counter.
+
 config BT_CTLR_SDC_ISO_RX_PDU_BUFFER_COUNT
 	int "BT_CTLR_SDC_ISO_RX_PDU_BUFFER_COUNT"
 	depends on BT_CTLR_SYNC_ISO || BT_CTLR_CONN_ISO
--- a/subsys/bluetooth/controller/hci_internal.c
+++ b/subsys/bluetooth/controller/hci_internal.c
@@ -644,6 +644,14 @@ static void vs_supported_commands(sdc_hci_vs_supported_vs_commands_t *cmds)
 #if defined(CONFIG_BT_CENTRAL)
 	cmds->central_acl_event_spacing_set = 1;
 #endif
+
+#if defined(CONFIG_BT_CTLR_SDC_EVENT_TRIGGER)
+	cmds->set_conn_event_trigger = 1;
+#endif
+
+#if defined(CONFIG_BT_CONN)
+	cmds->get_next_conn_event_counter = 1;
+#endif
 }
 #endif	/* CONFIG_BT_HCI_VS */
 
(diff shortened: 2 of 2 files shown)


subsys/bluetooth/controller/hci_internal.c
```diff
@@ -642,6 +642,8 @@ static void vs_supported_commands(sdc_hci_vs_supported_vs_commands_t *cmds)
 	cmds->set_power_control_request_params = 1;
 
 #endif
+	cmds->set_conn_event_trigger = 1;
```
rugeGerritsen, line 645: Add deps to BT_CONN?

rugeGerritsen (changes requested):
Missing dependencies. Otherwise it looks ok

subsys/bluetooth/controller/hci_internal.c
olivier-le-sage: I'm not sure I understand? This feature currently gets added to all SDC variants and can be used even without connections (for instance with scanning)

rugeGerritsen: get_next_conn_event_counter  is for connections only, right?

rugeGerritsen (approved):

knutel-nordic (approved):

subsys/bluetooth/controller/hci_internal.c
```diff
@@ -1565,6 +1570,15 @@ static uint8_t vs_cmd_put(uint8_t const * const cmd,
 #ifdef CONFIG_BT_CENTRAL
 	case SDC_HCI_OPCODE_CMD_VS_CENTRAL_ACL_EVENT_SPACING_SET:
 		return sdc_hci_cmd_vs_central_acl_event_spacing_set((void *)cmd_params);
+#endif
+	case SDC_HCI_OPCODE_CMD_VS_SET_CONN_EVENT_TRIGGER:
+		return sdc_hci_cmd_vs_set_conn_event_trigger((void *)cmd_params);
```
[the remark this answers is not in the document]
olivier-le-sage: I haven't checked. You mean something like CONFIG_BT_CTLR_CONN_EVENT_TRIGGER, right?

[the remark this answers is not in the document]
seko-nordic: yes. But better to check first, it may be neglectable

seko-nordic (changes requested):
need to check if unconditional flash increase is acceptable

subsys/bluetooth/controller/hci_internal.c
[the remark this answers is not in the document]
olivier-le-sage: [diff with without conn event trigger cmds.txt](https://github.com/nrfconnect/sdk-nrf/files/13316858/diff.with.without.conn.event.trigger.cmds.txt)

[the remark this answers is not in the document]
rugeGerritsen: So the connection event trigger is approximately 400 bytes. Then I believe it should be wrapped with a new Kconfig option

subsys/bluetooth/controller/Kconfig
```diff
@@ -345,6 +345,13 @@ config BT_CTLR_LE_POWER_CONTROL
 	  Enable support for LE Power Control feature that defined in the Bluetooth
 	  Core specification, Version 5.3 | Vol 6, Part B, Section 4.6.31.
 
+config BT_CTLR_SDC_CONN_EVENT_TRIGGER
+	bool "Connection Event Trigger"
+	help
+		Enable support for the Connection Event Trigger feature.
+		Connection event triggers are a proprietary SDC feature which can be used
+		to perform tasks in synchronization with radio activity.
+
```
rugeGerritsen, line 354:
```suggestion
config BT_CTLR_SDC_EVENT_TRIGGER
	bool "Event Trigger"
	help
	  Enable support for the Event Trigger feature.
	  The event trigger is  proprietary SDC feature which can be used
	  to perform tasks in synchronization with radio activity.
	  The feature triggers a PPI/DPPI task at the start of every N connection
	  event or start of a scanner/initiator event.
```

rugeGerritsen (changes requested):
Maybe we short rename this to "event trigger" as it can also be used for non-ACL events.
Suggest to add some more docs to make it more clear what this feature is about

subsys/bluetooth/controller/Kconfig
olivier-le-sage: I will copy some text from sdc_hci_vs.h.
We may extend this to other roles in the future so I prefer to keep the language generic ("start of radio events").

subsys/bluetooth/controller/Kconfig
```diff
@@ -345,6 +345,16 @@ config BT_CTLR_LE_POWER_CONTROL
 	  Enable support for LE Power Control feature that defined in the Bluetooth
 	  Core specification, Version 5.3 | Vol 6, Part B, Section 4.6.31.
 
+config BT_CTLR_SDC_EVENT_TRIGGER
+	bool "Event Trigger"
+	help
+		Enable support for the Event Trigger feature.
```
rugeGerritsen, line 351: Please fix the indentation here

rugeGerritsen (approved):

subsys/bluetooth/controller/hci_internal.c
```diff
@@ -1565,6 +1573,16 @@ static uint8_t vs_cmd_put(uint8_t const * const cmd,
 #ifdef CONFIG_BT_CENTRAL
 	case SDC_HCI_OPCODE_CMD_VS_CENTRAL_ACL_EVENT_SPACING_SET:
 		return sdc_hci_cmd_vs_central_acl_event_spacing_set((void *)cmd_params);
+#endif
+#if defined(CONFIG_BT_CTLR_SDC_EVENT_TRIGGER)
+	case SDC_HCI_OPCODE_CMD_VS_SET_CONN_EVENT_TRIGGER:
+		return sdc_hci_cmd_vs_set_conn_event_trigger((void *)cmd_params);
+#endif
+#if defined(CONFIG_BT_CONN)
+	case SDC_HCI_OPCODE_CMD_VS_GET_NEXT_CONN_EVENT_COUNTER:
```
seko-nordic, line 1582: can this be used for anything else than conn_event_trigger? I wander if it should go under CONFIG_BT_CTLR_SDC_EVENT_TRIGGER as well

olivier-le-sage: I'm not sure. Never really know what a customer might come up with. This command is also very small

seko-nordic (approved):

PR #12963, opened 2023-11-03
lib: pdn: unit tests and other improvements

eivindj-nordic:
* Do not attemt to free the pdn if already freed by another thread.
* PDP configuration is lost on modem side on cfun=0. We therefore has to
redo the configuration on CFUN state changes and not only on modem
initialization.
* Unit tests


gopiotr:
Please add to this PR commit with entry into CODEOWNERS file (https://github.com/nrfconnect/sdk-nrf/blob/main/CODEOWNERS) with path to test `/tests/lib/pdn/` and person/people responsible for this test.

lib/pdn/pdn.c
```diff
@@ -317,17 +321,25 @@ int pdn_default_ctx_cb_dereg(pdn_event_handler_t cb)
 	return 0;
 }
 
-static void pdn_ctx_free(struct pdn *pdn)
+static int pdn_ctx_free(struct pdn *pdn)
```
lemrey, line 325: This should probably remain `void`.

lib/pdn/pdn.c
```diff
@@ -347,13 +359,17 @@ int pdn_ctx_create(uint8_t *cid, pdn_event_handler_t cb)
 
 	err = nrf_modem_at_scanf("AT%XNEWCID?", "%%XNEWCID: %d", &ctx_id_tmp);
 	if (err < 0) {
-		pdn_ctx_free(pdn);
+		(void)pdn_ctx_free(pdn);
 		return err;
+	} else if (err == 0) {
+		/* no argument matched */
+		(void)pdn_ctx_free(pdn);
+		return -EFAULT;
```
lemrey, line 368: We shouldn't use `EFAULT` here, I suggest something else like `EBADMSG` to indicate that the parsing has failed.

lib/pdn/pdn.c
```diff
@@ -600,6 +621,20 @@ static void on_cfun(enum lte_lc_func_mode mode, void *ctx)
...
+		BUILD_ASSERT(sizeof(CONFIG_PDN_DEFAULT_USERNAME) > 1, "Username not defined");
+
+		err = pdn_ctx_auth_set(0, CONFIG_PDN_DEFAULT_AUTH,
+				CONFIG_PDN_DEFAULT_USERNAME,
+				CONFIG_PDN_DEFAULT_PASSWORD);
+		if (err) {
+			LOG_ERR("Failed to set auth params for default CID, err %d", err);
+			return;
+		}
+#endif /* CONFIG_PDN_DEFAULT_AUTH_PAP || CONFIG_PDN_DEFAULT_AUTH_CHAP */
```
lemrey, line 636: Hang on, in the `CFUN=1` callback, it might be too late to set these parameters, since they configure the default context itself. Should we have a `CFUN=0` callback instead? Otherwise we need to expose the `_init()` function. Might be good to do that regardless since the `pdn` library could be used without the `lte_lc` library.

eivindj-nordic: Yeah, that might be too late, indeed... Thanks!

lib/pdn/pdn.c
eivindj-nordic: Ok, as long as it is ok not to report an error if the context is already destroyed/does not exist, I am fine with that.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -363,7 +363,9 @@ Modem libraries
 
    * Added a mention about enabling TF-M logging while using modem traces in the :ref:`modem_trace_module`.
 
-|no_changes_yet_note|
+* :ref:`pdn_readme` library:
+
+  * Updated to fix a potential issue where the library tries to free the PDN context twice, causing the application to crash.
```
divipillai, line 368:
```suggestion
  * Fixed a potential issue where the library tries to free the PDN context twice, causing the application to crash.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -363,7 +363,10 @@ Modem libraries
 
    * Added a mention about enabling TF-M logging while using modem traces in the :ref:`modem_trace_module`.
 
-|no_changes_yet_note|
+* :ref:`pdn_readme` library:
+
+    * Fixed a potential issue where the library tries to free the PDN context twice, causing the application to crash.
+    * Updated the library to add PDP auto configuration to the :c:enumerator:`LTE_LC_FUNC_MODE_POWER_OFF` events.
```
divipillai, line 369:
```suggestion
* :ref:`pdn_readme` library:

   * Fixed a potential issue where the library tries to free the PDN context twice, causing the application to crash.
   * Updated the library to add PDP auto configuration to the :c:enumerator:`LTE_LC_FUNC_MODE_POWER_OFF` event.
```

divipillai (approved):

lib/pdn/pdn.c
```diff
@@ -215,15 +221,47 @@ static void parse_cgev_apn_rate_ctrl(const char *notif)
...
 	parse_cgev(notif);
 	parse_cgev_apn_rate_ctrl(notif);
 }
 
-NRF_MODEM_LIB_ON_INIT(modem_init_hook, on_modem_init, NULL);
+static void pdn_defaults_override(void)
+{
+	int err;
+
+	err = pdn_ctx_configure(0, CONFIG_PDN_DEFAULT_APN,
```
junqingzou, line 238: Here we hardcode CID=0 for "initial PDP context" as in TS 27.007.
But there are networks do allow CID=1, for example, when estanblishing primary PDN.

lib/pdn/Kconfig
```diff
@@ -29,6 +29,7 @@ config PDN_INIT_PRIORITY
 
 config PDN_DEFAULTS_OVERRIDE
 	bool "Override defaults for PDP context 0"
+	depends on LTE_LINK_CONTROL
```
junqingzou, line 32: I'd not to see this. 
Since libmodem init callback can still config PDN, there is definite need of dependency on lib_lte_lc.

lib/pdn/pdn.c
```diff
@@ -601,5 +639,11 @@ static void on_cfun(enum lte_lc_func_mode mode, void *ctx)
 			LOG_ERR("Unable to subscribe to +CGEREP=1, err %d", err);
 		}
 	}
+
+	if (mode == LTE_LC_FUNC_MODE_POWER_OFF) {
+#if defined(CONFIG_PDN_DEFAULTS_OVERRIDE)
+		pdn_defaults_override();
+#endif
```
junqingzou, line 646: This is a best-effort solution, i.e. if app does not use lib_lte_lc at all, it could still configure the default PDN by calling lib_pdn API.

lib/pdn/pdn.c
lemrey: We can't do much about that, I suggest we take this discussion with the modem team.
From what I understand, the initial PDP context has `cid=0`, from spec.
The network has nothing to do with CID numbers, which are just a modem-side handle for a PDP context.

junqingzou: Understood

junqingzou:
In order not to dependent on lib lte_lc, app can call pdn_ctx_configure() and pdn_ctx_auth_set() to config the default PDN before turning on radio.

junqingzou (approved):

PR #12966, opened 2023-11-03
Centralize FOTA support from nRF Cloud samples

plskeggs:
Move fota support code from the nRF Cloud REST FOTA and nRF Cloud Multi Service (CoAP mode) samples to an optional nrf_cloud_fota_poll.c file.

Allow CI to set CONFIG_REST_FOTA_DL_TIMEOUT_MIN, but use it to set the value of the newer FOTA_DL_TIMEOUT_MIN.

Rather than managing settings in nrf_cloud_fota.c and nrf_cloud_fota_poll.c, move them to one place in nrf_cloud_fota_common.c.

Also fix Coverity defect 201979.

Jira: IRIS-6793, IRIS-7159


samples/cellular/nrf_cloud_rest_fota/src/main.c
```diff
@@ -771,68 +417,23 @@ int main(void)
...
+			error_reboot(FOTA_REBOOT_SYS_ERROR);
 		}
 
-		err = wait_for_download();
-		if (err == -ETIMEDOUT) {
-			LOG_ERR("Timeout; FOTA download took longer than %d minutes",
-				CONFIG_REST_FOTA_DL_TIMEOUT_MIN);
-			fota_status = NRF_CLOUD_FOTA_TIMED_OUT;
-			fota_status_details = FOTA_STATUS_DETAILS_TIMEOUT;
+		err = nrf_cloud_fota_support_process(&fota_ctx);
```
jayteemo, line 425: since this is a sample, it might be good to add comments here describing what all this function does.
for example, it can block for awhile while the download occurs.

include/net/nrf_cloud.h
```diff
@@ -673,6 +673,28 @@ struct nrf_cloud_init_param {
...
+ * @brief  Event handler registered with the module to handle asynchronous
+ * events from the module.
+ *
+ * @param[in]  evt The event and any associated parameters.
+ */
+typedef void (*fota_reboot_handler_t)(enum nrf_cloud_fota_reboot_status status);
+
+struct nrf_cloud_fota_support_ctx {
+	struct nrf_cloud_rest_context *rest_ctx;
+	const char *device_id;
```
jayteemo, line 693: could the device_id buffer live in fota_support? 
and then `nrf_cloud_client_id_get()` could be called in `nrf_cloud_fota_support_init()`?

samples/cellular/nrf_cloud_rest_fota/prj.conf
```diff
@@ -1,6 +1,8 @@
 # Sample
 CONFIG_REST_FOTA_DO_JITP=n
-CONFIG_NRF_CLOUD_REST_FOTA_SAMPLE_LOG_LEVEL_INF=y
+CONFIG_NRF_CLOUD_REST_FOTA_SAMPLE_LOG_LEVEL_DBG=y
```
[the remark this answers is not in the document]
plskeggs: not for the final version, no

include/net/nrf_cloud.h
plskeggs: Yes, I could do that.  I did it this way to eliminate duplication of effort, but if you think moving it here is more important let me know.

jayteemo: if an application is only using nrf cloud for REST FOTA, i don't think it would be using the device id for anything else.

if an application is using other nrf cloud REST services, the fota support module, in its current form, might be a little awkward to use since the process function is blocking and using the shared rest context.

plskeggs: Your `nrf_cloud_rest_fota` sample prints out the device id; that's why I decided to reuse it rather than fetch it again.

The `nrf_cloud_fota_support_process()` function returns at the same points your code in `main()` used to loop around from.  If there is no update, for example, it returns right away.

It sounds like you would have preferred that the 8 smaller functions that `nrf_cloud_fota_support_process()` calls be made public.  That seems like a bigger maintenance burden than packaging it as I have.  Besides, we don't recommend people use REST for multiple services.

jayteemo: `Your nrf_cloud_rest_fota sample prints out the device id; that's why I decided to reuse it rather than fetch it again.`
ok, leave it up to the application.

`It sounds like you would have preferred that the 8 smaller functions `
fota_support in its current form is fine for the use case of doing only REST based FOTA in the sample, so for this ticket, no i would not prefer it be separated.

`we don't recommend people use REST for multiple services.`
interesting... i think there are use-cases which would allow it, but no need to discuss that here.

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_support.c
```diff
@@ -0,0 +1,554 @@
...
+ */
+#if defined(CONFIG_FOTA_USE_NRF_CLOUD_SETTINGS_AREA)
+#define FOTA_SETTINGS_NAME		NRF_CLOUD_SETTINGS_FULL_FOTA
+#define FOTA_SETTINGS_KEY_PENDING_JOB	NRF_CLOUD_SETTINGS_FOTA_JOB
+#define FOTA_SETTINGS_FULL		NRF_CLOUD_SETTINGS_FULL_FOTA_JOB
+#else
+#define FOTA_SETTINGS_NAME		CONFIG_FOTA_SETTINGS_NAME
+#define FOTA_SETTINGS_KEY_PENDING_JOB	CONFIG_FOTA_SETTINGS_KEY_PENDING_JOB
+#define FOTA_SETTINGS_FULL		FOTA_SETTINGS_NAME "/" FOTA_SETTINGS_KEY_PENDING_JOB
+#endif
```
glarsennordic, line 30: Should this be customizable specifically for these helpers?

I argue this is one of two built-in implementations of the FOTA job storage system, they should just straight up use the same KConfigs

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_support.c
```diff
@@ -0,0 +1,554 @@
...
+	if (sz == sizeof(pending_job)) {
+		LOG_INF("Saved job: %s, type: %d, validate: %d, bl: 0x%X",
+			pending_job.id, pending_job.type,
+			pending_job.validate, pending_job.bl_flags);
+	} else {
+		LOG_INF("FOTA settings size smaller than current, likely outdated");
+	}
+
+	return 0;
+}
```
glarsennordic, line 178: This is duplicate code with the implementation of `fota_settings_set` in `nrf_cloud_fota.c` right?

It might be nice to merge this with that over there. But it doesn't have to be for this PR. But if not this PR, can we write a ticket about it?

glarsennordic: ditto for other job storage functions

Maybe we need `nrf_cloud_fota_job.c` or something like that, with an internal private header. Again though, it's fine by me if you think that should be a followup PR

include/net/nrf_cloud.h
```diff
@@ -994,6 +1016,10 @@ bool nrf_cloud_fota_is_type_enabled(const enum nrf_cloud_fota_type type);
  */
 int nrf_cloud_fota_job_start(void);
 
+int nrf_cloud_fota_support_init(struct nrf_cloud_fota_support_ctx *ctx);
```
glarsennordic, line 1019: Since there will only ever be a single `nrf_cloud_fota_support` instance, maybe it would make sense to have `nrf_cloud_fota_support_init` copy the `ctx` struct to a private global context struct, and have `nrf_cloud_fota_support_start` and `nrf_cloud_fota_support_process` be parameter-less.

If you move in that direction, consider also renaming to `cfg` instead of `ctx`, and having it be pass by value rather than pass by pointer

glarsennordic (approved):
Overall looks great, so I am approving, but do consider my suggestions

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_support.c
plskeggs: I like the idea of moving it to `nrf_cloud_fota_job.c`.  I'm not certain I want to do it now, but maybe before the end of the sprint.

samples/cellular/nrf_cloud_multi_service/overlay_coap.conf
```diff
@@ -61,6 +61,7 @@ CONFIG_NRF_CLOUD_ALERT=y
 CONFIG_NRF_CLOUD_LOG_DIRECT=y
 CONFIG_NRF_CLOUD_LOG_OUTPUT_LEVEL=3
 CONFIG_NRF_CLOUD_LOG_LOG_LEVEL_INF=y
+CONFIG_NRF_CLOUD_FOTA_SUPPORT=y
```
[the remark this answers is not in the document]
plskeggs: I changed the Kconfig to `CONFIG_NRF_CLOUD_FOTA_POLL`.

I think it is better to only compile it when needed; otherwise this just adds to build time unnecessarily when it is not actually used.

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_support.c
plskeggs: I decided to consolidate the settings handling in nrf_cloud_fota_common.c, but leave the rest as it is for now.

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_poll.c
```diff
@@ -0,0 +1,473 @@
...
+	if (err) {
+		return err;
+	}
+
+#if defined(CONFIG_NRF_CLOUD_FOTA_FULL_MODEM_UPDATE)
+	struct dfu_target_fmfu_fdev fmfu_dev_inf = {
+		.size = 0,
+		.offset = 0,
+		.dev = IS_ENABLED(CONFIG_DFU_TARGET_FULL_MODEM_USE_EXT_PARTITION) ?
+		       NULL : DEVICE_DT_GET_ONE(jedec_spi_nor)
```
jayteemo, line 202: since this is in the common library now, the flash device name should be Kconfigurable?

plskeggs: It looks like ATv2 does not make it Kconfigurable -- they assume `jedec_spi_nor`.

jayteemo: Atv2 is for Nordic hardware.
This library should be usable by custom devices?

plskeggs: They intend people to use Atv2 as the basis for their products, not just for Nordic hardware.

plskeggs: I think it would be cleaner to stop supporting use of the external flash directly and just require using a partition, so the flash device can be managed elsewhere.

On the other hand, `subsys/net/lib/fota_download/src/util/fota_download_full_modem.c` does this:
`static const struct device *flash_dev = DEVICE_DT_GET_ANY(jedec_spi_nor);`
so maybe we should use the same if ext partitions are not used.

jayteemo: atv2 is an application and nrf_cloud is library.
a user shouldn't need to modify the library to get it to work with their product.

plskeggs: fota_download is a library as well

jayteemo: this PR is for nrf_cloud library.

if you don't want to make it configurable just say no and add a note in the docs.

jayteemo: that was easy 😅

include/net/nrf_cloud.h
plskeggs: I modeled this after what Justin did in the nrf_cloud_rest library, which accepts `struct nrf_cloud_fota_poll_ctx *` to be passed to every API.  Since this library uses that library (when built for REST), I'd like to leave it this way.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -249,10 +249,12 @@ Cellular samples (renamed from nRF9160 samples)
   * Changed:
 
     * The sample now explicitly uses the :c:func:`conn_mgr_all_if_connect` function to start network connectivity, instead of the :kconfig:option:`CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_START` and :kconfig:option:`CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_CONNECT` Kconfig options.
+    * The sample to use the FOTA support functions in the :file:`nrf_cloud_fota_poll.c` file and :file:`nrf_cloud_fota_common.c` file.
```
peknis, line 252:
```suggestion
    * The sample to use the FOTA support functions in the :file:`nrf_cloud_fota_poll.c` and :file:`nrf_cloud_fota_common.c` files.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -249,10 +249,12 @@ Cellular samples (renamed from nRF9160 samples)
...
 
 * :ref:`nrf_cloud_rest_fota` sample:
 
   * Added credential check before connecting to network.
+  * Changed the sample use the functions in the :file:`nrf_cloud_fota_poll.c` file and :file:`nrf_cloud_fota_common.c` file.
```
peknis, line 257:
```suggestion
  * Changed the sample use the functions in the :file:`nrf_cloud_fota_poll.c` and :file:`nrf_cloud_fota_common.c` files.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -510,6 +512,10 @@ Libraries for networking
 
   * Updated the device mode callback to send an event when the provisioning state changes.
 
+* :ref:`lib_nrf_cloud_fota` library:
+
+  * Added the :file:`nrf_cloud_fota_poll.c` to consolidate the FOTA polling code from the :ref:`nrf_cloud_multi_service` and :ref:`nrf_cloud_rest_fota` samples.
```
peknis, line 517:
```suggestion
  * Added the :file:`nrf_cloud_fota_poll.c` file to consolidate the FOTA polling code from the :ref:`nrf_cloud_multi_service` and :ref:`nrf_cloud_rest_fota` samples.
```

subsys/net/lib/nrf_cloud/include/nrf_cloud_fota.h
```diff
@@ -213,6 +213,24 @@ int nrf_cloud_fota_ble_job_update(
 	const struct nrf_cloud_fota_ble_job * const ble_job,
 	const enum nrf_cloud_fota_status status);
 
+/** @brief Initialize FOTA settings handling
```
peknis, line 216:
```suggestion
/** @brief Initialize FOTA settings handling.
```

subsys/net/lib/nrf_cloud/include/nrf_cloud_fota.h
```diff
@@ -213,6 +213,24 @@ int nrf_cloud_fota_ble_job_update(
...
+ *
+ *  @param job Pointer to FOTA job settings structure. The pointer must remain valid.
+ *
+ *  @retval 0 If successful.
+ *           Otherwise, a (negative) error code is returned.
+ */
+int nrf_cloud_fota_settings_load(struct nrf_cloud_settings_fota_job *job);
+
+/** @brief Save FOTA job info
```
peknis, line 225:
```suggestion
/** @brief Save FOTA job info.
```

peknis (changes requested):

peknis (approved):

PR #12971, opened 2023-11-06
net: nrf_provisioning: Remove attestation token

juhaylinen:
Remove authentication with attestation token.
Update test cases and documentation.


SeppoTakalo (approved):

samples/cellular/nrf_provisioning/prj.conf
```diff
@@ -74,6 +66,8 @@ CONFIG_MODEM_INFO_ADD_SIM=n
 
 # Modem key management
 CONFIG_MODEM_KEY_MGMT=y
+CONFIG_MODEM_JWT=y
+CONFIG_MODEM_ATTEST_TOKEN=y
```
melwee01, line 70: Is the sample moving from using the attestation token to JWT only? What are the implications for the user?

Also, why is the modem_attest_token config still set in this case?

I'm confused and want to make sure I understand.

juhaylinen: After this change the nrf_provisioning library will use only JWT for authentication.

CONFIG_NRF_PROVISIONING_JWT and CONFIG_NRF_PROVISIONING_ATTESTTOKEN needs to be removed from the user's application .conf file.

CONFIG_MODEM_ATTEST_TOKEN is enabled for provisioning shell command to read the attestation token from modem. The token is still used for claiming the device. It's only for testing.

jayteemo (approved):

jheiskan81 (approved):

samples/cellular/nrf_provisioning/prj.conf
plskeggs: @juhaylinen if CONFIG_MODEM_ATTEST_TOKEN is only for testing, how will customers using this service in production gather the identity attestation tokens for later claiming?  I realize this is out of scope of this PR, but curious what the plan is.

plskeggs (approved):
@juhaylinen have you tested this with an unclaimed device and confirmed it can be onboarded to nRF Cloud?

melwee01 (approved):

samples/cellular/nrf_provisioning/prj.conf
juhaylinen: I don't know what the plan is but it's possible to use AT command to read the identity attestation token.

juhaylinen:
> @juhaylinen have you tested this with an unclaimed device and confirmed it can be onboarded to nRF Cloud?

No, I haven't tested full onboarding with nRF Cloud yet.

PR #12980, opened 2023-11-06
bluetooth: controller: hci_driver: update iso rx pdu pool macro

sean-madigan:
Update to reflect change in sdc memory macro.

### DIFF
--- a/subsys/bluetooth/controller/Kconfig
+++ b/subsys/bluetooth/controller/Kconfig
@@ -359,6 +359,12 @@ config BT_CTLR_SDC_EVENT_TRIGGER
 	  When used for connections, the connection event trigger can be configured to trigger
 	  every N connection events starting from a given connection event counter.
 
+config BT_CTLR_SDC_ISO_RX_PDU_BUFFER_PER_STREAM_COUNT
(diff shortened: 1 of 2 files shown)


rugeGerritsen (approved):

subsys/bluetooth/controller/Kconfig
```diff
@@ -359,6 +359,12 @@ config BT_CTLR_SDC_EVENT_TRIGGER
 	  When used for connections, the connection event trigger can be configured to trigger
 	  every N connection events starting from a given connection event counter.
 
+config BT_CTLR_SDC_ISO_RX_PDU_BUFFER_PER_STREAM_COUNT
+	int "BT_CTLR_SDC_ISO_RX_PDU_BUFFER_PER_STREAM_COUNT"
+	depends on BT_CTLR_SYNC_ISO || BT_CTLR_CONN_ISO
+	range 0 10
+	default 1
```
alexstanoev-nordic, line 366: 3 might be a better default here?
This will keep the same behavior as we currently have via `BT_CTLR_SDC_ISO_RX_PDU_BUFFER_COUNT` when only configuring one RX stream.

alexstanoev-nordic (approved):

PR #12983, opened 2023-11-07
lib: lte_link_control: Support for PSM config in seconds

trantanen:
A new function for giving periodic TAU and active time in seconds. The library will handle conversion to bit field string representation specified in AT Commands Reference Guide, and 3GPP 24.008 Ch. 10.5.7.4a and Ch. 10.5.7.3.

Jira: NCSDK-23286

### DIFF
--- a/include/modem/lte_lc.h
+++ b/include/modem/lte_lc.h
@@ -1343,14 +1343,50 @@ int lte_lc_normal(void);
  * Set modem PSM parameters.
  *
  * Requested periodic TAU (RPTAU) and requested active time (RAT) are used when PSM mode is
- * subsequently enabled using lte_lc_psm_req().
+ * subsequently enabled using lte_lc_psm_req(). These are the requested values and network
+ * determines the final values.
  *
- * For reference see 3GPP 27.007 Ch. 7.38.
+ * For encoding of the variables, see nRF AT Commands Reference Guide, 3GPP 27.007 Ch. 7.38., and
+ * 3GPP 24.008 Ch. 10.5.7.4a and Ch. 10.5.7.3.
  *
- * @param[in] rptau Requested periodic TAU as a null-terminated string.
- *                  Set to @c NULL to use manufacturer-specific default value.
- * @param[in] rat Requested active time as a null-terminated string.
- *                Set to @c NULL to use manufacturer-specific default value.
+ * @param[in] rptau
+ * @parblock
+ *          Requested periodic TAU as a null-terminated 8 character long bit field string.
+ *          Set to @c NULL to use modem's default value.
+ *          See 3GPP 24.008 Ch. 10.5.7.4a for data format.
+ *
+ *          For example, value of 32400 s is represented as '00101001'.
+ *
+ *          Bits 5 to 1 represent the binary coded timer value that is multiplied by timer unit.
+ *
+ *          Bits 8 to 6 define the timer unit as follows:
+ *
+ *          - 000: 10 minutes
+ *          - 001: 1 hour
+ *          - 010: 10 hours
+ *          - 011: 2 seconds
+ *          - 100: 30 seconds
+ *          - 101: 1 minute
+ *          - 110: 320 hours
+ * @endparblock
+ *
+ * @param[in] rat
+ * @parblock
+ *          Requested active time as a null-terminated string.
+ *          Set to @c NULL to use modem's default value.
+ *          See 3GPP 24.008 Ch. 10.5.7.3 for data format.
+ *
+ *          For example, value of 120 s is represented as '00100010'.
+ *
+ *          Bits 5 to 1 represent the binary coded timer value that is multiplied by timer unit.
+ *
+ *          Bits 8 to 6 define the timer unit as follows:
+ *
+ *          - 000: 2 seconds
+ *          - 001: 1 minute
+ *          - 010: 6 minutes
+ *          - 111: Timer is deactivated
+ * @endparblock
  *
  * @retval 0 if successful.
  * @retval -EINVAL if an input parameter was invalid.
@@ -1358,10 +1394,56 @@ int lte_lc_normal(void);
 int lte_lc_psm_param_set(const char *rptau, const char *rat);
 
 /**
+ * Set modem PSM parameters.
+ *
+ * Requested periodic TAU (RPTAU) and requested active time (RAT) are used when PSM mode is
+ * subsequently enabled using lte_lc_psm_req(). These are the requested values and network
+ * determines the final values.
+ *
+ * This is a convenience function to set PSM parameters in seconds while lte_lc_psm_param_set()
+ * requires the caller to encode the values. However, note that 3GPP specifications do
+ * not support all possible integer values but the encoding limits the potential values.
+ * The values are rounded up to the next possible value. There may be significant
+ * rounding up, especially when rptau is tens and hundreds of hours.
+ *
+ * For more information about the encodings, see the description of the
+ * lte_lc_psm_param_set() function.
+ *
+ * @param[in] rptau
+ * @parblock
+ *          Requested periodic TAU in seconds as a non-negative integer. Range 0 - 35712000 s.
+ *          Set to @c -1 to use modem's default value.
+ *          The given value will be rounded up to the closest possible value that is
+ *          calculated by multiplying the timer unit with the timer value:
+ *
+ *          - Timer unit is one of: 2 s, 30 s, 60 s, 600 s (10 min), 3600 s (1 h), 36000 s (10 h),
+ *            1152000 s (320 h)
+ *          - Timer value range is 0-31
+ * @endparblock
+ *
+ * @param[in] rat
+ * @parblock
+ *          Requested active time in seconds as a non-negative integer. Range 0 - 11160s.
+ *          Set to @c -1 to use modem's default value.
+ *          The given value will be rounded up to the closest possible value that is
+ *          calculated by multiplying the timer unit with the timer value:
+ *
+ *          - Timer unit is one of: 2 s, 30 s, 360 s (6 min)
+ *          - Timer value range is 0-31
+ * @endparblock
+ *
+ * @retval 0 if successful.
+ * @retval -EINVAL if an input parameter was invalid.
+ */
+int lte_lc_psm_param_set_seconds(int rptau, int rat);
+
+/**
  * Request modem to enable or disable Power Saving Mode (PSM).
  *
- * PSM parameters can be set using @kconfig{CONFIG_LTE_PSM_REQ_RPTAU} and
- * @kconfig{CONFIG_LTE_PSM_REQ_RAT}, or by calling lte_lc_psm_param_set().
+ * PSM parameters can be set using @kconfig{CONFIG_LTE_PSM_REQ_FORMAT},
+ * @kconfig{CONFIG_LTE_PSM_REQ_RPTAU}, @kconfig{CONFIG_LTE_PSM_REQ_RAT},
+ * @kconfig{CONFIG_LTE_PSM_REQ_RPTAU_SECONDS} and @kconfig{CONFIG_LTE_PSM_REQ_RAT_SECONDS},
+ * or by calling lte_lc_psm_param_set() or lte_lc_psm_param_set_seconds().
  *
  * @note @kconfig{CONFIG_LTE_PSM_REQ} can be set to enable PSM, which is generally sufficient. This
  *       option allows explicit disabling/enabling of PSM requesting after modem initialization.
@@ -1379,9 +1461,9 @@ int lte_lc_psm_req(bool enable);
 /**
  * Get the current PSM (Power Saving Mode) configuration.
  *
- * @param[out] tau Periodic TAU interval in seconds. Positive integer,
+ * @param[out] tau Periodic TAU interval in seconds. A non-negative integer,
  *                 or @c -1 if timer is deactivated.
- * @param[out] active_time Active time in seconds. Positive integer,
+ * @param[out] active_time Active time in seconds. A non-negative integer,
  *                         or @c -1 if timer is deactivated.
  *
  * @retval 0 if successful.
@@ -1431,7 +1513,7 @@ int lte_lc_proprietary_psm_req(bool enable);
  *
  * @param[in] mode LTE mode to which the PTW value applies.
  * @param[in] ptw Paging Time Window value as a null-terminated string.
- *                Set to @c NULL to use manufacturer-specific default value.
+ *                Set to @c NULL to use modem's default value.
  *
  * @retval 0 if successful.
  * @retval -EINVAL if an input parameter was invalid.
@@ -1448,7 +1530,7 @@ int lte_lc_ptw_set(enum lte_lc_lte_mode mode, const char *ptw);
  *
  * @param[in] mode LTE mode to which the eDRX value applies.
  * @param[in] edrx eDRX value as a null-terminated string.
- *                 Set to @c NULL to use manufacturer-specific default.
+ *                 Set to @c NULL to use modem's default value.
  *
  * @retval 0 if successful.
  * @retval -EINVAL if an input parameter was invalid.
--- a/lib/lte_link_control/Kconfig
+++ b/lib/lte_link_control/Kconfig
@@ -69,9 +69,27 @@ config LTE_PSM_REQ
 	  the modem is initialized. This will cause the modem to include a PSM request in
 	  every subsequent LTE attach request.
 
+choice LTE_PSM_REQ_FORMAT
+	prompt "Format for PSM configuration"
+	default LTE_PSM_REQ_FORMAT_STRING
+	help
+	  Select the format in which default parameters for PSM are given.
+
+config LTE_PSM_REQ_FORMAT_STRING
+	bool "Use PSM configs given as bit field strings"
+	help
+	  Use LTE_PSM_REQ_RPTAU and LTE_PSM_REQ_RAT.
+
+config LTE_PSM_REQ_FORMAT_SECONDS
+	bool "Use PSM configs given in seconds"
+	help
+	  Use LTE_PSM_REQ_RPTAU_SECONDS and LTE_PSM_REQ_RAT_SECONDS.
+
+endchoice
+
 config LTE_PSM_REQ_RPTAU
 	string "Requested periodic TAU for PSM"
-	default "00000011"
+	default "00000011" # 30 minutes
 	help
 	  Power saving mode setting for requested periodic TAU.
 	  See 3GPP 27.007 Ch. 7.38.
@@ -79,11 +97,27 @@ config LTE_PSM_REQ_RPTAU
 
 config LTE_PSM_REQ_RAT
 	string "Requested active time for PSM"
-	default "00100001"
+	default "00100001" # 1 minute
 	help
 	  Power saving mode setting for requested active time.
 	  See 3GPP 27.007 Ch. 7.38.
 	  And 3GPP 24.008 Ch. 10.5.7.3 for data format.
+
+config LTE_PSM_REQ_RPTAU_SECONDS
+	int "Requested periodic TAU for PSM in seconds"
+	default 1800
+	help
+	  Power saving mode setting for requested periodic TAU in seconds.
+	  See lte_lc_psm_param_set_seconds() on limitations although any
+	  non-negative value can be given.
+
+config LTE_PSM_REQ_RAT_SECONDS
+	int "Requested active time for PSM in seconds"
+	default 60
+	help
+	  Power saving mode setting for requested active time in seconds.
+	  See lte_lc_psm_param_set_seconds() on limitations although any
+	  non-negative value can be given.
 
 config LTE_PROPRIETARY_PSM_REQ
 	bool "Enable use of proprietary PSM"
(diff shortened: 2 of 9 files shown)


lib/lte_link_control/lte_lc.c
```diff
@@ -67,12 +67,25 @@ static char edrx_param_nbiot[5] = CONFIG_LTE_EDRX_REQ_VALUE_NBIOT;
...
+#endif
+
+#if (CONFIG_LTE_PSM_REQ_RPTAU_SECONDS != -1)
+
+/* Default PSM RPTAU setting */
 static char psm_param_rptau[9] = CONFIG_LTE_PSM_REQ_RPTAU;
-/* Request PSM to be disabled */
+#else
+static char psm_param_rptau[9] = CONFIG_LTE_PSM_REQ_RPTAU_SECONDS;
+#endif
```
trantanen, line 84: This won't work at all and I have no idea how to implement those new Kconfigs nicely.

samples/cellular/modem_shell/src/link/link_shell.c
```diff
@@ -417,6 +423,8 @@ static struct option long_options[] = {
 	{ "uname", required_argument, 0, 'U' },
 	{ "rptau", required_argument, 0, 'p' },
 	{ "rat", required_argument, 0, 't' },
+	{ "srptau", required_argument, 0, 'P' },
+	{ "rsat", required_argument, 0, 'T' },
```
eivindj-nordic, line 427:
```suggestion
	{ "srat", required_argument, 0, 'T' },
```

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,85 @@ int lte_lc_normal(void);
...
+ * subsequently enabled using lte_lc_psm_req().
+ *
+ * This is a convenience function to set PSM parameters in seconds while lte_lc_psm_param_set()
+ * requires the caller to encode the values. However, notice that the underlying AT commands do
+ * not support all possible values but their encoding limits the potential values.
+ * The values are rounded up to the next possible value. Be aware that there may be significant
+ * rounding up especially when rptau is tens and hundreds of hours.
+ * More information about the encodings can be found from lte_lc_psm_param_set().
+ *
+ * @param[in] rptau Requested periodic TAU in seconds as a positive integer. Range 1 - 35711999s.
```
tokangas, line 1405: Why the upper limit is not 35712000? The lower limit could be 2 seconds, because that's the smallest number that can be encoded.

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,85 @@ int lte_lc_normal(void);
...
+ *                  Set to @c -1 to use manufacturer-specific default value.
+ *                  Given value will be rounded up to the closest possible value, which are
+ *                  represented by multiplying timer unit with a timer value:
+ *
+ *                  - Timer unit is one of: 2s, 30s, 60s, 600s (10min), 3600s (1h), 36000s (10h),
+ *                                          1152000s (320h)
+ *                  - Timer value range is 1-31
+ *
+ * @param[in] rat Requested active time in seconds as a positive integer. Range 1 - 11159s.
```
tokangas, line 1414: The upper limit should be 18600 seconds. And same comment for the lower limit as above.

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,85 @@ int lte_lc_normal(void);
...
+ *                Set to @c -1 to use manufacturer-specific default value.
+ *                Given value will be rounded up to the closest possible value, which are
+ *                represented by multiplying timer unit with a timer value:
+ *
+ *                - Timer unit is one of: 2s, 30s, 600s (10min)
+ *                - Timer value range is 1-31
+ *
+ * @retval 0 if successful.
+ * @retval -EINVAL if an input parameter was invalid.
```
tokangas, line 1423: One option would be, that the function would only accept values which can be encoded without rounding. If we anyway list the multiplier and the value, users would be able to select a suitable value. And the function could return `-EINVAL` if the value can not be encoded. Opinions? @trantanen @jtguggedal

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,85 @@ int lte_lc_normal(void);
...
- * For reference see 3GPP 27.007 Ch. 7.38.
+ * For encoding of the variables, see nRF AT commands Reference Guide, 3GPP 27.007 Ch. 7.38., and
+ * 3GPP 24.008 Ch. 10.5.7.4a and Ch. 10.5.7.3.
  *
- * @param[in] rptau Requested periodic TAU as a null-terminated string.
+ * @param[in] rptau Requested periodic TAU as a null-terminated 8 character long bit field string.
  *                  Set to @c NULL to use manufacturer-specific default value.
+ *                  See 3GPP 24.008 Ch. 10.5.7.4a for data format.
+ *
+ *                  Value of 12600s is represented as '00101001'.
```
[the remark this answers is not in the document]
trantanen: Yep, I updated the string because I wanted an example with non-zero time unit and naturally forgot to update the seconds.

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,85 @@ int lte_lc_normal(void);
...
 int lte_lc_psm_param_set(const char *rptau, const char *rat);
 
+/**
+ * Set modem PSM parameters.
+ *
+ * Requested periodic TAU (RPTAU) and requested active time (RAT) are used when PSM mode is
+ * subsequently enabled using lte_lc_psm_req().
+ *
+ * This is a convenience function to set PSM parameters in seconds while lte_lc_psm_param_set()
+ * requires the caller to encode the values. However, notice that the underlying AT commands do
```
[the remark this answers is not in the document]
trantanen: I'll change "AT commands" to "3GPP specifications"

include/modem/lte_lc.h
trantanen: Upper limit is 360s * 31 = 11160s

include/modem/lte_lc.h
trantanen: Yep, should be 35712000.

lib/lte_link_control/lte_lc.c
```diff
@@ -875,6 +888,33 @@ int lte_lc_psm_param_set(const char *rptau, const char *rat)
...
+int lte_lc_psm_param_set_seconds(int rptau, int rat)
+{
+	int ret;
+
+	if (rptau <= 0) {
+		*psm_param_rptau = '\0';
+		LOG_DBG("RPTAU use default");
+	}
+
+	if (rat <= 0) {
```
jtguggedal, line 900: I think we should support requesting 0 active time, it's a valid value

include/modem/lte_lc.h
jtguggedal: I actually liked the rounding approach. Maybe we could return a positive integer if it was rounded and 0 if it was exact?

lib/lte_link_control/lte_lc.c
jtguggedal: How about adding a Kconfig choice to select if binary string or seconds are used as default values. Setting the choice to binary string as default would maintain current behavior

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,96 @@ int lte_lc_normal(void);
...
+ * 3GPP 24.008 Ch. 10.5.7.4a and Ch. 10.5.7.3.
  *
- * @param[in] rptau Requested periodic TAU as a null-terminated string.
- *                  Set to @c NULL to use manufacturer-specific default value.
- * @param[in] rat Requested active time as a null-terminated string.
- *                Set to @c NULL to use manufacturer-specific default value.
+ * @param[in] rptau
+ * @parblock
+ *          Requested periodic TAU as a null-terminated 8 character long bit field string.
+ *          Set to @c NULL to use manufacturer-specific default value.
```
trantanen, line 1354: I guess we should be more specific here what manufacturer means here? User might be wondering if it's LTE LC Kconfig options, or modem defaults.

tokangas: Yes, wouldn't hurt to be make it more precise. The "manufacturer-specific default" wording is also used in the AT command specification which implies that it refers to modem defaults. But, it is confusing because LTE LC also has it's own default values.

include/modem/lte_lc.h
tokangas: I'm fine with this approach, just wanted to hear your opinions as well. If the user wants to check what was the value that was used, the debug traces will show it.

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -435,6 +435,141 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg)
...
+
+/* Lookup table for T3412-extended timer used for periodic TAU. Unit is seconds.
+ * Ref: GPRS Timer 3 in 3GPP TS 24.008 Table 10.5.163a/3GPP TS 24.008.
+ */
+static const uint32_t t3412_ext_lookup[8] = {600, 3600, 36000, 2, 30, 60, 1152000, 0};
+
+/* Lookup table for T3412 (legacy) timer used for periodic TAU. Unit is seconds.
+ * Ref: GPRS Timer in 3GPP TS 24.008 Table 10.5.172/3GPP TS 24.008.
+ */
+static const uint32_t t3412_lookup[8] = {2, 60, 360, 60, 60, 60, 60, 0};
```
tokangas, line 460: Could the tables be defined without fixed size:

`static const uint32_t t3324_lookup[] = {2, 60, 360};`

And the number of elements determined with the ARRAY_SIZE macro?

`encode_psm_timer(active_time_str, rat, t3324_lookup, ARRAY_SIZE(t3324_lookup));`

trantanen: > Could the tables be defined without fixed size:
> 
> `static const uint32_t t3324_lookup[] = {2, 60, 360};`
> 
> And the number of elements determined with the ARRAY_SIZE macro?
> 
> `encode_psm_timer(active_time_str, rat, t3324_lookup, ARRAY_SIZE(t3324_lookup));`

I'm not using the array size as a parameter for `encode_psm_timer()`. I could but it made sense to me to just handle the values that are different in the tables.

tokangas: But why there are numbers which are not used in the tables? I'm not following. 😄

trantanen: They are all used when parsing (decoding) to comply with 3GPP spec but they are not needed when encoding.
So yes, ARRAY_SIZE is already used in parsing but I'd rather use defines that I've introduced in the encoding side.

The tables are not new. `[8]` could be `[]` but I didn't change these from what they have originally been by @jtguggedal .

tokangas: Ok, I didn't see that those were already present earlier, just moved to a different place. I guess the decoding shouldn't need the extra values either, but the implementation would obviously have to be changed also there.

trantanen: Decoding needs those extra values to easily support the spec:
"Other values shall be interpreted as multiples of 1 minute in this version of the protocol."

lib/lte_link_control/lte_lc.c
trantanen: I've implemented a choice symbol and code into `init_and_config()` to handle this.

include/modem/lte_lc.h
trantanen: I've changed "manufacturer-specific" default value to "modem default value"

lib/lte_link_control/Kconfig
```diff
@@ -85,6 +85,22 @@ config LTE_PSM_REQ_RAT
 	  See 3GPP 27.007 Ch. 7.38.
 	  And 3GPP 24.008 Ch. 10.5.7.3 for data format.
 
+config LTE_PSM_REQ_RPTAU_SECONDS
+	int "Requested periodic TAU for PSM in seconds"
+	default 1800
+	help
+	  Power saving mode setting for requested periodic TAU in seconds.
+	  See 3GPP 27.007 Ch. 7.38.
+	  And 3GPP 24.008 Ch. 10.5.7.4a for data format.
```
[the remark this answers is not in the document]
trantanen: True. These are still copy paste and are now modified.

include/modem/lte_lc.h
trantanen: Leaving the current algorithm as is.

lib/lte_link_control/lte_lc_helpers.c
trantanen: Please resolve this if there are no further questions.

lib/lte_link_control/lte_lc.c
jtguggedal: This looks like a good solution 👍

jtguggedal (approved):

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,96 @@ int lte_lc_normal(void);
...
+ *
+ * @param[in] rptau
+ * @parblock
+ *          Requested periodic TAU in seconds as a non-negative integer. Range 0 - 35712000s.
+ *          Set to @c -1 to use modem's default value.
+ *          Given value will be rounded up to the closest possible value, which are
+ *          represented by multiplying timer unit with a timer value:
+ *
+ *          - Timer unit is one of: 2s, 30s, 60s, 600s (10min), 3600s (1h), 36000s (10h),
+ *            1152000s (320h)
```
peknis, line 1418:
```suggestion
 *          - Timer unit is one of: 2 s, 30 s, 60 s, 600 s (10 min), 3600 s (1 h), 36000 s (10 h),
 *            1152000 s (320 h)
```

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,96 @@ int lte_lc_normal(void);
...
+ *          - Timer value range is 0-31
+ * @endparblock
+ *
+ * @param[in] rat
+ * @parblock
+ *          Requested active time in seconds as a non-negative integer. Range 0 - 11160s.
+ *          Set to @c -1 to use modem's default value.
+ *          Given value will be rounded up to the closest possible value, which are
+ *          represented by multiplying timer unit with a timer value:
```
peknis, line 1427:
```suggestion
 *          The given value will be rounded up to the closest possible value that is
 *          calculated by multiplying the timer unit with the timer value:
```

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,96 @@ int lte_lc_normal(void);
...
+ *
+ *          - Timer unit is one of: 2s, 30s, 360s (6min)
```
peknis, line 1429:
```suggestion
 *          - Timer unit is one of: 2 s, 30 s, 360 s (6 min)
```

lib/lte_link_control/lte_lc_helpers.h
```diff
@@ -206,6 +206,17 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg);
 int parse_psm(const char *active_time_str, const char *tau_ext_str,
 	      const char *tau_legacy_str, struct lte_lc_psm_cfg *psm_cfg);
 
+/* @brief Encodes Periodic TAU timer and active time strings.
```
peknis, line 209:
```suggestion
/* @brief Encode Periodic TAU timer and active time strings.
```

peknis (changes requested):

lib/lte_link_control/lte_lc_helpers.c
tokangas: I'm not sure if modem would actually ever send such values, even if received from the network. Modem decodes the value and interprets the length of the timer, I don't think it should send such illegal values in the AT interface. The comment you copied is from the 3GPP specification, not from the AT command spec. But anyway, this could be reworked in another PR.

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -435,6 +435,147 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg)
 	return err;
 }
 
+/* Differenet values in the T3324 lookup table. */
```
tokangas, line 438:
```suggestion
/* Different values in the T3324 lookup table. */
```

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -435,6 +435,147 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg)
...
+#define T3324_LOOKUP_DIFFERENT_VALUES 3
+/* Differenet values in the T3412 extended lookup table. */
```
tokangas, line 440:
```suggestion
/* Different values in the T3412 extended lookup table. */
```

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -435,6 +435,147 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg)
...
+#define T3412_EXT_LOOKUP_DIFFERENT_VALUES 7
+/* Maximum value for the timer value field of the timer information elements in both
+ * T3324 and T3412 extended lookup tables.
+ */
+#define TIMER_VALUE_MAX 31
+
+/* Lookup table for T3324 timer used for PSM active time in seconds.
+ * Ref: GPRS Timer 2 IE in 3GPP TS 24.008 Table 10.5.163/3GPP TS 24.008.
```
tokangas, line 448:
```suggestion
 * Ref: GPRS Timer 2 IE in 3GPP TS 24.008 Table 10.5.163.
```

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -435,6 +435,147 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg)
...
+ */
+static const uint32_t t3324_lookup[8] = {2, 60, 360, 60, 60, 60, 60, 0};
+
+/* Lookup table for T3412-extended timer used for periodic TAU. Unit is seconds.
+ * Ref: GPRS Timer 3 in 3GPP TS 24.008 Table 10.5.163a/3GPP TS 24.008.
```
tokangas, line 453:
```suggestion
 * Ref: GPRS Timer 3 in 3GPP TS 24.008 Table 10.5.163a.
```

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -435,6 +435,147 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg)
...
+ */
+static const uint32_t t3412_ext_lookup[8] = {600, 3600, 36000, 2, 30, 60, 1152000, 0};
+
+/* Lookup table for T3412 (legacy) timer used for periodic TAU. Unit is seconds.
+ * Ref: GPRS Timer in 3GPP TS 24.008 Table 10.5.172/3GPP TS 24.008.
```
tokangas, line 458:
```suggestion
 * Ref: GPRS Timer in 3GPP TS 24.008 Table 10.5.172.
```

lib/lte_link_control/Kconfig
```diff
@@ -69,22 +69,56 @@ config LTE_PSM_REQ
 	  the modem is initialized. This will cause the modem to include a PSM request in
 	  every subsequent LTE attach request.
 
+choice LTE_PSM_REQ_FORMAT
+	prompt "Select network mode"
```
tokangas, line 73: Copy paste.

include/modem/lte_lc.h
tokangas: I had a better look at the AT command specification and the requested periodic TAU parameter is not optional after all. The spec does not say that omitting the value would use "manufacturer-specific" default, as it does say for the Active-Time parameter.

include/modem/lte_lc.h
```diff
@@ -1345,18 +1345,96 @@ int lte_lc_normal(void);
...
+ *          - 100: 30 seconds
+ *          - 101: 1 minute
+ *          - 110: 320 hours
+ *          - 111: Timer is deactivated
+ * @endparblock
+ *
+ * @param[in] rat
+ * @parblock
+ *          Requested active time as a null-terminated string.
+ *          Set to @c NULL to use a manufacturer-specific default value.
```
tokangas: "Modem's default" also here?

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -435,6 +435,147 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg)
...
+
+	if (rat >= 0) {
+		ret |= encode_psm_timer(
+			active_time_str,
+			rat,
+			t3324_lookup,
+			T3324_LOOKUP_DIFFERENT_VALUES);
+	} else {
+		*active_time_str = '\0';
+		LOG_DBG("RPTAU use default");
```
tokangas, line 573: Should be RAT, not RPTAU.
```suggestion
		LOG_DBG("Using modem default value for RAT");
```

include/modem/lte_lc.h
trantanen: As discussed offline, TAU is optional from the perspective that all parameters can be left out. It's true that AT reference guide doesn't give an option to give active time while not giving TAU. This works with the modem although AT guide doesn't mention it. If you think this requires further investigation, please create a ticket.

include/modem/lte_lc.h
```diff
@@ -1343,25 +1343,108 @@ int lte_lc_normal(void);
...
+ *          Bits 8 to 6 define the timer unit as follows:
+ *
+ *          - 000: 10 minutes
+ *          - 001: 1 hour
+ *          - 010: 10 hours
+ *          - 011: 2 seconds
+ *          - 100: 30 seconds
+ *          - 101: 1 minute
+ *          - 110: 320 hours
+ *          - 111: Timer is deactivated
```
tokangas, line 1371: As far as I understand, `111` should be removed from the description because TAUs can not be disabled. I think that `111` for T3412_EXT just means that T3412 is used instead. `111` makes sense in the %XMONITOR notification, but not here, I think.

trantanen: Line with '111' removed.

tokangas (approved):

peknis (approved):

jhirsi (approved):

PR #12991, opened 2023-11-07
lib: nrf_modem: lte_net_if: change defaults

lemrey:
Change Kconfig defaults for the `_AUTO_` options, to reflect the intended defaults for NCS samples.


simensrostad (approved):
:fireworks: :fireworks:

lemrey:
No changelog entry required, we are changing a default but the sample behavior is unaffected..

rlubos (approved):

PR #12994, opened 2023-11-07
Bluetooth: Controller: Update docs for BT_CTLR_SDC_(RX/TX)_PACKET_COUNT

rugeGerritsen:
This makes it more clear what the entries represent and how they should be set.


rugeGerritsen:
Created from https://github.com/nrfconnect/sdk-nrf/pull/12919/files

hubuhubu (approved):

subsys/bluetooth/controller/Kconfig
```diff
@@ -163,19 +163,20 @@ config BT_CTLR_SDC_TX_PACKET_COUNT
...
 	default 2
 	range 1 20
 	help
 	  The number Link Layer ACL RX packets reserved per connection.
-	  With the default count, the application is able to empty the buffers during a connection event.
-	  That is, non-default values should only be used if reduced throughput is accepted, or when the CPU
-	  utilization is so high that the application is not able to read data fast enough during connection events.
+	  With the default count, the application is expected to be able to empty the buffers
+	  during a connection event. That is, non-default values (>2) should only be used when
+	  the CPU utilization is so high or set to 1 if reduced throughput is accepted.
```
seko-nordic, line 179:
```suggestion
	  during a connection event. That is, non-default values (>2) should only be used when
	  the CPU utilization is so high that the application is not able to read data fast enough during connection events
	  Value 1 should be used to save memory when reduced throughput is accepted.
```

subsys/bluetooth/controller/Kconfig
```diff
@@ -163,19 +163,23 @@ config BT_CTLR_SDC_TX_PACKET_COUNT
 	range 1 20
 	help
 	  The number Link Layer ACL TX packets reserved per connection.
-	  With the default count, the application is able to refill the buffers during a connection event.
-	  That is, non-default values should only be used if reduced throughput is accepted, or when the CPU
-	  utilization is so high that the application is not able to provide data fast enough during connection events.
+	  With the default count, the application is expected to be able to refill
+	  the buffers during a connection event. That is, non-default (>3) should only
```
seko-nordic, line 167:
```suggestion
	  the buffers during a connection event. That is, non-default values (>3) should only
```
we should either put it back here or remove from line 179

seko-nordic (approved):

PR #12999, opened 2023-11-07
manifest: sdk-nrfxlib: Pull Wi-Fi test spec update

krish2718:
Fix issue of Wi-Fi tests not being run for OS agnostic driver changes.


sachinthegreen (approved):

krish2718:
manifest update done as part of #12993

PR #13001, opened 2023-11-08
samples: cellular: mss Fix NRF_MODEM_LIB configuration

glarsennordic:
A recent commit changed the default values for
CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_START
CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_CONNECT
CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_DOWN

but did not update this sample. This commit rectifies that.


plskeggs (approved):

jayteemo (commented):
https://github.com/nrfconnect/sdk-nrf/pull/12991 should have added a change log entry since it changed a public library's default behavior.

feel free to add one if you'd like.

jayteemo (approved):

jayteemo (approved):

peknis (approved):

lemrey (changes requested):
We did agree that samples in NCS shall manage network interfaces and connectivity all in the same way, by manually calling the API functions. I would be fine with keeping `_AUTO_DOWN=y`  though, that is different from the other two.

lemrey (approved):
Thank you for your patience and for updating the changelog.

PR #13007, opened 2023-11-08
bootloader: keygen: Fix regression in 0xFFFF check

sigvartmh:
A change calculates the digest out of the public key encoded in DER, and makes sure that it does not contain ‘ffff’, but provision.py still calculates the digest out of the raw public key (raw EC point), hence it may contain ‘ffff' and raise an error

This changes `keygen.py` to check the uncompressed points(raw EC points) digest if it contains `ffff`.

### DIFF
--- a/scripts/bootloader/keygen.py
+++ b/scripts/bootloader/keygen.py
@@ -23,10 +23,13 @@ def generate_legal_key():
     while True:
         key = ec.generate_private_key(ec.SECP256R1())
         public_bytes = key.public_key().public_bytes(
-                encoding=serialization.Encoding.DER,
-                format=serialization.PublicFormat.SubjectPublicKeyInfo,
-                )
-        digest = sha256(public_bytes).digest()[:16]
+                encoding=serialization.Encoding.X962,
+                format=serialization.PublicFormat.UncompressedPoint,
+        )
+
+        # The digest don't contain the first byte as it denotes
+        # if it is compressed/UncompressedPoint.
+        digest = sha256(public_bytes[1:]).digest()[:16]
         if not (any([digest[n:n + 2] == b'\xff\xff'
                      for n in range(0, len(digest), 2)])):
             return key


scripts/bootloader/keygen.py
```diff
@@ -23,10 +23,11 @@ def generate_legal_key():
...
         public_bytes = key.public_key().public_bytes(
-                encoding=serialization.Encoding.DER,
-                format=serialization.PublicFormat.SubjectPublicKeyInfo,
-                )
-        digest = sha256(public_bytes).digest()[:16]
+                encoding=serialization.Encoding.X962,
+                format=serialization.PublicFormat.UncompressedPoint,
+        )
+
+        digest = sha256(public_bytes[1:]).digest()[:16]
```
sigvartmh, line 30: Just incase anyone wonder why we take off 1 byte here it's the encoding output which gives 0x4 for uncompressed points. It wasn't possible to just get the Raw EC points out of the library in the same way as the python implementation.

Damian-Nordic: How about adding a comment on that?

Damian-Nordic (approved):
Thanks, did you maybe verified that this outputs the same digest as provision.py?

osaether (approved):

sigvartmh:
> Thanks! Did you maybe verify that this outputs the same digest as provision.py?

Will add comment, and yeah i've verified that it outputs the same
```python
In [89]: key.public_key().public_bytes(encoding=serialization.Encoding.X962, format=serialization.PublicFormat.UncompressedPoint)
Out[89]: b'\x04\xda\xf1\xa9\x07\xf9!:\x136hN\x8c\x96\x8d\x00\x90\xf4\xbc\xf3uw\x1e H\xe4\x9a_\x1bl\x1dr\xe1\xbc"F\xf4\xac7\xfa\x95Rvhi\xe2\xd38\x036u\xf2\x86\xed\x89\xdd#\xde\x9b\x9e\xaf>\x93\xe1\x14'

In [90]: ec_lib_verifykey.to_string()
Out[90]: b'\xda\xf1\xa9\x07\xf9!:\x136hN\x8c\x96\x8d\x00\x90\xf4\xbc\xf3uw\x1e H\xe4\x9a_\x1bl\x1dr\xe1\xbc"F\xf4\xac7\xfa\x95Rvhi\xe2\xd38\x036u\xf2\x86\xed\x89\xdd#\xde\x9b\x9e\xaf>\x93\xe1\x14'

scripts/bootloader/keygen.py
sigvartmh: added

PR #13015, opened 2023-11-08
kconfig: menu cleanup

lemrey:
Cleaning up the Kconfig menu; this has expanded with no structure and gotten a bit messy.

- I have added a `Subsystem` menu entry and put all the subsytems there so that the menu structure reflects the tree's
- I changed some `config` into `menuconfig` so they wouldn't expand in the top-level menu with dozens of entries
- I created groups for a few lib entries that had no group
- I moved some TFM and nRF Security configs, mainly the peripheral configuration (@joerchan) and the mbedTLS version choice 

~I am not finished, might add a few more things; feel free to point out other things that you feel we could fix here.~


kkasperczyk-no (approved):
Approving Matter related part

krish2718 (approved):
For Wi-Fi.

modules/tfm/zephyr/Kconfig
```diff
@@ -4,6 +4,8 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
+menu "TFM"
```
joerchan, line 7:
```suggestion
menu "TF-M"
```
Sorry

rugeGerritsen (approved):
MPSL change is good

modules/tfm/zephyr/Kconfig
lemrey: Thanks, no problem; have a question though.
There are now two places with Kconfigs related to TF-M and I wonder if we could move these to one place. 

The top level `Kconfig.nrf` pulls `rsource "modules/Kconfig"`, and that in turns pulls `rsource "tfm/zephyr/Kconfig"`, with this menu.

But TF-M also has another menu containing all the Kconfig in `nrf/modules/trusted-firmware-m` and that menu is accessible in the `Modules` menu (beside `nrf` and all the other modules).

jheiskan81 (approved):

subsys/dm/Kconfig
```diff
@@ -4,17 +4,15 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
-config DM_MODULE
-	bool "DM (Distance Measurement) module [EXPERIMENTAL]"
+menuconfig DM_MODULE
+	bool "nRF DM (Distance Measurement) [EXPERIMENTAL]"
```
maje-emb, line 8: Maybe something like that:
```suggestion
	bool "nRF DM (Distance Measurement) module [EXPERIMENTAL]"
```
Similar to this: https://github.com/nrfconnect/sdk-nrfxlib/blob/main/nrf_dm/Kconfig#L8
Also to distinguish a library from module.

lemrey: What is the distinction between a library and a module?

maje-emb: The library is in the nrfxlib repository. Module in subsys integrates this library with NCS.

modules/tfm/zephyr/Kconfig
joerchan: There may be some issues with that, because that may change the evaluation order and cause issues.
Initially in was only TFM_BOARD config in this file so I think the rest can possible be moved. But moving TFM_BOARD could potentially be a problem. It's worth a shot if you want to try it though.

SeppoTakalo (approved):

juhaylinen (approved):

subsys/dm/Kconfig
lemrey: I can do that change, but the usage of "module" sounds a bit storage here.

In my opinion the difference between library and module as you intend it is not clear to the user.
I believe this term is used for other things elsewhere -- what do we call components under `/nrf/modules`?
Are those modules in the same way that this component is a module?

But then from a user perspective the difference is minimal anyhow, this "module" is what pulls in the library and enable its API usage in the application. From a user perspective I argue that enabling this is equivalent to adding the library to their project, so I argue that the prompt in the menu doesn't need to differentiate between "library" or "module" it could just say `nRF DM (Distance measurement)` -- you select that and you can use the API provided by the distance measurement library.

Food for thoughts, I can keep "module" in the prompt if you want, but I think this is a good chance to think about it.

maje-emb: I see, you're right, let's remove "module" in the prompt.

maje-emb (approved):
For DM

joerchan (approved):

joerchan (changes requested):
Just waiting for "TF-M" to be updated :)

KAGA164 (approved):

lemrey:
@sigvartmh please have a look at my changes to `FW_INFO`.
@joerchan I have updated the menu title

joerchan (approved):

carlescufi:
@lemrey please remove the DNM once @sigvartmh has reviewed.

sigvartmh (approved):
FW_INFO changes looks good.

lemrey:
No changelog entry necessary, this is not visible to the user.

PR #13017, opened 2023-11-08
wifi: Fix nRF70 patch relocation

krish2718:


drivers/wifi/nrf700x/src/fw_load.c
```diff
@@ -89,7 +95,7 @@ enum nrf_wifi_status nrf_wifi_fw_load(void *rpu_ctx)
 		LOG_ERR("%s: nrf_wifi_fmac_fw_load failed", __func__);
 	}
 
-#if defined(CONFIG_NRF_WIFI_PATCHES_EXT_FLASH) && defined(CONFIG_NORDIC_QSPI_NOR)
+#if defined(CONFIG_NRF_WIFI_PATCHES_EXT_FLASH_XIP) && defined(CONFIG_NORDIC_QSPI_NOR)
```
Damian-Nordic, line 98: How about making `NRF_WIFI_PATCHES_EXT_FLASH_XIP` dependent on `NORDIC_QSPI_NOR` in Kconfig if both are needed?

drivers/wifi/nrf700x/src/fw_load.c
```diff
@@ -63,23 +64,28 @@ LOG_MODULE_DECLARE(wifi_nrf, CONFIG_WIFI_NRF700X_LOG_LEVEL);
...
 
+#ifdef CONFIG_NRF_WIFI_PATCHES_BUILTIN
 	status = nrf_wifi_fmac_fw_parse(rpu_ctx, (const uint8_t *)_bin_nrf70_fw_start,
 					_bin_nrf70_fw_end - _bin_nrf70_fw_start, &fw_info);
 	if (status != NRF_WIFI_STATUS_SUCCESS) {
 		LOG_ERR("%s: nrf_wifi_fmac_fw_parse failed", __func__);
 		return status;
 	}
+#else
+	BUILD_ASSERT(0, "CONFIG_NRF_WIFI_PATCHES_BUILTIN must be enabled");
```
Damian-Nordic, line 85: So this option is not really useful until non-XIP patch support is added. Is that going to happen in NCS 2.6.0?

drivers/wifi/nrf700x/Kconfig
```diff
@@ -72,11 +72,20 @@ config NRF_WIFI_IF_AUTO_START
 	bool "Enable Wi-Fi interface auto start on boot"
 	default y
 
-config NRF_WIFI_PATCHES_EXT_FLASH
-	bool "Store nRF700x FW patches in external flash"
+config NRF_WIFI_PATCHES_BUILTIN
+	bool "Store nRF700x FW patches as part of the driver"
+	help
+	  Select this option to store nRF700x FW patches as part of the driver.
+	  This option is impacts the flash memory size of the application.
```
Damian-Nordic, line 79:
```suggestion
	  This option impacts the flash memory size of the application.
```

Damian-Nordic (approved):
Some nits found but looking good overall.

drivers/wifi/nrf700x/src/fw_load.c
krish2718: No, it works even now, but for this option we also need `CONFIG_NRF_WIFI_PATCHES_BUILTIN` as that builds the patch in to the image and then XIP relocates.

krish2718: > Is that going to happen in NCS 2.6.0?

Yes, this patch is in preparation for that. But it's limited to non-XIP over QSPI i.e., 5340/52840.

sachinthegreen (approved):

PR #13022, opened 2023-11-09
MPSL/SoftDevice Controller: Lower footprint of assertion print

rugeGerritsen:
Previously it was a requirement that logging was enabled to print out assertions. As this configuration has significant memory footprint, allow printing the assertion message using verbose assertions or printk if it is enabled.

### DIFF
--- a/subsys/bluetooth/controller/hci_driver.c
+++ b/subsys/bluetooth/controller/hci_driver.c
@@ -250,8 +250,18 @@ void sdc_assertion_handler(const char *const file, const uint32_t line)
 #else /* !IS_ENABLED(CONFIG_BT_CTLR_ASSERT_HANDLER) */
 void sdc_assertion_handler(const char *const file, const uint32_t line)
 {
+#if defined(CONFIG_ASSERT) && defined(CONFIG_ASSERT_VERBOSE) && !defined(CONFIG_ASSERT_NO_MSG_INFO)
+	__ASSERT(false, "SoftDevice Controller ASSERT: %s, %d\n", file, line);
+#elif defined(CONFIG_LOG)
 	LOG_ERR("SoftDevice Controller ASSERT: %s, %d", file, line);
 	k_oops();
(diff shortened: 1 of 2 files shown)


subsys/bluetooth/controller/hci_driver.c
```diff
@@ -250,8 +250,18 @@ void sdc_assertion_handler(const char *const file, const uint32_t line)
 #else /* !IS_ENABLED(CONFIG_BT_CTLR_ASSERT_HANDLER) */
 void sdc_assertion_handler(const char *const file, const uint32_t line)
 {
+#if defined(CONFIG_ASSERT) && defined(CONFIG_ASSERT_VERBOSE) && !defined(CONFIG_ASSERT_NO_MSG_INFO)
+	__ASSERT(false, "SoftDevice Controller ASSERT: %s, %d\n", file, line);
+#elif defined(CONFIG_LOG)
 	LOG_ERR("SoftDevice Controller ASSERT: %s, %d", file, line);
 	k_oops();
+#elif defined(CONFIG_PRINTK)
+	printk("SoftDevice Controller ASSERT: %s, %d\n", file, line);
```
Frodevan, line 259: Would it be an idea to add "\<err\> " to the start of the lines for printk, or will that come through the string?

rugeGerritsen: The `k_oops()` below will print out a fatal error message

lemrey: I think it would be nicer to let the application run, maybe it can recover from this situation by re-initializing `sdc`?

rugeGerritsen: It may be hard to recover from such unexpected situations. We trigger these when the HW or stack is in an unexpected state. Rebooting the application is currently the most fool-proof solution of recovery

Tschet1 (approved):

minkrg (approved):

PR #13024, opened 2023-11-09
applications: asset_tracker_v2: Update LwM2M P-GPS configuration

tokangas:
Previously, A-GNSS and P-GPS were not supported at the same time with LwM2M. This is no longer true, so the configuration has been updated accordingly.

### DIFF
--- a/applications/asset_tracker_v2/src/modules/Kconfig.cloud_module
+++ b/applications/asset_tracker_v2/src/modules/Kconfig.cloud_module
@@ -15,12 +15,8 @@ if CLOUD_MODULE
 CLOUD_SERVICE_SELECTOR := choice
 
 # nRF Cloud A-GNSS is enabled by default when building for all supported cloud transport services.
-# For LwM2M builds, A-GNSS is disabled if P-GPS is enabled.
-# This is because the two methods are currently mutually exclusive for LwM2M builds.
-# For the other cloud configurations, A-GNSS and P-GPS can coexist.
 config NRF_CLOUD_AGNSS
 	depends on LOCATION_MODULE
-	default n if NRF_CLOUD_PGPS && LWM2M_INTEGRATION
 	default y
 
 # Enable MQTT clean session by default. This is to ensure that the configured cloud MQTT service


simensrostad (approved):

applications/asset_tracker_v2/src/modules/Kconfig.cloud_module
```diff
@@ -15,12 +15,8 @@ if CLOUD_MODULE
 CLOUD_SERVICE_SELECTOR := choice
 
 # nRF Cloud A-GNSS is enabled by default when building for all supported cloud transport services.
-# For LwM2M builds, A-GNSS is disabled if P-GPS is enabled.
-# This is because the two methods are currently mutually exclusive for LwM2M builds.
-# For the other cloud configurations, A-GNSS and P-GPS can coexist.
 config NRF_CLOUD_AGNSS
 	depends on LOCATION_MODULE
-	default n if NRF_CLOUD_PGPS && LWM2M_INTEGRATION
```
trantanen, line 23: So right now LWM2M+AGNSS+PGPS works because ANGSS and PGPS data requests are not triggered at the same time by Location library. If this changes, does LWM2M still continue to work?
I don't know if we will always guarantee in Location library that AGNSS and PGPS are not requested when another one is still fetching data.

trantanen (changes requested):

applications/asset_tracker_v2/src/modules/Kconfig.cloud_module
tokangas: This is a perfectly valid comment. After having a look at the implementation, the P-GPS request is currently delayed because the Location lib assistance implementation was simplified in #12815. The functionality change was unintentional, both A-GNSS and P-GPS data requests should be sent practically at the same time.

It's a limitation of the LwM2M location assistance library that A-GNSS and P-GPS requests can't be handled simultaneously. #13031 makes it easier to get the status of the request from the library and my intention is to use that functionality to handle A-GNSS and P-GPS requests sequentially in the ATv2 LwM2M integration.

tokangas:
I'll open a new PR because I had to broaden the scope a bit.

PR #13033, opened 2023-11-09
applications: nrf5340_audio: Add timeout when creating PA sync

alexsven:
- Delete PA sync if BIG sync is lost
- Restart scanning if PA sync create times out
- Increase stack size to account for additional pa_sync_delete

### DIFF
--- a/applications/nrf5340_audio/src/audio/Kconfig
+++ b/applications/nrf5340_audio/src/audio/Kconfig
@@ -356,7 +356,7 @@ config BUTTON_MSG_SUB_STACK_SIZE
 
 config LE_AUDIO_MSG_SUB_STACK_SIZE
 	int "Stack size for LE Audio subscriber"
-	default 1250
+	default 2048
 
 config CONTENT_CONTROL_MSG_SUB_STACK_SIZE
 	int "Stack size for content control subscriber"
--- a/applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
+++ b/applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
@@ -244,24 +244,16 @@ static void le_audio_msg_sub_thread(void)
 		case LE_AUDIO_EVT_SYNC_LOST:
 			LOG_INF("Sync lost");
 
+			ret = bt_mgmt_pa_sync_delete(msg.pa_sync);
+			if (ret) {
+				LOG_WRN("Failed to delete PA sync");
+			}
+
 			if (strm_state == STATE_STREAMING) {
 				stream_state_set(STATE_PAUSED);
 				audio_system_stop();
 				ret = led_on(LED_APP_1_BLUE);
 				ERR_CHK(ret);
-			}
-
-			if (IS_ENABLED(CONFIG_BT_OBSERVER)) {
-				ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
-				if (ret) {
-					if (ret != -EALREADY) {
-						LOG_ERR("Failed to restart scanning: %d", ret);
-					}
-					break;
-				}
-
-				/* NOTE: The string below is used by the Nordic CI system */
-				LOG_INF("Restarted scanning for broadcaster");
 			}
 
 			break;
--- a/applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
@@ -42,6 +42,33 @@ struct broadcast_source {
 	uint32_t broadcast_id;
 };
 
+static void scan_restart_worker(struct k_work *work)
+{
+	int ret;
+
+	/* Delete pending PA sync before restarting scan */
+	ret = bt_mgmt_pa_sync_delete(pa_sync);
+	if (ret) {
+		LOG_WRN("Failed to delete pending PA sync: %d", ret);
+	}
+
+	ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
+	if (ret) {
+		LOG_WRN("Failed to restart scanning for broadcast: %d", ret);
+	}
+}
+
+K_WORK_DEFINE(scan_restart_work, scan_restart_worker);
+
+static void pa_sync_timeout(struct k_timer *timer)
+{
+	LOG_WRN("PA sync create timed out, restarting scanning");
+
+	k_work_submit(&scan_restart_work);
+}
+
+K_TIMER_DEFINE(pa_sync_timer, pa_sync_timeout, NULL);
+
 static uint16_t interval_to_sync_timeout(uint16_t interval)
 {
 	uint16_t timeout;
@@ -78,6 +105,9 @@ static void periodic_adv_sync(const struct bt_le_scan_recv_info *info, uint32_t 
 	param.timeout = interval_to_sync_timeout(info->interval);
 
 	broadcaster_broadcast_id = broadcast_id;
+
+	/* Set timeout to same value as PA sync timeout in ms */
+	k_timer_start(&pa_sync_timer, K_MSEC(param.timeout * 10), K_NO_WAIT);
 
 	ret = bt_le_per_adv_sync_create(&param, &pa_sync);
 	if (ret) {
@@ -171,6 +201,8 @@ static void pa_synced_cb(struct bt_le_per_adv_sync *sync,
 	}
 
 	LOG_DBG("PA synced");
+
+	k_timer_stop(&pa_sync_timer);
 
 	msg.event = BT_MGMT_PA_SYNCED;
 	msg.pa_sync = sync;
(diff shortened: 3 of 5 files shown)


alexsven:
To avoid losing just the BIG sync and not the PA_sync and thereby be stuck in an in-between state we now delete the PA_sync when BIG sync is lost to start "fresh" everytime. Tested by going out of range and staying in the transition between out-of-range and in-range and we always resume the BIS when we are in stream.

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
```diff
@@ -42,6 +42,33 @@ struct broadcast_source {
...
+		LOG_WRN("Failed to delete pending PA sync: %d", ret);
+	}
+
+	ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
+	if (ret) {
+		LOG_WRN("Failed to restart scanning for braodcast: %d", ret);
+	}
+}
+
+K_WORK_DEFINE(scan_restart_work, scan_restart_worker);
```
rick1082, line 61: Probably we can move all the K_WORK_DEFINE to the top of the file.

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
```diff
@@ -42,6 +42,33 @@ struct broadcast_source {
...
+
+static void pa_sync_timeout(struct k_timer *timer)
+{
+	LOG_WRN("PA sync create timed out, restarting scanning");
+
+	k_work_submit(&scan_restart_work);
+}
+
+K_TIMER_DEFINE(pa_sync_timer, pa_sync_timeout, NULL);
```
rick1082, line 70: Same like previous comment.

rick1082 (approved):

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
alexsven: They have to be under the function that they call, to either we have to add a prototype at the top of the file or we keep the K_DEFINEs were they are. I had a look in the rest of our code, and K_WORKs and K_TIMERs are all in the midst of the other code (except one K_TIMER_DEFINE in bt_mgmt_ctlr_cfg.c were we have a prototype)

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
```diff
@@ -79,6 +106,9 @@ static void periodic_adv_sync(const struct bt_le_scan_recv_info *info, uint32_t
 
 	broadcaster_broadcast_id = broadcast_id;
 
+	/* Set timeout to same value as PA sync timeout in ms */
```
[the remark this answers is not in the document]
alexsven: I don't think that can happen no, but if it does it is not a problem. This is the documentation for k_timer_start:

>  Attempting to start a timer that is already running is permitted.
>  The timer's status is reset to zero and the timer begins counting down
>  using the new duration and period values.

applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
```diff
@@ -246,26 +246,18 @@ static void le_audio_msg_sub_thread(void)
...
-			if (IS_ENABLED(CONFIG_BT_OBSERVER)) {
-				ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
-				if (ret) {
-					if (ret != -EALREADY) {
-						LOG_ERR("Failed to restart scanning: %d", ret);
-					}
-					break;
-				}
-
-				/* NOTE: The string below is used by the Nordic CI system */
```
erikrobstad, line 265: Is the removal of this string handled in CI?

gWacey (approved):

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
```diff
@@ -495,6 +503,8 @@ int broadcast_sink_pa_sync_set(struct bt_le_per_adv_sync *pa_sync, uint32_t broa
 		return ret;
 	}
 
+	pa_sync_stored = pa_sync;
```
erikrobstad, line 506: Hmmm it seems like we can use broadcast_sink->pa_sync.
Then we dont need to save it here.

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
```diff
@@ -42,6 +42,33 @@ struct broadcast_source {
...
+	}
+}
+
+K_WORK_DEFINE(scan_restart_work, scan_restart_worker);
+
+static void pa_sync_timeout(struct k_timer *timer)
+{
+	LOG_WRN("PA sync create timed out, restarting scanning");
+
+	k_work_submit(&scan_restart_work);
```
erikrobstad, line 67: I think using delayable work instead of timer + work is cleaner.
k_work_schedule() and k_work_cancel_delayable()

erikrobstad: See e.g. unicast_client.c for reference

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
```diff
@@ -42,6 +42,33 @@ struct broadcast_source {
...
+
+	/* Delete pending PA sync before restarting scan */
+	ret = bt_mgmt_pa_sync_delete(pa_sync);
+	if (ret) {
+		LOG_WRN("Failed to delete pending PA sync: %d", ret);
+	}
+
+	ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
+	if (ret) {
+		LOG_WRN("Failed to restart scanning for braodcast: %d", ret);
```
erikrobstad, line 57:
```suggestion
		LOG_WRN("Failed to restart scanning for broadcast: %d", ret);
```

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
```diff
@@ -42,6 +42,33 @@ struct broadcast_source {
...
 };
 
+static void scan_restart_worker(struct k_work *work)
+{
+	int ret;
+
+	/* Delete pending PA sync before restarting scan */
+	ret = bt_mgmt_pa_sync_delete(pa_sync);
+	if (ret) {
+		LOG_WRN("Failed to delete pending PA sync: %d", ret);
```
erikrobstad, line 52: Should do LOG_ERR here and return?

applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
```diff
@@ -246,26 +246,18 @@ static void le_audio_msg_sub_thread(void)
 		case LE_AUDIO_EVT_SYNC_LOST:
 			LOG_INF("Sync lost");
 
+			ret = bt_mgmt_pa_sync_delete(msg.pa_sync);
+			if (ret) {
+				LOG_WRN("Failed to delete pa sync");
```
erikrobstad, line 251:
```suggestion
				LOG_WRN("Failed to delete PA sync");
```

applications/nrf5340_audio/src/audio/Kconfig
```diff
@@ -355,7 +355,7 @@ config BUTTON_MSG_SUB_STACK_SIZE
 
 config LE_AUDIO_MSG_SUB_STACK_SIZE
 	int "Stack size for LE Audio subscriber"
-	default 1250
+	default 2048
```
erikrobstad, line 358: This seems like a fairly large increase.
Is it necessary to increase it that much?

erikrobstad (changes requested):

applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
alexsven: Yes, since it is also printed on the case below and that case will always happen after this

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
alexsven: No, it is set to NULL when we lose the BIG sync

applications/nrf5340_audio/src/audio/Kconfig
alexsven: The stacj usage was 1350, so about 65% usage. That leaves us some room for additional variables etc in the future

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
alexsven: I'd rather end up in a scanning state than in between

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
alexsven: Not sure I agree. It requires a separate init function to be run and the timer has the benefit of being able to be started multiple times and it will just reset the timer. If you want similar behaviour for k_work_delayable I believe you have to first cancel k_work if it is pending and then reschedule

alexsven: Using a timer is also closer to a traditional watchdog, so I think it is easier to read

PR #13035, opened 2023-11-09
ci: Add a new Wi-Fi QA required label

krish2718:
This should be automatically added for nrf Wi-Fi related changes. This is used to enforce a mandatory QA upvote even though it is ready to be merged.

We are seeing an issue where the PRs are getting merged as they pass all mandatory checks but QA hasn't had a chance to verify it. So, this label which is enabled by default can be used similar to `doc-required`, and if its trivial and developer is confident enough then the label can be manually removed.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -23,3 +23,9 @@
   - "samples/bluetooth/mesh/*"
   - "doc/nrf/libraries/bluetooth_services/mesh/*"
   - "doc/nrf/ug_bt_mesh*"
+
+"wifi-qa-required":
+  - "drivers/wifi/**/*"
+  - "samples/wifi/**/*"
+  - "modules/hostap/**/*"
+  - "modules/lib/hostap/**/*"


krish2718:
Test PR: https://github.com/krish2718/sdk-nrf/pull/3 Still debugging

krish2718:
> Test PR: [krish2718#3](https://github.com/krish2718/sdk-nrf/pull/3) Still debugging

Working fine now, was a bug in the regex (auto-complete).

carlescufi:
@thst-nordic @jangalda-nsc please take a look

jangalda-nsc:
Does it only add label? I guess it does not create strict requirement which blocks merging. Is it fine for you?

krish2718:
> Does it only add label? I guess it does not create strict requirement which blocks merging. Is it fine for you?

Just following existing methodology like `doc-required`. I don't think we can block merging based on a label, at the most we can change the repo settings to need an approval from code-owner but that doesn't help here.

sachinthegreen (approved):

jangalda-nsc (approved):

krish2718:
@nrfconnect/ncs-merge-access FYI for enforcing this label, any member of @nrfconnect/ncs-test-leads can approve, for Wi-Fi it will be @krga2022 .

PR #13036, opened 2023-11-09
samples: Add nRF9161 Wi-Fi tests to twister

krish2718:
Apart from checking the build these are also needed for generating memory footprints.

### DIFF
--- a/samples/wifi/scan/sample.yaml
+++ b/samples/wifi/scan/sample.yaml
@@ -16,14 +16,20 @@ tests:
       - nrf5340dk_nrf5340_cpuapp
       - nrf52840dk_nrf52840
       - nrf9160dk_nrf9160_ns
-    platform_allow: nrf5340dk_nrf5340_cpuapp nrf52840dk_nrf52840 nrf9160dk_nrf9160_ns
+      - nrf9161dk_nrf9161_ns
+    platform_allow:
+      - nrf5340dk_nrf5340_cpuapp
+      - nrf52840dk_nrf52840
+      - nrf9160dk_nrf9160_ns
+      - nrf9161dk_nrf9161_ns
     tags: ci_build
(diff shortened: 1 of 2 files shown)


divipillai:
The docs for these samples need updates. I will be working on the docs.

sachinthegreen (approved):

samples/cellular/location/sample.yaml
```diff
@@ -34,7 +34,8 @@ tests:
     build_only: true
     integration_platforms:
       - nrf9160dk_nrf9160_ns
-    platform_allow: nrf9160dk_nrf9160_ns
+      - nrf9161dk_nrf9161_ns
```
trantanen, line 37: We'd need to add nrf9161dk_nrf9161_ns to all configurations of this file. Should I do that with a separate PR?

krish2718: Yes please that would be better, this PR is only for Wi-Fi parts.

divipillai: I plan to cover the document updates for the location sample as part of this PR. Should that be covered in the new PR?

krish2718: Preferably, as I would then need to add you as a colloborator.

trantanen: PR https://github.com/nrfconnect/sdk-nrf/pull/13048 for Location sample

trantanen: You can remove the changes in this file from this PR.

trantanen (approved):

divipillai:
I have covered the doc updates of shell sample here: https://github.com/nrfconnect/sdk-nrf/pull/13052/files#diff-7fd937b453cbda230e07e8f3fd883f26ba21b2b9af7631819d70c4f8a997dd47 . So, we are good from doc side.

PR #13043, opened 2023-11-10
samples: bluetooth: delete bond when disconnecting.

martintv:
Some tests test that you can reconnect after one device resets. To do that now that sample is encrypted,
 you need to delete the bond on the device that doesnt reset.


KAGA164 (approved):

martintv:
I was wrong about this, sample already delets bond. test failure was for another reason.

PR #13046, opened 2023-11-10
applications: nrf5340_audio: Standardize controller version printout

Frodevan:
It is expected by the CI system that the controller version-string print is of the same structure regardless of which controller is used. This PR fixes a discrepancy with this printout.


Frodevan:
This PR depends on a pending change in the integration tests. When this is ready to merge, the integration-test PR will have to be merged first.

rugeGerritsen (approved):

erikrobstad (approved):

rick1082 (approved):

PR #13050, opened 2023-11-10
twister: quarantine: Add quarantine for tests from sdk-zephyr

PerMac:
Adds a quarantine required for a new workflow, where tests from sdk-zephyr are executed within nrf context (using manifest from nrf)


katgiadla (approved):

gopiotr (approved):

jangalda-nsc (changes requested):
There are still 4 tests failing https://jenkins-ncs.nordicsemi.no/blue/organizations/jenkins/buran%2Fusers%2Fjag1%2Fsdk-nrf/detail/PR-13050/4/tests

PerMac:
These all failures are on nrf52833dk. The idea is that we will bring back nrf52840dk to the scope and remove nrf52833dk, hence no need for those.

jangalda-nsc (approved):

PR #13051, opened 2023-11-10
applications: nrf5340_audio: Implement zbus MSG_SUBSCRIBER

erikrobstad:
MSG_SUBSCRIBER guarantees that
messages are not lost (default
FIFO size is 16 messages).

### DIFF
--- a/applications/nrf5340_audio/Kconfig
+++ b/applications/nrf5340_audio/Kconfig
@@ -32,10 +32,6 @@ config TRANSPORT_CIS
 
 endchoice
 
-config ZBUS_RUNTIME_OBSERVERS_POOL_SIZE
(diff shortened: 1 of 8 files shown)


alexsven (approved):

erikrobstad:
> Topics for discussion:
> 
> * Should we now change bt_mgmt from listener to msg_subscriber? Is there a need for running this callback in the BT context?
> * Should content control and/or volume be changed from subscriber to msg_subscriber? Do we see any scenarios where these events are happening so fast that they might get lost?
> * I guess button events can still me subscriber, since it is not critical if some of those are lost?

I have created OCT-2791 where we can consider and potentially implement these changes.

rick1082 (approved):

PR #13056, opened 2023-11-10
scripts: Fix cryptography python package minimum version requirement

sigvartmh:
Fix-up for 1994a1bbe5872429b7592b9e61aa46af3231c2ad as the back-end argument was dropped in 36 release.

See: pyca/cryptography/commit/f7e2ec8fe11485f7725b840f811db0d2422e37f5


joerchan:
I seem to remember calls actually using this argument though. Wouldn't this be a breaking change then?

sigvartmh:
My wording was maybe a bit inprecise it was changed to an `Any` type meaning it can be dropped when calling the method providing the argument is still valid just optional.

joerchan (approved):

PR #13059, opened 2023-11-13
applications: serial_lte_modem: fix connection to FOTA servers

tomi-font:
In certain cases connect() would fail with ETIMEDOUT.


wiba-nordic (approved):
Table looks good.

Should this have a Known Issue entry?

SeppoTakalo (approved):

MarkusLassila (approved):
This if completely fine, but we should maybe delay adding this until we see the network logs from the failure.

tomi-font:
Seems like this may have been somehow fixed. Hasn't happened for some time in CI. We'll keep an eye on it.

PR #13061, opened 2023-11-13
crypto: Use Matter version of SPAKE2+

vili-nordic:
Set's the define in Oberon could so Matter compatible of SPAKE2+ is used.


SebastianBoe (approved):

joerchan (changes requested):
I suggest we set the define in the config header instead. This avoids changing the external source code.

LuDuda (approved):

joerchan (approved):

PR #13066, opened 2023-11-13
samples: cellular: udp: cleanup and add way to shutdown modem

MirkoCovizzi:
This commit adds a new Kconfig option for configuring
the number of data transmissions to the server.
Shutdown the modem library when there are no more
transmissions to execute.
Moved sample to `CI-iot-samples-test` plan.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -316,6 +316,10 @@ Cellular samples
 
   * Added the configuration overlay file :file:`overlay-supl.conf` for building the sample with SUPL assistance support.
 
+* :ref:`udp` sample:
+
+  * Added the :ref:`CONFIG_UDP_DATA_UPLOAD_ITERATIONS <CONFIG_UDP_DATA_UPLOAD_ITERATIONS>` Kconfig option for configuring the number of data transmissions to the server.
+
 Cryptography samples
 --------------------
 
--- a/samples/cellular/udp/Kconfig
+++ b/samples/cellular/udp/Kconfig
@@ -14,6 +14,14 @@ config UDP_DATA_UPLOAD_FREQUENCY_SECONDS
 	int "How often data is transmitted to the server"
 	default 900
 
+config UDP_DATA_UPLOAD_ITERATIONS
+	int "Number of data transmissions to the server before shutdown"
+	range -1 2147483647
+	default -1
+	help
+	  Number of data transmissions to the server before shutdown.
+	  Set to -1 to transmit indefinitely.
+
 config UDP_SERVER_ADDRESS_STATIC
 	string "UDP server IP address"
 	default "8.8.8.8"
--- a/samples/cellular/udp/README.rst
+++ b/samples/cellular/udp/README.rst
@@ -62,6 +62,12 @@ CONFIG_UDP_DATA_UPLOAD_SIZE_BYTES - UDP data upload size configuration
 CONFIG_UDP_DATA_UPLOAD_FREQUENCY_SECONDS - UDP data upload frequency configuration
    This configuration option sets the frequency with which the sample transmits data to the server.
 
+.. _CONFIG_UDP_DATA_UPLOAD_ITERATIONS:
+
+CONFIG_UDP_DATA_UPLOAD_ITERATIONS - UDP data upload iterations configuration
+   This configuration option sets the number of times the sample transmits data to the server before shutting down.
+   Set to ``-1`` to transmit indefinitely.
+
 .. _CONFIG_UDP_SERVER_ADDRESS_STATIC:
 
 CONFIG_UDP_SERVER_ADDRESS_STATIC - UDP Server IP Address configuration
--- a/samples/cellular/udp/src/main.c
+++ b/samples/cellular/udp/src/main.c
@@ -18,7 +18,7 @@ static struct sockaddr_storage host_addr;
 static struct k_work_delayable socket_transmission_work;
 static int data_upload_iterations = CONFIG_UDP_DATA_UPLOAD_ITERATIONS;
 
-K_SEM_DEFINE(lte_connected, 0, 1);
+K_SEM_DEFINE(lte_connected_sem, 0, 1);
 K_SEM_DEFINE(modem_shutdown_sem, 0, 1);
 
 static void socket_transmission_work_fn(struct k_work *work)
@@ -98,7 +98,7 @@ static void lte_handler(const struct lte_lc_evt *const evt)
 		printk("Network registration status: %s\n",
 		       evt->nw_reg_status == LTE_LC_NW_REG_REGISTERED_HOME ?
 		       "Connected - home" : "Connected - roaming");
-		k_sem_give(&lte_connected);
+		k_sem_give(&lte_connected_sem);
 		break;
 	case LTE_LC_EVT_PSM_UPDATE:
 		printk("PSM parameter update: TAU: %d s, Active time: %d s\n",
@@ -183,7 +183,7 @@ int main(void)
 		return -1;
 	}
 
-	k_sem_take(&lte_connected, K_FOREVER);
+	k_sem_take(&lte_connected_sem, K_FOREVER);
 
 	err = socket_connect();
 	if (err) {
--- a/samples/cellular/udp/src/main.c
+++ b/samples/cellular/udp/src/main.c
@@ -16,8 +16,10 @@
 static int client_fd;
 static struct sockaddr_storage host_addr;
 static struct k_work_delayable socket_transmission_work;
+static int data_upload_iterations = CONFIG_UDP_DATA_UPLOAD_ITERATIONS;
 
 K_SEM_DEFINE(lte_connected, 0, 1);
+K_SEM_DEFINE(modem_shutdown_sem, 0, 1);
 
 static void socket_transmission_work_fn(struct k_work *work)
 {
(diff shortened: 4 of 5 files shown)


samples/cellular/udp/src/main.c
```diff
@@ -126,6 +128,19 @@ static void lte_handler(const struct lte_lc_evt *const evt)
...
+	char resp[32];
+
+	if (nrf_modem_at_cmd(resp, sizeof(resp), "AT+CGMM") == 0) {
+		if (strstr(resp, "nRF9160") != NULL) {
+			return true;
+		}
+	}
+
+	return false;
+}
```
lemrey, line 142: This sample only works on nRF 9 series SoCs. It may have had support for `qemu_cortex_m3` earlier, but that is no longer the case. I do not see why this function is necessary anymore, can we remove it?

samples/cellular/udp/Kconfig
```diff
@@ -14,6 +14,14 @@ config UDP_DATA_UPLOAD_FREQUENCY_SECONDS
 	int "How often data is transmitted to the server"
 	default 900
 
+config UDP_DATA_UPLOAD_FOREVER
+	bool "Enable that data is transmitted to the server forever"
+	default y
+
+config UDP_DATA_UPLOAD_ITERATIONS
+	int "NUmber of data transmissions to the server before shutdown" if !UDP_DATA_UPLOAD_FOREVER
+	default 5
```
lemrey, line 23: We shouldn't need two entries for this; one config for the number of iterations should be sufficient.
If set to `-1` let the loop run forever.

samples/cellular/udp/src/main.c
```diff
@@ -239,5 +269,14 @@ int main(void)
 
 	k_work_schedule(&socket_transmission_work, K_NO_WAIT);
 
+#if defined(CONFIG_NRF_MODEM_LIB) && !defined(CONFIG_UDP_DATA_UPLOAD_FOREVER)
```
lemrey, line 272: You can remove all the occurrences where we check for `NRF_MODEM_LIB`. This sample runs only on nRF9 series boards, so this is always defined. Keep it a separate commit if you can.

samples/cellular/udp/src/main.c
```diff
@@ -174,6 +189,21 @@ static int modem_connect(void)
...
+	int err;
+
+	err = nrf_modem_lib_shutdown();
+	if (err) {
+		printk("Failed to shutdown modem library, error: %d\n", err);
+		return err;
+	}
+
+	return 0;
+}
```
lemrey, line 205: There is no point to defined this function to call `nrf_modem_lib_shutdown()`; just call `nrf_modem_lib_shutdown()` directly.

samples/cellular/udp/src/main.c
MirkoCovizzi: I'll update the sample accordingly then. I was under the assumption that qemu target was still required.

samples/cellular/udp/Kconfig
MirkoCovizzi: I thought about the -1 idea before, but I don't like the fact that a different value has a completely different behavior. While these two options make sense just by reading their name, in the -1 case the user will have to find such behavior in the documentation instead.

MirkoCovizzi: @MechanicV1 made me aware that the `download_client` uses `-1` already in a similar way https://github.com/nrfconnect/sdk-nrf/blob/8f8c2689151063428e0f52645762ebedaecf78ab/subsys/net/lib/download_client/Kconfig#L112 so I updated this sample to align to that.

samples/cellular/udp/Kconfig
```diff
@@ -14,6 +14,13 @@ config UDP_DATA_UPLOAD_FREQUENCY_SECONDS
 	int "How often data is transmitted to the server"
 	default 900
 
+config UDP_DATA_UPLOAD_ITERATIONS
+	int "NUmber of data transmissions to the server before shutdown"
```
eivindj-nordic, line 18: nit:
```suggestion
	int "Number of data transmissions to the server before shutdown"
```

samples/cellular/udp/Kconfig
```diff
@@ -14,6 +14,13 @@ config UDP_DATA_UPLOAD_FREQUENCY_SECONDS
...
+	default -1
+	help
+	  NUmber of data transmissions to the server before shutdown.
```
eivindj-nordic, line 21:
```suggestion
	  Number of data transmissions to the server before shutdown.
```

eivindj-nordic (approved):

samples/cellular/udp/Kconfig
```diff
@@ -14,6 +14,13 @@ config UDP_DATA_UPLOAD_FREQUENCY_SECONDS
 	int "How often data is transmitted to the server"
 	default 900
 
+config UDP_DATA_UPLOAD_ITERATIONS
```
eivindj-nordic, line 17: Do we want a range here as well, as in the download client?

MirkoCovizzi: I thought about it and to me it doesn't make sense to add it

samples/cellular/udp/src/main.c
```diff
@@ -63,6 +65,17 @@ static void socket_transmission_work_fn(struct k_work *work)
...
 
+	if (data_upload_iterations > 0) {
+		data_upload_iterations--;
+		goto transmit;
+	} else if (data_upload_iterations == 0) {
+		k_sem_give(&modem_shutdown_sem);
+		return;
+	} else if (data_upload_iterations < 0) {
+		goto transmit;
+	}
```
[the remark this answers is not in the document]
MirkoCovizzi: That looks smart but smart is not always better. I'd rather be explicit and clear. It's not going to make it more efficient anyway.

MechanicV1 (approved):

samples/cellular/udp/src/main.c
```diff
@@ -63,6 +65,21 @@ static void socket_transmission_work_fn(struct k_work *work)
...
+		goto transmit;
+	} else if (data_upload_iterations == 0) {
+		k_sem_give(&modem_shutdown_sem);
+		return;
+	}
+
+	/* Transmit indefinitely. */
+	if (CONFIG_UDP_DATA_UPLOAD_ITERATIONS == -1) {
+		goto transmit;
+	}
```
eivindj-nordic, line 80: This is not necessary

samples/cellular/udp/Kconfig
```diff
@@ -14,6 +14,14 @@ config UDP_DATA_UPLOAD_FREQUENCY_SECONDS
...
+	int "Number of data transmissions to the server before shutdown"
+	range -1 2147483647
+	default -1
```
lemrey, line 20: I think the sample should terminate by default, so that we can run it in CI.

lemrey (changes requested):

samples/cellular/udp/Kconfig
MirkoCovizzi: @MechanicV1 From CI point of view should this sample terminate by default?

eivindj-nordic: CI can change the Kconfig value when building it. Though many other samples are terminating by default.

MechanicV1: Test automation will take care of setting the Kconfig ! @MirkoCovizzi

MechanicV1 (approved):

samples/cellular/udp/Kconfig
lemrey: I think we should implement the same behavior as other sample. It's easy enough to change that for users if needed.
CI might want to override the value anyway to shorten the time it takes to complete the test.

lemrey (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -316,6 +316,10 @@ Cellular samples
 
   * Added the configuration overlay file :file:`overlay-supl.conf` for building the sample with SUPL assistance support.
 
+* :ref:`udp` sample:
+
+  * Added the :kconfig:option:`CONFIG_UDP_DATA_UPLOAD_ITERATIONS` Kconfig option for configuring the number of data transmissions to the server.
```
divipillai, line 321:
```suggestion
  * Added the :ref:`CONFIG_UDP_DATA_UPLOAD_ITERATIONS <CONFIG_UDP_DATA_UPLOAD_ITERATIONS>` Kconfig option for configuring the number of data transmissions to the server.
```

samples/cellular/udp/README.rst
```diff
@@ -62,6 +62,12 @@ CONFIG_UDP_DATA_UPLOAD_SIZE_BYTES - UDP data upload size configuration
 CONFIG_UDP_DATA_UPLOAD_FREQUENCY_SECONDS - UDP data upload frequency configuration
    This configuration option sets the frequency with which the sample transmits data to the server.
 
+.. _CONFIG_UDP_DATA_UPLOAD_ITERATIONS:
+
+CONFIG_UDP_DATA_UPLOAD_ITERATIONS - UDP data upload iterations configuration
+   This configuration option sets the number of times the sample transmits data to the server before shutting down.
+   Set to -1 to transmit indefinitely.
```
divipillai, line 69:
```suggestion
   Set to ``-1`` to transmit indefinitely.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
MirkoCovizzi: @divipillai Is this the new standard syntax to use from now on?

divipillai: This is the way to represent the application-specific Kconfig options. For example, for UDP sample, the ones that are defined here: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/cellular/udp/README.html#configuration. The Kconfig options defined in the Kconfig reference (library-specific) are provided with the ":kconfig:option:" syntax.

divipillai (approved):

PR #13071, opened 2023-11-13
event_manager: Remove unexpected reboot on alloc fail

ball-hayden:
Having the application reboot unexpectedly is ... unexpected. By convention, any function that is an allocator is expected to return `null` if it is unable to allocate the requested memory.

Remove the confusing reboot, and instead return `NULL` so that the application can decide what to do.

### DIFF
--- a/subsys/app_event_manager/app_event_manager.c
+++ b/subsys/app_event_manager/app_event_manager.c
@@ -121,11 +121,7 @@ void * __weak app_event_manager_alloc(size_t size)
 	if (unlikely(!event)) {
 		LOG_ERR("Application Event Manager OOM error\n");
 		__ASSERT_NO_MSG(false);
-		if (IS_ENABLED(CONFIG_REBOOT)) {
-			sys_reboot(SYS_REBOOT_WARM);
-		} else {
-			k_panic();
-		}
+		k_panic();
 		return NULL;
 	}
 


lemrey:
Thank you for your contribution to NCS, I have enabled CI and added reviewers :)

pdunaj (changes requested):
This is by design. The assumption is that during an integration period, heap size would be chosen to meet the memory requirements for the device.
Assert and LOG are there to help.

If integrator thinks that this is not needed, they can add their own implementation but note this is not handled in standard event handlers. As mentioned above it is assumed that heap is set to meet required level.

pdunaj:
> This is by design. The assumption is that during an integration period, heap size would be chosen to meet the memory requirements for the device. Assert and LOG are there to help.
> 
> If integrator thinks that this is not needed, they can add their own implementation but note this is not handled in standard event handlers. As mentioned above it is assumed that heap is set to meet required level.

Maybe I should add a word of justification. This is not a standard memory allocation function. This is a core of application framework. If it does not work, you cannot process event and so my extension you can assume nothing works. Processing this null later is just delaying inevitable that you have entered an area your application will not able able to handle by itself and needs to reboot.

If you need something special, that does not require event system to work (since it does not work as there was no memory for it to work), you an reimplement your own allocation function. The function you refer is weak so you can override it with your own implementation. There you can add whatever you want that device should do to recover sanity. Personally I suggest reset. Maybe you can record an error status in some UICR and use it after reboot to inform the user of the device that something is very, very wrong.

ball-hayden:
> Assert and LOG are there to help.

With async logging, the reboot kicks in before the log message is printed.
The log messages we were seeing were:

```
Everything is fine
Everything is fine
Everything is fine

*** Booting Zephyr OS build v3.3.99-ncs1-1 ***
```

It took us a long time to find this.

If this is by design, I feel we should always use `k_panic` (even if `CONFIG_REBOOT` is set) - when panicking, async logs will be flushed properly?

ball-hayden:
> inform the user of the device that something is very, very wrong.

In our use case, this is kind-of expected. We're generating app events when we receive data over UART, and the remote device is bursty in its communication.

We expect, sometimes, that we might not be able to allocate memory for all of them, and we can request repeats of any that we miss.

> you an implement your own allocation function. The function you refer is weak so you can override it

Thank you - I didn't spot this. This will solve our immediate problem.

I do feel that using `k_panic` always would be a better option, though, for the reason above about async logging.

ball-hayden:
@pdunaj would you consider this updated PR?

pdunaj:
> @pdunaj would you consider this updated PR?

Missed that comment. It was dig out by @MarekPieta . We have discussed this internally. Marek proposed to add the config option so behavior can be selectable. He can issue a PR.

We could live it as `k_panic` only as this is essentially just that. The reason I would keep reboot by default is the behavior of `k_panic`. In production devices it should cause a reset. Theoretically this would still be reset by watchdog but only if enabled.

pdunaj:
@MarekPieta , before introducing a new config, please check if k_sys_fatal_error_handler from NCS is triggered in this k_panic if so  check if we can gracefully enable CONFIG_RESET_ON_FATAL_ERROR in the module. Then k_panic would be enough.

pdunaj:
btw - when this change is done you can as well visit what k_sys_fatal_error_handler  does as it seems there is a dead code in the source (the file is never compiled in unless CONFIG_RESET_ON_FATAL_ERROR is enabled but this option is still checked in the file).

MarekPieta:
Hi @pdunaj, @ball-hayden 
I prepared a PR with fix that would allow to explicitly select between `k_panic` and `sys_reboot` in Event Manager configuration: https://github.com/nrfconnect/sdk-nrf/pull/14546. The PR also maintains backwards compatibility, Please let me know what do you think about it.

I would prefer to explicitly call reset, instead of always going through the `k_panic` here. It may be relevant for an application to easily distinguish Event Manager OOM from `k_panic` triggered by something else. An application may simply reboot on OOM and perform some additional actions on `k_panic` triggered by something else (e.g. it may want to clear settings NVM area to prevent incompatibility of non-volatile data).

PR #13073, opened 2023-11-13
applications: asset_tracker_v2: Use path relative to Zephyr SDK

simensrostad:
Use path relative to Zephyr SDK to support out of tree builds.


didrikr (approved):
Thanks

KajaKoren (approved):

PR #13075, opened 2023-11-14
Add HTTP server sample

simensrostad:
Add HTTP server sample


samples/net/http_server/src/main.c
```diff
@@ -0,0 +1,678 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
```
jukkar, line 4: The base of the application (minus the http protocol handling) seems to be taken from upstream
`samples/net/sockets/dumb_http_server_mt` sample. In this respect changing the license does not look right.
Perhaps it would make sense to send this application to upstream instead, there does not seem to be any Nordic specific stuff here.
cc: @carlescufi

simensrostad: I forgot to include the orignal intel license, will add that. We are using the connection manager + net l2 backend for wifi which is only downstream. There might even be support for LTE at some point. But wifi was prioritized first.

simensrostad: I don't how to properly handle this licensing. Some pointers here would do good. Maybe @ioannisg also have some ideas.

jukkar: There are two things here, copyright and license.
We should add original copyright together with the Nordic one, I do not see any issue here.
The more problematic one is the license which does not look to be compatible with Apache-2 so adding two licenses here feels questionable. But I am not a lawyer so this might be wrong.

About the sample itself. There is actually effort in upstream zephyr to create a HTTP server library and sample application (in PR 64465). I took the abandoned GSoC 2023 HTTP server project and continued its work on my free time. Unfortunately my progress have been very slow. Should I have known about this PR, I would have suggested that our efforts would have been concentrated enhancing the upstream version as it also supports HTTP/2.

simensrostad: A HTTP server library would do good, the HTTP handling is quite raw. Avoiding direct socket and protocol handling would provide a nice abstraction. This will also make it easier for customers to add their own logic.

This sample was made in parallel with the wifi dev academy course, as some exercises in the course have been based on this sample. This gave us a good opportunity to make a _proper_ sample that could live in NCS,  supporting both wifi and lte. There has also been some interest in the wifi space for this sample in relation to home automation which has pushed the sense of urgency a bit.

But I guess the biggest blocker is the fact that an upstream version of this sample wouldn't support LTE, AFAIK.

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,151 @@
...
+
+.. table-from-sample-yaml::
+
+.. include:: /includes/tfm.txt
+
+Overview
+********
+
+The sample supports setting and getting the state of LED 1 and 2 on a supported development kit via the default server URL: *http://httpserver.local:8080/led/<1/2>* or *https://httpserver.local:443/led/<1/2>* (TLS).
+The server supports both incoming IPv4 and IPv6 connections, but only works on localhost, meaning that the server is not reachable from the open internet.
```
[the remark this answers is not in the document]
simensrostad: yes, thanks will update.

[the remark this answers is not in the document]
simensrostad: Ill probably have to specify this part a bit and separate it when LTE support is added. Ill reach out when the final docs are in place.

samples/net/http_server/Kconfig
```diff
@@ -0,0 +1,49 @@
...
+
+config HTTP_SERVER_SAMPLE_SERVER_PORT_TLS
+	int "Server Port (TLS)"
+	default 443
+
+config HTTP_SERVER_SAMPLE_CLIENTS_MAX
+	int "Maximum clients"
+	default 2
+	help
+	  This option sets the maximum number of concurrent clients.
```
ioannisg, line 21: Might be interesting to add a sentence on the impact of increasing the number of clients (e..g memory footprint, performance, etc.)

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,151 @@
+.. _http_server:
+
+HTTP server
+###########
+
+The HTTP server sample demonstrates how to host a HTTP server on a nRF70 Series device.
```
ioannisg, line 6:
```suggestion
The HTTP server sample demonstrates how to host an HTTP server on an nRF70 Series device.
```

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,151 @@
...
+
+.. include:: /includes/tfm.txt
+
+Overview
+********
+
+The sample supports setting and getting the state of LED 1 and 2 on a supported development kit via the default server URL: *http://httpserver.local:8080/led/<1/2>* or *https://httpserver.local:443/led/<1/2>* (TLS).
+The server supports both incoming IPv4 and IPv6 connections, but only works on localhost, meaning that the server is not reachable from the open internet.
+The sample supports mDNS queries with a hostname set by the :kconfig:option:`CONFIG_NET_HOSTNAME` option.
+Depending on the performed HTTP call and state of the device the device will return typical HTTP reponses such as *HTTP/1.1 200 OK* and *HTTP/1.1 400 Bad Request*.
```
ioannisg, line 27:
```suggestion
Depending on the performed HTTP call and state of the device, the device will return typical HTTP responses such as *HTTP/1.1 200 OK* and *HTTP/1.1 400 Bad Request*.
```

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,151 @@
...
+Requirements
+************
+
+The sample supports the following development kits:
+
+.. table-from-sample-yaml::
+
+.. include:: /includes/tfm.txt
+
+Overview
```
ioannisg, line 21: I would add a sentence listing the supported transport options
- Wi-Fi
- LTE (future)
and the fact that this is using the connection manager by default.

ioannisg (commented):
Some minor review comments from me, @simensrostad :)

samples/net/http_server/src/main.c
ioannisg: Regarding the license, I agree @carlescufi would be best to assist.

Regarding the sample itself:
- The HTTP server library would be very nice @jukkar. Without being an expert, I believe that any HTTP server sample (upstream or downstream) could be made to use this HTTP server library, once this library is available. 
- I do not see, however, much harm of having a dedicated downstream HTTP server sample (in addition to an upstream one). My thinking:
  - @simensrostad intended to use the connection manager library in the sample, and then share the sample code seamlessly between Wi-Fi and LTE. However, to the best of my knowledge, using the connectivity manager in upstream samples is of little value as there is no transport backend supported directly.
  - As mentioned, the upstream HTTP server example would not be directly usable for Nordic users/customers upstream (there is neither Nordic Wi-Fi, nor LTE support in the upstream project, for now)
  -  We could potentially scope or add functionality to this example that would be specific to Nordic needs, so it helps if the example is hosted in Nordic SDK.

Alternatively, one option would be to host the sample upstream, and support it on nRF70 Series in Nordic's downstream fork of Zephyr (with a Wi-Fi overlay, as @rlubos has done for other upstream Zephyr net examples..). But I am not sure how it could be made to work for LTE and nRF91 Series. Note that the original reason for creating the `samples/net` directory in `sdk-nrf` repository was to offer a set of NRF SDK networking examples, with a common code base for LTE and Wi-Fi using the connectivity (abstraction) manager feature.

simensrostad:
Ill address review comments next week when I have tested with LTE. Since LTE support is only config additions ill add it to this PR. Keeping this PR DNM until everything is in place.

samples/net/http_server/README.rst
simensrostad: Will refurbish the docs a bit after I've tested with cellular

samples/net/http_server/src/main.c
```diff
@@ -1,7 +1,8 @@
 /*
+ * Copyright (c) 2019 Intel Corporation
  * Copyright (c) 2024 Nordic Semiconductor ASA
  *
- * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ * SPDX-License-Identifier: Apache-2.0
```
jukkar, line 5: This change should be in the first commit.
The license does not seem to be compatible with nordic one?

simensrostad: Im going to squash the commit

simensrostad: :confused:

simensrostad: cc @ioannisg Can you comment regarding the current copyright/license header

ioannisg: Using Apache 2.0 license header in sdk-nrf source files is OK. If we copied this from upstream Zephyr we should use the upstream SPDX.

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,187 @@
...
+
+Add instructions on how to use IPv6.
+ - Enable IPv6 in router settings, passthrough
+ - Find IPv6 address using net ipv6 command
+ - Specify IPv6 address in HTTP request: http PUT 'http://[2001:8c0:5140:895:f6ce:36ff:fe00:1970]:8080/led/1' --raw="1"
+
+LTE
+===
+
+SIM card dependent, etc...
```
ioannisg, line 51: Fix?

samples/net/http_server/boards/nrf7002dk_nrf5340_cpuapp_ns.conf
```diff
@@ -0,0 +1,51 @@
+#
```
ioannisg, line 1: @simensrostad  do you see a benefit of including also an overlay for the default (Secure) 5340 Application Core build?

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,167 @@
+.. _http_server:
+
+HTTP server
+###########
+
+The HTTP server sample demonstrates how to host an HTTP server that runs on a Internet connected Nordic device via cellular or Wi-Fi.
+Cellular connectivity is supported on the nRF91 Series SiP, while Wi-Fi connectivity is supported on the nRF52 or nRF53 Series SoCs hosting the nRF70 Series Wi-Fi companion IC.
```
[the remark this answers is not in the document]
ioannisg: Not for Wi-Fi Connectivity (at the moment at least)

samples/net/http_server/boards/nrf7002dk_nrf5340_cpuapp_ns.conf
simensrostad: Not really. We should inspire customers to build for NS. For security reasons.

samples/net/http_server/README.rst
simensrostad: Jepp, working on it.

samples/net/http_server/Kconfig
```diff
@@ -0,0 +1,51 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menu "HTTP server sample"
+
+config HTTP_SERVER_SAMPLE_SERVER_PORT
```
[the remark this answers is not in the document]
simensrostad: Sounds good, Ill make the change.

samples/net/http_server/Kconfig
```diff
@@ -0,0 +1,51 @@
...
+	int "Server Port"
+	default 8080
```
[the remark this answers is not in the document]
simensrostad: No particular reason, just a port number: Adopted from the upstream zephyr sample. I can change it to 80 if that is desired.

samples/net/http_server/Kconfig
```diff
@@ -0,0 +1,51 @@
...
+
+config HTTP_SERVER_SAMPLE_SERVER_PORT_TLS
```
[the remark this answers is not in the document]
simensrostad: Yes, that a good improvement. Ill fix that.

samples/net/http_server/Kconfig
```diff
@@ -0,0 +1,51 @@
...
+	int "Server Port (TLS)"
+	default 443
+
+config HTTP_SERVER_SAMPLE_CLIENTS_MAX
+	int "Maximum clients"
+	default 2
+	help
+	  This option sets the maximum number of concurrent clients.
```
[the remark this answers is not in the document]
simensrostad: True, that should be 4 in total so its 2 per IP family. Ill add some help text.

samples/net/http_server/src/main.c
```diff
@@ -0,0 +1,673 @@
...
+			break;
+		} else if (received < 0) {
+			/* Socket error */
+			ret = -errno;
+			LOG_ERR("[%d] Connection error %d", request.socket, ret);
+			break;
+		}
+
+		/* Parse the received data as HTTP request */
+		(void)http_parser_execute(&request.parser,
```
gregersrygg, line 441: Should check error from `http_parser_execute`.

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,167 @@
...
+The following serial UART output is displayed in the terminal emulator when running the sample:
+
+Wi-Fi
+
+.. code-block:: console
+
+   *** Booting nRF Connect SDK v1.1.0-rc1-15416-gb0a135cfb3a4 ***
+   [00:00:00.535,430] <inf> http_server: HTTP Server sample started
+   [00:00:00.563,537] <inf> http_server: Network interface brought up
+   uart:~$ wifi connect <ssid> <password>
```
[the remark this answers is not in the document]
simensrostad: Good point, wifi cred add actually stores credentials to PSA storage, upon reboot the device will automatically use stored values. Wifi connect doesn't do that (atm atleast). Ill update the logs to use wifi cred.

samples/net/http_server/boards/nrf9160dk_nrf9160_ns.conf
```diff
@@ -0,0 +1 @@
+COPY
```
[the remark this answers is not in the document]
simensrostad: tyy :1st_place_medal:

samples/net/http_server/src/main.c
```diff
@@ -0,0 +1,673 @@
...
+			&client_addr_len);
+	if (client < 0) {
+		LOG_ERR("Error in accept %d, try again", -errno);
+		return;
+	}
+
+	slot = get_free_slot(accepted);
+	if (slot < 0 || slot >= MAX_CLIENT_QUEUE) {
+		LOG_ERR("Cannot accept more connections");
+		close(client);
```
gregersrygg, line 495: Ideally, we should return «503 Service Unavailable» to the client.

samples/net/http_server/Kconfig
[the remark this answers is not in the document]
gregersrygg: Please do. Then the user doesn't need to specify port number in the URL. So the non-https URLs can also be updated to not include `:8080`.

samples/net/http_server/src/main.c
```diff
@@ -0,0 +1,673 @@
...
+		return;
+	}
+}
+
+/* Update the LED states. Returns 0 if it was updated, otherwise -1. */
+static int led_update(uint8_t index, uint8_t state)
+{
+	int ret;
+
+	if (state <= 1) {
```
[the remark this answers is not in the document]
simensrostad: Idea here was to keep the conversion from led id to index outside the LED update function. So that the LED update function only need to care about the index and state that it gets.

samples/net/http_server/src/main.c
```diff
@@ -0,0 +1,673 @@
...
+	char *resp_ptr = response_200;
+	char get_response_buffer[100] = { 0 };
+
+	/* Handle the request method */
+	switch (request->method) {
+	case HTTP_PUT:
+		ret = handle_put(request);
+		if (ret) {
+			LOG_WRN("Incoming HTTP PUT request, error: %d", ret);
+			resp_ptr = response_400;
```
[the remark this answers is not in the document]
simensrostad: Sounds good, ill make the distinction using different retvals.

samples/net/http_server/credentials/private_key.pem.h
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+static const unsigned char private_key[] =
```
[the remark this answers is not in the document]
balaji-nordic: Try this method shown here https://github.com/nrfconnect/sdk-nrf/blob/c95907c3271db44ab0a7d6e89bae3a649f473c94/samples/net/download/src/main.c#L40C7-L40C7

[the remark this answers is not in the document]
simensrostad: Ill create a ticket for this to be added in all of the samples that provisions and includes credentials like we do here.

samples/net/http_server/src/main.c
simensrostad: I think ill leave this one as a feature improvement as there is no good clean way of doing it.

gregersrygg (approved):

samples/net/http_server/src/main.c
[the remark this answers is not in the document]
simensrostad: fixed

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,266 @@
...
+.. include:: /includes/net_connection_manager.txt
+
+.. contents::
+   :local:
+   :depth: 2
+
+Requirements
+************
+
+The sample supports the followiURLng development kits:
```
gregersrygg, line 17:
```suggestion
The sample supports the following development kits:
```

simensrostad: ^^  @divipillai  Could you include this suggestion?

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
```diff
@@ -143,7 +143,7 @@ static int add_network_from_credentials_struct_personal(struct wifi_credentials_
 		return -ENOEXEC;
 	}
 
-	LOG_INF("Connection requested");
+	LOG_DBG("Connection requested");
```
rlubos, line 146: Was this change intentinal here? If so, I'd put it in a separate commit, it's not really related to the sample.

samples/net/http_server/boards/nrf7002dk_nrf5340_cpuapp_ns.conf
```diff
@@ -0,0 +1,56 @@
...
+# Wi-Fi
+CONFIG_WIFI=y
+CONFIG_WIFI_NRF700X=y
+CONFIG_WPA_SUPP=y
+CONFIG_NET_L2_ETHERNET=y
+
+# IP address config
+CONFIG_NET_CONFIG_INIT_TIMEOUT=0
+CONFIG_NET_CONFIG_SETTINGS=y
+CONFIG_NET_CONFIG_MY_IPV4_ADDR="192.0.2.1"
```
rlubos, line 28: Given that DHCP is used, maybe there's no point setting static address? That'd make the conn manager to notify interface readiness before we have a "valid" (DHCP assigned) address.

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,342 @@
...
+.. tabs::
+
+   .. tab:: Wi-Fi
+
+         Non TLS (unsecure):
+
+         .. code-block:: console
+
+            http PUT http://httpserver.local:80/led/1 --raw="1"
+            http GET http://httpserver.local:80/led/1
```
rlubos, line 224: Is there a way to enforce IPv6 with this command? If so, maybe it'd make sense to add a sample command for that case as well, given the server uses a different port for IPv6. Or at least add a comment about different port being used for IPv6 in the testing section, I think this info could be overlooked given it's only provided in the Kconfig seciton.

subsys/net/lib/wifi_mgmt_ext/wifi_mgmt_ext.c
simensrostad: I didn't intent to include this change, ill remove it.

samples/net/http_server/boards/nrf9160dk_nrf9160_ns.conf
```diff
@@ -0,0 +1,25 @@
...
+# This file is merged with prj.conf in the application folder, and options
+# set here will take precedence if they are present in both files.
+
+# General
+CONFIG_PICOLIBC=y
+
+# Disable Duplicate Address Detection (DAD)
+# due to not being properly implemented for offloaded interfaces.
+CONFIG_NET_IPV6_NBR_CACHE=n
+CONFIG_NET_IPV6_MLD=n
```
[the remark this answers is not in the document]
simensrostad: Without these changes the sample hard faults. Initially discovered when the _modem_net_if_ module was developed which integrates with zephyr NET conn mgr API for LTE. Fixing this has probably been forgotten or down prioritized, it might be that @glarsennordic created a ticket for it. Easy to reproduce.

samples/net/http_server/README.rst
simensrostad: Update the server to use the same port for IPv6, using the SO_REUSEADDR socketopt. Previously I got -112 without setting this option. EADDRINUSE

@jukkar Interesting, how would I do that?. I'm not sure if httpie supports forcing IPv6 other than explicitly using the IPv6 address directly in the request. Ill have to check that out.

samples/net/http_server/boards/nrf9160dk_nrf9160_ns.conf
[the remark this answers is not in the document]
jukkar: I have some doubts if this PR will work properly with IPv6 networks if MLD is disabled. MLD is pretty much mandatory for IPv6. How well we have tested this?

samples/net/http_server/README.rst
jukkar: > Interesting, how would I do that?. I'm not sure if httpie supports forcing IPv6 other than explicitly using the IPv6 address directly in the request. Ill have to check that out.

See for example the `setup_tcp_socket()` in echo-service sample in upstream
https://github.com/zephyrproject-rtos/zephyr/blob/68bea801119fe9a479f4a53185dfa0c95d26ce95/samples/net/sockets/echo_service/src/main.c#L117
how it is done.
So `CONFIG_NET_IPV4_MAPPING_TO_IPV6=y` needs to be set and `IPV6_V6ONLY` socket option needs to be turned off. The system creates IPv6 socket only, but it allows IPv4 to connect to it.

samples/net/http_server/boards/nrf7002dk_nrf5340_cpuapp_ns.conf
simensrostad: It shouldn't be necessary, I have removed it

samples/net/http_server/boards/nrf9160dk_nrf9160_ns.conf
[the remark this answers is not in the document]
simensrostad: This is only disabled for cellular builds. It shouldn't "have any effect" as IP is offloaded and handled by the modem anyway. I haven't had any related issues with IPv6 + conn mgr and modem net if before.
At least not related to this in particular.

So perhaps this is more about making this work properly when the net interface is offloaded.

samples/net/http_server/README.rst
simensrostad: I don't believe our modem supports this feature I'm afraid which will cause some displacement between wifi and lte builds. Alternative is to offload tcp/ip to native but I think we should use the modems tcp/ip stack.

samples/net/http_server/boards/nrf9160dk_nrf9160_ns.conf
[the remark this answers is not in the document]
jukkar: If modem handles it (MLD) itself, then it should be ok. But for wifi we would need to enable it.

samples/net/http_server/README.rst
jukkar: Indeed, for offloaded case we cannot use this atm.

samples/net/http_server/boards/nrf9160dk_nrf9160_ns.conf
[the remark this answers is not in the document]
simensrostad: Jepp, this is only for 91 builds since its in the 91 specific board file.

samples/net/http_server/README.rst
simensrostad: @rlubos Even though I use the same ports now I still have issues using the hostname directly in the request for IPv6. Ill add some docs on this shortcoming and add some cmnds on how to use IPv6 address direclty insead, until its resolved. It might be that this is only on my network setup.

samples/net/http_server/README.rst
```diff
@@ -0,0 +1,357 @@
...
+| Cellular   | Global\*         | \http://<ip>/led/<led_id>             | TLS (optional) | Server and client\**          | IPv4 IPv6              |
++------------+------------------+---------------------------------------+----------------+-------------------------------+------------------------+
+
+| * - When running the server on a cellular board, you are dependent on a SIM card with a fixed/static IP to be reachable from the global internet.
+      These SIM cards can be provided by Mobile Network Operators (MNO) and are often referred to as IoT SIM cards.
+| ** - The sample includes pregenerated server and client certificates with the local hosthame *httpserver.local* used as the Common Name (CN). These ceritifcates will work out of the box when building the sample for Wi-Fi.
+       On cellular, a static IP is required to reach the server.
+       This means that a new set of server and client certificates needs to be generated, with a CN matching the assigned IP.
+| *** - IPv4 and/or IPv6 may not be enabled in your W-FI Access Point (AP) or cellular network.
```
divipillai, line 48:
```suggestion
| \* - When running the server on a cellular board, you are dependent on a SIM card with a fixed/static IP to be reachable from the global internet.
       These SIM cards can be provided by Mobile Network Operators (MNO) and are often referred to as IoT SIM cards.
| \** - The sample includes pregenerated server and client certificates with the local hosthame *httpserver.local* used as the Common Name (CN).
        These ceritifcates will work out of the box when building the sample for Wi-Fi.
        On cellular, a static IP is required to reach the server.
        This means that a new set of server and client certificates needs to be generated, with a CN matching the assigned IP.
| \*** - IPv4 and/or IPv6 may not be enabled in your W-FI Access Point (AP) or cellular network.
```

divipillai (approved):
fixes the doc build error.

samples/net/http_server/src/main.c
```diff
@@ -0,0 +1,718 @@
...
+ */
+static uint8_t led_states[2];
+
+/* HTTP responses for demonstration */
+#define response_200 "HTTP/1.1 200 OK\r\n"
+#define response_400 "HTTP/1.1 400 Bad Request\r\n\r\n"
+#define response_403 "HTTP/1.1 403 Forbidden\r\n\r\n"
+#define response_404 "HTTP/1.1 404 Not Found\r\n\r\n"
+#define response_405 "HTTP/1.1 405 Method Not Allowed\r\n\r\n"
+#define response_500 "HTTP/1.1 500 Internal Server Error\r\n\r\n"
```
bjda, line 79: Should we make the macros all caps?
```suggestion
#define RESPONSE_200 "HTTP/1.1 200 OK\r\n"
#define RESPONSE_400 "HTTP/1.1 400 Bad Request\r\n\r\n"
#define RESPONSE_403 "HTTP/1.1 403 Forbidden\r\n\r\n"
#define RESPONSE_404 "HTTP/1.1 404 Not Found\r\n\r\n"
#define RESPONSE_405 "HTTP/1.1 405 Method Not Allowed\r\n\r\n"
#define RESPONSE_500 "HTTP/1.1 500 Internal Server Error\r\n\r\n"
```

bjda (commented):
Makes sense to me. Left a couple of comments

samples/net/http_server/src/main.c
simensrostad: added

samples/net/http_server/boards/nrf9160dk_nrf9160_ns.conf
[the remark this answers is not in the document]
glarsennordic: @simensrostad No ticket that I am aware of -- as you mention, I don't think it's an actual bug, is it?

bjda (approved):

samples/net/http_server/boards/nrf9160dk_nrf9160_ns.conf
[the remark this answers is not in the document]
simensrostad: Well, if you remove those all samples that use the modem net if + conn mgrf implementation hard faults. So depends on how you look at it. You shouldn't really be allowed to enable MLD (and likely other IP related modules as well) if the TCP/IP stack is offloaded.

Ok, I thought this was scheduled to be addressed. Ill see if I can locate the discussions we had regarding it and formulate a ticket. Thanks

rlubos (approved):

jukkar (approved):

PR #13089, opened 2023-11-15
samples: matter: Fixed crash after exceeding providers number

kkasperczyk-no:
The Matter bridge application crashes after adding provider that exceeds the max allowed number. It happens only if the max number of allowed bridged endpoints is bigger and would allow to complete this operation.

Removed unique pointers used to manage the objects lifetime, as AddSingleDevice method already releases the memory, what leads to double-delete operations and application crash.


markaj-nordic (approved):
LGTM, although I'd suggest adding doxygen documentation of the AddSingleDevice to explain method's implementation (even though it's not a part of API)

LuDuda (approved):

PR #11845, opened 2023-07-20
doc: introduction: rework section

greg-fer:
Changed the structure and contents of the Introduction section. NCSDK-22322.

----

- This is step 1 of 2. Step 2 will add new information on the Intro page, after discussing with stakeholders what exactly we want there.
- The Welcome to the SDK page and Introduction page will coexist now, but the Welcome to the SDK page will be dropped later this year.

PR #13090, opened 2023-11-15
doc: Rearranged items on the ncs intro page

annwoj:
Substituted the ncs intro page with the introduction content.


greg-fer (changes requested):
I'm not convinced we should be touching this page. The network protocols are listed in full here: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/introduction.html

As mentioned in #11845, the Welcome page will be dropped once we switch 100% to TechDocs, so I don't see the point of updating it.

doc/nrf/index.rst
```diff
@@ -1,22 +1,42 @@
 .. _index:
```
greg-fer, line 1:
```suggestion
.. _index:
.. _ncs_introduction:
```

greg-fer (changes requested):
This is great, but we need to keep the old page tag. The proposed edit should fix the doc build.

greg-fer (approved):

DeidreCasey (approved):

PR #13097, opened 2023-11-15
samples: cellular: modem_shell: link: Improve command line parsing

trantanen:
Improve error handling of invalid command line options.

Jira: MOSH-530

### DIFF
--- a/samples/cellular/modem_shell/src/link/link_shell.c
+++ b/samples/cellular/modem_shell/src/link/link_shell.c
@@ -51,17 +51,17 @@ enum link_shell_command {
 	LINK_CMD_PROPRIPSM
 };
 
-enum link_shell_common_options {
-	LINK_COMMON_NONE = 0,
-	LINK_COMMON_READ,
-	LINK_COMMON_WRITE,
-	LINK_COMMON_ENABLE,
-	LINK_COMMON_DISABLE,
-	LINK_COMMON_SUBSCRIBE,
-	LINK_COMMON_UNSUBSCRIBE,
-	LINK_COMMON_START,
-	LINK_COMMON_STOP,
-	LINK_COMMON_RESET
+enum link_shell_operation {
+	LINK_OPERATION_NONE = 0,
+	LINK_OPERATION_READ,
+	LINK_OPERATION_WRITE,
+	LINK_OPERATION_ENABLE,
+	LINK_OPERATION_DISABLE,
+	LINK_OPERATION_SUBSCRIBE,
+	LINK_OPERATION_UNSUBSCRIBE,
+	LINK_OPERATION_START,
+	LINK_OPERATION_STOP,
+	LINK_OPERATION_RESET
 };
 
 static const char link_settings_usage_str[] =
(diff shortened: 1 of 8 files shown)


trantanen:
I've modified few subcommands for now so that we can agree on the style before I do it for all subcommands as there is a bit of similar work and don't want to fix same style issues to ~15 places.

samples/cellular/modem_shell/src/link/link_shell.c
```diff
@@ -1012,7 +1012,7 @@ static int link_shell_dnsaddr(const struct shell *shell, size_t argc, char **arg
 static int link_shell_edrx(const struct shell *shell, size_t argc, char **argv)
 {
 	int ret = 0;
-	enum link_shell_common_options common_option = LINK_COMMON_NONE;
+	enum link_shell_common_options operation = LINK_COMMON_NONE;
```
trantanen, line 1015: This is intentionally still using the enum kind of approach while other functions use a boolean for each common option.

samples/cellular/modem_shell/src/link/link_shell.c
```diff
@@ -1076,12 +1082,17 @@ static int link_shell_edrx(const struct shell *shell, size_t argc, char **argv)
 				return -EINVAL;
 			}
 			break;
+
+		case '?':
+			mosh_error("Unknown option. See usage:");
+			goto show_usage;
+		case 'h':
 		default:
```
tokangas, line 1090: With the current approach, an option which is supported by some command, but not the current command, does not cause the "Unknown option" error to be printed. One way to fix this (and probably the way `getopt_long()` is intended to be used) would be to have separate option lists for each subcommand, with only the options that are supported by that particular subcommand. However, I wonder if there would be an easier way to work around this by printing the error also from the `default` branch? Would that have any side effects?

samples/cellular/modem_shell/src/link/link_shell.c
trantanen: Decided to use enum

samples/cellular/modem_shell/src/link/link_shell.c
trantanen: This has been changed pretty much as suggested throughout the MOSH

samples/cellular/modem_shell/src/link/link_shell.c
```diff
@@ -1739,9 +1955,13 @@ static int link_shell_search(const struct shell *shell, size_t argc, char **argv
 	int opt;
 
 	while ((opt = getopt_long(argc, argv, short_options, long_options, NULL)) != -1) {
-		link_shell_getopt_common_option(opt, &common_option);
-
 		switch (opt) {
+		case 'r':
+		case LINK_SHELL_OPT_WRITE:
+		case LINK_SHELL_OPT_START:
```
tokangas, line 1961: Should there  be `LINK_SHELL_OPT_RESET` as well?

tokangas (changes requested):

samples/cellular/modem_shell/src/link/link_shell.c
trantanen: Yep. This is mapped from --clear.

samples/cellular/modem_shell/src/rest/rest_shell.c
```diff
@@ -239,17 +239,22 @@ static int rest_shell(const struct shell *shell, size_t argc, char **argv)
...
 		default:
+			mosh_error("Unknown option. See usage:");
 			show_usage = true;
 			goto end;
 		}
 	}
 
+	if (optind < argc) {
+		mosh_error("Arguments without '-' not supported: %s", argv[argc - 1]);
+		goto end;
```
tokangas, line 255: `show_usage` should be set here.

samples/cellular/modem_shell/src/startup_cmd/startup_cmd_shell.c
```diff
@@ -93,14 +93,20 @@ static int startup_cmd_shell(const struct shell *shell, size_t argc, char **argv
 			startup_cmd_mem_slot = 3;
 			break;
 
-		case '?':
-			mosh_error("Unknown option. See usage:");
-			goto show_usage;
 		case 'h':
```
tokangas, line 96: `-h` and `--help` options do not work with `startup_cmd`:

```
mosh:~$ startup_cmd -h
Unknown option. See usage:
...
```

```
mosh:~$ startup_cmd --help
Unknown option. See usage:
...
```

tokangas (changes requested):

samples/cellular/modem_shell/src/startup_cmd/startup_cmd_shell.c
trantanen: Fixed

samples/cellular/modem_shell/src/rest/rest_shell.c
trantanen: Fixed

tokangas (approved):

trantanen:
I'm copying memory report for these changes (some debug logging enabled but same config for both reports) which show 2728 bytes of additional flash usage:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:      633284 B       960 KB     64.42%
             RAM:      126216 B     197400 B     63.94%
        IDT_LIST:          0 GB         2 KB      0.00%
```
And plain ncs/main without this PR:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:      630556 B       960 KB     64.14%
             RAM:      126216 B     197400 B     63.94%
        IDT_LIST:          0 GB         2 KB      0.00%
```

jhirsi (approved):

PR #13106, opened 2023-11-16
Sysbuild fixes/improvements

nordicjm:
sysbuild: Fix secure boot image names
    
    Fixes not using the correct image names for secure boot targets

    sysbuild: Fix Kconfig dependency for b0 package
    
    Fixes an issue whereby this Kconfig was gated on the wrong
    Kconfig

    cmake: sybuild: partition_manager: Fix wrongly use all images
    
    Fixes wrongly using all images for the network core when it should
    instead only use those dedicated to the network core

    tfm: Fix wrong sysbuild hex file when MCUboot is disabled
    
    Fixes an issue whereby the wrong (default) zephyr hex file was
    used for TF-M images when using sysbuild, when it should instead
    use the merged hex file

    sysbuild: Abort if invalid secure boot configuration is set
    
    Errors out with a fatal error if the secure boot configuration is
    invalid, this prevents the user from waiting for a long period of
    time as all images are configured before seeing a fatal error

    applications: asset_tracker_v2: Add sysbuild secure boot core
    
    Enables secure boot on the application core when using sysbuild

    applications: asset_tracker_v2: Enable sysbuild builds
    
    Enables sysbuild builds for this sample

    sysbuild: Add rpc_host netcore image
    
    Allows selecting rpc_host as the network core image for nrf5340dk


applications/asset_tracker_v2/sample.yaml
```diff
@@ -10,6 +10,15 @@ tests:
...
     tags: ci_build
+  applications.asset_tracker_v2.nrf_cloud.sysbuild:
+    build_only: true
+    sysbuild: true
+    build_on_all: true
+    platform_allow: nrf9160dk_nrf9160_ns thingy91_nrf9160_ns
+    integration_platforms:
+      - nrf9160dk_nrf9160_ns
+      - thingy91_nrf9160_ns
+    tags: ci_build sysbuild
```
SebastianBoe, line 21: Is this going to be the first time that sysbuild is in NCS CI?

Or do sysbuild tags not run in CI?

nordicjm: Sysbuild has been in CI for quite some time in samples/bluetooth/fast_pair/input_device, samples/bluetooth/peripheral_uart, samples/multicore/hello_world, and samples/zigbee/shell

SebastianBoe: ah, I was not aware.

SebastianBoe (approved):

nordicjm:
@frkv @Vge0rge @vili-nordic @tejlmand @simensrostad @jtguggedal @jhn-nordic @coderbyheart please take a look

jtguggedal (approved):

PR #13118, opened 2023-11-16
samples: lwm2m_carrier: Fix use of implicitly defined function

joerchan:
Fix use of implicitly defined function lte_lc_connect_async. When LWM2M_CARRIER is not enabled the header file is not included.

### DIFF
--- a/samples/cellular/lwm2m_carrier/src/main.c
+++ b/samples/cellular/lwm2m_carrier/src/main.c
@@ -6,11 +6,11 @@
 
 #include <zephyr/kernel.h>
 #include <modem/nrf_modem_lib.h>
+#include <modem/lte_lc.h>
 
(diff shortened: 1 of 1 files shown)


lemrey:
@kacperradoszewski can you please have a look here?

kacperradoszewski:
Right, we were aware of this at one point and seemingly forgot about it... Then I suggest we move the lte_lc header outside of the CONFIG_LWM2M_CARRIER scope, next to the other non carrier-related headers.

joerchan:
> Right, we were aware of this at one point and seemingly forgot about it... Then I suggest we move the lte_lc header outside of the CONFIG_LWM2M_CARRIER scope, next to the other non carrier-related headers.

Thy will be done.

rlubos (approved):

PR #13119, opened 2023-11-16
samples: matter: Initial integration with sysbuild 

markaj-nordic:
Scope:
* added multiprotocol_rpmsg target to the existing sysbuild NCS infrastructure
* integrated Matter samples with sysbuild for basic configuration without DFU and factory data support enabled
* the nRF52 integration works already out of the box, for nRF53 the multiprotocol_rpmsg is needed

Tests:
* built and flashed all Matter samples with the following commands:
`mkdir build && cd build`
`cmake -GNinja -DBOARD=nrf5340dk_nrf5340_cpuapp -DAPP_DIR=.. ~/ncs/zephyr/share/sysbuild/ -DCONFIG_CHIP_FACTORY_DATA=n -DCONFIG_CHIP_FACTORY_DATA_BUILD=n -DCONF_FILE="prj_no_dfu.conf"`
`cd ..`
`west flash --erase`
* commissioned Matter lock sample to the Matter over Thread network with `chip-tool` and verified basic Matter commands

Comments/TODOs:
* the next step is to add support for DFU and factory data
* the net core KConfig configuration should be unified to avoid boilerplate - like it is done currently when using the native build system (https://github.com/nrfconnect/sdk-connectedhomeip/blob/master/config/nrfconnect/chip-module/Kconfig.multiprotocol_rpmsg.root)


samples/matter/light_bulb/Kconfig.sysbuild
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+source "${ZEPHYR_BASE}/share/sysbuild/Kconfig"
+
+config NRF_DEFAULT_MULTIPROTOCOL
```
kkasperczyk-no, line 9: Is it going to work for the hci_rpmsg case (Wi-Fi) as well?

samples/matter/light_bulb/sysbuild/multiprotocol_rpmsg/prj.conf
```diff
@@ -0,0 +1,45 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+## Disable serial and UART interface.
```
kkasperczyk-no, line 7: Would it be possible to inherit this configuration from the Kconfig.multiprotocol_rpmsg.defaults, similarly, as we have it for the child images?

samples/matter/light_bulb/sysbuild/multiprotocol_rpmsg/prj.conf
```diff
@@ -0,0 +1,45 @@
...
+CONFIG_SERIAL=n
+CONFIG_UART_CONSOLE=n
```
kkasperczyk-no, line 9: Generic comment to the file, is it going to work with all kinds of prj+suffix.confs? E.g. for child images we had to duplicate it in case of having several different prj config names.

kkasperczyk-no (commented):
Did you manage to test nRF7002 DK as well?

kkasperczyk-no:
One more comment, please align Matter weather station and Matter bridge applications to the changes introduced in samples.

samples/matter/light_bulb/Kconfig.sysbuild
markaj-nordic: No, as I don't see support for nRF70 family added. @nordicjm, can you share what the plan is?

samples/matter/light_bulb/sysbuild/multiprotocol_rpmsg/prj.conf
markaj-nordic: This is what I suggested in TODO, but I guess we will need more integration bits for Matter samples to work with OTA and factory data, so this is just an initial effort to make samples working with sysbuild. But this is an important topic, @nordicjm is it possible to inherit domain/image configuration from NCS modules, like connectedhomeip as mentioned?

samples/matter/light_bulb/Kconfig.sysbuild
nordicjm: What is the question?

samples/matter/light_bulb/sysbuild/multiprotocol_rpmsg/prj.conf
nordicjm: > Would it be possible to inherit this configuration from the Kconfig.multiprotocol_rpmsg.defaults, similarly, as we have it for the child images?

If you mean the old child/parent way of doing things with inheritance, no, not supported, will not be supported

samples/matter/light_bulb/sysbuild/multiprotocol_rpmsg/prj.conf
nordicjm: Not supported, there is a plan for this but it is a long way off

markaj-nordic: Looks like there is a chance:
https://github.com/nrfconnect/sdk-nrf/tree/main/samples/zigbee/light_switch/sysbuild/multiprotocol_rpmsg
Again, @nordicjm  can you please comment on that? :)

nordicjm: Comment on what, what is the question? If you have an application and configure it to use prj_my_config.conf, it will apply to that image only, the rest of the images will use prj.conf as normal vs in child/parent image they would all use prj_my_config.conf

samples/matter/light_bulb/Kconfig.sysbuild
markaj-nordic: If the nRF70 boards are currently supported or if there is a plan to add support for them soon.

samples/matter/light_bulb/sysbuild/multiprotocol_rpmsg/prj.conf
markaj-nordic: So is there a different way to reuse the same multiprotocol configuration for selected samples, e.g. Matter samples? Or do we need to maintain copies of the same .conf file all over the place?

samples/matter/light_bulb/Kconfig.sysbuild
nordicjm: What does the nRF70 board need?

samples/matter/light_bulb/sysbuild/multiprotocol_rpmsg/prj.conf
nordicjm: Needs copies of the .conf

samples/matter/light_bulb/sysbuild/multiprotocol_rpmsg/prj.conf
markaj-nordic: Currently whenever we want a dedicated application configuration, for instance prj_release.conf, we need to provide the file with the exact same name in the `child_image/image_name` location even if prj_release.conf is the same as the default prj.conf for `image_name`. For example:
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/lock/prj_release.conf and
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/lock/child_image/multiprotocol_rpmsg/prj_release.conf

The question is: does it work the same in sysbuild (sysbuild/multiprotocol_rpmsg/*)?

nordicjm: > Not supported, there is a plan for this but it is a long way off

samples/matter/light_bulb/Kconfig.sysbuild
markaj-nordic: `nrf7002dk_nrf5340_cpuapp` and `nrf7002dk_nrf5340_cpunet` targets. We need them to support all required Nordic platforms in Matter. I mean, this is not needed right now because this is an initial port. I want to know if there is a plan to support these any soon.

nordicjm (approved):

samples/matter/light_bulb/Kconfig.sysbuild
nordicjm: This is gated on hwmv2 support upstream. If you want support for it before then, the board names can be added to the Kconfigs and cmake files in `sysbuild` and it will work like the nrf5340dk

markaj-nordic: Is it just a matter of extending https://github.com/nrfconnect/sdk-nrf/blob/d97238ff9328533a7590e934b7680db949c0ad71/sysbuild/Kconfig.netcore#L7 and other board related bits in this file?

nordicjm: that is part of it, the other part is the cmake parts in the same folder

markaj-nordic:
> One more comment, please align Matter weather station and Matter bridge applications to the changes introduced in samples.

@kkasperczyk-no applications do not support no-dfu configuration, so it's pointless to integrate them with sysbuild until we implement the proper OTA handling with sysbuild

kkasperczyk-no (approved):
Despite discussions in the comments, I still have many doubts about how should we handle this port. I'm not sure how the functionalities that we currently support (e.g. different build types) will be handled in the future. The question is if the current sysbuild support is mature enough, so we would be able to move everything we have to it. But yeah, these are only my thoughts. Let's hope more information will come with the next PRs.

PR #12619, opened 2023-10-10
manifest: Update upmerged respositories

de-nordic:
Upmerging:
  - sdk-mcuboot to 0c0470e294dcfb52aab92299356a5f3caa0aa52b
  - sdk-zephyr to 23cf38934c0f68861e403b22bc3dd0ce6efbfa39
  
test-sdk-nrf: test_upmerge_nov_23

PR #13124, opened 2023-11-17
bt: bas_client: Fix CONTAINER_OF invocation

trond-snekvik:
The bas_client used the k_delayed_work struct instead of the k_work struct to resolve the bt_bas_client with CONTAINER_OF. As k_work is the first member in k_delayed_work, this isn't an active bug, but zephyrproject-rtos/zephyr#61962 adds a static assert for this, which starts failing in the upmerge.

### DIFF
--- a/subsys/bluetooth/services/bas_client.c
+++ b/subsys/bluetooth/services/bas_client.c
@@ -181,7 +181,7 @@ static void bas_read_value_handler(struct k_work *work)
 	struct bt_bas_client *bas;
 
 	bas = CONTAINER_OF(work, struct bt_bas_client,
-			     periodic_read.read_work);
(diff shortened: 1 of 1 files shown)


nordicjm (commented):
Part of https://github.com/nrfconnect/sdk-nrf/pull/12619

joerchan (approved):

nordicjm (changes requested):
-1 to prevent merge, fix is OK but is already part of upmerge and would rather not have to fix a rebase

joerchan:
> -1 to prevent merge, fix is OK but is already part of upmerge and would rather not have to fix a rebase

Might as well close it then.

PR #13127, opened 2023-11-17
samples: matter: Align matter samples data model

doublemis1:
Based on Matter 1.1 align the samples data model values like e.g. cluster revision, feature map, mandatory commands/attributes.


kkasperczyk-no (approved):
Could you apply the same changes to the Matter bridge? I know it's not production ready, but maybe we could try to keep consistency between the samples.

doublemis1:
> Could you apply the same changes to the Matter bridge? I know it's not production ready, but maybe we could try to keep consistency between the samples.

Sure I will wait only for the tests to be sure that I don;t need to fix other things

PR #13130, opened 2023-11-17
samples: matter: Enable mandatory thermostat attributes in zap

doublemis1:
Enabled mandatory thermostat attributes: occupied cooling setpoint and occupied heating setpoint.


samples/matter/thermostat/src/thermostat.matter
```diff
@@ -1809,3 +1813,5 @@ endpoint 1 {
     handle command SetpointRaiseLower;
   }
 }
+
```
kkasperczyk-no, line 1816: I don't know why, but ZAP always adds two extra lines to this sample. You have to remove it manually from this file to fix the compliance check.

kkasperczyk-no (approved):

LuDuda (approved):

PR #13133, opened 2023-11-17
caf: power_manager: Postpone system off state enter

rakons:
This commit adds a delay between receiving the power off request and powering the CPU down.
This allows processing of wake up events that are in the event queue, generated by some modules in reaction on power down event.

JIRA: NCSDK-24656


subsys/caf/modules/power_manager.c
```diff
@@ -153,6 +151,18 @@ static void system_off_on_error(void)
 	sys_poweroff();
 }
 
+static void system_off_handler(struct k_work *work)
+{
+	LOG_WRN("System turned off");
+	LOG_PANIC();
```
gmarull, line 157: won't panic halt?

rakons: LOG_PANIC means only to process all logs.

gmarull (approved):

zycz (approved):
Ok, then it's fine.

gte1:
Could this be added to milestone 2.5.1 ?

PR #13135, opened 2023-11-17
samples: net: wifi: update twister tests

KajaKoren:
samples built for wifi is no longer using static credentials as default.
the static credential configs are now removed from twister tests.

### DIFF
--- a/samples/net/download/sample.yaml
+++ b/samples/net/download/sample.yaml
@@ -7,9 +7,12 @@ tests:
     integration_platforms:
       - nrf9160dk_nrf9160_ns
       - nrf9161dk_nrf9161_ns
-    platform_allow: nrf9160dk_nrf9160_ns nrf9161dk_nrf9161_ns
+    platform_allow:
+      - nrf9160dk_nrf9160_ns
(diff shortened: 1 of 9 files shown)


samples/net/download/sample.yaml
```diff
@@ -7,9 +7,12 @@ tests:
     integration_platforms:
       - nrf9160dk_nrf9160_ns
       - nrf9161dk_nrf9161_ns
-    platform_allow: nrf9160dk_nrf9160_ns nrf9161dk_nrf9161_ns
+    platform_allow:
```
gregersrygg, line 10: I think you forgot to remove `.lte` from the test name here.

gregersrygg (approved):
Nice cleanup! Just a small name change that I think slipped.

samples/net/download/sample.yaml
KajaKoren: removed the whole test, since we build with tfm as default.

bjda (approved):

PerMac (approved):

simensrostad (approved):

PR #13140, opened 2023-11-20
lib: location: Use NCELLMEAS=3,X

trantanen:
Change neighbor cell searches as follows:
- Do NCELLMEAS=1
- If more than one cell is requested, do NCELLMEAS=3,15
- If not enough GCI cells are found (by default 3), do NCELLMEAS=4,X, where X=3 by default

Even more significant change compared to the algorithm is that the cell count is not taking normal neighbors into account but only the current cell and GCI cells.

Jira: NCSDK-24585

### DIFF
--- a/doc/nrf/libraries/modem/location.rst
+++ b/doc/nrf/libraries/modem/location.rst
@@ -47,8 +47,9 @@ The supported location methods are as follows:
 * Cellular positioning
 
   * Uses :ref:`lte_lc_readme` for getting a list of nearby cellular base stations.
-  * Neighbor cell measurement is performed with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_EXTENDED_COMPLETE` search type.
-    A GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_LIGHT` search type is performed if the previous search did not find enough cells.
+  * Neighbor cell measurement is performed with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_EXTENDED_LIGHT` search type.
+    If more than one cell is requested, a GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_DEFAULT` search type is performed to find the cells based on the history information.
+    If a sufficient number of cells still has not been found, a GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_LIGHT` search type is performed.
     For more details on GCI search, see :c:member:`location_cellular_config.cell_count`.
   * The ``cloud location`` method handles sending cell information to the selected location service and getting the calculated location back to the device.
 
--- a/include/modem/location.h
+++ b/include/modem/location.h
@@ -371,16 +371,17 @@ struct location_cellular_config {
 	 *
 	 * Maximum value is 15.
 	 *
-	 * Zero indicates that only normal neighbor cell search is performed but no GCI search.
-	 *
-	 * If there are less than requested number of neighbor cells (including current cell),
-	 * GCI (surrounding) cells are requested also.
+	 * The number of cells mean GCI cells including current cell.
+	 * Normal neighbor cells are received and used but they do not count towards the cell count.
+	 * If the number of cells is bigger than one, GCI (surrounding) cells are requested.
+	 * Hence, zero and one mean that only normal neighbor cell search is performed
+	 * but no GCI search.
 	 *
 	 * Note that even if there are a lot of cells available, the number of cells
 	 * used for positioning may be lower than the requested number of cells due to
 	 * the behavior of the search algorithm. Also, the number of cells used for
 	 * positioning may be higher than the requested number of cells if there are
-	 * more neighbor cells (including current cell).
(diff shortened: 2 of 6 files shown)


trantanen:
I still need to double check the documentation part and some header file etc. comments that they are valid but functionally this is ready for review.

include/modem/location.h
```diff
@@ -365,22 +365,23 @@ struct location_cellular_config {
...
+	 * If the number of cells is bigger than one, GCI (surrounding) cells are requested.
+	 * Hence, zero and one mean that only normal neighbor cell search is performed
+	 * but no GCI search.
 	 *
 	 * Note that even if there are a lot of cells available, the number of cells
 	 * used for positioning may be lower than the requested number of cells due to
 	 * the behavior of the search algorithm. Also, the number of cells used for
 	 * positioning may be higher than the requested number of cells if there are
-	 * more neighbor cells (including current cell).
+	 * more cells inthe history information (including current cell).
```
peknis, line 384:
```suggestion
	 * more cells in the history information (including current cell).
```

doc/nrf/libraries/modem/location.rst
```diff
@@ -48,7 +48,8 @@ The supported location methods are as follows:
 
   * Uses :ref:`lte_lc_readme` for getting a list of nearby cellular base stations.
   * Neighbor cell measurement is performed with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_EXTENDED_COMPLETE` search type.
-    A GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_LIGHT` search type is performed if the previous search did not find enough cells.
+    If more than one cell is requested, a GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_DEFAULT` search type is performed to find the cells based on the history information.
+    If enough cells still have not been found, a GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_LIGHT` search type is performed.
```
peknis, line 52:
```suggestion
    If a sufficient number of cells still has not been found, a GCI search with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_GCI_EXTENDED_LIGHT` search type is performed.
```

peknis (changes requested):

doc/nrf/libraries/modem/location.rst
```diff
@@ -48,7 +48,8 @@ The supported location methods are as follows:
 
   * Uses :ref:`lte_lc_readme` for getting a list of nearby cellular base stations.
   * Neighbor cell measurement is performed with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_EXTENDED_COMPLETE` search type.
```
tokangas, line 50:
```suggestion
  * Neighbor cell measurement is performed with :c:enum:`LTE_LC_NEIGHBOR_SEARCH_TYPE_EXTENDED_LIGHT` search type.
```

lib/location/scan_cellular.c
```diff
@@ -148,31 +147,59 @@ void scan_cellular_execute(int32_t timeout, uint8_t cell_count)
...
+	err = lte_lc_neighbor_cell_measurement(&ncellmeas_params);
+	if (err) {
+		LOG_WRN("Failed to initiate GCI cell measurements: %d", err);
+		/* Clearing 'err' because previous neighbor search has succeeded
+		 * so those are still valid and positioning can proceed with that data
+		 */
+		err = 0;
+		goto end;
+	}
+	k_sem_take(&scan_cellular_sem_ncellmeas_evt, K_FOREVER);
```
tokangas, line 171:
```suggestion
	err = k_sem_take(&scan_cellular_sem_ncellmeas_evt, K_FOREVER);
```

tokangas (changes requested):

tests/lib/location/prj.conf
```diff
@@ -16,6 +16,9 @@ CONFIG_LOCATION_METHOD_CELLULAR=y
 CONFIG_LOCATION_SERVICE_HERE=y
 CONFIG_LOCATION_SERVICE_HERE_API_KEY="MyApiKey"
 
+# Increase AT monitor heap because %NCELLMEAS notifications can be large
+CONFIG_AT_MONITOR_HEAP_SIZE=512
```
jhirsi, line 20: I think that the requirements for this should be considered to be documented. Seems that also location lib sample is having same config for this and based on my knowledge from the past, 512 bytes ain't large after all, especially when now using NCELLMEAS=3,15 in cellular location algorithm. 
I recall that the theoretical max for NCELL meas response is ~4k bytes and that's why for I wrote that comment on mosh config:  https://github.com/nrfconnect/sdk-nrf/blob/main/samples/cellular/modem_shell/prj.conf#L87-L90

trantanen: Internal information shows that 4020 is theoretical maximum. In practice the maximum is 1654 because GCI cells never have neighbor cell information. And actually it's 1485 bytes because serving cell will neither have normal neighbors with GCI search.
I'll make a separate PR about the CONFIG_AT_MONITOR_HEAP_SIZE related configs in different apps/samples and asserting proper size in Location library.

jhirsi (approved):

PR #13146, opened 2023-11-20
Applications: nrf5340_audio: synchronize L and R with SDC.

henrla:
Use timestamps in SDUs to synchronize the left and right channels. The approach to synchronize the left and right channels are the following. If two channels send their first SDU simultaneously, it is without a timestamp as the controller cannot provide a timestamp value using read_iso_tx_sync until after the first SDU has been sent. The first two SDUs will be given a timestamp by the controller. This timestamp is based on time of arrival. Unfortunate timings can lead to the two timestamps belonging to different CIG reference points. In this situation, read_iso_tx_sync will return different tx timestamps for the two channels. The channel that is behind must use the timestamp of the channel that is in front in time to catch up and synchronize.

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
@@ -25,6 +25,7 @@
 #include "macros_common.h"
 #include "nrf5340_audio_common.h"
 #include "le_audio.h"
+#include "audio_sync_timer.h"
 
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(unicast_client, CONFIG_UNICAST_CLIENT_LOG_LEVEL);
@@ -778,7 +779,7 @@ static void stream_sent_cb(struct bt_bap_stream *stream)
 
 	ret = channel_index_get(stream->conn, &channel_index);
 	if (ret) {
-		LOG_ERR("Channel index not found");
+		LOG_WRN("Channel index not found");
 	} else {
 		atomic_dec(&headsets[channel_index].iso_tx_pool_alloc);
 	}
@@ -1153,7 +1154,7 @@ static void disconnected_headset_cleanup(uint8_t chan_idx)
 	headsets[chan_idx].num_source_eps = 0;
 }
 
-static int iso_stream_send(uint8_t const *const data, size_t size, struct le_audio_headset *headset)
+static int iso_stream_send(uint8_t const *const data, size_t size, struct le_audio_headset *headset, uint32_t ts)
 {
 	int ret;
 	struct net_buf *buf;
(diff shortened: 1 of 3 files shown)


applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -778,7 +779,7 @@ static void stream_sent_cb(struct bt_bap_stream *stream)
 
 	ret = channel_index_get(stream->conn, &channel_index);
 	if (ret) {
-		LOG_ERR("Channel index not found");
+		LOG_WRN("Channel index not found");
```
koffes, line 782: Why this change?

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -1306,11 +1308,17 @@ int unicast_client_stop(void)
 	return 0;
 }
 
+
+#define SDU_INTERVAL_US (10000u)
```
koffes, line 1312: I guess this is just for testing as we need to get the actual interval.

applications/nrf5340_audio/src/modules/audio_sync_timer.h
```diff
@@ -17,6 +17,13 @@
  */
 uint32_t audio_sync_timer_capture(void);
 
+/**
+ * @brief Add some documentation :)
+ * 
+ * @retval
+*/
+uint32_t audio_sync_frame_send_timer_capture(void);
```
koffes, line 25: Never used. Can we omit these two files?

koffes (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
henrla: Correct.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
henrla: For debugging purposes. Should not be included.

applications/nrf5340_audio/src/modules/audio_sync_timer.h
henrla: For testing purposes, I didn't have to use it. For the real thing, I think it might have to be used to make the solution more robust, due to how timestamped SDUs works on our controller. I think we should discuss the scenario where I think this is needed elsewhere.

koffes:
Resolved in https://github.com/nrfconnect/sdk-nrf/pull/13589

PR #13147, opened 2023-11-20
applications: serial_lte_modem: allow modem AT command interception

tomi-font:
For the interception to catch all AT commands, they all need to be sent through nrf_modem_at_cmd() because this is the only function covered by the modem library's callback feature.

This was already the case for AT commands received externally, but not for AT commands sent from the SLM code.

This implements alternative functions to nrf_modem_at_printf() and nrf_modem_at_scanf() as well as trap macros to prevent their use.

Also, because the interception is case-sensitive, all the AT commands sent are now all-uppercase.

### DIFF
--- a/applications/serial_lte_modem/src/slm_at_commands.c
+++ b/applications/serial_lte_modem/src/slm_at_commands.c
@@ -89,7 +89,7 @@ bool verify_datamode_control(uint16_t time_limit, uint16_t *time_limit_min);
 static bool is_modem_functional_mode(enum lte_lc_func_mode mode)
 {
 	int cfun;
-	int rc = nrf_modem_at_scanf("AT+CFUN?", "+CFUN: %d", &cfun);
+	int rc = slm_util_at_scanf("AT+CFUN?", "+CFUN: %d", &cfun);
 
 	return (rc == 1 && cfun == mode);
 }
@@ -104,7 +104,7 @@ static void modem_power_off(void)
 		 * This will return once the modem responds, which means it has actually
 		 * stopped. This has been observed to take between 1 and 2 seconds.
 		 */
-		nrf_modem_at_printf("AT+CFUN=0");
+		slm_util_at_printf("AT+CFUN=0");
 	}
 }
 
@@ -394,7 +394,7 @@ int handle_at_carrier(enum at_cmd_type cmd_type);
 #endif
 
 static struct slm_at_cmd {
-	char *string;
+	const char *string;
 	slm_at_handler_t handler;
 } slm_at_cmd_list[] = {
 	/* Generic commands */
@@ -519,24 +519,32 @@ int handle_at_clac(enum at_cmd_type cmd_type)
 	return ret;
 }
 
-int slm_at_parse(const char *at_cmd)
+int slm_at_parse(const char *cmd_str, size_t cmd_name_len)
 {
 	int ret = UNKNOWN_AT_COMMAND_RET;
 	int total = ARRAY_SIZE(slm_at_cmd_list);
 
 	for (int i = 0; i < total; i++) {
-		if (slm_util_cmd_casecmp(at_cmd, slm_at_cmd_list[i].string)) {
-			enum at_cmd_type type = at_parser_cmd_type_get(at_cmd);
+		const struct slm_at_cmd *const at_cmd = &slm_at_cmd_list[i];
 
-			at_params_list_clear(&slm_at_param_list);
-			ret = at_parser_params_from_str(at_cmd, NULL, &slm_at_param_list);
-			if (ret) {
-				LOG_ERR("Failed to parse AT command %d", ret);
-				return -EINVAL;
-			}
-			ret = slm_at_cmd_list[i].handler(type);
-			break;
+		/* For the match to happen the AT command names must be identical,
+		 * which requires both names to have the same characters and the same length.
+		 */
+		if (!(!strncmp(cmd_str, at_cmd->string, cmd_name_len)
+		      && at_cmd->string[cmd_name_len] == '\0')) {
+			continue;
 		}
+
+		const enum at_cmd_type type = at_parser_cmd_type_get(cmd_str);
+
+		at_params_list_clear(&slm_at_param_list);
+		ret = at_parser_params_from_str(cmd_str, NULL, &slm_at_param_list);
+		if (ret) {
+			LOG_ERR("Failed to parse AT command %d", ret);
+			return -EINVAL;
+		}
+		ret = at_cmd->handler(type);
+		break;
 	}
 
 	return ret;
--- a/applications/serial_lte_modem/src/slm_at_host.c
+++ b/applications/serial_lte_modem/src/slm_at_host.c
@@ -10,7 +10,6 @@
 #include <zephyr/drivers/uart.h>
 #include <zephyr/sys/ring_buffer.h>
 #include <string.h>
-#include "slm_util.h"
 #include "slm_at_host.h"
 #include "slm_at_fota.h"
 #include "slm_uart_handler.h"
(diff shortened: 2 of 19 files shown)


applications/serial_lte_modem/src/slm_trap_macros.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef SLM_TRAP_MACROS_
+#define SLM_TRAP_MACROS_
```
SeppoTakalo, line 8:
```suggestion
#ifndef SLM_TRAP_MACROS_H
#define SLM_TRAP_MACROS_H
```

It is more common to have full filename in uppercase.
https://en.wikipedia.org/wiki/Include_guard

SeppoTakalo (approved):
LGTM.

applications/serial_lte_modem/src/slm_util.c
```diff
@@ -7,65 +7,129 @@
...
+
+	if (ret == 0) {
+		ret = -NRF_EBADMSG;
+	}
+	return ret;
+}
+
+int slm_util_at_cmd_fwd_from_cb(char *buf, size_t len, char *at_cmd)
+{
+/* Purposefully bypass nrf_modem_at_cmd() to avoid infinite recursion. */
```
MarkusLassila, line 95: Comment meaning is bit obscure. I take it that this callback is to be called from nrf_modem_at_cmd and that is why we are redefining the nrf_modem_at_printf, to be able to print it without infinite recursion.

applications/serial_lte_modem/src/slm_util.c
```diff
@@ -7,65 +7,129 @@
...
+#undef nrf_modem_at_printf
+	const int ret = nrf_modem_at_printf("%s", at_cmd);
+#define nrf_modem_at_printf nrf_modem_at_scanf
```
MarkusLassila, line 98: I take that this again restores the trap macros, this is pretty obscure stuff and should be commented so that casual reader has a hope of understanding what is going on.

I am bit unsure if the trap macros are not more trouble than they are worth. They are currently defined in some generic .h files, which does not mean that they would always be included anyway and as we control the code base, it might be enough to document this and the reasons for it.

applications/serial_lte_modem/src/slm_at_host.c
```diff
@@ -427,19 +425,28 @@ static void format_final_result(char *buf, size_t buf_len, size_t buf_max_len)
...
+	char *at_cmd = buf;
 
 	LOG_HEXDUMP_DBG(buf, cmd_length, "RX");
 
 	/* UART can send additional characters when the device is powered on.
 	 * We ignore everything before the start of the AT-command.
 	 */
 	while (offset + 1 < cmd_length) {
-		if (toupper((int)buf[offset]) == 'A' && toupper((int)buf[offset + 1]) == 'T') {
+		if (toupper(buf[offset]) == 'A' && toupper(buf[offset + 1]) == 'T') {
```
MarkusLassila, line 449: Why remove the explicit conversion here when you keep it elsewhere?

MarkusLassila (changes requested):

applications/serial_lte_modem/src/slm_at_host.c
```diff
@@ -427,19 +425,28 @@ static void format_final_result(char *buf, size_t buf_len, size_t buf_max_len)
 	}
 }
 
-static void cmd_send(uint8_t *buf, uint16_t cmd_length, uint16_t buf_size)
+static void cmd_toupper(char *cmd)
+{
+	/* Set/test command names are delimited by '=', read by '?'. */
+	while (*cmd && *cmd != '=' && *cmd != '?') {
```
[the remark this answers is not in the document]
MarkusLassila: So I did.

applications/serial_lte_modem/src/slm_at_commands.c
```diff
@@ -519,24 +519,28 @@ int handle_at_clac(enum at_cmd_type cmd_type)
...
-			if (ret) {
-				LOG_ERR("Failed to parse AT command %d", ret);
-				return -EINVAL;
-			}
-			ret = slm_at_cmd_list[i].handler(type);
-			break;
+		const struct slm_at_cmd *const at_cmd = &slm_at_cmd_list[i];
+
+		if (strncmp(cmd_str, at_cmd->string, cmd_name_len)
+		 || at_cmd->string[cmd_name_len] != '\0') {
```
MarkusLassila, line 531: If we have a long at command coming and we compare to shorter one, doesn't the at_cmd->string[cmd_name_len] actually point over the allocated memory area?

MarkusLassila: Should you compare the strlens of the commands instead?

applications/serial_lte_modem/src/slm_at_host.c
```diff
@@ -427,19 +424,32 @@ static void format_final_result(char *buf, size_t buf_len, size_t buf_max_len)
 	}
 }
 
-static void cmd_send(uint8_t *buf, uint16_t cmd_length, uint16_t buf_size)
+static size_t cmd_name_toupper(char *cmd, size_t cmd_len)
+{
+	size_t i = 0;
+
+	/* Set/test command names are delimited by '=', read by '?'. */
+	while (i != cmd_len && cmd[i] != '=' && cmd[i] != '?') {
```
MarkusLassila, line 432: Would it be better to utilize for?

applications/serial_lte_modem/src/slm_util.c
MarkusLassila: Yeah, the documentation for this is tricky as well. I think that we can go for this approach for now and let's change in future, if it becomes too confusing.

lib/at_cmd_custom/at_cmd_custom.c
```diff
@@ -39,17 +40,16 @@ int at_cmd_custom_respond(char *buf, size_t buf_size,
 	size_t buf_size_required;
 
 	if (buf == NULL) {
-		LOG_ERR("%s called with NULL buffer", __func__);
```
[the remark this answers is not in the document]
MarkusLassila: I am not sure if you should change this, given that we are not using this and it has an affect on development, but I leave this up to you. Feel free to resolve.

MarkusLassila (changes requested):

applications/serial_lte_modem/src/slm_at_commands.c
MarkusLassila: Ah so it does, please have a comment for this. This is getting bit clever.

applications/serial_lte_modem/src/slm_at_host.c
MarkusLassila: Looks bit weird with while, so a slight preference on for. But you can also resolve this if you want.

applications/serial_lte_modem/src/slm_at_commands.c
MarkusLassila: Not sure which way I prefer it, but the comment clarifies it.

MarkusLassila (approved):

applications/serial_lte_modem/src/slm_util.c
```diff
@@ -7,65 +7,136 @@
...
+		return -EINVAL;
 	}
 
-	return true;
+	/* We are not interested in the AT response, so avoid having it copied.
+	 * However the modem library requires "OK\r\n" in the buffer to accept
+	 * a successful response if the command is intercepted so make room for it (5 bytes).
+	 */
+	ret = nrf_modem_at_cmd(buf, 5, "%s", buf);
+	if (ret == -NRF_E2BIG) {
```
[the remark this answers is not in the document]
tomi-font: Good point. Changed.

eivindj-nordic (approved):

PR #13151, opened 2023-11-20
lib: lte_link_control: Refactor network mode fallback

tokangas:
Refactored the network mode fallback implementation. Earlier, connecting started from the previously used network mode. Now connecting always starts with the preferred network mode (set using Kconfig options or lte_lc_system_mode_set()) and if connecting times out, fallback network mode is tried.

Fallback network mode is now determined from the current network mode (from Kconfig options or lte_lc_system_mode_set()) at runtime. If current network mode is LTE-M, fallback mode is NB-IoT, and vice versa.

Fixed a bug, where after a timeout or an error, the modem was left in online functional mode.


tokangas:
Added unit tests.

lib/lte_link_control/lte_lc.c
```diff
@@ -171,7 +157,7 @@ static void at_handler_cereg(const char *response)
 	enum lte_lc_lte_mode lte_mode;
 	struct lte_lc_psm_cfg psm_cfg = {0};
 
-	LOG_DBG("+CEREG notification: %s", response);
+	LOG_DBG("+CEREG notification: %.*s", strlen(response) - strlen("\r\n"), response);
```
jtguggedal, line 160: Nit: can use sizeof -1 for `\r\n` to save a few instructions

lemrey, line 160: nice

jtguggedal (approved):
LGTM

lemrey (approved):

lib/lte_link_control/lte_lc.c
tokangas: I tried changing the code to use `sizeof()` instead, but at least the flash size of the binary is exactly the same in both cases. I believe the compiler actually optimizes the `strlen()` call out with a constant string.

jtguggedal: Ah, then it's all good

lib/lte_link_control/lte_lc.c
```diff
@@ -598,64 +623,82 @@ static int connect_lte(bool blocking)
...
+			if (err) {
+				err = -ETIMEDOUT;
+				goto exit;
+			}
 
-		err = lte_lc_func_mode_set(LTE_LC_FUNC_MODE_NORMAL);
-		if (err || !blocking) {
-			goto exit;
-		}
+			LOG_INF("Using fallback system mode");
```
trantanen, line 679: If you include this info into log on line 590, maybe this could be removed.

tests/lib/lte_lc_api/src/lte_lc_api_test.c
```diff
@@ -283,6 +293,214 @@ void test_lte_lc_connect_async_null(void)
...
+	test_event_data[5].type = LTE_LC_EVT_CELL_UPDATE;
+	test_event_data[5].cell.id = LTE_LC_CELL_EUTRAN_ID_INVALID;
+	test_event_data[5].cell.tac = LTE_LC_CELL_TAC_INVALID;
+
+	test_event_data[6].type = LTE_LC_EVT_LTE_MODE_UPDATE;
+	test_event_data[6].lte_mode = LTE_LC_LTE_MODE_NONE;
+
+	/* Schedule +CEREG notification to be dispatched after timeout to trigger fallback */
+	strcpy(at_notif, "+CEREG: 5,\"003F\",\"0013BEEF\",9\r\n");
+	k_work_schedule(&at_notif_dispatch_work, K_SECONDS(CONFIG_LTE_NETWORK_TIMEOUT + 1));
```
trantanen, line 445: Maybe set CONFIG_LTE_NETWORK_TIMEOUT=1 in prj.conf. Unfortunately it cannot be made smaller.
But +1 could be made smaller and this could be like 1.3 or 1.5 seconds overall.

trantanen (changes requested):
See my previous comments

lib/lte_link_control/lte_lc.c
```diff
@@ -598,64 +623,82 @@ static int connect_lte(bool blocking)
...
+	}
 
-		/* Change the modem sys-mode only if it's not running or is meant to change */
-		if (!IS_ENABLED(CONFIG_LTE_NETWORK_MODE_DEFAULT) &&
-		    ((current_func_mode == LTE_LC_FUNC_MODE_POWER_OFF) ||
-		     (current_func_mode == LTE_LC_FUNC_MODE_OFFLINE))) {
-			err = lte_lc_system_mode_set(lte_lc_sys_mode, lte_lc_sys_mode_pref);
+	/* Reset the semaphore, it may have already been given
+	 * by an earlier +CEREG notification.
+	 */
```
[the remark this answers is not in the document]
tokangas: It barely does. 😄

lib/lte_link_control/lte_lc.c
```diff
@@ -566,11 +552,50 @@ static int enable_notifications(void)
...
+	case LTE_LC_SYSTEM_MODE_LTEM_GPS:
+		fallback_sys_mode = LTE_LC_SYSTEM_MODE_NBIOT_GPS;
+		break;
+
+	case LTE_LC_SYSTEM_MODE_NBIOT_GPS:
+		fallback_sys_mode = LTE_LC_SYSTEM_MODE_LTEM_GPS;
+		break;
+
+	default:
+		LOG_DBG("Fallback enabled, but not possible from current system mode");
```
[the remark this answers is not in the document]
tokangas: I made this a debug log, because this is not an error. Fallback is possible from some system modes, but not from all of them. I would keep the level as it is now. Not sure if there's need to log the current system mode, but I can add it if you want. This PR already changes `lte_lc_system_mode_set()` so, that system mode changes are logged.

lib/lte_link_control/lte_lc.c
```diff
@@ -566,11 +552,50 @@ static int enable_notifications(void)
...
+
+	err = nrf_modem_at_printf("AT%%XSYSTEMMODE=%s,%c",
+				  system_mode_params[fallback_sys_mode],
+				  system_mode_preference[lte_lc_sys_mode_pref]);
+	if (err) {
+		LOG_ERR("Failed to set system mode, error: %d", err);
+		return -EFAULT;
+	}
+
+	LOG_DBG("System mode set to %d, preference %d", fallback_sys_mode, lte_lc_sys_mode_pref);
```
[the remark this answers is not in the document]
tokangas: I was going to point to line 679, which provides the same information. I already considered doing the change myself, but couldn't figure out what to exactly say in the trace. :) I'll try to figure out something.

lib/lte_link_control/lte_lc.c
tokangas: Yes, I'll remove this.

tests/lib/lte_lc_api/src/lte_lc_api_test.c
tokangas: Set the timeout to 1 second and optimized the sleep times.

lib/lte_link_control/lte_lc.c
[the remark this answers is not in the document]
trantanen: Isn't this log something we'd like to be aware of even if debug logging is not turned ON? Developer might have misconfigured the application or used system modes in an undesired manner. Whether it would be LOG_WRN or maybe LOG_INF is another question. I'm not sure LOG_WRN is wrong or a spamming too much here.

Logging system mode adds value as this log is not normal and having the necessary information for the "problem" in the log is nice. And doesn't really cost anything.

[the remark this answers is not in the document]
tokangas: Currently, `CONFIG_LTE_NETWORK_USE_FALLBACK` is enabled by default, unless network mode is LTE-M + NB-IoT or LTE-M + NB-IoT + GNSS. Thus, it's enabled by default when `CONFIG_LTE_NETWORK_MODE_DEFAULT` is selected, but fallback can not be performed. If we make this LOG_WRN, I think we should change the Kconfig so, that fallback is not enabled with `CONFIG_LTE_NETWORK_MODE_DEFAULT` either.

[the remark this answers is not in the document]
tokangas: Updated as agreed offline.

PR #13153, opened 2023-11-20
samples: various: Add sysbuild configurations

nordicjm:
Adds sysbuild configurations to samples to check that they are building properly in CI


SebastianBoe:
Can we do something more sophisticated than duplicating all of our test configs to regression-test sysbuild?

I'm concerned about the maintenance cost.

How long will we support both sysbuild and NCS child images?

nordicjm:
> Can we do something more sophisticated than duplicating all of our test configs to regression-test sysbuild?

Like what?

> I'm concerned about the maintenance cost.
> 
> How long will we support both sysbuild and NCS child images?

3.6 will include sysbuild, child/parent image might be deprecated at that point, at which point the entries in the sample/testcase yaml files can be removed

SebastianBoe:
Like for instance if the

sysbuild: yes

entry

caused two tests to be performed. One with and one without sysbuild.

"3.6 will include sysbuild, child/parent image might be deprecated at that point,"

I suppose we can live with the duplication if we are unable to develop something better.

nordicjm:
> Like for instance if the
> 
> sysbuild: yes
> 
> entry caused two tests to be performed. One with and one without sysbuild.
> 
> "3.6 will include sysbuild, child/parent image might be deprecated at that point,"
> 
> I suppose we can live with the duplication if we are unable to develop something better.

At the moment, want both sets to run so that if there is a failure it can be pinpointed, this does have the unfortunate effect of increasing CI load/time. Then we should probably have a script that compares the output. Once we have 3.6 ready then maybe we will go back to having 1 set though, though all the CI tests will need to be updated to test with the sysbuild artifacts, I will begin communicating with those in charge of the tests about this soon.

nordicjm:
> I'm also concerned about CI response time.
> 
> Is the intention to double the CI twister runtime, or will only a selection of samples be enabled with sysbuild?

Adding sysbuild-equivalent samples/tests one by one and fixing/logging issues until they're all duplicated (with and without sysbuild). For example, the original run here showed we can't run native_posix or any qemu tests in ncs with sysbuild because it forces PM

shanthanordic:
@PerMac  can you take look for twister building  here ? Also can u also check if it take more time to build ?

shanthanordic (changes requested):
Giving a -1 till i  get a green from all test leads

PerMac:
> Gazell samples support nRF52 Series but not nRF5340. Are sysbuild tests beneficial for these single-core samples?

AFAIK sysbuild is not only for multicore devices, but in general, for any project that requires more than a single image. It can be multiple one-core devices in a setup, single multicore device, or even single device with single core but with multiple images, like bootloader and application. I think the last scenario is heavily resolved with child images in NCS and the idea is to replace it with sysbuid approach.

PerMac:
It seems it doesn't add that much to the CI. I compared the number of "tests" in latest main and on this PR and there is increase of ~200 new "tests" which is ~10% of the total. I don't see much difference in the execution time but it also wont be measurable unless we run it many more times and get proper statistics. It is safe to assume it adds ~10% load on top of 30min average on 10 nodes being used in parallel. 
How much is 10%? For waiting time, probably not much. For resources used, might be influential. 
I propose a "simple" solution. We can mark all the new sysbuild variants with "sysbuild" tag.  On-PR scope we will exclude all tests marked with this tag. However, they will be included in the nightly scope. With such approach a regression affecting sysbuild can still leak into the main, but will be caught the same or next day it enters. When we are done with the full transition child img ->sysbuild we could remove the no-longer-supported versions and not exclude sysbuild by default
Later on, when we progress with dynamically selected scope, we can figure out how and when to include sysbuild version in the scope.

stephen-nordic:
There is no UART output on crypto samples on _ns. Pointed this PR to the test PR that runs crypto tests with sysbuild.

magnev:
@nordicjm: Have you tried to run any samples on _ns targets?

nordicjm:
> @nordicjm: Have you tried to run any samples on _ns targets?

TF-M PSA template, yes, works

PerMac:
@jamie Will you add tags to all these sysbuild configurations? I could then easily run all of them on our on-target setup to verify the PR

nordicjm:
> @jamie Will you add tags to all these sysbuild configurations? I could then easily run all of them on our on-target setup to verify the PR

Has been added

samples/net/aws_iot/sample.yaml
```diff
@@ -18,3 +18,21 @@ tests:
...
       - CONFIG_WIFI_CREDENTIALS_STATIC_SSID="ssid"
       - CONFIG_WIFI_CREDENTIALS_STATIC_PASSWORD="psk"
+  sample.net.aws_iot.sysbuild:
+    sysbuild: true
+    tags: ci_build sysbuild
+    build_only: true
+    build_on_all: true
+    integration_platforms:
+      - nrf9160dk_nrf9160_ns
+      - nrf9160dk_nrf9160_ns
```
gregersrygg, line 28: I guess this one was supposed to be `nrf9161dk_nrf9161_ns` :)

gregersrygg:
I'm also concerned about Twister build/execution time in CI for PRs. The current total CI execution time including downstream jobs is already over 1 hour on average (excluding re-runs that skips tests).
If it's possible to do as @PerMac suggests with tags and skip sysbuild for PRs, that would be great!

nordicjm:
> I'm also concerned about Twister build/execution time in CI for PRs. The current total CI execution time including downstream jobs is already over 1 hour on average (excluding re-runs that skips tests). If it's possible to do as @PerMac suggests with tags and skip sysbuild for PRs, that would be great!

There are 2 parts to this, the main job that builds all the samples/test and ensures they build, then the downstream jobs that run the tests using the generated files. The first part should generate both sets of files and ensure neither fail to compile/build, the second part should ideally be switched over to exclusively use the sysbuild files (this PR is to find the issues with those), so yes, the main job will take longer but downstream jobs should take the same amount of time.

gregersrygg:
> There are 2 parts to this, the main job that builds all the samples/test and ensures they build, then the downstream jobs that run the tests using the generated files. The first part should generate both sets of files and ensure neither fail to compile/build, the second part should ideally be switched over to exclusively use the sysbuild files (this PR is to find the issues with those), so yes, the main job will take longer but downstream jobs should take the same amount of time.

Most of the tests have multiple boards to compile for. How much time will this PR increase build times? Not sure if these runs are representable or not, but they don't look promising:
~97 minutes for Twister: https://jenkins-ncs.nordicsemi.no/plugin/pipeline-timeline/index.html?build_url=https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/job/PR-13153/17/
~50 minutes for Twister: https://jenkins-ncs.nordicsemi.no/plugin/pipeline-timeline/index.html?build_url=https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/job/PR-13153/12/

@PerMac Is it possible to avoid building the tagged tests?

nordicjm:
It is going to vary depending upon CPU load, it just finished the current part in 1 hour 2 minutes, what is average for a normal PR without those changes? Looking at https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/view/change-requests/ it varies from 18 minutes to 3 hours

PerMac:
> ~97 minutes for Twister: https://jenkins-ncs.nordicsemi.no/plugin/pipeline-timeline/index.html?build_url=https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/job/PR-13153/17/ ~50 minutes for Twister: https://jenkins-ncs.nordicsemi.no/plugin/pipeline-timeline/index.html?build_url=https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/job/PR-13153/12/
> 
> @PerMac Is it possible to avoid building the tagged tests?

Not sure where the time for run # 17  come, twister failed there after 3 mins. During run # 12 9 out of 10 twister subsets finished in ~20min. The whole stage was lagged by a single subset, which took 50 mins. My guess is that there was some issue with build node. However, It is impossible to give any consistent value while the PR is still wip and varying builds are failing. The values Jamie mentioned are for the whole run, not only twister. We have to be sure we are comparing apples to apples. 
Yes, we will exclude configurations with sysbuild tags for on-PR runs. However, we won't do this before this PR is finalized and passing.

plskeggs (approved):

glarsennordic (approved):

alstrzebonski (approved):

juhaylinen (approved):

samples/bluetooth/mesh/light/sample.yaml
```diff
@@ -24,3 +39,12 @@ tests:
     platform_allow: nrf52840dk_nrf52840 nrf21540dk_nrf52840
     extra_args: OVERLAY_CONFIG=overlay-dfu.conf
     tags: bluetooth ci_build
+  sample.bluetooth.mesh.light.dfu.sysbuild:
```
ludvigsj, line 42: @alxelax could you take a look at this? I believe this build target will run into the known issue you reported

dchat-nordic (approved):
ipc_radio LGTM, just compliance for blank lines.

samples/bluetooth/mesh/light/sample.yaml
alxelax: Yep. Seems this is not sufficient. If to build sample with `--sysbuild` option and the overlay it doesn't add mcuboot to the final image.
Probably it might have been fixed but this is the known issue on rc1.

alxelax: sure

nordicjm: Most applications will fail to build until the hwmv2/sysbuild in ncs upmerge is merged

tony-le-24 (approved):

kkasperczyk-no:
@nordicjm sorry, I wanted to add one more reviewer and it somehow removed other ones. I can't fix it, because github says limit of reviewers is 15.

koffes (approved):
Only reviewed Audio. Looks good.

applications/machine_learning/sample.yaml
```diff
@@ -37,3 +66,13 @@ tests:
...
+  applications.machine_learning.zrelease.sysbuild:
+    sysbuild: true
+    build_only: true
+    platform_allow: nrf52840dk/nrf52840 thingy53/nrf5340/cpuapp thingy53/nrf5340/cpuapp/ns
+    integration_platforms:
+      - nrf52840dk/nrf52840
+      - thingy53/nrf5340/cpuapp
+      - thingy53/nrf5340/cpuapp/ns
+    tags: ci_build sysbuild
+    extra_args: CONF_FILE=prj_release.conf
```
maje-emb, line 78: Shouldn't we use `FILE_SUFFIX`?
```suggestion
    extra_args: FILE_SUFFIX="release" 
```
Also in other places.

SeppoTakalo (approved):

sysbuild/Kconfig.zip
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+menuconfig DFU_ZIP
```
Andrewpini, line 5: I've tried to run the mesh light sample with DFU with this option enabled in the sample sysbuild.conf file. But it seems no zip is generated at `light/build/dfu_application.zip` as expected. Later, I've also added `DFU_ZIP_BLUETOOTH_MESH_METADATA`. Then the zip file is generated, but it is missing the binary and manifest.

Is this going to be added later, or is it an oversight?

applications/machine_learning/sample.yaml
nordicjm: Fixed

applications/matter_bridge/sample.yaml
```diff
@@ -3,56 +3,70 @@ sample:
   name: Matter Bridge application
 tests:
   applications.matter_bridge.release:
+    sysbuild: true
```
kkasperczyk-no, line 6: Does it mean we are going to stop testing old solution at all? I think that we are not going to stop supporting it with this release, we should keep testing both solutions (at least some basic smoke tests). @doublemis1 fyi.

samples/matter/lock/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ set(multiprotocol_rpmsg_KCONFIG_ROOT "\\\${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/co
 set(hci_ipc_KCONFIG_ROOT "\\\${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/config/nrfconnect/chip-module/Kconfig.hci_ipc.root")
 
 # For prj.conf the CONF_FILE is empty. In other cases extract the exact file name from the path string.
-if(CONF_FILE)
-    get_filename_component(CONF_FILE_NAME ${CONF_FILE} NAME CACHE)
-endif()
-
-find_package(Zephyr HINTS $ENV{ZEPHYR_BASE})
+#if(CONF_FILE)
```
kkasperczyk-no, line 15: Do you want to remove it or it was commented out for some testing purposes?

applications/matter_bridge/sample.yaml
nordicjm: there are no child/parent image builds left in this but some do need to be added to ensure nothing breaks but this will only be a couple of projects (still tbd)

doublemis1: If the sysbuild includes all necessary features which will enabled building our solution in all configuration then I will move   directly to sysbuild, however create additional configuration per platform without sysbuild to validate previous (still supported) approach:
* Template sample for all supported major platform
* Matter Bridge App for supported platform
* Matter Weather Station App for supported platform

sysbuild/CMakeLists.txt
```diff
@@ -331,6 +336,10 @@ function(${SYSBUILD_CURRENT_MODULE_NAME}_post_cmake)
     include_mesh_dfu_metadata()
   endif()
 
+  if(SB_CONFIG_MATTER_FACTORY_DATA)
+    include_matter_factory_data()
```
ArekBalysNordic, line 377: I can see that factory data hex is generated in the build directory but is not merged with the firmware. It seems that config `MATTER_FACTORY_DATA_MERGE_WITH_FIRMWARE` is not used anywhere. I found a problem, I've added a comment on a PR to sdk-connectedhomeip.

nordicjm: Fixed

samples/matter/lock/sample.yaml
```diff
@@ -22,24 +25,32 @@ tests:
...
       - nrf21540dk/nrf52840
     platform_allow: >
       nrf21540dk/nrf52840
+    tags: sysbuild
   sample.matter.lock.thread_wifi_switched.smp_dfu:
+    sysbuild: true
     build_only: true
-    extra_args: SHIELD=nrf7002ek multiprotocol_rpmsg_SHIELD=nrf7002ek_coex
-      CONF_FILE=prj_thread_wifi_switched.conf CONFIG_NRF_WIFI_PATCHES_EXT_FLASH_STORE=y
+    extra_args: lock_SHIELD=nrf7002ek multiprotocol_rpmsg_SHIELD=nrf7002ek_coex
```
ArekBalysNordic, line 40: This configuration seems not to work. There is a problem with `MCUBOOT_UPDATEABLE_IMAGES` because I'm getting an error: `warning: default value 2 on MCUBOOT_UPDATEABLE_IMAGES ... clamped to 3 due to being outside the active range ([3, 3])`.
Does it mean that currently, sysbuild doesn't support 3 mcuboot images?

samples/matter/lock/Kconfig.sysbuild
```diff
@@ -10,20 +10,13 @@ config NRF_DEFAULT_MULTIPROTOCOL
...
-	default n
-
-config BOOTLOADER_NONE
-	default y
-
-endif
-
 if BOOTLOADER_MCUBOOT && (BOARD_NRF5340DK_NRF5340_CPUAPP || BOARD_NRF7002DK_NRF5340_CPUAPP)
 
 config MCUBOOT_UPDATEABLE_IMAGES
```
ArekBalysNordic, line 22:
```suggestion
config MCUBOOT_UPDATEABLE_IMAGES
	default 3 if NRF_WIFI_PATCHES_EXT_FLASH_STORE
```
Then the `sample.matter.lock.thread_wifi_switched.smp_dfu` target can be built.

samples/matter/template/sample.yaml
```diff
@@ -3,19 +3,32 @@ sample:
...
+    sysbuild: true
     build_only: true
     extra_args: CONFIG_NCS_SAMPLE_MATTER_OPERATIONAL_KEYS_MIGRATION_TO_ITS=y
     integration_platforms:
       - nrf52840dk/nrf52840
       - nrf5340dk/nrf5340/cpuapp
       - nrf7002dk/nrf5340/cpuapp
+    platform_allow: nrf52840dk/nrf52840 nrf5340dk/nrf5340/cpuapp nrf7002dk/nrf5340/cpuapp
+    tags: sysbuild
+  sample.matter.template.debug.nrf54l15:
```
doublemis1, line 15: @nordicjm 
Do you see any issue on your side why the different configuration is needed for nRF54L15 for debug configuration?
Or this is just leftover from old sample.yaml?

mkapala-nordic (commented):
Some initially found issues in case of nRF Desktop app.
CC: @MarekPieta

applications/nrf_desktop/Kconfig.sysbuild
```diff
@@ -0,0 +1,27 @@
...
+
+choice MCUBOOT_MODE
+	default MCUBOOT_MODE_SINGLE_APP if BOARD_NRF52833DK || BOARD_NRF52820DONGLE || BOARD_NRF52833DONGLE
+endchoice
+
+config SECURE_BOOT
+	default y if BOARD_NRF5340DK || BOARD_NRF52840DONGLE || BOARD_NRF52840DK
+
+config SECURE_BOOT_APPCORE
+	default y if BOARD_NRF5340DK || BOARD_NRF52840DONGLE || BOARD_NRF52840DK
```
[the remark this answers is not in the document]
nordicjm: > There are some missing targets, for example nrf52840gmouse has SECURE_BOOT enabled in prj.conf by default but with sysbuild it builds without it by default (west build --pristine --sysbuild -d build_gmouse -b nrf52840gmouse)

You need to use the config from sample.yaml to have the correct images included

> Moreover, most of the nRF Desktop configurations that has MCUboot enabled are configured in the direct-xip mode, why in sysbuild configurations it was changed to swap by default?

Has been fixed

> but I did not found the image for the secondary slot which in child/parent built automatically.
> How can I enable building the secondary slot image for MCUboot in direct-xip mode?

Had no support, has now been added

> Taking into the account the amount of different configurations in the nRF Desktop app, is it possible to have sysbuild configs per target and per file_suffix?

No, sysbuild does not support sysbuild.conf files with boards in the name

samples/matter/template/sample.yaml
nordicjm: They are needed for the different build options

samples/matter/lock/Kconfig.sysbuild
nordicjm: Issue was in sysbuild Kconfig, has been fixed

sysbuild/Kconfig.zip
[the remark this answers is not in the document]
nordicjm: Did not have support, has been added

applications/nrf_desktop/sample.yaml
```diff
@@ -102,40 +122,41 @@ tests:
...
     platform_allow: >
-      nrf52dmouse/nrf52832 nrf52kbd/nrf52832 nrf52810dmouse/nrf52810 nrf52820dongle/nrf52820
-      nrf52833dk/nrf52820 nrf52833dk/nrf52833 nrf52833dongle/nrf52833 nrf52840dk/nrf52840
+      nrf52dmouse/nrf52832 nrf52kbd/nrf52832 nrf52810dmouse/nrf52810
+      nrf52833dk/nrf52833 nrf52833dongle/nrf52833 nrf52840dk/nrf52840
       nrf52840dongle/nrf52840 nrf52840gmouse/nrf52840 nrf5340dk/nrf5340/cpuapp
       nrf54l15pdk/nrf54l15/cpuapp nrf54l15pdk@0.3.0/nrf54l15/cpuapp
     integration_platforms:
       - nrf52dmouse/nrf52832
       - nrf52kbd/nrf52832
```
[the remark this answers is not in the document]
nordicjm: Added

samples/matter/template/sample.yaml
doublemis1: However on main we have the same configuration for all supported platform
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/template/sample.yaml#L5-L15

Can we align it to the version as it is on main?

nordicjm: No, the options for this build differ, they are not done in the main applications prj.conf as it is with child/parent image

applications/nrf_desktop/Kconfig.sysbuild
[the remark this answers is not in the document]
MarekPieta: A complex application may actually need a way to provide multiple `sysbuild.conf` files. The content may depend on board and build type (file suffix). Avoiding passing multiple CLI parameters would greatly reduce risk of error for end-user who builds an application (the sysbuild configuration file would contain all of the needed sysbuild configuration).

sysbuild/Kconfig.zip
```diff
@@ -0,0 +1,44 @@
...
+	bool "Create DFU zip package"
+	depends on BOOTLOADER_MCUBOOT
```
mkapala-nordic, line 7: Why such dependency on MCUboot?
AFAIK the `dfu_application.zip` was also created when B0 was enabled (I think here: `nrf/subsys/bootloader/cmake/sign.cmake`) in child/parent.
Tested on nRF Desktop with `twister -v --build-only -T . -p nrf52kbd/nrf52832 --sub-test applications.nrf_desktop.zrelease.nrf52kbd` the zip file was not created.
Checked also in main using `west build --pristine -d build_kbd -b nrf52kbd -- -DCONF_FILE=prj_release.conf` and a dfu_application.zip was created.

applications/nrf_desktop/Kconfig.sysbuild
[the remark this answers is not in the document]
mkapala-nordic: >    > but I did not found the image for the secondary slot which in child/parent built automatically.
>    > How can I enable building the secondary slot image for MCUboot in direct-xip mode?
>
> Had no support, has now been added

signing of the secondary slot seems to fail:
```
Generating files from /home/mak4/ncs/nrf/applications/nrf_desktop/twister-out/nrf52840gmouse_nrf52840/applications.nrf_desktop.zdebug_fast_pair.gmouse/mcuboot_secondary_app/zephyr/zephyr.elf for board: nrf52840gmouse
Usage: imgtool sign [OPTIONS] INFILE OUTFILE
Try 'imgtool sign -h' for help.

Error: Invalid value for '-H' / '--header-size': Minimum value for -H/--header-size is 32
FATAL ERROR: command exited with status 2: /home/mak4/.local/bin/imgtool sign --version 0.0.0+0 --align 4 --header-size 0 --slot-size 430080 --overwrite-only --align 1 --key /home/mak4/ncs/bootloader/mcuboot/root-ec-p256.pem /home/mak4/ncs/nrf/applications/nrf_desktop/twister-out/nrf52840gmouse_nrf52840/applications.nrf_desktop.zdebug_fast_pair.gmouse/mcuboot_secondary_app/zephyr/zephyr.bin /home/mak4/ncs/nrf/applications/nrf_desktop/twister-out/nrf52840gmouse_nrf52840/applications.nrf_desktop.zdebug_fast_pair.gmouse/mcuboot_secondary_app/zephyr/zephyr.signed.bin
ninja: build stopped: subcommand failed.
```

applications/nrf_desktop/sample.yaml
```diff
@@ -23,40 +24,56 @@ tests:
...
+    sysbuild: true
     build_only: true
     platform_allow: nrf52840dk/nrf52840 nrf52840gmouse/nrf52840
     integration_platforms:
       - nrf52840dk/nrf52840
       - nrf52840gmouse/nrf52840
-    tags: bluetooth ci_build
-    extra_args: CONF_FILE=prj_mcuboot_smp.conf
+    tags: bluetooth ci_build sysbuild
+    extra_args: FILE_SUFFIX=mcuboot_smp
```
mkapala-nordic, line 71: I think `SB_CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=n` might be missing compared to the `prj_mcuboot_smp.conf`

applications/nrf_desktop/sample.yaml
```diff
@@ -23,40 +24,56 @@ tests:
...
+    sysbuild: true
     build_only: true
     platform_allow: nrf52840dk/nrf52840 nrf52840gmouse/nrf52840
     integration_platforms:
       - nrf52840dk/nrf52840
       - nrf52840gmouse/nrf52840
-    tags: bluetooth ci_build
-    extra_args: CONF_FILE=prj_fast_pair.conf
+    tags: bluetooth ci_build sysbuild
+    extra_args: FILE_SUFFIX=fast_pair
```
mkapala-nordic, line 44: for `nrf52840dk/nrf52840` target:
`SB_CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=n` might be missing compared to the `prj_fast_pair.conf`

mkapala-nordic: There is also an issue with disabling `SB_CONFIG_SECURE_BOOT_APPCORE` on `nrf52840dk/nrf52840`:
```
warning: SECURE_BOOT_APPCORE (defined at
/home/mak4/ncs/nrf/applications/nrf_desktop/Kconfig.sysbuild:22,
/home/mak4/ncs/nrf/sysbuild/Kconfig.secureboot:13) was assigned the value 'n' but got the value 'y'.
See http://docs.zephyrproject.org/latest/kconfig.html#CONFIG_SECURE_BOOT_APPCORE and/or look up
SECURE_BOOT_APPCORE in the menuconfig/guiconfig interface. The Application Development Primer,
Setting Configuration Values, and Kconfig - Tips and Best Practices sections of the manual might be
helpful too.
```
Seems to be present on all `nrf52840dk` configs which enables mcuboot.

sysbuild/Kconfig.zip
nordicjm: What's the point in having a zip if you can't do anything with it? The zip is for use with nrfutil or the mobile apps to load the files using MCUmgr. Without mcuboot you either have to design your own way of loading the updates or you have to flash an update using jlink

mkapala-nordic: Well, we can. In case of nRF Desktop application there is a tool `nrf/scripts/hid_configurator` which is used for the DFU also in case of B0.

CC: @MarekPieta

sysbuild/CMakeLists.txt
```diff
@@ -340,6 +349,11 @@ function(${SYSBUILD_CURRENT_MODULE_NAME}_post_cmake)
     get_property(PM_MCUBOOT_PRIMARY_SIZE TARGET partition_manager PROPERTY PM_MCUBOOT_PRIMARY_SIZE)
     get_property(PM_MCUBOOT_PAD_SIZE TARGET partition_manager PROPERTY PM_MCUBOOT_PAD_SIZE)
 
+    if(SB_CONFIG_MATTER_FACTORY_DATA)
+      get_property(PM_FACTORY_DATA_ADDRESS TARGET partition_manager PROPERTY PM_FACTORY_DATA_ADDRESS)
```
ArekBalysNordic, line 411: What about support for it if there is no Partition Manager support? Due to that our samples for nRF54L15 fail.
For this purpose, we need to read out the `factory_data` partition from a DTS file. We had it in the previous cmake file.

This part seems not to work: 
https://github.com/nrfconnect/sdk-connectedhomeip/pull/438/files#diff-02ef2c92f4124826f9ff5e08958a8644a90f37a4433043d80d5258c38f7341ddR168-R174

ArekBalysNordic (changes requested):

ArekBalysNordic:
> It seems like there is an issue with regeneration of `dfu_application.zip` when rebuilding a project without pristine build:
> 
> To reproduce:
> 
> 1. Added SB_CONFIG_DFU_ZIP=y and SB_CONFIG_DFU_ZIP_BLUETOOTH_MESH_METADATA=y to `sysbuild.conf` in Mesh distributor sample.
> 2. `nrf/samples/bluetooth/mesh/dfu/distributor$ west build -b nrf52840dk/nrf52840 -p --sysbuild`
> 3. `manifest.json` shows version_MCUBOOT = 1.0.0+0 in the `dfu_application.zip` which is correct
> 4. Change CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION="2.0.0+0" in prj.conf
> 5. `nrf/samples/bluetooth/mesh/dfu/distributor$ west build --sysbuild`
> 6. `manifest.json` **still** shows version_MCUBOOT = 1.0.0+0 in the `dfu_application.zip` which is incorrect

There is a similar issue with adding the WiFi patch to multi-image dfu files. There is a missing WiFi patch image entry when building:
`samples/matter/lock/sample.matter.lock.thread_wifi_switched.smp_dfu`

ArekBalysNordic (changes requested):
It will repair the WiFi external patch usage in Matter CI.

tony-le-24:
Found this issue with at_client build in our Cloud CI. This sample is crucial for our tests, thus need to be built with our cloud samples as well. Saw this in the error log, somehow the merged.hex for this build wasn't created:
`Generating files from /jenkins_workspace/workspace/nnect-nrf-iot_cloud_build_master/nrf/samples/cellular/at_client/build_at_nrf9161/at_client/zephyr/zephyr.elf for board: nrf9161dk`
`[7/10] No install step for 'at_client'`
`[8/10] Completed 'at_client'`
Full build log here: https://jenkins-ncs.nordicsemi.no/job/latest/job/test-fw-nrfconnect-nrf-iot_cloud_build/job/master/1681/artifact/test-fw-nrfconnect-nrf_iot/outcomes/logs/compiler_outputs/at_client/build_at_nrf9161/west_build_stdout_stderr.txt

ktaborowski (changes requested):
Sidewalk porting to sysbuild in not ready yet

- [x] nrf53 dfu
- [x] unit tests
- [x] ci prepatation for addition tests cases
- [x] documetnation updates
- [ ] ...

samples/matter/template/sample.yaml
```diff
@@ -3,20 +3,21 @@ sample:
...
 tests:
   sample.matter.template.debug:
+    sysbuild: true
     build_only: true
     extra_args: CONFIG_NCS_SAMPLE_MATTER_OPERATIONAL_KEYS_MIGRATION_TO_ITS=y
     integration_platforms:
       - nrf52840dk/nrf52840
       - nrf5340dk/nrf5340/cpuapp
       - nrf7002dk/nrf5340/cpuapp
-      - nrf54l15pdk/nrf54l15/cpuapp
```
doublemis1, line 12: @nordicjm can you update the sample.yaml and add nrf54l platform to `sample.matter.template.debug`

nordicjm: Done

ktaborowski:
minor: threre is an invalid waringing when using parartioin manager:
```
          ---------------------------------------------------------------------
          --- WARNING: Using a bootloader without pm_static.yml.            ---
          --- There are cases where a deployed product can consist of       ---
          --- multiple images, and only a subset of these images can be     ---
          --- upgraded through a firmware update mechanism. In such cases,  ---
          --- the upgradable images must have partitions that are static    ---
          --- and are matching the partition map used by the bootloader     ---
          --- programmed onto the device.                                   ---
          ---------------------------------------------------------------------
          

Call Stack (most recent call first):
  /home/ktab/ncs/v2.6.0/zephyr/cmake/modules/kernel.cmake:248 (include)
  /home/ktab/ncs/v2.6.0/zephyr/cmake/modules/zephyr_default.cmake:141 (include)
  /home/ktab/ncs/v2.6.0/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:66 (include)
  /home/ktab/ncs/v2.6.0/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate)
  CMakeLists.txt:10 (find_package)


-- Configuring done
-- Generating done
-- Build files have been written to: /home/ktab/ncs/v2.6.0/sidewalk/samples/sid_end_device/build/sid_end_device
-- Found Python3: /home/ktab/ncs/toolchains/2be090971e/usr/local/bin/python3.9 (found version "3.9.18") found components: Interpreter 
-- Found partition manager static configuration : /home/ktab/ncs/v2.6.0/sidewalk/samples/sid_end_device/pm_static_nrf54l15pdk_nrf54l15_cpuapp.yml
```

nordicjm:
> minor: threre is an invalid waringing when using parartioin manager:
> 
> ```
>           ---------------------------------------------------------------------
>           --- WARNING: Using a bootloader without pm_static.yml.            ---
>           --- There are cases where a deployed product can consist of       ---
>           --- multiple images, and only a subset of these images can be     ---
>           --- upgraded through a firmware update mechanism. In such cases,  ---
>           --- the upgradable images must have partitions that are static    ---
>           --- and are matching the partition map used by the bootloader     ---
>           --- programmed onto the device.                                   ---
>           ---------------------------------------------------------------------
>           
> 
> Call Stack (most recent call first):
>   /home/ktab/ncs/v2.6.0/zephyr/cmake/modules/kernel.cmake:248 (include)
>   /home/ktab/ncs/v2.6.0/zephyr/cmake/modules/zephyr_default.cmake:141 (include)
>   /home/ktab/ncs/v2.6.0/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:66 (include)
>   /home/ktab/ncs/v2.6.0/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate)
>   CMakeLists.txt:10 (find_package)
> 
> 
> -- Configuring done
> -- Generating done
> -- Build files have been written to: /home/ktab/ncs/v2.6.0/sidewalk/samples/sid_end_device/build/sid_end_device
> -- Found Python3: /home/ktab/ncs/toolchains/2be090971e/usr/local/bin/python3.9 (found version "3.9.18") found components: Interpreter 
> -- Found partition manager static configuration : /home/ktab/ncs/v2.6.0/sidewalk/samples/sid_end_device/pm_static_nrf54l15pdk_nrf54l15_cpuapp.yml
> ```

Fixed

RobertGalatNordic:
I have an issue with twister and unit_testing board.
When I added `sysbuild: true` to testcase.yaml, I realized that the tests build for unit_testing target are failing, because there is no such board.

```
tests:
  unity.sidewalk.unit_tests.interrupts:
    sysbuild: true
    tags: Sidewalk
    type: unit
```

Currently we overcome this issue by explicitly disabling sysbuild ( `sysbuild: false` ) in testcase.yaml where unit_testing board is used.

I also tried to build this test manually with twister `build -b unit_testing --sysbuild `, and it failed the same way as twister - it failed to recognize the board. 

```
Loading Zephyr module(s) (Zephyr base (cached)): extensions sysbuild_extensions python west root zephyr_module boards shields hwm_v2 sysbuild_kconfig native_simulator_sb_extensions
-- Cache files will be written to: /home/vscode/.cache/zephyr
-- Found west (found suitable version "1.2.0", minimum required is "0.14.0")
No board named 'unit_testing' found.

Please choose one of the following boards:
...
```
I just wonder if this is expected

PerMac:
> I have an issue with twister and unit_testing board. When I added `sysbuild: true` to testcase.yaml, I realized that the tests build for unit_testing target are failing, because there is no such board.
> 
> ```
> tests:
>   unity.sidewalk.unit_tests.interrupts:
>     sysbuild: true
>     tags: Sidewalk
>     type: unit
> ```
> 
> Currently we overcome this issue by explicitly disabling sysbuild ( `sysbuild: false` ) in testcase.yaml where unit_testing board is used.
> 
> I also tried to build this test manually with twister `build -b unit_testing --sysbuild `, and it failed the same way as twister - it failed to recognize the board.
> 
> ```
> Loading Zephyr module(s) (Zephyr base (cached)): extensions sysbuild_extensions python west root zephyr_module boards shields hwm_v2 sysbuild_kconfig native_simulator_sb_extensions
> -- Cache files will be written to: /home/vscode/.cache/zephyr
> -- Found west (found suitable version "1.2.0", minimum required is "0.14.0")
> No board named 'unit_testing' found.
> 
> Please choose one of the following boards:
> ...
> ```
> 
> I just wonder if this is expected

AFAIK (@tejlmand) sysbuild is not supported on "unit_testing" board. "Currently we overcome this issue by explicitly disabling sysbuild ( `sysbuild: false` ) in testcase.yaml where unit_testing board is used." it means you had to turn it on in the first place. So why turning on and then off? I am missing something.

ktaborowski (approved):
sidewalk aligned with sysbuild changes. no need to change (if integration tests are green)

ArekBalysNordic (approved):

SeppoTakalo (approved):

magnev (approved):

applications/asset_tracker_v2/Kconfig.sysbuild
```diff
@@ -4,10 +4,18 @@
...
+
 config SECURE_BOOT
 	default y if BOARD_NRF9160DK_NRF9160_NS
 
 config SECURE_BOOT_APPCORE
 	default y if BOARD_NRF9160DK_NRF9160_NS
 
+config PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY
+	default y if BOARD_NRF9151DK_NRF9151_NS || BOARD_NRF9161DK_NRF9161_NS
+	default y if BOARD_NRF9160DK_NRF9160_NS && "${BOARD_REVISION}" = "0.14.0"
```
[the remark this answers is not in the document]
nordicjm: it means exact match, if it needs higher revisions then those need to be added manually

maciejbaczmanski (approved):

shanthanordic (approved):
Approved with accepted failures

PR #13164, opened 2023-11-21
bluetooth: add VS command

olivier-le-sage:
Allows the controller to establish connections through the initiator and PAwR advertiser at the same time.

### DIFF
--- a/subsys/bluetooth/controller/hci_internal.c
+++ b/subsys/bluetooth/controller/hci_internal.c
@@ -655,6 +655,10 @@ static void vs_supported_commands(sdc_hci_vs_supported_vs_commands_t *cmds)
 #if defined(CONFIG_BT_CONN)
 	cmds->get_next_conn_event_counter = 1;
 #endif
+
+#if defined(CONFIG_BT_CTLR_SDC_PAWR_ADV)
+	cmds->allow_parallel_connection_establishments = 1;
+#endif
 }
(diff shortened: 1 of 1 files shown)


rugeGerritsen (approved):

subsys/bluetooth/controller/hci_internal.c
```diff
@@ -1585,6 +1589,10 @@ static uint8_t vs_cmd_put(uint8_t const *const cmd, uint8_t *const raw_event_out
 		*param_length_out += sizeof(sdc_hci_cmd_vs_get_next_conn_event_counter_return_t);
 		return sdc_hci_cmd_vs_get_next_conn_event_counter((void *)cmd_params,
 							    (void *)event_out_params);
+#endif
+#if defined(CONFIG_BT_CTLR_SDC_PAWR_ADV)
```
hermabe, line 1593: Should there be an `&& defined(CONFIG_BT_CENTRAL)` here? There is no use for the command if the application only uses pawr and never runs an initiator.

olivier-le-sage: Are you thinking about the use-case where we would support periodic advertising with responses + central links, but not scanning?

I don't think it really matters, given the command is just one line

hermabe: I was thinking about supporting just pawr and no connections.

hermabe (approved):

PR #13186, opened 2023-11-22
net: nrf_provisioning: Add events to device mode callback

juhaylinen:
Update callback to send an event when provisioning state is changed.


jayteemo (approved):
this will need a change log entry.
does the nrf_provisioning rst doc need updated?

subsys/net/lib/nrf_provisioning/src/nrf_provisioning.c
```diff
@@ -574,8 +577,9 @@ int nrf_provisioning_req(void)
 				LOG_DBG("Saving the latest command id");
 				commit_latest_cmd_id();
 			}
-			dm.cb(dm.user_data);
+			dm.cb(NRF_PROVISIONING_EVENT_STOP, dm.user_data);
 		}
+		dm.cb(NRF_PROVISIONING_EVENT_DONE, dm.user_data);
```
SeppoTakalo, line 582: This looks like now that the `STOP` event is only coming if the connection was succesful.

I think the START-STOP should come in pairs.. They must be indicated regardless of the outcome.

The `DONE` should only be indicated if there was new data.

subsys/net/lib/nrf_provisioning/src/nrf_provisioning.c
```diff
@@ -560,21 +563,26 @@ int nrf_provisioning_req(void)
...
 			/* Provisioning finished */
 			if (IS_ENABLED(CONFIG_NRF_PROVISIONING_SAVE_CMD_ID)) {
 				LOG_DBG("Saving the latest command id");
 				commit_latest_cmd_id();
 			}
-			dm.cb(dm.user_data);
+			dm.cb(NRF_PROVISIONING_EVENT_DONE, dm.user_data);
+		} else {
+			dm.cb(NRF_PROVISIONING_EVENT_STOP, dm.user_data);
 		}
```
SeppoTakalo, line 587: This is still not what we agreed.. STOP must follow the START unconditionally. Just to indicate that network activities have stopped.

So it should be
```c
		dm.cb(NRF_PROVISIONING_EVENT_START, dm.user_data);
		if (IS_ENABLED(CONFIG_NRF_PROVISIONING_HTTP)) {
			ret = nrf_provisioning_http_req(&rest_ctx);
		} else {
			ret = nrf_provisioning_coap_req(&coap_ctx);
		}
		dm.cb(NRF_PROVISIONING_EVENT_STOP, dm.user_data);
```

Then `dm.cb(NRF_PROVISIONING_EVENT_DONE, dm.user_data);` should only come if there was provisioning.

SeppoTakalo (changes requested):

subsys/net/lib/nrf_provisioning/src/nrf_provisioning.c
juhaylinen: Oops, sorry my bad. Fixed the events.

SeppoTakalo (approved):

doc/nrf/libraries/networking/nrf_provisioning.rst
```diff
@@ -65,8 +65,8 @@ The provisioning client can be initialized in one of the following ways:
 
 The function uses the following arguments:
 
-*  A pointer to a callback function, which is called when the modem state changes.
-*  A pointer that must be called once provisioning is done.
+*  A pointer to a :c:struct:`nrf_provisioning_mm_change` which holds a callback function to be called when the modem state changes.
+*  A pointer to a :c:struct:`nrf_provisioning_dm_change` which holds a callback function to be called when the provisioning state changes.
```
peknis, line 69:
```suggestion
*  A pointer to a :c:struct:`nrf_provisioning_mm_change` structure that holds a callback function to be called when the modem state changes.
*  A pointer to a :c:struct:`nrf_provisioning_dm_change` structure that holds a callback function to be called when the provisioning state changes.
```

include/net/nrf_provisioning.h
```diff
@@ -49,23 +49,38 @@ struct nrf_provisioning_mm_change {
...
 /**
  * @struct nrf_provisioning_dm_change
- * @brief Holds the callback to be called once provisioning is done together with data set by the
- * callback provider.
+ * @brief Holds the callback to be called once provisioning state changes together with data
+ * set by the callback provider.
  *
  * @param cb        The callback function.
- * @param user_data App specific data to be fed to the callback once it's called.
+ * @param user_data Application specific data to be fed to the callback once it's called.
```
peknis, line 80:
```suggestion
 * @param user_data Application-specific data to be fed to the callback once it is called.
```

include/net/nrf_provisioning.h
```diff
@@ -49,23 +49,38 @@ struct nrf_provisioning_mm_change {
...
+	/** Provisioning complete. "Finished" command received from the provisioning service. */
+	NRF_PROVISIONING_EVENT_DONE
+};
+
+/**
+ * @typedef nrf_provisioning_event_cb_t
+ * @brief Called when provisioning state changes.
+ *
+ * @param event nrf_provisioning event code.
+ * @param user_data Application specific data.
```
peknis, line 70:
```suggestion
 * @param user_data Application-specific data.
```

juhaylinen: @peknis Thanks for the review. I have addressed the findings.

peknis (approved):

PR #13189, opened 2023-11-22
Applications: nrf5340_audio: Added timestamps when sending SDUs

koffes:
To ensure SDUs are sent in the same ISO interval
Co-authored-by henrla

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
@@ -1153,7 +1153,7 @@ static void disconnected_headset_cleanup(uint8_t chan_idx)
 	headsets[chan_idx].num_source_eps = 0;
 }
 
-static int iso_stream_send(uint8_t const *const data, size_t size, struct le_audio_headset *headset)
+static int iso_stream_send(uint8_t const *const data, size_t size, struct le_audio_headset *headset, uint32_t timestamp)
 {
 	int ret;
 	struct net_buf *buf;
@@ -1191,7 +1191,8 @@ static int iso_stream_send(uint8_t const *const data, size_t size, struct le_aud
 
 	ret = bt_bap_stream_send(&headset->sink_stream, buf,
 				 get_and_incr_seq_num(&headset->sink_stream),
-				 BT_ISO_TIMESTAMP_NONE);
+				 timestamp);
+				 //BT_ISO_TIMESTAMP_NONE);
 	if (ret < 0) {
 		LOG_WRN("Failed to send audio data to %s: %d", headset->ch_name, ret);
 		net_buf_unref(buf);
@@ -1306,9 +1307,11 @@ int unicast_client_stop(void)
 	return 0;
 }
 
+static bool channels_are_synched = false;
+
 int unicast_client_send(struct le_audio_encoded_audio enc_audio)
 {
-	int ret;
+	int ret = 1;
 	size_t data_size_pr_stream;
 	struct bt_iso_tx_info tx_info = {0};
 	struct sdu_ref_msg msg;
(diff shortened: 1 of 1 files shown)


applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -1306,12 +1317,58 @@ int unicast_client_stop(void)
...
+ * @param tx_info
+ * @param ts_state
+ * @return uint32_t
+ */
+uint32_t timestamp_select(struct bt_iso_tx_info const *const tx_info)
+{
+
+	if (!ts_valid(ts_state[AUDIO_CH_L]) && !ts_valid(ts_state[AUDIO_CH_R])) {
+		LOG_WRN("No valid timestamp found");
+		return 0;
```
henrla, line 1338: You could return `BT_ISO_TIMESTAMP_NONE` instead of zero here

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -1306,12 +1317,58 @@ int unicast_client_stop(void)
...
+	}
+
+	/* From this point, we know that we have valid ts_state for both channels.
+	 * We need to check for wrapping. If both values have wrapped, this is fine.
+	 * However, if only one has wrapped, we must select the smaller value.
+	 */
+
+	uint32_t diff = abs((int)(tx_info[AUDIO_CH_L].ts - tx_info[AUDIO_CH_R].ts));
+
+	if (diff > (UINT32_MAX / 2)) {
```
henrla, line 1356: To my understanding, they may not have wrapped if the difference is larger than `UINT32_MAX / 2 `but smaller than `UINT32_MAX`. Is there an underlying assumption that allows this if to check for wrapping? Or am I misunderstanding something?

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -1306,12 +1317,58 @@ int unicast_client_stop(void)
...
+uint32_t timestamp_select(struct bt_iso_tx_info const *const tx_info)
+{
+
+	if (!ts_valid(ts_state[AUDIO_CH_L]) && !ts_valid(ts_state[AUDIO_CH_R])) {
+		LOG_WRN("No valid timestamp found");
+		return 0;
+	}
+
+	if (ts_valid(ts_state[AUDIO_CH_L]) && !ts_valid(ts_state[AUDIO_CH_R])) {
+		return tx_info[AUDIO_CH_L].ts;
```
henrla, line 1342: at this point, it should be ok to do `ts_state[AUDIO_CH_R] = TS_VALID`.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -1306,12 +1317,58 @@ int unicast_client_stop(void)
...
+	}
+
+	if (!ts_valid(ts_state[AUDIO_CH_L]) && ts_valid(ts_state[AUDIO_CH_R])) {
+		return tx_info[AUDIO_CH_R].ts;
```
henrla, line 1346: at this point, it should be ok to do `ts_state[AUDIO_CH_L] = TS_VALID`.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -1306,12 +1317,58 @@ int unicast_client_stop(void)
...
+ * @brief
+ *
+ * @param tx_info
+ * @param ts_state
+ * @return uint32_t
+ */
+uint32_t timestamp_select(struct bt_iso_tx_info const *const tx_info)
+{
+
+	if (!ts_valid(ts_state[AUDIO_CH_L]) && !ts_valid(ts_state[AUDIO_CH_R])) {
```
henrla, line 1336: The two following `if` statements could be `else if` to show that the states we are checking excludes each other.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
koffes: I see your point, but that would imply that the timestamp we already have in tx_info[AUDIO_CH_R] is valid, which it may not be.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
koffes: See above.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
koffes: Discuss

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
henrla: I agree that it is not valid at this point in time, but in this state, I believe we can safely anticipate that it will be valid the next time we are here. However, I see that for the purposes of the state machine it more tidy and clear to not change state in here.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
koffes: We have talked, and have found an alternative way to handle this. Fix will be pushed today.

PR #13193, opened 2023-11-22
scripts: Add memory thresholds' list

katgiadla:
This commit is part of a solution addressing developers' need to be aware of increases of memory footprints of applications under their maintainership.
The list added in this commit stores information about configurations, whose memory footprints should be monitored in the CI. Additionally, this list contains thresholds passing of which will trigger an automated warning on PRs introducing such changes. This dictionary is used as an input by a script computing memory increases, called by the CI after the twister stage.

jira: NCSDK-23524

### DIFF
--- a/scripts/memory-threshold-list.yaml
+++ b/scripts/memory-threshold-list.yaml
@@ -1,0 +1,36 @@
+- scenarios:
+    - sample.matter.lock.release
+    - sample.matter.light_bulb.release
+  platforms:
+    - nrf52840dk_nrf52840
+  ram_size: 241664
+  rom_size: 778240
+
+- scenarios:
+    - sample.matter.lock.debug
+    - sample.matter.light_bulb.debug
+  platforms:
+    - nrf52840dk_nrf52840
+  ram_size: 241664
+  rom_size: 931840
+
+- scenarios:
+    - sample.matter.lock.release
+  platforms:
+    - nrf7002dk_nrf5340_cpuapp
+  ram_size: 503808
+  rom_size: 778240
+
+- scenarios:
+    - sample.matter.lock.debug
+  platforms:
+    - nrf7002dk_nrf5340_cpuapp
+  ram_size: 503808
+  rom_size: 942080
+
+- scenarios:
+    - applications.nrf_desktop.z.*
+  platforms:
+    - all
+  rom_related_usage: 512
+  ram_related_usage: 256


scripts/memory-threshold-list.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - nrf7002dk_nrf5340_cpuapp
+  ram_size: 942080
+  rom_size: 503808
+
+- scenarios:
+    - applications.nrf_desktop.z.*
+  platforms:
+    - all
+  rom_related_usage: 256
+  ram_related_usage: 1024
```
MarekPieta, line 36: Hi @katgiadla
Can we slightly update thresholds here?:
```suggestion
  rom_related_usage: 512
  ram_related_usage: 256
```

MarekPieta: btw. Can we also monitor sizes of child images (bootloaders or netcore image)?

MarekPieta (changes requested):

scripts/memory-threshold-list.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - nrf52840dk_nrf52840
+  ram_size: 778240
+  rom_size: 241664
+
+- scenarios:
+    - sample.matter.lock.debug
+    - sample.matter.light_bulb.debug
+  platforms:
+    - nrf52840dk_nrf52840
+  ram_size: 931840
```
MarekPieta, line 14: btw. Aren't ROM and RAM values actually swapped for all of the records? `nrf52840` has 256kB of RAM so it would never reach the threshold here

katgiadla: yes, my mistake. Thanks

scripts/memory-threshold-list.yaml
katgiadla: > Hi @katgiadla Can we slightly update thresholds here?:

yes, applied

MarekPieta (approved):

scripts/memory-threshold-list.yaml
PerMac: The monitoring is based on the value in twister reports. This value is calculated by a script which parses zephyr.elf file. It means it has knowledge only about the main app. Let's see how useful it will be in action. If people enjoy the feature and keep using it we could think about something more sophisticated which would require some extra scripting and changes to the workflows etc.

PR #13197, opened 2023-11-22
cmake: sysbuild: partition_manager: Fix compatibility

andy3469:
Make the partition manager compatible with other brands MCU while using sysbuild


andy3469:
I would like to make the partition manager more compatible as a whole but I don't have the time needed right now.
For example, when you use it with another brand MCU, you need to reconfigure the `CONFIG_PM_SRAM_BASE` and `CONFIG_PM_SRAM_SIZE` to the MCU specs. But right now those configs are not visible.

thedjnK:
> I would like to make the partition manager more compatible as a whole but I don't have the time needed right now. For example, when you use it with another brand MCU, you need to reconfigure the `CONFIG_PM_SRAM_BASE` and `CONFIG_PM_SRAM_SIZE` to the MCU specs. But right now those configs are not visible.

You do realise this is completely against the https://github.com/nrfconnect/sdk-nrf/blob/main/LICENSE (license) of this whole repository?

andy3469:
> > I would like to make the partition manager more compatible as a whole but I don't have the time needed right now. For example, when you use it with another brand MCU, you need to reconfigure the `CONFIG_PM_SRAM_BASE` and `CONFIG_PM_SRAM_SIZE` to the MCU specs. But right now those configs are not visible.
> 
> You do realise this is completely against the https://github.com/nrfconnect/sdk-nrf/blob/main/LICENSE (license) of this whole repository?

Ah yes sorry. I will close the PR and move to the main repository of Zephyr as I need to use Nordic and other brands MCUs.

nordicjm:
> > > I would like to make the partition manager more compatible as a whole but I don't have the time needed right now. For example, when you use it with another brand MCU, you need to reconfigure the `CONFIG_PM_SRAM_BASE` and `CONFIG_PM_SRAM_SIZE` to the MCU specs. But right now those configs are not visible.
> > 
> > 
> > You do realise this is completely against the https://github.com/nrfconnect/sdk-nrf/blob/main/LICENSE (license) of this whole repository?
> 
> Ah yes sorry. I will close the PR and move to the main repository of Zephyr as I need to use Nordic and other brands MCUs.

You cannot use sdk-nrf with other brand of MCU, you must use upstream zephyr without any of the code from here, that includes PM.

andy3469:
Is it possible to disable the PM which is inside the sdk-nrf ? I only need the sysbuild system already present inside the main Zephyr repository. Because even with the `SB_CONFIG_PARTITION_MANAGER=n`, it's still enabled.

nordicjm:
> Is it possible to disable the PM which is inside the sdk-nrf ? I only need the sysbuild system already present inside the main Zephyr repository. Because even with the `SB_CONFIG_PARTITION_MANAGER=n`, it's still enabled.

You need to use upstream zephyr i.e. direct from https://github.com/zephyrproject-rtos/zephyr/ you cannot use sdk-nrf or sdk-zephyr. That means if you are developing for nordic and non-nordic boards you need to either do all of your nordic development in sdk-nrf and non-nordic development in zephyr, or do all of your nordic and non-nordic development in zephyr

andy3469:
OK, will close the PR then

PR #13199, opened 2023-11-23
applications: Fixed setting parent ep id for the Matter bridge

kkasperczyk-no:
Currently in the Matter bridge the dynamic endpoints doesn't have parent id set, which makes them do not work with some commercial ecosystems. This change fixes it by setting parent endpoint it to 1 (Aggregator).

### DIFF
--- a/samples/matter/common/src/bridge/Kconfig
+++ b/samples/matter/common/src/bridge/Kconfig
@@ -16,6 +16,10 @@ config BRIDGE_MAX_BRIDGED_DEVICES_NUMBER_PER_PROVIDER
 	int "Maximum number of endpoints paired to the one non-Matter provider device"
 	default 2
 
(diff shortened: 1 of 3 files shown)


markaj-nordic (approved):
Good catch! I'd only suggest to define the parent endpoint id as a const variable somewhere to make to code more reader friendly (assuming it's not possible to retrieve it from existing objects).

kkasperczyk-no:
> Good catch! I'd only suggest to define the parent endpoint id as a const variable somewhere to make to code more reader friendly (assuming it's not possible to retrieve it from existing objects).

It's not actually my catch, I got it from our FAE :smile: Sure, I will try to wrap it somehow.

Damian-Nordic (approved):

LuDuda (approved):
Thanks @lats1980 and @kkasperczyk-no!

PR #13201, opened 2023-11-23
applications: MatterBridge: added support for mandatory clusters

markaj-nordic:
* implemented Identify cluster on all bridged devices
* aligned mandatory clusters' configuration
* removed redundant Descriptor cluster implementation

### DIFF
--- a/applications/matter_bridge/src/bridged_device_types/humidity_sensor.cpp
+++ b/applications/matter_bridge/src/bridged_device_types/humidity_sensor.cpp
@@ -10,6 +10,7 @@ namespace
 {
 DESCRIPTOR_CLUSTER_ATTRIBUTES(descriptorAttrs);
 BRIDGED_DEVICE_BASIC_INFORMATION_CLUSTER_ATTRIBUTES(bridgedDeviceBasicAttrs);
+IDENTIFY_CLUSTER_ATTRIBUTES(identifyAttrs);
 }; /* namespace */
 
 using namespace ::chip;
@@ -27,7 +28,8 @@ DECLARE_DYNAMIC_ATTRIBUTE(Clusters::RelativeHumidityMeasurement::Attributes::Mea
 DECLARE_DYNAMIC_CLUSTER_LIST_BEGIN(bridgedHumidityClusters)
 DECLARE_DYNAMIC_CLUSTER(Clusters::RelativeHumidityMeasurement::Id, humiSensorAttrs, nullptr, nullptr),
 	DECLARE_DYNAMIC_CLUSTER(Clusters::Descriptor::Id, descriptorAttrs, nullptr, nullptr),
-	DECLARE_DYNAMIC_CLUSTER(Clusters::BridgedDeviceBasicInformation::Id, bridgedDeviceBasicAttrs, nullptr,
+	DECLARE_DYNAMIC_CLUSTER(Clusters::BridgedDeviceBasicInformation::Id, bridgedDeviceBasicAttrs, nullptr, nullptr),
+	DECLARE_DYNAMIC_CLUSTER(Clusters::Identify::Id, identifyAttrs, sIdentifyIncomingCommands,
 				nullptr) DECLARE_DYNAMIC_CLUSTER_LIST_END;
 
 DECLARE_DYNAMIC_ENDPOINT(bridgedHumidityEndpoint, bridgedHumidityClusters);
@@ -56,7 +58,8 @@ CHIP_ERROR HumiditySensorDevice::HandleRead(ClusterId clusterId, AttributeId att
 	switch (clusterId) {
 	case Clusters::RelativeHumidityMeasurement::Id:
 		return HandleReadRelativeHumidityMeasurement(attributeId, buffer, maxReadLength);
-		break;
+	case Clusters::Identify::Id:
+		return HandleReadIdentify(attributeId, buffer, maxReadLength);
 	default:
 		return CHIP_ERROR_INVALID_ARGUMENT;
 	}
@@ -102,6 +105,8 @@ CHIP_ERROR HumiditySensorDevice::HandleAttributeChange(chip::ClusterId clusterId
 	switch (clusterId) {
 	case Clusters::BridgedDeviceBasicInformation::Id:
 		return HandleWriteDeviceBasicInformation(clusterId, attributeId, data, dataSize);
+	case Clusters::Identify::Id:
+		return HandleWriteIdentify(attributeId, data, dataSize);
 	case Clusters::RelativeHumidityMeasurement::Id: {
 		switch (attributeId) {
 		case Clusters::RelativeHumidityMeasurement::Attributes::MeasuredValue::Id: {
--- a/samples/matter/common/src/bridge/matter_bridged_device.cpp
+++ b/samples/matter/common/src/bridge/matter_bridged_device.cpp
@@ -5,12 +5,13 @@
  */
 
 #include "matter_bridged_device.h"
+#include <app-common/zap-generated/callback.h>
 #include <lib/support/ZclString.h>
 
 using namespace ::chip;
 using namespace ::chip::app;
 
-CHIP_ERROR MatterBridgedDevice::CopyAttribute(void *attribute, size_t attributeSize, void *buffer,
+CHIP_ERROR MatterBridgedDevice::CopyAttribute(const void *attribute, size_t attributeSize, void *buffer,
 					      uint16_t maxBufferSize)
 {
 	if (maxBufferSize < attributeSize) {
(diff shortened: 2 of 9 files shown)


samples/matter/common/src/bridge/matter_bridged_device.cpp
```diff
@@ -73,6 +74,61 @@ CHIP_ERROR MatterBridgedDevice::HandleReadDescriptor(AttributeId attributeId, ui
...
 	}
+	case Clusters::Descriptor::Attributes::DeviceTypeList::Id: {
+		return CopyAttribute(mDeviceTypeList, sizeof(mDeviceTypeList), buffer, maxReadLength);
+	}
+	case Clusters::Descriptor::Attributes::ServerList::Id: {
+		return CopyAttribute(mServerIdsList, sizeof(mServerIdsList), buffer, maxReadLength);
+	}
+	case Clusters::Descriptor::Attributes::ClientList::Id: {
+		// no clients enabled
+		return CopyAttribute(nullptr, 0, buffer, maxReadLength);
```
kkasperczyk-no, line 85: What do you think about creating an `mClientsList` field in the class and assigning nullptr to it? This would make the implementation more generic, as we do not support client and parts list, but user would be able to set this field value to something else without modifying the core code. The same applies to parts list.

samples/matter/common/src/bridge/matter_bridged_device.h
```diff
@@ -46,20 +76,27 @@ class MatterBridgedDevice {
...
 	};
-
+	using IdentifyType = chip::app::Clusters::Identify::IdentifyTypeEnum;
 	static constexpr uint8_t kDefaultDynamicEndpointVersion = 1;
 	static constexpr uint8_t kNodeLabelSize = 32;
 	static constexpr uint8_t kDescriptorAttributeArraySize = 254;
 
 	explicit MatterBridgedDevice(const char *nodeLabel)
+		: mIdentifyServer(mEndpointId, IdentifyStartDefaultCb, IdentifyStopDefaultCb,
+				  IdentifyType::kVisibleIndicator)
```
kkasperczyk-no, line 86: Could we create an mIdentifyType field in this class and assign `kVisibleInditcatior` to it? I know for our examples we are not going to use other type of indicator, but if customer would like to use something else, it will not be possible without core code modification with the current implementation.

markaj-nordic: I guess, we can just set it to kNone and let the developer to select the supported mode in the concrete device implementation. I chose kVisibleInditcatior, because I saw it was the solution suggested by the Matter spec somewhere. But then I'm not sure if the identify-server will process the identify command as expected, I will check that.

Anyway, `ConfigureIdentifyServer()` method can be implemented in the concrete device class to overwrite the default identify-server configuration.

kkasperczyk-no: Yeah, but it works the same way for all other attributes. The user could also bypass our bridged device layer and just set the cluster value directly. I just think it may not be obvious for the user why we skip representing one selected attribute in the class, while we represent the other ones.

kkasperczyk-no: > I chose kVisibleInditcatior, because I saw it was the solution suggested by the Matter spec somewhere.

I'm not against it, I just think it should not be hardcoded in the constructor, but selected as a default value for the class field.

samples/matter/common/src/bridge/matter_bridged_device.cpp
```diff
@@ -73,6 +74,61 @@ CHIP_ERROR MatterBridgedDevice::HandleReadDescriptor(AttributeId attributeId, ui
 		uint32_t featureMap = GetDescriptorFeatureMap();
 		return CopyAttribute(&featureMap, sizeof(featureMap), buffer, maxReadLength);
 	}
+	case Clusters::Descriptor::Attributes::DeviceTypeList::Id: {
+		return CopyAttribute(mDeviceTypeList, sizeof(mDeviceTypeList), buffer, maxReadLength);
```
Damian-Nordic, line 78: Have you tested this works correctly? Looking at the `ReadSingleClusterData`, it doesn't appear that lists are supported in the default implementation. Isn't the default implementation of the descriptor cluster supposed to work with bridged endpoints as well (in which case this method wouldn't be needed at all).

markaj-nordic: Well I tested and it worked but it doesn't mean it won't work without this patch :) Will check that.
Btw. @kkasperczyk-no maybe that would explain why the cluster revision of the Descriptor cluster set in the application is always ignored.

applications/matter_bridge/src/bridged_device_types/humidity_sensor.cpp
```diff
@@ -10,6 +10,7 @@ namespace
 {
 DESCRIPTOR_CLUSTER_ATTRIBUTES(descriptorAttrs);
 BRIDGED_DEVICE_BASIC_INFORMATION_CLUSTER_ATTRIBUTES(bridgedDeviceBasicAttrs);
+BRIDGED_DEVICE_IDENTIFY_CLUSTER_ATTRIBUTES(bridgedDeviceIdentifyAttrs);
```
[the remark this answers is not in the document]
markaj-nordic: Sure.

samples/matter/common/src/bridge/matter_bridged_device.h
markaj-nordic: > Yeah, but it works the same way for all other attributes. The user could also bypass our bridged device layer and just set the cluster value directly. I just think it may not be obvious for the user why we skip representing one selected attribute in the class, while we represent the other ones.

But which attributes are you referring to? The IdentifyType is represented already in the Identify instance object which is not the case for IdentifyTime, therefore we need to keep in the class to handle it as an external attribute (support reading and writing). I assume the user needs to use ConfigureIdentifyServer(Identify*) method which provides the direct access to the Identify instance to align Identify configuration. I get your point, but in the current implementation the IdentifyTime kept as a field would never change its value anyway.

samples/matter/common/src/bridge/matter_bridged_device.cpp
markaj-nordic: Ok, we can do it.

samples/matter/common/src/bridge/matter_bridged_device.cpp
markaj-nordic: Also, there is a bug here as I take the wrong size of the list, so this code can't work.

Damian-Nordic: You can check the implementation in `descriptor.cpp` but for me it should work for dynamic endpoints as well. If that's the case, you can probably drop all the code you have for the descriptor attribute reads.

Damian-Nordic: Descriptor cluster revision seems to be fixed there: `static constexpr uint16_t ClusterRevision = 2;`

samples/matter/common/src/bridge/matter_bridged_device.h
kkasperczyk-no: Ok, in my opinion we create sort of interface for managing the dynamic endpoints and their clusters. If something can be customized by the user, we provide a possibility to set it, without invoking directly the clusters code.

For example, the device node label is typically something that is assigned during object creation and will not be modified ever again. Despite that fact we do not hardcode it, but we let user set some custom value (in this particular case using constructor).

I would imagine we should do the similar thing in this scenario. The IdentifyType is something constant at device life time, but not something what cannot be changed by the customer (e.g. ClusterRevision). Of course, if the mIdentifyType object exists somewhere we do not have to duplicate it, but I would expect we should allow to set its value in the constructor (maybe not the best option imo) or provide a Set method that would call  `ConfigureIdentifyServer(Identify*)` underneath. I just feel that providing API to customize all attributes but one that has to be modified using cluster specific code is not quite right.

samples/matter/common/src/bridge/matter_bridged_device.cpp
markaj-nordic: Yep, the whole code which handles descriptor cluster in the dynamic endpoints is useless. It's handled automatically by the core.

samples/matter/common/src/bridge/matter_bridged_device.h
kkasperczyk-no: Discussed F2F

kkasperczyk-no (approved):

samples/matter/common/src/bridge/matter_bridged_device.cpp
```diff
@@ -62,17 +63,42 @@ CHIP_ERROR MatterBridgedDevice::HandleReadBridgedDeviceBasicInformation(Attribut
...
 
-CHIP_ERROR MatterBridgedDevice::HandleReadDescriptor(AttributeId attributeId, uint8_t *buffer, uint16_t maxReadLength)
+CHIP_ERROR MatterBridgedDevice::HandleWriteIdentify(chip::AttributeId attributeId, void *data, size_t dataSize)
 {
 	switch (attributeId) {
-	case Clusters::Descriptor::Attributes::ClusterRevision::Id: {
-		uint16_t clusterRevision = GetDescriptorClusterRevision();
+	case Clusters::Identify::Attributes::IdentifyTime::Id:
+		if (data && dataSize == sizeof(mIdentifyTime)) {
+			mIdentifyTime = (*reinterpret_cast<decltype(mIdentifyTime) *>(data));
```
Damian-Nordic, line 71: There's no guarantee `data` is aligned.  This should use `memcpy`, instead.

samples/matter/common/src/bridge/matter_bridged_device.cpp
```diff
@@ -62,17 +63,42 @@ CHIP_ERROR MatterBridgedDevice::HandleReadBridgedDeviceBasicInformation(Attribut
...
+			/* Externally stored attribute was updated, now we need to notify identify-server to
+			   leverage the Identify cluster implementation */
+			app::ConcreteAttributePath attributePath{ mEndpointId, Clusters::Identify::Id, attributeId };
+			MatterIdentifyClusterServerAttributeChangedCallback(attributePath);
```
Damian-Nordic, line 75: Actually, I'm not sure if the identify-server will be able to pick up the new value stored externally. I trust it will, but just giving you a heads up that it's worth checking as in the past attribute accessors that identify-server uses didn't work with external attributes.

Damian-Nordic (approved):
Feel free to merge, but take a look at the comments and fix it as a followup if they're valid.

samples/matter/common/src/bridge/matter_bridged_device.cpp
markaj-nordic: Yes, fortunately it works as expected, the identify-server eventually calls `emAfReadOrWriteAttribute()` and `emberAfAttributeWriteAccessCallback()`. The latter checks the attribute bitmask for `ATTRIBUTE_MASK_EXTERNAL_STORAGE` and calls `emberAfExternalAttributeWriteCallback`/`emberAfExternalAttributeWriteCallback` accordingly.

samples/matter/common/src/bridge/matter_bridged_device.cpp
markaj-nordic: Yes, thanks for the catch. I will fix it in the followup PR.

samples/matter/common/src/bridge/matter_bridged_device.cpp
Damian-Nordic: ah, got it. Thanks for explanation!

PR #13202, opened 2023-11-23
tfm: Fix wrong sysbuild hex file when MCUboot is disabled

nordicjm:
Properly fixes an issue whereby the wrong (default) zephyr hex file was used for TF-M images when using sysbuild, when it should instead use the merged hex file


frkv (approved):
LGTM

PR #13203, opened 2023-11-23
sysbuild: Disable PM by default for other CI boards

nordicjm:
Disables PM from being used on non-physical boards used in CI


aescolar (approved):

tejlmand (approved):
🙈

PR #13218, opened 2023-11-23
wifi: Backport MAC address handling fixes

krish2718:


samples/wifi/shutdown/src/main.c
```diff
@@ -217,11 +217,13 @@ int main(void)
...
 
-		/* Set a local MAC address with Nordic OUI */
-		ret = net_if_down(iface);
-		if (ret) {
-			LOG_ERR("Cannot bring down iface (%d)", ret);
-			return ret;
+		if (net_if_is_up(iface)) {
+			/* Set a local MAC address with Nordic OUI */
+			ret = net_if_down(iface);
+			if (ret) {
```
rado17, line 223: We cant check for ` if (ret && ret != -EALREADY)` to avoid the no-op case

rado17 (changes requested):

samples/wifi/shutdown/src/main.c
krish2718: Good point, but this is backported fix, so, don't want to diverge from main fix.

sachinthegreen (approved):

PR #13221, opened 2023-11-24
Applications: nrf5340_audio: synchronize L-R channels with SDC.

henrla:
Use timestamps in SDUs to to synchronize L-R channels using SDC. When only one channel is streaming, there is no need to synchronize so data on this channel will be sent without timestamp. At the time when the other channel is set in streaming state, the channels must be synchronized. This is achieved by first sending one SDU without a timestamp on one channel. Read_iso_tx_sync will now return this SDU's P->C SDU synchronization reference, which determines when the SDU will be sent. This P->C SDU synchronization reference can be used as a timestamp for the SDU on the other channel, ensuring that they both will have the same P->C SDU synchronization reference, and sent at the same time.

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
@@ -1153,7 +1153,7 @@ static void disconnected_headset_cleanup(uint8_t chan_idx)
 	headsets[chan_idx].num_source_eps = 0;
 }
 
-static int iso_stream_send(uint8_t const *const data, size_t size, struct le_audio_headset *headset)
+static int iso_stream_send(uint8_t const *const data, size_t size, struct le_audio_headset *headset, uint32_t timestamp)
 {
 	int ret;
 	struct net_buf *buf;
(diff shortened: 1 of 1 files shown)


applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -1325,22 +1328,86 @@ int unicast_client_send(struct le_audio_encoded_audio enc_audio)
...
+		channels_are_synched = true;
+
+	} else {
+
+		bool all_channels_in_stream_state = true;
+
+		if (ep_state_check(headsets[AUDIO_CH_L].sink_stream.ep, BT_BAP_EP_STATE_STREAMING))	{
+			
+			ret = iso_stream_send(enc_audio.data, data_size_pr_stream,
+						&headsets[AUDIO_CH_L], BT_ISO_TIMESTAMP_NONE);
```
koffes, line 1374: Can NONE be used here?

koffes: Don't we need a ts here which is the same as we send for the other channel?

henrla: If the channels were previously synced, the "time of arrival" based timestamp that the controller will assign these packets should be the same when they are sent at the "same" time to the controller.

koffes:
This work continues in #13241

PR #13222, opened 2023-11-24
tfm: Move modules/tfm together with modules/trusted-firmware-m

joerchan:
Move modules/tfm together with modules/trusted-firmware-m.
The separation between these folders are not really clear to me.
Having two places in the Kconfig menu where TF-M configurations appear are confusing to the user.


subsys/partition_manager/pm.yml.tfm
```diff
@@ -19,7 +19,7 @@ tfm_secure:
 tfm_nonsecure:
   span: [app]
 
-#ifdef CONFIG_TFM_STORAGE
+#ifdef CONFIG_PM_PARTITION_TFM_STORAGE
```
SebastianBoe, line 22: maybe grep for CONFIG_TFM_STORAGE and check if there are any SYSBUILD scripts using it, assuming we are changing it's semantics.

joerchan: I haven't seen anything at all regarding sysbuild and TF-M flash/ram layout.
Our out-of-tree support still rely only on PM support.
@nordicjm Any input here?

nordicjm: Sysbuild uses the existing cmake code for TF-M, you could try the PSA sample building using sysbuild (if it uses this) to check it is still working, or any other sample that does, and ensure it builds and runs properly and that the partitions.yml file has the same sizes and offsets as the child/parent image build

Vge0rge (commented):
LGTM

SebastianBoe (approved):
A much better organization

Vge0rge (approved):

PR #13224, opened 2023-11-24
applications: nrf5340_audio:Fixed changing broadcast source runtime i…

rick1082:
…ssue

* Delete PA sync before switch broadcaster
* Only trigger scan in PA sync lost if reason is not from local

OCT-2806

### DIFF
--- a/applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
+++ b/applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
@@ -93,6 +93,12 @@ static void button_msg_sub_thread(void)
 					LOG_WRN("Failed to stop broadcast sink: %d", ret);
 				}
 			} else if (strm_state == STATE_PAUSED) {
+				ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
+				if (ret) {
+					LOG_WRN("Failed to start scanning for broadcaster: %d",
+						ret);
+				}
+
 				ret = broadcast_sink_start();
 				if (ret) {
 					LOG_WRN("Failed to start broadcast sink: %d", ret);
@@ -123,6 +129,10 @@ static void button_msg_sub_thread(void)
 			ret = broadcast_sink_change_active_audio_stream();
 			if (ret) {
 				LOG_WRN("Failed to change active audio stream: %d", ret);
+			}
+			ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
+			if (ret) {
+				LOG_WRN("Failed to start scanning for broadcaster: %d", ret);
 			}
 
 			break;
@@ -242,11 +252,26 @@ static void le_audio_msg_sub_thread(void)
 			break;
 
 		case LE_AUDIO_EVT_SYNC_LOST:
-			LOG_INF("Sync lost");
+			LOG_INF("Sync lost, reason = %d", msg.sync_lost_reason);
 
 			ret = bt_mgmt_pa_sync_delete(msg.pa_sync);
 			if (ret) {
 				LOG_WRN("Failed to delete PA sync");
+			}
+
+			if (msg.sync_lost_reason != BT_HCI_ERR_LOCALHOST_TERM_CONN) {
+				if (IS_ENABLED(CONFIG_BT_OBSERVER)) {
+					ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST,
+								 NULL);
+					if (ret != -EALREADY) {
+						LOG_WRN("Scanning already started: %d", ret);
+					} else {
+						/* NOTE: The string below is used by the Nordic CI
+						 * system
+						 */
+						LOG_INF("Restarted scanning for broadcaster");
+					}
+				}
 			}
 
 			if (strm_state == STATE_STREAMING) {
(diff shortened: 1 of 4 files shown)


applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
```diff
@@ -191,11 +193,12 @@ static void stream_started_cb(struct bt_bap_stream *stream)
 
 static void stream_stopped_cb(struct bt_bap_stream *stream, uint8_t reason)
 {
+	stream_stopped_reason = reason;
 
 	switch (reason) {
 	case BT_HCI_ERR_LOCALHOST_TERM_CONN:
 		LOG_INF("Stream stopped by user");
-		le_audio_event_publish(LE_AUDIO_EVT_NOT_STREAMING);
+		le_audio_event_publish(LE_AUDIO_EVT_SYNC_LOST);
```
koffes, line 201: Why is LE_AUDIO_EVT_SYNC_LOST sent when the reason actually is BT_HCI_ERR_LOCALHOST_TERM_CONN ?

applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
```diff
@@ -93,6 +93,12 @@ static void button_msg_sub_thread(void)
 					LOG_WRN("Failed to stop broadcast sink: %d", ret);
 				}
 			} else if (strm_state == STATE_PAUSED) {
+				ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
```
koffes, line 96: Do we need to scan if the stream is just paused?

applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
```diff
@@ -93,6 +93,12 @@ static void button_msg_sub_thread(void)
...
+				if (ret) {
+					LOG_WRN("Failed to start scanning for broadcaster: %d",
```
koffes, line 98: LOG_ERR?

koffes: Should the error be handled?

applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
```diff
@@ -242,13 +252,28 @@ static void le_audio_msg_sub_thread(void)
...
 			if (ret) {
 				LOG_WRN("Failed to delete PA sync");
 			}
 
+			if (msg.sync_lost_reason != BT_HCI_ERR_LOCALHOST_TERM_CONN) {
+				if (IS_ENABLED(CONFIG_BT_OBSERVER)) {
+					ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST,
+								 NULL);
+					if (ret != -EALREADY) {
+						LOG_WRN("Failed to restart scanning: %d", ret);
```
koffes, line 267: Suggest: 
```suggestion
						LOG_WRN("Scanning already started: %d", ret);
```

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
```diff
@@ -53,6 +53,7 @@ static struct audio_codec_info audio_codec_info[CONFIG_BT_BAP_BROADCAST_SNK_STRE
 static uint32_t bis_index_bitfields[CONFIG_BT_BAP_BROADCAST_SNK_STREAM_COUNT];
 
 static struct bt_le_per_adv_sync *pa_sync_stored;
+static uint8_t stream_stopped_reason;
```
koffes, line 56: We need to be cautious here. Can we pass this info in line 201?

koffes (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
rick1082: The reason is that:
1. No matter if the reason is LOCALHOST_TERM_CONN or TIMEOUT, the sink needs to delete PA sync for making sure the sink has the ability to sync again
2. Sending LE_AUDIO_EVT_SYNC_LOST will trigger the "PA sync delete"
https://github.com/nrfconnect/sdk-nrf/blob/a29db7cfe09ab59edcf4e92517006a756c252ae7/applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c#L257

applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
rick1082: Yes, since if user push button when in "pause state", then the sink needs to start the stream again, so sink needs to start scan and get sync with source.

applications/nrf5340_audio/src/audio/streamctrl_broadcast_sink.c
rick1082: I will do more check to see if there's a racing condition for trigger the scan multiple times.
Too bad there's no way to check if there's a scan on going.

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
rick1082: We need to check if the reason is from TIMEOUT (lost signal from gateway) or LOCALHOST_TERM (sink drop the sync by user's input). So we need to pass this info in the beginning.

PR #13225, opened 2023-11-24
applications: nrf_desktop: Non-volatile memory instead of flash in docs

MarekPieta:
Change switches to using non-volatile memory term instead of flash in the documentation. Other non-volatile memory may be used for the update image slot. Change also applies a few additional doc improvements.

Jira: NCSDK-24518

### DIFF
--- a/applications/nrf_desktop/doc/dfu.rst
+++ b/applications/nrf_desktop/doc/dfu.rst
@@ -9,8 +9,8 @@ Device Firmware Upgrade module
 
 Use the DFU module to perform the following:
 
-* Obtain the update image from :ref:`nrf_desktop_config_channel` and store it in the appropriate flash memory partition.
-* Erase the flash memory partition in the background before storing the update image.
+* Obtain the update image from :ref:`nrf_desktop_config_channel` and store it in the appropriate non-volatile memory partition.
+* Erase the non-volatile memory partition in the background before storing the update image.
 
 Module events
 *************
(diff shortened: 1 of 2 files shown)


applications/nrf_desktop/doc/dfu.rst
kapi-no, line 1: Can you also cover the DFU lock module, dfu_mcumgr, settings_loader and general nRF Desktop documentation (main Readme)?

Also, remember to add one changelog entry for this change (e.g. "Changed the flash memory term to the non-volatile memory term to generalize the nRF Desktop documentation")

MarekPieta: Will be done under follow-up ticket: https://nordicsemi.atlassian.net/browse/NCSDK-24793

applications/nrf_desktop/doc/dfu.rst
```diff
@@ -103,8 +103,8 @@ The DFU module implementation is centered around the transmission and the storag
 * `Partition preparation`_ - How the module prepares for receiving an image.
 
 The firmware transfer operation can also be carried out either by the :ref:`nrf_desktop_ble_smp` or :ref:`nrf_desktop_dfu_mcumgr`.
-The application user must not perform more than one firmware upgrade at a time.
-The modification of the data by multiple application modules would result in a broken image that would be rejected by the bootloader.
+Make sure that all enabled modules used for the firmware transfer operation use :ref:`nrf_desktop_dfu_lock` to synchronize non-volatile memory access.
+If a modules does not use the :ref:`nrf_desktop_dfu_lock`, the simultaneous modification of the data by the module and another application module would result in a broken image that would be rejected by the bootloader.
```
peknis, line 107:
```suggestion
If a module does not use the :ref:`nrf_desktop_dfu_lock`, simultaneous modification of the data by the module and another application module results in a broken image that would be rejected by the bootloader.
```

applications/nrf_desktop/doc/dfu.rst
```diff
@@ -173,7 +174,7 @@ start
 
 data
    Perform the set operation on this option to pass the chunk of the update image that should be stored at the current offset.
-   The data is initially buffered in RAM, and then it is written to the flash partition after a fetch operation is performed on the ``sync`` option.
+   The data is initially buffered in RAM, and then it is written to the non-volatile memory partition after a fetch operation is performed on the ``sync`` option.
```
peknis, line 177:
```suggestion
   The data is initially buffered in RAM, and then written to the non-volatile memory partition after a fetch operation is performed on the ``sync`` option.
```

peknis (changes requested):

applications/nrf_desktop/doc/dfu.rst
MarekPieta: Changed to `may result` to make information more accurate

peknis (approved):

kapi-no (approved):

applications/nrf_desktop/doc/dfu.rst
```diff
@@ -212,35 +213,37 @@ sync
...
 .. _dfu_bootloader_var:
 
 module_variant
    Perform the fetch operation on this option to get the information about the bootloader variant that is being used on the device.
 
 
-Writing data to flash
-=====================
+Writing data to non-volatile memory
+===================================
```
divipillai, line 225: Shouldn't the dfu_lock.rst also be updated (line no: 24). We have mention of flash memory.

mkapala-nordic (approved):

applications/nrf_desktop/doc/dfu.rst
MarekPieta: It will be updated in a separate PR.

divipillai (approved):

PR #13226, opened 2023-11-24
nrf_security: Remove unused vendor max ECC bits

Vge0rge:
This code is not useful anymore for two reasons.
First of all the CONFIG_PSA_CRYPTO_DRIVER_*
symbols are now removed and so this will never
be executed.

The vendor max ECC bits are now configured
by the Oberon PSA core in the file:

core/include/psa/crypto_sizes.h

Depending on the PSA_WANT_* enabled curves.


SebastianBoe (approved):

vili-nordic (approved):

joerchan (changes requested):
Remove it from the template as well:

```
/* Max curve bits for PSA APIs */
#cmakedefine PSA_VENDOR_ECC_MAX_CURVE_BITS                      @PSA_VENDOR_ECC_MAX_CURVE_BITS@
```

Vge0rge:
@joerchan  done :)

joerchan (approved):

PR #13235, opened 2023-11-24
manifest: Update sdk-zephyr with nrfx 3.1 with fixes for NCS 2.5.1

awojasinski-nordicsemi:
Pulls in switch of the remote for hal_nordic to nrfconnect/sdk-hal_nordic which contains hal_nordic with nrfx 3.1.0 with MDK 8.58.0 + minor fixes requested for the release

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -128,7 +128,7 @@ manifest:
           compare-by-default: true
     - name: mcuboot
       repo-path: sdk-mcuboot
-      revision: v1.10.99-ncs1
+      revision: edfe1e1465dbc698bf9a195816247913490ab391
       path: bootloader/mcuboot
     - name: qcbor
       url: https://github.com/laurencelundblade/QCBOR.git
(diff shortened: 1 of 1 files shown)


nika-nordic (approved):

anangl (approved):

cvinayak:
@divipillai any doc updates required here?

awojasinski-nordicsemi:
> > @divipillai any doc updates required here?
> 
> Should this change be updated in the release notes of 2.5.1 release (if the fixes were requested by a customer)? If so, I will be creating a PR for release note next week, and a comment could be added to that PR for the release note entry.
> 
> Please disregard my early comment.

@divipillai Yes it should have an entry in the change log. I would suggest it should look like this:

"Changed nrfx QSPI driver behaviour. Now QSPI peripheral is not activated during the driver initialization but rather when the first operation is requested or when `nrfx_qspi_activate()` is called."

"Fixed issue when external memory requires a relatively long time for the operation to happen by adding a `timeout` member to `nrfx_qspi_config_t`. It allows for configurable in the runtime timeout value. If the timeout value is 0 old value is used instead (500 milliseconds)."

Feel free to modify it if you see such a need.

PR #13241, opened 2023-11-27
Applications: nrf5340_audio: synchronize L-R channels with SDC.

koffes:
Use timestamps in SDUs to to synchronize L-R channels using SDC. When only one channel is streaming, there is no need to synchronize so data on this channel will be sent without timestamp. At the time when the other channel is set in streaming state, the channels must be synchronized. This is achieved by first sending one SDU without a timestamp on one channel. Read_iso_tx_sync will now return this SDU's P->C SDU synchronization reference, which determines when the SDU will be sent. This P->C SDU synchronization reference can be used as a timestamp for the SDU on the other channel, ensuring that they both will have the same P->C SDU synchronization reference, and sent at the same time.


koffes:
Replaced by: https://github.com/nrfconnect/sdk-nrf/pull/13589

PR #13248, opened 2023-11-28
nfc: Introduce Zero Latency IRQ

dchat-nordic:
Introduce Zero Latency IRQ in NFC Platform.
The Zero Latency IRQ is an improvement on NFC
processing latency with cost of a single SWI interrupt.

### DIFF
--- a/subsys/nfc/lib/Kconfig
+++ b/subsys/nfc/lib/Kconfig
@@ -81,14 +81,16 @@ config NFC_THREAD_PRIORITY
 
 endif# NFC_OWN_THREAD
 
-config NFC_LOW_LATENCY_IRQ
-	bool "Use low latency IRQ for NFC"
-	default y
+config NFC_ZERO_LATENCY_IRQ
+	bool "Use zero latency IRQ for NFC"
+	default n
+	select ZERO_LATENCY_IRQS
+	select EXPERIMENTAL
 	help
-	  Using this option ensures fast NFC interrupt
+	  Using this option ensures zero latency NFC interrupt
 	  execution on the cost of using additional interrupt (SWI).
 
-if NFC_LOW_LATENCY_IRQ
+if NFC_ZERO_LATENCY_IRQ
 
 config NFC_SWI_NUMBER
 	int
--- a/subsys/nfc/lib/platform.c
+++ b/subsys/nfc/lib/platform.c
@@ -4,8 +4,6 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 #include <zephyr/kernel.h>
-#include <zephyr/drivers/clock_control.h>
-#include <zephyr/drivers/clock_control/nrf_clock_control.h>
 #include <nrfx_nfct.h>
 #include <nrfx_timer.h>
 #include <nfc_platform.h>
@@ -28,8 +26,11 @@ LOG_MODULE_REGISTER(nfc_platform, CONFIG_NFC_PLATFORM_LOG_LEVEL);
 
 #define DT_DRV_COMPAT nordic_nrf_clock
 
-static struct onoff_manager *hf_mgr;
-static struct onoff_client cli;
+#ifdef CONFIG_NFC_ZERO_LATENCY_IRQ
+#define NFCT_IRQ_FLAGS IRQ_ZERO_LATENCY
+#else
+#define NFCT_IRQ_FLAGS 0
+#endif /* CONFIG_NFC_ZERO_LATENCY_IRQ */
 
 ISR_DIRECT_DECLARE(nfc_isr_wrapper)
 {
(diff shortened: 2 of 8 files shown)


subsys/nfc/lib/platform_internal_thread.c
```diff
@@ -185,28 +213,35 @@ static void schedule_callback(void)
 #endif /* CONFIG_NFC_OWN_THREAD */
 }
 
-#ifdef CONFIG_NFC_LOW_LATENCY_IRQ
+#ifdef CONFIG_NFC_ZERO_LATENCY_IRQ
 ISR_DIRECT_DECLARE(nfc_swi_handler)
 {
+	atomic_val_t tmp;
+
+	tmp = atomic_set(&clk_mode, HFCLK_NONE);
```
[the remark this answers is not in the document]
dchat-nordic: I think that atomic makes sense in the following scenario:
-The ISR is being serviced in the same time that the NFC interrupt fires.
-When swi isr reads the clk_mode as start/stop it needs to clean the clk_mode to HFCLK_NONE. These are 2 separate operations
- In between these operations the NFC interrupt (higher priority) could fire and set the clk_mode to another value start/stop and now the 2nd operation - clear would destroy the set value.
- Using atomics this situation is avoided.

subsys/nfc/lib/platform_internal_thread.c
```diff
@@ -185,28 +213,35 @@ static void schedule_callback(void)
 #endif /* CONFIG_NFC_OWN_THREAD */
 }
 
-#ifdef CONFIG_NFC_LOW_LATENCY_IRQ
+#ifdef CONFIG_NFC_ZERO_LATENCY_IRQ
 ISR_DIRECT_DECLARE(nfc_swi_handler)
 {
+	atomic_val_t tmp;
```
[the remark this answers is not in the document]
dchat-nordic: I actually can't do it, because other client could have altered the clock status and if nfc repeated for example clock start request, the old request would be the same as current and clock would not be started.

peknis (changes requested):

subsys/nfc/lib/platform.c
```diff
@@ -43,21 +38,21 @@ ISR_DIRECT_DECLARE(nfc_isr_wrapper)
...
 	int err;
 
-	hf_mgr = z_nrf_clock_control_get_onoff(CLOCK_CONTROL_NRF_SUBSYS_HF);
-	__ASSERT_NO_MSG(hf_mgr);
+	err = nfc_platform_internal_hfclk_init();
+	__ASSERT_NO_MSG(err == 0);
 
+#ifdef CONFIG_NFC_ZERO_LATENCY_IRQ
+	IRQ_DIRECT_CONNECT(NFCT_IRQn, CONFIG_NFCT_IRQ_PRIORITY,
+			   nfc_isr_wrapper, IRQ_ZERO_LATENCY);
```
maje-emb, line 50:
```suggestion
	IRQ_DIRECT_CONNECT(NFCT_IRQn, CONFIG_NFCT_IRQ_PRIORITY,
			   nfc_isr_wrapper, NFCT_IRQ_FLAGS);
```
What do you think about using the NFCT_IRQ_FLAGS definition?
This way we avoid conditional compilation at this place.

subsys/nfc/lib/platform.c
```diff
@@ -43,21 +38,21 @@ ISR_DIRECT_DECLARE(nfc_isr_wrapper)
...
+#else
 	IRQ_DIRECT_CONNECT(NFCT_IRQn, CONFIG_NFCT_IRQ_PRIORITY,
 			   nfc_isr_wrapper, 0);
+#endif /* CONFIG_NFC_ZERO_LATENCY_IRQ */
```
maje-emb, line 54:
```suggestion
IRQ_DIRECT_CONNECT(NFCT_IRQn, CONFIG_NFCT_IRQ_PRIORITY,
			   nfc_isr_wrapper, NFCT_IRQ_FLAGS);
```
What do you think about using the NFCT_IRQ_FLAGS definition?
This way we avoid conditional compilation at this place.

subsys/nfc/lib/Kconfig
```diff
@@ -80,14 +80,15 @@ config NFC_THREAD_PRIORITY
 
 endif# NFC_OWN_THREAD
 
-config NFC_LOW_LATENCY_IRQ
-	bool "Use low latency IRQ for NFC"
+config NFC_ZERO_LATENCY_IRQ
```
KAGA164, line 83: Experimental?

subsys/nfc/lib/platform_internal_thread.c
```diff
@@ -251,9 +295,36 @@ void nfc_platform_cb_request(const void *ctx,
...
 	schedule_callback();
-#endif /* CONFIG_NFC_LOW_LATENCY_IRQ */
+#endif /* CONFIG_NFC_ZERO_LATENCY_IRQ */
+}
+
+void nfc_platform_event_handler(nrfx_nfct_evt_t const *event)
+{
+	switch (event->evt_id) {
+	case NRFX_NFCT_EVT_FIELD_DETECTED:
+#ifdef CONFIG_NFC_ZERO_LATENCY_IRQ
```
KAGA164, line 309: My general suggestion is to avoid conditional compilation by using `IS_ENABLED()`

KAGA164:
I also noticed error log:
E: Tried to read header: 4 bytes, read 0.
E: Reading nfc ring buffer failed, resetting ring buffer.

subsys/nfc/lib/platform_internal_thread.c
dchat-nordic: This can't be swapped here because of clk_mode variable which is conditional.

peknis (approved):

PR #13257, opened 2023-11-28
nrf_security: Fix CMAC_ALT requiring CBC cipher mode

joerchan:
Fix CMAC_ALT implementation requiring CBC cipher mode to be enabled.


frkv (approved):
LGTM

PR #13270, opened 2023-11-29
samples: matter: Added support for mcuboot and DFU with sysbuild

kkasperczyk-no:
Introduced configuration that allows to enable mcuboot and DFU (over BT SMP) for the Matter samples build with sysbuild. Additionally the support for sysbuild in Matter weather station and Matter bridge was added.


applications/matter_bridge/Kconfig.sysbuild
```diff
@@ -0,0 +1,47 @@
...
+
+choice BOOTLOADER
+	default BOOTLOADER_MCUBOOT
+endchoice
+
+if BOOTLOADER_MCUBOOT && ($(BOARD) = "nrf7002dk_nrf5340_cpuapp")
+
+config MCUBOOT_UPDATEABLE_IMAGES
+	default 2
+
```
nordicjm, line 18: The problem I have with things like this, is it's not so much setting a default for an application for e.g. a bootloader, it's setting up a tonne of configuration for an image, which I think would be best done in a .conf file that sysbuild uses - thoughts @tejlmand ? This will set the application up in a certain way if the user meets the criteria, but if they for example disable mcuboot in the main configuration command, then enable it in `menuconfig`, none of this configuration will be applied.

applications/matter_weather_station/sysbuild.cmake
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+if("${OVERLAY_CONFIG}" STREQUAL "overlay-factory_data.conf")
+    set(PM_STATIC_YML_FILE ${APPLICATION_CONFIG_DIR}/pm_static_factory_data.yml PARENT_SCOPE)
+endif()
```
nordicjm, line 9: This too, it would make more sense for this to be an application sysbuild Kconfig which, if set, will include the overlay and set the static PM file

samples/matter/light_bulb/Kconfig.sysbuild
```diff
@@ -4,10 +4,47 @@
...
 	default y if ($(BOARD) = "nrf5340dk_nrf5340_cpuapp")
 
 config NRF_DEFAULT_BLUETOOTH
 	default y if ($(BOARD) = "nrf7002dk_nrf5340_cpuapp")
+
+choice BOOTLOADER
+	default BOOTLOADER_MCUBOOT
+endchoice
+
+if BOOTLOADER_MCUBOOT && (($(BOARD) = "nrf5340dk_nrf5340_cpuapp")||($(BOARD) = "nrf7002dk_nrf5340_cpuapp"))
```
nordicjm, line 17: spaces before/after `||`

samples/matter/light_bulb/sysbuild.cmake
```diff
@@ -0,0 +1,14 @@
...
+#
+
+if(CONF_FILE)
+    get_filename_component(CONFIG_FILE_NAME ${CONF_FILE} NAME)
+endif()
+
+# Set the paritions configuration
+if(NOT "${CONFIG_FILE_NAME}" STREQUAL "prj_no_dfu.conf")
+    set(PM_STATIC_YML_FILE ${ZEPHYR_NRF_MODULE_DIR}/samples/matter/light_bulb/configuration/${BOARD}/pm_static_dfu.yml PARENT_SCOPE)
+endif()
```
nordicjm, line 14: this might work fine on linux, but what happens on windows if someone specifies `PRJ_NO_DFU.CONF` as the filename ?

applications/matter_bridge/Kconfig.sysbuild
kkasperczyk-no: Ok, actually it's not a problem for me to move it to the `sysbuild.conf`. I added it to Kconfig, as a defaults, because we used to do that in child-image pattern. The reason was that we had multiple `prj_x.confs`, so it was easier to set it in one place instead of copying multiple times across different files.

samples/matter/light_bulb/sysbuild.cmake
kkasperczyk-no: I think it will not work, but isn't it expected behavior? The same is going to happen if I call the main application file `PRJ.CONF` instead of `prj.conf`. With the current build system it will fail as well. I mean, I think that generally our build system is case sensitive regarding file names.

applications/matter_bridge/Kconfig.sysbuild
kkasperczyk-no: @nordicjm the only problem is that we use different SB configs for different boards, so we cannot apply single `sysbuild.conf` to the multiple targets.

markaj-nordic (approved):

applications/matter_bridge/Kconfig.sysbuild
nordicjm: `-DSB_CONF_FILE=with_dfu.conf` ?

samples/matter/light_bulb/sysbuild.cmake
nordicjm: Note **windows** in the above, windows is case insensitive, so `PRJ.CONF` and `prj.conf` both work and give you the same file.

nordicjm: Will ignore this for now, can be opened as an issue later if a customer raises it as a problem

applications/matter_bridge/Kconfig.sysbuild
kkasperczyk-no: This file is maybe not the perfect example, because Matter bridge is supported only for one board - nRF7002 DK. However the other examples are supported for nRF52840, nRF5340 and nRF7002. Following the `-SB_CONF_FILE` approach we would need to create separate conf files for them and instruct the user how to attach them depending on used board. Unless, is it possible to create in `boards` directory for sysbuild and add `nrf52840dk_nrf52840.conf`, `nrf5340dk_nrf5340_cpuapp.conf` and `nrf7002dk_nrf5340_cpuapp.conf` there, so it would be automatically merged with all others SB configs? We had it for child images, but I'm not sure if it's even possible with sysbuild

ArekBalysNordic (approved):

kkasperczyk-no:
We are going to merge this PR. There is no request changes mark and the PR is open for too long due to discussion in comments without getting responses in a timely manner. Keeping it open also blocks the further development on the sysbuild integration. If there are serious issues with the current change, please fill the bug ticket.

PR #13271, opened 2023-11-29
doc: PCD library: link fix

peknis:
Fixing a reported wrong link.

### DIFF
--- a/doc/nrf/libraries/dfu/pcd.rst
+++ b/doc/nrf/libraries/dfu/pcd.rst
@@ -28,8 +28,8 @@ This is done to avoid compromised code from being able to perform updates of the
 The network core uses the PCD library to look for instructions on where to find the updates.
 Once an update instruction is found, this library is used to transfer the firmware update image.
 
-On the application core, the PCD library is used by the :doc:`mcuboot:index-ncs` sample.
-On the network core, the PCD library is used by the :ref:`nc_bootloader` sample.
+On the application core, this library is used :ref:`MCUboot <ug_bootloader_adding_upgradable_mcuboot>`.
+On the network core, the library is used by the :ref:`nc_bootloader` sample.
 
 
 API documentation


doc/nrf/libraries/dfu/pcd.rst
```diff
@@ -28,8 +28,8 @@ This is done to avoid compromised code from being able to perform updates of the
 The network core uses the PCD library to look for instructions on where to find the updates.
 Once an update instruction is found, this library is used to transfer the firmware update image.
 
-On the application core, the PCD library is used by the :doc:`mcuboot:index-ncs` sample.
-On the network core, the PCD library is used by the :ref:`nc_bootloader` sample.
+On the application core, the this library is used when having :ref:`MCUboot as an immutable bootloader <ug_bootloader_adding_upgradable_mcuboot>`.
```
hieuv, line 31: @hakonfam Is PCD only used by MCUboot when it is the immutable bootloader? When NSIB is the immutable one, PCD is still used by MCUboot, right?

doc/nrf/libraries/dfu/pcd.rst
```diff
@@ -28,8 +28,8 @@ This is done to avoid compromised code from being able to perform updates of the
 The network core uses the PCD library to look for instructions on where to find the updates.
 Once an update instruction is found, this library is used to transfer the firmware update image.
 
-On the application core, the PCD library is used by the :doc:`mcuboot:index-ncs` sample.
-On the network core, the PCD library is used by the :ref:`nc_bootloader` sample.
```
hieuv, line 32: @peknis I wonder if we should call the Network Core Bootloader a "sample." It's located in samples, but it is meant to be used as-is, and is being used as-is.

doc/nrf/libraries/dfu/pcd.rst
```diff
@@ -28,8 +28,8 @@ This is done to avoid compromised code from being able to perform updates of the
...
+On the application core, the this library is used when having :ref:`MCUboot as an immutable bootloader <ug_bootloader_adding_upgradable_mcuboot>`.
+On the network core, the library is used by the :ref:`nc_bootloader` sample.
```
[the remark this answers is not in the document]
peknis: Sorry, cannot really answer this.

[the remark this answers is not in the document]
sigvartmh: It's provided as a sample in-case you want to modify it to add some special functionality. I don't know if you actually can override the location of it in the build system and have your own out of tree version of this.

doc/nrf/libraries/dfu/pcd.rst
hieuv: I understand that PCD is used by MCUboot and not NSIB. What I mean is that the doc is saying:
_[PCD] is used when having MCUboot as an immutable bootloader_

If I understand correctly, PCD is used whether MCUboot is immutable or not (i.e. whether NSIB is the immutable BL or not).

peknis:
We are a bit stuck with this. @sigvartmh and @hakonfam, please suggest better wording.

sigvartmh (approved):

doc/nrf/libraries/dfu/pcd.rst
sigvartmh: That's correct. PCD is used in MCUBoot in either case.

hieuv: Then I propose this instead:
<code>On the application core, this library is used by :ref:\`MCUboot <ug_bootloader_adding_upgradable_mcuboot>\`.</code>

doc/nrf/libraries/dfu/pcd.rst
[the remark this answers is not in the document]
hieuv: This bit has always been tricky. Using "sample" is OK, I guess.
I haven't seen b0n location being overridden yet, but I have seen it being done for hci_rpmsg location. It's most likely possible.

PR #13279, opened 2023-11-29
[backport-v2.5] manifest: Update zephyr revision

anangl:
Pull in a fix in the XIP enabling function in the nrf_qspi_nor driver so that XIP works without any other QSPI transfer performed before.


awojasinski-nordicsemi:
The change in sdk-zephyr will be integrated as part of #13294

PR #13285, opened 2023-11-29
samples: Add sample for PSA PBKDF2

joerchan:


mswarowsky (approved):

samples/crypto/pbkdf2/boards/nrf5340dk_nrf5340_cpuapp_ns.conf
```diff
@@ -0,0 +1,4 @@
+CONFIG_TFM_PROFILE_TYPE_NOT_SET=y
+
+# Using hardware crypto accelerator
+CONFIG_PSA_CRYPTO_DRIVER_CC3XX=y
```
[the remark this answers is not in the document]
joerchan: It's copy-paste from the HKDF sample.

It's skipped because TF-M makes sure of this, TF-M depend on key derivation from oberon.

samples/crypto/pbkdf2/README.rst
```diff
@@ -0,0 +1,52 @@
...
+1. Initialization of the Platform Security Architecture (PSA) API.
+
+#. Key derivation:
+
+   a. The input key is imported into the PSA crypto keystore.
+   #. The output key is derived.
+
+#. Cleanup:
+
+   a. The input and the output keys are removed from the PSA crypto keystore.
```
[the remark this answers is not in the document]
joerchan: I've updated to say input password, instead of input key, since we are deriving from a password in this sample.

Also, the sample is using output_bytes, instead of output_key, so there is no output key after the key derivation.
I'm removing any mention of output key in the key store.

The end results is the derived key, in the application, instead of the key store.
Usually you would want to keep the derived key in the key_store and only use it by reference, but in this case we are outputting the key to the application

samples/crypto/pbkdf2/boards/nrf9160dk_nrf9160.conf
```diff
@@ -0,0 +1,5 @@
+# Enable both oberon driver and hardware crypto accelerator
+# Key derivation is only supported in software driver but is using chained
+# driver to accelerate sub-operations.
```
frkv, line 3:
```suggestion
# driver which will use hardware accelerated crypto for low-level crypto operations, if enabled.
```

samples/crypto/pbkdf2/boards/nrf5340dk_nrf5340_cpuapp.conf
```diff
@@ -0,0 +1,5 @@
+# Enable both oberon driver and hardware crypto accelerator
+# Key derivation is only supported in software driver but is using chained
+# driver to accelerate sub-operations.
```
frkv, line 3: `# driver which will use hardware accelerated crypto for low-level crypto operations, if enabled.`

frkv, line 3:
```suggestion
# driver which will use hardware accelerated crypto for low-level operations, if enabled.
```

melwee01 (approved):

PR #13291, opened 2023-11-30
lib: location: NCELLMEAS=3,X to use MAX(5,cell_count)

trantanen:
Changing cellular positioning history search cell count from 15 to the maximum of 5 and requested cell count. While we can get more than requested number of cells fairly cheaply in terms of power and time, the measurements done with history search still require some power, and adding more and more cells doesn't provide clear benefits.

Jira: NCSDK-24585

### DIFF
--- a/lib/location/scan_cellular.c
+++ b/lib/location/scan_cellular.c
@@ -114,6 +114,7 @@ void scan_cellular_execute(int32_t timeout, uint8_t cell_count)
 		.gci_count = 0
 	};
 	int err;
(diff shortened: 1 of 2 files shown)


tokangas (approved):
I don't have strong opinion whether to use `cell_count` or some other value here. As discussed offline, we could use a value like 6 or 8 here, but it might not result in any better cell positioning results, while still consuming some energy.

jhirsi (approved):

PR #13311, opened 2023-12-01
net: lib: mqtt_helper: Add password support

jtguggedal:
Add support for setting MQTT password.


gregersrygg (approved):

jtguggedal:
> Tests?

Yeah, but it's more involved than the current mqtt_helper tests as we would need to check the parameters passed to `mqtt_connect`. I figured it could go in a separate PR where we update all the test cases to do that where it's relevant. CIA-1122 created for tracking

rlubos (approved):

divipillai (approved):

PR #13313, opened 2023-12-01
SDC Kconfig fixes for simulated nrf5340

aescolar:
Minor fixes so the Softdevice is configured in the same way for the simulated nrf5340 as for the real one.
SOC_COMPATIBLE_NRF5<x> is set both by the corresponding simulated and real targets.

### DIFF
--- a/Kconfig.nrf
+++ b/Kconfig.nrf
@@ -80,7 +80,7 @@ config MCUMGR_TRANSPORT_NETBUF_SIZE
 
 # When using HCI on the nRF5340 we need a larger command buffer.
 config BT_BUF_CMD_TX_COUNT
-	default 10 if SOC_NRF5340_CPUAPP || SOC_NRF5340_CPUNET
+	default 10 if SOC_COMPATIBLE_NRF5340_CPUAPP || SOC_COMPATIBLE_NRF5340_CPUNET
 
 # Set ENTROPY_GENERATOR to true for TF-M builds with enabled
 # RNG provided from secure services.


subsys/bluetooth/controller/Kconfig
```diff
@@ -38,7 +38,7 @@ config BT_LL_SOFTDEVICE
 	select BT_HAS_HCI_VS
 	select BT_CTLR_DF_SUPPORT if HAS_HW_NRF_RADIO_DFE
 	select BT_CTLR_LE_POWER_CONTROL_SUPPORT
-	depends on (SOC_SERIES_BSIM_NRFXX || SOC_SERIES_NRF52X || SOC_NRF5340_CPUNET)
+	depends on (SOC_COMPATIBLE_NRF52X || SOC_COMPATIBLE_NRF5340_CPUNET)
```
rugeGerritsen, line 41: @jori-nordic , do we still need this, or should we depend on that the dragoon west group is enabled?

jori-nordic: will be determined today. It depends if we want to keep building with the prebuilt lib or only allow from source.

jori-nordic: we don't need it anymore, I'll post a PR to remove it

aescolar:
@martintv  @rugeGerritsen @jori-nordic May I ask for a +1? (the CI reported failure seems bogus(?), as in Jenkins it seems to appear as passed)

jori-nordic (approved):

rugeGerritsen (approved):

PR #13316, opened 2023-12-01
lib: location: Wait RRC idle before GCI search

trantanen:
Since GCI search won't return GCI cells but only normal neighbors in RRC connected mode, we will wait for RRC idle mode for 10s and proceed to GCI searches when RRC idle mode is reached. If RRC is still connected after 10s, GCI searches are skipped and we use results from NCELLMEAS=1 for cellular positioning.

Jira: NCSDK-24814

### DIFF
--- a/lib/location/scan_cellular.c
+++ b/lib/location/scan_cellular.c
@@ -22,6 +22,13 @@ BUILD_ASSERT(CONFIG_AT_MONITOR_HEAP_SIZE >= 1024,
 	    "and other notifications at the same time");
 #endif
 
+/**
+ * Waiting time (in seconds) of RRC connection going into idle mode. Because GCI search
+ * cannot scan GCI cells during RRC connected mode, we will wait for the RRC idle mode before
+ * initiating GCI search.
+ */
+#define SCAN_CELLULAR_RRC_IDLE_WAIT_TIME 10
+
 static struct lte_lc_ncell neighbor_cells[CONFIG_LTE_NEIGHBOR_CELLS_MAX];
 static struct lte_lc_cell gci_cells[CONFIG_LTE_NEIGHBOR_CELLS_MAX];
 static struct lte_lc_cells_info scan_cellular_info = {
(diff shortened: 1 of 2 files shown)


lib/location/scan_cellular.c
```diff
@@ -162,6 +180,23 @@ void scan_cellular_execute(int32_t timeout, uint8_t cell_count)
...
+
+	/* If semaphore is reset during the waiting period, the position request was canceled.*/
+	if (k_sem_take(&entered_rrc_idle, K_SECONDS(SCAN_CELLULAR_RRC_IDLE_WAIT_TIME)) == -EAGAIN) {
+		if (!running) {
+			goto end;
+		}
+		/* We're still running, i.e., the wait for RRC idle timed out */
+		LOG_WRN("RRC connection was not released in %d seconds.",
+			SCAN_CELLULAR_RRC_IDLE_WAIT_TIME);
+		goto end;
```
trantanen, line 196: We didn't get to RRC idle and we will use NCELLMEAS=1 results which are 10s old. I'm thinking if that's find...

@tokangas, @jhirsi

tokangas: I think we just have to accept that. Also, if the RRC connection is released, there's time difference between the measurements, might be even more than 10 seconds.

lib/location/scan_cellular.c
```diff
@@ -108,6 +117,15 @@ void scan_cellular_lte_ind_handler(const struct lte_lc_evt *const evt)
 
 		k_sem_give(&scan_cellular_sem_ncellmeas_evt);
 	} break;
+	case LTE_LC_EVT_RRC_UPDATE:
+		if (evt->rrc_mode == LTE_LC_RRC_MODE_CONNECTED) {
+			/* Prevent GNSS from starting while RRC is in connected mode. */
```
jhirsi, line 122: copy paste from GNSS side?

jhirsi (approved):

lib/location/scan_cellular.c
trantanen: If we go into GCI searches, only old data we have is for PCI neighbors. Everything else is fresh from the GCI search including current cell.

peknis (approved):

PR #13318, opened 2023-12-01
applications: Matter Bridge: added support for Generic Switch

markaj-nordic:
* implemented new Matter device type: Generic Switch that is based on the Switch cluster
* combined the Generic Switch with the LBS BT device implementation
* updated documentation
* updated changelog

### DIFF
--- a/applications/matter_bridge/Kconfig
+++ b/applications/matter_bridge/Kconfig
@@ -17,6 +17,10 @@ config BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE
 	bool "Support for Humidity Sensor bridged device"
 	default y
 
+config BRIDGE_GENERIC_SWITCH_BRIDGED_DEVICE
+	bool "Support for Generic Switch bridged device"
+	default y
+
 choice BRIDGED_DEVICE_IMPLEMENTATION
 	prompt "Bridged Device implementation"
 	default BRIDGED_DEVICE_SIMULATED
--- a/applications/matter_bridge/README.rst
+++ b/applications/matter_bridge/README.rst
@@ -46,6 +46,7 @@ The Matter bridge has capability of representing non-Matter bridged devices as d
 The application supports bridging the following Matter device types:
 
 * On/Off Light
+* Generic Switch
 * Temperature Sensor
 * Humidity Sensor
 
@@ -99,7 +100,7 @@ The application supports two bridged device configurations that are mutually exc
 * Bluetooth LE bridged device - This configuration allows to connect a real peripheral Bluetooth LE device to the Matter bridge and represent its functionalities using :ref:`Matter Data Model <ug_matter_overview_data_model>`.
   The application supports the following Bluetooth LE services:
 
-  * Nordic Semiconductor's :ref:`LED Button Service <lbs_readme>` - represented by the Matter On/Off Light device type.
+  * Nordic Semiconductor's :ref:`LED Button Service <lbs_readme>` - represented by the Matter On/Off Light and Generic Switch device types.
   * Zephyr's :ref:`Environmental Sensing Service <peripheral_esp>` - represented by the Matter Temperature Sensor and Humidity Sensor device types.
 
 Depending on the bridged device you want to support in your application, :ref:`enable it using the appropriate Kconfig option <matter_bridge_app_bridged_support_configs>`.
@@ -207,6 +208,7 @@ Adding a simulated bridged device to the Matter bridge
    * *<device_type>* is the Matter device type to use for the bridged device.
      The argument is mandatory and accepts the following values:
 
+      * ``15`` - Generic Switch.
       * ``256`` - On/Off Light.
       * ``770`` - Temperature Sensor.
       * ``775`` - Humidity Sensor.
@@ -220,7 +222,7 @@ Adding a simulated bridged device to the Matter bridge
 
       uart:~$ matter_bridge add 256 "Kitchen Light"
 
-Controlling a simulated OnOff Light bridged device
+Controlling a simulated On/Off Light bridged device
    Use the following command:
 
    .. parsed-literal::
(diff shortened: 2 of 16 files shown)


applications/matter_bridge/README.rst
```diff
@@ -493,6 +499,16 @@ After building the sample and programming it to your development kit, complete t
...
 
+
+            Read the value of the *current-position* attribute from the *switch* cluster using the following command:
+
+            .. parsed-literal::
+               :class: highlight
+
+               ./chip-tool switch read current-position *<bridge_node_ID>* *<bridged_device_endpoint_ID>*
+
+            Note that the Generic Switch implements the momentary switch which, in contrast to the latching switch, remains switched on only as long as the physical button is compressed.
```
wiba-nordic, line 510:
```suggestion
            Note that the Generic Switch implements the momentary switch.
            This means that, in contrast to the latching switch, it remains switched on only as long as the physical button is pressed.
```
Is it correct to say "implements the momentary switch", or should it be "is implemented as a momentary switch"?

applications/matter_bridge/src/ble_providers/ble_onoff_light_data_provider.cpp
```diff
@@ -32,11 +32,12 @@ uint8_t BleOnOffLightDataProvider::GattNotifyCallback(bt_conn *conn, bt_gatt_sub
 	VerifyOrExit(length == sizeof(mOnOff), );
 	VerifyOrExit(provider, );
 
-	/* TODO: Implement invoking command through the binding for OnOff light switch or update state of generic
-	 * switch. */
+	/* Save data received in the notification. */
+	memcpy(&provider->mOnOff, data, length);
```
kkasperczyk-no, line 36: I'm not sure we should modify `mOnOff` here, as this is attribute from completely different cluster, unless we want to sort of bind the switch with light functionality and make button to control LED. I think we should just create a separate member called `mSwitchCurrentPosition` or sth.

applications/matter_bridge/src/bridged_device_types/generic_switch.h
```diff
@@ -0,0 +1,36 @@
...
+	CHIP_ERROR HandleWrite(chip::ClusterId clusterId, chip::AttributeId attributeId, uint8_t *buffer) override
+	{
+		return CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE;
+	};
+
+private:
+	CHIP_ERROR HandleReadSwitch(chip::AttributeId attributeId, uint8_t *buffer, uint16_t maxReadLength);
+
+	static constexpr uint16_t GetSwitchClusterRevision() { return 1; }
+	static constexpr uint32_t GetSwitchClusterFeatureMap() { return 1; }
```
kkasperczyk-no, line 32: Shouldn't it be 2 (MS), instead of 1 (LS)? Also maybe it would be worth to add comment what this feature map value means. I know we have not done it before, but now I see it would be good to do that.

applications/matter_bridge/Kconfig
```diff
@@ -17,6 +17,10 @@ config BRIDGE_HUMIDITY_SENSOR_BRIDGED_DEVICE
 	bool "Support for Humidity Sensor bridged device"
 	default y
 
+config BRIDGE_GENERIC_SWITCH_BRIDGED_DEVICE
+	bool "Support for Generic Switch bridged device"
+	default y
```
kkasperczyk-no, line 22: If the generic switch is not supported for simulated devices, we should add `depends on BRIDGED_DEVICE_BT`.

kkasperczyk-no (commented):
Please update `Overview` and `Bridged device support` readme sections as well. Also we should mention in readme that generic switch is supported only for BT devices.

applications/matter_bridge/src/bridged_device_types/generic_switch.h
markaj-nordic: Ahh of course, it's bit 1 in the case of our implementation.

ArekBalysNordic (approved):
LGTM

applications/matter_bridge/src/ble_providers/ble_onoff_light_data_provider.cpp
```diff
@@ -195,14 +196,22 @@ int BleOnOffLightDataProvider::ParseDiscoveredData(bt_gatt_dm *discoveredData)
...
-void BleOnOffLightDataProvider::NotifyAttributeChange(intptr_t context)
+void BleOnOffLightDataProvider::NotifyOnOffAttributeChange(intptr_t context)
 {
 	BleOnOffLightDataProvider *provider = reinterpret_cast<BleOnOffLightDataProvider *>(context);
 
 	provider->NotifyUpdateState(Clusters::OnOff::Id, Clusters::OnOff::Attributes::OnOff::Id, &provider->mOnOff,
 				    sizeof(provider->mOnOff));
 }
 
+void BleOnOffLightDataProvider::NotifySwitchCurrentPositionAttributeChange(intptr_t context)
```
kkasperczyk-no, line 207: One more minor, would you mind changing this module name from `BleOnOffLightDataProvider` to ` BleLBSDataProvider`? I think it would also follow the naming scheme for environmental profile.

applications/matter_bridge/README.rst
markaj-nordic: I'm not sure, for me both versions seem correct. I can use the latter, no problem.

applications/matter_bridge/README.rst
```diff
@@ -479,9 +482,14 @@ After building the sample and programming it to your development kit, complete t
...
 
                I: Added device to dynamic endpoint 3 (index=0)
+               I: Added device to dynamic endpoint 4 (index=1)
                I: Created 0x100 device type on the endpoint 3
+               I: Created 0xf device type on the endpoint 4
+
+            For the LED Button Service and the Environmental Sensor, two endpoints are created:
 
-            For the Environmental Sensor, two endpoints are created: one implements the Temperature Sensor, and the other implements the Humidity Sensor.
+            * For the LED Button Service, one implements the OnOff Light Device and the other implements the Generic Switch Device.
```
wiba-nordic, line 491: This is written OnOff here, but On/Off on Line 210. Which is correct?

wiba-nordic (approved):

markaj-nordic:
> I discovered a crash when adding the bridged device after commissioning the device to the Matter network, fix needed. DNM.

Let's merge this one. The discovered problem is generic and applies also to the current main branch. There is already a ticket for that https://nordicsemi.atlassian.net/browse/KRKNWK-18171, so it will be fixed separately.

PR #13322, opened 2023-12-04
Fix docs for nRF7001

D-Triveni:
Fixes SHEL-2267

### DIFF
--- a/samples/wifi/provisioning/sample.yaml
+++ b/samples/wifi/provisioning/sample.yaml
@@ -9,6 +9,13 @@ tests:
       - nrf7002dk_nrf5340_cpuapp
     platform_allow: nrf7002dk_nrf5340_cpuapp
     tags: ci_build
+  sample.nrf7001.provisioning:
+    build_only: true
+    integration_platforms:
+      - nrf7002dk_nrf7001_nrf5340_cpuapp
+    platform_allow: nrf7002dk_nrf7001_nrf5340_cpuapp
+    tags: ci_build
+    skip: true
   sample.nrf7002_eks.provisioning:
     build_only: true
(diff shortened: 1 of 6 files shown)


krish2718 (approved):
The PR title should be `Fix twister builds` (sample document is misleading).

D-Triveni:
> The PR title should be `Fix twister builds` (sample document is misleading).

ok, will change.

krish2718:
> [samples: wifi: Add support for nRF7001 target]
>
> Add entry for emulated nRF7001 target but skip
> twister run.

Can you please add a reason why this has to be skipped? IIUC, the title should be 

```
Fix docs for nRF700

 Add entry for emulated nRF7001 target but skip
the twister build, this is mainly for the target to show
up in the supported targets list but don't want to increase
CI time by duplicating (using nRF7002 for CI).
```

bama-nordic (approved):

ioannisg (commented):
Could you also add support for nrf7000 in the scan only related samples (in wifi directory and maybe also the location sample in nrf cloud). Guess the shutdown same is also a good candidate for 7000. 

Or are you planning to fix the support for 7000 in a separate PR?

D-Triveni:
> Could you also add support for nrf7000 in the scan only related samples (in wifi directory and maybe also the location sample in nrf cloud). Guess the shutdown same is also a good candidate for 7000.
> 
> Or are you planning to fix the support for 7000 in a separate PR?

The support for nRF7000 was already added in the Wi-Fi samples (shell, scan and shutdown).

ioannisg (approved):
Thanks

krga2022 (approved):

PR #13328, opened 2023-12-04
manifest: sdk-hostap: Fix race when getting scan results

jukkar:
Fix race when getting wifi scan results.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -111,7 +111,7 @@ manifest:
     - name: hostap
       repo-path: sdk-hostap
       path: modules/lib/hostap
-      revision: 865b82abbc2c01838428ca0cfde77d18c68b1fec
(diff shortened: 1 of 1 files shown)


west.yml
```diff
@@ -111,7 +111,7 @@ manifest:
     - name: hostap
       repo-path: sdk-hostap
       path: modules/lib/hostap
-      revision: 865b82abbc2c01838428ca0cfde77d18c68b1fec
+      revision: pull/138/head
```
krish2718, line 114: Please update to `c11edbd593fc9a22d93c99ee94e4597272678e68`

krish2718:
We can do the manifest update in https://github.com/nrfconnect/sdk-nrf/pull/13343 as that supercedes this.

PR #13330, opened 2023-12-04
bluetooth: gatt_dm: Correct the behavior of bt_gatt_dm_attr_next.

dchat-nordic:
Corrected the behavior of skipping the first attribute.
Also corrected the test which also skipped first attribute.

Jira: NCSDK-23749

### DIFF
--- a/subsys/bluetooth/gatt_dm.c
+++ b/subsys/bluetooth/gatt_dm.c
@@ -558,7 +558,7 @@ const struct bt_gatt_dm_attr *bt_gatt_dm_attr_next(
 	const struct bt_gatt_dm_attr *prev)
 {
 	if (!prev) {
-		prev = dm->attrs;
+		return dm->attrs;
 	}
 
 	if (dm->attrs <= prev) {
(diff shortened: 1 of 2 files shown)


subsys/bluetooth/gatt_dm.c
```diff
@@ -558,7 +558,7 @@ const struct bt_gatt_dm_attr *bt_gatt_dm_attr_next(
 	const struct bt_gatt_dm_attr *prev)
 {
 	if (!prev) {
-		prev = dm->attrs;
```
doki-nordic, line 561: From the function documentation: `@param[in] prev Previous attribute or NULL if we wish to get first attribute (just after service).`.

Returning `dm->attrs` you will return service attribute and that's not intention of this function. Can you verify behavior of this function. Maybe the "print" function should be fixed instead? We are not using this function anywhere but "print". It's a public API function, so changes in behavior should be made with caution.

dchat-nordic:
The behavior of function is correct. No need to change

PR #13338, opened 2023-12-05
lib: location: Skip GNSS if not in RRC idle

trantanen:
If timeout for waiting for RRC idle mode expires and we are still in RRC connected mode, let's not try to get GNSS fix. It will only spend time and power without getting a fix. We also spend the location request timer and fallbacks are not even tried.

In addition, lowering the waiting time for getting to RRC idle mode and PSM from 5mins to 2mins.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -521,6 +521,7 @@ Modem libraries
       Previously, 1-2 searches were performed and now 1-3 will be done depending on the requested number of cells and the number of found cells.
       Also, only GCI cells are counted towards the requested number of cells, and normal neighbors are ignored from this perspective.
     * Cellular positioning not to use GCI search when the device is in RRC connected mode, because the modem cannot search for GCI cells in that mode.
+    * GNSS is not started at all if the device does not enter RRC idle mode within two minutes.
 
 * :ref:`lte_lc_readme` library:
 
--- a/lib/location/method_gnss.c
+++ b/lib/location/method_gnss.c
@@ -41,11 +41,11 @@ BUILD_ASSERT(
 	"CONFIG_LOCATION_SERVICE_EXTERNAL must be enabled");
 #endif
 
-/* Maximum waiting time before GNSS is started regardless of RRC or PSM state [min]. This prevents
+/* Maximum waiting time before proceeding regardless of RRC or PSM state [min]. This prevents
  * Location library from getting stuck indefinitely if the application keeps LTE connection
  * constantly active.
  */
-#define SLEEP_WAIT_BACKSTOP 5
+#define SLEEP_WAIT_BACKSTOP 2
 #if !defined(CONFIG_NRF_CLOUD_AGNSS)
 /* range 10240-3456000000 ms, see AT command %XMODEMSLEEP */
 #define MIN_SLEEP_DURATION_FOR_STARTING_GNSS 10240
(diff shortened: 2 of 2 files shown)


lib/location/method_gnss.c
```diff
@@ -41,11 +41,11 @@ BUILD_ASSERT(
 	"CONFIG_LOCATION_SERVICE_EXTERNAL must be enabled");
 #endif
 
-/* Maximum waiting time before GNSS is started regardless of RRC or PSM state [min]. This prevents
+/* Maximum waiting time before proceeding regardless of RRC or PSM state [min]. This prevents
  * Location library from getting stuck indefinitely if the application keeps LTE connection
  * constantly active.
  */
-#define SLEEP_WAIT_BACKSTOP 5
+#define SLEEP_WAIT_BACKSTOP 2
```
trantanen, line 48: Just to make sure you understand the change. This timeout value is for waiting for RRC idle always when GNSS fix is requested but it also applies to PSM waiting when A-GNSS is disabled.

2 minutes should be fine for RRC idle mode waiting time but is this enough for PSM waiting time?
I'm trying to find values with which we would still get fallbacks to work with default 5min location request timeout. And I realize it may not be possible with A-GNSS disabled when we wait both for RRC idle and PSM.

tokangas (changes requested):

lib/location/method_gnss.c
tokangas: The idea behind the selected value has been that it should be long enough not to cause problems in cases where the RRC connection release or entering PSM takes longer, but to prevent the GNSS method from waiting forever.

Both cases currently use the same timeout, but it does not necessarily have to be like that. Starting from RRC connection release, there's not much point in starting GNSS before the RRC connection release, because GNSS will not get runtime and can not get a fix. So, I think we could even wait for RRC connection release until the GNSS method timeout.

Waiting for PSM is different. We only do that when A-GNSS is disabled. The idea is, that getting a fix in RRC idle is not possible, because GNSS can not decode data from the satellite broadcast. That's not the whole story, though. First, GNSS might not need to decode any data to get a fix if it's hot enough. And secondly, we do not take eDRX into account. If both PSM and eDRX are configured, it would be possible for GNSS to get enough runtime already during PSM active time (depending on the eDRX cycle).

The problem with waiting for PSM is that we don't know for sure if it's necessary or not. We could add another check for eDRX, but we don't know if GNSS needs to decode ephemerides or TOW to get a fix. It's not very likely that GNSS has all the data it needs without A-GNSS, I guess. Delaying GNSS start when getting a fix is not possible is beneficial, because GNSS would get short time windows during legacy DRX and unnecessarily consume power.

@hiltunent Any thoughts?

lib/location/method_gnss.c
```diff
@@ -755,18 +754,12 @@ static bool method_gnss_allowed_to_start(void)
 	}
 
 	LOG_DBG("%s", k_sem_count_get(&entered_rrc_idle) == 0 ?
-		"Waiting for the RRC connection release..." :
+		"Waiting for the RRC connection release for 2 minutes..." :
 		"RRC already in idle mode");
 
 	/* If semaphore is reset during the waiting period, the position request was canceled.*/
```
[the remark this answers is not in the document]
trantanen: I don't want to remove the condition text "If semaphore is reset" because comparing to `-EAGAIN` doesn't say that directly.

tokangas (approved):

lib/location/method_gnss.c
trantanen: Decided to keep the implementation as is after some discussions.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -521,6 +521,7 @@ Modem libraries
       Previously, 1-2 searches were performed and now 1-3 will be done depending on the requested number of cells and the number of found cells.
       Also, only GCI cells are counted towards the requested number of cells, and normal neighbors are ignored from this perspective.
     * Cellular positioning not to use GCI search when the device is in RRC connected mode, because the modem cannot search for GCI cells in that mode.
+    * GNSS acquisition not to occur at all if the device does not get into RRC idle mode within 2 minutes.
```
peknis, line 524:
```suggestion
    * GNSS acquisition not to occur at all if the device does not enter RRC idle mode within two minutes.
```

peknis (changes requested):

peknis (approved):

hiltunent (approved):

jhirsi (approved):

PR #13344, opened 2023-12-05
cmake: define edge_impulse as a static imported library

tejlmand:
The edge impulse library is created using ExternalProject_Add.
This means the library is not a regular CMake library but an imported
library.

In general CMake is happy to link directly to a library using its path,
however in the edge_impulse library case, the edge_impulse itself
needs to link against the Zephyr kernel library as it calls kernel
functions. Therefore a link dependency to Zephyr kernel lib must be
created and to do so, edge impulse must be a known library target.
A static imported Edge impulse pointing to the external project
generated library file is created so that a CMake library target exists.

Jira: NCSDK-24520


MarekPieta:
FYI @rakons, @nvlsianpu

MarekPieta (approved):

rakons (approved):

PR #13352, opened 2023-12-06
doc: known_issues: Add known issue peripherals are secure on nrf91

joerchan:
Add known issue for nrf9160 peripherals NRF_PDM and NRF_I2S are configure as secure peripherals and cannot be accessed by the nonsecure application without triggering a security violation.


mia-ko (approved):

frkv (approved):
LGTM

PR #13357, opened 2023-12-06
lib: dk_buttons_and_leds: add shell for buttons

RobertGalatNordic:
There are applications where it would be convinient, especially during testing to be able to simulate button press.

The dk_buttons_and_leds module seems to be a good place, to inject simulated button presses for the application above.

This addition is disabled by default, and can be enabled through Kconfg.


ktaborowski (approved):
looks good

lib/dk_buttons_and_leds/Kconfig
```diff
@@ -22,4 +22,8 @@ module = DK_LIBRARY
 module-str = DK library
 source "${ZEPHYR_BASE}/subsys/logging/Kconfig.template.log_config"
 
+config DK_LIBRARY_SHELL
+	bool "Enable shell to control DK_LIBRARY and allow to simulate button press"
```
tejlmand, line 26: please avoid `Enable` in prompts.

Also remember that the people who will see the line are those using menuconfig of similar, therefore we should avoid long texts as prompt, instead the `help` keyword should be used for detailed description.

The bonus on that is that the help text will automatically be rendered into NCS documentation.

Like this:
```
config DK_LIBRARY_SHELL
	bool "<Short and as descriptive as possible title of setting>"
Comment
	depends on SHELL
	help
	  <Long and detailed description of this setting>
```

tejlmand (changes requested):

lib/dk_buttons_and_leds/Kconfig
RobertGalatNordic: refactored Kconfig entry as suggested

tejlmand (approved):

PR #13358, opened 2023-12-06
applications: Matter Bridge: Fixed a crash when subscribing to LBS

markaj-nordic:
* The crash was caused by the uninitialized bt_gatt_subscribe_params structure that implied function calls from random memory location in GATT implementation internals.
* The crash was not a case for ESP, because there we had a default initialization of class members added.
* Added proper initializaton of GATT related structures to the LSB data provider to avoid mentioned crash.
* Moved ParseDiscoveredData call after the connection callback is executed, as it makes more sense to accept and forward BLE notifications only when we have the connection handled and Matter device counterpart ready.


kkasperczyk-no (approved):
Good job!

LuDuda (approved):

PR #13362, opened 2023-12-07
mpsl: make mpsl sw interrupt kconfig configurable

Tschet1:
Make the SW interrupt that MPSL is using configurable through KConfig. Like that, MPSL can better cooperate with other libraries that might use the same SW interrupt.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -141,7 +141,7 @@ manifest:
     - name: nrfxlib
       repo-path: sdk-nrfxlib
       path: nrfxlib
-      revision: 3398c8491f302e2a48d300d8f08b96cb505e1a64
+      revision: dcb888fd890fa409a15ce030ff10ffe2debd0d6c
     - name: trusted-firmware-m
       repo-path: sdk-trusted-firmware-m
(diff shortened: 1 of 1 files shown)


subsys/mpsl/init/Kconfig
```diff
@@ -56,6 +56,14 @@ config MPSL_TRIGGER_IPC_TASK_ON_RTC_START_CHANNEL
 	  The application core part of the code needs to be aware of
 	  which channel is used.
 
+config MPSL_LOW_PRIO_IRQN
+	int
+	default $(dt_node_array_prop_int,$(dt_nodelabel_path,swi5),interrupts,0) if SOC_COMPATIBLE_NRF52X
```
jori-nordic, line 61: discussed OOB: I suggest to drop the dts complexity and "just hardcode" the number. 

When we do have to revisit this, I propose we do the right thing (IMO) and add a mpsl driver and configure the SWI through a property.

jori-nordic (approved):

rugeGerritsen (approved):

PR #13145, opened 2023-11-20
drivers: wifi: Add option to skip local MAC in scan results

rado17:
Add option to suppress the AP entries in scan results which have the local bit set in the MAC address, or the MAC belongs to the reserved IANA Unicast MAC address range.

Fixes SHEL-2024.

PR #13365, opened 2023-12-07
samples: cellular: Skip WiFi APs with local MAC

trantanen:
Take CONFIG_WIFI_NRF700X_SKIP_LOCAL_ADMIN_MAC Kconfig option into use from PR #13145. WiFi APs with local MAC will not be returned as scan results as they are not useful for positioning purposes. nrf_cloud library is ignoring these but from memory perspective, it's better to drop them at the driver level.


samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
```diff
@@ -95,7 +96,7 @@ CONFIG_LOCATION_TRACKING_WIFI=y
 CONFIG_LOCATION_METHOD_WIFI=y
 # Align this with CONFIG_LOCATION_METHOD_WIFI_SCANNING_RESULTS_MAX_CNT.
 # Also see comments for CONFIG_HEAP_MEM_POOL_SIZE.
-# We set this to 30 in this overlay (as opposed to 60 in the wifi-scan-only overlay) so that
+# We set this to 30 in this overlay so that
```
trantanen, line 99: @glarsennordic I don't know what you'd like to have here but 60 is wrong as it's currently 20 in overlay-nrf7002ek-wifi-scan-only.conf

jayteemo (approved):

gregersrygg (approved):
ATv2 change looks good to me.

Is «WiFi APs with local MAC» useful in any scenario? Could it be disabled by default instead of explicitly disabled in each sample?

Side-note; would it make sense to move overlay-nrf7002ek-wifi-scan-only, or parts of it, out as a re-usable snippet? No need to do in this PR, but please consider as a future improvement.

trantanen:
> Is «WiFi APs with local MAC» useful in any scenario? Could it be disabled by default instead of explicitly disabled in each sample?

«WiFi APs with local MAC» is useful in data transfer and there are rare scanning cases also where they could be used.

> Side-note; would it make sense to move overlay-nrf7002ek-wifi-scan-only, or parts of it, out as a re-usable snippet? No need to do in this PR, but please consider as a future improvement.

I've created a separate ticket so I will check this. Parts of it could be made into a snippet surely. I guess different apps and samples still would need these conf files but I need to see if some could actually get rid of it.

jhirsi (approved):

glarsennordic (approved):

samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
trantanen: > If you wind up rebasing, could you set this to 20?

So you want `CONFIG_WIFI_MGMT_SCAN_MAX_BSS_CNT=20`?

glarsennordic: Yes, but don't bother with it unless you are already making changes to the area, we can always change it ourselves when we next adjust configs. Just making the comment accurate (as you have done) is just fine

trantanen: Ok, I was about to change this already but I guess `CONFIG_HEAP_MEM_POOL_SIZE` comment and maybe even the value should be tuned so I'll leave it to you.

PR #13370, opened 2023-12-08
lib: pdn: add function to retrieve dynamic parameters

tomi-font:
To retrieve parameters returned by AT+CGCONTRDP=<cid>.

### DIFF
--- a/lib/pdn/pdn.c
+++ b/lib/pdn/pdn.c
@@ -12,6 +12,7 @@
 #include <limits.h>
 #include <zephyr/init.h>
 #include <zephyr/kernel.h>
+#include <zephyr/net/socket.h>
 #include <zephyr/sys/slist.h>
 #include <zephyr/logging/log.h>
 #include <nrf_modem_at.h>
@@ -596,6 +597,35 @@ int pdn_id_get(uint8_t cid)
 	}
 
 	return strtoul(p + 1, NULL, 10);
+}
+
+int pdn_dynamic_params_get(uint8_t cid, struct in_addr *dns4_pri,
+			   struct in_addr *dns4_sec, unsigned int *ipv4_mtu)
+{
+	int ret;
+	char resp[256];
+	const char *fmt;
+	char dns4_pri_str[INET_ADDRSTRLEN];
+	char dns4_sec_str[INET_ADDRSTRLEN];
+
+	ret = nrf_modem_at_cmd(resp, sizeof(resp), "AT+CGCONTRDP=%u", cid);
+	if (ret) {
+		LOG_ERR("Failed to read dynamic params for CID %u, err %d", cid, ret);
+		return ret;
+	}
+	   /* "+CGCONTRDP: 0,,"example.com","","","198.276.154.230","12.34.56.78",,,,,1464" */
+	fmt = "+CGCONTRDP: %*u,,\"%*[^\"]\",\"\",\"\",\"%15[0-9.]\",\"%15[0-9.]\",,,,,%u";
+
+	/* If IPv4 is enabled, it will be the first response line. */
+	if (sscanf(resp, fmt, &dns4_pri_str, &dns4_sec_str, ipv4_mtu) != 3) {
+		return -EBADMSG;
+	}
+
+	if (zsock_inet_pton(AF_INET, dns4_pri_str, dns4_pri) != 1
+	 || zsock_inet_pton(AF_INET, dns4_sec_str, dns4_sec) != 1) {
+		return -EFAULT;
+	}
+	return 0;
 }
 
 int pdn_default_apn_get(char *buf, size_t len)
--- a/include/modem/pdn.h
+++ b/include/modem/pdn.h
@@ -10,6 +10,7 @@
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <zephyr/net/net_ip.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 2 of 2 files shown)


lib/pdn/pdn.c
```diff
@@ -598,6 +599,37 @@ int pdn_id_get(uint8_t cid)
...
+	if (ret) {
+		LOG_ERR("Failed to read dynamic params for CID %u, err %d", cid, ret);
+		return ret;
+	}
+	   /* "+CGCONTRDP: 0,,"example.com","","","198.276.154.230","198.276.154.203",,,,,1464" */
+	fmt = "+CGCONTRDP: %*u,,\"%*[^\"]\",\"\",\"\",\"%n%*15[0-9.]\",\"%n%*15[0-9.]\",,,,,%u";
+	ret = sscanf(resp, fmt, &dns4_pri_offset, &dns4_sec_offset, ipv4_mtu);
+	if (ret != 1) {
+		return -EBADMSG;
+	}
```
SeppoTakalo, line 621: Why don't you read the addresses into a string buffer?
It looks a lot cleaner.

```suggestion
	char pri[IPV6_ADDR_STR_LEN];
	char sec[IPV6_ADDR_STR_LEN];
	char *fmt = "+CGCONTRDP: %*u,,\"%*[^\"]\",\"\",\"\",\"%[^\"]\",\"%[^\"]\",,,,,%u";

	ret = sscanf(resp, fmt, &pri, &sec, &ipv4_mtu);
	if (ret != 3) {
		printf("EBADMSG\n");
		return -1;
	}
	inet_pton(AF_INET, pri, dns4_pri); // ...
```

lib/pdn/pdn.c
```diff
@@ -598,6 +599,37 @@ int pdn_id_get(uint8_t cid)
...
+	/* Null terminate the DNS IP addresses to be able to parse them into in_addr structs. */
+	*strchr(resp + dns4_pri_offset, '"') = '\0';
+	*strchr(resp + dns4_sec_offset, '"') = '\0';
+	if (zsock_inet_pton(AF_INET, resp + dns4_pri_offset, dns4_pri) != 1
+	 || zsock_inet_pton(AF_INET, resp + dns4_sec_offset, dns4_sec) != 1) {
```
SeppoTakalo, line 626: Can we really assume AF_INET? How about IPv6?

MarkusLassila (approved):

lib/pdn/pdn.c
```diff
@@ -598,6 +599,35 @@ int pdn_id_get(uint8_t cid)
...
 }
 
+int pdn_dynamic_params_get(uint8_t cid, struct in_addr *dns4_pri,
+			   struct in_addr *dns4_sec, unsigned int *ipv4_mtu)
+{
+	int ret;
+	char resp[256];
+	const char *fmt;
+	char dns4_pri_str[INET_ADDRSTRLEN];
+	char dns4_sec_str[INET_ADDRSTRLEN];
```
lemrey, line 609: This is quite a big stack

lib/pdn/pdn.c
```diff
@@ -598,6 +599,35 @@ int pdn_id_get(uint8_t cid)
...
+
+	ret = nrf_modem_at_cmd(resp, sizeof(resp), "AT+CGCONTRDP=%u", cid);
+	if (ret) {
+		LOG_ERR("Failed to read dynamic params for CID %u, err %d", cid, ret);
+		return ret;
+	}
+	   /* "+CGCONTRDP: 0,,"example.com","","","198.276.154.230","12.34.56.78",,,,,1464" */
+	fmt = "+CGCONTRDP: %*u,,\"%*[^\"]\",\"\",\"\",\"%15[0-9.]\",\"%15[0-9.]\",,,,,%u";
```
lemrey, line 617: If you have a scanf format, why not use `nrf_modem_at_scanf` instead of `nrf_modem_at_cmd`?
Using `nrf_modem_at_scanf` which will scanf directly from shared memory, so you won't need any intermediate buffer on the stack (which in this case is quite big, 256 bytes).

lib/pdn/pdn.c
```diff
@@ -598,6 +599,35 @@ int pdn_id_get(uint8_t cid)
...
+
+	/* If IPv4 is enabled, it will be the first response line. */
+	if (sscanf(resp, fmt, &dns4_pri_str, &dns4_sec_str, ipv4_mtu) != 3) {
+		return -EBADMSG;
+	}
+
+	if (zsock_inet_pton(AF_INET, dns4_pri_str, dns4_pri) != 1
+	 || zsock_inet_pton(AF_INET, dns4_sec_str, dns4_sec) != 1) {
+		return -EFAULT;
```
lemrey, line 626: I suggest to use a different value here, `EFAULT` should be used for bad memory access.

lib/pdn/pdn.c
```diff
@@ -598,6 +599,35 @@ int pdn_id_get(uint8_t cid)
...
+	int ret;
+	char resp[256];
+	const char *fmt;
+	char dns4_pri_str[INET_ADDRSTRLEN];
+	char dns4_sec_str[INET_ADDRSTRLEN];
+
+	ret = nrf_modem_at_cmd(resp, sizeof(resp), "AT+CGCONTRDP=%u", cid);
+	if (ret) {
+		LOG_ERR("Failed to read dynamic params for CID %u, err %d", cid, ret);
+		return ret;
```
lemrey, line 614: This returns both positive and negative values, I think we should not mix those. Elsewhere this library takes care of converting positive values to `ENOEXEC`, we should do the same here.

lemrey:
We need a changelog entry for this as well, since it adds a function to the new library.

tomi-font:
Good remarks. I will put up another PR to address them.

PR #13373, opened 2023-12-08
applications: serial_lte_modem: small changes

tomi-font:


SeppoTakalo (approved):

applications/serial_lte_modem/src/slm_at_icmp.c
```diff
@@ -470,11 +470,11 @@ static int ping_test_handler(const char *target)
 
 	/* Use the first result to decide which address family to use */
 	if (res->ai_family == AF_INET) {
-		char ipv4_addr[NET_IPV4_ADDR_LEN] = {0};
+		char ipv4_addr[INET_ADDRSTRLEN];
 
 		LOG_INF("Ping target's IPv4 address");
 		util_get_ip_addr(ping_argv.pdn, ipv4_addr, NULL);
-		if (strlen(ipv4_addr) == 0) {
+		if (!*ipv4_addr) {
```
MarkusLassila, line 477: Might be more understandable for casual reader, if you left the = {0} initialization's in place and removed the zeroing of the first byte of array's from util_get_ip_addr.

One more alternative would be to supply these as ** and allow the util_get_ip_addr to set them to null, if there is no IP address.

MarkusLassila (approved):

PR #13374, opened 2023-12-08
tfm: Fix TF-M initial attestation missing dependency on RNG

joerchan:
Fix TF-M initial attestation missing dependency on RNG. RNG is required during the ECDSA signing.
RNG is required by the tfm_plat_get_boot_seed function.


joerchan:
This is needed after this change: https://github.com/nrfconnect/sdk-zephyr/pull/1406

SebastianBoe (approved):

PR #13375, opened 2023-12-08
Fix MAC address parsing

krish2718:


sachinthegreen (approved):

sachinthegreen (approved):

modules/hostap/src/supp_events.c
```diff
@@ -94,15 +101,17 @@ static int wpa_supp_process_status(struct supp_int_event_data *event_data, char
...
 		event_data->data_len = sizeof(data->bss_added);
 		break;
 	case WPA_SUPP_EVENT_BSS_REMOVED:
 		ret = sscanf(wpa_supp_status + strlen("CTRL-EVENT-BSS-REMOVED "),
 			"%u "MACSTR,
 			&data->bss_removed.id,
-			MAC2STR(data->bss_removed.bssid));
+			MACADDR2STR(tmp_mac_addr));
 		event_data->data_len = sizeof(data->bss_removed);
+		os_memcpy(data->bss_removed.bssid, tmp_mac_addr, ETH_ALEN);
```
jukkar, line 114: The `tmp_mac_addr` is 4 * 6 = 24 bytes long array, but the `bss_removed.bssid` is 6 bytes long, the memcpy will copy garbage to bssid. You need to copy the fields one by one.

modules/hostap/src/supp_events.c
```diff
@@ -28,12 +31,26 @@ static char *wpa_supp_event_map[] = {
 	"CTRL-EVENT-DSCP-POLICY",
 };
 
+static void copy_mac_addr(unsigned int *src, uint8_t *dst)
+{
+	int i;
+
+	for (i = 0; i < ETH_ALEN; i++) {
+#if defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+		dst[i] = (uint8_t)(__bswap_32(src[i]) & 0xFF);
```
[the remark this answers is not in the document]
jukkar: Hmm, I think compiler will generate code that assign variables in this case just fine, only if the value is > 255, then the upper bits are discarded.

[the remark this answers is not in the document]
krish2718: Well, I don't see any warnings with your suggested code, so, only concern would be endianness.

modules/hostap/src/supp_events.c
```diff
@@ -28,12 +31,26 @@ static char *wpa_supp_event_map[] = {
...
+static void copy_mac_addr(const unsigned int *src, uint8_t *dst)
+{
+	int i;
+
+	for (i = 0; i < ETH_ALEN; i++) {
+		if (IS_ENABLED(CONFIG_BIG_ENDIAN)) {
+			dst[i] = (uint8_t)(__bswap_32(src[i]) & 0xFF);
+		} else {
+			dst[i] = (uint8_t)(src[i] & 0xFF);
+		}
```
jukkar, line 43: If the src is already in host byte order this is not needed. You can just do `dst[i] = src[i]` here.

You can test this by this small program yourself:
```
#include <stdio.h>

void main(void)
{
	unsigned int mac_int[] = { 100, 200, 0xff03, 0xff04, 0xff00005, 0xef000006 };
	unsigned char mac_uint8[6];

	for (int i = 0; i < sizeof(mac_uint8); i++) {
		mac_uint8[i] = mac_int[i];
		printf("[%d] = %d  %d\n", i, mac_uint8[i], mac_int[i]);
	}
}
```

krish2718: Yes, I did try a similar test and it worked, and also as you say as both are host byte order endianness shouldn't impact, will update.

jukkar (approved):

PR #13307, opened 2023-12-01
manifest: zephyr: Remove CPU_HAS_NRF_IDAU's dependencies

mswarowsky:
Simplify NRF_IDAU config

PR #13383, opened 2023-12-08
manifest: Update zephyr to fix flash samples and tests with TF-M

joerchan:
Update zephyr to fix flash samples and tests with TF-M. These need to use a valid flash partition configured as non-secure.


rlubos:
Manifest will be updated in #13307

PR #13384, opened 2023-12-08
applications: serial_lte_modem: allow slm_util_at_cmd to receive AT response contents

tomi-font:
Using nrf_modem_at_scanf() instead of nrf_modem_at_printf() allows to parse some of the AT response contents.
With some gymnastics we are effectively able to receive AT responses of up to two lines, which covers most use cases.

This need was discovered when realizing that interception of AT write commands also intercepts the corresponding test commands. Test commands have one response line of usage explanation, which must be forwarded properly to the caller.

### DIFF
--- a/applications/serial_lte_modem/src/slm_util.c
+++ b/applications/serial_lte_modem/src/slm_util.c
@@ -3,6 +3,7 @@
  *
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
+#include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 #include <errno.h>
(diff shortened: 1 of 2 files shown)


applications/serial_lte_modem/src/slm_util.c
```diff
@@ -93,32 +94,67 @@ int slm_util_at_scanf(const char *cmd, const char *fmt, ...)
...
+	LOG_DBG("Forwarding \"%s\" to the modem.", at_cmd);
+
+/* This function gets called by interception callbacks invoked by nrf_modem_at_cmd().
  * Here nrf_modem_at_cmd() must be bypassed because it would otherwise call
  * the same interception callbacks and infinite recursion would happen.
  */
-#undef nrf_modem_at_printf
-	const int ret = nrf_modem_at_printf("%s", at_cmd);
+#undef nrf_modem_at_scanf
+	const int line_count = nrf_modem_at_scanf(at_cmd, format_str, buf, second_line);
```
MarkusLassila, line 129: What happens if the response is 3 lines or more? If that is not the case, is then the OK is not in the resp_str and whole response is overwritten with ERROR?

applications/serial_lte_modem/src/slm_util.c
```diff
@@ -93,32 +94,67 @@ int slm_util_at_scanf(const char *cmd, const char *fmt, ...)
...
+	return success_ret;
+}
+
+int slm_util_at_cmd_no_intercept(char *buf, size_t len, const char *at_cmd)
+{
+	int ret;
+	char second_line[strlen("+CMx ERROR: xxxx")];
+	char format_str[24];
+
+	/* Construct the format string to read two response lines (into buf and second_line). */
```
MarkusLassila, line 117: Could mention to see the scanf format specifier for more information.

applications/serial_lte_modem/src/slm_util.c
```diff
@@ -93,32 +94,67 @@ int slm_util_at_scanf(const char *cmd, const char *fmt, ...)
...
+			" The buffer needs to be made bigger.", printf_ret - len + 1);
+		return -NRF_E2BIG;
+	}
+	return success_ret;
+}
+
+int slm_util_at_cmd_no_intercept(char *buf, size_t len, const char *at_cmd)
+{
+	int ret;
+	char second_line[strlen("+CMx ERROR: xxxx")];
```
MarkusLassila, line 114: When AT-command response returns some actual values, this expects that the following line (OK), fits in here. I take it that if there is multiple lines, then we just respond with ERROR, even if the AT-command itself was successful (as we did not receive OK)?

applications/serial_lte_modem/src/slm_util.c
```diff
@@ -93,32 +94,67 @@ int slm_util_at_scanf(const char *cmd, const char *fmt, ...)
...
+
+#define AT_ERROR (NRF_MODEM_AT_ERROR << 16)
+	assert(nrf_modem_at_err_type(AT_ERROR) == NRF_MODEM_AT_ERROR);
 
-	if (!ret || err == NRF_MODEM_AT_ERROR) {
-		snprintf(buf, len, "%s\r\n", ret ? "ERROR" : "OK");
+	/* Reconstruct the response string and code from the parsed response line(s). */
+	if (!strcmp(resp_str, "OK")) {
+		/* The command succeeded and its response is no longer than two lines. */
+		ret = terminate_at_response(buf, len, (line_count == 1) ? "\r\n" : "\r\nOK\r\n", 0);
```
MarkusLassila, line 150: Another way to go about this method would have been to have the whole answer in the single response buf and then search for the responses and line changes from there. What was the reasoning in scanning to two lines?

MarkusLassila (changes requested):

applications/serial_lte_modem/src/slm_util.c
tomi-font: Correct. If the response is longer than two lines, in which case the OK would not be in the first two lines.

applications/serial_lte_modem/src/slm_util.c
tomi-font: This would have been ideal, but this is limited by the fact that `nrf_modem_at_cmd()` cannot be called from here.
And with `scanf()` to my knowledge it's not possible to scan the whole response with a single format specifier. Only one line at a time. So because it doesn't scale well and (as of now) this is only used for commands that return at most two-line responses, I went for this.
`nrf_modem_at_cmd_async()` could allow to receive the whole response, but it comes with potential concurrency issues. Though it might be possible to work around them. Would require some thinking. Could be explored if/when there is a need to receive more than two lines of response from commands.

Do you have any suggestion regarding this?

applications/serial_lte_modem/src/slm_util.c
tomi-font: Correct.

applications/serial_lte_modem/src/slm_util.c
MarkusLassila: Spent some time reading the scanf description, and I am not entirely sure if you could get it to fetch the newline as well, if you use for example the negated scanset for something else. Unsure if it is possible to use scanset for everything/empty negated scanset.

That said, it might not be a good practice, even if it did work. Also, I am not sure we need to add the nrf_modem_at_cmd_async at this point. As this is completely new implementation, it is OK, to have a limitation at this point. 

Previously, I just did not understand the reason for the limitation. You can resolve this.

tomi-font: Unless there is some special character that can be used and is guaranteed to not appear, I'm afraid that we cannot get all the output at once. It's not possible to use an empty scanset, and putting `'\0'` obviously cannot work.

@lemrey @eivindj-nordic If you have some more insight/idea of a scanf format specifier that could fetch all the response at once, that would be very welcome.

If not, we will (for now) continue with this limitation as it isn't (yet) blocking a use case.

tomi-font: I tried, and actually specifying a character that doesn't appear in the response (e.g. `'\x80'` allows to fetch the whole response into a single buffer.

So we would just need a guarantee for a certain character that it cannot appear in an AT response.

MarkusLassila (approved):

applications/serial_lte_modem/src/slm_util.c
tomi-font: Agreed.

rlubos (approved):

PR #13390, opened 2023-12-11
samples: matter: common: Added binding handler

LipinskiPNordicSemi:
This commit remove binding handlers from thermostat and light switch, and creates one unified module in common section.


samples/matter/common/src/binding_handler.cpp
```diff
@@ -27,62 +15,78 @@ using namespace chip::app;
...
 
-	TemperatureMeasurementUnicastBindingRead(data, binding, peer_device);
+	/* If session was recovered and communication works, reset flag to the initial state. */
+	if (bindingData->CaseSessionRecovered)
+		bindingData->CaseSessionRecovered = false;
 }
 
-void BindingHandler::TemperatureMeasurementUnicastBindingRead(BindingData *data, const EmberBindingTableEntry &binding,
-							      OperationalDeviceProxy *peer_device)
+void BindingHandler::OnInvokeCommandFailure(BindingData *bindingData, CHIP_ERROR aError)
```
Damian-Nordic, line 30: Let's be consistent wrt to using `aArg` notation for arguments - I would personally not use it but at least let it be consistent in a single source file :).

samples/matter/common/src/binding_handler.cpp
```diff
@@ -27,62 +15,78 @@ using namespace chip::app;
...
-		ChipLogError(NotSpecified, "Read Thermostat attribute failed: %" CHIP_ERROR_FORMAT, error.Format());
-	};
+		/* Set flag to not try recover session multiple times. */
+		bindingData->CaseSessionRecovered = true;
 
-	CHIP_ERROR ret = CHIP_NO_ERROR;
+		/* Allocate new object to make sure its life time will be appropriate. */
+		BindingHandler::BindingData *data = Platform::New<BindingHandler::BindingData>();
+		VerifyOrReturn(data != nullptr, LOG_ERR("New object allocation failed"));
+		data = bindingData;
```
Damian-Nordic, line 43: `data` is leaked here, right? Do we need to allocate a new object?

Damian-Nordic (commented):
I like the idea, but I think the common API can be a bit more type-safe, and I found a few possible bugs. The comments for the API design can, of course, be addressed in a followup (whatever suits you better) :).

samples/matter/common/src/binding_handler.h
```diff
@@ -0,0 +1,47 @@
...
+		uint8_t Value;
+		bool IsGroup{ false };
+		bool CaseSessionRecovered{ false };
+	};
+
+	static void Init();
+	void PrintBindingTable();
+	bool IsGroupBound();
+
+	static void DeviceWorkerHandler(intptr_t aContext);
```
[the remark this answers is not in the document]
LipinskiPNordicSemi: Good point, thanks!

samples/matter/light_switch/src/app_task.cpp
```diff
@@ -204,6 +186,24 @@ CHIP_ERROR AppTask::Init()
 	 */
 	PlatformMgr().AddEventHandler(ChipEventHandler, 0);
 
+	LightSwitch::GetInstance().Init(kLightSwitchEndpointId);
```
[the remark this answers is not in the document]
LipinskiPNordicSemi: It has to be initialized later than in previous one, or we can go back to SheduleWork(Init) in the binding handler.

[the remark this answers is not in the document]
Damian-Nordic: I see, thanks!

samples/matter/thermostat/src/temperature_measurement/real_temp.cpp
```diff
@@ -23,15 +31,39 @@ constexpr chip::EndpointId mTemperatureMeasurementEndpointId = 1;
...
+}
+
+void RealSensor::TemperatureMeasurementReadHandler(const EmberBindingTableEntry &binding,
+						   OperationalDeviceProxy *deviceProxy,
+						   BindingHandler::BindingData &bindingData)
+{
+	auto onSuccess = [dataPointer = Platform::New<BindingHandler::BindingData>(bindingData)](
+				 const ConcreteDataAttributePath &attributePath, const auto &dataResponse) {
+		ChipLogProgress(NotSpecified, "Read Temperature Sensor attribute succeeded");
+		int16_t responseValue = dataResponse.Value();
```
Damian-Nordic, line 54: Are you sure you don't need to check if `dataResponse.HasData()`? The attribute is nullable.

samples/matter/common/src/binding_handler.cpp
```diff
@@ -0,0 +1,155 @@
...
+{
+	CHIP_ERROR error;
+
+	if (Error == CHIP_ERROR_TIMEOUT && !bindingData->CaseSessionRecovered) {
+		LOG_INF("Response timeout for invoked command, trying to recover CASE session.");
+
+		/* Set flag to not try recover session multiple times. */
+		bindingData->CaseSessionRecovered = true;
+
+		VerifyOrDie(bindingData != nullptr);
```
Damian-Nordic, line 47: This should be moved to the beginning (or before bindingData is dereferenced);

samples/matter/common/src/binding_handler.cpp
```diff
@@ -0,0 +1,155 @@
...
+void BindingHandler::RunBoundClusterAction(BindingData *bindingData)
+{
+	VerifyOrReturn(bindingData != nullptr, LOG_ERR("Invalid binding data"));
+
+	DeviceLayer::PlatformMgr().ScheduleWork(DeviceWorkerHandler, reinterpret_cast<intptr_t>(bindingData));
+}
+
+void BindingHandler::OnInvokeCommandSucces(BindingData *bindingData)
+{
+	LOG_DBG("Binding command applied successfully!");
```
Damian-Nordic, line 30: It seems bindingData must be released in these callbacks.

Damian-Nordic (approved):
3 last issues found, but LGTM overall.

kkasperczyk-no (approved):
Good job, let's finally ship it!

kkasperczyk-no:
@LipinskiPNordicSemi please squash commits and push once again to pass compliance check.

Damian-Nordic:
fyi, squashed the commits on behalf of Patryk to enable merging the PR today if there are no more comments.

samples/matter/light_switch/src/light_switch.h
```diff
@@ -25,8 +27,8 @@ class LightSwitch {
 		Off /* Turn off light on lighting-app device */
 	};
 
-	void Init(chip::EndpointId aLightSwitchEndpoint);
-	void InitiateActionSwitch(Action);
+	void Init(chip::EndpointId LightSwitchEndpoint);
+	void InitiateActionSwitch(Action mAction);
```
markaj-nordic, line 31:
```suggestion
	void InitiateActionSwitch(Action action);
```
to avoid confusing with members.

samples/matter/light_switch/src/light_switch.h
```diff
@@ -25,8 +27,8 @@ class LightSwitch {
 		Off /* Turn off light on lighting-app device */
 	};
 
-	void Init(chip::EndpointId aLightSwitchEndpoint);
-	void InitiateActionSwitch(Action);
+	void Init(chip::EndpointId LightSwitchEndpoint);
```
markaj-nordic, line 30:
```suggestion
	void Init(chip::EndpointId lightSwitchEndpoint);
```

samples/matter/thermostat/src/temperature_measurement/simulated_temp.cpp
```diff
@@ -7,6 +7,8 @@
 #include "sensor.h"
 #include "temp_sensor_manager.h"
 
+#include <atomic>
```
markaj-nordic, line 10: Why is it needed?

samples/matter/thermostat/src/temperature_measurement/real_temp.cpp
```diff
@@ -23,15 +31,44 @@ constexpr chip::EndpointId mTemperatureMeasurementEndpointId = 1;
...
 {
 	BindingHandler::BindingData *data = Platform::New<BindingHandler::BindingData>();
-	data->clusterId = Clusters::TemperatureMeasurement::Id;
-	data->localEndpointId = mTemperatureMeasurementEndpointId;
-	data->isGroup = BindingHandler::GetInstance().IsGroupBound();
-	DeviceLayer::PlatformMgr().ScheduleWork(BindingHandler::ThermostatWorkerHandler,
-						reinterpret_cast<intptr_t>(data));
+	data->ClusterId = Clusters::TemperatureMeasurement::Id;
+	data->EndpointId = mTemperatureMeasurementEndpointId;
+	data->IsGroup = BindingHandler::IsGroupBound();
```
markaj-nordic, line 42: We are reading this flag from BindingHandler and then passing into the BindingHandler::RunBoundClusterAction... so it looks to me like something that can be handled solely by the BindingHandler with no need for additional action in the application. Or do I miss something here?

samples/matter/common/src/binding_handler.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <app-common/zap-generated/ids/Clusters.h>
+#include <app-common/zap-generated/ids/Commands.h>
+#include <app/CommandSender.h>
+#include <app/clusters/bindings/BindingManager.h>
+#include <controller/InvokeInteraction.h>
+#include <platform/CHIPDeviceLayer.h>
+
+class BindingHandler {
+public:
```
markaj-nordic, line 16: What I miss here is the documentation of the public API. We can however deffer that to followup task/PR.

markaj-nordic: Also, isn't the BindingHandler stateless class? Look like we are using class only to provide the meaningful namespace. The same can be achieved with just `namespace` and regular functions. So object oriented code makes sense to me if we foresee possible extensions of the class in the future. I don't have a string opinion on that, and I accept the contribution. It just looks a bit odd to have a class with all members static :)

samples/matter/common/src/binding_handler.cpp
```diff
@@ -0,0 +1,156 @@
...
+			break;
+		default:
+			break;
+		}
+	}
+}
+
+void BindingHandler::DeviceWorkerHandler(intptr_t context)
+{
+	VerifyOrDie(context != 0);
```
markaj-nordic, line 146: Isn't it too aggressive? Should be really kill the app? BTW. Isn't is better to check for the nullptr after line 147?

markaj-nordic (approved):
LGTM overall. Thanks for addressing my comments from the internal review. Added some final remarks though.

samples/matter/common/src/binding_handler.cpp
Damian-Nordic: This is not passed by a class user, so if this ends up nullptr it means this component is broken so I would say assertions/VerifyOrDie are exactly for these situations :)

samples/matter/thermostat/src/temperature_measurement/real_temp.cpp
Damian-Nordic: Asked Patryk about it as well but we thought we could change that in the followup. Long story short - shell commands use this force unicast/groupcast binding (I don't think that's very useful tbh and I'm +1 for making this transparent for a user).

samples/matter/common/src/binding_handler.cpp
markaj-nordic: Fair enough :)

samples/matter/thermostat/src/temperature_measurement/real_temp.cpp
markaj-nordic: Ok, I'm ok with follow ups as long as we we create a task in the according epic right away :)

markaj-nordic: Approving on condition that we create aforementioned task.

PR #13394, opened 2023-12-11
applications: serial_lte_modem: Allow building on older DK

SeppoTakalo:
nRF9160 DK board revision 0.14 added external flash that is required for Full Modem FOTA.

Separate overlays for board revisions so we can build for older DK without using the flash.

### DIFF
--- a/applications/serial_lte_modem/boards/nrf9160dk_nrf9160_ns_0_14_0.overlay
+++ b/applications/serial_lte_modem/boards/nrf9160dk_nrf9160_ns_0_14_0.overlay
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/* Enable external flash */
+&arduino_spi {
+	mx25r64: mx25r6435f@1 {
+		status = "okay";
+	};
+};
--- a/applications/serial_lte_modem/doc/FOTA_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/FOTA_AT_commands.rst
@@ -35,7 +35,8 @@ Syntax
   * ``3`` - Start FOTA for full modem update.
     Can only be used when the :file:`overlay-full_fota.conf` configuration file is used.
 
-    Not supported on the :ref:`Thingy:91 <thingy91_ug_intro>` as it lacks an external flash to store the firmware image.
+    Not supported on the :ref:`Thingy:91 <thingy91_ug_intro>` or nRF9160 DK board revisions older than 0.14.0 as they lack an external flash to store the firmware image.
+    See :ref:`nrf9160_board_revisions` for more details.
 
   * ``7`` - Read modem DFU area size and firmware image offset (for modem delta update).
   * ``9`` - Erase modem DFU area (for modem delta update).
--- a/applications/serial_lte_modem/overlay-full_fota.conf
+++ b/applications/serial_lte_modem/overlay-full_fota.conf
@@ -4,6 +4,9 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
+# This overlay requires a board with an external flash.
+# For nRF9160 DKs, only revisions 0.14.0 and later have one.
+
 # Enable external flash for full FOTA
 CONFIG_SPI_NOR=y
 
--- a/applications/serial_lte_modem/boards/nrf9160dk_nrf9160_ns.overlay
+++ b/applications/serial_lte_modem/boards/nrf9160dk_nrf9160_ns.overlay
@@ -53,10 +53,3 @@
 		};
 	};
 };
-
-/* Enable external flash */
-&spi3 {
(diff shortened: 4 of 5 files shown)


applications/serial_lte_modem/doc/FOTA_AT_commands.rst
```diff
@@ -35,7 +35,8 @@ Syntax
   * ``3`` - Start FOTA for full modem update.
     Can only be used when the :file:`overlay-full_fota.conf` configuration file is used.
 
-    Not supported on the :ref:`Thingy:91 <thingy91_ug_intro>` as it lacks an external flash to store the firmware image.
+    Not supported on the :ref:`Thingy:91 <thingy91_ug_intro>` or nRF9160 DK board revisions older than 0.14.0 as these lacks an external flash to store the firmware image.
```
tomi-font, line 38:
```suggestion
    Not supported on the :ref:`Thingy:91 <thingy91_ug_intro>` or nRF9160 DK board revisions older than 0.14.0 as these lack an external flash to store the firmware image.
```

applications/serial_lte_modem/overlay-full_fota.conf
```diff
@@ -4,6 +4,9 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
+# This overlay requires nRF9160 DK board revision 0.14.0 or later.
```
tomi-font, line 7: This phrasing could be interpreted such that this overlay only works on nRF9160 DKs, which is not the case (e.g. nRF9161 DK).
I think it could be changed to state more precisely that it's only nRF9160 DK board revisions earlier than 0.14.0 that are not supported.

applications/serial_lte_modem/boards/nrf9160dk_nrf9160_ns_0_14_0.overlay
tomi-font, line 1: Is this overlay picked up for revisions older than 0.14.0?
Could this instead be renamed to remove the `_0_14_0`? Just as is done with the nRF9161 DK. (not sure)

applications/serial_lte_modem/overlay-full_fota.conf
SeppoTakalo: Changed the phrasing a bit. Is it better now?

applications/serial_lte_modem/boards/nrf9160dk_nrf9160_ns_0_14_0.overlay
tomi-font: So does this mean that we need to change how it is done for the nRF9161 DK? (The overlay for older revisions has _0_7_0, while the one for newer revisions has nothing.)

tomi-font: No, actually it's maybe not quite the same story. In the case of the nRF9161 DK it's probably as follows: all revisions include the base overlay, and only the 0.7.0 includes the extra overlay aimed at it.

applications/serial_lte_modem/overlay-full_fota.conf
```diff
@@ -4,6 +4,9 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
+# This overlay requires board with external flash.
+# For example, nRF9160 DK revision 0.14.0 or later.
```
tomi-font, line 8:
```suggestion
# This overlay requires a board with an external flash.
# For nRF9160 DKs, only revisions 0.14.0 and later have one.
```

applications/serial_lte_modem/doc/FOTA_AT_commands.rst
```diff
@@ -35,7 +35,8 @@ Syntax
   * ``3`` - Start FOTA for full modem update.
     Can only be used when the :file:`overlay-full_fota.conf` configuration file is used.
 
-    Not supported on the :ref:`Thingy:91 <thingy91_ug_intro>` as it lacks an external flash to store the firmware image.
+    Not supported on the :ref:`Thingy:91 <thingy91_ug_intro>` or nRF9160 DK board revisions older than 0.14.0 as these lack an external flash to store the firmware image.
```
peknis, line 38:
```suggestion
    Not supported on the :ref:`Thingy:91 <thingy91_ug_intro>` or nRF9160 DK board revisions older than 0.14.0 as they lack an external flash to store the firmware image.
```

applications/serial_lte_modem/overlay-full_fota.conf
tomi-font: Yes, almost perfect!

peknis (approved):
Approved with a minor nit.

applications/serial_lte_modem/boards/nrf9160dk_nrf9160_ns_0_14_0.overlay
SeppoTakalo: I changed the naming.. Looks like the common overlay is picked on all versions.

MarkusLassila (approved):

tomi-font (approved):

PR #13398, opened 2023-12-11
samples: bluetooth: Fix central_uart and peripheral_uart samples

maje-emb:
- Increases the size of the main stack. The settings subsystem needs more stack.
- Increases the sysworkq stack size for minimal configuration.
- Corrects the timeout setting for the `uart_rx_enable' function.


samples/bluetooth/central_uart/prj.conf
```diff
@@ -25,8 +25,9 @@ CONFIG_BT_SCAN_UUID_CNT=1
 CONFIG_BT_GATT_DM=y
 CONFIG_HEAP_MEM_POOL_SIZE=2048
 
-# This example requires more workqueue stack
+# This example requires more stack
 CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=2048
+CONFIG_MAIN_STACK_SIZE=1152
```
jori-nordic, line 30: those numbers look very precise. did you get them using the thread analyzer?
Some more padding would be good, so another PR like this is not needed for a while.
At least, getting the values from running with `CONFIG_DEBUG=y`, as I imagine customers might want to do that.

KAGA164 (approved):

samples/bluetooth/central_uart/prj.conf
maje-emb: I based the configuration on the peripheral_uart sample.
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/bluetooth/peripheral_uart/prj.conf#L42
You are right with CONFIG_DEBUG=y is a value that is too small. Increased it to 1.5kB

jori-nordic (approved):

dchat-nordic (approved):

PR #13400, opened 2023-12-11
[DNM] Upmerge 20231208

de-nordic:


de-nordic:
It has been decided that the upmerge will not be updated to the latest sha, closing.

PR #13401, opened 2023-12-12
sample: http_update: Apply modem update after downloading

junqingzou:
Call lib_fota_download API to init DFU target and apply modem update, instead of using app function in main.

### DIFF
--- a/samples/cellular/http_update/modem_delta_update/src/main.c
+++ b/samples/cellular/http_update/modem_delta_update/src/main.c
@@ -22,6 +22,7 @@
 #include <modem/modem_info.h>
 #include <modem/nrf_modem_lib.h>
 #include <net/fota_download.h>
+#include <fota_download_util.h>
 
 #define FOTA_TEST "FOTA-TEST"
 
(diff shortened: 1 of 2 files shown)


eivindj-nordic:
I think the changes look good in itself. The question is whether we want to introduce another library to the sample or keep it closer to the modem library API.

junqingzou:
> I think the changes look good in itself. The question is whether we want to introduce another library to the sample or keep it closer to the modem library API.

Hi Eivind, the API is from the fota_download lib, which has already been enabled by tha sample CONFIG_FOTA_DOWNLOAD=y

junqingzou:
@rlubos could you help to invite Juha Heiskanen (jheiskan81) to review this PR?
For some reason I could not invite.

eivindj-nordic:
After looking at this a bit more and discussing with @lemrey I think it makes sense to keep the modem library init and shutdown in the sample. Using the util might make sense in some applications, though I think we want to expose what is happening with the modem in the sample. The modem shutdown, re-init and its error handling is documented in the modem library documentation, while for the util one has to dive into the code to understand what is happening (I don't see any documentation anywhere). In the end the util does the same, just hides the modem calls, which I think makes it more complex to get started with and to understand.

junqingzou:
> After looking at this a bit more and discussing with @lemrey I think it makes sense to keep the modem library init and shutdown in the sample. Using the util might make sense in some applications, though I think we want to expose what is happening with the modem in the sample. The modem shutdown, re-init and its error handling is documented in the modem library documentation, while for the util one has to dive into the code to understand what is happening (I don't see any documentation anywhere). In the end the util does the same, just hides the modem calls, which I think makes it more complex to get started with and to understand.

I understand. The original point of this PR is to be the first sample that utilizes the new APIs to simplify application development i.e. for customers' awareness. I'll close this PR now.

PR #13403, opened 2023-12-12
wifi: Fix raw tx mode dependency

krish2718:


drivers/wifi/nrf700x/Kconfig
```diff
@@ -62,12 +62,22 @@ config NRF700X_STA_MODE
 
 config NRF700X_AP_MODE
 	bool "Enable Access point mode"
+	depends on NRF700X_STA_MODE
```
sachinthegreen, line 65: Can we remove this dependency, so that AP mode can be compiled independent of STA mode, providing an opportunity to optimize memory by excluding STA specific functionality?

sachinthegreen (changes requested):

drivers/wifi/nrf700x/Kconfig
ioannisg: This is not related to raw TX mode dependencies, right? 

Makes sense to me to have an option for AP-only support - to minimize footprint.

krish2718: > This is not related to raw TX mode dependencies, right?

No, please see the commit log.

krish2718: > Can we remove this dependency, so that AP mode can be compiled independent of STA mode, providing an opportunity to optimize memory by excluding STA specific functionality?

This needs a wider review, so, dropped that commit.

sachinthegreen (approved):

ioannisg (approved):
seems trivial

PR #13407, opened 2023-12-12
boards: arm: thingy91: Do not enable uart1 by default

nordic-krch:
uart1 was enabled by default which was causing initialization in bootloader. It impacted application which was using uart1 because probably it was not uninitialized by the bootloader.

Ref: NCSIDB-1073.

### DIFF
--- a/applications/connectivity_bridge/app.overlay
+++ b/applications/connectivity_bridge/app.overlay
@@ -4,6 +4,10 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 
+&uart1 {
+	status = "okay";
+};
+
 &zephyr_udc0 {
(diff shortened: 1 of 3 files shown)


applications/connectivity_bridge/app.overlay
```diff
@@ -4,6 +4,10 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 
+&uart1 {
+	status = "okay";
+};
+,
```
anangl, line 10: I guess it won't work with this comma here.

anangl (approved):

nordic-krch:
@balaji-nordic does this look ok?

balaji-nordic (approved):
Looks good. Verified that the modem traces functionality on the thing91 is not affected by this change.

PR #13411, opened 2023-12-12
tfm: kconfig: Config SHARE_INSTANCE from prj.conf instead of Kconfig

SebastianBoe:
It is bad practice for Kconfig sources to encode the configuration of their users.

Instead users should enable/disable features in Kconfig fragments like prj.conf.

TFM_REGRESSION_S is one of many users of
TFM_SECURE_UART_SHARE_INSTANCE.


mswarowsky (approved):

modules/trusted-firmware-m/Kconfig
```diff
@@ -310,7 +310,6 @@ if TFM_SECURE_UART
 config TFM_SECURE_UART_SHARE_INSTANCE
 	bool
 	prompt "Share TF-M's UART instance with the non-secure application"
-	depends on !TFM_REGRESSION_S # Secure regressen tests require UART output after boot.
```
joerchan, line 313: If we can do this now, and we don't have to worry about the sdk-zephyr tfm_regression_tests sample, then I think there are more opportunities to clear up the Kconfigs.

config TFM_CRYPTO_IOVEC_BUFFER_SIZE
config MBEDTLS_SHA1_C

config TFM_REGRESSION_NS
	bool
	select NRF_TIMER0_SECURE

config PM_PARTITION_SIZE_TFM
config PM_PARTITION_SIZE_TFM_SRAM

Vge0rge (approved):

PR #13415, opened 2023-12-12
applications: nrf5340_audio: Fix netcore scheduler

pirun:
Turn off sys clock related kconfig to prevent network core scheduler abnormal.

### DIFF
--- a/applications/nrf5340_audio/CMakeLists.txt
+++ b/applications/nrf5340_audio/CMakeLists.txt
@@ -31,6 +31,7 @@ if ((CONFIG_AUDIO_DFU EQUAL 1) OR (CONFIG_AUDIO_DFU EQUAL 2))
     list(APPEND empty_net_core_OVERLAY_CONFIG
         "${CMAKE_CURRENT_LIST_DIR}/dfu/conf/overlay-empty_net_core.conf"
     )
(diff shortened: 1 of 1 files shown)


rick1082 (approved):

erikrobstad (changes requested):
Thank you for the help on this.
Maybe it is better to do CONFIG_SOC_NRF53_RTC_PRETICK=n instead of using CONFIG_SYS_CLOCK_EXISTS?

koffes:
Thanks. Please coordinate this with @andvib.

andvib (changes requested):
I agree with Erik - using CONFIG_SOC_NRF53_RTC_PRETICK=n disables only the functionality that caused the bug to appear.

Maybe we should add a comment to why this must be disabled as well? Mentioning that that the controller isn't affected by the anomaly this fixes, and that there's a resource collision between B0N and the controller

erikrobstad:
> I agree with Erik - using CONFIG_SOC_NRF53_RTC_PRETICK=n disables only the functionality that caused the bug to appear.
> 
> Maybe we should add a comment to why this must be disabled as well? Mentioning that that the controller isn't affected by the anomaly this fixes, and that there's a resource collision between B0N and the controller

Agreed

PR #13417, opened 2023-12-12
net: lib: nrf_cloud: Add config support for ground-fix

plskeggs:
nRF Cloud now supports 3 options for ground-fix requests over REST and MQTT, and will soon support for CoAP.

Add support for these options. Default for it to be disabled for CoAP until cloud support is ready.

Jira: IRIS-7554


tomi-font (approved):
SLM looks good. (CI failures are unrelated, CI should be re-triggered and pass.)

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -644,11 +649,13 @@ Libraries for networking
     * An :c:struct:`nrf_cloud_obj_shadow_data` structure to the :c:struct:`nrf_cloud_evt` structure to be used during shadow update events.
     * The :kconfig:option:`CONFIG_NRF_CLOUD_SEND_SERVICE_INFO_FOTA` Kconfig option to enable sending configured FOTA service info on the device's initial connection to nRF Cloud.
     * The :kconfig:option:`CONFIG_NRF_CLOUD_SEND_SERVICE_INFO_UI` Kconfig option to enable sending configured UI service info on the device's initial connection to nRF Cloud.
+    * An :c:struct:`nrf_cloud_location_config` structure for specifying the desired behavior of an nRF Cloud ground-fix request.
 
   * Updated:
 
     * The :c:func:`nrf_cloud_obj_object_add` function to reset the added object on success.
     * Custom shadow data is now passed to the application during shadow update events.
+    * The :c:func:`nrf_cloud_obj_location_request_create` function and the :c:func:`nrf_cloud_location_request` function to accept the :c:struct:`nrf_cloud_location_config` structure.
```
peknis, line 658:
```suggestion
    * The :c:func:`nrf_cloud_obj_location_request_create` and :c:func:`nrf_cloud_location_request` functions to accept the :c:struct:`nrf_cloud_location_config` structure.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -556,6 +558,8 @@ Modem libraries
     * Cellular positioning not to use GCI search when the device is in RRC connected mode, because the modem cannot search for GCI cells in that mode.
     * GNSS is not started at all if the device does not enter RRC idle mode within two minutes.
 
+  * Added do_reply, hi_conf, and fallback flags to the c:struct:`location_config` structure.
```
peknis, line 561:
```suggestion
  * Added ``do_reply``, ``hi_conf``, and ``fallback`` flags to the c:struct:`location_config` structure.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -663,6 +670,10 @@ Libraries for networking
 
   * Added support for using a password when connecting to a broker.
 
+* :ref:`lib_nrf_cloud_rest` library:
+
+  * Updated the :c:struct:`nrf_cloud_rest_location_request` structure to accept a pointer to a :c:struct:`nrf_cloud_location_config` structure in place of the single disable_response flag.
```
peknis, line 675:
```suggestion
  * Updated the :c:struct:`nrf_cloud_rest_location_request` structure to accept a pointer to a :c:struct:`nrf_cloud_location_config` structure in place of the single ``disable_response`` flag.
```

include/modem/location.h
```diff
@@ -485,6 +485,34 @@ struct location_config {
 	 * location_config_defaults_set() function is called.
 	 */
 	enum location_req_mode mode;
+
+	/**
+	 * @brief Flag to request location back from the cloud.
```
peknis, line 490:
```suggestion
	 * @brief Flag to request location from the cloud.
```

include/modem/location.h
```diff
@@ -485,6 +485,34 @@ struct location_config {
...
+	 * cloud will return the current location, if known. The cloud default is true.
+	 */
+	bool do_reply;
+
+	/**
+	 * @brief Flag to control horizontal positioning error (HPE) for a location.
+	 *
+	 * @details For cellular and Wi-Fi location methods, if this flag is set, the
+	 * cloud will use a 95% confidence interval for the returned location and radius
+	 * than the default 68%. The cloud default is false.
```
peknis, line 502: "...radius than" does not sound right. Maybe "...radius larger than the default 68%.", or something like that.

include/modem/location.h
```diff
@@ -485,6 +485,34 @@ struct location_config {
...
+	 */
+	bool hi_conf;
+
+	/**
+	 * @brief Flag to control handling of position not found errors.
```
peknis, line 507:
```suggestion
	 * @brief Flag to control handling of "position not found" errors.
```

include/modem/location.h
```diff
@@ -485,6 +485,34 @@ struct location_config {
...
+	 *
+	 * @details For cellular and Wi-Fi location methods, if this flag is set, and the
+	 * cloud cannot determine the location based on the cell tower identification and/or
```
peknis, line 510:
```suggestion
	 * cloud cannot determine the location based on the cell tower identification or
```
We try to avoid and/or

include/net/nrf_cloud_location.h
```diff
@@ -57,6 +57,19 @@ struct nrf_cloud_location_result {
 	enum nrf_cloud_error err;
 };
 
+/** @brief Location request config */
+struct nrf_cloud_location_config {
+	/** If true, nRF Cloud will send the location back to the device. */
```
peknis, line 62:
```suggestion
	/** If true, nRF Cloud will send the location to the device. */
```

include/modem/location.h
```diff
@@ -485,6 +485,34 @@ struct location_config {
...
+	 * Wi-Fi access points nearby, it will fallback to coarse position based on the
```
peknis, line 511:
```suggestion
	 * Wi-Fi access points nearby, it will fall back to coarse position based on the
```

include/net/nrf_cloud_location.h
```diff
@@ -92,8 +105,8 @@ typedef void (*nrf_cloud_location_response_t)(const struct nrf_cloud_location_re
  *                 To omit a request parameter, use the appropriate
  *                 `NRF_CLOUD_LOCATION_WIFI_OMIT_` define.
  *                 Can be NULL if cell info is provided.
- * @param request_loc If true, cloud will send location to the device.
- *                    If false, cloud will not send location to the device.
+ * @param config   Optional configuration of request.  If NULL, fall back to default
```
peknis, line 108:
```suggestion
 * @param config   Optional configuration of request. If NULL, fall back to default
```

peknis (changes requested):

SeppoTakalo (approved):

include/modem/location.h
```diff
@@ -485,6 +485,34 @@ struct location_config {
 	 * location_config_defaults_set() function is called.
 	 */
 	enum location_req_mode mode;
+
```
tokangas, line 488: Our intention has been to keep the Location library API as simple as possible. Probably we don't want to have these options in the API at all. In any case, the current approach has several problems.

At least for now, I think the used configuration should be just hard coded in `cloud_service_nrf.c` like the `request_loc` parameter was earlier and all other Location library changes should be removed. Or what do you think, @trantanen?

tokangas (changes requested):

include/modem/location.h
trantanen: Yep, we've tried to keep the the API configuration minimal as this is an abstraction layer for location functionality and advanced users can always use other APIs to implement the details the way they want.
Also, we don't really have any cloud specific configurations in the location request configuration. We've been told that the use of nRF Cloud service is here mainly for testing and prototyping purposes and not that much for final products, which is another reason adding cloud service specific stuff would be nice to avoid.

include/net/nrf_cloud_location.h
```diff
@@ -57,6 +57,19 @@ struct nrf_cloud_location_result {
...
+struct nrf_cloud_location_config {
+	/** If true, nRF Cloud will send the location back to the device. */
+	bool do_reply;
+	/** If true, uncertainty of result will be 95%, otherwise 68%. */
+	bool hi_conf;
+	/** If true, and location cannot be found, result will fall back to rough estimate
+	 *  based on the cell tower tracking area code.  If false, will return an error
+	 *  if the cloud cannot provide a higher accuracy response.
+	 */
+	bool fallback;
```
trantanen, line 70: Is this somehow related to `hi_conf` as it refers to `higher accuracy response`?

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap
```diff
@@ -37,6 +37,15 @@ config NON_RESP_RETRIES
...
 	  by waits for a response.
 
+config NRF_CLOUD_COAP_GF_CONF
+	bool "Enable configuration options for ground-fix requests"
+	help
+	  The initial release of CoAP for nRF Cloud did not provide a way to
+	  configure the response; it defaulted to always returning the location
+	  and method used to acquire the location. The cloud will soon support
+	  configuration values do_reply, fallback, and hi_conf. When the cloud
+	  support is ready, NRF_CLOUD_COAP_GF_CONF can be set true.
```
trantanen, line 47: When cloud supports this, do we need this Kconfig option? OR I don't remember we would have similar for other transports so just thinking to be minimalistic to somplify things.

trantanen (changes requested):
We can have offline discussion also on the Location API configurations if needed.

include/net/nrf_cloud_location.h
plskeggs: No, according to Cole fallback is different. "Higher accuracy response" means the location based on the serving cell plus 0 or more neighbors or GCI cells. "Rough estimate" is less accurate than even a single cell tower location.

He said: "TAC is less accurate than single cell. TAC is unique within a single carrier can cover several towers. TAC is thousands of meters while single cell is hundreds of meters."

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap
plskeggs: I will remove it when the time comes.

include/modem/location.h
plskeggs: I changed it to be "the cloud will use a 95% confidence interval for the returned location and radius instead of the default 68% confidence interval."

include/modem/location.h
plskeggs: Excluding these configuration options from the location library does not work for us, since we changed our multi service sample to use the location library, as did others for ATv2. You are basically saying we have feature-frozen the positioning services as far as our samples are concerned.

Excluding these options also means you will not be able to exercise them from mosh. I would have thought you would be curious and interested to see how they work and want to try them.

We do still have the nrf_cloud_rest_cell_location sample which uses lower level positioning directly, but that one only does cell location not Wi-Fi (per the name).

samples/cellular/modem_shell/src/location/location_srv_ext_nrf_cloud.c
```diff
@@ -105,6 +105,11 @@ static int location_srv_ext_nrf_cloud_location_get(
 {
 	int err;
 	nrf_cloud_location_response_t callback = NULL;
+	struct nrf_cloud_location_config conf = {
+		.do_reply = (callback != NULL), /* Defaults to true on cloud side */
+		.fallback = true,   /* Defaults to true on cloud side */
+		.hi_conf = false    /* Defaults to false on cloud side */
+	};
```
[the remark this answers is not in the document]
plskeggs: They are relevant if you want to make the REST and CoAP requests as small as possible.

samples/cellular/modem_shell/src/location/location_srv_ext_nrf_cloud.c
```diff
@@ -113,8 +118,11 @@ static int location_srv_ext_nrf_cloud_location_get(
 	}
 
 	mosh_print("Sending positioning request to cloud via MQTT");
+	/* Below, no need to send conf if it matches the cloud defaults. */
 	err = nrf_cloud_location_request(
-		cell_data, scanning_results, (callback != NULL), callback);
+		cell_data, scanning_results,
+		(!conf.do_reply || !conf.fallback || conf.hi_conf) ? &conf : NULL,
```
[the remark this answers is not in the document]
plskeggs: The cloud team works hard to avoid changing defaults because they know older devices will be unaware of the change and may no longer work correctly.  So it is fine to make assumptions here about the cloud defaults.

It does cost something, albeit very little, to pass the additional parameters. But since this is mosh, I'll remove this optimization.

subsys/net/lib/nrf_cloud/src/nrf_cloud_rest.c
```diff
@@ -517,7 +518,25 @@ int nrf_cloud_rest_location_get(struct nrf_cloud_rest_context *const rest_ctx,
...
 
-	req.url = request->disable_response ? API_GET_LOCATION_NO_REPLY : API_GET_LOCATION;
+	char url[sizeof(API_GET_LOCATION_CONFIG)];
+	bool do_reply;
+
+	if (request->config) {
+		ret = snprintk(url, sizeof(url) - 1, API_GET_LOCATION_CONFIG,
+			       request->config->do_reply,
+			       request->config->fallback,
+			       request->config->hi_conf);
```
glarsennordic, line 528: Hmm, this client is in HTTP_POST mode, but we are using GET-style URL parameters? It's not **_impossible_** to do that, but it does seem odd

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -980,11 +981,48 @@ int nrf_cloud_obj_location_request_create(struct nrf_cloud_obj *const obj,
...
+		}
+		if (!err) {
+			/* At least one entry differed from defaults, so add it to the obj. */
+			err = nrf_cloud_obj_object_add(obj, NRF_CLOUD_LOCATION_JSON_KEY_CONFIG,
+						       &config_obj, false);
+			if (err) {
+				goto cleanup;
+			}
+			/* The config object now belongs to the location request object */
+			config_obj.json = NULL;
```
jayteemo, line 1019: this line can be removed, `nrf_cloud_obj_object_add` resets it.

```
/**
 * @brief Add a key string and object to the provided object.
 *
 * @details If successful, obj_to_add will be reset with @ref nrf_cloud_obj_reset
 *          since its data has been moved to obj.
```

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -980,11 +981,48 @@ int nrf_cloud_obj_location_request_create(struct nrf_cloud_obj *const obj,
...
+		} else {
+			(void)nrf_cloud_obj_free(&config_obj);
```
jayteemo, line 1021: this else can be removed (assuming you init the config_obj inside the if statement).

subsys/net/lib/nrf_cloud/src/nrf_cloud_rest.c
```diff
@@ -561,16 +580,16 @@ int nrf_cloud_rest_location_get(struct nrf_cloud_rest_context *const rest_ctx,
 	req.body = payload_obj.encoded_data.ptr;
```
glarsennordic, line 580: I am surprised that `do_reply`, `fallback` and `hi_conf` are not sent in the body. Is this imposed by cloud? If not, perhaps `nrf_cloud_obj_location_request_payload_add` can be updated to accept these parameters, instead of encoding them as query strings in the URL?

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -980,11 +981,48 @@ int nrf_cloud_obj_location_request_create(struct nrf_cloud_obj *const obj,
 		goto cleanup;
 	}
 
-	/* By default, nRF Cloud will send the location to the device */
-	if (!request_loc &&
-	    nrf_cloud_obj_num_add(&data_obj, NRF_CLOUD_LOCATION_KEY_DOREPLY, 0, false)) {
+	if (config && (obj->type == NRF_CLOUD_OBJ_TYPE_JSON)) {
+		/* Assume no differences from defaults. */
```
jayteemo, line 985: i believe `config_obj` needs to be initialized.

```        
        err = nrf_cloud_obj_init(&config_obj);
	if (err) {
		goto cleanup;
	}
```

glarsennordic (commented):
Mostly LGTM but the URL stuff is at least worth discussing

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -980,11 +981,48 @@ int nrf_cloud_obj_location_request_create(struct nrf_cloud_obj *const obj,
 		goto cleanup;
 	}
 
-	/* By default, nRF Cloud will send the location to the device */
-	if (!request_loc &&
-	    nrf_cloud_obj_num_add(&data_obj, NRF_CLOUD_LOCATION_KEY_DOREPLY, 0, false)) {
+	if (config && (obj->type == NRF_CLOUD_OBJ_TYPE_JSON)) {
```
jayteemo, line 984: looking at this function, i suppose it doesn't make sense to continue if the type is not JSON, since that is the only one that is currently supported.
can you add a:
```
if (obj->type != NRF_CLOUD_OBJ_TYPE_JSON) {
		ret = -ENOTSUP;
}
```
to the beginning, and then this JSON check and the original switch statement below can be removed.

subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
```diff
@@ -292,8 +292,35 @@ static void copy_wifi_info(struct wifi_ob *wifi_encode,
...
 
+static bool copy_config(struct config *config_encode,
+			struct nrf_cloud_location_config const *const conf)
+{
+	bool needed = false;
+
+	/** Check each field and compare to the cloud's default. If different,
+	 *  enable output of the field and set the new value of it.
+	 */
+	if (conf->do_reply != true) {
```
jayteemo, line 303: can you make defines for the cloud default values?
to be used here and in `nrf_cloud_obj_location_request_create` and in `nrf_cloud_rest_location_get`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
jayteemo: it looks like there is an errant blank line in the header comments for `nrf_cloud_obj_object_add`... could you please remove it 😄

subsys/net/lib/nrf_cloud/src/nrf_cloud_rest.c
plskeggs: This is the REST API doc for it:
https://api.nrfcloud.com/v1/#tag/Ground-Fix/operation/GetLocationFromCellsOrWifiNetworks

subsys/net/lib/nrf_cloud/src/nrf_cloud_rest.c
plskeggs: See the link above.

include/modem/location.h
trantanen: Replying quickly so you can still respond today...

> Excluding these configuration options from the location library does not work for us, since we changed our multi service sample to use the location library, as did others for ATv2. You are basically saying we have feature-frozen the positioning services as far as our samples are concerned.

I know this is a little complicated and would be nice to have all kinds of parameters in Location library (and we've discussed about many possible params related to cloud and other functionality such as NCELLMEAS search type) but:
- ATv2 is the only application in our SDK that uses Location library (I made the integration) and that's not a problem because it uses `CONFIG_LOCATION_SERVICE_EXTERNAL` and can hence send nRF Cloud request in the format it wants to.
- Having cloud service support in Location library has been questioned. There hasn't been serious discussions about removing it and I've always wanted to have it for quick testing and prototyping purposes. If the user wants to do something more, `CONFIG_LOCATION_SERVICE_EXTERNAL` is there.
- Multiservice sample could use `CONFIG_LOCATION_SERVICE_EXTERNAL` too instead of `CONFIG_LOCATION_SERVICE_NRF_CLOUD` if there are some cloud related functionalities not supported in Location library. But let's close this discussion first before rushing into it.
- If we would add these new cloud specific detailed parameters, I'd say there are probably dozen or two other parameters that are equally nice or important to have in the API. So this is why we try to carefully select the parameters we add to the API. And there are some that could probably be removed based on how things have evolved since the first versions of the API.

> Excluding these options also means you will not be able to exercise them from mosh. I would have thought you would be curious and interested to see how they work and want to try them.

We'll discuss this within our team but:
- In the past year or so, we've mainly sent direct REST requests to cloud through scripts using pre-collected NCELLMEAS results as an input. This way we can exercise more data points than using MOSH on a live device.
- We can still use any nRF Cloud Library API parameters through the `CONFIG_LOCATION_SERVICE_EXTERNAL` configuration, for which you also modified some default values.

include/net/nrf_cloud_location.h
trantanen: I guess we don't know from the returned result whether fallback was used, right? But it would be visible as a higher accuracy value?

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap
trantanen: Good!

include/net/nrf_cloud_location.h
plskeggs: Correct, the returned result does not specify that a fallback to a more coarse position was done, but if fallback occurs, the position will be *lower* accuracy (location based only on TAC instead of one or more specific cell towers).

glarsennordic (approved):

subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
jayteemo, line 303:
```suggestion
	if (conf->do_reply != NRF_CLOUD_LOCATION_DOREPLY_DEFAULT) {
```

subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
```diff
@@ -292,8 +292,35 @@ static void copy_wifi_info(struct wifi_ob *wifi_encode,
...
+		config_encode->_config_doReply_present = true;
+		config_encode->_config_doReply._config_doReply = conf->do_reply;
+		needed = true;
+	}
+	if (conf->fallback != true) {
+		config_encode->_config_fallback_present = true;
+		config_encode->_config_fallback._config_fallback = conf->fallback;
+		needed = true;
+	}
+	if (conf->hi_conf != false) {
```
jayteemo, line 313:
```suggestion
	if (conf->hi_conf != NRF_CLOUD_LOCATION_HICONF_DEFAULT) {
```

subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
```diff
@@ -292,8 +292,35 @@ static void copy_wifi_info(struct wifi_ob *wifi_encode,
...
+
+	/** Check each field and compare to the cloud's default. If different,
+	 *  enable output of the field and set the new value of it.
+	 */
+	if (conf->do_reply != true) {
+		config_encode->_config_doReply_present = true;
+		config_encode->_config_doReply._config_doReply = conf->do_reply;
+		needed = true;
+	}
+	if (conf->fallback != true) {
```
jayteemo, line 308:
```suggestion
	if (conf->fallback != NRF_CLOUD_LOCATION_FALLBACK_DEFAULT) {
```

subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
plskeggs: good catch

samples/cellular/modem_shell/src/location/location_srv_ext_nrf_cloud.c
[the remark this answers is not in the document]
tokangas: I didn't have a look at the nRF Cloud library implementation, but my opinion is, that if the request should be optimized (default parameters are left out), it should be done inside the nRF Cloud library. Any other components shouldn't need to be aware of cloud default values.

peknis (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_rest.c
```diff
@@ -517,7 +518,25 @@ int nrf_cloud_rest_location_get(struct nrf_cloud_rest_context *const rest_ctx,
...
 
-	req.url = request->disable_response ? API_GET_LOCATION_NO_REPLY : API_GET_LOCATION;
+	char url[sizeof(API_GET_LOCATION_CONFIG) + sizeof("false") * 3];
+	bool do_reply;
+
+	if (request->config) {
+		ret = snprintk(url, sizeof(url) - 1, API_GET_LOCATION_CONFIG,
+			       request->config->do_reply ? "true" : "false",
+			       request->config->fallback ? "true" : "false",
+			       request->config->hi_conf  ? "true" : "false");
```
trantanen, line 528: If we wanted to optimize the size of the REST URL, you could add just those params that are non-default. I'm not sure you want to go there though...

samples/cellular/modem_shell/src/location/location_srv_ext_nrf_cloud.c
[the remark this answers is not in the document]
tokangas: Yes, passing parameters does cost something. But the check whether to pass the parameters or not probably costs more, unless it can be optimized out by the compiler.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -651,11 +656,13 @@ Libraries for networking
     * An :c:struct:`nrf_cloud_obj_shadow_data` structure to the :c:struct:`nrf_cloud_evt` structure to be used during shadow update events.
     * The :kconfig:option:`CONFIG_NRF_CLOUD_SEND_SERVICE_INFO_FOTA` Kconfig option to enable sending configured FOTA service info on the device's initial connection to nRF Cloud.
     * The :kconfig:option:`CONFIG_NRF_CLOUD_SEND_SERVICE_INFO_UI` Kconfig option to enable sending configured UI service info on the device's initial connection to nRF Cloud.
+    * An :c:struct:`nrf_cloud_location_config` structure for specifying the desired behavior of an nRF Cloud ground-fix request.
```
trantanen, line 659: We should mention somewhere that `bool request_loc` parameter has been converted to the new structure. This is important because it's backwards incompatible change.

subsys/net/lib/nrf_cloud/src/nrf_cloud_rest.c
```diff
@@ -517,7 +518,25 @@ int nrf_cloud_rest_location_get(struct nrf_cloud_rest_context *const rest_ctx,
...
+		if ((ret < 0) || (ret > (sizeof(url) - 1))) {
+			ret = -EIO;
+			goto clean_up;
+		}
+		do_reply = request->config->do_reply;
+	} else {
+		strncpy(url, API_GET_LOCATION, sizeof(url) - 1);
+		url[sizeof(url) - 1] = '\0';
+		do_reply = NRF_CLOUD_LOCATION_DOREPLY_DEFAULT;
```
trantanen, line 537: How about handling here cloud default values so the application doesn't need to think about whether to pass values (which might be cloud defaults) or NULL config?
It would resolve this also https://github.com/nrfconnect/sdk-nrf/pull/13417#discussion_r1441651810

subsys/net/lib/nrf_cloud/src/nrf_cloud_rest.c
plskeggs: Yes, I drew the line at that point.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
plskeggs: Added to the line that describes updates to the two functions that used to take `bool request_loc`:

>     * The :c:func:`nrf_cloud_obj_location_request_create` and :c:func:`nrf_cloud_location_request` functions to accept the :c:struct:`nrf_cloud_location_config` structure in place of the ``bool request_loc`` parameter.

samples/cellular/modem_shell/src/location/location_srv_ext_nrf_cloud.c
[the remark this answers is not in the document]
plskeggs: I was speaking of data costs and radio power costs accumulated over time, not CPU clock cycle costs.

MarkusLassila (approved):

include/modem/location.h
plskeggs: I've removed changes to the location library.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
trantanen: Was this free needed? I guess it's needed when there are no errors up to this point. Another free call if there is an error later in the function will occur but I guess that doesn't matter...

include/net/nrf_cloud_location.h
```diff
@@ -57,6 +57,19 @@ struct nrf_cloud_location_result {
...
 };
 
+/** @brief Location request config */
+struct nrf_cloud_location_config {
+	/** If true, nRF Cloud will send the location to the device. */
+	bool do_reply;
+	/** If true, uncertainty of result will be 95%, otherwise 68%. */
+	bool hi_conf;
+	/** If true, and location cannot be found, result will fall back to rough estimate
+	 *  based on the cell tower tracking area code.  If false, will return an error
```
trantanen, line 67:
```suggestion
	 *  based on the cell tower tracking area code. If false, will return an error
```

include/net/nrf_cloud_location.h
```diff
@@ -92,8 +105,8 @@ typedef void (*nrf_cloud_location_response_t)(const struct nrf_cloud_location_re
  *                 To omit a request parameter, use the appropriate
  *                 `NRF_CLOUD_LOCATION_WIFI_OMIT_` define.
  *                 Can be NULL if cell info is provided.
- * @param request_loc If true, cloud will send location to the device.
- *                    If false, cloud will not send location to the device.
+ * @param config   Optional configuration of request. If NULL, fall back to default
+ *                 which is do_reply = true, hi_conf = false, and fallback = true.
  * @param cb Callback function to receive parsed location result. Only used when
  *           request_loc is true. If NULL, JSON result will be sent to the cloud event
```
trantanen, line 111: `request_loc` is not there anymore so need to refer to the `config->do_reply` etc. and the comment should take care of config=NULL case too which means this callback would be valid.

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c
```diff
@@ -415,7 +415,10 @@ int nrf_cloud_coap_location_get(struct nrf_cloud_rest_location_request const *co
 	size_t len = sizeof(buffer);
 	int err;
 
-	err = coap_codec_ground_fix_req_encode(request->cell_info, request->wifi_info, buffer, &len,
+	err = coap_codec_ground_fix_req_encode(request->cell_info,
+					       request->wifi_info,
+					       request->config,
+					       buffer, &len,
```
trantanen, line 421: `coap_codec.c` changes are likely missing from the PR. CI also fails because of that now.

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap
```diff
@@ -37,6 +37,15 @@ config NON_RESP_RETRIES
 	  result in up to this number of retransmissions of the request followed
 	  by waits for a response.
 
+config NRF_CLOUD_COAP_GF_CONF
```
trantanen, line 40: This is not used anywhere at the moment but I assume it would be in `coap_codec.c`

plskeggs:
I pushed before I completed some changes to the CoAP support.  The new parameters need to be handled differently.

________________________________

include/net/nrf_cloud_location.h
plskeggs: The comment does take care of config=NULL.
```
* @param config   Optional configuration of request. If NULL, fall back to default
 *                 which is do_reply = true, hi_conf = false, and fallback = true.

Comment (plskeggs) [reply include/net/nrf_cloud_location.h:111]: OK, missed that.

File: subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c (hunk 30 above)
Comment (plskeggs) [reply subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c:421]: Premature push.

File: subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap (hunk 31 above)
Comment (plskeggs) [reply subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap:40]: Premature push.

File: subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c (hunk 18 above)
Comment (plskeggs) [reply subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c:1029]: I set `err = -ENOMEM` before checking the 3 flags to see if they differ from the default (I basically use `err` for two purposes).

If any of them differ, I add JSON bools for them and set the values accordingly.  The act of adding sets the `err` variable to 0 when successful; when non-zero, we bail out to `cleanup`.

If none of them differ, `err == -ENOMEM` still, which indicates that `config_obj` was not needed after all, so I delete it.

This works fine but is obscure. I'll make it clearer.

File: subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c (hunk 30 above)
Comment (plskeggs) [reply subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c]: CoAP passing of config flags needs to be via a query string for now. Refactored all this and added a central function to generate the URI for both CoAP and REST.

File: samples/cellular/nrf_cloud_rest_cell_location/README.rst [hunk 32]
```diff
@@ -62,6 +62,8 @@ You only need to do this once for each device.
 
 After the sample completes any requested JITP or shadow updates, pressing the **button 1** toggles between single-cell and multi-cell mode.
 
+Set the :kconfig:`CONFIG_REST_CELL_CHANGE_CONFIG` config value to try all combinations of the :c:struct:`nrf_cloud_location_config` structure values ``hi_conf`` and ``fallback``.
```
peknis, line 65:
```suggestion
Set the :ref:`CONFIG_REST_CELL_CHANGE_CONFIG <CONFIG_REST_CELL_CHANGE_CONFIG>` Kconfig config value to try all combinations of the :c:struct:`nrf_cloud_location_config` structure values ``hi_conf`` and ``fallback``.
```

samples/cellular/nrf_cloud_rest_cell_location/README.rst
```diff
@@ -77,6 +79,22 @@ Check and configure the following Kconfig options for the sample:
 CONFIG_REST_CELL_LOCATION_DO_JITP - Enable prompt to perform JITP via REST
    This configuration option defines whether the application prompts the user for just-in-time provisioning on startup.
 
+.. _CONFIG_REST_CELL_CHANGE_CONFIG:
+
+CONFIG_REST_CELL_CHANGE_CONFIG - Enable changing request configuration
+   Set this to use the next combination of hi_conf and fallback flags after performing both single- and multi-cell location requests.
```
peknis, line 85:
```suggestion
   Set this to use the next combination of ``hi_conf`` and ``fallback`` flags after performing both single and multi-cell location requests.
```

subsys/net/lib/nrf_cloud/include/nrf_cloud_codec_internal.h
```diff
@@ -395,6 +395,14 @@ int nrf_cloud_pgps_req_data_json_encode(const struct gps_pgps_request *const req
 int nrf_cloud_json_to_url_params_convert(char *const buf, const size_t buf_size,
 					 const cJSON * const obj);
 
+/** @brief Generate URL for ground-fix request.
+ *  Return size of required buffer if buf = NULL, otherwise generate and store
+ *  url into buf based on config, and return 0 on no error or negative error code.
```
peknis, line 400:
```suggestion
 *  URL into buf based on config, and return 0 on no error or negative error code.
```

peknis (changes requested):

peknis (approved):

samples/cellular/nrf_cloud_rest_cell_location/README.rst
```diff
@@ -77,6 +79,22 @@ Check and configure the following Kconfig options for the sample:
...
+.. _CONFIG_REST_CELL_DEFAULT_HICONF_VAL:
+
+CONFIG_REST_CELL_DEFAULT_HICONF_VAL - Enable high confidence result
+   Enable a 95% confidence interval for the location, instead of the default 68%.
+
+.. _CONFIG_REST_CELL_DEFAULT_FALLBACK_VAL:
+
+CONFIG_REST_CELL_DEFAULT_FALLBACK_VAL - Enable fallback to coarse location
+   If enabled and the location of the cell tower or Wi-Fi access points cannot be found, return area-level location based on the cellular tracking area code.
+   Otherwise a 404 error will be returned indicating location is not known.
```
trantanen, line 96: Not sure I would mention 404 as that's not visible to the user of the sample, right?

```suggestion
   Otherwise an error will be returned indicating location is not known.
```

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
trantanen: > Another free call if there is an error later in the function will occur but I guess that doesn't matter...

Is double free an issue? It can happen when we go to cleanup from line 1041 or 1047.

Looking at `nrf_cloud_obj_free()`, I guess you can call it twice for the same object and the underlying JSON etc. data is only freed on the first instance. The second call will simply return `-ENOTSUP`, right?

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap
trantanen: Still cannot find this used anywhere...

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -3713,6 +3713,64 @@ int nrf_cloud_json_to_url_params_convert(char *const buf, const size_t buf_size,
 	return 0;
 }
 
+int nrf_cloud_ground_fix_url_encode(char *buf, size_t size, const char *base,
+				    const struct nrf_cloud_location_config *config)
+{
+	#define ADJ(ptr, sz) do { \
```
jayteemo, line 3719: a comment for this macro would be nice.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
```diff
@@ -980,40 +984,74 @@ int nrf_cloud_obj_location_request_create(struct nrf_cloud_obj *const obj,
...
-		err = -ENOMEM;
-		goto cleanup;
-	}
-
-	switch (obj->type) {
-	case NRF_CLOUD_OBJ_TYPE_JSON:
-	{
-		/* Add cell/wifi info */
-		err = nrf_cloud_obj_location_request_payload_add(&data_obj, cells_inf, wifi_inf);
+	if (config) {
```
jayteemo, line 987: if you want to avoid the change_needed flag and potential init and free of an unnecessary object:

```suggestion
	if (config && 
	    ((config->do_reply != NRF_CLOUD_LOCATION_DOREPLY_DEFAULT) || 
	     (config->hi_conf != NRF_CLOUD_LOCATION_HICONF_DEFAULT) || 
	     (config->fallback != NRF_CLOUD_LOCATION_FALLBACK_DEFAULT))) {
```

jayteemo (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec.c
plskeggs: Yes, I know, and I didn't do it that way to make the function slightly less ugly. But I'll go ahead and add it.

PR #13421, opened 2023-12-13
applications: serial_lte_modem: first half implementation of PPP

tomi-font:
This adds basic PPP functionality to SLM.
As of these changes, PPP is started manually using AT#XPPP=1.
Its overlay files configure it to use uart1 at a baud rate of 1M.


applications/serial_lte_modem/src/slm_ppp.c
```diff
@@ -0,0 +1,311 @@
...
+		}
+	} else if (!*addr6) {
+		LOG_ERR("No connectivity.");
+		return false;
+	}
+	if (*addr6) {
+		struct in6_addr in6;
+		struct ipv6cp_options *const opts = &ctx->ipv6cp.my_options;
+
+		if (inet_pton(AF_INET6, addr6, &in6) != 1)
```
rlubos, line 106: Missing `{ }`

applications/serial_lte_modem/src/slm_ppp.c
```diff
@@ -0,0 +1,311 @@
...
+
+	for (size_t i = 0; i != ARRAY_SIZE(fds); ++i) {
+		fds[i].fd = ppp_fds[i];
+		fds[i].events = POLLIN;
+	}
+
+	while (true) {
+		const int poll_ret = poll(fds, ARRAY_SIZE(fds), 1000);
+
+		if (poll_ret == 0) {
```
rlubos, line 256: If no action is taken in case of no events, maybe just set the poll timeout to `-1` (infinite)?

applications/serial_lte_modem/src/slm_ppp.c
```diff
@@ -0,0 +1,311 @@
...
+	net_if_flag_set(ppp_iface, NET_IF_POINTOPOINT);
+
+	LOG_DBG("PPP initialized.");
+	return 0;
+}
+
+/* Handles AT#XPPP commands. */
+int handle_at_ppp(enum at_cmd_type cmd_type)
+{
+	if (cmd_type != AT_CMD_TYPE_SET_COMMAND
```
rlubos, line 216: I'd move the check after the variable declarations, for me it looks a bit awkward that way.

applications/serial_lte_modem/src/slm_ppp.c
tomi-font: True.

SeppoTakalo (approved):

applications/serial_lte_modem/src/slm_ppp.c
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	ret = pdn_dynamic_params_get(PDP_CID, &ctx->ipcp.my_options.dns1_address,
+				     &ctx->ipcp.my_options.dns2_address, &mtu);
+	if (ret) {
+		return ret;
+	}
+
+	/* Set the PPP MTU to that of the LTE link. */
+	mtu = MIN(mtu, sizeof(ppp_data_buf));
```
MarkusLassila, line 136: Just a note that in some networks (either Elisa or DNA sim that I have), the MTU returned by CGCONTRDP was problematic when trying to send full size UDP packages. The actual transmittable package was much less than what this returned (headers were taken into account).

applications/serial_lte_modem/src/slm_ppp.c
```diff
@@ -0,0 +1,309 @@
...
+			return;
+		}
+
+		for (size_t src = 0; src != ARRAY_SIZE(fds); ++src) {
+			const short revents = fds[src].revents;
+
+			if (!revents) {
+				continue;
+			}
+			if (!(revents & POLLIN)) {
```
MarkusLassila, line 269: How does this work if we have both POLLIN and POLLERR? I think that it then does only latter part and misses the POLLERR. Does the next poll then catch the error?

MarkusLassila (commented):
Looks good, couple of clarifications.

applications/serial_lte_modem/src/slm_ppp.c
MarkusLassila: This is the only one that I am really interested in this point.

tomi-font: The way I understand it is that they cannot happen at the same time.
From `man poll`:
> The bits returned in revents can include any of those specified in events, or one of the values POLLERR, POLLHUP, or POLLNVAL.

applications/serial_lte_modem/src/slm_ppp.c
tomi-font: Interesting. Do you recall the numbers? We might have to make the MTU configurable at some point, or something of the sort.

applications/serial_lte_modem/src/slm_ppp.c
MarkusLassila: At least with TCP, it is possible to receive data in connection closure.

applications/serial_lte_modem/src/slm_ppp.c
```diff
@@ -0,0 +1,309 @@
...
+}
+
+/* Handles AT#XPPP commands. */
+int handle_at_ppp(enum at_cmd_type cmd_type)
+{
+	int ret;
+	unsigned int op;
+	enum {
+		OP_STOP,
+		OP_RESTART,
```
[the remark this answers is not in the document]
tomi-font: It's meant as a start-or-restart operation.
When auto-connect is enabled and PPP is started automatically on network registration (which is coming up), it is effectively a restart from the first time.

applications/serial_lte_modem/src/slm_ppp.c
MarkusLassila: Unfortunately not and anyway it differed between operators.

applications/serial_lte_modem/src/slm_ppp.c
tomi-font: Here we are dealing with raw sockets, with one end being the modem (not the network) and the other end being the PPP interface. From testing those simply return error events when they are shut down, and I don't think any of them would ever send any data along.
I still updated the condition to take this possibility into account.

MarkusLassila: My point was to handle the possible POLLIN event before handling the error events. Now the logic will miss the possible data from POLLIN, should it be possible to receive that with errors.

tomi-font: Oh well. Then the original implementation most likely handled it the way you wanted. If there were any errors along with a `POLLIN`, then I'm pretty sure it would instantly be received another time on the next loop iteration.

MarkusLassila: It will likely receive another error when calling poll. I still think that it would be better to switch the order that you process the data in here, to first check the POLLIN and doing the data processing part and then doing the error handling part. but I leave it up to you. Especially since this will be continued.

MarkusLassila (approved):

applications/serial_lte_modem/src/slm_ppp.c
tomi-font: I consider it to be pretty solid as it is now, so will leave it like that. Switching the order would require another level of indentation, which would require quite some lines to be further split into several lines.

PR #13425, opened 2023-12-13
Fix sysbuild showing child/parent config options and add alternative zephyr test configuration

nordicjm:
sysbuild: Disable child image Kconfigs when PM is disabled
    
    Prevents PM being enabled when it should be disabled in zephyr
    tests by disabling child image Kconfigs that enable the PM
    Kconfig option

    sysbuild: Hide child/parent Kconfig options
    
    Disables configuration options for child/parent images

    sysbuild: Do not set signing script if PM is disabled
    
    Leave the default zephyr signing script if partition manager is
    not enabled when using sysbuild

    alt-test-config: Add alternative zephyr test configuration
    
    Adds alternative test configuration files that will be used
    instead of the zephyr ones to ensure partition manager is
    disabled. Base files taken from zephyr commit
    23cf38934c0f68861e403b22bc3dd0ce6efbfa39

Hiding child/parent Kconfig will not work as requires commit from ongoing upmerge in sdk-zephyr

### DIFF
--- a/sysbuild/CMakeLists.txt
+++ b/sysbuild/CMakeLists.txt
@@ -184,8 +184,10 @@ function(${SYSBUILD_CURRENT_MODULE_NAME}_pre_cmake)
       )
     endif()
 
-    # Use NCS signing script with support for PM
-    set(${DEFAULT_IMAGE}_SIGNING_SCRIPT "${ZEPHYR_NRF_MODULE_DIR}/cmake/sysbuild/image_signing.cmake" CACHE INTERNAL "MCUboot signing script" FORCE)
+    if(SB_CONFIG_PARTITION_MANAGER)
+      # Use NCS signing script with support for PM
+      set(${DEFAULT_IMAGE}_SIGNING_SCRIPT "${ZEPHYR_NRF_MODULE_DIR}/cmake/sysbuild/image_signing.cmake" CACHE INTERNAL "MCUboot signing script" FORCE)
+    endif()
 
     set(imgtool_extra)
     if(SB_CONFIG_MCUBOOT_HARDWARE_DOWNGRADE_PREVENTION)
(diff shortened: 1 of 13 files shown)


sysbuild/CMakeLists.txt
```diff
@@ -68,6 +68,16 @@ function(${SYSBUILD_CURRENT_MODULE_NAME}_pre_cmake)
   cmake_parse_arguments(PRE_CMAKE "" "" "IMAGES" ${ARGN})
   restore_ncs_vars()
 
+  if(NOT SB_CONFIG_PARTITION_MANAGER)
+    # Disable any Kconfigs for child images in the main application
+    foreach(config NCS_SAMPLE_EMPTY_APP_CORE_CHILD_IMAGE NCS_INCLUDE_RPMSG_CHILD_IMAGE NCS_SAMPLE_REMOTE_SHELL_CHILD_IMAGE
```
tejlmand, line 73: isn't some of this handled through e35d9e2d94e8bcda399568819a47bc6e70a6fd2a ?

afaik the `BUILD_STRATEGY` settings will not be available with linked sha.

tejlmand (approved):

sysbuild/CMakeLists.txt
nordicjm: Yes, this was an earlier try to get them unselected, could trim some of the options

nordicjm: Fixed

PerMac:
@nordicjm @tejlmand 
I am playing with this --alt-test-config but I couldn't find an easy solution to make it work in our CI without reworking the feature itself. The implementation of this feature requires quite strict usage, and is based on how twister resolves paths, which is screwed for non-zephyr repos.
What about another approach: maybe this SB_CONFIG_PARTITION_MANAGER=n can be applied as an overlay to the tests in need? Instead of full copies of yamls, add a commit adding required overlays  as a noup patch to sdk-zephyr? I've heard that a noup path modifying those yamls in a "no go" due to "upmerge hell" those tends to create (although I am not convinced in this case and I find maintaining duplicates more risky) maybe a noup with overlays won't cause such issues?

PR #13427, opened 2023-12-13
tests: samples: Add platform_allow section to configs without it

PerMac:
Entries in platform_allow are copies of integration_platforms. This change is needed for dynamic scope optimization for twister. A test_plan.py script selects which configurations to add to the scope. If a test is modified, it is scoped for all applicable platforms. This means for cases in the commit, selecting >600 platforms available in the upstream and nrf. Platform_allow works as a filter limiting the scope.

### DIFF
--- a/samples/app_event_manager/sample.yaml
+++ b/samples/app_event_manager/sample.yaml
@@ -15,6 +15,14 @@ common:
 tests:
   sample.app_event_manager:
     build_only: false
+    platform_allow:
+      - qemu_cortex_m3
+      - nrf52dk_nrf52832
+      - nrf52840dk_nrf52840
+      - nrf5340dk_nrf5340_cpuapp
+      - nrf5340dk_nrf5340_cpuapp_ns
+      - nrf9160dk_nrf9160_ns
+      - nrf21540dk_nrf52840
     integration_platforms:
       - qemu_cortex_m3
       - nrf52dk_nrf52832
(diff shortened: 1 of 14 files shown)


MarekPieta (approved):

samples/nrf_profiler/sample.yaml
```diff
@@ -9,5 +9,10 @@ tests:
       - nrf52840dk_nrf52840
       - nrf9160dk_nrf9160_ns
       - nrf21540dk_nrf52840
+    platform_allow:
+      - nrf52dk_nrf52832
+      - nrf52840dk_nrf52840
+      - nrf9160dk_nrf9160_ns
+      - nrf21540dk_nrf52840
```
alstrzebonski, line 16: Everywhere else `platform_allow` is above `integration_platforms`. Here, order is reversed.

tests/unity/example_test/testcase.yaml
```diff
@@ -3,3 +3,5 @@ tests:
     tags: example
     integration_platforms:
       - native_posix
+    platform_allow:
+      - native_posix
```
alstrzebonski, line 7: Same here

alstrzebonski (approved):

samples/app_event_manager/sample.yaml
```diff
@@ -15,6 +15,14 @@ common:
 tests:
   sample.app_event_manager:
     build_only: false
+    platform_allow:
```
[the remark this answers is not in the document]
PerMac: This new format was added recently in the upstream, to keep consistency with integration_platforms. Now both string are sequence are allowed. Here I copied integration_platforms and changed the key to platform_allow. I think your suggestion is legit, but it is very low on priority list:)

samples/nrf_profiler/sample.yaml
PerMac: good catch:) my mistake. there is no required order anyhow nor I see a point in recommending any. I see it is not a blocker for you. Therefore, I won't change it since it affects nothing and would just cause CI to restart and dismiss reviews.

tests/unity/example_test/testcase.yaml
PerMac: explained above

PR #13434, opened 2023-12-14
secure_storage: Fix counter bug

Vge0rge:
The counter used as a nonce for the secure storage was 16 bytes but the implementation on the AEAD backend expected a 12 byte nonce. This corrupted the key material when the nonce was written.
This fixes the issue by making the counter to be 12 bytes as expected by the AEAD backend. It also ensures the AEAD backend and the counter have the same length and fails at runtime without corrupting any buffers if they differ.

### DIFF
--- a/subsys/secure_storage/src/aead/aead_ctr_nonce.c
+++ b/subsys/secure_storage/src/aead/aead_ctr_nonce.c
@@ -18,15 +18,12 @@
 #include "aead_nonce.h"
 
 /*
- * Nonce is a 128 bits counter.
+ * Nonce is a 96 bits counter.
  */
-
-#define NONCE_MAX_LENGTH 16
-
 static struct aead_ctr_nonce {
 	uint64_t low;
-	uint64_t high;
-} g_nonce;
+	uint32_t high;
+} __packed g_nonce;
 
 /* Return an incrementing nonce */
 psa_status_t secure_storage_get_nonce(uint8_t *nonce, size_t nonce_len)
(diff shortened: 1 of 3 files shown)


subsys/secure_storage/src/aead/aead_ctr_nonce.c
```diff
@@ -35,7 +32,7 @@ psa_status_t secure_storage_get_nonce(uint8_t *nonce, size_t nonce_len)
 		return PSA_ERROR_INVALID_ARGUMENT;
 	}
 
-	if (nonce_len > NONCE_MAX_LENGTH) {
+	if (nonce_len != sizeof(g_nonce)) {
 		return PSA_ERROR_NOT_SUPPORTED;
 	}
```
vili-nordic, line 38: comment on line 43 is wrong

Vge0rge: true! fixed!

subsys/secure_storage/src/aead/aead_ctr_nonce.c
```diff
@@ -35,24 +32,23 @@ psa_status_t secure_storage_get_nonce(uint8_t *nonce, size_t nonce_len)
...
 
 	if (nonce_len == 0) {
 		return PSA_SUCCESS;
 	}
 
-	/* Incrementing is implemented by using 2 64bit numbers */
+	/* Incrementing is implemented by using a 64bit an a 32bit number */
```
frkv, line 43:
```suggestion
	/* Incrementing is implemented by using a 64-bit and a 32-bit number */
```

maciejbaczmanski (approved):
Tested on OT CLI samples on nrf52 and nrf53, Issue with asserts after board restart does not reproduce anymore. Thanks!

ArekBalysNordic (approved):
It repairs the issue with resetting nRF53/52 DKs also in Matter samples when PSA crypto is enabled.

vili-nordic (approved):

PR #11593, opened 2023-06-21
samples: openthread: align serial transport configuration 

maciejbaczmanski:
After changes in Zephyr, `nrf52840dongle_nrf52840` uses CDC ACM USB
transport by default, so building for this target should not require
providing additional snippets/ configuration files.

Additionally, remove `CONFIG_USB_DEVICE_INITIALIZE_AT_BOOT=n` from
usb configuration files, as they are no longer required after https://github.com/zephyrproject-rtos/zephyr/pull/58060

PR #13442, opened 2023-12-14
doc: openthread: Update Thread tools documentation.

canisLupus1313:
Commit fixes:
	- `ot-daemon` invoke command
	- change default `ot-daemon` build command to build with newest thread version.

### DIFF
--- a/doc/nrf/protocols/thread/tools.rst
+++ b/doc/nrf/protocols/thread/tools.rst
@@ -281,13 +281,13 @@ Build the OpenThread POSIX applications by performing the following steps:
       ./script/bootstrap
 
 #. Build the applications with the required options.
-   For example, to build the ``ot-cli`` application with support for Thread v1.1, run the following command::
+   For example, to build the ``ot-cli`` application, run the following command::
 
-      ./script/cmake-build posix -DOT_THREAD_VERSION=1.1
+      ./script/cmake-build posix
 
-   Alternatively, to build the ``ot-daemon`` and ``ot-ctl`` applications with support for Thread v1.2, run the following command::
+   Alternatively, to build the ``ot-daemon`` and ``ot-ctl`` applications, run the following command::
 
-      ./script/cmake-build posix -DOT_THREAD_VERSION=1.2 -DOT_DAEMON=ON
+      ./script/cmake-build posix -DOT_DAEMON=ON
 
 You can find the generated applications in :file:`./build/posix/src/posix/`.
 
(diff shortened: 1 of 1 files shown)


doc/nrf/protocols/thread/tools.rst
```diff
@@ -87,7 +87,7 @@ To program the nRF device with the RCP application, complete the following steps
 
          .. code-block:: console
 
-            west build -p always -b nrf52840dongle_nrf52840 nrf/samples/openthread/coprocessor/
+            west build -p always -b nrf52840dongle_nrf52840 nrf/samples/openthread/coprocessor/ -S usb
```
maciejbaczmanski, line 90: after https://github.com/nrfconnect/sdk-nrf/pull/11593 `usb` snippet should not be required for `nrf52840dongle_nrf52840` build target. Does the sample break without it?

doc/nrf/protocols/thread/tools.rst
```diff
@@ -281,13 +281,13 @@ Build the OpenThread POSIX applications by performing the following steps:
       ./script/bootstrap
 
 #. Build the applications with the required options.
-   For example, to build the ``ot-cli`` application with support for Thread v1.1, run the following command::
+   For example, to build the ``ot-cli`` application with support for Thread v1.2, run the following command::
```
maciejbaczmanski, line 284: this part sounds a bit confusing to me. Maybe we could split this into 3 cases explaining default build, changing versions and enabling ot-daemon?
- `./script/cmake-build posix`
- `./script/cmake-build posix -DOT_THREAD_VERSION=1.2`
- `./script/cmake-build posix -DOT_DAEMON=ON`

doc/nrf/protocols/thread/tools.rst
```diff
@@ -281,13 +281,13 @@ Build the OpenThread POSIX applications by performing the following steps:
...
 
-      ./script/cmake-build posix -DOT_THREAD_VERSION=1.1
+      ./script/cmake-build posix -DOT_THREAD_VERSION=1.2
 
-   Alternatively, to build the ``ot-daemon`` and ``ot-ctl`` applications with support for Thread v1.2, run the following command::
+   Alternatively, to build the ``ot-daemon`` and ``ot-ctl`` applications with support for Thread, run the following command::
```
wiba-nordic, line 288: This is something that was here previously already, but is the change from 'ot-cli' to 'ot-ctl' intentional? Anything isn't specified in either example command.

Does support for Thread need to be mentioned if no version is specified?

doc/nrf/protocols/thread/tools.rst
canisLupus1313: Let's remove info about building other versions of thread since we tend to support only the newest default version. for this we also generate libraries.

doc/nrf/protocols/thread/tools.rst
canisLupus1313: @wiba-nordic The statement that we can build thread for version 1.2 is kind of obsolete since the newest is now 1.3. We are also providing usually only libs for the newest version of thread. So instructing users to build thread for 1.2 is confusing.

wiba-nordic (approved):

edmont (approved):

PR #13449, opened 2023-12-15
wifi: Basic AP mode support

krish2718:
Beaconing only


modules/hostap/src/supp_api.c
```diff
@@ -398,14 +380,47 @@ int z_wpa_supplicant_connect(const struct device *dev,
...
+			   "Interface %s is down, dropping connect",
+			   dev->name);
+		return -1;
+	}
+
+	k_mutex_lock(&wpa_supplicant_mutex, K_FOREVER);
+
+	wpa_s = get_wpa_s_handle(dev);
+	if (!wpa_s) {
+		ret = -1;
```
rado17, line 452: Can we set `ret` to `-ENODEV` instead of `-1`?

drivers/wifi/nrf700x/src/net_if.c
```diff
@@ -38,6 +38,35 @@ static struct net_mgmt_event_callback ip_maddr4_cb;
...
+		goto out;
+	}
+
+	if (!event) {
+		LOG_ERR("%s: event is NULL",
+			__func__);
+		goto out;
+	}
+
+	(void)event_len;
```
rado17, line 59: What does this do?

modules/hostap/src/supp_api.c
krish2718: Sure, I guess we need to fix this at all places, is it okay if I do the cleanup together in another PR?

drivers/wifi/nrf700x/src/net_if.c
krish2718: Tells the compiler that `not using this variable is intended and not a bug` and then we don't see `unused variable` warning.

modules/hostap/src/supp_api.c
rado17: Sure.

rado17 (approved):

PR #13453, opened 2023-12-15
applications: serial_lte_modem: second half implementation of PPP

tomi-font:
This adds auto-start functionality so that the PPP link goes up/down when the network connection goes up/down.

Also, auto-restart is performed when the PPP peer disconnects so that reconnection can happen successfully.

Documentation for the #XPPP AT command is added. The rest is not yet documented as this is still a work in progress.

### DIFF
--- a/applications/serial_lte_modem/doc/PPP_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/PPP_AT_commands.rst
@@ -1,0 +1,95 @@
+.. _SLM_AT_PPP:
+
+PPP AT commands
+****************
+
+.. contents::
+   :local:
+   :depth: 2
+
+This page describes PPP-related AT commands.
+
+Control PPP #XPPP
+=================
+
+Set command
+-----------
+
+The set command allows you to start and stop PPP.
+
+Syntax
+~~~~~~
+
+::
+
+   #XPPP=<op>
+
+* The ``<op>`` parameter can be the following:
+
+  * ``0`` - Stop PPP.
+  * ``1`` - Start PPP.
+    If PPP is already started, it is first stopped before being started again.
+
+.. note::
+
+   PPP is automatically started and stopped when the default PDN connection is established and lost, respectively.
+   This happens even if PPP has previously been stopped or started with this command.
+
+Read command
+------------
+
+The read command allows you to get the status of PPP.
+
+Syntax
+~~~~~~
+
+::
+
+   #XPPP?
+
+Response syntax
+~~~~~~~~~~~~~~~
+
+::
+
+   #XPPP: <running>,<peer_connected>
+
+* The ``<running>`` parameter is an integer that indicates whether PPP is running.
+  It is ``1`` for running or ``0`` for stopped.
+
+* The ``<peer_connected>`` parameter is an integer that indicates whether a peer is connected to PPP.
+  It is ``1`` for connected or ``0`` for not connected.
+
+Test command
+------------
+
+The test command is not supported.
+
+Example
+-------
+
+::
+
+  AT+CFUN=1
+
+  OK
+  // Wait for network registration. PPP is automatically started.
+  AT#XPPP?
+
+  #XPPP: 1,0
+
+  OK
+  // Have the peer connect to SLM's PPP. The read command shows that a peer is connected.
+  AT#XPPP?
+
+  #XPPP: 1,1
+
+  OK
+  AT+CFUN=4
+
+  OK
+  AT#XPPP?
+
+  #XPPP: 0,0
+
+  OK
--- a/applications/serial_lte_modem/src/slm_at_host.c
+++ b/applications/serial_lte_modem/src/slm_at_host.c
@@ -658,6 +658,14 @@ AT_MONITOR(at_notify, ANY, notification_handler);
 static void notification_handler(const char *notification)
 {
 	if (get_slm_mode() == SLM_AT_COMMAND_MODE) {
+
+#if defined(CONFIG_SLM_PPP)
+		if (!slm_fwd_cgev_notifs
+		 && !strncmp(notification, "+CGEV: ", strlen("+CGEV: "))) {
+			/* CGEV notifications are silenced. Do not forward them. */
+			return;
+		}
+#endif
 		(void)slm_uart_tx_write(CRLF_STR, strlen(CRLF_STR), true, true);
 		(void)slm_uart_tx_write(notification, strlen(notification), true, false);
 	}
--- a/applications/serial_lte_modem/src/slm_ppp.c
+++ b/applications/serial_lte_modem/src/slm_ppp.c
@@ -7,6 +7,8 @@
 #include "slm_ppp.h"
 #include "slm_at_host.h"
 #include "slm_util.h"
+#include <modem/at_cmd_custom.h>
+#include <modem/lte_lc.h>
 #include <modem/pdn.h>
 #include <zephyr/net/net_if.h>
 #include <zephyr/net/ppp.h>
@@ -14,6 +16,12 @@
 #include <assert.h>
 
 LOG_MODULE_REGISTER(slm_ppp, CONFIG_SLM_LOG_LEVEL);
+
+/* This keeps track of whether the user is registered to the CGEV notifications.
+ * We need them to know when to start/stop the PPP link, but that should not
+ * influence what the user receives, so we do the filtering based on this.
+ */
+bool slm_fwd_cgev_notifs;
 
 static const struct device *ppp_uart_dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_ppp_uart));
 static int ppp_iface_idx;
@@ -25,6 +33,10 @@ static struct sockaddr_ll ppp_zephyr_dst_addr;
 static struct k_thread ppp_data_passing_thread_id;
 static K_THREAD_STACK_DEFINE(ppp_data_passing_thread_stack, KB(2));
 static void ppp_data_passing_thread(void*, void*, void*);
+
+static struct k_work ppp_restart_work;
+
+static bool ppp_peer_connected;
 
 /* We use only the default PDP context. */
 enum { PDP_CID = 0 };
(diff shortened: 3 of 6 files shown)


applications/serial_lte_modem/src/slm_at_host.c
```diff
@@ -658,6 +658,14 @@ AT_MONITOR(at_notify, ANY, notification_handler);
...
 {
 	if (get_slm_mode() == SLM_AT_COMMAND_MODE) {
+
+#if defined(CONFIG_SLM_PPP)
+		if (!slm_fwd_cgev_notifs
+		 && !strncmp(notification, "+CGEV: ", strlen("+CGEV: "))) {
+			/* CGEV notifications are silenced. Do not forward them. */
+			return;
+		}
+#endif
```
SeppoTakalo, line 668:
```suggestion
		if (IS_ENABLED(CONFIG_SLM_PPP) && !slm_fwd_cgev_notifs
		    && !strncmp(notification, "+CGEV: ", strlen("+CGEV: "))) {
			/* CGEV notifications are silenced. Do not forward them. */
			return;
		}
```

I would prefer `IS_ENABLED()` over `#if defined` on a code that can actually compile when the feature is disabled. Compiler will then still do the syntax checking on the disabled part, even when its going to be finally optimized away.

SeppoTakalo (approved):

applications/serial_lte_modem/src/slm_at_host.c
tomi-font: True. Though in that case the variable is defined in `slm_ppp.h` that is only included `#if defined(CONFIG_SLM_PPP)`.

tomi-font:
> /home/runner/work/sdk-nrf/sdk-nrf/ncs/nrf/doc/_build/nrf/src/applications/serial_lte_modem/doc/PPP_AT_commands.rst:30: ERROR: Unexpected indentation.

@wiba-nordic I fail to understand how [line 30](https://github.com/nrfconnect/sdk-nrf/pull/13453/files#diff-e7ca8cde48588741afafd50071684296e5ff8847c04730ddd426a4553afd35e6R30) has the wrong indentation?

applications/serial_lte_modem/src/slm_at_host.c
SeppoTakalo: You can move the variable as well. As it is unused, compiler should drop it.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,94 @@
...
+Syntax
+~~~~~~
+
+::
+
+   #XPPP=<op>
+
+* The ``<op>`` parameter can be the following:
+  * ``0`` - Stop PPP.
+  * ``1`` - (Re)start PPP.
```
wiba-nordic, line 29:
```suggestion
  * ``1`` - Start PPP.
```

applications/serial_lte_modem/src/slm_at_host.c
tomi-font: Yeah, though I have a comment on its definition that pertains to PPP...

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,94 @@
...
+   #XPPP?
+
+Response syntax
+~~~~~~~~~~~~~~~
+
+::
+
+   #XPPP: <running>,<peer_connected>
+
+* The ``<running>`` parameter tells whether PPP is running.
```
wiba-nordic, line 56:
```suggestion
* The ``<running>`` parameter is an integer that indicates whether PPP is running.
```

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,94 @@
...
+  It is ``1`` when PPP is running and ``0`` when PPP is stopped.
+
+* The ``<peer_connected>`` parameter tells whether a peer is connected to PPP.
```
wiba-nordic, line 59:
```suggestion
* The ``<peer_connected>`` parameter is an integer that indicates whether a peer is connected to PPP.
```

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Response syntax
+~~~~~~~~~~~~~~~
+
+::
+
+   #XPPP: <running>,<peer_connected>
+
+* The ``<running>`` parameter tells whether PPP is running.
+  It is ``1`` when PPP is running and ``0`` when PPP is stopped.
```
wiba-nordic, line 57:
```suggestion
  It is ``1`` for running or ``0`` for stopped.
```

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,94 @@
...
+
+* The ``<peer_connected>`` parameter tells whether a peer is connected to PPP.
+  It is ``1`` when a peer is connected and ``0`` when there is no peer connected.
```
wiba-nordic, line 60:
```suggestion
  It is ``1`` for connected or ``0`` for not connected.
```

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Syntax
+~~~~~~
+
+::
+
+   #XPPP=<op>
+
+* The ``<op>`` parameter can be the following:
+  * ``0`` - Stop PPP.
```
wiba-nordic, line 28:
```suggestion
* The ``<op>`` parameter can be the following:

  * ``0`` - Stop PPP.
```
This is where the actual error is.

tomi-font: Ah! Thanks.

wiba-nordic (approved):

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,95 @@
+.. _SLM_AT_PPP:
+
```
MarkusLassila, line 2: CGEREP special handling needs to be documented.

applications/serial_lte_modem/src/slm_ppp.c
```diff
@@ -192,8 +204,154 @@ static void slm_ppp_stop(void)
...
+	}
+}
+
+/* We need to receive CGEV notifications at all times.
+ * CGEREP AT commands are intercepted to prevent the user
+ * from unsubcribing us and make that behavior invisible.
+ */
+AT_CMD_CUSTOM(at_cgerep_interceptor, "AT+CGEREP", at_cgerep_callback);
+
+static int at_cgerep_callback(char *buf, size_t len, char *at_cmd)
```
MarkusLassila, line 235: This is kind of cute way of handling this.

MarkusLassila (commented):
Looks good. Couple of comments.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,95 @@
...
+  AT+CFUN=1
+
+  OK
+  // Wait for network registration. PPP is automatically started.
+  AT#XPPP?
+
+  #XPPP: 1,0
+
+  OK
+  // Connect to PPP. The read command shows that a peer is connected.
```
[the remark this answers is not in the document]
tomi-font: From SLM's perspective yes, but from the user doing this not necessarily, e.g. here you would start `pppd`.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
tomi-font: Why do you think so? Everything is made so that the behavior is not changed for the end user and it's all invisible.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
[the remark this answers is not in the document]
MarkusLassila: Should this example then include the AT#PPP=1?

applications/serial_lte_modem/doc/PPP_AT_commands.rst
MarkusLassila: The events are not really stopped. Take for example the case where there is another processor for these events. This will be very hard to investigate if there is no mention of it. Same goes for the errors that are sent from the AT-command versions which you ignore. If there is no mention of special handling of these commands anywhere, it will be problematic to investigate.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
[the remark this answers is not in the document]
tomi-font: No. The start happens indeed automatically on the SLM side with the changes from this PR. PPP will get auto-started when the network registration happens, which in this example is between `AT+CFUN=1` and the first `AT#XPPP?`.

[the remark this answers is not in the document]
MarkusLassila: OK, so it is as I first thought. "Connect to PPP." to me reads as an action that you would do. So that is why I thought that this was missing a command, with that phrasing.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
tomi-font: What potential other processor for these events are you thinking of?

Regarding the alternative AT command syntax, okay. My reasoning was that these should not be allowed because of how weird they are, but it's actually a deviation from the modem behavior. The code could be updated to accept those.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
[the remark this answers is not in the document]
tomi-font: I mean yes, that is what I was subtly trying to communicate. At that point you may want to start `pppd`, for example. But that may need to be made more explicit. How about something like below?
```suggestion
  // Have the peer connect to SLM's PPP. The read command shows that a peer is connected.
```

applications/serial_lte_modem/doc/PPP_AT_commands.rst
MarkusLassila: Anything that subscribes to AT_MONITOR(at_notify, ANY, notification_handler) (or specific events).

Rather than update the event, I think that it would be enough to document that we have special handling for AT+CGEREP and +CGEV, in which SLM intercepts and resends the commands.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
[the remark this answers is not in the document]
MarkusLassila: Ah yes, forgot the starting of the pppd at this point and thus had problems in understanding this. This is definitively better.

Not on the scope for this, but should we document the pppd command as an example? It would aid in trying this out with Linux, but then there is the question of Windows usage...

rlubos (approved):

applications/serial_lte_modem/doc/PPP_AT_commands.rst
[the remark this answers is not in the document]
tomi-font: Regarding additional documentation, yes definitely. This is planned.

applications/serial_lte_modem/doc/PPP_AT_commands.rst
MarkusLassila: I think that the command would be possible from modules other than SLM, not sure how likely that is. But you are correct that then we are on the developer side.

If you can get the error cases to work (or at least log indicative errors from them), then this will not necessarily need to be indicated in the documentation.

applications/serial_lte_modem/src/slm_ppp.c
```diff
@@ -192,8 +204,154 @@ static void slm_ppp_stop(void)
...
+{
+	int ret;
+	unsigned int subscribe;
+	const bool set_cmd = (sscanf(at_cmd, "AT+CGEREP=%u", &subscribe) == 1);
+
+	if (!set_cmd && strcmp(at_cmd, "AT+CGEREP?") && strcmp(at_cmd, "AT+CGEREP=?")) {
+		/* The modem interprets AT+CGEREP and AT+CGEREP= as AT+CGEREP=0.
+		 * Prevent those forms, only allowing AT+CGEREP=0, for simplicty.
+		 */
+		return -EINVAL;
```
MarkusLassila, line 245: At minimum, give some indication why the commands are not succeeding.

MarkusLassila (approved):

applications/serial_lte_modem/doc/PPP_AT_commands.rst
```diff
@@ -0,0 +1,95 @@
...
+  AT+CFUN=1
+
+  OK
+  // Wait for network registration. PPP is automatically started.
+  AT#XPPP?
+
+  #XPPP: 1,0
+
+  OK
+  // Have the peer connect to SLM's to PPP. The read command shows that a peer is connected.
```
wiba-nordic, line 82: SLM's what?

tomi-font: Woops. SLM's ~~to~~ PPP.

wiba-nordic (approved):

PR #13456, opened 2023-12-15
applications: nrf5340_audio: Disable RTC pretick workaround for B0N

andvib:
Disable the RTC pretick anomaly workaround for B0N. The workaround causes a resource conflict between B0N and the controller, causing the controller to never start advertising. As the controller does not use RTC to wake up from sleep it is not affected by the anomaly.

### DIFF
--- a/applications/nrf5340_audio/CMakeLists.txt
+++ b/applications/nrf5340_audio/CMakeLists.txt
@@ -31,6 +31,12 @@ if ((CONFIG_AUDIO_DFU EQUAL 1) OR (CONFIG_AUDIO_DFU EQUAL 2))
     list(APPEND empty_net_core_OVERLAY_CONFIG
         "${CMAKE_CURRENT_LIST_DIR}/dfu/conf/overlay-empty_net_core.conf"
     )
+
+    # Disable pretick anomaly workaround. The workaround causes a resource conflict between B0N and
+    # the controller, and as the controller does not use RTC to wake up from sleep it is not
+    # affected by the anomaly.
+    list(APPEND empty_net_core_b0n_CONFIG_SOC_NRF53_RTC_PRETICK n)
+
     if (CONFIG_B0N_MINIMAL)
         set(min_b0n_flag "-m")
         list(APPEND empty_net_core_b0n_OVERLAY_CONFIG overlay-minimal-size.conf)


applications/nrf5340_audio/CMakeLists.txt
```diff
@@ -31,6 +31,12 @@ if ((CONFIG_AUDIO_DFU EQUAL 1) OR (CONFIG_AUDIO_DFU EQUAL 2))
     list(APPEND empty_net_core_OVERLAY_CONFIG
         "${CMAKE_CURRENT_LIST_DIR}/dfu/conf/overlay-empty_net_core.conf"
     )
+
+    # Disable pretick anomaly workaround. The workaround causes a resource conflict between B0N and
+    # the controller, and as the controller does not use RTC to wake up from sleep it is not
+    # affected by the anomaly.
+    list(APPEND empty_net_core_b0n_CONFIG_SOC_NRF53_RTC_PRETICK n)
+
```
cvinayak, line 39: I am assuming here we are addressing one particular controller implementation. Should we take care now if conditionally disabling depending on `CONFIG_BT_LL_ACS_NRF53` ?

cvinayak (changes requested):

applications/nrf5340_audio/CMakeLists.txt
erikrobstad: This is a good point, but it is actually only here (when DFU is enabled) that changing this config has any effect.
In "non-DFU mode" we do not add the B0N bootloader, and we are running the controller directly on the NET core. It is only when the B0N bootloader is added that this config have an effect.

andvib: Makes sense to me - we would then just wrap the `list()` call in an `if(CONFIG_BT_LL_ACS_NRF53)`? That can be done here AFAICT

cvinayak: > I prefer we move on as it is,

Fine by me.

cvinayak (approved):

Frodevan (approved):
Passes DFU tests on main

applications/nrf5340_audio/CMakeLists.txt
erikrobstad: Ah seems like I misread/misunderstood the question the first time.
For this 2.5.1 release I guess it is sufficient to have DFU functionality only for BT_LL_ACS_NRF53.

PR #13445, opened 2023-12-14
samples: cellular: Remove nrf7002ek positioning overlay

trantanen:
The shield for nrf7002ek disables UART1 so overlays for nrf7002ek Wi-Fi positioning are not needed.

PR #13471, opened 2023-12-19
applications: asset_tracker_v2: Fix requested data in sample request

tokangas:
Removed location from sample request when the Location module is disabled. Earlier also location was requested and because of this the sending of data was delayed until sampling timed out.

Removed battery and environmental data sampling when Sensor module is disabled.

Improved handling of case where all configured location methods are disabled using the run-time configuration.

Added missing Kconfig option for Wi-Fi sampling default value.

### DIFF
--- a/applications/asset_tracker_v2/src/main.c
+++ b/applications/asset_tracker_v2/src/main.c
@@ -381,14 +381,17 @@ static void data_get(void)
 
 	/* Specify which data that is to be included in the transmission. */
 	app_module_event->data_list[count++] = APP_DATA_MODEM_DYNAMIC;
-	app_module_event->data_list[count++] = APP_DATA_BATTERY;
-	app_module_event->data_list[count++] = APP_DATA_ENVIRONMENTAL;
-
 	if (!modem_static_sampled) {
 		app_module_event->data_list[count++] = APP_DATA_MODEM_STATIC;
 	}
 
-	if (!app_cfg.no_data.neighbor_cell || !app_cfg.no_data.gnss || !app_cfg.no_data.wifi) {
+	if (IS_ENABLED(CONFIG_SENSOR_MODULE)) {
+		app_module_event->data_list[count++] = APP_DATA_BATTERY;
+		app_module_event->data_list[count++] = APP_DATA_ENVIRONMENTAL;
+	}
+
+	if (IS_ENABLED(CONFIG_LOCATION_MODULE) &&
+	    (!app_cfg.no_data.neighbor_cell || !app_cfg.no_data.gnss || !app_cfg.no_data.wifi)) {
 		app_module_event->data_list[count++] = APP_DATA_LOCATION;
 
 		/* Set application module timeout when location sampling is requested.
--- a/applications/asset_tracker_v2/src/modules/Kconfig.data_module
+++ b/applications/asset_tracker_v2/src/modules/Kconfig.data_module
@@ -169,6 +169,14 @@ config DATA_SAMPLE_NEIGHBOR_CELLS_DEFAULT
 	  data in sample requests sent to other modules. This configuration can be overwritten by
 	  changing the application's real-time configuration using the cloud-side state.
 
+config DATA_SAMPLE_WIFI_DEFAULT
+	bool "Include Wi-Fi APs in sample requests"
+	default y
+	help
+	  If this configuration is enabled the application will by default include Wi-Fi AP data in
+	  sample requests sent to other modules. This configuration can be overwritten by changing
+	  the application's real-time configuration using the cloud-side state.
+
 config DATA_GRANT_SEND_ON_CONNECTION_QUALITY
 	bool "Grant or deny encoding and sending of data based on LTE connection quality"
 	select EXPERIMENTAL
--- a/applications/asset_tracker_v2/src/modules/data_module.c
+++ b/applications/asset_tracker_v2/src/modules/data_module.c
@@ -94,7 +94,8 @@ static struct cloud_data_cfg current_cfg = {
 	.accelerometer_inactivity_threshold	= CONFIG_DATA_ACCELEROMETER_INACT_THRESHOLD,
 	.accelerometer_inactivity_timeout	= CONFIG_DATA_ACCELEROMETER_INACT_TIMEOUT_SECONDS,
 	.no_data.gnss		 = !IS_ENABLED(CONFIG_DATA_SAMPLE_GNSS_DEFAULT),
-	.no_data.neighbor_cell	 = !IS_ENABLED(CONFIG_DATA_SAMPLE_NEIGHBOR_CELLS_DEFAULT)
+	.no_data.neighbor_cell	 = !IS_ENABLED(CONFIG_DATA_SAMPLE_NEIGHBOR_CELLS_DEFAULT),
+	.no_data.wifi		 = !IS_ENABLED(CONFIG_DATA_SAMPLE_WIFI_DEFAULT)
 };
 
 static struct k_work_delayable data_send_work;
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -168,7 +168,9 @@ This section provides detailed lists of changes by :ref:`application <applicatio
 Asset Tracker v2
 ----------------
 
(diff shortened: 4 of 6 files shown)


gregersrygg (approved):

applications/asset_tracker_v2/src/main.c
```diff
@@ -388,7 +388,8 @@ static void data_get(void)
 		app_module_event->data_list[count++] = APP_DATA_MODEM_STATIC;
 	}
 
-	if (!app_cfg.no_data.neighbor_cell || !app_cfg.no_data.gnss || !app_cfg.no_data.wifi) {
+	if (IS_ENABLED(CONFIG_LOCATION_MODULE) &&
```
trantanen, line 391: @gregersrygg Do we have similar issue with `APP_DATA_ENVIRONMENTAL` and `CONFIG_SENSOR_MODULE`?

applications/asset_tracker_v2/src/modules/Kconfig.data_module
```diff
@@ -169,6 +169,14 @@ config DATA_SAMPLE_NEIGHBOR_CELLS_DEFAULT
 	  data in sample requests sent to other modules. This configuration can be overwritten by
 	  changing the application's real-time configuration using the cloud-side state.
 
+config DATA_SAMPLE_WIFI_DEFAULT
+	bool "Include Wi-Fi APs in sample requests"
+	default y
```
trantanen, line 174: This is problematic to be `y` by default and should likely be `n` because WiFi is not always configured into the build. We should likely set this on only when WiFi is enabled in the build. Otherwise we request `APP_DATA_LOCATION` with non-WiFi configuration even if GNSS and neighbor_cell are disabled.

applications/asset_tracker_v2/src/modules/data_module.c
```diff
@@ -94,7 +94,8 @@ static struct cloud_data_cfg current_cfg = {
 	.accelerometer_inactivity_threshold	= CONFIG_DATA_ACCELEROMETER_INACT_THRESHOLD,
 	.accelerometer_inactivity_timeout	= CONFIG_DATA_ACCELEROMETER_INACT_TIMEOUT_SECONDS,
 	.no_data.gnss		 = !IS_ENABLED(CONFIG_DATA_SAMPLE_GNSS_DEFAULT),
-	.no_data.neighbor_cell	 = !IS_ENABLED(CONFIG_DATA_SAMPLE_NEIGHBOR_CELLS_DEFAULT)
+	.no_data.neighbor_cell	 = !IS_ENABLED(CONFIG_DATA_SAMPLE_NEIGHBOR_CELLS_DEFAULT),
+	.no_data.wifi		 = !IS_ENABLED(CONFIG_DATA_SAMPLE_WIFI_DEFAULT)
```
trantanen, line 98: This `.no_data.wifi` line is something we should add to fix a bug with WiFi build.
If discussions lead to removal of `CONFIG_DATA_SAMPLE_WIFI_DEFAULT` from this PR, I guess this could be:
`.no_data.wifi = !IS_ENABLED(CONFIG_LOCATION_METHOD_WIFI)`

applications/asset_tracker_v2/src/modules/Kconfig.data_module
```diff
@@ -169,6 +169,14 @@ config DATA_SAMPLE_NEIGHBOR_CELLS_DEFAULT
 	  data in sample requests sent to other modules. This configuration can be overwritten by
 	  changing the application's real-time configuration using the cloud-side state.
 
+config DATA_SAMPLE_WIFI_DEFAULT
```
trantanen, line 172: This has been left out intentionally because WiFi sampling has been kind of assumed to be wanted when WiFi is built into the image. So it's different in nature compared to GNSS and neighbor_cell which are always available in the build.

applications/asset_tracker_v2/src/main.c
```diff
@@ -388,7 +388,8 @@ static void data_get(void)
...
+	    (!app_cfg.no_data.neighbor_cell || !app_cfg.no_data.gnss || !app_cfg.no_data.wifi)) {
```
trantanen, line 392: The idea has been to set the `CONFIG_DATA_SAMPLE_*` Kconfig options to `n` so this has worked.
Except there's been a bug with WiFi as there wasn't a configuration for that so that should be fixed anyway.

trantanen (changes requested):
Let's talk offline on how to deal with this.

applications/asset_tracker_v2/src/main.c
trantanen: So we'll keep IS_ENABLED(CONFIG_LOCATION_MODULE) here because you could easily enable some method from cloud and the build wouldn't support location stuff. Also, this is easier to read from the code.

applications/asset_tracker_v2/src/modules/data_module.c
trantanen: By default without setting specific configurations, we should leave `.no_data.wifi` false just like other `no_data` variables.
The main reason for this is that if we set this to true, it will be written to cloud and when you build and flash WiFi config, WiFi positioning will be unintentionally disabled in the cloud.

applications/asset_tracker_v2/src/modules/Kconfig.data_module
trantanen: This will be handled elsewhere in `location_module.c:start_search()`

applications/asset_tracker_v2/src/main.c
tokangas: Yes, if Sensor module is disabled, battery and environmental data should not be sampled. I added flagging for these as well.

applications/asset_tracker_v2/src/modules/Kconfig.data_module
trantanen: Let's add it for consistency although it's quite useless and similar Kconfig's for GNSs and cellular could be even removed.

trantanen: This KConfig has to be added to document though.

trantanen (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -168,7 +168,9 @@ This section provides detailed lists of changes by :ref:`application <applicatio
 Asset Tracker v2
 ----------------
 
-  * Removed nRF7002 EK DTC overlay file ``nrf91xxdk_with_nrf7002ek.overlay`` because UART1 is disabled through the shield configuration.
+* Added the :kconfig:option:`CONFIG_DATA_SAMPLE_WIFI_DEFAULT` Kconfig option to configure if Wi-Fi APs are included in sample requests by default.
+
+* Removed nRF7002 EK DTC overlay file ``nrf91xxdk_with_nrf7002ek.overlay`` because UART1 is disabled through the shield configuration.
```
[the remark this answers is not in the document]
trantanen: This sentence came in from PR https://github.com/nrfconnect/sdk-nrf/pull/13445 and is in 3 other places of the release note.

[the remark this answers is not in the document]
trantanen: And do we want `:file:` tag for a file that's removed from the repository?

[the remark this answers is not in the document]
peknis: Yep, I was sloppy with the previous review, sorry. The :file: tag does not really have much functionality and does not have a big impact on anything. It is just for formatting and double backticks would do the same trick.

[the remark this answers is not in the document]
tokangas: This PR only changed the indentation, otherwise this line is unchanged. But we can of course do some fine tuning, if necessary. 😄

simensrostad (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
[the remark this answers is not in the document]
trantanen: > Yep, I was sloppy with the previous review, sorry. The :file: tag does not really have much functionality and does not have a big impact on anything. It is just for formatting and double backticks would do the same trick.

Ok, I was thinking about :file: tag originally but didn't do it since the doc generation results haven't been properly available although they should be now.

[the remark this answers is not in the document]
tokangas: Updated all four occurrences.

peknis (approved):

PR #13474, opened 2023-12-19
samples: matter: common initialization module

markaj-nordic:
* Added common initialization module for all Matter samples to reduce maintenance effort and simplify code
* Documented Matter::Init module API
* Adapted all Matter samples and applications
* Cleaned up includes

- [x] Fix CI builds
- [x] Adapt Weather Station application
- [x] Make the Matter::Init module more flexible by extending InitData with other Matter interfaces
- [x] Test more thoroughly (commissioned weather station to Google ecosystem, tested multiple build configurations, integration passed, tested Matter Bridge BLE connections and loading of stored devices after reboot)
- [x] Cleanup namespaces
- [x] Avoid instantiating unused default static objects
- [x] Stress test with nightly suite (nighty plan is green: https://jenkins-ncs.nordicsemi.no/view/z_CHIP/job/latest/job/acc/job/test-fw-nrfconnect-chip/job/master/87/)

### DIFF
--- a/applications/matter_bridge/src/app_task.cpp
+++ b/applications/matter_bridge/src/app_task.cpp
@@ -5,10 +5,11 @@
  */
 
 #include "app_task.h"
-#include "task_executor.h"
+
 #include "bridge_manager.h"
 #include "bridge_storage_manager.h"
-#include "fabric_table_delegate.h"
+#include "matter_init.h"
+#include "task_executor.h"
 
 #ifdef CONFIG_BRIDGED_DEVICE_BT
 #include "ble_bridged_device_factory.h"
@@ -17,33 +18,21 @@
 #else
 #include "simulated_bridged_device_factory.h"
 #endif /* CONFIG_BRIDGED_DEVICE_BT */
-#include <platform/CHIPDeviceLayer.h>
 
 #include <app-common/zap-generated/ids/Attributes.h>
 #include <app-common/zap-generated/ids/Clusters.h>
 #include <app/server/OnboardingCodesUtil.h>
-#include <app/server/Server.h>
-#include <credentials/DeviceAttestationCredsProvider.h>
-#include <credentials/examples/DeviceAttestationCredsExample.h>
-#include <lib/support/CHIPMem.h>
-#include <lib/support/CodeUtils.h>
-#include <system/SystemError.h>
-
-#include <app/clusters/network-commissioning/network-commissioning.h>
-#include <platform/nrfconnect/wifi/NrfWiFiDriver.h>
 
 #ifdef CONFIG_CHIP_OTA_REQUESTOR
 #include "ota_util.h"
 #endif
 
-#include <zephyr/kernel.h>
 #include <zephyr/logging/log.h>
 
 LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
 
 using namespace ::chip;
 using namespace ::chip::app;
-using namespace ::chip::Credentials;
 using namespace ::chip::DeviceLayer;
 
 namespace
@@ -57,112 +46,7 @@ static constexpr uint8_t kUuidServicesNumber = ARRAY_SIZE(sUuidServices);
 
 } /* namespace */
 
-app::Clusters::NetworkCommissioning::Instance
-	sWiFiCommissioningInstance(0, &(NetworkCommissioning::NrfWiFiDriver::Instance()));
-
-CHIP_ERROR AppTask::Init()
-{
-	/* Initialize CHIP stack */
-	LOG_INF("Init CHIP stack");
-
-	CHIP_ERROR err = chip::Platform::MemoryInit();
-	if (err != CHIP_NO_ERROR) {
-		LOG_ERR("Platform::MemoryInit() failed");
-		return err;
-	}
-
-	err = PlatformMgr().InitChipStack();
-	if (err != CHIP_NO_ERROR) {
-		LOG_ERR("PlatformMgr().InitChipStack() failed");
-		return err;
-	}
-
-#if defined(CONFIG_CHIP_WIFI)
-	sWiFiCommissioningInstance.Init();
-#else
-	return CHIP_ERROR_INTERNAL;
-#endif /* CONFIG_CHIP_WIFI */
-
-	if (!GetBoard().Init()) {
-		LOG_ERR("User interface initialization failed.");
-		return CHIP_ERROR_INCORRECT_STATE;
-	}
-
-#ifdef CONFIG_CHIP_OTA_REQUESTOR
-	/* OTA image confirmation must be done before the factory data init. */
-	OtaConfirmNewImage();
-#endif
-
-#ifdef CONFIG_MCUMGR_TRANSPORT_BT
-	/* Initialize DFU over SMP */
-	GetDFUOverSMP().Init();
-	GetDFUOverSMP().ConfirmNewImage();
-#endif
-
-	/* Initialize CHIP server */
-#if CONFIG_CHIP_FACTORY_DATA
-	ReturnErrorOnFailure(mFactoryDataProvider.Init());
-	SetDeviceInstanceInfoProvider(&mFactoryDataProvider);
-	SetDeviceAttestationCredentialsProvider(&mFactoryDataProvider);
-	SetCommissionableDataProvider(&mFactoryDataProvider);
-#else
-	SetDeviceInstanceInfoProvider(&DeviceInstanceInfoProviderMgrImpl());
-	SetDeviceAttestationCredentialsProvider(Examples::GetExampleDACProvider());
-#endif
-
-	static chip::CommonCaseDeviceServerInitParams initParams;
-	(void)initParams.InitializeStaticResourcesBeforeServerInit();
-
-	ReturnErrorOnFailure(chip::Server::GetInstance().Init(initParams));
-	ConfigurationMgr().LogDeviceConfig();
-	PrintOnboardingCodes(chip::RendezvousInformationFlags(chip::RendezvousInformationFlag::kBLE));
-	AppFabricTableDelegate::Init();
-
-#ifdef CONFIG_BRIDGED_DEVICE_BT
-	/* Initialize BLE Connectivity Manager before the Bridge Manager, as it must be ready to recover devices loaded
-	 * from persistent storaged during the bridge init. */
-	err = BLEConnectivityManager::Instance().Init(sUuidServices, kUuidServicesNumber);
-	if (err != CHIP_NO_ERROR) {
-		LOG_ERR("BLEConnectivityManager initialization failed");
-		return err;
-	}
-#endif
-
-	/* Initialize bridge manager */
-	err = BridgeManager::Instance().Init(RestoreBridgedDevices);
-	if (err != CHIP_NO_ERROR) {
-		LOG_ERR("BridgeManager initialization failed");
-		return err;
-	}
-
-	/*
-	 * Add CHIP event handler and start CHIP thread.
-	 * Note that all the initialization code should happen prior to this point to avoid data races
-	 * between the main and the CHIP threads.
-	 */
-	PlatformMgr().AddEventHandler(ChipEventHandler, 0);
-
-	err = PlatformMgr().StartEventLoopTask();
-	if (err != CHIP_NO_ERROR) {
-		LOG_ERR("PlatformMgr().StartEventLoopTask() failed");
-		return err;
-	}
-
-	return CHIP_NO_ERROR;
-}
-
-CHIP_ERROR AppTask::StartApp()
-{
-	ReturnErrorOnFailure(Init());
-
-	while (true) {
-		TaskExecutor::DispatchNextTask();
-	}
-
-	return CHIP_NO_ERROR;
-}
-
-void AppTask::ChipEventHandler(const ChipDeviceEvent *event, intptr_t /* arg */)
+void AppTask::MatterEventHandler(const ChipDeviceEvent *event, intptr_t /* arg */)
 {
 	bool isNetworkProvisioned = false;
 
@@ -174,7 +58,8 @@ void AppTask::ChipEventHandler(const ChipDeviceEvent *event, intptr_t /* arg */)
 		break;
 #if defined(CONFIG_CHIP_WIFI)
 	case DeviceEventType::kWiFiConnectivityChange:
-		isNetworkProvisioned = ConnectivityMgr().IsWiFiStationProvisioned() && ConnectivityMgr().IsWiFiStationEnabled();
+		isNetworkProvisioned =
+			ConnectivityMgr().IsWiFiStationProvisioned() && ConnectivityMgr().IsWiFiStationEnabled();
 #if CONFIG_CHIP_OTA_REQUESTOR
 		if (event->WiFiConnectivityChange.Result == kConnectivity_Established) {
 			InitBasicOTARequestor();
@@ -242,5 +127,49 @@ CHIP_ERROR AppTask::RestoreBridgedDevices()
 							    chip::Optional<uint16_t>(endpointId));
 #endif
 	}
+	return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR AppTask::Init()
+{
+	/* Initialize Matter stack */
+	ReturnErrorOnFailure(Nordic::Matter::PrepareServer(
+		MatterEventHandler, Nordic::Matter::InitData{ .mPostServerInitClbk = [] {
+#ifdef CONFIG_BRIDGED_DEVICE_BT
+			/* Initialize BLE Connectivity Manager before the Bridge Manager, as it must be ready to recover
+			 * devices loaded from persistent storage during the bridge init. */
+			CHIP_ERROR bleInitError =
+				BLEConnectivityManager::Instance().Init(sUuidServices, kUuidServicesNumber);
+			if (bleInitError != CHIP_NO_ERROR) {
+				LOG_ERR("BLEConnectivityManager initialization failed");
+				return bleInitError;
+			}
+#endif
+
+			/* Initialize bridge manager */
+			CHIP_ERROR bridgeMgrInitError = BridgeManager::Instance().Init(RestoreBridgedDevices);
+			if (bridgeMgrInitError != CHIP_NO_ERROR) {
+				LOG_ERR("BridgeManager initialization failed");
+				return bridgeMgrInitError;
+			}
+			return CHIP_NO_ERROR;
+		} }));
+
+	if (!GetBoard().Init()) {
+		LOG_ERR("User interface initialization failed.");
+		return CHIP_ERROR_INCORRECT_STATE;
+	}
+
+	return Nordic::Matter::StartServer();
+}
+
+CHIP_ERROR AppTask::StartApp()
+{
+	ReturnErrorOnFailure(Init());
+
+	while (true) {
+		TaskExecutor::DispatchNextTask();
+	}
+
 	return CHIP_NO_ERROR;
 }
--- a/applications/matter_weather_station/src/app_task.cpp
+++ b/applications/matter_weather_station/src/app_task.cpp
@@ -5,44 +5,37 @@
  */
 
 #include "app_task.h"
+
 #include "board.h"
 #include "task_executor.h"
 
 #include "battery.h"
 #include "buzzer.h"
-#include "fabric_table_delegate.h"
 #include "led_widget.h"
-#include <platform/CHIPDeviceLayer.h>
+#include "matter_init.h"
 
 #include <DeviceInfoProviderImpl.h>
-
 #include <app-common/zap-generated/attributes/Accessors.h>
 #include <app-common/zap-generated/cluster-objects.h>
 #include <app/clusters/ota-requestor/OTATestEventTriggerDelegate.h>
 #include <app/server/OnboardingCodesUtil.h>
 #include <app/server/Server.h>
-#include <app/util/attribute-storage.h>
-#include <credentials/DeviceAttestationCredsProvider.h>
-#include <credentials/examples/DeviceAttestationCredsExample.h>
-
-#ifdef CONFIG_CHIP_WIFI
-#include <app/clusters/network-commissioning/network-commissioning.h>
-#include <platform/nrfconnect/wifi/NrfWiFiDriver.h>
-#endif
 
 #ifdef CONFIG_CHIP_OTA_REQUESTOR
 #include "ota_util.h"
 #endif
 
-#include <dk_buttons_and_leds.h>
+#ifdef CONFIG_MCUMGR_TRANSPORT_BT
+#include "dfu_over_smp.h"
+#endif
+
 #include <zephyr/drivers/sensor.h>
-#include <zephyr/kernel.h>
-#include <zephyr/logging/log.h>
 
 using namespace ::chip;
-using namespace ::chip::Credentials;
 using namespace ::chip::DeviceLayer;
 using namespace ::chip::app;
+
+#include <zephyr/logging/log.h>
 
 LOG_MODULE_DECLARE(app);
 
@@ -88,13 +81,7 @@ constexpr size_t kIdentifyTimerIntervalMs = 500;
 k_timer sMeasurementsTimer;
 k_timer sIdentifyTimer;
 
-/* NOTE! This key is for test/certification only and should not be available in production devices!
- * If CONFIG_CHIP_FACTORY_DATA is enabled, this value is read from the factory data.
- */
-uint8_t sTestEventTriggerEnableKey[TestEventTriggerDelegate::kEnableKeyLength] = { 0x00, 0x11, 0x22, 0x33, 0x44, 0x55,
-										   0x66, 0x77, 0x88, 0x99, 0xaa, 0xbb,
-										   0xcc, 0xdd, 0xee, 0xff };
-chip::DeviceLayer::DeviceInfoProviderImpl gExampleDeviceInfoProvider;
+const device *sBme688SensorDev = DEVICE_DT_GET_ONE(bosch_bme680);
 
 /* Add identify for all endpoints */
 Identify sIdentifyTemperature = { chip::EndpointId{ kTemperatureMeasurementEndpointId }, AppTask::OnIdentifyStart,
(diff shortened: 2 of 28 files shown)


samples/matter/common/src/matter_init.h
```diff
@@ -0,0 +1,83 @@
...
+#include <lib/core/Optional.h>
+#include <lib/support/Variant.h>
+#include <platform/PlatformManager.h>
+#if CONFIG_CHIP_FACTORY_DATA
+#include <platform/nrfconnect/FactoryDataProvider.h>
+#else
+#include <platform/nrfconnect/DeviceInstanceInfoProviderImpl.h>
+#endif
+
+namespace Nordic
```
Damian-Nordic, line 19: Should we have a task to move all common files to this namespace? I mean this should probably be unified, right?

Damian-Nordic (commented):
Some early comments/questions

samples/matter/common/src/matter_init.h
```diff
@@ -0,0 +1,83 @@
...
+{
+namespace Matter
+{
+	using ChipCustomInit = CHIP_ERROR (*)(void);
+	struct ChipInitData {
+		/** @brief Pointer to the user provided NetworkCommissioning instance. */
+		chip::app::Clusters::NetworkCommissioning::Instance *mNetworkingInstance{ nullptr };
+		/** @brief Optional custom initialization code that will be scheduled in the CHIP main event loop. */
+		ChipCustomInit mCustomInit{ nullptr };
+#if CONFIG_CHIP_FACTORY_DATA
```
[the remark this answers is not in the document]
markaj-nordic: Yes, I have this in mind. I guess weather station is the good example of customized application and it gives an idea what might be desired from customer's perspective. I will try to align to that while integrating WS (which is an early stage for now.)

applications/matter_weather_station/src/app_task.cpp
```diff
@@ -191,71 +167,26 @@ CHIP_ERROR AppTask::Init()
...
 	k_timer_init(
-		&sIdentifyTimer, [](k_timer *) { TaskExecutor::PostTask([] { IdentifyTimerHandler(); }); }, nullptr);
+		&sIdentifyTimer, [](k_timer *) {
+		TaskExecutor::PostTask([] { IdentifyTimerHandler(); }); }, nullptr);
 	k_timer_start(&sMeasurementsTimer, K_MSEC(kMeasurementsIntervalMs), K_MSEC(kMeasurementsIntervalMs));
 
 	/* Initialize CHIP server */
 	static chip::CommonCaseDeviceServerInitParams initParams;
-	static OTATestEventTriggerDelegate testEventTriggerDelegate{ ByteSpan(sTestEventTriggerEnableKey) };
+	// static OTATestEventTriggerDelegate testEventTriggerDelegate{ ByteSpan(sTestEventTriggerEnableKey) };
```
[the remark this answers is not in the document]
markaj-nordic: Yes, I've opened the PR already mainly to check if all build variants pass to spare manual verification.

samples/matter/common/src/matter_init.h
```diff
@@ -0,0 +1,83 @@
...
+#else
+#include <platform/nrfconnect/DeviceInstanceInfoProviderImpl.h>
+#endif
+
+namespace Nordic
+{
+namespace Matter
+{
+	using ChipCustomInit = CHIP_ERROR (*)(void);
+	struct ChipInitData {
```
kkasperczyk-no, line 24: If we already started using Matter name (e.g. in namespace), shouldn't we drop naming new structures with `Chip`?

samples/matter/lock/src/app_task.cpp
```diff
@@ -78,54 +57,19 @@ constexpr uint32_t kSwitchTransportTimeout = 10000;
...
-		return err;
-	}
-#elif defined(CONFIG_CHIP_WIFI)
-	sWiFiCommissioningInstance.Init();
+	/* In the Thread/Wi-Fi switchable mode, it is the ThreadWifiSwitch module that owns the NetworkCommissioning
+	   instances, so offload the initialization module from controlling that by explicitly setting the
+	   mNetworkingInstance parameter to nullptr. Otherwise, the initialization module instantiates the
+	   NetworkCommissioning object and handles it by default internally. */
+	ReturnErrorOnFailure(Nordic::Matter::InitChipServer(
+		ChipEventHandler, chip::MakeOptional<Nordic::Matter::ChipInitData>(
```
kkasperczyk-no, line 69: Is there a big value from making this argument optional instead of using a simple pointer with default value set to nullptr? I mean, we still need to do additional check (nullcheck vs Valid() check) and the mapping makes call looking much more complex.

samples/matter/common/src/matter_init.h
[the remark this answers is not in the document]
markaj-nordic: Regarding the switchable variant, this was already considered and solved by explicitly feeding ChipInitData::mNetworkingInstance=nullptr in the Nordic::Matter::InitChipServer call.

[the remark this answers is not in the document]
Damian-Nordic: Yeah, but it's assigned only if not null:
```
			if (customInitData.mNetworkingInstance) {
				sCommissioningInstance = customInitData.mNetworkingInstance;
			}
```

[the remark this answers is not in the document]
markaj-nordic: Ahh, right, thanks for the catch. The intention is to allow to explicitly register null networking instance (the sanity check is done later on anyway).

samples/matter/common/src/matter_init.cpp
```diff
@@ -0,0 +1,216 @@
...
+	VerifyInitResultOrReturn(sInitResult, "Cannot register CHIP event handler");
+
+	if (sCustomInit) {
+		sInitResult = sCustomInit();
+		VerifyInitResultOrReturn(sInitResult, "Custom initialization failed");
+	}
+
+	k_mutex_lock(&sInitMutex, K_FOREVER);
+	sInitDone = true;
+	k_condvar_signal(&sInitCondVar);
```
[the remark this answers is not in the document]
markaj-nordic: I added RAII object for that. Please re-check if you see any issues :)

samples/matter/common/src/matter_init.cpp
```diff
@@ -0,0 +1,230 @@
...
+	CHIP_ERROR InitChipServer(PlatformManager::EventHandlerFunct chipEventHandler, Optional<ChipInitData> initData)
+	{
+		if (initData.HasValue()) {
+			auto customInitData = initData.Value();
+			sCommissioningInstance = customInitData.mNetworkingInstance;
+			sCustomInit = customInitData.mCustomInit;
+			sServerInitParams = customInitData.mServerInitParams;
+#if CONFIG_CHIP_FACTORY_DATA
+			sFactoryDataProvider = customInitData.mFactoryDataProvider;
+#endif
```
ArekBalysNordic, line 195: Do we need custom init data at all? We didn't have it at all previously. I guess, customers don't need a custom implementation because they can just ignore calling the DoInitChipServer method in their apptask.cpp code and implement it on its own. Moreover, it makes the code more complicated because a customer needs to understand the whole "custom" dependencies first, and it makes our solution less attractive. If they saw only one method "InitChipServer" (located anywhere) they would have understood it without thinking about any other stuff.

samples/matter/lock/src/app_task.cpp
```diff
@@ -152,50 +100,9 @@ CHIP_ERROR AppTask::Init()
 	/* Initialize lock manager */
 	BoltLockMgr().Init(LockStateChanged);
 
-#ifdef CONFIG_CHIP_OTA_REQUESTOR
-	/* OTA image confirmation must be done before the factory data init. */
-	OtaConfirmNewImage();
-#endif
+	ReturnErrorOnFailure(Nordic::Matter::StartChipServer());
```
kkasperczyk-no, line 103: Could `StartChipServer` call `WaitForReadiness` internally? It's weird to me that user has to remember about calling some synchronization mechanisms of our module.

samples/matter/light_switch/src/app_task.cpp
```diff
@@ -60,106 +41,25 @@ bool sWasDimmerTriggered = false;
...
-	 * between the main and the CHIP threads.
-	 */
-	PlatformMgr().AddEventHandler(ChipEventHandler, 0);
-
-	err = PlatformMgr().StartEventLoopTask();
-	if (err != CHIP_NO_ERROR) {
-		LOG_ERR("PlatformMgr().StartEventLoopTask() failed");
-		return err;
-	}
+	ReturnErrorOnFailure(Nordic::Matter::StartChipServer());
```
kkasperczyk-no, line 60: Could `StartChipServer` call `WaitForReadiness` internally? It seems weird to me that user has to care about it and remember to call synchronization method for our module.

applications/matter_weather_station/src/app_task.cpp
```diff
@@ -229,35 +177,13 @@ CHIP_ERROR AppTask::Init()
...
 
-	err = PlatformMgr().StartEventLoopTask();
-	if (err != CHIP_NO_ERROR) {
-		LOG_ERR("PlatformMgr().StartEventLoopTask() failed");
-		return err;
-	}
+	ReturnErrorOnFailure(Nordic::Matter::StartChipServer());
 
-	return CHIP_NO_ERROR;
+	return Nordic::Matter::WaitForReadiness();
```
ArekBalysNordic, line 186: Hmm, I don't get the WaitForReadiness and that InitGuard functionality. First of all, why does a user need to think about that and use it in their application? It is a bit confusing.

samples/matter/common/src/matter_init.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+Nordic::Matter::ChipCustomInit sCustomInit{ nullptr };
+
+K_MUTEX_DEFINE(sInitMutex);
+K_CONDVAR_DEFINE(sInitCondVar);
+CHIP_ERROR sInitResult;
+bool sInitDone{ false };
+
+/* RAII utility to implement automatic cleanup and signalling. */
+struct InitGuard {
```
ArekBalysNordic, line 60: Why do we need the whole InitGuard functionality? I can see that it is "working" when the DoInitChipServer method is in progress, but what can interrupt that method? As far as I remember the whole init is run sequentially until we call `PlatformMgr().StartEventLoopTask();`. Do you observe some multithread problems here, and due to that have you decided to add the Init Guard?

samples/matter/common/src/matter_init.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+		/* Before we schedule anything to execute in the CHIP thread, the platform memory
+		   and the stack itself must be initialized first. */
+		CHIP_ERROR err = Platform::MemoryInit();
+		VerifyInitResultOrReturnError(err, "Platform::MemoryInit() failed");
+		err = PlatformMgr().InitChipStack();
+		VerifyInitResultOrReturnError(err, "PlatformMgr().InitChipStack() failed");
+
+		/* Schedule all CHIP initializations to the CHIP thread for better synchronization. */
+		return PlatformMgr().ScheduleWork(DoInitChipServer, reinterpret_cast<intptr_t>(chipEventHandler));
```
ArekBalysNordic, line 206: Hmm, ok it may make the whole initialization better synchronous. But for users, it will be confusing a lot I guess. Personally, I liked the previous way because it worked fine, and the whole inits were done one by one, and it was super clear for users. Doing it in a chip thread may confuse the user because they can just forget about calling "StartChipServer" and nothing will be initialized. :)

samples/matter/common/src/matter_init.cpp
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	CHIP_ERROR StartChipServer()
```
ArekBalysNordic, line 209: I can see that the `StartChipServer` method doesn't start the chip server at all :)

ArekBalysNordic (commented):
First round of the review.

samples/matter/common/src/matter_init.cpp
markaj-nordic: The thing is we had, but this was hidden in each sample initialization sequence. On example might be deferred attribute persister in the light bulb or customized factory data provider in the weather station app. Regarding customers perspective, I guess we need to find a consensus. One approach is to hard-code everything with #ifdefs just to support our current samples structure. The second one is to make everything configurable, so may as well do nothing with the current state. I guess none of these is what we want. The singular `InitChipServer` won't work because it gives no means to inject customized CHIP interfaces implementation. Also we need to keep in mind that we have at least 2 threads here that need to be synchronized. Otherwise our solution is prone to races, which actually happened in the past.

samples/matter/common/src/matter_init.cpp
markaj-nordic: Because to achieve thread safety and synchronization we initialize all Matter interfaces in CHIP thread and all application object in the application thread. The second part can only be started once the CHIP initialization is finalized. That's why the application thread must wait until the CHIP thread signals that the initialization is done.

applications/matter_weather_station/src/app_task.cpp
markaj-nordic: The `InitGuard` object is just an utility which does the cleanup automatically, i.e. it unblocks the waiting thread on each failure can happen in the `DoInitChipServer`. Otherwise we need to call `k_condvar_signal` each time `DoInitChipServer` returns prematurely.

samples/matter/common/src/matter_init.cpp
markaj-nordic: It starts the main CHIP event loop so essentially it starts the server :) But we can think of a batter name, I have no problem with that (e.g. StartChipThread or StartChipEventLoop)

samples/matter/light_switch/src/app_task.cpp
markaj-nordic: Well we can, I thought it was better to make an explicit call for that. I guess we can embed waiting stage in the StartChipServer provided the API documentation is very clear on that.

samples/matter/lock/src/app_task.cpp
markaj-nordic: I was thinking about it. With the pointer argument we'd need to keep data passed to the method alive all the time. If we want to pass the raw ChipInitData temporary object we'd need to either check validity of all fields in the passed structure or add additional field like `IsDataSet`. But I'll think about it more.

samples/matter/common/src/matter_init.h
markaj-nordic: I don't know myself :) Chip is still used in most of the Matter SDK interfaces. I'm ok with using only Matter in all added code. @Damian-Nordic, @ArekBalysNordic any thoughts?

Damian-Nordic: I don't have a strong opinion whether to use `chip` or `matter` (except that it should be consistent in the common module) but I agree that if the namespace is `matter` then it's unnecessary to add either `chip` or `matter` to class or function names.

samples/matter/common/src/matter_init.cpp
markaj-nordic: But that's the point of using the documented API, isn't it? :) Currently the CHIP server also needs to be started with  `PlatformMgr().StartEventLoopTask();` but it's nowhere documented. So for me it sort of addresses the lack of Matter SDK API documentation.

markaj-nordic: Discussed f2f.

samples/matter/common/src/matter_init.cpp
```diff
@@ -0,0 +1,230 @@
...
+		sInitResult = sCustomInit();
+		VerifyInitResultOrReturn(sInitResult, "Custom initialization failed");
+	}
+}
+} // namespace
+
+/* Public API */
+namespace Nordic
+{
+namespace Matter
```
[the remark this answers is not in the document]
markaj-nordic: Discussed f2f.

samples/matter/common/src/matter_init.cpp
markaj-nordic: Discussed f2f.

samples/matter/common/src/matter_init.h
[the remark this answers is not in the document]
markaj-nordic: The problem I see with accepting `CommissionableDataProvider` and other Matter interfaces instead of `FactoryDataProxy` is the fact that the latter contains the `Init()` method which is specific only to the nrfconnect FactoryData implementation (as a matter of fact most of other platforms do similar). Therefore I guess the initialization code must remain nrfconnect platform biased. As a workaround the user of MatterInit module can explicitly set the `mFactoryDataProvider` to nullptr and provide the customized implementation by directly using Matter API in the callback, e.g:

ReturnErrorOnFailure(Nordic::Matter::InitServer(
		MatterEventHandler, chip::MakeOptional<Nordic::Matter::InitData>(Nordic::Matter::InitData{
					    .mFactoryDataProvider = nullptr, .mPostServerInitClbk = [] {
						    chip::DeviceLayer::SetDeviceInstanceInfoProvider(cusomInfoProvider);
						    chip::Credentials::SetDeviceAttestationCredentialsProvider(customAttProvider);
                                                    chip::DeviceLayer::SetCommissionableDataProvider(customCommProvider);
						    return CHIP_NO_ERROR;
					    } })));

samples/matter/lock/src/app_task.cpp
markaj-nordic: Got rid of the Optional.

applications/matter_bridge/src/app_task.cpp
```diff
@@ -244,3 +127,44 @@ CHIP_ERROR AppTask::RestoreBridgedDevices()
 	}
 	return CHIP_NO_ERROR;
 }
+
+CHIP_ERROR AppTask::Init()
+{
+	/* Initialize Matter stack */
+	ReturnErrorOnFailure(Nordic::Matter::InitServer(MatterEventHandler));
```
Damian-Nordic, line 134: nit: I wonder if it might be confusing for people to realize that `InitServer` doesn't actually initialize the Matter `Server` object. I am not sure if we can come up with a better name though (sth like `InitPlatform` perhaps), so you can leave it as it is.

samples/matter/common/src/matter_init.h
```diff
@@ -0,0 +1,109 @@
...
+	 */
+	struct InitData {
+		/** @brief Pointer to the user provided NetworkCommissioning instance. */
+#ifdef CONFIG_CHIP_WIFI
+		chip::app::Clusters::NetworkCommissioning::Instance *mNetworkingInstance{ &sWiFiCommissioningInstance };
+#else
+		chip::app::Clusters::NetworkCommissioning::Instance *mNetworkingInstance{ nullptr };
+#endif
+		/** @brief Pointer to the user provided custom server initialization parameters. */
+		chip::CommonCaseDeviceServerInitParams *mServerInitParams{ &sServerInitParamsDefault };
```
Damian-Nordic, line 45: This doesn't have to be `CommonCase` object. Any `ServerInitParams` will do, right?

samples/matter/window_covering/src/app_task.cpp
```diff
@@ -261,3 +152,30 @@ void AppTask::ChipEventHandler(const ChipDeviceEvent *event, intptr_t /* arg */)
...
+{
+	/* Initialize Matter stack */
+	ReturnErrorOnFailure(Nordic::Matter::InitServer(MatterEventHandler));
+
+	if (!GetBoard().Init(ButtonEventHandler)) {
+		LOG_ERR("User interface initialization failed.");
+		return CHIP_ERROR_INCORRECT_STATE;
+	}
+
+	WindowCovering::Instance().PositionLEDUpdate(WindowCovering::MoveType::LIFT);
```
Damian-Nordic, line 166: This runs the code that accesses the attribute store before the Matter data model is initialized. Even if it works, the behavior is undefined, so maybe this should be moved to `emberAfWindowCoveringClusterInitCallback`?

samples/matter/thermostat/src/app_task.cpp
```diff
@@ -5,43 +5,25 @@
...
-#include <platform/nrfconnect/wifi/NrfWiFiDriver.h>
-#endif
 
 #ifdef CONFIG_CHIP_OTA_REQUESTOR
 #include "ota_util.h"
 #endif
 
-#include <dk_buttons_and_leds.h>
-#include <zephyr/kernel.h>
-#include <zephyr/logging/log.h>
```
Damian-Nordic, line 38: Logging is still used in these samples and so are button constants AFAICS

samples/matter/thermostat/src/app_task.cpp
```diff
@@ -250,3 +112,39 @@ void AppTask::ChipEventHandler(const ChipDeviceEvent *event, intptr_t /* arg */)
...
+		return CHIP_ERROR_INCORRECT_STATE;
+	}
+
+	CHIP_ERROR err = TempSensorManager::Instance().Init();
+	if (err != CHIP_NO_ERROR) {
+		LOG_ERR("TempSensorManager Init fail");
+		return err;
+	}
+
+	err = TemperatureManager::Instance().Init();
```
Damian-Nordic, line 132: This accesses the attribute store before the Matter data model is initialized.

samples/matter/light_switch/src/app_task.cpp
```diff
@@ -319,3 +187,33 @@ void AppTask::UserTimerTimeoutCallback(k_timer *timer)
...
+CHIP_ERROR AppTask::Init()
+{
+	/* Initialize Matter stack */
+	ReturnErrorOnFailure(Nordic::Matter::InitServer(MatterEventHandler));
+
+	/* Initialize application timers */
+	k_timer_init(&sDimmerPressKeyTimer, AppTask::UserTimerTimeoutCallback, nullptr);
+	k_timer_init(&sDimmerTimer, AppTask::UserTimerTimeoutCallback, nullptr);
+
+	LightSwitch::GetInstance().Init(kLightSwitchEndpointId);
```
Damian-Nordic, line 200: This initializes BindingHandler that accesses Server's members before `Server` is initialized.

samples/matter/light_bulb/src/app_task.cpp
```diff
@@ -428,3 +279,56 @@ void AppTask::UpdateClusterState()
...
+		return chip::System::MapErrorZephyr(ret);
+	}
+#endif
+
+	/* Initialize trigger effect timer */
+	k_timer_init(&sTriggerEffectTimer, &AppTask::TriggerEffectTimerTimeoutCallback, nullptr);
+
+	/* Initialize lighting device (PWM) */
+	uint8_t minLightLevel = kDefaultMinLevel;
+	Clusters::LevelControl::Attributes::MinLevel::Get(kLightEndpointId, &minLightLevel);
```
Damian-Nordic, line 311: Access of the attribute store before the Matter data model is initialied

samples/matter/common/src/matter_init.cpp
```diff
@@ -0,0 +1,238 @@
...
+{
+	CHIP_ERROR InitServer(PlatformManager::EventHandlerFunct matterEventHandler, InitData initData)
+	{
+		sCommissioningInstance = initData.mNetworkingInstance;
```
Damian-Nordic, line 203: You could save the entire `initData` instead of all members separately.

Damian-Nordic (commented):
I think what might be tricky in the current design (do all initialization in CHIP thread) is making sure that objects that access Matter structures are not initialized before Server::Init. Previously, it was a bit more straightforward to inject code between `Server::Init` and `StartEventLoopTask` and now either the post-init callback or ZCL init callbacks must be used for those components.

Maybe the part of the confusion is that `InitServer` doesn't actually initialize the `Server` object, so it's easy to assume Server members can already be accessed. In any case, make sure that our nightly is run before merging this PR :)

applications/matter_bridge/src/app_task.cpp
markaj-nordic: Maybe `PrepareServer` would work?

samples/matter/common/src/matter_init.cpp
markaj-nordic: Fair point. But if we keep the whole `InitData` structure in the matter_init.cpp, will the default static objects (implementing Matter interfaces) of this instance be stripped by the compiler/linker? It is the case when we initialize the structure immediately (e.g. when passing it by copy to the function) but I wonder if the behavior is the same if we first create the default-initialized object and then overwrite it in `InitServer` function.

samples/matter/common/src/matter_init.h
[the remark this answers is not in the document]
Damian-Nordic: I see, I guess then a user would need to call `Init` on their own. I don't have a good solution if we want to keep `Init` inside the common module...

samples/matter/common/src/matter_init.cpp
Damian-Nordic: Good point. I guess you could include `matterEventHandler` as `initData` member and then allocate the copy of `initData` on heap and pass it directly to `ScheduleWork`. But feel free to ignore this comment as it's not that important after all.

samples/matter/common/src/matter_init.h
markaj-nordic: The `ServerInitParams` API does not include `InitializeStaticResourcesBeforeServerInit` which is CommonCase specific... so again, it looks similar to the FactoryDataProxy problem. IMO It's a pity that instead of always using pure Matter interfaces sometimes we need to know implementation specific details. On the other hand CommonCase is explicitly discouraged to be used in the final product firmware, so it's a bit confusing. I wonder if our customers are aware of that :)
https://github.com/nrfconnect/sdk-connectedhomeip/blob/master/src/app/server/Server.h#L153
Transitioning to handling all CommonCase internals in the application now seems a bit overwhelming.

samples/matter/light_bulb/src/app_task.cpp
markaj-nordic: Of course I agree, but just to clarify, isn't it the problem also for the current implementation? I mean even without changes from this PR.

samples/matter/window_covering/src/app_task.cpp
markaj-nordic: Yes, or `mPostServerInitClbk` I guess.

samples/matter/light_switch/src/app_task.cpp
markaj-nordic: Right, this should be done in the mPostServerInitClbk.

samples/matter/common/src/matter_init.cpp
markaj-nordic: Or we can initialize the local InitData with `nullptr`s instead of using default initialization, right?

Damian-Nordic: Yes, we can.

samples/matter/thermostat/src/app_task.cpp
markaj-nordic: dk_buttons_and_leds is a part of `board` module now. Without logging it compiles but I agree we cannot depend on other modules to include it for us.

ArekBalysNordic (approved):
LGTM :)

applications/matter_bridge/src/app_task.cpp
Damian-Nordic: Yes, I think it's better.

samples/matter/common/src/matter_init.h
Damian-Nordic: Well, I think in this case we can somehow get rid off the `CommonCase` but we can clean that later*. But overall, yes, it looks like default initializers make it difficult to use default objects with custom init. 

\* in some cases we could do:
```
struct InitData
{
  ServerInitParams* mServerInitParams = [] {
    sDefaultServerInitParams.InitializeStaticResourcesBeforeServerInit();
    return &sDefaultServerInitParams
  }();
};
```
But I'm not sure if I'm not proposing monstrosity that would only confuse people :).

samples/matter/light_bulb/src/app_task.cpp
Damian-Nordic: Yes, it is :). I guess it's not a big deal when loading "constant" attributes. The problem would be more visible with persistent attributes (like the light brightness) as the attributes wouldn't manage to be loaded from flash before `Get`.

samples/matter/common/src/matter_init.h
markaj-nordic: I have created a task for that, to revisit the usage of CommonCase in NCS Matter samples:
https://nordicsemi.atlassian.net/browse/KRKNWK-18274

samples/matter/lock/src/app_task.cpp
```diff
@@ -413,3 +263,59 @@ void AppTask::NUSUnlockCallback(void *context)
...
+
+CHIP_ERROR AppTask::Init()
+{
+	/* Initialize Matter stack */
+#ifdef CONFIG_THREAD_WIFI_SWITCHING
+	/* In the Thread/Wi-Fi switchable mode, it is the ThreadWifiSwitch module that owns the NetworkCommissioning
+	   instances, so offload the initialization module from controlling that by explicitly setting the
+	   mNetworkingInstance parameter to nullptr. Otherwise, the initialization module instantiates the
+	   NetworkCommissioning object and handles it by default internally. */
+	ReturnErrorOnFailure(Nordic::Matter::InitServer(MatterEventHandler,
```
Damian-Nordic, line 275: Not updated.

Damian-Nordic (approved):

samples/matter/lock/src/app_task.cpp
markaj-nordic: Sorry, fixed.

PR #13475, opened 2023-12-19
samples: openthread: disable `mx25r64` in OpenThread samples

maciejbaczmanski:
`mx25r64` is used to enable nrf_qspi_nor in nrf samples. It is not used in OpenThread and should be disabled by default.

In the future, disabling `mx25r64` will be handled directly in boards `.dts` zephyr files, but until then, we should disable it in project overlays.

### DIFF
--- a/samples/openthread/coap_client/boards/nrf52840dk_nrf52840.overlay
+++ b/samples/openthread/coap_client/boards/nrf52840dk_nrf52840.overlay
@@ -1,4 +1,4 @@
-/* Copyright (c) 2020 Nordic Semiconductor ASA
+/* Copyright (c) 2023 Nordic Semiconductor ASA
  *
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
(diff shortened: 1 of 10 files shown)


edmont (approved):

samples/openthread/coap_client/boards/nrf52840dk_nrf52840.overlay
```diff
@@ -41,6 +41,9 @@
 &qspi {
 	status = "disabled";
 };
+&mx25r64 {
+	status = "disabled";
+};
```
canisLupus1313, line 46: The convention normally is different. node should be disabled by default and then enabled when needed.
```suggestion
&mx25r64 {
	status = "okay";
};
```

canisLupus1313 (changes requested):
```
&mx25r64 {
	status = "disabled";
};

maciejbaczmanski:
As discussed with @canisLupus1313 , we will wait for `mx25r64` to be disabled by default in zephyr to keep the correct convention

PR #13478, opened 2023-12-19
lib: Sample rate converter library

andvib:
Implements sample rate converter library that uses CMSIS DSP functionality to filter an array of samples before adding/removing samples as needed. Supports converting to/from 48kHz from/to 24kHz or 16kHz.


gWacey (approved):

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+ * The internal state buffer must for each context fulfill the following equations:
+ *	Interpolation:
+ *		number of filter taps + block size - 1
+ *	Decimation:
+ *		(number of filter taps / conversion ratio) + block size - 1
+ *
+ * The equation for interpolation is used as size as this gives the largest number.
+ */
+#define SAMPLE_RATE_CONVERTER_STATE_BUFFER_SIZE                                                    \
+	CONFIG_SAMPLE_RATE_CONVERTER_BLOCK_SIZE + CONFIG_SAMPLE_RATE_CONVERTER_MAX_FILTER_SIZE - 1
```
koffes, line 39:
```suggestion
	(CONFIG_SAMPLE_RATE_CONVERTER_BLOCK_SIZE + CONFIG_SAMPLE_RATE_CONVERTER_MAX_FILTER_SIZE - 1)
```

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+
+/** Filter types supported by the sample rate converter */
+enum sample_rate_converter_filter {
+	SAMPLE_RATE_FILTER_SIMPLE,
```
koffes, line 43:
```suggestion
	SAMPLE_RATE_FILTER_SIMPLE  = 1,
```

koffes: Suggest renaming SIMPLE to TRIVIAL

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+#define SAMPLE_RATE_CONVERTER_INPUT_BUFFER_NUMBER_OVERFLOW_SAMPLES 2
+
+/**
+ * To maintain filter requirements the output buffer must in some cases store six samples between
+ * each block processed. The output buffer must be able to store a full block in addition to these
+ * overflow samples.
+ */
+#define SAMPLE_RATE_CONVERTER_OUTPUT_BUFFER_NUMBER_OVERFLOW_SAMPLES 6
+
+#ifdef CONFIG_SAMPLE_RATE_CONVERTER_BIT_DEPTH_16
```
koffes, line 60: Add parentheses on all these defines.
```
#define FOO (1 + 1)
#define BAR 1 + 1

4*FOO != 4*BAR
```

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+#define SAMPLE_RATE_CONVERTER_INPUT_BUF_SIZE                                                       \
+	SAMPLE_RATE_CONVERTER_INPUT_BUFFER_NUMBER_OVERFLOW_SAMPLES * sizeof(uint16_t)
+#define SAMPLE_RATE_CONVERTER_RINGBUF_SIZE                                                         \
+	(CONFIG_SAMPLE_RATE_CONVERTER_BLOCK_SIZE +                                                 \
+	 SAMPLE_RATE_CONVERTER_OUTPUT_BUFFER_NUMBER_OVERFLOW_SAMPLES) *                            \
+		sizeof(uint16_t)
+#elif CONFIG_SAMPLE_RATE_CONVERTER_BIT_DEPTH_32
```
koffes, line 67: Add an elif and error if the bit depth is 24

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+#endif
+
+/** Buffer used for storing input bytes to the sample rate converter */
+struct buf_ctx {
+	uint8_t buf[SAMPLE_RATE_CONVERTER_INPUT_BUF_SIZE];
+	size_t bytes_in_buf;
+};
+
+enum sample_rate_converter_direction {
+	CONVERSION_DIR_UP,
```
koffes, line 83:
```suggestion
	CONVERSION_DIR_UP = 1 ,
```

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+	CONVERSION_DIR_DOWN,
```
koffes, line 84: Suggest rename to UPSAMPLE and DOWNSAMPLE

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+};
+
+/** Context for the sample rate conversion */
+struct sample_rate_converter_ctx {
+	/* Input and out sample rate to be used for the conversion. */
+	int input_sample_rate;
```
koffes, line 90: Can these be unsigned?

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * @param[out]	ctx	Pointer to the sample rate conversion context.
+ *
+ * @retval	0	On success.
+ * @retval	-EINVAL	NULL pointer given for context.
+ */
+int sample_rate_converter_open(struct sample_rate_converter_ctx *ctx);
+
+/**
+ * @brief	Process input samples and produce output samples with new samplerate.
```
koffes, line 144:
```suggestion
 * @brief	Process input samples and produce output samples with new sample rate.
```

include/sample_rate_converter.h
```diff
@@ -0,0 +1,179 @@
...
+ * @retval	0	On success.
+ * @retval	-EINVAL	Invalid parameters for sample rate conversion.
+ * @retval	-EFAULT	Output ring buffer has either not enough bytes to output, or not enough
+ *			space to store bytes.
+ */
+int sample_rate_converter_process(struct sample_rate_converter_ctx *ctx,
+				  enum sample_rate_converter_filter filter, void *input,
+				  size_t input_size, uint32_t input_sample_rate, void *output,
+				  size_t output_size, size_t *output_written,
+				  uint32_t output_sample_rate);
```
koffes, line 173:
```suggestion
int sample_rate_converter_process(struct sample_rate_converter_ctx *ctx,
				  enum sample_rate_converter_filter filter, void const * const input,
				  size_t input_size, uint32_t input_sample_rate, void *output,
				  size_t output_size, size_t *output_written,
				  uint32_t output_sample_rate);
```

lib/sample_rate_converter/Kconfig
```diff
@@ -0,0 +1,61 @@
...
+#
+
+config SAMPLE_RATE_CONVERTER
+	bool "Sample Rate conversion library"
+	select CMSIS_DSP
+	select CMSIS_DSP_FILTERING
+	select TIMING_FUNCTIONS
+	help
+	  Enable the sample rate conversion library. The library uses CMSIS DSP filters to
+	  preserve quality during the conversion. 16k, 24k and 48k are supported.
```
koffes, line 14: Is 24 supported?

lib/sample_rate_converter/Kconfig
```diff
@@ -0,0 +1,61 @@
...
+config SAMPLE_RATE_CONVERTER_MAX_FILTER_SIZE
+	int "Maximum number of filter taps the module supports"
+	default 34
+	help
+	  The maximum number of filter taps the sample rate converter supports. Increasing this
+	  number will increase the memory usage of the converter.
+
+config SAMPLE_RATE_CONVERTER_BLOCK_SIZE
+	int "Number of samples per conversion call"
+	default 480
```
koffes, line 43: I thought this would be dependent on the sample rate input? In which case, we may want to rename to SAMPLE_RATE_CONVERTER_BLOCK_SIZE_MAX

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+#include "sample_rate_converter_filter.h"
+
+#include <errno.h>
+#include <stdbool.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(sample_rate_converter, CONFIG_SAMPLE_RATE_CONVERTER_LOG_LEVEL);
+
+/**
+ * The input buffer must be able to store maximum two samples in addition to the block size to meet
```
koffes, line 17:
```suggestion
 * The input buffer must be able to store two samples in addition to the block size to meet
```

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+ * filter requirements.
+ */
+#define SAMPLE_RATE_CONVERTER_INPUT_BUF_NUMBER_SAMPLES
```
koffes, line 20: Parentheses on defines.

koffes, line 20: Not defines as anything.

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+				output_sample_rate);
+			return -EINVAL;
+		}
+
+		if ((input_sample_rate != 24000) && (input_sample_rate != 16000)) {
+			LOG_ERR("Invalid input sample rate for upsampling: %d", input_sample_rate);
+			return -EINVAL;
+		}
+	} else {
+		LOG_ERR("Input and out samplerates are the same");
```
koffes, line 62:
```suggestion
		LOG_ERR("Input and out sample rates are the same");
```

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+
+/**
+ * @brief Reconfigures the sample rate converter context.
+ *
+ * @details Validates and sets all sample rate conversion parameters for the context. If buffering
+ *	    is needed for the conversion, the input buffer will be padded with two samples to
+ *	    ensure there will always be enough samples for a valid conversion.
+ *
+ * @param[in,out]	ctx			Pointer to the sample rate conversion context.
+ * @param[in]		input_sample_rate	Sample rate of the input samples.
```
koffes, line 97: Usually, we go from generic to specific in names. Suggest `sample_rate_input `and `sample_rate_output`

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+	if (ret) {
+		LOG_ERR("Invalid sample rate given (%d)", ret);
+		return ret;
+	}
+
+	ctx->input_sample_rate = input_sample_rate;
+	ctx->output_sample_rate = output_sample_rate;
+
+	ctx->conversion_ratio = calculate_conversion_ratio(input_sample_rate, output_sample_rate);
+	ctx->conversion_direction =
```
koffes, line 126: If input_sample_rate == output_sample_rate then CONVERSION_DIR_DOWN will be set which is a bit misleading. Suggest making a static function and handle this case.

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+		LOG_ERR("Failed to get filter (%d)", ret);
+		return ret;
+	} else if (filter_size > CONFIG_SAMPLE_RATE_CONVERTER_MAX_FILTER_SIZE) {
+		LOG_ERR("Filter is larger than max size");
+		return -EINVAL;
+	}
+
+	if (conversion_needs_buffering(ctx, ctx->conversion_ratio)) {
+		LOG_DBG("Conversion needs buffering, start with the input buffer filled");
+#ifdef CONFIG_SAMPLE_RATE_CONVERTER_BIT_DEPTH_16
```
koffes, line 142: Instead of #ifdef, try to use if (IS_DEFINED())

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+#endif
+	if (arm_err == ARM_MATH_LENGTH_ERROR) {
+		LOG_ERR("Filter size is not a multiple of conversion ratio");
+		return -EINVAL;
+	} else if (arm_err != ARM_MATH_SUCCESS) {
+		LOG_ERR("Unknown error during interpolator/decimator initialization (%d)", arm_err);
+		return -EINVAL;
+	}
+
+	LOG_DBG("Sample rate converter initialized. Input sample rate: %d, Output sample rate: %d, "
```
koffes, line 194: Suggest 
```suggestion
	LOG_INF("Sample rate converter initialized. Input sample rate: %d, Output sample rate: %d, "
```

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+	}
+
+	if ((ctx == NULL) || (input == NULL) || (output == NULL) || (output_written == NULL)) {
+		LOG_ERR("Null pointer received");
+		return -EINVAL;
+	}
+
+	if ((ctx->input_sample_rate != input_sample_rate) ||
+	    (ctx->output_sample_rate != output_sample_rate) || (ctx->filter_type != filter)) {
+		LOG_DBG("State has changed, re-initializing filter");
```
koffes, line 252: Do we expect this to change run-time? Perhaps this should be a WRN?

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+
+	if (*output_written > output_size) {
+		LOG_ERR("Conversion process will produce more bytes than the output buffer can "
+			"hold");
+		return -EINVAL;
+	}
+
+	if (*output_written > SAMLPE_RATE_CONVERTER_INTERNAL_OUTPUT_BUF_SIZE) {
+		LOG_ERR("Conversion process will produce more bytes than the internal output "
+			"buffer can hold");
```
koffes, line 283:
```suggestion
	if (*output_written > SAMLPE_RATE_CONVERTER_INTERNAL_OUTPUT_BUF_SIZE) {
		LOG_ERR("Conversion process will produce more bytes than the internal output buffer can hold");
```

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+		return -EINVAL;
+	}
+
+	if (!conversion_needs_buffering(ctx, ctx->conversion_ratio)) {
```
koffes, line 287: Can we switch around this if check, so we check for `conversion_needs_buffering` and not `!conversion_needs_buffering` ?

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,407 @@
...
+				     samples_to_process);
+	}
+#endif
+
+	if (!conversion_needs_buffering(ctx, ctx->conversion_ratio)) {
+		/* Nothing needs to be done in output buffer */
+		return 0;
+	}
+
+	if (samples_to_process < samples_in) {
```
koffes, line 340: Would appreciate a walkthrough of the last part of this function.

lib/sample_rate_converter/sample_rate_converter_filter.c
```diff
@@ -0,0 +1,128 @@
...
+
+#include "sample_rate_converter.h"
+#include "sample_rate_converter_filter.h"
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(sample_rate_converter_filter, CONFIG_SAMPLE_RATE_CONVERTER_LOG_LEVEL);
+
+#ifdef CONFIG_SAMPLE_RATE_CONVERTER_FILTER_SIMPLE
+#ifdef CONFIG_SAMPLE_RATE_CONVERTER_BIT_DEPTH_16
+static q15_t filter_48khz_24kz_16bit_simple[] = {0x3ffe, 0x3fff};
```
koffes, line 15:
```suggestion
static q15_t filter_48khz_24kz_16bit_simple[] = {0x3ffe, 0x3fff};
```
Discuss naming.

lib/sample_rate_converter/sample_rate_converter_filter.c
```diff
@@ -0,0 +1,128 @@
...
+
+#elif CONFIG_SAMPLE_RATE_CONVERTER_BIT_DEPTH_32
+static q31_t filter_48khz_16khz_32bit_small[] = {
+	0xff34e0fc, 0xfd20fa09, 0x00b8d644, 0x01b78c3a, 0x01b60933, 0xff379d35, 0xfcec701a,
+	0xfde17e67, 0x020dad5a, 0x051dfe81, 0x027ae300, 0xfb0ed811, 0xf6869fd3, 0xfd49086d,
+	0x0fea99ac, 0x24820268, 0x2d7765d4, 0x24820268, 0x0fea99ac, 0xfd49086d, 0xf6869fd3,
+	0xfb0ed811, 0x027ae300, 0x051dfe81, 0x020dad5a, 0xfde17e67, 0xfcec701a, 0xff379d35,
+	0x01b60933, 0x01b78c3a, 0x00b8d644, 0xfd20fa09, 0xff34e0fc};
+
+static q31_t filter_48khz_24khz_32bit_small[] = {
```
koffes, line 46:
```suggestion
static q31_t filter_48khz_to_24khz_32bit_small[] = {
```

tests/lib/sample_rate_converter/prj.conf
```diff
@@ -0,0 +1,9 @@
+CONFIG_ZTEST=y
+CONFIG_ZTEST_NEW_API=y
```
koffes, line 2: I think this is enabled by default.

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	zassert_equal(ring_buf_size_get(&conv_ctx.output_ringbuf), 0,
+		      "Number of bytes in output ringbuffer not as expected %d",
+		      ring_buf_size_get(&conv_ctx.output_ringbuf));
+
+	zassert_equal(output_written, expected_output_samples * sizeof(uint16_t),
+		      "Output size was not as expected (%d)", output_written);
+
+	/* Verify output bytes, expected two input samples per output sample */
+	for (int i = 0; i < expected_output_samples; i++) {
+		zassert_within(output_samples[i], input_samples[i * 2], 550);
```
koffes, line 61: We need to define 550 and provide some reasoning behind this value.

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	zassert_equal(ring_buf_size_get(&conv_ctx.output_ringbuf), 0,
+		      "Number of bytes in output ringbuffer not as expected %d",
+		      ring_buf_size_get(&conv_ctx.output_ringbuf));
+
+	zassert_equal(output_written, expected_output_samples * sizeof(uint16_t),
+		      "Output size was not as expected (%d)", output_written);
+
+	/* Verify output bytes, expected three input samples per output sample */
+	for (int i = 0; i < expected_output_samples; i++) {
+		zassert_within(output_samples[i], input_samples[i * 3], 1100);
```
koffes, line 107: See earlier comment.

koffes (changes requested):

katgiadla:
Does this tests use Twister to test? 
If yes, please attach testcase.yaml.
Expect that, please specify the code owners.

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	enum sample_rate_converter_filter filter = SAMPLE_RATE_FILTER_SIMPLE;
+	size_t output_written;
+
+	ret = sample_rate_converter_process(
+		&conv_ctx, filter, input_samples, num_samples * sizeof(uint32_t), input_sample_rate,
+		output_samples, expected_output_samples * sizeof(uint32_t), &output_written,
+		output_sample_rate);
+
+	zassert_equal(ret, 0, "Sample rate conversion process failed");
+	zassert_equal(conv_ctx.input_sample_rate, input_sample_rate,
```
koffes, line 377: Suggest that these checks which are repeated are moved to a static function.

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	zassert_equal(ring_buf_size_get(&conv_ctx.output_ringbuf), 0,
+		      "Number of bytes in output ringbuffer not as expected %d",
+		      ring_buf_size_get(&conv_ctx.output_ringbuf));
+
+	zassert_equal(output_written, expected_output_samples * sizeof(uint32_t),
+		      "Output size was not as expected (%d)", output_written);
+
+	/* Verify output bytes, expected two samples in output per input */
+	for (int i = 0; i < expected_output_samples; i++) {
+		zassert_true(output_samples[i] < input_samples[(i / 2)], "not correct");
```
koffes, line 441: Discuss these checks. We may need to add another step to verify the output.

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	zassert_equal(output_written, expected_output_samples * sizeof(uint32_t),
+		      "Output size was not as expected (%d)", output_written);
+
+	/* Verify output bytes, expected three samples in output per input */
+	for (int i = 0; i < expected_output_samples; i++) {
+		zassert(output_samples[i] < input_two[(i / 3)],
+			"Output samples is not smaller than corresponding input");
+	}
+
+	/* Third run */
```
koffes, line 530: Several tests have multiple runs which is good, but can we execute in a loop?

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+		output_samples, expected_output_samples, &output_written, output_sample_rate);
+
+	zassert_equal(ret, -EINVAL, "Sample rate conversion process did not fail");
+}
+
+ZTEST(suite_sample_rate_converter, test_invalid_process_buffer_null_ptr)
+{
+	int ret;
+
+	uint16_t input_samples[] = {1000, 2000, 3000, 40000};
```
koffes, line 766:
```suggestion
	uint16_t input_samples[] = {1000, 2000, 3000, 4000};
```

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+
+	zassert_equal(ret, -EINVAL,
+		      "Sample rate conversion process did not fail when output pointer is NULL");
+}
+
+ZTEST(suite_sample_rate_converter, test_invalid_process_output_written_null_ptr)
+{
+	int ret;
+
+	uint16_t input_samples[] = {1000, 2000, 3000, 40000};
```
koffes, line 795:
```suggestion
	uint16_t input_samples[] = {1000, 2000, 3000, 4000};
```

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	ret = sample_rate_converter_open(NULL);
+
+	zassert_equal(ret, -EINVAL, "Call to open did not fail");
+}
+
+ZTEST(suite_sample_rate_converter, test_valid_process_zero_size_input)
+{
+	int ret;
+
+	uint16_t input_samples[] = {1000, 2000, 3000, 40000};
```
koffes, line 826:
```suggestion
	uint16_t input_samples[] = {1000, 2000, 3000, 4000};
```

koffes (changes requested):

lib/sample_rate_converter/Kconfig
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config SAMPLE_RATE_CONVERTER
```
lemrey, line 7: Please make this a menuconfig to avoid these entries appearing in the top-level menu.

include/sample_rate_converter.h
andvib: I don't think I agree. Currently the simple filter still does some filtering with a very small filter, but it's still doing DSP filtering. In my head trivial is more like the implementation that was made earlier that blindly adds/removes samples, which is more of a trivial method of doing it.

Writing this comment made me realize that maybe SIMPLE and SMALL is the wrong way to go though. We could just call it SMALL and BIG, and refactor what is the big and small filters in the future _if_ we add more filters?

include/sample_rate_converter.h
andvib: These kconfigs are specific to the sample rate converter library, so only kconfigs for 32 and 16 exists. So if I understand you correctly, if the user tries to configure the audio project to use the sample rate converter with a 24 bit depth system, we should raise an error at the audio application level

include/sample_rate_converter.h
andvib: So `CONVERSION_DIR_UPSAMPLE` or just `UPSAMPLE?`

I think I prefer the full name, though I agree that upsample/downsample is better than just up/down

include/sample_rate_converter.h
andvib: Yes (and should be 🤠)

lib/sample_rate_converter/Kconfig
andvib: It is - but I realized this text is a bit confusing as it stands. I've specified that it's referring to kHz, so hopefully that makes it clearer what the numbers are.

lib/sample_rate_converter/Kconfig
andvib: MAX is a better name, changing to that. It is indeed dependent on the input size, but as the text says the actual number of samples can be lower, so max is the correct name

lib/sample_rate_converter/sample_rate_converter.c
andvib: If `input_sample_rate == output_sample_rate`  the conversion will be rejected by `validate_sample_rates()`. So in my opinion we are already handling that case in a proper manner to ensure this case can't happen here. Do you still think we need to add another check for it?

lib/sample_rate_converter/sample_rate_converter.c
andvib: My understanding is that the audio application allows the stream to change sample rate during run time, and I wrote it like this so the application does not need to care if the sample rate changes. If this is not the case we should definitely discuss changing it! 👍 

As for the log message I think it makes the most sense to have debug as it is a completely valid case the way the lib is written now.

lib/sample_rate_converter/sample_rate_converter.c
andvib: This log message is too long to be on one line, and the autoformatter splits the line like this.  Is this not the correct way to split the line?

lib/sample_rate_converter/sample_rate_converter.c
andvib: Sure, let's do that offline. I was also surprised by how long this last part ended up being

tests/lib/sample_rate_converter/prj.conf
andvib: Seems that it actually isn't, it is disabled when I remove it from the .conf file, and the definition doesn't have a default.

tests/lib/sample_rate_converter/src/main.c
andvib: I agree, they are way to arbitrary and magic as they are now - but I'm not sure how to find a good, scientific reasoning for it. Can we go through the tests offline and do some brainstorming for it?

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	zassert_equal(conv_ctx.input_buf.bytes_in_buf, 0, "Bytes in input buffer not as expected");
+	zassert_equal(ring_buf_size_get(&conv_ctx.output_ringbuf), 0,
+		      "Number of bytes in output ringbuffer not as expected %d",
+		      ring_buf_size_get(&conv_ctx.output_ringbuf));
+
+	zassert_equal(output_written, expected_output_samples * sizeof(uint16_t),
+		      "Output size was not as expected (%d)", output_written);
+
+	/* Verify output bytes, expect two samples in output per input */
+	for (int i = 0; i < expected_output_samples; i++) {
```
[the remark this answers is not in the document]
andvib: It proved difficult to find a good way to check the actual output values, as a small number of samples really isn't enough for the filter to produce good output. So for this test case I found that a sanity check could be that the output samples aren't higher than the input samples (since the filter is a low pass filter). We can of course hard code the expected output values, but IMO that doesn't give us any more confidence and we get tests that are harder to maintain.

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+		      "Conversion direction is not as expected");
+	zassert_equal(conv_ctx.filter_type, filter, "Filter set incorrectly");
+	zassert_equal(conv_ctx.input_buf.bytes_in_buf, 2,
+		      "Bytes in input buffer not as expected %d", conv_ctx.input_buf.bytes_in_buf);
+	zassert_equal(ring_buf_size_get(&conv_ctx.output_ringbuf), 6,
+		      "Number of bytes in output ringbuffer not as expected %d",
+		      ring_buf_size_get(&conv_ctx.output_ringbuf));
+
+	uint16_t expected_input_buf_two[] = {8000};
+	zassert_mem_equal(conv_ctx.input_buf.buf, expected_input_buf_two, 1 * sizeof(uint16_t));
```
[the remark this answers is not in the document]
andvib: Updated to use size of array

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	zassert_equal(conv_ctx.conversion_direction, CONVERSION_DIR_UP,
+		      "Conversion direction is not as expected");
+	zassert_equal(conv_ctx.filter_type, filter, "Filter set incorrectly");
+	zassert_equal(conv_ctx.input_buf.bytes_in_buf, 4,
+		      "Bytes in input buffer not as expected %d", conv_ctx.input_buf.bytes_in_buf);
+	zassert_equal(ring_buf_size_get(&conv_ctx.output_ringbuf), 0,
+		      "Number of bytes in output ringbuffer not as expected %d",
+		      ring_buf_size_get(&conv_ctx.output_ringbuf));
+
+	uint16_t expected_input_buf_three[] = {11000, 12000};
```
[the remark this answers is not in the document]
andvib: This array contains the values we expect to be stored in the input buffer for the conversion context. I agree the name is a bit messy, so name ideas are appreciated!

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1065 @@
...
+	zassert_mem_equal(conv_ctx.input_buf.buf, expected_input_buf_three, 2 * sizeof(uint16_t));
```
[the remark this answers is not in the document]
andvib: Updated to use `sizeof(expected_input_buf_three)`

tests/lib/sample_rate_converter/src/main.c
andvib: I prefer to have all checks in the test, I think that makes it more clear what is tested and easier to understand the test case when debugging.

tests/lib/sample_rate_converter/src/main.c
andvib: We probably could, but we would have to keep track of at least three values that change between each iteration, so I think the test case is clearer without additional loop-logic

lib/sample_rate_converter/Kconfig
andvib: Good point, done!

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,406 @@
...
+#endif
+	if (arm_err == ARM_MATH_LENGTH_ERROR) {
+		LOG_ERR("Filter size is not a multiple of conversion ratio");
+		return -EINVAL;
+	} else if (arm_err != ARM_MATH_SUCCESS) {
+		LOG_ERR("Unknown error during interpolator/decimator initialization (%d)", arm_err);
+		return -EINVAL;
+	}
+
+	LOG_INF("Sample rate converter initialized. Input sample rate: %d, Output sample rate: %d, "
```
alexsven, line 193:
```suggestion
	LOG_DBG("Sample rate converter initialized. Input sample rate: %d, Output sample rate: %d, "
```

lib/sample_rate_converter/sample_rate_converter.c
alexsven: The conversion can absolutely change run-time. If you are going from a phone call to high quality music for instance, or even medium quality (24kHz)

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,406 @@
...
+		*output_written = input_size / ctx->conversion_ratio;
+	}
+
+	if (*output_written > output_size) {
+		LOG_ERR("Conversion process will produce more bytes than the output buffer can "
+			"hold");
+		return -EINVAL;
+	}
+
+	if (*output_written > SAMLPE_RATE_CONVERTER_INTERNAL_OUTPUT_BUF_SIZE) {
```
alexsven, line 280:
```suggestion
	if (*output_written > SAMPLE_RATE_CONVERTER_INTERNAL_OUTPUT_BUF_SIZE) {
```
It obviously needs to be changed on line:27 as well ;)

alexsven (changes requested):

lib/sample_rate_converter/sample_rate_converter.c
koffes: As long as it passes CI it is fine.

tests/lib/sample_rate_converter/prj.conf
koffes: Yes. Rebase on main now as the upmerge is done and remove this.

tests/lib/sample_rate_converter/src/main.c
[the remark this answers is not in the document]
koffes: OK, let's discuss

koffes (changes requested):
Great! The remaining few issues from me needs to be discussed.

include/sample_rate_converter.h
```diff
@@ -0,0 +1,180 @@
...
+ * @brief Implements functionality to increase or decrease the sample rate of a data array using
+ * the CMSIS DSP filtering library.
+ *
+ * @{
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/sys/ring_buffer.h>
+#include <dsp/filtering_functions.h>
```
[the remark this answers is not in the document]
andvib: Good catch, even kernel.h wasn't used. Removed.

include/sample_rate_converter.h
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * @param[in,out]	ctx			Pointer to the sample rate conversion context.
+ * @param[in]		filter			Filter type to be used for the conversion.
+ * @param[in]		input			Pointer to samples to process.
+ * @param[in]		input_size		Size of the input in bytes.
+ * @param[in]		input_sample_rate	Sample rate of the input bytes.
+ * @param[out]		output			Array that output will be written.
+ * @param[in]		output_size		Size of the output array in bytes.
+ * @param[out]		output_written		Number of bytes written to output.
+ * @param[in]		output_sample_rate	Sample rate of output.
```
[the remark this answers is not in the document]
andvib: I'm not sure I see the reason for the change here, isn't it standard to have the argument for buffer pointer first, then the meta?

tests/lib/sample_rate_converter/prj.conf
andvib: That seems to be correct - removed

lib/sample_rate_converter/Kconfig
```diff
@@ -0,0 +1,62 @@
...
+	  Includes the simple filters for the sample rate converter. The simple filters uses the
+	  least amount of space and time for the conversion, at the expense of quality.
+
+config SAMPLE_RATE_CONVERTER_FILTER_SMALL
+	bool "Include the small sample rate converter filters"
+	help
+	  Includes the small filters for the sample rate converter. The small filters uses a small
+	  amount of space and time for the conversion, while also giving some low-pass filter
+	  capabilities.
+
```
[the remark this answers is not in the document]
andvib: Both can be selected, so it can't be a choice. The idea is that you select at compile time what filters to include (so you can save space if needed), and then change between the selected filters at run time.

lib/sample_rate_converter/Kconfig
```diff
@@ -0,0 +1,62 @@
...
+config SAMPLE_RATE_CONVERTER_MAX_FILTER_SIZE
+	int "Maximum number of filter taps the module supports"
+	default 34
+	help
+	  The maximum number of filter taps the sample rate converter supports. Increasing this
+	  number will increase the memory usage of the converter.
```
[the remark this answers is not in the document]
andvib: No, it doesn't affect processing time.

I felt it was good to point out the memory usage since this kconfig directly changes the size of the context struct which may often be statically allocated, not dynamically. So basically just to make it clear that this shouldn't mindlessly be set high to "allow for all cases".

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,406 @@
...
+	(CONFIG_SAMPLE_RATE_CONVERTER_BLOCK_SIZE_MAX +                                             \
+	 SAMPLE_RATE_CONVERTER_INPUT_BUFFER_NUMBER_OVERFLOW_SAMPLES)
+
+#ifdef CONFIG_SAMPLE_RATE_CONVERTER_BIT_DEPTH_16
+#define SAMPLE_RATE_CONVERTER_INTERNAL_INPUT_BUF_SIZE                                              \
+	(INTERNAL_INPUT_BUF_NUMBER_SAMPLES * sizeof(uint16_t))
+#define SAMLPE_RATE_CONVERTER_INTERNAL_OUTPUT_BUF_SIZE                                             \
+	(CONFIG_SAMPLE_RATE_CONVERTER_BLOCK_SIZE_MAX * sizeof(uint16_t))
+#elif CONFIG_SAMPLE_RATE_CONVERTER_BIT_DEPTH_32
+#define SAMPLE_RATE_CONVERTER_INTERNAL_INPUT_BUF_SIZE                                              \
```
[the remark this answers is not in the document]
andvib: The input buffer (which has size `SAMPLE_RATE_CONVERTER_INPUT_BUF_SIZE`) is stored in the context and contains the bytes that overflows between each process call. The internal input buffer (which is size `SAMPLE_RATE_CONVERTER_INTERNAL_INPUT_BUF_SIZE`) is the size of the buffer allocated by the process function in order to prepare the sample array for the filter, and therefore needs to be able to store both a full frame worth of samples and the samples that may have overflowed from last process call.

I get that these are named confusingly similar, so I'll be happy to discuss the naming!

lib/sample_rate_converter/sample_rate_converter.c
```diff
@@ -0,0 +1,406 @@
...
+
+/**
+ * When upsampling from 16kHz to 48kHz the input and output bytes must be buffered. This is to
+ * fulfill the requirement for the filter that the number of input bytes must be divisible by the
+ * conversion factor. This function returns true when this is the case.
+ */
+static inline bool conversion_needs_buffering(struct sample_rate_converter_ctx *ctx,
+					      uint8_t conversion_ratio)
+{
+	return ((ctx->conversion_direction == CONVERSION_DIR_UP) && (conversion_ratio == 3));
```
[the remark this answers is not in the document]
andvib: I like the idea, but I think this would complicate more cases than it simplifies. E.g where the code is currently checking for up/down it would have to check for two cases instead of one, and as the conversion ratio is needed as an integer several places it would have to be calculated either way.

lib/sample_rate_converter/sample_rate_converter.c
andvib: Haha, team work at it's best!

I'll change it back to debug then, since that was my initial gut feeling as well.

andvib:
> ~Does this tests use Twister to test? If yes, please attach testcase.yaml.~ Expect that, please specify the code owners.

Both has been added now :)

CODEOWNERS
```diff
@@ -138,6 +138,7 @@ Kconfig*                                  @tejlmand
 /lib/pcm_mix/                             @koffes @alexsven @erikrobstad @rick1082 @gWacey
 /lib/pcm_stream_channel_modifier/         @koffes @alexsven @erikrobstad @rick1082 @gWacey
 /lib/tone/                                @koffes @alexsven @erikrobstad @rick1082 @gWacey
+/lib/sample_rate_converter/               @andvib @gWacey
```
katgiadla, line 141: If you are codeowner for module, please add your responsibility for tests too.

katgiadla, line 141: If you are codeowner for module, please add your responsibility for tests too.

katgiadla (changes requested):

lib/sample_rate_converter/sample_rate_converter.c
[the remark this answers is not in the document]
andvib: After offline discussion I will try to implement the ratio as a signed value

lib/sample_rate_converter/Kconfig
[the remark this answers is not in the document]
andvib: Will find a way to do this in code and not kconfig

katgiadla (approved):

CODEOWNERS
```diff
@@ -295,6 +296,7 @@ Kconfig*                                  @tejlmand
 /tests/lib/pcm_mix/                       @koffes @alexsven @erikrobstad @rick1082 @gWacey
 /tests/lib/pcm_stream_channel_modifier/   @koffes @alexsven @erikrobstad @rick1082 @gWacey
 /tests/lib/tone/                          @koffes @alexsven @erikrobstad @rick1082 @gWacey
+/tests/lib/sample_rate_converter/         @andvib @gWacey
```
katgiadla, line 299: please keep the alphabetical order of the modules and tests

katgiadla (changes requested):

lib/sample_rate_converter/sample_rate_converter_filter.c
```diff
@@ -0,0 +1,131 @@
...
+
+int sample_rate_converter_filter_get(enum sample_rate_converter_filter filter_type,
+				     uint8_t conversion_ratio, void **filter_ptr,
+				     size_t *filter_size)
+{
+
+	__ASSERT(filter_ptr != NULL, "Filter pointer cannot be NULL");
+	__ASSERT(filter_size != NULL, "Filter size pointer cannot be NULL");
+
+	/* Filters only support conversion between 16kHz, 14kHz and 48kHz, which means that the
```
alexsven, line 63:
```suggestion
	/* Filters only support the following conversions:
	 * 16 <--> 48
	 * 24 <--> 48 
	 * which means that the
```

alexsven (changes requested):

alexsven (approved):

CODEOWNERS
andvib: I'm not exactly sure how to follow the structure as most entries aren't alphabetical, but at least I moved the entry above `tone` so that it's sorted within the Octave libs

katgiadla (approved):

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1155 @@
...
+	zassert_equal(conv_ctx.filter_type, filter, "Filter not as expected");
+	zassert_equal(conv_ctx.input_buf.bytes_in_buf, 0, "Bytes in input buffer not as expected");
+	zassert_equal(ring_buf_size_get(&conv_ctx.output_ringbuf), 0,
+		      "Number of bytes in output ringbuffer not as expected %d",
+		      ring_buf_size_get(&conv_ctx.output_ringbuf));
+
+	zassert_equal(output_written, expected_output_samples * sizeof(uint16_t),
+		      "Received %d output samples when none was expected", output_written);
+
+	zassert_within(output_samples[0], input_samples[0], 550);
```
koffes, line 401: These are still 550.

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -0,0 +1,1155 @@
...
```
koffes, line 442: 550

tests/lib/sample_rate_converter/prj.conf
```diff
@@ -2,7 +2,6 @@ CONFIG_ZTEST=y
 CONFIG_MAIN_STACK_SIZE=8192
 CONFIG_ZTEST_STACK_SIZE=8192
 CONFIG_SAMPLE_RATE_CONVERTER=y
-CONFIG_SAMPLE_RATE_CONVERTER_FILTER_SIMPLE=y
-CONFIG_SAMPLE_RATE_CONVERTER_FILTER_SMALL=y
+CONFIG_SAMPLE_RATE_CONVERTER_FILTER_TEST=y
```
koffes, line 5: Nice, but are we preventing use of the non-TEST-filters?

lib/sample_rate_converter/sample_rate_converter_filter.c
```diff
@@ -7,21 +7,30 @@
...
 
-static q31_t filter_48khz_to_24khz_32bit_simple[] = {0x3fffffff, 0x40000000};
+static q31_t filter_48khz_to_24khz_32bit_test[] = {0x3fffffff, 0x40000000};
+
+static q31_t filter_24khz_to_48khz_32bit_test[] = {0x7fffffff, 0x7fffffff};
+
+static q31_t filter_48khz_to_16khz_32bit_test[] = {0x2aaaaaaa, 0x2aaaaaaa, 0x2aaaaaab};
 
-static q31_t filter_48khz_to_16khz_32bit_simple[] = {0x2aaaaaaa, 0x2aaaaaaa, 0x2aaaaaab};
+static q31_t filter_16khz_to_48khz_32bit_test[] = {0x7fffffff, 0x7fffffff, 0x7fffffff};
```
koffes, line 31: Let's have a word about the sum of these filters.

koffes (changes requested):

koffes (approved):

PR #13482, opened 2023-12-20
applications: nrf5340_audio: Broadcast stop scan improvement

erikrobstad:
Moved bt_le_scan_stop()


erikrobstad:
@Thalley Please have a look at this when you have time.
This was something we changed during the hackathon.

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
```diff
@@ -46,6 +46,11 @@ static void scan_restart_worker(struct k_work *work)
 {
 	int ret;
 
+	ret = bt_le_scan_stop();
+	if (ret) {
+		LOG_WRN("Stop scan failed: %d", ret);
+	}
```
Thalley, line 52: Since we are starting scan right after, is there a reason why we stop and restart scanning? 

i.e. could this be removed together with 
```c
ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_BROADCAST, NULL);
```

Thalley (commented):
Changes look good, but the commit should address why it is being moved. 

IIRC the reason we moved it, was to control the scan parameters in the application, rather than the implicit scanning from calling `bt_le_per_adv_sync_create`, and since we are now using explicit scanning, it is necessary to also explicitly stop it.

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/bt_mgmt_scan_for_broadcast.c
erikrobstad: I discussed this with the team.
The reason why we do bt_mgmt_scan_start() here is to re-add the scan callbacks. And that is why scanning is stopped first.

The way we now register and unregister the scan callbacks and use scan_cb_registered for "checking if we are scanning or not" might not be optimal. This could potentially be improved, maybe in this PR? @alexsven @koffes @rick1082 

I will be leaving it as is, for now at least.

alexsven (approved):

PR #13484, opened 2023-12-20
nrf_security: Add Kconfig for SSL max fragment

Vge0rge:
Add Kconfig option for MBEDTLS_SSL_MAX_FRAGMENT_LENGTH since we have it as configurable in the configuration file but there was way of enabling it.


SebastianBoe (approved):

maciejbaczmanski (approved):
Thanks!

mswarowsky (approved):

PR #13485, opened 2023-12-20
samples: matter: common: Cleaned up samples common directory

LipinskiPNordicSemi:
This commit:
-  creates directories for commons in the samples matter.
- fix issue for matter bridge binding handler (one common BH used for every sample)


west.yml
```diff
@@ -149,7 +149,7 @@ manifest:
     - name: matter
       repo-path: sdk-connectedhomeip
       path: modules/lib/matter
-      revision: d95eb1d43d6e8247045d549802642ec66ad1d322
+      revision: pull/370/head
```
Damian-Nordic, line 152: This is not required by this change, right?

Damian-Nordic (approved):

samples/matter/window_covering/CMakeLists.txt
```diff
@@ -43,19 +43,19 @@ target_sources(app PRIVATE
     src/zcl_callbacks.cpp
     src/zap-generated/IMClusterCommandHandler.cpp
     src/zap-generated/callback-stub.cpp
-    ${COMMON_ROOT}/src/led_widget.cpp
-    ${COMMON_ROOT}/src/pwm_device.cpp
-    ${COMMON_ROOT}/src/task_executor.cpp
-    ${COMMON_ROOT}/src/board.cpp
+    ${COMMON_ROOT}/src/led/led_widget.cpp
+    ${COMMON_ROOT}/src/pwm/pwm_device.cpp
```
ArekBalysNordic, line 47: What about Matter Bridge and Weather Station applications? :)

ArekBalysNordic (changes requested):

samples/matter/lock/src/app_task.cpp
```diff
@@ -5,17 +5,17 @@
...
 #include "app_task.h"
-#include "task_executor.h"
+#include "tasks/task_executor.h"
 
 #include "bolt_lock_manager.h"
 #include "fabric_table_delegate.h"
 
 #ifdef CONFIG_MCUMGR_TRANSPORT_BT
-#include "dfu_over_smp.h"
+#include "dfu/dfu_over_smp.h"
```
markaj-nordic, line 14:
```suggestion
#include "dfu/smp/dfu_over_smp.h"
```

markaj-nordic (changes requested):
Please align Matter WS and Bridge apps.

applications/matter_bridge/CMakeLists.txt
```diff
@@ -47,13 +47,13 @@ target_sources(app PRIVATE
...
     src/zap-generated/IMClusterCommandHandler.cpp
     src/zap-generated/callback-stub.cpp
-    ${COMMON_ROOT}/src/led_widget.cpp
-    ${COMMON_ROOT}/src/persistent_storage_util.cpp
-    ${COMMON_ROOT}/src/task_executor.cpp
-    ${COMMON_ROOT}/src/board.cpp
+    ${COMMON_ROOT}/src/led/led_widget.cpp
+    ${COMMON_ROOT}/src/ps_storage/persistent_storage_util.cpp
+    ${COMMON_ROOT}/src/tasks/task_executor.cpp
+    ${COMMON_ROOT}/src/board/board.cpp
```
ArekBalysNordic, line 56: These changes should be added to the [samples: matter: common: Cleaned up samples common directory](https://github.com/nrfconnect/sdk-nrf/pull/13485/commits/4b0337997497c09ea2ce1d1cc3e80ea87100edbe) commit because they are not related to common binding manager.

In this situation, there is a risk that someone may checkout only the "clean up samples" commit and the bridge won't work. This is not so important, because the main branch is experimental anyway and customers should use main releases, but still, if we can do something to keep our devices working commit by commit we should do it. :)

samples/matter/common/src/bridge/bridge_storage_manager.h
```diff
@@ -7,7 +7,7 @@
 #pragma once
 
 #include "matter_bridged_device.h"
-#include "persistent_storage_util.h"
+#include "ps_storage/persistent_storage_util.h"
```
ArekBalysNordic, line 10: Regarding the commit description, I can see the commit doesn't implement anything new. I guess, you've used the previously created common binding manager/handler in the Matter Bridge, right? :)

kkasperczyk-no (approved):

applications/matter_bridge/CMakeLists.txt
```diff
@@ -47,13 +47,13 @@ target_sources(app PRIVATE
...
     ${COMMON_ROOT}/src/bridge/bridged_device_data_provider.cpp
-    ${COMMON_ROOT}/src/bridge/binding_handler.cpp
+    ${COMMON_ROOT}/src/binding/binding_handler.cpp
     src/zap-generated/IMClusterCommandHandler.cpp
     src/zap-generated/callback-stub.cpp
-    ${COMMON_ROOT}/src/led_widget.cpp
-    ${COMMON_ROOT}/src/persistent_storage_util.cpp
-    ${COMMON_ROOT}/src/task_executor.cpp
-    ${COMMON_ROOT}/src/board.cpp
+    ${COMMON_ROOT}/src/led/led_widget.cpp
```
markaj-nordic, line 53: This commit mixes cleanup of the `common` directory and utilization of the common binding handler in the bridge application. I don't know what's other opinion but to me it should be fixed to keep the git history clean (@kkasperczyk-no, @ArekBalysNordic ?). I also suggest to change the commit message: "...Implement common binding manager" to "...Use common binding handler"

markaj-nordic: EDIT: Ok I see Arek shares the same concerns I do, so this definitely must be fixed :)

ArekBalysNordic: Yeah, I agree with Kamil, as far as we as Nordic developers are aware of that because sometimes we checkout between different commits.

LipinskiPNordicSemi: Sorry for that mess, by mistake I've added a whole bridge to one commit. It should be cleaner right now

ArekBalysNordic (approved):

ArekBalysNordic (approved):

markaj-nordic (approved):

PR #13487, opened 2023-12-20
ci: use composite checkout action

karhama:
Use separate action for checking out the source code. Composite action can be found from here: https://github.com/NordicBuilder/action-checkout-west-update/

### DIFF
--- a/.github/workflows/license-reusable.yml
+++ b/.github/workflows/license-reusable.yml
@@ -18,7 +18,6 @@ name: Reusable License Check
 on:
   workflow_call:
     inputs:
-
       path:
         # Required path to your module relative to west workspace.
         type: string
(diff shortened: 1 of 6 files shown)


.github/workflows/license-reusable.yml
```diff
@@ -70,32 +59,24 @@ jobs:
...
       working-directory: ncs
       env:
         PR_REF: ${{ github.event.pull_request.head.sha }}
       run: |
         export PATH="$HOME/.local/bin:$PATH"
         export PATH="$HOME/bin:$PATH"
-        if [ "${{ inputs.path }}" = "nrf" ]; then
-            west init -l nrf
-        else
-            west init -m ${{ inputs.nrf_repo }} --mr ${{ inputs.nrf_rev }}
```
DatGizmo, line 96: It looks like the new action does not support this use case.

karhama: Correct. I tried to search if these inputs are actually used somewhere but didn't find hits (but I might have missed something).

.github/workflows/oss-history.yml
```diff
@@ -6,22 +6,20 @@ jobs:
...
-        uses: actions/checkout@v3
+      - name: Checkout sources
+        uses: nordicbuilder/action-checkout-west-update@main
         with:
-          path: ncs/nrf
-          fetch-depth: 0
+          git-fetch-depth: 0
+          west-update-args: ''
 
       - name: Install extra python dependencies
```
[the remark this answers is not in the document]
thst-nordic: pip doesn't read anything except 'nrf' which is same order

thst-nordic (approved):

.github/workflows/license-reusable.yml
DatGizmo: Checked as well and couldn't find it being used.

DatGizmo (approved):

PR #13491, opened 2023-12-21
applications: nrf5340_audio: Fix volume up to maximum cause volume sa…

rick1082:
…turation issue

Fixed volume up to maximum volume cause saturation.

### DIFF
--- a/applications/nrf5340_audio/src/modules/hw_codec.c
+++ b/applications/nrf5340_audio/src/modules/hw_codec.c
@@ -202,7 +202,7 @@ int hw_codec_volume_adjust(int8_t adjustment_db)
 		if (new_volume_reg_val <= 0) {
 			LOG_WRN("Volume at MIN (-64dB)");
 			new_volume_reg_val = 0;
-		} else if (volume_reg_val >= MAX_VOLUME_REG_VAL) {
+		} else if (new_volume_reg_val >= MAX_VOLUME_REG_VAL) {
 			LOG_WRN("Volume at MAX (0dB)");
 			new_volume_reg_val = MAX_VOLUME_REG_VAL;
 		}


applications/nrf5340_audio/src/modules/hw_codec.c
```diff
@@ -202,7 +202,7 @@ int hw_codec_volume_adjust(int8_t adjustment_db)
 		if (new_volume_reg_val <= 0) {
 			LOG_WRN("Volume at MIN (-64dB)");
 			new_volume_reg_val = 0;
-		} else if (volume_reg_val >= MAX_VOLUME_REG_VAL) {
+		} else if (volume_reg_val > MAX_VOLUME_REG_VAL) {
```
koffes, line 205: What does this fix?

Let's say you move from volume 0x78 to 0x80 (MAX_VOLUME_REG_VAL). 
Earlier, you would get a WRN print as this happened. Now, you will get the print if you try to increase beyond this limit.

koffes (changes requested):

applications/nrf5340_audio/src/modules/hw_codec.c
rick1082: If we keep push buttons to reach the maximum volume, then we cannot low down the volume.
Since in that condition, `volume_reg_val` is larger than `MAX_VOLUME_REG_VAL`, and we will jump into the condition for setting `new_volume_reg_val = MAX_VOLUME_REG_VAL;`, and then "new_volume_reg_val" will be trapped in here,

alexsven: I believe the correct solution to this bug should be:
```suggestion
		} else if (new_volume_reg_val >= MAX_VOLUME_REG_VAL) {
```

alexsven (changes requested):

alexsven (approved):

applications/nrf5340_audio/src/modules/hw_codec.c
```diff
@@ -202,7 +202,7 @@ int hw_codec_volume_adjust(int8_t adjustment_db)
 		if (new_volume_reg_val <= 0) {
 			LOG_WRN("Volume at MIN (-64dB)");
 			new_volume_reg_val = 0;
-		} else if (volume_reg_val >= MAX_VOLUME_REG_VAL) {
+		} else if (new_volume_reg_val >= MAX_VOLUME_REG_VAL) {
 			LOG_WRN("Volume at MAX (0dB)");
 			new_volume_reg_val = MAX_VOLUME_REG_VAL;
 		}
```
koffes, line 208: This whole section should be moved further down to before we set the volume

koffes (approved):

PR #13495, opened 2023-12-21
Make TF-M more portable

SebastianBoe:
Make TF-M more portable


mswarowsky (approved):

modules/trusted-firmware-m/tfm_boards/board/RTE_Device.h
```diff
@@ -8,45 +8,37 @@
...
+		DT_PINCTRL_BY_NAME(DT_NODELABEL(uart##idx), default, 0), \
 		DT_FOREACH_PROP_ELEM, psels, UART_PIN_INIT \
 	) \
 }
 
-#endif /* defined(CONFIG_TFM_SECURE_UART1) */
-
-/* Configuration settings for Driver_FLASH0. */
-#define RTE_FLASH0 1
+#define RTE_USART0_PINS RTE_USART_PINS(0)
```
Vge0rge, line 41: Is that correct? Before the create of the nodes were ifdefed by the TFM_SECURE_UART0/1 , right? 

Now you generate both of the nodes which also same the same names and idx:
DT_PINCTRL_BY_NAME(node_id, name, idx)

Does that make sense to define these inside the ifdefs?

SebastianBoe: discussed offline

Vge0rge (approved):

PR #13496, opened 2023-12-21
kconfig: openthread: enable `MBEDTLS_SSL_MAX_FRAGMENT_LENGTH`

maciejbaczmanski:
enable `MBEDTLS_SSL_MAX_FRAGMENT_LENGTH` for configurations that use DTLS


edmont (approved):
Is there maybe some other config from https://github.com/openthread/openthread/blob/main/third_party/mbedtls/mbedtls-config.h needed?

maciejbaczmanski:
> Is there maybe some other config from https://github.com/openthread/openthread/blob/main/third_party/mbedtls/mbedtls-config.h needed?

all remaining configs that are not enabled in PSA implementation but enabled in OT's configuration have been checked in KRKNWK-18230:
- `MBEDTLS_HAVE_ASM`  is unnecessary, as it is about the built-in crypto which we don't use
- `MBEDTLS_SHA256_SMALLER` has been disabled in non-psa configuration. Also, it is used only by legacy mbedtls implementation, now replaced by PSA API
`- MBEDTLS_SSL_MAX_FRAGMENT_LENGTH` to be added (this PR)
- `MBEDTLS_HMAC_DRBG_C` legacy random APIS have been disabled by psa_generate_random, so this is not needed anymore
- `MBEDTLS_ECP_NIST_OPTIM` is needed only for legacy APIs (MBEDTLS_ECP_C)

canisLupus1313 (approved):

PR #13498, opened 2023-12-21
applications: asset_tracker_v2: Fix air pressure value

tokangas:
Air pressure value was scaled incorrectly and the values sent to the cloud were not in Pascals. When the BSEC library is used, the value is already in Pascals and no scaling is needed, however the BME680 driver reports the pressure in kPascals.


tokangas:
@maxd-nordic @jtguggedal @simensrostad I noticed that the reported air pressure values with Thingy:91 default configuration were something like 0.1 Pascals, when the correct value should be around 100000 Pascals. It seems that the unit depends on the driver configuration, but the implementation in the PR works correctly with both BME680 drivers (as far as I can tell). The old implementation reported incorrect values with both.

simensrostad (approved):
:+1:

jtguggedal (approved):

PR #13507, opened 2023-12-22
lib: nrf_modem_lib: lte_net_if: Add handling of reset loop event

simensrostad:
Add handling of reset loop event to notify users when this occurs.

### DIFF
--- a/lib/nrf_modem_lib/lte_net_if/lte_net_if.c
+++ b/lib/nrf_modem_lib/lte_net_if/lte_net_if.c
@@ -302,29 +302,33 @@ static void pdn_event_handler(uint8_t cid, enum pdn_event event, int reason)
 
 static void lte_reg_handler(const struct lte_lc_evt *const evt)
 {
-	if (evt->type != LTE_LC_EVT_NW_REG_STATUS) {
-		return;
-	}
+	if (evt->type == LTE_LC_EVT_MODEM_EVENT && evt->modem_evt == LTE_LC_MODEM_EVT_RESET_LOOP) {
+		LOG_WRN("The modem has detected a reset loop. LTE network attach is now "
+			"restricted for the next 30 minutes.");
 
-	switch (evt->nw_reg_status) {
-	case LTE_LC_NW_REG_REGISTERED_HOME:
-		__fallthrough;
-	case LTE_LC_NW_REG_REGISTERED_ROAMING:
-		/* Mark serving cell as available. */
-		LOG_DBG("Registered to serving cell");
-		update_has_cell(true);
-		break;
-	case LTE_LC_NW_REG_SEARCHING:
-		/* Searching for a new cell, do not consider this cell loss unless it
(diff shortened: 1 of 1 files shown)


gregersrygg (approved):

lib/nrf_modem_lib/lte_net_if/lte_net_if.c
```diff
@@ -301,29 +301,36 @@ static void pdn_event_handler(uint8_t cid, enum pdn_event event, int reason)
...
-	default:
-		LOG_DBG("Not registered to serving cell");
-		/* Mark the serving cell as lost. */
-		update_has_cell(false);
-		break;
+	if (evt->type == LTE_LC_EVT_MODEM_EVENT && evt->modem_evt == LTE_LC_MODEM_EVT_RESET_LOOP) {
+		LOG_WRN("The modem has detected a reset loop. LTE network attach is now "
+			"restricted for the next 30 minutes. Power-cycle the device to "
+			"circumvent this restriction or reset the modem to factory settings by "
+			"issuing the AT\%XFACTORYRESET AT command.");
```
eivindj-nordic, line 308: I'm not sure if this is something we want to state as a solution for a reset loop. It is good as a workaround for development, though I'm afraid someone might get the idea to add the factory reset to production code from this, causing a live device to continue spamming attach requests to the network to "eternity".

A suggestion is to keep the first sentence and remove the second.

simensrostad: Fair point, I can remove the last part of the sentence.

simensrostad: Removed

eivindj-nordic (approved):

lib/nrf_modem_lib/lte_net_if/lte_net_if.c
```diff
@@ -302,29 +302,34 @@ static void pdn_event_handler(uint8_t cid, enum pdn_event event, int reason)
...
-		/* Mark the serving cell as lost. */
-		update_has_cell(false);
-		break;
+	if (evt->type == LTE_LC_EVT_MODEM_EVENT && evt->modem_evt == LTE_LC_MODEM_EVT_RESET_LOOP) {
+		LOG_WRN("The modem has detected a reset loop. LTE network attach is now "
+			"restricted for the next 30 minutes.");
+
+		LOG_DBG("For more information see the nRF91 AT Commands - Command "
+			"Reference Guide v2.0 - chpt. 5.36");
+
```
tokangas, line 311: I think this should be made more generic, because we now have separate AT command documentation for nRF9160 and nRF91x1. Also, it would be better not to refer to any specific document version.

How about something like this?
```suggestion
		LOG_DBG("For more information, see the AT command documentation for %MDMEV notification");
```

bjda (approved):

tokangas (approved):

lemrey (approved):

PR #13515, opened 2023-12-29
samples: cellular: modem_shell: Take Custom AT commands library into use

tokangas:
Refactored the custom AT command implementation to use the Custom AT commands libary and underlying Modem library implementation.

MOSH-259


samples/cellular/modem_shell/src/at/at_cmd_mode_custom.c
```diff
@@ -0,0 +1,124 @@
...
+
+#if defined(CONFIG_MOSH_PING)
+#include "icmp_ping.h"
+#endif
+
+#define MOSH_AT_CMD_MAX_PARAM 8
+
+extern bool at_cmd_mode_dont_print;
+
+#if defined(CONFIG_MOSH_PING)
```
trantanen, line 27: I know this is in old code but would probably be cleaner to exclude the entire file in `CMakeLists.txt`

eivindj-nordic (approved):
Looks good from modem lib/custom AT commands point of view.

samples/cellular/modem_shell/src/at/at_cmd_mode_custom.c
```diff
@@ -0,0 +1,124 @@
+/*
+ * Copyright (c) 2022 Nordic Semiconductor ASA
```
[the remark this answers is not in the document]
tokangas: The file was renamed, so I kept the old copyright year. Not sure if that was the right thing to do or not.

samples/cellular/modem_shell/src/at/at_cmd_mode_custom.c
tokangas: I wanted to do it like this because this module is not ping specific. This module can contain handlers for different custom AT commands. It just happens to contain only ping at the moment. 😄

trantanen (approved):

PR #13512, opened 2023-12-29
wifi: AP connection support

krish2718:

PR #13518, opened 2024-01-02
manifest: sdk-hostap: Pull fix for shell freeze

krish2718:
Fixes the shell freeze seen when WPA supplicant OOM causes an infinite loop.


sachinthegreen (approved):

rado17 (approved):

krish2718:
manifest will be updated as part of #13512

PR #13519, opened 2024-01-02
net: lib: nrf_cloud_pgps: Fix bug in prediction set update

tokangas:
Fixed a bug which caused the prediction set update to fail. The failure happened when all downloaded predictions could not be written to flash sequentially.

NCSDK-25224

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -671,6 +671,10 @@ Libraries for networking
     * The :c:func:`nrf_cloud_obj_object_add` function to reset the added object on success.
     * Custom shadow data is now passed to the application during shadow update events.
 
+* :ref:`lib_nrf_cloud_pgps` library:
+
+  * Fixed a bug in prediction set update when the :kconfig:option:`CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD` Kconfig option was set to non-zero value.
+
 * :ref:`lib_nrf_provisioning` library:
 
   * Renamed nRF Device provisioning to :ref:`lib_nrf_provisioning`.
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
@@ -78,7 +78,7 @@ struct pgps_index {
 	uint8_t cur_pnum;
 	bool partial_request;
 	bool stale_server_data;
-	uint32_t storage_extent;
+	int32_t storage_extent;
 	int store_block;
 
 	/* Array of memory offsets to predictions, in sorted time order.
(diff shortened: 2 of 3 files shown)


tokangas:
I also added a couple of error checks. Previously, the library continued to the end of the download as if nothing had happened, because the error from `store_prediction()` was ignored.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -645,6 +645,10 @@ Libraries for networking
     * The :c:func:`nrf_cloud_obj_object_add` function to reset the added object on success.
     * Custom shadow data is now passed to the application during shadow update events.
 
+* :ref:`lib_nrf_cloud_pgps` library:
+
+  * Fixed a bug in prediction set update when :kconfig:option:`CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD` Kconfig option was set to non-zero value.
```
peknis, line 650:
```suggestion
  * Fixed a bug in prediction set update when the :kconfig:option:`CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD` Kconfig option was set to non-zero value.
```

peknis (approved):
Approved with a minor nit.

subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
```diff
@@ -1557,6 +1561,10 @@ static int consume_pgps_data(uint8_t pnum, const char *buf, size_t buf_len)
 					LOG_ERR("Error opening storage again:%d", err);
 					goto fail;
 				}
+			} else if (index.storage_extent < 0) {
+				LOG_ERR("Unexpected storage extent");
```
jayteemo, line 1565:
```suggestion
				LOG_ERR("Unexpected storage extent:%d", index.storage_extent);
```

subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps_utils.c
```diff
@@ -328,16 +328,21 @@ void npgps_free_block(int block)
...
 	int i;
 	int len = 0;
+	int current_block = store_block;
 
 	for (i = 0; i < num_blocks; i++) {
-		if (pool.block_used[block]) {
+		/* The first block has already been allocated for storage, so the usage check
+		 * is skipped.
+		 */
+		if (current_block != store_block &&
```
jayteemo, line 341:
```suggestion
		if ((current_block != store_block) &&
```

jayteemo (approved):

glarsennordic (approved):

tokangas:
We did a test run with the fix and `npgps_get_block_extent()` still didn't return the correct number of blocks in all cases. I rewrote the function how I think it should work. It should not wrap-around, but calculate the number of blocks that can be written before the flash storage needs to be re-opened from the beginning of the flash area.

I'll be on vacation until 17th of January. @plskeggs, please remove the DNM label if the fix is fine.

plskeggs (approved):

PR #13524, opened 2024-01-02
slm: preliminary CMUX implementation

tomi-font:
2 channels are configured and simply echo what is received.

### DIFF
--- a/applications/serial_lte_modem/src/slm_cmux.c
+++ b/applications/serial_lte_modem/src/slm_cmux.c
@@ -1,0 +1,187 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+#include "slm_cmux.h"
+#include "slm_at_host.h"
+#include "slm_uart_handler.h"
+#include <zephyr/logging/log.h>
+#include <zephyr/modem/backend/uart.h>
+#include <zephyr/modem/cmux.h>
+#include <zephyr/modem/pipe.h>
+#include <assert.h>
+
+LOG_MODULE_REGISTER(slm_cmux, CONFIG_SLM_LOG_LEVEL);
+
+/* This makes use of some of the Zephyr modem subsystem that has a CMUX module implementation.
+ */
+
+static struct {
+	/* UART backend */
+	struct modem_pipe *uart_pipe;
+	struct modem_backend_uart uart_backend;
+	uint8_t uart_backend_receive_buf[512];
+	uint8_t uart_backend_transmit_buf[512];
+
+	/* CMUX */
+	struct modem_cmux instance;
+	uint8_t cmux_receive_buf[128];
+	uint8_t cmux_transmit_buf[256];
+
+	/* CMUX channels (Data Link Connection Identifier); index = address - 1 */
+	struct cmux_dlci {
+		struct modem_cmux_dlci instance;
+		struct modem_pipe *pipe;
+		uint8_t *recv_buf;
+		uint16_t recv_buf_size;
+		uint8_t address;
+	} dlcis[2];
+} cmux;
+
+static void dlci_pipe_event_handler(struct modem_pipe *pipe,
+				    enum modem_pipe_event event, void *user_data)
+{
+	const struct cmux_dlci *const dlci = user_data;
+
+	switch (event) {
+	case MODEM_PIPE_EVENT_OPENED:
+	case MODEM_PIPE_EVENT_CLOSED:
+		LOG_INF("DLCI %u %s.", dlci->address,
+			(event == MODEM_PIPE_EVENT_OPENED) ? "opened" : "closed");
+		break;
+	case MODEM_PIPE_EVENT_RECEIVE_READY:
+		static uint8_t recv_buf[256];
+		const int recv_len = modem_pipe_receive(pipe, recv_buf, sizeof(recv_buf));
+
+		if (recv_len > 0) {
+			LOG_DBG("Received %u bytes on DLCI %u.", recv_len, dlci->address);
+			modem_pipe_transmit(pipe, recv_buf, recv_len);
+		} else {
+			LOG_ERR("Failed to receive data from DLCI %u (%d).",
+				dlci->address, recv_len);
+		}
+		break;
+	}
+}
+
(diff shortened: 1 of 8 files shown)


applications/serial_lte_modem/src/slm_cmux.c
```diff
@@ -0,0 +1,187 @@
...
+LOG_MODULE_REGISTER(slm_cmux, CONFIG_SLM_LOG_LEVEL);
+
+/* This makes use of some of the Zephyr modem subsystem that has a CMUX module implementation.
+ */
+
+static struct {
+	/* UART backend */
+	struct modem_pipe *uart_pipe;
+	struct modem_backend_uart uart_backend;
+	uint8_t uart_backend_receive_buf[512];
```
MarkusLassila, line 24: Any reason to have these buffers as configurable? Also, these could perhaps utilize the UART buffers in future.

applications/serial_lte_modem/src/slm_cmux.c
```diff
@@ -0,0 +1,187 @@
...
+
+	modem_pipe_attach(dlci->pipe, dlci_pipe_event_handler, dlci);
+}
+
+static int cmux_start(void)
+{
+	int ret;
+
+	/* Disable SLM UART so that CMUX can take over it. */
+	ret = slm_uart_rx_disable();
```
MarkusLassila, line 99: Note: This does prevent TX, so something that is already running could theoretically be sending. Given that you have disabled most of the clients with PPP, that is unlikely to be a problem. But something to stay aware. 

Mechanism to disable TX would be reasonably simple, but unnecessary unless this is observed to cause problems.

applications/serial_lte_modem/src/slm_cmux.c
```diff
@@ -0,0 +1,187 @@
...
+	ret = modem_pipe_open(cmux.uart_pipe);
+	return ret;
+}
+
+static void cmux_starter(struct k_work *)
+{
+	const int ret = cmux_start();
+
+	if (ret) {
+		LOG_ERR("Failed to start CMUX (%d).", ret);
```
MarkusLassila, line 151: Should likely enable uart_rx here as otherwise the device will stay unresponsive.

applications/serial_lte_modem/src/slm_cmux.c
```diff
@@ -0,0 +1,187 @@
...
+	/* Disable SLM UART so that CMUX can take over it. */
+	ret = slm_uart_rx_disable();
+	if (ret) {
+		return ret;
+	}
+
+	{
+		const struct modem_backend_uart_config uart_backend_config = {
+			.uart = DEVICE_DT_GET(DT_CHOSEN(ncs_slm_uart)),
+			.receive_buf = cmux.uart_backend_receive_buf,
```
MarkusLassila, line 107: Does this use ISR or async uart? I am wondering as with async uart, there should be several buffers, or the performance is going to suffer drastically.

applications/serial_lte_modem/src/slm_cmux.c
```diff
@@ -0,0 +1,187 @@
...
+}
+
+/* Handles AT#CMUX commands. */
+int handle_at_cmux(enum at_cmd_type cmd_type)
+{
+	static struct k_work_delayable cmux_start_work;
+	int ret;
+	unsigned int op;
+	enum {
+		OP_START,
```
MarkusLassila, line 164: Did I understand correctly that in future you should be able to still send the STOP AT-command via CMUX:ed UART? Or how is this going to be stopped?

MarkusLassila (approved):
Looks like a fine start of implementation. My comments can be addressed in future PR's. Just wanted to point them out.

applications/serial_lte_modem/src/slm_cmux.c
tomi-font: The (end) goal is to have both RX and TX go through CMUX. As of this PR the TX part is indeed not handled, as this is just echoing back what is received.
Though maybe `slm_uart_rx_disable()` should be renamed? Do you have a suggestion?

applications/serial_lte_modem/src/slm_cmux.c
tomi-font: Async. Internally ring buffers are used, so I imagine that it's equivalent to using several buffers?

applications/serial_lte_modem/src/slm_cmux.c
tomi-font: Correct. This is actually something I am currently working on.

applications/serial_lte_modem/src/slm_cmux.c
tomi-font: Possibly. I am a little bit confused regarding this because the 3GPP spec for the `+CMUX` command says
> This command is used to enable/**disable** the 3GPP TS 27.010 [45] multiplexing protocol control channel.

But it doesn't clearly define how to stop CMUX, and even states
> If a +CMUX command is issued whilst in any multiplexer mode then that +CMUX command shall be ignored and the MT/TA shall return a +CME ERROR: <err> response.

Though we may want to be able to stop CMUX regardless of the (admittedly confusing) official spec.
Currently the only way is via an `AT#XRESET`.

SeppoTakalo (approved):

applications/serial_lte_modem/src/slm_cmux.c
MarkusLassila: I am not sure what they have done internally, but the idea is to have several buffers so that uart receiving can continue while you process the other buffer. Otherwise this will always stop the uart from receiving when you are processing the buffer.

applications/serial_lte_modem/src/slm_cmux.c
MarkusLassila: I think that the name is in line what it does, so no need to remove at this point.

applications/serial_lte_modem/src/slm_cmux.c
tomi-font: I checked, and internally the async UART backend splits the supplied buffer into several sub-buffers (in 4; 2 buffers and 2 ring buffers).
Also, there is actually quite a bit of copying happening between the modules (which is why we have so many buffers declared), so I think that by the time data goes through a `modem_pipe` the end that transmitted it (e.g. the async UART backend when receiving) doesn't need to keep the data anymore even though it may not even have reached the application yet.

PR #13528, opened 2024-01-03
applications: Added BT security support for the Matter bridge

kkasperczyk-no:
Extended the Matter bridge BLE connectivity manager by the support for BT SMP and allowing to establish secure connections with the Bluetooth LE bridged devices.


kkasperczyk-no:
The documentation is missing yet, but I think it's still worth to do the first round of code review.

kkasperczyk-no:
The docs were added as well.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -353,6 +387,32 @@ Build the target using the following command in the project directory to enable
...
+* Security Level 4 - supports ECDHE communication encryption, requires authentication and pairing.
+
+By default, the Matter bridge device has SMP enabled and supports security levels 2, 3 and 4.
+
+You can disable the Bluetooth LE security mechanisms by setting :kconfig:option:`CONFIG_BT_SMP` to ``n``.
+However it is strongly not recommended configuration, as it leads to unencrypted communication with bridged devices, which makes them vulnerable for the security attacks.
+
+You can select the minimum security level required by you application.
+It means the Matter bridge will require setting the selected minimum level from the connected Bluetooth LE bridged device.
+If the bridged device supports minimum criteria and also one of the higher levels, the devices may negotiate using the highest security level possible to achieve.
```
wiba-nordic, line 412:
```suggestion
If the bridged device supports also levels higher than the selected minimum, the devices may negotiate using the highest shared security level.
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -353,6 +387,32 @@ Build the target using the following command in the project directory to enable
...
+In case the bridged device does not support minimum required level, the connection will be terminated.
```
wiba-nordic, line 413:
```suggestion
In case the bridged device does not support the minimum required level, the connection will be terminated.
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -353,6 +387,32 @@ Build the target using the following command in the project directory to enable
...
+To select the minimum security level, set :kconfig:option:`CONFIG_BRIDGE_BT_MINIMUM_SECURITY_LEVEL` to ``2``, ``3`` or ``4``.
```
wiba-nordic, line 414:
```suggestion
To select the minimum security level, set the :kconfig:option:`CONFIG_BRIDGE_BT_MINIMUM_SECURITY_LEVEL` Kconfig option to ``2``, ``3`` or ``4``.
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -496,6 +557,49 @@ After building the sample and programming it to your development kit, complete t
...
+
+         #. Write down the authentication pincode value from the Bluetooth LE bridged device terminal.
+            The terminal output is similar to the following one:
+
+            .. parsed-literal::
+               :class: highlight
+
+               Passkey for FD:D6:53:EB:92:3A (random): 350501
+
+            The displayed pincode value is equal to ``350501`` in the presented example.
```
wiba-nordic, line 578:
```suggestion
            In the above example output, the displayed pincode value is  ``350501``.
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -496,6 +557,49 @@ After building the sample and programming it to your development kit, complete t
...
+            It will be used in the next steps as *<bluetooth_authentication_pincode>*.
+         #. Insert the authentication pincode of a bridged device in the Matter bridge terminal.
```
wiba-nordic, line 580:
```suggestion
         #. Insert the authentication pincode of the bridged device in the Matter bridge terminal.
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -496,6 +557,49 @@ After building the sample and programming it to your development kit, complete t
...
+            To do that, run the following :ref:`Matter CLI command <matter_bridge_cli>` with *<bluetooth_authentication_pincode>* replaced by the value read in the previous step:
```
wiba-nordic, line 581:
```suggestion
            To insert the pincode, run the following :ref:`Matter CLI command <matter_bridge_cli>` with *<bluetooth_authentication_pincode>* replaced by the value read in the previous step:
```

wiba-nordic (changes requested):

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -353,6 +387,32 @@ Build the target using the following command in the project directory to enable
...
+   Do not confuse the Bluetooth LE Security Manager Protocol (SMP) abbreviation with the Bluetooth LE Simple Management Protocol (SMP) used for the Device Firmware Upgrade purposes.
+
+There are following security levels defined by the Bluetooth LE specification:
+
+* Security Level 1 - supports communication without security.
+* Security Level 2 - supports AES-CMAC communication encryption, but does not require device authentication.
+* Security Level 3 - supports AES-CMAC communication encryption, requires device authentication and pairing.
+* Security Level 4 - supports ECDHE communication encryption, requires authentication and pairing.
+
+By default, the Matter bridge device has SMP enabled and supports security levels 2, 3 and 4.
```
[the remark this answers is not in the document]
kkasperczyk-no: Yes, it should. Applied.

wiba-nordic (approved):

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -142,6 +176,98 @@ void BLEConnectivityManager::DisconnectionHandler(bt_conn *conn, uint8_t reason)
...
+
+void BLEConnectivityManager::AuthenticationCancel(struct bt_conn *conn)
+{
+	char addr[BT_ADDR_LE_STR_LEN];
+
+	bt_addr_le_to_str(bt_conn_get_dst(conn), addr, sizeof(addr));
+
+	LOG_ERR("Pairing cancelled: %s\n", addr);
+
+	bt_addr_le_t a = *bt_conn_get_dst(conn);
```
markaj-nordic, line 197: Can we select more descriptive variable name? Or just do:
`Instance().RemoveBLEProvider(*bt_conn_get_dst(conn));`

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -142,6 +176,98 @@ void BLEConnectivityManager::DisconnectionHandler(bt_conn *conn, uint8_t reason)
...
+
+		bt_addr_le_t addr = Instance().mConnectedProviders[i]->GetBtAddress();
+		if (memcmp(&addr, bt_conn_get_dst(conn), sizeof(addr)) == 0) {
+			provider = Instance().mConnectedProviders[i];
+			break;
+		}
+	}
+
+	/* The device with given address was not found.  */
+	if (!provider) {
```
markaj-nordic, line 233: Should we log error then?

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -182,6 +186,23 @@ class BLEConnectivityManager {
...
+
+#ifdef CONFIG_BT_SMP
+	static void SecurityChangedHandler(struct bt_conn *conn, bt_security_t level, enum bt_security_err err);
+	static void AuthenticationCancel(struct bt_conn *conn);
+	static void PasskeyEntry(struct bt_conn *conn);
+	static void PairingComplete(struct bt_conn *conn, bool bonded);
+	static void PairingFailed(struct bt_conn *conn, enum bt_security_err reason);
+
+	/**
+	 * @brief Set authentication pincode to confirm the specific Bluetooth LE connection.
```
markaj-nordic, line 199: Why only this one is documented? If the rest of the methods guarded with CONFIG_BT_SMP are used only internally we should move them to the private scope of the class.

applications/matter_bridge/Kconfig
```diff
@@ -98,6 +98,30 @@ config BT_GATT_CLIENT
...
 
+config BT_SMP
+	default y
+
+if BT_SMP
+
+config BT_SETTINGS
+	default y
+
+# Configure how many Bluetooth LE devices will be paired with Matter bridge. Assuming it will be only bridged devices, it should be set to BT_MAX_CONN - 1.
```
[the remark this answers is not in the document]
kkasperczyk-no: It was my plan as well, but I think unfortunately it can't. It is because we redefine the default value of `BT_MAX_CONN` above and then we would try to use it for calculation, but it is done before sourcing Zephyr root Kconfig. So at this stage the `BT_MAX_CONN` config has no type defined and cannot be used for calculation as non-integer value.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: Right. Actually I started to think what should be done if it fails, because the GATT discovery will not be started and the device adding process will not be completed. On the other hand, on other failures we usually just call `RemoveBLEProvider`, but here it has no point, because provider does not exist. The question is, do we consider even realistic scenario, as provider should always be valid at this point.

samples/matter/common/src/bridge/ble_connectivity_manager.h
kkasperczyk-no: Yes, these are only static handlers used internally and cannot be called by the user directly. Sure, I added it to public, similarly to other currently existing handlers, but I agree all of them could be moved to the private scope. I will do that.

applications/matter_bridge/child_image/hci_rpmsg/prj.conf
```diff
@@ -17,3 +17,4 @@ CONFIG_RESET_ON_FATAL_ERROR=n
 
 CONFIG_BT_MAX_CONN=10
 CONFIG_BT_CENTRAL=y
+CONFIG_BT_EXT_ADV=y
```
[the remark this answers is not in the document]
kkasperczyk-no: Good point, I can see the `prj.conf` for a sysbuild hci_rpmsg image is already off for the bridge application. I pushed an update for it.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -353,6 +387,32 @@ Build the target using the following command in the project directory to enable
...
+   Do not confuse the Bluetooth LE Security Manager Protocol (SMP) abbreviation with the Bluetooth LE Simple Management Protocol that has the same abbreviation and is used for the Device Firmware Upgrade process.
+
+The following security levels are defined by the Bluetooth LE specification:
+
+* Security Level 1 - supports communication without security.
+* Security Level 2 - supports AES-CMAC communication encryption, but does not require device authentication.
+* Security Level 3 - supports AES-CMAC communication encryption, requires device authentication and pairing.
+* Security Level 4 - supports ECDHE communication encryption, requires authentication and pairing.
+
+By default, the Matter bridge application has SMP enabled and supports security levels 2, 3 and 4.
```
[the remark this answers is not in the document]
kkasperczyk-no: Sure, I was not aware that it exists. All in all I have impression that our documentation is less technical and easier to understand, that the one from Zephyr, but I can also add a link as a way to read more about this topic.

applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -39,7 +59,12 @@ static int AddBridgedDeviceHandler(const struct shell *shell, size_t argc, char
 		shell_fprintf(shell, SHELL_ERROR, "Invalid Bluetooth LE device index.\n");
 	}
 
+#ifdef CONFIG_BT_SMP
+	result = BleBridgedDeviceFactory::CreateDevice(uuid, address, nodeLabel, BluetoothConnectionSecurityRequest,
+						       const_cast<struct shell *>(shell));
```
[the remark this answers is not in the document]
markaj-nordic: Sorry, I guess we still need the full type name when casting.

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.h
```diff
@@ -76,10 +76,12 @@ CHIP_ERROR CreateDevice(int deviceType, bt_addr_le_t btAddress, const char *node
  * @param uuid the Bluetooth LE service UUID of a bridged device provider that will be paired with bridged device
  * @param btAddress the Bluetooth LE address of a device to be bridged with created Matter device
  * @param nodeLabel node label of a Matter device to be created
+ * @param callback callback that will be called if connection requires additional security information to be inserted
```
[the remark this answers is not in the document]
kkasperczyk-no: Done, please re-review.

samples/matter/common/src/bridge/ble_connectivity_manager.h
kkasperczyk-no: As discussed F2F, it turned out that they have to be public. Added comment to the methods declaration in the class.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
markaj-nordic: So it's up to you :)

markaj-nordic (approved):

kkasperczyk-no:
Adding DNM for now to allow QA guys make tests alignment before breaking the nightly runs.

ArekBalysNordic (approved):

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -353,6 +387,33 @@ Build the target using the following command in the project directory to enable
...
+   Do not confuse the Bluetooth LE Security Manager Protocol (SMP) abbreviation with the Bluetooth LE Simple Management Protocol that has the same abbreviation and is used for the Device Firmware Upgrade process.
+
+The following security levels are defined by the Bluetooth LE specification:
+
+* Security Level 1 - supports communication without security.
+* Security Level 2 - supports AES-CMAC communication encryption, but does not require device authentication.
+* Security Level 3 - supports AES-CMAC communication encryption, requires device authentication and pairing.
+* Security Level 4 - supports ECDHE communication encryption, requires authentication and pairing.
+
+To read more about the Bluetooth LE security implementation in Zephyr, see ``Security`` subsection of the :ref:`bluetooth-arch` page.
```
wiba-nordic, line 405:
```suggestion
To read more about the Bluetooth LE security implementation in Zephyr, see the Security section of the :ref:`bluetooth-arch` page.
```

wiba-nordic (changes requested):

wiba-nordic (approved):

PR #13534, opened 2024-01-04
net: lib: nrf_cloud: fix auto info section send

jayteemo:
Fix the behavior of the automatic shadow info section sending.
Only send section data if the config is enabled.
Do not clear section data if the config is disabled. 
IRIS-6851

### DIFF
--- a/subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap_transport.c
+++ b/subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap_transport.c
@@ -25,6 +25,7 @@
 #include <nrf_modem_at.h>
 #include <date_time.h>
 #include <net/nrf_cloud.h>
+#include <net/nrf_cloud_codec.h>
 #include <net/nrf_cloud_coap.h>
 #include <cJSON.h>
 #include <version.h>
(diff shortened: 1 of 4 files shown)


jayteemo:
I still need to test with CoAP.

jayteemo:
CoAP tested ✔️ 
Ready for review 👓

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap_transport.c
```diff
@@ -181,28 +182,43 @@ static int update_configured_info_sections(const char * const app_ver)
...
-		err = 0;
-	} else {
-		err = nrf_cloud_coap_shadow_device_status_update(&dev_status);
+	if (err) {
+		LOG_ERR("Failed to initialize object: %d", err);
+		return err;
+	}
+
+	/* Encode the enabled info sections */
+	err = nrf_cloud_enabled_info_sections_json_encode(info_obj.json, app_ver);
```
glarsennordic, line 195: Would it make more sense for us to just pass `info_obj`, and name the function `nrf_cloud_enabled_info_sections_encode` and return -ENOTSUP for anything other than JSON? I know we've used that pattern in the past

glarsennordic (approved):

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap_transport.c
jayteemo: if this wasn't in the internal codec, yes that would be preferred.
this function is also used by the MQTT code, so it was simpler to keep it as a cJSON *.

plskeggs (approved):

PR #13542, opened 2024-01-04
Bluetooth: Controller: New default event length and spacing for ISO

rugeGerritsen:
When ISO is enabled, we want to set asside as much time as possible for ISO activities.

By setting BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT=0, we ensure that the ACL will utilize as little time as possible and avoid increasing the data length capabilities.

The BT_CTLR_SDC_CENTRAL_ACL_EVENT_SPACING_DEFAULT needs to be updated accordingly - we want the ACLs to be spaced apart to avoid scheduling conflicts. By selecting the value of 30 ms, we ensure that:
 - The ACLs are non-overlapping given that they are using an interval which is greater than 30 ms.
 - The ACL events are not overlapping with ISO events, given that the ISO interval is a multiple of 7.5 ms or 10 ms.

### DIFF
--- a/subsys/bluetooth/controller/Kconfig
+++ b/subsys/bluetooth/controller/Kconfig
@@ -136,7 +136,7 @@ config BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT_OVERRIDE
 
 config BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT
 	int "Default max connection event length [us]"
-	default 10000 if BT_ISO && !BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT_OVERRIDE
+	default 0 if BT_ISO && !BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT_OVERRIDE
 	default 7500 if !BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT_OVERRIDE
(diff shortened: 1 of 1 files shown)


Tschet1 (approved):

rugeGerritsen:
> Did you choose 30ms because of BT_GAP_INIT_CONN_INT_MIN / the default connection interval parameter in zephyr?
> 
> I was thinking if we should create a warning when a shorter value is chosen for the ACL interval or if we should otherwise document the scheduling changes when using ISO.

I choose 30 as this the least common multiple of the commonly used ISO intervals of 7.5 ms and 10 ms. 
We should ensure that we establish the connections with some interval above 30 ms to make it to work. The other approach is to set this value to 10 ms and rather make the application handle this in case it needs an ISO interval of 7.5 ms.

Tschet1:
> We should ensure that we establish the connections with some interval above 30 ms to make it to work. The other approach is to set this value to 10 ms and rather make the application handle this in case it needs an ISO interval of 7.5 ms.

I see that a lot of the BAP configurations use 7.5ms. So maybe 30ms is a sane value.
I think we could in the commit message mention why we choose 30ms and then update documentation for the SoftDevice controller in a follow-up.

Tschet1 (approved):

yuxCai (approved):

PR #13550, opened 2024-01-05
test: aws_iot: Fix segfault in unit test

balaji-nordic:
- net: lib: aws_iot: Add null pointer check in strnstr
- test: aws_iot: Fix segfault in unit test

### DIFF
--- a/scripts/quarantine_no_optimization.yaml
+++ b/scripts/quarantine_no_optimization.yaml
@@ -2,7 +2,7 @@
 # will be skipped if quarantine is used. More details here:
 # https://docs.zephyrproject.org/latest/guides/test/twister.html#quarantine
 # To have an empty list use:
+# - scenarios:
+#     - None
 - scenarios:
-    - net.lib.aws_iot
-  platforms:
-    - native_posix
+    - None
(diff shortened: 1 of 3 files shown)


simensrostad (approved):

scripts/quarantine_no_optimization.yaml
```diff
@@ -1,8 +0,0 @@
-# The configurations resulting as a product of scenarios and platforms
-# will be skipped if quarantine is used. More details here:
-# https://docs.zephyrproject.org/latest/guides/test/twister.html#quarantine
-# To have an empty list use:
-- scenarios:
-    - net.lib.aws_iot
-  platforms:
-    - native_posix
```
PerMac, line 8: Please, don't remove this file, as it is now required by the plan. Clear it out instead by replacing this entry with:
```
- scenarios:
    - None
```

PerMac (changes requested):

scripts/quarantine_no_optimization.yaml
balaji-nordic: Thanks. I was unsure what to do. I will fix as you suggest.

balaji-nordic: @PerMac  Fixed. Also updated the comment in the header of the quarantine file. Please have a look.

PerMac (approved):

bjda (approved):

PR #13553, opened 2024-01-07
bluetooth: hids: Don't leak conn_ctx on error paths in writes

ndreys:
A number of erros paths in various _write functions do early returns after acquiring conn_ctx without releasing the context first. Fix this by converting affected early returns in to error cleanup gotos and sharing the "happy path" cleanup sequence.

### DIFF
--- a/subsys/bluetooth/services/hids.c
+++ b/subsys/bluetooth/services/hids.c
@@ -121,6 +121,7 @@ static ssize_t hids_protocol_mode_write(struct bt_conn *conn,
 					void const *buf, uint16_t len,
 					uint16_t offset, uint8_t flags)
 {
+	ssize_t ret = len;
 	LOG_DBG("Writing to Protocol Mode characteristic.");
 
 	struct bt_hids_pm_data *pm = attr->user_data;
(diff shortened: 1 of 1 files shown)


subsys/bluetooth/services/hids.c
```diff
@@ -321,7 +326,8 @@ static ssize_t hids_outp_rep_write(struct bt_conn *conn,
 	rep_data = conn_data->outp_rep_ctx + rep->offset;
 
 	if (offset + len > rep->size) {
-		return BT_GATT_ERR(BT_ATT_ERR_INVALID_OFFSET);
+		len = BT_GATT_ERR(BT_ATT_ERR_INVALID_OFFSET);
```
jori-nordic, line 329: `ret =`

jori-nordic (changes requested):

subsys/bluetooth/services/hids.c
ndreys: Updated. Thanks for catching that!

jori-nordic (approved):
thanks for the patch!

alwa-nordic (approved):

KAGA164 (approved):
Good catch. Thanks

PR #13560, opened 2024-01-08
samples: cellular: modem_trace_flash: doc: fix

MirkoCovizzi:
Remove `|connect_terminal|` entry to prevent having two serial terminals connected to the device at the same time when choosing the nRF Connect for Desktop Serial Terminal app. Having two serial terminals connected to the device will cause a garbled output.

### DIFF
--- a/samples/cellular/modem_trace_flash/README.rst
+++ b/samples/cellular/modem_trace_flash/README.rst
@@ -83,7 +83,6 @@ Testing
 After programming the sample and board controller firmware (as mentioned in Requirements section) to your development kit, test it by performing the following steps:
 
 #. |connect_kit|
-#. |connect_terminal|
 #. Open the `Cellular Monitor`_ desktop application and connect the DK.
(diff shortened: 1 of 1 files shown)


samples/cellular/modem_trace_flash/README.rst
```diff
@@ -83,7 +83,6 @@ Testing
 After programming the sample and board controller firmware (as mentioned in Requirements section) to your development kit, test it by performing the following steps:
 
 #. |connect_kit|
-#. |connect_terminal|
```
eivindj-nordic, line 86: Then we should open the nRF Connect Serial terminal instead to observe the output. This can be opened separately or through the Cellular Monitor.

MirkoCovizzi: That step is mentioned in the following steps. If it's a hard requirement I could remove the `(optional)` string.

eivindj-nordic (approved):

divipillai (approved):

PR #13569, opened 2024-01-09
net: fota_download: Allow passing sec tag list

jtguggedal:
net: lib: fota_download: Use sec tag list instead of single entry

Changes to use a list of sec tags instead of limiting tag count
to one. This lets a user set up several sec tags, which is already
supported by the download_client used by fota_download.

--- 

net: fota_download: Add event for failed connection attempt

The errnos -ECONNREFUSED and -ECONNABORTED are most likely to
happen during connection establishment, and in particular because
of issues with the TLS handshake.

The patch attempts to distinguish between general errors from the
download_client and errors arrising from the connection establishment.


jtguggedal:
DNM because the external Memfault SDK version must be updated in the same commit as the signature of `fota_download_start()` changes.

subsys/net/lib/fota_download/src/fota_download.c
```diff
@@ -420,17 +425,18 @@ int fota_download_s0_active_get(bool *const s0_active)
...
 int fota_download_start_with_image_type(const char *host, const char *file,
-	int sec_tag, uint8_t pdn_id, size_t fragment_size,
+	const int *sec_tag_list, uint8_t sec_tag_count, uint8_t pdn_id, size_t fragment_size,
 	const enum dfu_target_image_type expected_type)
 {
-	static int sec_tag_list[1];
 	uint32_t host_hash = 0;
 	uint32_t file_hash = 0;
 	int err = -1;
+	static int sec_tag_list_copy[CONFIG_FOTA_DOWNLOAD_SEC_TAG_LIST_SIZE];
```
jtguggedal, line 434: I wasn't sure where to put the sec tag list storage, it might as well live in the `download_client_cfg` struct.

subsys/net/lib/fota_download/Kconfig
```diff
@@ -56,6 +56,12 @@ config FOTA_DOWNLOAD_RESOURCE_LOCATOR_LENGTH
 	int "Size of buffer used for Resource locator"
 	default 512
 
+config FOTA_DOWNLOAD_SEC_TAG_LIST_SIZE
```
simensrostad, line 59: Maybe slap on a '_MAX' to indicate that this is not the actual size but rather the maximum allowed.

simensrostad (approved):

SeppoTakalo:
I don't like this, it changes the API without announcing the deprecation beforehand.

I would propose, instead, that you create a new API with these new parameters and just deprecate the old ones, and make them to call the new API.

New API could be:
* `fota_download()` with all the parameter that you proposed for `fota_download_start_with_image_type()`.
* `fota_download_any()` with just alias of calling `fota_download(...,IMAGE_TYPE_ANY)`

Then current API could be deprecated and just forward `fota_download_start_with_image_type()` to call the `fota_download()`

subsys/net/lib/fota_download/src/fota_download.c
```diff
@@ -280,6 +280,11 @@ static int download_client_callback(const struct download_client_evt *event)
 			/* Fall through and return 0 below to tell
 			 * download_client to retry
 			 */
+		} else if ((event->error == -ECONNABORTED) || (event->error == -ECONNREFUSED)) {
+			LOG_ERR("Download client failed to connect to server");
+			set_error_state(FOTA_DOWNLOAD_ERROR_CAUSE_CONNECT_FAILED);
+
+			goto error_and_close;
```
SeppoTakalo, line 287: Can this be combined with the `else` case below?
Or why do we not need to call `dfu_target_done()` here, but we do below.

jtguggedal: `dfu_target_done(false)` is called also under the `error_and_close` label, so should not need to call it here too. I can instead remove it below in the `else`

SeppoTakalo: Oh, I see.
Then only difference is the verbosity, which I don't think is relevant.. I'm fine with how it is.

jtguggedal:
> I don't like this, it changes the API without announcing the deprecation beforehand.
> 
> I would propose, instead, that you create a new API with these new parameters and just deprecate the old ones, and make them to call the new API.
> 
> New API could be:
> 
> * `fota_download()` with all the parameter that you proposed for `fota_download_start_with_image_type()`.
> * `fota_download_any()` with just alias of calling `fota_download(...,IMAGE_TYPE_ANY)`
> 
> Then current API could be deprecated and just forward `fota_download_start_with_image_type()` to call the `fota_download()`

I'm open to do something along those lines. Don't know how this is usually handled. For simplicity I'd like to keep the same function names, as naming is one of the hardest things in programming, but I see your point.

@rlubos How do you usually do this in upstream networking libraries?

rlubos:
> @rlubos How do you usually do this in upstream networking libraries?

This depends on the API maturity level, the whole process is described [here](https://docs.zephyrproject.org/latest/develop/api/api_lifecycle.html), although I'm not sure if we follow the same process in NCS.

Unstable/experimental APIs do not require deprecation, although it's a good practice to mention such API changes in the [migration guide](https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/releases/migration-guide-3.6.rst#networking) (I don't see a similar page for NCS though). Stable APIs require deprecation, so a function rename is likely the only option (which I'm not a big fan of).

subsys/net/lib/fota_download/src/fota_download.c
```diff
@@ -420,17 +425,18 @@ int fota_download_s0_active_get(bool *const s0_active)
...
-	static int sec_tag_list[1];
 	uint32_t host_hash = 0;
 	uint32_t file_hash = 0;
 	int err = -1;
+	static int sec_tag_list_copy[CONFIG_FOTA_DOWNLOAD_SEC_TAG_LIST_SIZE];
 
 	struct download_client_cfg config = {
 		.pdn_id = pdn_id,
 		.frag_size_override = fragment_size,
+		.sec_tag_list = sec_tag_list_copy,
```
jayteemo, line 439: you could remove this line and set it down below with `config.sec_tag_count = sec_tag_count;`,  just so its all done in one place 🤷

jayteemo (approved):

plskeggs:
> > I don't like this, it changes the API without announcing the deprecation beforehand.
> > I would propose, instead, that you create a new API with these new parameters and just deprecate the old ones, and make them to call the new API.
> > New API could be:
> > 
> > * `fota_download()` with all the parameter that you proposed for `fota_download_start_with_image_type()`.
> > * `fota_download_any()` with just alias of calling `fota_download(...,IMAGE_TYPE_ANY)`
> > 
> > Then current API could be deprecated and just forward `fota_download_start_with_image_type()` to call the `fota_download()`
> 
> I'm open to do something along those lines. Don't know how this is usually handled. For simplicity I'd like to keep the same function names, as naming is one of the hardest things in programming, but I see your point. I'll do what the consensus here prefers.

In general we change things too much and making it hard for users to upgrade to the next version.  I agree with Seppo that this could be better handled with a new function name, and keep the old name but deprecate it (though honestly, if the old function simply calls the new, why bother with deprecation?).

jtguggedal:
I have changed the PR to implement @SeppoTakalo's suggestions about adding new functions and let the old ones call the new functions:
* `fota_download_start_with_image_type()` calls the new function `fota_download(...)`
* `fota_download_start()` calls the new function `fota_download_any(...)`

And I agree with @plskeggs that if we add new functions, why bother with deprecation. Deprecated functions will be removed in a couple of releases, and NCS users would then have the same trouble of upgrading NCS version. If they can work with the current APIs that allow only one sec tag, there's no need to force them over to using a sec tag list.
I'm not a big fan of the slightly confusing naming we now get of the 4 functions that do almost the same thing, but I couldn't come up with anything better. Let me know if you have suggestions.

subsys/net/lib/fota_download/Kconfig
```diff
@@ -56,6 +56,12 @@ config FOTA_DOWNLOAD_RESOURCE_LOCATOR_LENGTH
 	int "Size of buffer used for Resource locator"
 	default 512
 
+config FOTA_DOWNLOAD_SEC_TAG_LIST_SIZE_MAX
```
stig-bjorlykke, line 59: We already have `NRF_SOCKET_TLS_MAX_SEC_TAG_LIST_SIZE` which is the maximum size supported by modem.

stig-bjorlykke (commented):
Be aware that if one of the sec_tag's in the list has an invalid certificate then the `connect()` will fail with `ECONNREFUSED` without trying to connect.

subsys/net/lib/fota_download/Kconfig
jtguggedal: Yeah, maybe I should set an allowed range when TLS in modem is used. Not sure how to do that, though

samples/cellular/http_update/modem_delta_update/src/main.c
```diff
@@ -307,9 +309,9 @@ static int update_download(void)
 	}
 
 	/* Functions for getting the host and file */
-	err = fota_download_start(CONFIG_DOWNLOAD_HOST, file, SEC_TAG, 0, 0);
+	err = fota_download_any(CONFIG_DOWNLOAD_HOST, file, &sec_tag, sec_tag_count, 0, 0);
```
SeppoTakalo, line 312: Actually.. As we are touching these, why don't we change it to use `DFU_TARGET_IMAGE_TYPE_MODEM_DELTA`

After all, this is DELTA update demo.

So it is kind of wrong to accept any kind of image.

SeppoTakalo (approved):
I'm OK with the change.

And to be honest, maybe we could leave old APIs not deprecated. Does not matter much.

However, I suggest that either on this, or some follow up PR, we should change the FOTA samples to use proper image types, one of:
* DFU_TARGET_IMAGE_TYPE_ANY_APPLICATION
* DFU_TARGET_IMAGE_TYPE_ANY_MODEM
* DFU_TARGET_IMAGE_TYPE_MODEM_DELTA
* DFU_TARGET_IMAGE_TYPE_FULL_MODEM

As it does not really make sense that delta-update would silently accept application image.

include/net/fota_download.h
```diff
@@ -103,6 +107,8 @@ struct fota_download_evt {
  */
 typedef void (*fota_download_callback_t)(const struct fota_download_evt *evt);
 
+
+
```
balaji-nordic, line 111: Remove?

applications/serial_lte_modem/src/slm_at_fota.c
```diff
@@ -221,9 +221,10 @@ static int do_fota_start(int op, const char *file_uri, int sec_tag,
 			LOG_ERR("Missing sec_tag");
 			return -EINVAL;
 		}
-		ret = fota_download_start_with_image_type(hostname, path, sec_tag, pdn_id, 0, type);
+		ret = fota_download(hostname, path, &sec_tag, 1, pdn_id, 0,
```
balaji-nordic, line 224: Can be written in one line right?

include/net/fota_download.h
```diff
@@ -112,6 +118,36 @@ typedef void (*fota_download_callback_t)(const struct fota_download_evt *evt);
...
+ * @param host Name of host to start downloading from. Can include scheme
+ *             and port number, for example https://google.com:443
+ * @param file Path to the file you wish to download. See fota_download_any()
+ *             for details on expected format.
+ * @param sec_tag_list Security tags that you want to use with HTTPS. Pass NULL to disable TLS.
+ * @param sec_tag_count Number of TLS security tags in list. If set to 0, TLS is disabled.
+ * @param pdn_id Packet Data Network ID to use for the download, or 0 to use the default.
+ * @param fragment_size Fragment size to be used for the download.
+ *			If 0, @kconfig{CONFIG_DOWNLOAD_CLIENT_HTTP_FRAG_SIZE} is used.
+ *
```
MarkusLassila, line 139:
```suggestion
 *  @param expected_type Type of firmware file to be downloaded and installed.
 *
```

include/net/fota_download.h
```diff
@@ -103,6 +107,8 @@ struct fota_download_evt {
  */
 typedef void (*fota_download_callback_t)(const struct fota_download_evt *evt);
 
+
```
MarkusLassila, line 110:
```suggestion
```

include/net/fota_download.h
```diff
@@ -112,6 +118,36 @@ typedef void (*fota_download_callback_t)(const struct fota_download_evt *evt);
  */
 int fota_download_init(fota_download_callback_t client_callback);
 
+/**@brief Start downloading the given file from the given host. Validate that the
```
balaji-nordic, line 121: Try to keep the brief short with just one sentence. You can bring the second sentence to the description.

include/net/fota_download.h
```diff
@@ -112,6 +118,36 @@ typedef void (*fota_download_callback_t)(const struct fota_download_evt *evt);
...
+ * file type matches the expected type before starting the installation.
```
balaji-nordic, line 122:
```suggestion
 * file type matches the expected type before starting the download.
```

MarkusLassila (approved):

include/net/fota_download.h
```diff
@@ -112,6 +118,36 @@ typedef void (*fota_download_callback_t)(const struct fota_download_evt *evt);
...
+ *
+ * URI parameters (host and file) are not copied, so pointers must stay valid
+ * until download is finished.
+ *
+ * @param host Name of host to start downloading from. Can include scheme
+ *             and port number, for example https://google.com:443
+ * @param file Path to the file you wish to download. See fota_download_any()
+ *             for details on expected format.
+ * @param sec_tag_list Security tags that you want to use with HTTPS. Pass NULL to disable TLS.
+ * @param sec_tag_count Number of TLS security tags in list. If set to 0, TLS is disabled.
```
balaji-nordic, line 135:
```suggestion
 * @param sec_tag_count Number of TLS security tags in list.  Pass 0 to disable TLS. 
```
To stay consistent with previous line.

include/net/fota_download.h
```diff
@@ -112,6 +118,36 @@ typedef void (*fota_download_callback_t)(const struct fota_download_evt *evt);
...
+ * @param pdn_id Packet Data Network ID to use for the download, or 0 to use the default.
+ * @param fragment_size Fragment size to be used for the download.
+ *			If 0, @kconfig{CONFIG_DOWNLOAD_CLIENT_HTTP_FRAG_SIZE} is used.
```
balaji-nordic, line 138: Add

`* @param expected_type Type of firmware file to be downloaded and installed`

include/net/fota_download.h
```diff
@@ -130,15 +166,39 @@ int fota_download_init(fota_download_callback_t client_callback);
  *              See <a href="https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/ug_bootloader.html">
  *              Secure Bootloader Chain Docs</a> for details regarding the upgradable
  *              bootloader slots.
- * @param sec_tag Security tag you want to use with HTTPS set to -1 to Disable.
+ * @param sec_tag_list Security tags that you want to use with HTTPS. Pass NULL to disable TLS.
+ * @param sec_tag_count Number of TLS security tags in list. If set to 0, TLS is disabled.
```
balaji-nordic, line 170:
```suggestion
 * @param sec_tag_count Number of TLS security tags in list. Pass 0 to disable TLS.
```

include/net/fota_download.h
```diff
@@ -130,15 +166,39 @@ int fota_download_init(fota_download_callback_t client_callback);
...
  * @param pdn_id Packet Data Network ID to use for the download, or 0 to use the default.
  * @param fragment_size Fragment size to be used for the download.
  *			If 0, @kconfig{CONFIG_DOWNLOAD_CLIENT_HTTP_FRAG_SIZE} is used.
+ * @param expected_type Type of firmware file to be downloaded and installed.
```
balaji-nordic, line 174: Remove

include/net/fota_download.h
```diff
@@ -130,15 +166,39 @@ int fota_download_init(fota_download_callback_t client_callback);
...
  *                   Otherwise, a negative value is returned.
  */
+int fota_download_any(const char *host, const char *file, const int *sec_tag_list,
+		      uint8_t sec_tag_count, uint8_t pdn_id, size_t fragment_size);
+
+/**@brief Start downloading the given file from the given host.
+ *
+ * When the download is complete, the secondary slot of MCUboot is tagged as having
+ * valid firmware inside it. The completion is reported through an event.
+ *
```
balaji-nordic, line 189: Add 

```
 * URI parameters (host and file) are not copied, so pointers must stay valid
 * until download is finished.
```

include/net/fota_download.h
```diff
@@ -163,11 +223,10 @@ int fota_download_start(const char *host, const char *file, int sec_tag,
  *
  * @retval 0	     If download has started successfully.
  * @retval -EALREADY If download is already ongoing.
- *                   Otherwise, a negative value is returned.
```
balaji-nordic, line 166: Revert this (as other negative values are still returned by this function)

include/net/fota_download.h
```diff
@@ -163,11 +223,10 @@ int fota_download_start(const char *host, const char *file, int sec_tag,
...
  */
-int fota_download_start_with_image_type(const char *host, const char *file,
-			int sec_tag, uint8_t pdn_id, size_t fragment_size,
-			const enum dfu_target_image_type expected_type);
+int fota_download_start_with_image_type(const char *host, const char *file, int sec_tag,
+					uint8_t pdn_id, size_t fragment_size,
+					const enum dfu_target_image_type expected_type);
```
balaji-nordic, line 229: Revert this too as these are just white space changes?

samples/cellular/http_update/modem_full_update/src/main.c
```diff
@@ -456,9 +458,9 @@ static int update_download(void)
 	}
 
 	/* Functions for getting the host and file */
-	err = fota_download_start(CONFIG_DOWNLOAD_HOST, file, SEC_TAG, 0, 0);
+	err = fota_download_any(CONFIG_DOWNLOAD_HOST, file, &sec_tag, sec_tag_count, 0, 0);
```
balaji-nordic, line 461: Use `fota_download()` instead with  `DFU_TARGET_IMAGE_TYPE_FULL_MODEM` here?

subsys/net/lib/fota_download/Kconfig
balaji-nordic:
```
default NRF_SOCKET_TLS_MAX_SEC_TAG_LIST_SIZE  if NRF_MODEM_LIB
```

subsys/net/lib/fota_download/src/util/fota_download_util.c
```diff
@@ -195,12 +195,12 @@ static void start_fota_download(struct k_work *work)
 {
 	int ret;
 
-	ret = fota_download_start_with_image_type(fota_host, fota_path, fota_sec_tag, 0, 0,
+	ret = fota_download(fota_host, fota_path, NULL, 0, 0, 0,
 						  active_dfu_type);
```
balaji-nordic, line 199: Can be in one line

subsys/net/lib/fota_download/src/util/fota_download_util.c
```diff
@@ -195,12 +195,12 @@ static void start_fota_download(struct k_work *work)
...
 	if (ret) {
 		struct fota_download_evt evt;
 
-		LOG_ERR("fota_download_start() failed, return code %d", ret);
+		LOG_ERR("fota_download_any() failed, return code %d", ret);
```
balaji-nordic, line 203:
```suggestion
		LOG_ERR("fota_download() failed, return code %d", ret);
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -693,6 +693,13 @@ Libraries for networking
 
   * Added support for using a password when connecting to a broker.
 
+* :ref:`lib_fota_download` library:
+
+  * Added:
+
+    * The functions :c:func:`fota_download` and :c:func:`fota_download_any` that can accept a security tag list and security tag count as arguments instead of a single security tag.
+    * :c:enumerator:`FOTA_DOWNLOAD_ERROR_CAUSE_CONNECT_FAILED` as potential error casue in  :c:enumerator:`FOTA_DOWNLOAD_EVT_ERROR` events.
```
balaji-nordic, line 701:
```suggestion
    * Added :c:enumerator:`FOTA_DOWNLOAD_ERROR_CAUSE_CONNECT_FAILED` as potential error cause in  :c:enumerator:`FOTA_DOWNLOAD_EVT_ERROR` events.
```

tests/subsys/net/lib/fota_download/src/test_fota_download.c
```diff
@@ -20,7 +20,7 @@
 static char buf[1024];
```
balaji-nordic, line 20: How about some test cases for `fota_download_start` and the `fota_download`?

balaji-nordic (changes requested):

jayteemo:
> I'm not a big fan of the slightly confusing naming we now get of the 4 functions that do almost the same thing

yeah, while breaking the API is not ideal, i find these new functions more confusing.
i prefer the original implementation.  it will break projects that are upgraded, but it is a pretty obvious and easy fix.

suggestion if you want to not-break the api, just add _stl (for sec_tag list) to the existing function names 🤷 
`fota_download_start_stl`
`fota_download_start_stl_with_image_type`

still approving.

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_poll.c
```diff
@@ -333,9 +334,8 @@ static int start_download(void)
 	}
 
 	LOG_INF("Starting FOTA download of %s/%s", job.host, job.path);
-	int err = fota_download_start_with_image_type(job.host, job.path,
```
jayteemo, line 336: with your latest update, this all changes can be reverted from this file?
unless the plan is to deprecate `fota_download_start_with_image_type`?

subsys/net/lib/fota_download/Kconfig
jtguggedal: Yes, that one works if TLS stack in the modem is used. But if TLS is ran in the app core, `NRF_SOCKET_TLS_MAX_SEC_TAG_LIST_SIZE` is not relevant anymore. I think I'm going to leave the deafult at 5 as an okay trade-off between RAM usage and large-enough-for-most-users

include/net/fota_download.h
jtguggedal: Rewrote this a bit, please have a look

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota_poll.c
jtguggedal: Yeah, you're right, reverted as there is no plan to deprecate

jtguggedal:
@balaji-nordic and @divipillai: I addressed your comments, and it should be ready for review again.

I did not do anything with the tests as they are incomplete as is. Will do changes in a separate commit and probably in a spearate PR as this PR already has a few approvals

divipillai (approved):

plskeggs (approved):

PR #13575, opened 2024-01-10
drivers: wifi: Notify app of local TWT tear down

ajayparida:
UMAC is not sending any feedback for the local teardown CMD, to keep the application design consistent, driver should always inform with the teardown status, it would be always success till UMAC implements the feedback.

### DIFF
--- a/drivers/wifi/nrf700x/src/wifi_mgmt.c
+++ b/drivers/wifi/nrf700x/src/wifi_mgmt.c
@@ -397,6 +397,7 @@ int nrf_wifi_twt_teardown_flows(struct nrf_wifi_vif_ctx_zep *vif_ctx_zep,
 	struct nrf_wifi_umac_config_twt_info twt_info = {0};
 	enum nrf_wifi_status status = NRF_WIFI_STATUS_FAIL;
 	int ret = 0;
+	struct wifi_twt_params twt_params = {0};
 
 	if (!vif_ctx_zep) {
 		LOG_ERR("%s: vif_ctx_zep is NULL", __func__);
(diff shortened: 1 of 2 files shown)


drivers/wifi/nrf700x/src/wifi_mgmt.c
```diff
@@ -428,6 +429,10 @@ int nrf_wifi_twt_teardown_flows(struct nrf_wifi_vif_ctx_zep *vif_ctx_zep,
 		}
 		/* UMAC doesn't send TWT teardown event for host initiated teardown */
 		nrf_wifi_twt_update_internal_state(vif_ctx_zep, false, flow_id);
+		/* Raise a tear down event from host itself, till UMAC supports */
```
sachinthegreen, line 432: Don't we need to encode the status i.e. `SUCCESS` or `FAIL` ?

ajayparida: For host genearted, it will be always success. There is no command status field. Only for peer response of sucess/failure. 
Do I need to add for a command status field as this is temporary till UMAC gives proper event?

ajayparida: Yeah, I think it's better to keep success/failure. I will update accordingly

krish2718:
```suggestion
		/* TODO: Remove this once UMAC sends the status */
```

krish2718: Yes, we should always inform the status

ajayparida: Yeah, I will update.

ioannisg:
@ajayparida @sachinthegreen is this event-notification to be demonstrated in the TWT sample (or elsewhere)?

krish2718 (approved):

sachinthegreen (approved):

ajayparida:
> @ajayparida @sachinthegreen is this event-notification to be demonstrated in the TWT sample (or elsewhere)?



> @ajayparida @sachinthegreen is this event-notification to be demonstrated in the TWT sample (or elsewhere)?

Yes in the sample twt app user will be  notified of the teardown status(success/failure) with flow id. 
Also in the shell application, once user initiates teardowm, will be notified.

rado17 (approved):

rlubos:
@ajayparida We've been into lot of struggle due to broken BLE CI. I've updated manifest to the latest sdk-zephyr revision in https://github.com/nrfconnect/sdk-nrf/pull/13708 - please drop the `west.yml` change and we can focus on Wi-Fi specific part only, hopefully w/o fighting with CI anymore.

ajayparida:
> @ajayparida We've been into lot of struggle due to broken BLE CI. I've updated manifest to the latest sdk-zephyr revision in #13708 - please drop the `west.yml` change and we can focus on Wi-Fi specific part only, hopefully w/o fighting with CI anymore.

Dropped west.yml change.

PR #13577, opened 2024-01-10
manifest: sdk-zephyr: Pull fixes for Wi-Fi shell string args

krish2718:
Fixes case sensitivity in processing string args.

Fixes SHEL-2360.


sachinthegreen (approved):

krish2718:
Manifest will be updated as part of https://github.com/nrfconnect/sdk-nrf/pull/13633

PR #13580, opened 2024-01-11
samples: matter: Added Nordic namespace to all common modules

kkasperczyk-no:
The common modules delivered by Nordic should be clearly separated from the generic code, so the namespace Nordic was added to all of them.

By the way refactored directories structure a bit to group files of similar purpose in a one location.


samples/matter/common/src/app/task_executor.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
+
+constexpr size_t kTaskQueueSize = CONFIG_NCS_SAMPLE_MATTER_APP_TASK_QUEUE_SIZE;
+
+K_MSGQ_DEFINE(sTaskQueue, sizeof(Nordic::Matter::Task), kTaskQueueSize, alignof(Nordic::Matter::Task));
+
+namespace Nordic
```
Damian-Nordic, line 18: Note that starting from C++17, you could use nested namespaces:
```
namespace Nordic::Matter
{
```
It's not relevant, just fyi, that you can type fewer characters to achieve the same :).

samples/matter/common/src/board/board.cpp
```diff
@@ -18,14 +18,17 @@
...
+namespace Nordic {
+namespace Matter {
 
 bool Board::Init(button_handler_t buttonHandler, LedStateHandler ledStateHandler)
 {
 #ifdef CONFIG_DK_LIBRARY
 	/* Initialize LEDs */
-	LEDWidget::InitGpio();
-	LEDWidget::SetStateUpdateCallback(LEDStateUpdateHandler);
+	Nordic::LEDWidget::InitGpio();
```
Damian-Nordic, line 30: This definition is already in the Nordic namespace so you don't need such changes.

samples/matter/common/src/board/led_widget.cpp
```diff
@@ -11,7 +11,9 @@
 #endif
 #include <zephyr/kernel.h>
 
-static LEDWidget::LEDWidgetStateUpdateHandler sStateUpdateCallback;
+static Nordic::LEDWidget::LEDWidgetStateUpdateHandler sStateUpdateCallback;
```
Damian-Nordic, line 14: Consider moving such global definitions inside `Nordic` namespace to shorten the type.

Damian-Nordic (approved):
Looking good but IMHO the code in `matter/common` that is already in `Nordic` namespace doesn't need to specify this namespace explicitly when referencing other objects in the same namespace.
I also thought that "Nordic" namespace could be replaced with shorter "nrf", but don't have a strong opinion on that, so ignore this comment unless others would also like this idea.

samples/matter/common/src/board/board.cpp
kkasperczyk-no: Yes, you are right, the reason is I modified files one by one and fixed compilation errors gradually, so there are some unnecessary Nordic oraz Nordic::Matter usages in a modules that also use them. I will go through all files once again and remove such cases.

markaj-nordic: @kkasperczyk-no I hope you leaked "oraz" here on purpose because it sounds great to me :)

kkasperczyk-no: Wtf :smile: I have no idea how it happened. Maybe I was thinking about too many things at the same time.

samples/matter/common/src/board/board.cpp
```diff
@@ -18,6 +18,8 @@
 
 LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
 
+namespace Nordic::Matter {
```
markaj-nordic, line 21:
```suggestion
namespace Nordic {
```
I'd like to keep Matter namespace for more Matter related APIs, like init and event handling.

samples/matter/common/src/board/board.h
```diff
@@ -8,8 +8,11 @@
 
 #include "board/board_config.h"
 #include "board/board_consts.h"
-#include "led/led_util.h"
-#include "led/led_widget.h"
+#include "board/led_util.h"
+#include "board/led_widget.h"
+
+namespace Nordic::Matter
```
markaj-nordic, line 14:
```suggestion
namespace Nordic
```

samples/matter/common/src/board/board_consts.h
```diff
@@ -8,35 +8,38 @@
 
 #include <cstdint>
 
-namespace FactoryResetConsts
+namespace Nordic::Matter
```
markaj-nordic, line 11:
```suggestion
namespace Nordic
```

samples/matter/common/src/app/task_executor.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
+
+constexpr size_t kTaskQueueSize = CONFIG_NCS_SAMPLE_MATTER_APP_TASK_QUEUE_SIZE;
+
+K_MSGQ_DEFINE(sTaskQueue, sizeof(Nordic::Matter::Task), kTaskQueueSize, alignof(Nordic::Matter::Task));
+
+namespace Nordic::Matter
```
markaj-nordic, line 18:
```suggestion
namespace Nordic
```
Same as for Board, I think this is more application level utility.

samples/matter/common/src/app/task_executor.h
```diff
@@ -0,0 +1,42 @@
...
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#pragma once
+
+#include <functional>
+
+namespace Nordic::Matter
```
markaj-nordic, line 11: ditto

samples/matter/common/src/board/board.cpp
kkasperczyk-no: Ok, aligned in all places.

markaj-nordic:
> Looking good but IMHO the code in `matter/common` that is already in `Nordic` namespace doesn't need to specify this namespace explicitly when referencing other objects in the same namespace. I also thought that "Nordic" namespace could be replaced with shorter "nrf", but don't have a strong opinion on that, so ignore this comment unless others would also like this idea. :)

Sorry for the late feedback, but actually, nrf instead of Nordic sounds good to me. We could then use `nrf::matter` to keep the consistent lowercase approach. @kkasperczyk-no  what do you think?

kkasperczyk-no:
> > Looking good but IMHO the code in `matter/common` that is already in `Nordic` namespace doesn't need to specify this namespace explicitly when referencing other objects in the same namespace. I also thought that "Nordic" namespace could be replaced with shorter "nrf", but don't have a strong opinion on that, so ignore this comment unless others would also like this idea. :)
> 
> Sorry for the late feedback, but actually, nrf instead of Nordic sounds good to me. We could then use `nrf::matter` to keep the consistent lowercase approach. @kkasperczyk-no what do you think?

It's ok to me, I just followed the existing approach.

kkasperczyk-no:
> > > Looking good but IMHO the code in `matter/common` that is already in `Nordic` namespace doesn't need to specify this namespace explicitly when referencing other objects in the same namespace. I also thought that "Nordic" namespace could be replaced with shorter "nrf", but don't have a strong opinion on that, so ignore this comment unless others would also like this idea. :)
> > 
> > 
> > Sorry for the late feedback, but actually, nrf instead of Nordic sounds good to me. We could then use `nrf::matter` to keep the consistent lowercase approach. @kkasperczyk-no what do you think?
> 
> It's ok to me, I just followed the existing approach.

@markaj-nordic nevertheless I can see we have namespaces names written using UpperCamelCase style, so actually changing it now would rise another inconsistencies. Do we even have some common agreement regarding the used case style?

kkasperczyk-no:
I changed name from `Nordic` to `Nrf`.

Damian-Nordic (approved):

markaj-nordic (approved):

PR #13590, opened 2024-01-11
samples: matter: Unified default Matter event handler

markaj-nordic:
Implemented common Matter event handler module that includes the default handler and API to register user provided handlers in the application code.

Added default Matter event handler for GPIO related operations in the Board utility module, that can be registered in the application.


kkasperczyk-no (approved):
LGTM

Damian-Nordic (approved):

PR #13602, opened 2024-01-12
crypto: Enable key management driver for Brainpool

vili-nordic:


mswarowsky (approved):

subsys/bootloader/bl_crypto/CMakeLists.txt
```diff
@@ -4,37 +4,39 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
-zephyr_library()
-zephyr_library_sources(bl_crypto.c)
-zephyr_library_link_libraries(nrfxlib_crypto)
-zephyr_link_libraries(nrfxlib_crypto)
+if(CONFIG_IS_SECURE_BOOTLOADER OR CONFIG_ZTEST)
```
SebastianBoe, line 7: I see that we have

add_subdirectory_ifdef(CONFIG_IS_SECURE_BOOTLOADER bootloader)

in nrf/subsys/CMakeLists.txt, so CONFIG_IS_SECURE_BOOTLOADER is always true.

Unless this CMakeLists.txt is sourced from elsewhere? (Which would be bad practice).

Making this if statement always true.

Would you mind explaining this if-statement?

SebastianBoe (changes requested):
Discussed offline, I think there is a simpler way to solve NCSDK-25144.

Will do some prototyping and report back.

PR #13603, opened 2024-01-12
drivers: wifi: Fix QSPI initialiation

krish2718:
QSPI initialization reads QSPI slave registers, and if RPU isn't powered on this causes undefined behaviour (works on DK, but doesn't work on customer board with different QSPI).

Fix this by dividing the RPU configuration into init and enable, init doesn't use QSPI, and both QSPI init and RPU enable use QSPI.

Fixes SHEL-2375.


Rallare (approved):
Tested successfully! Thanks again for helping, @krish2718 !

sachinthegreen (approved):

PR #13608, opened 2024-01-15
applications: Fixed duplicated add issue for Matter BT bridge

kkasperczyk-no:
Added check that prevents from adding the same Bluetooth LE device to the bridge multiple times.


markaj-nordic (approved):
LGTM

ArekBalysNordic (approved):

PR #13617, opened 2024-01-15
Download client IPv6 to IPv4 fallback improvements

stig-bjorlykke:
Change the IPv6 to IPv4 fallback to also try to connect before doing the fallback. This will fix two known issues:
  1. A DNS request for an IPv6 address may be successful even if the device does not have an IPv6 address.
  2. The host is not reachable using IPv6 because of a network routing issue.

Change IPv6 support from compile time to runtime.
  * Default attempt IPv6 connection first and fallback to IPv4 if DNS lookup or connect fails.
  * Add configuration to struct download_client_cfg for the device to optimize the download sequence when it only support IPv4 or IPv6.
  * Remove CONFIG_DOWNLOAD_CLIENT_IPV6.

Change the "Connecting to" log from repeating the full URL to show the resolved IP address. This will also show IP version used.

### DIFF
--- a/subsys/net/lib/download_client/src/download_client.c
+++ b/subsys/net/lib/download_client/src/download_client.c
@@ -429,15 +429,15 @@ static int client_connect(struct download_client *dl)
 		type |= SOCK_NATIVE_TLS;
 	}
 
-	/* Attempt IPv6 connection if configured, fallback to IPv4 */
-	if (IS_ENABLED(CONFIG_DOWNLOAD_CLIENT_IPV6)) {
+	/* Attempt IPv6 connection if configured, fallback to IPv4 on error */
+	if ((dl->config.family == AF_UNSPEC) || (dl->config.family == AF_INET6)) {
 		err = host_lookup(dl->host, AF_INET6, dl->config.pdn_id, &dl->remote_addr);
 		if (!err) {
 			err = client_socket_connect(dl, type, port);
 		}
 	}
 
-	if (err || !IS_ENABLED(CONFIG_DOWNLOAD_CLIENT_IPV6)) {
+	if (((dl->config.family == AF_UNSPEC) && err) || (dl->config.family == AF_INET)) {
 		err = host_lookup(dl->host, AF_INET, dl->config.pdn_id, &dl->remote_addr);
 		if (!err) {
 			err = client_socket_connect(dl, type, port);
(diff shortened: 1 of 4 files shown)


subsys/net/lib/download_client/src/download_client.c
```diff
@@ -409,14 +340,25 @@ static int client_connect(struct download_client *dl)
...
+		LOG_INF("Connecting to %s", ip_addr_str);
+	}
 	LOG_DBG("fd %d, addrlen %d, fam %s, port %d",
 		dl->fd, addrlen, str_family(dl->remote_addr.sa_family), port);
 
 	err = connect(dl->fd, &dl->remote_addr, addrlen);
 	if (err) {
+		LOG_ERR("Unable to connect, errno %d", errno);
 		err = -errno;
-		LOG_ERR("Unable to connect, errno %d", -err);
```
SeppoTakalo, line 419: Why switch the order? Is there guarantee that LOG() macros don't touch errno?

SeppoTakalo (approved):
It looks kind of OK, but I would prefer to refactor the whole `host_lookup()`.

Currently we only take the first address that is returned.
Then, this seem like still too much hand coding those address families.

We could refactor the `host_lookup()` to run a loop through all results, first one that `connect()` properly, would be the one we use. Then whatever is in `config.family` would be used in hints. So if someone puts `AF_UNSPEC` it works on both IPv4 and IPv6.

subsys/net/lib/download_client/src/download_client.c
stig-bjorlykke: This is the order used for other logs using errno.

stig-bjorlykke:
> We could refactor the `host_lookup()` to run a loop through all results, first one that `connect()` properly, would be the one we use. Then whatever is in `config.family` would be used in hints. So if someone puts `AF_UNSPEC` it works on both IPv4 and IPv6.

When using `AF_UNSPEC` with our modem implementation only the first successful DNS response is returned, so if an IPv6 address is found then this is the only address returned. Because of this we need to do several DNS requests anyway.

I agree we could try all IP addresses returned in case of multiple, but that's not the scope for this change.

subsys/net/lib/download_client/src/download_client.c
rlubos: IMO the original order (read `errno` once before any other action) makes more sense if we want to align the codebase.

rlubos (approved):

lucasvdp (approved):

PR #13621, opened 2024-01-16
ext: fix oberon driver uninitialized variable

ktaborowski:
Fix for gcc warning: 'res' may be used uninitialized

For some reason I only see the warning when CONFIG_DEBUG_OPTIMIZATIONS=y


RobertGalatNordic (approved):

Vge0rge:
It's interesting that you get it only in this file, there are many other place that uninitialized variables exist.
The Oberon code is an external delivery and doing commits on top of it is an overhead for us for each new version.
I informed Oberon about this issue and they will fix it in a later release.
If that causes you CI failures then we can merge it and we will handle this internally.
If it is just to remove the warning I would propose to close this PR and wait until the the next version of Oberon deliverable.

SebastianBoe (approved):
good find, thank you for the patch

PR #13624, opened 2024-01-16
applications: Fixed pincode command for the Matter bridge

kkasperczyk-no:
Fixed the string to int conversion for the Matter bridge pincode command. Before it was not able to correctly convert pincodes with leading 0 characters.

### DIFF
--- a/applications/matter_bridge/src/bridge_shell.cpp
+++ b/applications/matter_bridge/src/bridge_shell.cpp
@@ -199,7 +199,7 @@ static void BluetoothScanResult(Nrf::BLEConnectivityManager::ScannedDevice *devi
 static int InsertBridgedDevicePincodeHandler(const struct shell *shell, size_t argc, char **argv)
 {
 	int bleDeviceIndex = strtoul(argv[0], NULL, 0);
-	unsigned int pincode = strtoul(argv[1], NULL, 0);
+	unsigned int pincode = strtoul(argv[1], NULL, 10);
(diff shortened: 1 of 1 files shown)


applications/matter_bridge/src/bridge_shell.cpp
```diff
@@ -199,7 +199,15 @@ static void BluetoothScanResult(Nrf::BLEConnectivityManager::ScannedDevice *devi
 static int InsertBridgedDevicePincodeHandler(const struct shell *shell, size_t argc, char **argv)
 {
 	int bleDeviceIndex = strtoul(argv[0], NULL, 0);
-	unsigned int pincode = strtoul(argv[1], NULL, 0);
```
Damian-Nordic, line 202: If you only need to support decimal numbers, you can probably pass `10` instead of `0` as the last argument and that should fix the issue. I suppose the issue might have been that 0X numbers were parsed as octal, right?

Damian-Nordic (approved):

applications/matter_bridge/src/bridge_shell.cpp
kkasperczyk-no: Oh right, I haven't thought it is that simple, thanks :) I haven't investigated it deeply, but the problem was that the numbers with leading 0s were not converted, as it would be decimal numbers.

doublemis1 (approved):

PR #13629, opened 2024-01-16
net: lib: nrf_cloud: rename nct_ to nrfcloud_transport_

DematteisGiacomo:
Reason is conflicting naming with zephyr native simulator nct_if.h

### DIFF
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_transport.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_transport.c
@@ -1031,7 +1031,7 @@ static void nct_mqtt_evt_handler(struct mqtt_client *const mqtt_client,
 	}
 }
 
-int nct_init(const char * const client_id)
+int nct_initialize(const char * const client_id)
 {
 	int err;
 
--- a/subsys/net/lib/nrf_cloud/include/nrf_cloud_transport.h
+++ b/subsys/net/lib/nrf_cloud/include/nrf_cloud_transport.h
@@ -67,7 +67,7 @@ struct nct_evt {
 int nct_socket_get(void);
 
 /** @brief Initialization routine for the transport. */
-int nct_init(const char * const client_id);
+int nct_initialize(const char * const client_id);
 
 /** @brief Unintialize the transport; reset state and free allocated memory */
 void nct_uninit(void);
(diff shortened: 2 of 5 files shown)


jayteemo:
what is the actual conflict you are fixing?
is it only `nct_init`?
if so, can you just change the cloud function to `int nct_initialize(const char * const client_id)`

subsys/net/lib/nrf_cloud/src/nrf_cloud_log_backend.c
```diff
@@ -257,7 +257,7 @@ static void logger_process(const struct log_backend *const backend, union log_ms
 
 	logs_init_context(rest_ctx, device_id, level, src_id, src_name, dom_id, ts, &log_context);
 
-	log_output_func = log_format_func_t_get(log_format_current);
+	log_output_func = log_format_funrfct_get(log_format_current);
```
jorgenmk, line 260: This doesn't look correct.

subsys/net/lib/nrf_cloud/src/nrf_cloud_transport.c
```diff
@@ -236,19 +236,19 @@ static bool strings_compare(const char *s1, const char *s2, uint32_t s1_len,
 }
 
 /* Verify if the RX topic is a control channel topic or not. */
-static bool nrf_cloud_cc_rx_topic_decode(const struct mqtt_topic *topic, enum nct_cc_opcode *opcode)
+static bool nrf_cloud_cc_rx_topic_dec(const struct mqtt_topic *topic, enum nrfct_cc_opcode *opcode)
```
jorgenmk, line 239: Why this change? Is it related to max line length?

jorgenmk (changes requested):

subsys/net/lib/nrf_cloud/src/nrf_cloud_log_backend.c
DematteisGiacomo: it is not fixed :+1:

subsys/net/lib/nrf_cloud/src/nrf_cloud_transport.c
DematteisGiacomo: yes, it is due to max line

DematteisGiacomo:
> what is the actual conflict you are fixing? is it only `nct_init`? if so, can you just change the cloud function to `int nct_initialize(const char * const client_id)`

@jayteemo  yes it is only `nct_init`, I'll just rename that one then. 
The idea for a general rename was to avoid future possible conflicts.

jayteemo (approved):

glarsennordic (approved):

plskeggs (approved):

jorgenmk (approved):

PR #13639, opened 2024-01-17
manifest: Use NCS fork of psa-arch-tests instead of Zephyr's repo

SebastianBoe:
Use an NCS fork of psa-arch-tests instead of Zephyr's repo because we need to fromlist some patches while we wait for Zephyr to get the patches.


mswarowsky (approved):

vili-nordic (approved):

cvinayak:
@carlescufi and @tejlmand need your approvals here

tejlmand (approved):

PR #13645, opened 2024-01-17
applications: nrf5340_audio: Issues after upmerge

gWacey:
OCT-2852

After the upmerge, some of the prints where incorrect and there was a increase in the files built. 

The printing issue was due to the change in the return data from bt_audio_codec_cfg_get_freq() and bt_audio_codec_cfg_get_frame_dur(). These now return the frequency and duration as  enums IDs from bt_audio_codec_config_freq and bt_audio_codec_config_frame_dur respectively. Previously these functions returned the Hz and us values. Two new functions have been included that read the IDs and convert them to the required format. 

The number of build files has increased as the CMSIS-DSP library is build with all functions. The functions are in separate files and where optionally built by CONFIGs. These CONFIGs have been removed and the complete library is now built. Resulting in a  vastly increased number of files to built.

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
@@ -145,20 +145,48 @@ static void print_codec(const struct audio_codec_info *codec)
 static void get_codec_info(const struct bt_audio_codec_cfg *codec,
 			   struct audio_codec_info *codec_info)
 {
+	int ret;
+
 	if (codec->id == BT_HCI_CODING_FORMAT_LC3) {
 		/* LC3 uses the generic LTV format - other codecs might do as well */
 		LOG_DBG("Retrieve the codec configuration for LC3");
 		codec_info->id = codec->id;
 		codec_info->cid = codec->cid;
 		codec_info->vid = codec->vid;
-		codec_info->frequency = bt_audio_codec_cfg_get_freq(codec);
-		codec_info->frame_duration_us = bt_audio_codec_cfg_get_frame_dur(codec);
-		bt_audio_codec_cfg_get_chan_allocation(codec, &codec_info->chan_allocation);
-		codec_info->octets_per_sdu = bt_audio_codec_cfg_get_octets_per_frame(codec);
+
+		ret = le_audio_freq_hz_get(codec, &codec_info->frequency);
+		if (ret) {
+			LOG_ERR("Error retrieving sampling frequency: %d", ret);
+			return;
+		}
+
+		ret = le_audio_duration_us_get(codec, &codec_info->frame_duration_us);
+		if (ret) {
+			LOG_ERR("Error retrieving frame duration: %d", ret);
+			return;
+		}
+
+		ret = bt_audio_codec_cfg_get_chan_allocation(codec, &codec_info->chan_allocation);
+		if (ret) {
+			LOG_ERR("Error retrieving channel allocation: %d", ret);
+			return;
+		}
+
+		ret = le_audio_octets_per_frame_get(codec, &codec_info->octets_per_sdu);
+		if (ret) {
+			LOG_ERR("Error retrieving octets per frame: %d", ret);
+			return;
+		}
+
 		codec_info->bitrate =
 			(codec_info->octets_per_sdu * 8 * 1000000) / codec_info->frame_duration_us;
-		codec_info->blocks_per_sdu =
-			bt_audio_codec_cfg_get_frame_blocks_per_sdu(codec, true);
+
+		ret = le_audio_frame_blocks_per_sdu_get(codec, &codec_info->blocks_per_sdu);
+		if (codec_info->octets_per_sdu < 0) {
+			LOG_ERR("Error retrieving frame blocks per SDU: %d",
+				codec_info->octets_per_sdu);
+			return;
+		}
 	} else {
 		LOG_WRN("Codec is not LC3, codec_id: 0x%2hhx", codec->id);
 	}
@@ -166,7 +194,14 @@ static void get_codec_info(const struct bt_audio_codec_cfg *codec,
 
 static bool bitrate_check(const struct bt_audio_codec_cfg *codec)
 {
-	uint32_t octets_per_sdu = bt_audio_codec_cfg_get_octets_per_frame(codec);
+	int ret;
+	uint32_t octets_per_sdu;
+
+	ret = le_audio_octets_per_frame_get(codec, &octets_per_sdu);
+	if (ret) {
+		LOG_ERR("Error retrieving octets per frame: %d", ret);
+		return false;
+	}
 
 	if (octets_per_sdu < LE_AUDIO_SDU_SIZE_OCTETS(CONFIG_LC3_BITRATE_MIN)) {
 		LOG_WRN("Bitrate too low");
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
@@ -34,3 +34,76 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
 
 	return false;
 }
+
+int le_audio_freq_hz_get(const struct bt_audio_codec_cfg *codec, int *freq_hz)
+{
+	int ret;
+
+	ret = bt_audio_codec_cfg_get_freq(codec);
+	if (ret < 0) {
+		*freq_hz = 0;
+		return ret;
+	}
+
+	ret = bt_audio_codec_cfg_freq_to_freq_hz(ret);
+	if (ret < 0) {
+		*freq_hz = 0;
+		return ret;
+	}
+
+	*freq_hz = ret;
+
+	return 0;
+}
+
+int le_audio_duration_us_get(const struct bt_audio_codec_cfg *codec, int *frame_dur_us)
+{
+	int ret;
+
+	ret = bt_audio_codec_cfg_get_frame_dur(codec);
+	if (ret < 0) {
+		*frame_dur_us = 0;
+		return ret;
+	}
+
+	ret = bt_audio_codec_cfg_frame_dur_to_frame_dur_us(ret);
+	if (ret < 0) {
+		*frame_dur_us = 0;
+		return ret;
+	}
+
+	*frame_dur_us = ret;
+
+	return 0;
+}
+
+int le_audio_octets_per_frame_get(const struct bt_audio_codec_cfg *codec, uint32_t *octets_per_sdu)
+{
+	int ret;
+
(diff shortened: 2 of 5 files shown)


applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
```diff
@@ -162,6 +162,28 @@ BUILD_ASSERT(CONFIG_BT_ASCS_ASE_SRC_COUNT <= 1,
...
+	if (freq_id < 0) {
+		return freq_id;
+	}
+
+	return (bt_audio_codec_cfg_freq_to_freq_hz(freq_id));
+}
+
+static int codec_duration_us(const struct bt_audio_codec_cfg *codec)
+{
+	enum bt_audio_codec_config_frame_dur frame_dur = bt_audio_codec_cfg_get_frame_dur(codec);
```
koffes, line 178: Have a look at other places where `bt_audio_codec_cfg_get_freq` and 'bt_audio_codec_cfg_get_frame_dur' are used. Seems the commonly return to a type `int` and not an enum since they can contain negative values.

koffes (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
```diff
@@ -151,8 +173,8 @@ static void get_codec_info(const struct bt_audio_codec_cfg *codec,
 		codec_info->id = codec->id;
 		codec_info->cid = codec->cid;
 		codec_info->vid = codec->vid;
-		codec_info->frequency = bt_audio_codec_cfg_get_freq(codec);
-		codec_info->frame_duration_us = bt_audio_codec_cfg_get_frame_dur(codec);
+		codec_info->frequency = codec_freq_get_hz(codec);
```
koffes, line 176: If `codec_freq_get_hz` returns an error (bt_audio_codec_cfg_get_freq) returns an error, this is then stored directly in frequency with no error check. We need either to check for error or do an assert.

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
```diff
@@ -126,6 +126,28 @@ static void le_audio_event_publish(enum le_audio_evt_type event)
 	ERR_CHK(ret);
 }
 
+static int codec_freq_get_hz(const struct bt_audio_codec_cfg *codec)
```
koffes, line 129: This code is duplicated. We can discuss another place to put it which is accessible for both the broadcast_sink and broadcast_source.

koffes (changes requested):

alexsven (approved):

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
```diff
@@ -130,8 +130,8 @@ static void print_codec(const struct audio_codec_info *codec)
 {
 	if (codec->id == BT_HCI_CODING_FORMAT_LC3) {
 		LOG_INF("Codec config for LC3:");
-		LOG_INF("\tFrequency: %d Hz", codec->frequency);
-		LOG_INF("\tFrame Duration: %d us", codec->frame_duration_us);
+		LOG_INF("\tFrequency: %d Hz", le_audio_freq_get_hz(codec));
```
alexsven, line 133: This is already calculated in the function below, so should just need to print here, same with the with the one below

alexsven (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
gWacey: So this system will always run get_codec_info() before print_codec() and not directly printing the configuration?

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
gWacey: Moved to le_audio.[c, h] as broadcast and unicast can access this, plus it is codec related.

applications/nrf5340_audio/src/bluetooth/bt_stream/broadcast/broadcast_sink.c
alexsven: They are using two different structs, this print function is using our own `audio_codec_info` that is populated in the function `get_codec_info()`

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
```diff
@@ -34,3 +34,39 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
...
+	int dur_us;
+
+	frame_dur = bt_audio_codec_cfg_get_frame_dur(codec);
+	if (frame_dur < 0) {
+		return frame_dur;
+	}
+
+	dur_us = (bt_audio_codec_cfg_frame_dur_to_frame_dur_us(frame_dur));
+	if (dur_us < 0) {
+		LOG_WRN("Invalid codec duration detected, %d", dur_us);
```
koffes, line 68: IMO, all these prints should be LOG_ERR

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
```diff
@@ -34,3 +34,39 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
 
 	return false;
 }
+
+int le_audio_freq_get_hz(const struct bt_audio_codec_cfg *codec)
```
koffes, line 38: Nitpick 1: 
```suggestion
int le_audio_freq_hz_get(const struct bt_audio_codec_cfg *codec)
```

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
```diff
@@ -34,3 +34,39 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
...
+
+	freq_hz = (bt_audio_codec_cfg_freq_to_freq_hz(freq_id));
+	if (freq_hz < 0) {
+		LOG_WRN("Invalid codec frequency detected, %d", freq_hz);
+	}
+
+	return freq_hz;
+}
+
+int le_audio_duration_us(const struct bt_audio_codec_cfg *codec)
```
koffes, line 56: For consistency:
```suggestion
int le_audio_duration_us_get(const struct bt_audio_codec_cfg *codec)
```

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
```diff
@@ -167,6 +167,7 @@ static void print_codec(const struct bt_audio_codec_cfg *codec, enum bt_audio_di
 	if (codec->id == BT_HCI_CODING_FORMAT_LC3) {
 		/* LC3 uses the generic LTV format - other codecs might do as well */
 		enum bt_audio_location chan_allocation;
+		int dur_us = le_audio_duration_us(codec);
```
koffes, line 170: When these functions are called, we need to perform a check: `dur_us <= 0`
If so, return an error.

If `le_audio_duration_us` returns an error here, we will print the negative number.

koffes (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
gWacey: If I assert in the two get functions for an error with the ID or actual Hz/us value this would simplify the code and trap this error. Would this be acceptable?

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
gWacey: If I change this to an assert then it traps the error, would this be acceptable?

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
koffes: Reccomend we do the usual err_check as an Assert will reboot the device, and it quite harsh.

koffes: We let the layer above decide which action to take.

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
```diff
@@ -34,3 +34,84 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
...
+
+	ret = bt_audio_codec_cfg_get_octets_per_frame(codec);
+	if (ret < 0) {
+		*octets_per_sdu = 0;
+		return ret;
+	}
+
+	*octets_per_sdu = ret;
+
+	if (*octets_per_sdu < LE_AUDIO_SDU_SIZE_OCTETS(CONFIG_LC3_BITRATE_MIN)) {
```
koffes, line 92: The purpose of this function is to get the value and not check if it is valid.
In broadcast_sink this check is still performed. 
Remove line 92-98 and add the check back in unicast_server.

koffes (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.h
```diff
@@ -63,12 +63,53 @@ struct le_audio_encoded_audio {
  *		If the endpoint is NULL, it is not in the
  *		given state, and this function returns false.
  *
- * @param[in]	ep	The endpoint to check.
+ * @param[in]	ep		The endpoint to check.
```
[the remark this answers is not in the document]
gWacey: This is strange and looks like weird tabs!

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
```diff
@@ -34,3 +34,84 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
...
+	ret = bt_audio_codec_cfg_freq_to_freq_hz(ret);
+	if (ret < 0) {
+		*freq_hz = 0;
+		return ret;
+	}
+
+	*freq_hz = ret;
+
+	return 0;
+}
```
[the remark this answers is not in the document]
gWacey: This is better as the bt_audio_codec_cfg_get_freq(codec) will check the id gives a valid frequency, thus bt_audio_codec_cfg_freq_to_freq_hz(codec) will always pass.

koffes (approved):

PR #13653, opened 2024-01-18
drivers: sensor: bme68x_iaq: Add additional bsec channels and q-ULP mode

purpl3F0x:
This patch adds support for BSEC CO2 and VOC channels, as long as channels for indication of gas sensor status and accuracy of measurements. It also adds support for q-ULP mode, which samples for gasses every 300seconds and every 3s for temperature, pressure and humidity.

This commit is more of a draft at this point since there is no way for the user to read some of the measurements.
Maybe `bme_sample_result` should be moved to `include/drivers/bm68x_iaq.h ` so there is access to the raw data (maybe exposing a lock, unlock when user wants to access dev->data)

Also do the values inside `bme_sample_result` need to be double ? bsec returns floats


purpl3F0x:
Practically isn't this trigger SENSOR_TRIG_DATA_READY ?
https://github.com/nrfconnect/sdk-nrf/blob/3d4ff5711c49e2740b1cb60b8e3c05e1c9b52667/drivers/sensor/bme68x_iaq/bme68x_iaq.c#L515

maxd-nordic:
> Practically isn't this trigger SENSOR_TRIG_DATA_READY ?
> 
> https://github.com/nrfconnect/sdk-nrf/blob/3d4ff5711c49e2740b1cb60b8e3c05e1c9b52667/drivers/sensor/bme68x_iaq/bme68x_iaq.c#L515

I guess in that case you could use either of them when implementing the driver.

drivers/sensor/bme68x_iaq/bme68x_iaq.c
```diff
@@ -554,6 +594,10 @@ static int bme68x_channel_get(const struct device *dev,
 	} else if (chan == SENSOR_CHAN_IAQ) {
 		val->val1 = data->latest.air_quality;
 		val->val2 = 0;
+	} else if (chan == SENSOR_CHAN_CO2) {
+		sensor_value_from_double(val, data->latest.co2);
+	}else if (chan == SENSOR_CHAN_VOC) {
```
maxd-nordic, line 599:
```suggestion
	} else if (chan == SENSOR_CHAN_VOC) {
```

drivers/sensor/bme68x_iaq/bme68x_iaq.c
```diff
@@ -521,6 +559,8 @@ static int bme68x_trigger_set(const struct device *dev,
 	    || (trig->chan == SENSOR_CHAN_HUMIDITY)
 	    || (trig->chan == SENSOR_CHAN_AMBIENT_TEMP)
 	    || (trig->chan == SENSOR_CHAN_PRESS)
+		|| (trig->chan == SENSOR_CHAN_CO2)
```
maxd-nordic, line 562:
```suggestion
	    || (trig->chan == SENSOR_CHAN_CO2)
```

drivers/sensor/bme68x_iaq/bme68x_iaq.c
```diff
@@ -521,6 +559,8 @@ static int bme68x_trigger_set(const struct device *dev,
...
+		|| (trig->chan == SENSOR_CHAN_VOC)
```
maxd-nordic, line 563:
```suggestion
	    || (trig->chan == SENSOR_CHAN_VOC)
```

include/drivers/bme68x_iaq.h
```diff
@@ -19,6 +19,13 @@ extern "C" {
 
 #define SENSOR_CHAN_IAQ (SENSOR_CHAN_PRIV_START+1)
 
+enum bme68x_accuracy {
+    BME68X_ACCURACY_UNRELIABLE = 0,
```
maxd-nordic, line 23:
```suggestion
	BME68X_ACCURACY_UNRELIABLE = 0,
```

include/drivers/bme68x_iaq.h
```diff
@@ -19,6 +19,13 @@ extern "C" {
...
+    BME68X_ACCURACY_LOW,
```
maxd-nordic, line 24:
```suggestion
	BME68X_ACCURACY_LOW,
```

include/drivers/bme68x_iaq.h
```diff
@@ -19,6 +19,13 @@ extern "C" {
...
+    BME68X_ACCURACY_MEDIUM,
```
maxd-nordic, line 25:
```suggestion
	BME68X_ACCURACY_MEDIUM,
```

include/drivers/bme68x_iaq.h
```diff
@@ -19,6 +19,13 @@ extern "C" {
...
+    BME68X_ACCURACY_HIGH,
```
maxd-nordic, line 26:
```suggestion
	BME68X_ACCURACY_HIGH,
```

maxd-nordic (commented):
whitespace stuff

purpl3F0x:
> whitespace stuff

Sorry about those, but if I run clang-format, it formats quite a few parts of the file(s).


<br>

> For accessing the additional data, we could define custom sensor channels, just like `SENSOR_CHAN_IAQ`. That would uphold the sensor API access.

Okay, will do it this way. I suppose the status quo is to use the sensor api, despite adding some overhead

<br>

> > Practically isn't this trigger SENSOR_TRIG_DATA_READY ?
> > https://github.com/nrfconnect/sdk-nrf/blob/3d4ff5711c49e2740b1cb60b8e3c05e1c9b52667/drivers/sensor/bme68x_iaq/bme68x_iaq.c#L515
> 
> I guess in that case you could use either of them when implementing the driver.

Will add them both, to avoid get something broken.


<br><br>

There are also 1 other thing the library could add, BSEC currently it doesn't use the predefined configurations bsec provides that tune the internals of the library according to the operating voltage of the sensor/mode/Calibration Time etc..., For some reason they are not included in the generic bsec library, but  we could add the their Arduino library in the west dependencies.

maxd-nordic:
> > whitespace stuff
> 
> Sorry about those, but if I run clang-format, it formats quite a few parts of the file(s).
> 
> > For accessing the additional data, we could define custom sensor channels, just like `SENSOR_CHAN_IAQ`. That would uphold the sensor API access.
> 
> Okay, will do it this way. I suppose the status quo is to use the sensor api, despite adding some overhead
> 
> > > Practically isn't this trigger SENSOR_TRIG_DATA_READY ?
> > > https://github.com/nrfconnect/sdk-nrf/blob/3d4ff5711c49e2740b1cb60b8e3c05e1c9b52667/drivers/sensor/bme68x_iaq/bme68x_iaq.c#L515
> > 
> > 
> > I guess in that case you could use either of them when implementing the driver.
> 
> Will add them both, to avoid get something broken.
> 
> There are also 1 other thing the library could add, BSEC currently it doesn't use the predefined configurations bsec provides that tune the internals of the library according to the operating voltage of the sensor/mode/Calibration Time etc..., For some reason they are not included in the generic bsec library, but we could add the their Arduino library in the west dependencies.

Do you mean these? https://github.com/boschsensortec/BSEC-Arduino-library/tree/21213272be0fcb8f6d886089b38068b1ae25c3de/src/config
I wonder what they actually do - seems they ran some calibrations with a bunch of setups on their own boards?

purpl3F0x:
> > There are also 1 other thing the library could add, BSEC currently it doesn't use the predefined configurations bsec provides that tune the internals of the library according to the operating voltage of the sensor/mode/Calibration Time etc..., For some reason they are not included in the generic bsec library, but we could add the their Arduino library in the west dependencies.
> 
> Do you mean these? https://github.com/boschsensortec/BSEC-Arduino-library/tree/21213272be0fcb8f6d886089b38068b1ae25c3de/src/config I wonder what they actually do - seems they ran some calibrations with a bunch of setups on their own boards? I think there should be no need for these. Each board will yield a different calibration. If anything, one can prepare an initial calibration for an own board.

Yes I'm bit sure either. I think if not provided bsec internally sets the 3.3V 300s one. It needs further investigation, so let's leave this commit additions as is.

purpl3F0x:
> > > There are also 1 other thing the library could add, BSEC currently it doesn't use the predefined configurations bsec provides that tune the internals of the library according to the operating voltage of the sensor/mode/Calibration Time etc..., For some reason they are not included in the generic bsec library, but we could add the their Arduino library in the west dependencies.
> > 
> > 
> > Do you mean these? https://github.com/boschsensortec/BSEC-Arduino-library/tree/21213272be0fcb8f6d886089b38068b1ae25c3de/src/config I wonder what they actually do - seems they ran some calibrations with a bunch of setups on their own boards? I think there should be no need for these. Each board will yield a different calibration. If anything, one can prepare an initial calibration for an own board.
> 
> Yes I'm not sure either. I think if not provided bsec internally sets the 3.3V 300s one. It needs further investigation, so let's leave this commit additions as is.
> 
> P. S. Should the commit message be trimmed to pass compliance?

After some googling, it seems these configurations have 2 main purposes. 
- Practically 28days vs 4 days are for static vs portable implementations and will significantly affect the IAQ results.
- The voltage variable is to accommodate for the gas heater affection to the temperature measurements. That mainly seems to matter on LP mode where measurements are 3s apart. In ULP the effect is irrelevant since there is a 5 minute rest period and temp measurement proceeds the gas heater anyways.

purpl3F0x:
> Apparently, 72 chars is the line limit there.
I did some trimming 🧙‍♂️


> I didn't see a voltage input to bsec, so I'm assuming this is referring to a test setup.

It's the voltage the device is powered by.
I suppose the simplest thing is to have a KConfig magic with options for manually setting the config if wanted. Or else leave the driver as is :)

maxd-nordic (approved):

bjda (approved):

PR #13659, opened 2024-01-18
samples: matter: binding: Release binding implementation

LipinskiPNordicSemi:
This commit implements the possibility to add a callback function called when binding with device is released. Additionally, implementation of this solution for a thermostat is included.

### DIFF
--- a/samples/matter/common/src/binding/binding_handler.cpp
+++ b/samples/matter/common/src/binding/binding_handler.cpp
@@ -15,149 +15,176 @@ using namespace chip::app;
 
 namespace Nrf::Matter
 {
-	void BindingHandler::Init()
-	{
-		InitInternal();
-	}
+void BindingHandler::Init()
+{
+	InitInternal();
+}
 
-	void BindingHandler::RunBoundClusterAction(BindingData *bindingData)
-	{
-		VerifyOrReturn(bindingData != nullptr, LOG_ERR("Invalid binding data"));
-		VerifyOrReturn(bindingData->InvokeCommandFunc != nullptr,
-			       LOG_ERR("No valid InvokeCommandFunc assigned"););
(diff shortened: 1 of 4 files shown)


samples/matter/common/src/binding/binding_handler.cpp
```diff
@@ -15,149 +15,176 @@ using namespace chip::app;
...
+	mReleaseCallbackPointer = bindingReleaseCallback;
+
+	return CHIP_NO_ERROR;
+}
+
+void BindingHandler::BindingMatterEventHandler(const DeviceLayer::ChipDeviceEvent *event, intptr_t /* unused */)
+{
+	switch (event->Type) {
+	case DeviceLayer::DeviceEventType::kBindingsChangedViaCluster:
+		if (BindingTable::GetInstance().Size() == 0) {
```
Damian-Nordic, line 174: This won't really work if there are multiple bindings. Not sure if this can be made generically without modifying the Matter core as the event carries no information about the removed binding, so only the application can react to `kBindingsChangedViaCluster` properly by checking if the concerned binding is still present in the table.

LipinskiPNordicSemi: With kBindingsChangedViaCluster event, only fabricIndex value is given. Is it enough to identify which connection changed?

Damian-Nordic: Unfortunately not. kBindingsChangedViaCluster is generated once per entire binding table update and there might be multiple entries changed in one update. I guess you could replace `ReleaseCallback` with `UpdateCallback` and have the application determine in `UpdateCallback` if an entry for your thermostat exists.

LipinskiPNordicSemi: Yeah, that's good idea. Thanks!

PR #13664, opened 2024-01-18
samples: bluetooth: throughput: Negotiate con params before packet len

rakons:
This commit changes the order of the connection configuration. Negotiate the connection params with higher connection latency before changing the packet len.
This way we may be sure that the connection would not break due to timeout after the high packet len was set that cannot fit into configured connection latency.

JIRA: NCSDK-25371

### DIFF
--- a/samples/bluetooth/throughput/src/main.c
+++ b/samples/bluetooth/throughput/src/main.c
@@ -502,24 +502,6 @@ static int connection_configuration_set(const struct shell *shell,
 		return err;
 	}
 
-	if (info.le.data_len->tx_max_len != data_len->tx_max_len) {
-		data_length_req = true;
-
-		err = bt_conn_le_data_len_update(default_conn, data_len);
-		if (err) {
-			shell_error(shell, "LE data length update failed: %d",
-				    err);
-			return err;
-		}
-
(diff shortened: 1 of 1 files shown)


rakons:
The question is what if the user after the first run, would set the connection time to be short, but the payload to be short too. Then it would fail also. Do we wish to handle that? Now it runs with the default settings and in the situation mentioned here - it would work after the connection restart.

rafpelc (approved):
Tests on this PR passed, so from QA perspective it is good to merge

KAGA164 (approved):
LGTM

KAGA164:
> The question is what if the user after the first run, would set the connection time to be short, but the payload to be short too. Then it would fail also. Do we wish to handle that? Now it runs with the default settings and in the situation mentioned here - it would work after the connection restart.

@rakons Why will short payload with short connection interval fail?

rakons:
> Why will short payload with short connection interval fail?

It would happen if this is second run, after the first one with long payload. If we (after running the test once) change the settings to short latency and small payload, then on second run, we would set short latency first (having still long payload from previous run) and at this moment we would get timeout. To fix this we need to reset the payload before configuring. I will create separate task for that.

PR #13670, opened 2024-01-19
samples: matter: replace std::function with custom task

Damian-Nordic:
std::function<void()> is currently used as a task that is to be executed in the application thread, but this class is not trivially copyable so passing it over Zephyr's msgq may lead to resource leaks or other unexpected behavior in certain circumstances (for example, when a functor that std::function is initialized with is big enough that it cannot fit in the std::function static storage and must be allocated on heap).

Replace std::function with a lightweight replacement that ensures that it is only used with simple lambda functions that can be safely serialized.

### DIFF
--- a/samples/matter/common/src/Kconfig
+++ b/samples/matter/common/src/Kconfig
@@ -11,6 +11,13 @@ config NCS_SAMPLE_MATTER_APP_TASK_QUEUE_SIZE
 	  Define the maximum size of the queue dedicated for application tasks that
 	  have to be run in the application thread context.
 
+config NCS_SAMPLE_MATTER_APP_TASK_MAX_SIZE
+	int "Maximum size of application task in bytes"
+	default 16
+	help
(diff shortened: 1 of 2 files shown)


samples/matter/common/src/Kconfig
```diff
@@ -11,6 +11,13 @@ config NCS_SAMPLE_MATTER_APP_TASK_QUEUE_SIZE
 	  Define the maximum size of the queue dedicated for application tasks that
 	  have to be run in the application thread context.
 
+config NCS_SAMPLE_MATTER_APP_TASK_MAX_SIZE
+	int "Maximum size of application task"
```
kkasperczyk-no, line 15:
```suggestion
	int "Maximum size of application task in bytes"
```

kkasperczyk-no (approved):

ArekBalysNordic (approved):

PR #13674, opened 2024-01-19
ci: gitlint: Enforce Jira mention in commit log

krish2718:
This checks for a valid Jira ID in the commit log with two prefixes
    (Fixes or Implements), this helps in git-jira integration.
    
    Implements NCSDK-25550.


carlescufi (changes requested):
Let's bring it up in the next all-hands meeting.

PR #13683, opened 2024-01-22
samples: cellular: modem_shell: Flag reset reason printing

tokangas:
Reset reason can only be read once, because the register needs to be cleared after reading. Memfault implementation reads the reset reason before modem_shell, so the register was always empty and the reset reason was not correctly printed. Added flagging for reset reason printing to skip it when Memfault is enabled.


trantanen (approved):
Not nice but I guess there are no convenient alternatives.

jhirsi (approved):

PR #13686, opened 2024-01-22
lib: nrf_modem: fault: add initialization retry limit

eivindj-nordic:
Add a limit to the number of times the application will try to reinit the modem after a fault. This is usefull when a fault occurs during initialization.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -720,6 +720,7 @@ Modem libraries
   * Updated by renaming ``lte_connectivity`` module to ``lte_net_if``.
     All related Kconfig options have been renamed accordingly.
   * Changed the default value of the :kconfig:option:`CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_START`, :kconfig:option:`CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_CONNECT`, and :kconfig:option:`CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_DOWN` Kconfig options from enabled to disabled.
+  * The :kconfig:option:`CONFIG_NRF_MODEM_LIB_ON_FAULT_RESET_MODEM_RETRIES` Kconfig option to choose the number of times the modem fault handler will attempt to initialize the modem after a fault.
 
   * Fixed:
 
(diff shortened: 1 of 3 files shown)


lemrey (commented):
I'm wondering if we should retry at all.. Perhaps we can check the return value of the initialization function, and if it is an error then log an error, exit the loop and return from the thread.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -683,6 +683,7 @@ Modem libraries
   * Updated by renaming ``lte_connectivity`` module to ``lte_net_if``.
     All related Kconfig options have been renamed accordingly.
   * Changed the default value of the :kconfig:option:`CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_START`, :kconfig:option:`CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_CONNECT`, and :kconfig:option:`CONFIG_NRF_MODEM_LIB_NET_IF_AUTO_DOWN` Kconfig options from enabled to disabled.
+  * The Kconfig option :kconfig:option:`CONFIG_NRF_MODEM_LIB_ON_FAULT_RESET_MODEM_RETRIES` to choose the number of times the modem fault handler will attempt to initialize the modem after a fault.
```
MirkoCovizzi, line 686:
```suggestion
  * The :kconfig:option:`CONFIG_NRF_MODEM_LIB_ON_FAULT_RESET_MODEM_RETRIES` Kconfig option to choose the number of times the modem fault handler will attempt to initialize the modem after a fault.
```

eivindj-nordic:
> I'm wondering if we should retry at all.. Perhaps we can check the return value of the initialization function, and if it is an error then log an error, exit the loop and return from the thread.

For some errors that makes sense, as the modem will not succeed in the next initialization (e.g. RPC version mismatch, modem firmware authentication failure and modem firmware update failure). There are also errors that should not occur in this case: 
* EALREADY as we have called shutdown, 
* EFAULT, ENOLCK and EINVAL as they are returned by the modem library, and we keep the modem library in this case. 
 
Though, we also have ETIMEDOUT and EAGAIN. While it might be unlikely that these also succeeds on a retry, there is a chance. Though, for EAGAIN it might make sense to reschedule it at a later time for when the voltage has increased. Maybe this is too complex to keep as the default, and should be handled in an application specific handler. Though the question is how many will use this functionality directly. If it can save some (live) applications from stopping to work after a fault, I'd say lets keep it.

PR #13688, opened 2024-01-22
samples: matter: Extended BT bridge recovery implementation

kkasperczyk-no:
The bridge BT recovery mechanism was refactored to:
* add backoff mechanism, which makes it scan the less frequently, the longer there are no lost devices detected
* optimize the recovery time by not repeating scan operation before adding every device

### DIFF
--- a/samples/matter/common/src/bridge/ble_bridged_device.h
+++ b/samples/matter/common/src/bridge/ble_bridged_device.h
@@ -14,7 +14,8 @@
 #include <zephyr/bluetooth/conn.h>
 #include <zephyr/bluetooth/gatt.h>
 
-namespace Nrf {
+namespace Nrf
+{
 
 struct BLEBridgedDeviceProvider;
 
@@ -66,7 +67,7 @@ public:
 	bool IsInitiallyConnected() { return mDevice.mInitiallyConnected; }
 
 	/**
-	 * @brief Confirm that the @ref mFirstConnectionCallback has been already called .
+	 * @brief Confirm that the @ref mFirstConnectionCallback has been already called.
 	 *
 	 * This method informs the bridged device that the @ref mFirstConnectionCallback has been called
 	 * and connection has been established successfully.
@@ -76,8 +77,35 @@ public:
 
 	bt_addr_le_t GetBtAddress() { return mDevice.mAddr; }
 
+	/**
+	 * @brief Get a number of failed recovery attempts for this provider.
+	 */
+	uint16_t GetFailedRecoveryAttempts() { return mFailedRecoveryAttempts; }
+
+	/**
+	 * @brief Inform provider that recovery attempt for it failed.
+	 *
+	 * This method increments the number of failed recovery attempts.
+	 *
+	 */
+	void NotifyFailedRecovery()
+	{
+		if (mFailedRecoveryAttempts < UINT16_MAX) {
+			mFailedRecoveryAttempts++;
+		}
+	}
+
+	/**
+	 * @brief Inform provider that recovery attempt for it succeeded.
+	 *
+	 * This method resets the number of failed recovery attempts.
+	 *
+	 */
+	void NotifySuccessfulRecovery() { mFailedRecoveryAttempts = 0; }
+
 protected:
 	BLEBridgedDevice mDevice = { 0 };
+	uint16_t mFailedRecoveryAttempts = 0;
 };
 
 } /* namespace Nrf */
--- a/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
+++ b/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
@@ -56,16 +56,9 @@ void BLEConnectivityManager::FilterMatch(bt_scan_device_info *device_info, bt_sc
 		return;
 	}
 
-	if (filter_match->addr.match) {
-		/* Result comes from the recovery mechanism */
-		Instance().mRecovery.mScannedDevice.mAddr = *filter_match->addr.addr;
-		Instance().mRecovery.mScannedDevice.mConnParam = *device_info->conn_param;
-		Instance().mRecovery.mAddressMatched = true;
-		return;
-	}
-
 	auto scannedDevices = Instance().mScannedDevices;
 	auto scannedDevicesCounter = Instance().mScannedDevicesCounter;
+
 	/* Limit the number of devices that can be scanned. */
 	if (scannedDevicesCounter >= kMaxScannedDevices) {
 		return;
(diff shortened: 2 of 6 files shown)


samples/matter/common/src/bridge/ble_bridged_device.h
```diff
@@ -76,8 +77,35 @@ class BLEBridgedDeviceProvider : public BridgedDeviceDataProvider {
...
+	 */
+	uint16_t GetFailedRecoveryAttempts() { return mFailedRecoveryAttempts; }
+
+	/**
+	 * @brief Inform provider that recovery attempt for it failed.
+	 *
+	 * This method increments the number of failed recovery attempts.
+	 *
+	 */
+	void NotifyFailedRecovery()
```
ArekBalysNordic, line 91: We didn't notify anyone within this method, right?  What about renaming it to `IncreaseRecoveryAttempts`

samples/matter/common/src/bridge/ble_bridged_device.h
```diff
@@ -76,8 +77,35 @@ class BLEBridgedDeviceProvider : public BridgedDeviceDataProvider {
...
+		}
+	}
+
+	/**
+	 * @brief Inform provider that recovery attempt for it succeeded.
+	 *
+	 * This method resets the number of failed recovery attempts.
+	 *
+	 */
+	void NotifySuccessfulRecovery() { mFailedRecoveryAttempts = 0; }
```
ArekBalysNordic, line 104: The same here. I don't see any notification/callback here. What about renaming it to `ResetRecoveryAttempts`?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -118,17 +113,27 @@ void BLEConnectivityManager::ConnectionHandler(bt_conn *conn, uint8_t conn_err)
 		}
 	}
 
-	/* The device with given address was not found.  */
+	/* The device with given address was not found. */
```
ArekBalysNordic, line 116:
```suggestion
	/* The device with the given address was not found. */
```

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -444,30 +448,103 @@ CHIP_ERROR BLEConnectivityManager::Scan(ScanDoneCallback callback, void *context
...
+		if (!item) {
+			return false;
+		}
+
+		if (item->mProvider == provider) {
+			LOG_ERR("Provider found on a list to recover");
+			return true;
+		}
+		node = sys_slist_peek_next(node);
+	} while (node != nullptr);
```
ArekBalysNordic, line 499: Can we use Single-linked API for searching: https://docs.zephyrproject.org/apidoc/latest/group__single-linked-list__apis.html#gad6f1014e26d6cf9925d00b4f53076116 ?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -444,30 +448,103 @@ CHIP_ERROR BLEConnectivityManager::Scan(ScanDoneCallback callback, void *context
...
+	BLEBridgedDeviceProvider *provider;
+
+	sys_snode_t *node = sys_slist_peek_head(&Instance().mRecovery.mListToRecover);
+	Recovery::ListItem *item;
+	bool providerFound = false;
+	bt_addr_le_t providerAddress;
+
+	/* Filter out devices to recover from a scan result and put them to the queue to reconnect. */
+	do {
+		item = CONTAINER_OF(node, __typeof__(*item), mNode);
```
ArekBalysNordic, line 515: Ditto. And also, if we cannot use Zephyr's API maybe we can create a short helper method to not duplicate a code?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -444,30 +448,103 @@ CHIP_ERROR BLEConnectivityManager::Scan(ScanDoneCallback callback, void *context
...
+								 &Instance().mRecovery.mListToReconnect);
 			}
 		}
-		/* The device still should be recovered so put it back into the recovery ring buffer */
-		Instance().mRecovery.NotifyProviderToRecover(provider);
-	} else if (Instance().mRecovery.IsNeeded()) {
-		/* There is another device that needs to be recovered */
+
+		/* Increase failed attempts counter of all devices to recover that were not detected. */
+		if (!providerFound) {
```
ArekBalysNordic, line 532: Do you measure the time of those search operations for a high amount of devices, for example, 100 simulated? I wonder if it doesn't block anything.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -514,9 +591,17 @@ CHIP_ERROR BLEConnectivityManager::Reconnect(BLEBridgedDeviceProvider *provider)
...
 
 	bt_conn *conn;
+	bt_le_conn_param *connParams = GetScannedDeviceConnParams(provider->GetBLEBridgedDevice().mAddr);
+
+	char addrStr[BT_ADDR_LE_STR_LEN];
+	bt_addr_le_to_str(&provider->GetBLEBridgedDevice().mAddr, addrStr, sizeof(addrStr));
 
-	int err = bt_conn_le_create(&Instance().mRecovery.mScannedDevice.mAddr, create_param,
-				    &Instance().mRecovery.mScannedDevice.mConnParam, &conn);
+	if (!connParams) {
```
ArekBalysNordic, line 599: Nit: Move to 595?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -674,67 +759,107 @@ bt_le_conn_param *BLEConnectivityManager::GetScannedDeviceConnParams(bt_addr_le_
...
 {
-	uint32_t deviceAddr = 0;
-	int ret = ring_buf_get(&mRingBuf, reinterpret_cast<uint8_t *>(&deviceAddr), sizeof(deviceAddr));
-	if (ret == sizeof(deviceAddr)) {
-		return reinterpret_cast<BLEBridgedDeviceProvider *>(deviceAddr);
+	sys_snode_t *node = sys_slist_get(list);
+	BLEBridgedDeviceProvider * provider = nullptr;
+
+	if (!node) {
+		return nullptr;
```
ArekBalysNordic, line 797: What happens if the method returns here and we don't call delete on an item?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -674,67 +759,107 @@ bt_le_conn_param *BLEConnectivityManager::GetScannedDeviceConnParams(bt_addr_le_
...
-		int ret = ring_buf_peek(&mRingBuf, reinterpret_cast<uint8_t *>(&deviceAddr), sizeof(deviceAddr));
-		if (ret == sizeof(deviceAddr)) {
-			mCurrentAddr = reinterpret_cast<BLEBridgedDeviceProvider *>(deviceAddr)->GetBtAddress();
-			return &mCurrentAddr;
+	sys_snode_t *node = sys_slist_peek_head(&Instance().mRecovery.mListToRecover);
+	ListItem *item;
+	uint16_t attempts = UINT16_MAX;
+
+	/* Find the smallest number or failed recovery attempts from all providers. */
+	do {
```
ArekBalysNordic, line 835: ditto

samples/matter/common/src/bridge/ble_bridged_device.h
kkasperczyk-no: I mean, notify the provider that recovery attempt failed/succeeded, so internally it may take some actions due to that. I actually followed the naming scheme used by you in the `Recovery`. We have there `NotifyProviderToRecover` method that is used to notify `Recover` module that new provider was lost.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: I haven't noticed this API. It looks good, I will give it a try!

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: This code is not used by the simulated devices, so for now max we should care about is 19. Well, we could try to schedule it to execute in the CHIP thread. I didn't want increase the code complexity, but I can try to address it.

ArekBalysNordic (approved):

markaj-nordic (approved):

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -36,39 +38,50 @@ class BLEConnectivityManager {
...
+		/* Recovery intervals in seconds. */
+		constexpr static auto kRecoveryInterval = 1;
+		constexpr static auto kRecoveryMaxInterval = CONFIG_BRIDGE_BT_RECOVERY_MAX_INTERVAL;
+
 		constexpr static auto kRecoveryScanTimeoutMs = CONFIG_BRIDGE_BT_RECOVERY_SCAN_TIMEOUT_MS;
 
+		struct ListItem {
+			sys_snode_t mNode;
+			BLEBridgedDeviceProvider *mProvider = nullptr;
+		};
```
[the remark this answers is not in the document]
kkasperczyk-no: Thanks for suggestion, I applied it.

PR #13696, opened 2024-01-22
[DNM] Test for the CI

PerMac:
Let's see if this fails CI.
This reverts commit c88a6e6b227076e78db3caacdb4aa925f0476f0b.


PerMac:
The CI failed as it should

PR #13709, opened 2024-01-23
applications: matter: Add unpair to BLE disconnection handler.

ArekBalysNordic:
We need to unpair the bounded device while the connection is closed.
Without that, the newconnection to the same device causes a fault.

### DIFF
--- a/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
+++ b/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
@@ -34,7 +34,8 @@ static const bt_conn_auth_cb auth_callbacks = {
 	.cancel = Nrf::BLEConnectivityManager::AuthenticationCancel,
 };
 
-static bt_conn_auth_info_cb conn_auth_info_callbacks = { .pairing_complete = Nrf::BLEConnectivityManager::PairingComplete,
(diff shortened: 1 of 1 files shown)


samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -159,6 +161,13 @@ void BLEConnectivityManager::DisconnectionHandler(bt_conn *conn, uint8_t reason)
 	bt_addr_le_to_str(bt_conn_get_dst(conn), addrStr, sizeof(addrStr));
 	LOG_INF("Disconnected: %s (reason %u)", addrStr, reason);
 
+#ifdef CONFIG_BT_SMP
+	int ret = bt_unpair(BT_ID_DEFAULT, bt_conn_get_dst(conn));
```
kkasperczyk-no, line 165: It's definitely not a good idea to unpair on disconnect. It will result in starting pairing again on every re-connect or reboot and this is not what we want to do.
I think we should only unpair the device if it was permanently removed from the bridge (not temporarily disconnected).

ArekBalysNordic: Ok, repaired.

kkasperczyk-no (approved):

LuDuda (approved):

PR #13716, opened 2024-01-24
Change certificate PEM files to PEM format

lucasvdp:
- Removed the quotes and comments out of the PEM files that should not be there.
- Automatically generate hex files that can be included.


jtguggedal:
This is a nice improvement! I think you can also move `Amazon root CA 1` from the download sample into the common certs folder. And the common certs folder should probably be documented somewhere, though not sure quite where to put it

lucasvdp:
- I moved the certificate from the download sample as well.
- For now I undid the changes to the mqtt sample, with the different overlays this requires a lot more changes. I suggest to limit this PR to only the root certificates.
- If someone has suggestions about what to document where for the common certificate folder let me know.

lemrey:
This is fine, but I remember I had discussed moving the certs to a common folder some time ago, and one argument I found against doing that is that it becomes a bit harder to know whether a certificate is still used or not. If you modify a sample to update the certificate it is using, you don't know right away what to do with the old certificate. You can't just remove it, because there might other samples still using it, so you need to look at the other TLS samples too (which now are in two folders, `samples/cellular/` and `samples/net/`).

On the other hand, keeping the certificate next to each sample makes that task easier, and it avoids creating a `common` folder under `samples/` which I don't like. There is a risk that others might want to put things there, which are not certificates. If we do end putting the certificates in one common folder, I argue we it should be called, e.g. `samples/certificates` so that there is no ambiguity as to what can be put there.

What are the disadvantages to using a dedicated folder beside each sample to store the certificates, beyond have more files checked in `git` ?

lucasvdp:
I got the request for putting the certificates in a common place, while I was touching them anyway. But don't really care either way myself.
It is nice to keep the sample folders smaller and distinguish between files that might need editing by the user and files that should never be changed.
On the other hand which certificate to use is something they likely need to change and just having it in the sample itself does emphasise that.

The common folder already existed, I just added the certificates folder inside there.

gregersrygg (approved):
This looks like a nice improvement to me. Regarding where to document it, I think it would be nice to document this in each sample where it is used. It's not necessarily obvious to the user of the sample that the certificate is included using CMakeLists.txt.
This section should also be updated: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/libraries/modem/modem_key_mgmt.html#certificates

lemrey:
I think we should keep the certificates beside each sample, I do not see any advantages putting them all in one folder.

I could see a point in doing that if we had 20 samples, connecting to the same server with the same certificate.
Then if that server changed its certificate, you'd only need to update the certificate in one place rather than 20.
But right now we have only 4 samples which need a certificate, and they don't even all connect to the same server.

Plus I dislike "common" folders, because they tend to become dumping grounds for things people don't know where to put.

lucasvdp:
Thanks,
- I removed the second commit with the certificate move to a common folder.
- Added a few lines of documentation to the samples.
- Changed the modem_key_mgmt.rst document.

jorgenmk (approved):
LGTM, tested working on https_client and download samples

nordic-hani (approved):

samples/cellular/ciphersuites/README.rst
```diff
@@ -36,6 +36,8 @@ Obtaining a certificate
 
 The sample connects to ``www.example.com``, which requires an X.509 certificate.
 This certificate is provided in the :file:`samples/cellular/https_client/cert` folder.
```
divipillai, line 38:
```suggestion
This certificate is provided in the :file:`samples/cellular/ciphersuites/cert` folder.
```

samples/cellular/ciphersuites/README.rst
```diff
@@ -36,6 +36,8 @@ Obtaining a certificate
...
+It is automatically converted to a hex file in :file:`CMakeLists.txt`.
+This hex file is included in :file:`main.c` where it is provisioned to the modem.
```
divipillai, line 40:
```suggestion
It is automatically converted to a hex file in :file:`CMakeLists.txt`.
This hex file is included in :file:`main.c`, where it is provisioned to the modem.
```

samples/net/https_client/README.rst
```diff
@@ -36,6 +36,8 @@ Obtaining a certificate
 
 The sample connects to ``www.example.com``, which requires an X.509 certificate.
 This certificate is provided in the :file:`samples/net/https_client/cert` folder.
+It is automatically converted to a hex file in :file:`CMakeLists.txt`.
+This hex file is included in :file:`main.c` where it is provisioned to the modem.
```
divipillai, line 40:
```suggestion
This hex file is included in :file:`main.c`, where it is provisioned to the modem.
```

divipillai (approved):

PR #13717, opened 2024-01-24
lib: modem_key_mgmt: Protect shared buffer with mutex

SeppoTakalo:
Add mutex to protect the scratch_buf so it would be thread safe.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -734,6 +734,10 @@ Modem libraries
   * Added the :kconfig:option:`CONFIG_AT_HOST_STACK_SIZE` Kconfig option.
     This option allows the stack size of the AT host workqueue thread to be adjusted.
 
+* :ref:`modem_key_mgmt` library:
+
+  * Fixed a potential race condition, where two threads might corrupt a shared response buffer.
+
 Libraries for networking
 ------------------------
 
--- a/lib/modem_key_mgmt/modem_key_mgmt.c
+++ b/lib/modem_key_mgmt/modem_key_mgmt.c
@@ -17,6 +17,8 @@
 
 LOG_MODULE_REGISTER(modem_key_mgmt, CONFIG_MODEM_KEY_MGMT_LOG_LEVEL);
 
+/* Protect the shared scratch_buf with a mutex. */
+static K_MUTEX_DEFINE(key_mgmt_mutex);
 static char scratch_buf[4096];
 
 static bool cmee_is_active(void)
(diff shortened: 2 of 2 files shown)


lib/modem_key_mgmt/modem_key_mgmt.c
```diff
@@ -17,6 +17,8 @@
 
 LOG_MODULE_REGISTER(modem_key_mgmt, CONFIG_MODEM_KEY_MGMT_LOG_LEVEL);
 
+/* Protect the shared scratch_buf with a mutex. */
+K_MUTEX_DEFINE(key_mgmt_mutex);
```
lemrey, line 21: This can be `static`, also, a very minor comment, I would use a binary semaphore here, since I don't expect this API to be reentrant.

lib/modem_key_mgmt/modem_key_mgmt.c
```diff
@@ -194,36 +202,46 @@ int modem_key_mgmt_cmp(nrf_sec_tag_t sec_tag,
...
+		goto out;
 	}
 
 	if (memcmp(begin, buf, len)) {
 		LOG_DBG("Credential data mismatch");
-		return 1;
+		err = 1;
+		goto out;
 	}
+	err = 0;
```
lemrey, line 239: This shouldn't be needed

lib/modem_key_mgmt/modem_key_mgmt.c
```diff
@@ -269,14 +289,17 @@ int modem_key_mgmt_exists(nrf_sec_tag_t sec_tag,
...
+		goto out;
 	}
 
 	if (strlen(scratch_buf) > strlen("OK\r\n")) {
 		*exists = true;
 	} else {
 		*exists = false;
 	}
-
+	err = 0;
```
lemrey, line 301: Neither should this

lemrey (commented):
A few minor comments. Do we need a changelog entry for this?

lib/modem_key_mgmt/modem_key_mgmt.c
SeppoTakalo: There doesn't seem to be much difference  in cost

With mutex:
```
FLASH:      291868 B
```

With semaphore:
```
FLASH:      291872 B
```

So I kept it as a semaphore as it was 4 bytes smaller :D

lib/modem_key_mgmt/modem_key_mgmt.c
SeppoTakalo: removed

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -734,6 +734,10 @@ Modem libraries
   * Added the :kconfig:option:`CONFIG_AT_HOST_STACK_SIZE` Kconfig option.
     This option allows the stack size of the AT host workqueue thread to be adjusted.
 
+* :ref:`modem_key_mgmt` library:
+
+  * Fixed a potential race conditions where two threads might corrupt a shared response buffer.
```
peknis, line 739:
```suggestion
  * Fixed a potential race condition, where two threads might corrupt a shared response buffer.
```

peknis (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
SeppoTakalo: Fixed.

peknis (approved):

lib/modem_key_mgmt/modem_key_mgmt.c
lemrey: That's fine.

lemrey (approved):

rlubos (approved):

PR #13722, opened 2024-01-24
net: lwm2m_client_utils: Disable RAI on certain events

SeppoTakalo:
## net: lwm2m_client_utils: Disable RAI on certain events

RAI indications should be disabled in case we are not on the registered state.
This might otherwise cause bootstrap sessions to fail on first try.

Lower the LOG level for RAI otions.

## samples: lwm2m-client: Change eDRX cycle to 5.12s
      
When we have shortest eDRX cycle, it is most responsive
so there would be less likely CoAP retransmissions.
      
This is good setting for development phase, as it is most
compatible with default settings of common servers.
      
For power saving, developer might choose to go higher,
but it may cause more CoAP packet losses.


SeppoTakalo:
I'll drop the doc-required label as I only changed one number from changelog.

juhaylinen (approved):

rlubos (approved):

PR #13729, opened 2024-01-24
nrf_security: Don't enable RSA code by default

SebastianBoe:
Follow good practice by not enabling features by default.

The RSA code is particularly large.

This patch reduces the code and read-only data from 113kB to 80kB in the sha256 sample using 53_ns.


mswarowsky (approved):
Not sure if this is need for something particular, but if not good call to remove it

vili-nordic (approved):

frkv (approved):
LGTM

tomi-font (approved):

MarkusLassila (approved):

SeppoTakalo (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -856,7 +856,8 @@ Libraries for NFC
 nRF Security
 ------------
 
-|no_changes_yet_note|
+* Stop enabling RSA keys by default. This will reduce code size for those that are not using RSA by 30kB. It will also break the configuration of those that use RSA keys and are not explicitly enabling an RSA key size. To fix this enable the key size that is required, e.g. :kconfig:option:`CONFIG_PSA_WANT_RSA_KEY_SIZE_2048`, if 2048 bit RSA keys are required.
```
mia-ko, line 859:
```suggestion
* Updated the library to no longer enable RSA keys by default.
  This reduces the code size by 30 kB for those that are not using RSA keys.
  This will also break the configuration for those using the RSA keys without explicitly enabling an RSA key size.
  Enable the required key size to fix the configuration, for example by setting the Kconfig option :kconfig:option:`CONFIG_PSA_WANT_RSA_KEY_SIZE_2048` if 2048-bit RSA keys are required.
```

mia-ko (changes requested):

mia-ko (approved):

PR #13733, opened 2024-01-24
tfm: __assert.h: Fix include

SebastianBoe:
__assert.h declares the macro

\#define __ASSERT_POST_ACTION() tfm_core_panic()

which references tfm_core_panic().

But __assert.h is not including utilities.h which is the header that provides tfm_core_panic.

Fix the include dependency.


vili-nordic (approved):

mswarowsky (approved):

frkv (commented):
LGTM

frkv (approved):

PR #13740, opened 2024-01-25
samples: wifi: Initialize wifi_connect_req_params structure

D-Triveni:
This fix addresses the issue where update to the channel value was not taking effect.

Fixes SHEL-2417.

### DIFF
--- a/samples/wifi/raw_tx_packet/src/wifi_connection.c
+++ b/samples/wifi/raw_tx_packet/src/wifi_connection.c
@@ -196,7 +196,7 @@ static int __wifi_args_to_params(struct wifi_connect_req_params *params)
 static int wifi_connect(void)
 {
 	struct net_if *iface = NULL;
-	static struct wifi_connect_req_params cnx_params;
+	static struct wifi_connect_req_params cnx_params = { 0 };
 
 	iface = net_if_get_first_wifi();
 	if (!iface) {
--- a/samples/wifi/sr_coex/src/main.c
+++ b/samples/wifi/sr_coex/src/main.c
@@ -224,7 +224,7 @@ static int __wifi_args_to_params(struct wifi_connect_req_params *params)
 static int wifi_connect(void)
 {
 	struct net_if *iface = net_if_get_default();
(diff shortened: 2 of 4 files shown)


krish2718 (commented):
The commit log should explain the root cause not just the issue description.

D-Triveni:
> The commit log should explain the root cause not just the issue description.

Updated.

chiranjeevi2776 (approved):

bama-nordic (approved):

muraliThokala (approved):

samples/wifi/raw_tx_packet/src/wifi_connection.c
```diff
@@ -196,7 +196,7 @@ static int __wifi_args_to_params(struct wifi_connect_req_params *params)
 static int wifi_connect(void)
 {
 	struct net_if *iface = NULL;
-	static struct wifi_connect_req_params cnx_params;
+	static struct wifi_connect_req_params cnx_params = { 0 };
```
krish2718, line 199: Why is `static` in use in the first place? I don't see any necessity.

krish2718 (changes requested):
Also, I don't understand how this can cause SHEL-2417, there the band and channel are being overwritten and not relying on initial values.

samples/wifi/raw_tx_packet/src/wifi_connection.c
krish2718: And just initialising zero doesn't mean that the values are sane, you need to have a `set_defaults` which you did in https://github.com/nrfconnect/sdk-nrf/pull/13715

D-Triveni: > Why is `static` in use in the first place? I don't see any necessity.

Will remove and raise new PR.

D-Triveni:
> Also, I don't understand how this can cause SHEL-2417, there the band and channel are being overwritten and not relying on initial values.

Closing this PR. Will do necessary changes as per the discussion on SHEL-2417

PR #13772, opened 2024-01-26
samples: cellular: modem_shell: Add sock getaddrinfo

stig-bjorlykke:
* Add a sock command to do getaddrinfo().
* Allow PDN CID 0 for connect().

### DIFF
--- a/samples/cellular/modem_shell/src/sock/sock.c
+++ b/samples/cellular/modem_shell/src/sock/sock.c
@@ -307,12 +307,11 @@ static int sock_protocol_resolve(
 	return proto;
 }
 
-static int sock_getaddrinfo(
-	struct sock_info *socket_info,
+static int sock_getaddrinfo_req(
+	struct addrinfo **result,
 	int family,
 	int type,
 	char *address,
-	int port,
 	int pdn_cid)
 {
 	int err;
@@ -336,7 +335,7 @@ static int sock_getaddrinfo(
 			hints.ai_flags = AI_PDNSERV;
 		}
 
-		err = getaddrinfo(address, service, &hints, &socket_info->addrinfo);
+		err = getaddrinfo(address, service, &hints, result);
 		if (err) {
 			if (err == DNS_EAI_SYSTEM) {
 				mosh_error("getaddrinfo() failed, err %d errno %d", err, errno);
@@ -344,17 +343,6 @@ static int sock_getaddrinfo(
 				mosh_error("getaddrinfo() failed, err %d", err);
 			}
 			return -EADDRNOTAVAIL;
-		}
-
-		/* Set port to address info */
-		if (family == AF_INET) {
-			((struct sockaddr_in *)socket_info->addrinfo->ai_addr)->sin_port =
-				htons(port);
-		} else if (family == AF_INET6) {
-			((struct sockaddr_in6 *)socket_info->addrinfo->ai_addr)->sin6_port =
-				htons(port);
-		} else {
-			assert(0);
 		}
 	}
 	return 0;
@@ -505,6 +493,29 @@ exit:
 	return err;
 }
 
+int sock_getaddrinfo(int family, int type, char *hostname, int pdn_cid)
+{
+	struct addrinfo *ai = NULL;
+	int err;
+
+	mosh_print("getaddrinfo query family=%d, type=%d, pdn_cid=%d, hostname=%s",
+		   family, type, pdn_cid, hostname);
+
+	err = sock_getaddrinfo_req(&ai, family, type, hostname, pdn_cid);
+	if (err) {
+		return err;
+	}
+
+	for (struct addrinfo *addr = ai; addr; addr = addr->ai_next) {
+		mosh_print("getaddrinfo answer family=%d, address=%s",
+			   addr->ai_family, net_utils_sckt_addr_ntop(addr->ai_addr));
+	}
+
+	freeaddrinfo(ai);
+
+	return err;
+}
+
 int sock_open_and_connect(
 	int family,
 	int type,
@@ -547,9 +558,18 @@ int sock_open_and_connect(
 	}
 
 	/* Get address */
-	err = sock_getaddrinfo(socket_info, family, type, address, port, pdn_cid);
+	err = sock_getaddrinfo_req(&socket_info->addrinfo, family, type, address, pdn_cid);
 	if (err) {
 		goto connect_error;
+	}
+
+	/* Set port to address info */
+	if (family == AF_INET) {
+		((struct sockaddr_in *)socket_info->addrinfo->ai_addr)->sin_port = htons(port);
+	} else if (family == AF_INET6) {
+		((struct sockaddr_in6 *)socket_info->addrinfo->ai_addr)->sin6_port = htons(port);
+	} else {
+		assert(0);
 	}
 
 	/* Create socket */
--- a/samples/cellular/modem_shell/src/sock/sock_shell.c
+++ b/samples/cellular/modem_shell/src/sock/sock_shell.c
@@ -30,6 +30,7 @@
 enum sock_shell_command {
 	SOCK_CMD_CONNECT = 0,
 	SOCK_CMD_CLOSE,
+	SOCK_CMD_GETADDRINFO,
 	SOCK_CMD_SEND,
 	SOCK_CMD_RECV,
 	SOCK_CMD_RAI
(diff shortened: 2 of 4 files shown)


samples/cellular/modem_shell/src/sock/sock.c
```diff
@@ -505,6 +505,36 @@ static int sock_bind_to_pdp_context(struct sock_info *socket_info, int pdn_cid)
 	return err;
 }
 
+int sock_getaddrinfo(int family, int type, char *address, int pdn_cid)
+{
+	struct addrinfo *addrinfo = NULL;
```
MirkoCovizzi, line 510:
```suggestion
	struct addrinfo *ai = NULL;
```

Less confusing.

stig-bjorlykke: ok

samples/cellular/modem_shell/src/sock/sock.c
```diff
@@ -505,6 +495,36 @@ static int sock_bind_to_pdp_context(struct sock_info *socket_info, int pdn_cid)
...
+		return err;
+	}
+
+	for (struct addrinfo *addr = ai; addr; addr = addr->ai_next) {
+		net_addr_ntop(addr->ai_family, &((struct sockaddr_in *)addr->ai_addr)->sin_addr,
+			      addr_str, sizeof(addr_str));
+		mosh_print("getaddrinfo answer family=%d, address=%s",
+			   addr->ai_family, addr_str);
+
+		if (addr->ai_canonname) {
```
tokangas, line 518: I tried the `-C` option, but did not get any canonical names. I checked the code and it seems that `nrf91_sockets.c` is missing implementation for copying the CNAME: https://github.com/nrfconnect/sdk-nrf/blob/f2d99e9c66e69d124eb051a7b3cef56a74b51fde/lib/nrf_modem_lib/nrf91_sockets.c#L308

I tried printing out the `ai_canonname` received from the modem, but it doesn't seem to work as expected. I checked from the modem log that the DNS query response contained a CNAME which was different from the requested hostname, but the response from the modem always contains the requested hostname. If the modem doesn't give the real CNAME, there's no point in fixing `nrf91_socket.c` and having the `-C` option here, either, I guess.

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -212,6 +231,104 @@ static void sock_print_usage(enum sock_shell_command command)
...
+				mosh_error(
+					"Unsupported address type=%s. Supported values are: "
+					"'any', 'stream' (tcp) or 'dgram' (udp)",
+					optarg);
+				return -EINVAL;
+			}
+			break;
+		case 'I': /* PDN CID */
+			arg_pdn_cid = atoi(optarg);
+			if (arg_pdn_cid <= 0) {
```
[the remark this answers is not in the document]
MirkoCovizzi: That needs to be addressed throughout the file, I see other occurrences for example in `cmd_sock_connect`.

samples/cellular/modem_shell/src/sock/sock.c
lemrey: Shall we create a task for this?

tokangas: I asked about this from the modem team and most likely this won't be changed for nRF91x1. If the `nrf91_socket.c` implementation would be completed, the canonical name would still be incorrect. Would it be better to leave the incorrect name out than provide it?

samples/cellular/modem_shell/src/sock/sock_shell.c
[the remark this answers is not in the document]
tokangas: @jhirsi Can you have a look? Isn't CID 0 the default PDP context?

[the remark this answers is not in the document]
stig-bjorlykke: CID 0 is the default PDP context, and providing `-I 0` should be the same as not provide a CID.

For testing purpose (and for completeness) it could be ok to allow this. I can change the existing check in connect and update this.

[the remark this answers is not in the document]
tokangas: Thanks, I agree that it would be good to allow also the default PDP context to be specified, even though the same functionality can be achieved without the parameter. `modem_shell` is basically a test application, so it should be possible to test this kind of scenarios. If the application is using more than one PDP context, I don't think it's far fetched to use CID 0 for traffic targeted to the default PDP context.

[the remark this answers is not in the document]
trantanen: Just note that atoi returns zero for invalid values so `-I hey` will be same as `-I 0`.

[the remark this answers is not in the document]
stig-bjorlykke: The same issue exists for many arguments in `sock_shell.c`, including socket_id.
This should be fixed in a separate PR.

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -212,6 +229,100 @@ static void sock_print_usage(enum sock_shell_command command)
...
+					"Unsupported address type=%s. Supported values are: "
+					"'any', 'stream' (tcp) or 'dgram' (udp)",
+					optarg);
+				return -EINVAL;
+			}
+			break;
+		case 'I': /* PDN CID */
+			arg_pdn_cid = atoi(optarg);
+			if (arg_pdn_cid < 0) {
+				mosh_error("PDN CID (%d) must be positive integer.", arg_pdn_cid);
```
tokangas, line 299: If you want to be precise, the error text should also be updated. Zero is not positive.

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -246,7 +357,7 @@ static int cmd_sock_connect(const struct shell *shell, size_t argc, char **argv)
 		switch (opt) {
 		case 'I': /* PDN CID */
 			arg_pdn_cid = atoi(optarg);
-			if (arg_pdn_cid <= 0) {
+			if (arg_pdn_cid < 0) {
 				mosh_error("PDN CID (%d) must be positive integer.", arg_pdn_cid);
```
tokangas, line 361: Same here.

tokangas (approved):
Looks good to me.

MirkoCovizzi (approved):

samples/cellular/modem_shell/src/sock/sock_shell.c
stig-bjorlykke: Actually I will change this check to be a range `0-10` according to the [AT+CGDCONT](https://docs.nordicsemi.com/bundle/ref_at_commands_nrf91x1/page/REF/at_commands/packet_domain/cgdcont_read.html) documentation.

trantanen: I wouldn't limit the number to 10 because this is a test tool where we can try to allow use of some invalid input in order to test modem behavior.
We don't check all validities due to this but I'm sure it's not consistent throughout the application.

stig-bjorlykke: If we want to be able to test any value I think we should remove the validation and accept anything into `arg_pdn_cid`.

trantanen: I think removing the validation is better than checking whether it's higher than 10.
I don't have strong opinion about checking negative values so I'm fine keeping it or removing it. There are few other places in MOSH that does some checks for CID but we can handle those later.

trantanen: As I said, I'd remove the check about being bigger than 10. And for negative values the check or no check is fine for me.

MirkoCovizzi: I think that something like this would be acceptable:
```suggestion
				mosh_error("PDN CID (%d) must be non-negative integer.", arg_pdn_cid);
```

stig-bjorlykke: ok 👍

trantanen (approved):

jhirsi (approved):

samples/cellular/modem_shell/src/sock/sock.c
```diff
@@ -547,11 +558,20 @@ int sock_open_and_connect(
...
 		goto connect_error;
 	}
 
+	/* Set port to address info */
+	if (family == AF_INET) {
+		((struct sockaddr_in *)socket_info->addrinfo->ai_addr)->sin_port = htons(port);
+	} else if (family == AF_INET6) {
+		((struct sockaddr_in6 *)socket_info->addrinfo->ai_addr)->sin6_port = htons(port);
+	} else {
+		assert(0);
```
trantanen, line 572: @stig-bjorlykke I need to remove this assert because raw sockets are hitting it always.

stig-bjorlykke: @trantanen I did not add this assert, I only moved the code. It's ok for me.

trantanen: Yep, moving the code doesn't work here for raw sockets that were treated separately in the previous location.
And while reviewing the code, Github collapses some important conditions from the previous location so it was tough to figure this out.
Fixing in PR #15723.

PR #13776, opened 2024-01-28
lib: nrf_modem: Copy ai_canonname from addrinfo

stig-bjorlykke:
Copy ai_canonname when converting between nrf_addrinfo and zsock_addrinfo.


lemrey (approved):
Looks good, consider adding one line to the changelog.

MirkoCovizzi:
I believe we should instead update the documentation https://github.com/nrfconnect/sdk-nrf/pull/13772#discussion_r1470678335

PR #13792, opened 2024-01-30
net: lwm2m_client_utils: Change resource data type

juhaylinen:
Change data type of last_change from time_t to int64_t.


SeppoTakalo:
Why?
If we store seconds, `time_t` should be used.

If Zephyr SDK is used, `time_t` should already be 64bit.

```c
// zephyr-sdk-0.16.4/arm-zephyr-eabi/arm-zephyr-eabi/sys-include/sys/types.h
#if !defined(__time_t_defined) && !defined(_TIME_T_DECLARED)
typedef	_TIME_T_	time_t;
#define	__time_t_defined
#define	_TIME_T_DECLARED
#endif
```

```c
// zephyr-sdk-0.16.4/arm-zephyr-eabi/arm-zephyr-eabi/sys-include/sys/_types.h
#if defined(_USE_LONG_TIME_T) || __LONG_MAX__ > 0x7fffffffL
#define	_TIME_T_ long
#else
#define	_TIME_T_ __int_least64_t
#endif
typedef	_TIME_T_	__time_t;
```

juhaylinen:
I'll close the PR. Coverity warning was false positive.

PR #13793, opened 2024-01-30
Samples: Bluetooth: Mesh: Migrate samples to new Sensor API

ludvigsj:
Sensor util improvements:

 * Access to sensor formats, needed to statically initialize
   `struct bt_mesh_sensor_value`.
 * Add back `bt_mesh_sensor_value_in_column`
 * Fix a couple of smaller bugs

Shell: Make it possible to compile the Sensor Client Shell with the new
API enabled.

Samples: Rewrite the samples to use the new Sensor APIs and disable the legacy
sensor value in the project config.

Signed-off-by: Ludvig Jordet <ludvig.jordet@nordicsemi.no>


doc/nrf/libraries/bluetooth_services/mesh/sensor_cli.rst
```diff
@@ -144,7 +144,7 @@ mesh models sensor series-entry get <SensorID> <Column>
    Get the value of a column for a sensor on the server.
 
    * ``SensorID`` - Selects the sensor for which to get the entry value.
-   * ``Column`` - Start value of the column for which to get the entry value.
+   * ``Column`` - For three channel sensors: Start value of the column for which to get the entry value. For other sensors: Index of the column for which to get the entry value.
```
mia-ko, line 147:
```suggestion
   * ``Column`` - For the three channel sensors: Start value of the column for which to get the entry value. For other sensors: Index of the column for which to get the entry value.
```

mia-ko (changes requested):
One tiny, tiny change suggestion.
And I was wondering if we need a changelog entry for this one.? :)

doc/nrf/libraries/bluetooth_services/mesh/sensor_cli.rst
ludvigsj: Really? "The three channel sensors" is not a defined set of anything so "the" feels out of place. I'd rather do "For sensors with three channels"

mia-ko: Aaaah, now I understand :) Three-channel sensors! I thought channel sensors as a type of sensor and 3 of them LOL.
It makes sense.
Either "For three-channel sensors" or "For sensors with three channels" then :)

PavelVPV:
Have you checked our sensors samples tests? Do they still pass?

ludvigsj:
> Have you checked our sensors samples tests? Do they still pass?

Oh I forgot we had one, I thought it was one of the ones we still tested manually. I'll run it and check after fixing the other test failure. Will DNM this in the meantime.

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -176,10 +195,9 @@ static void button_handler_cb(uint32_t pressed, uint32_t changed)
 		}
 	}
 	if (pressed & changed & BIT(1)) {
-		err = bt_mesh_sensor_cli_setting_set(
-			&sensor_cli, NULL, &bt_mesh_sensor_present_dev_op_temp,
-			&bt_mesh_sensor_dev_op_temp_range_spec,
-			&temp_ranges[temp_idx++][0], NULL);
+		err = setting_set_int(&bt_mesh_sensor_present_dev_op_temp,
+				      &bt_mesh_sensor_dev_power_range_spec,
```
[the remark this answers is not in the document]
ludvigsj: ... becuase IntelliSense took my typo and ran with it. I'll fix

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -15,22 +15,29 @@
...
+		enum bt_mesh_sensor_value_status status =
+			bt_mesh_sensor_value_to_micro(value, &time_since_presence_detected);
+
+		if (status == BT_MESH_SENSOR_VALUE_UNKNOWN) {
 			printk("Unknown last presence detected\n");
+		} else if (!time_since_presence_detected) {
+			printk("Presence detected, or under 1 second since presence detected\n");
 		} else {
 			printk("%s second(s) since last presence detected\n",
 			       bt_mesh_sensor_ch_str(value));
```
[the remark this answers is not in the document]
ludvigsj: No it prints seconds? The unit for the format is seconds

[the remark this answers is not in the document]
HaavardRei: Ahh ok, just got a little confused by the "_to_micro"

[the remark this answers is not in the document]
ludvigsj: Right, in case there still is some confusion, the `_to_micro` call is just to get a decoded integer out of it (`time_since_presence_detected`) so I can check for 0 on line 37. `value` stays as (encoded) seconds and is what is printed here.

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -66,9 +73,8 @@ static void sensor_cli_desc_cb(struct bt_mesh_sensor_cli *cli, struct bt_mesh_ms
 			       const struct bt_mesh_sensor_info *sensor)
 {
 	printk("Descriptor of sensor with ID 0x%04x:\n", sensor->id);
-	printk("\ttolerance: { positive: %s",
-	       bt_mesh_sensor_ch_str(&sensor->descriptor.tolerance.positive));
-	printk(" negative: %s }\n", bt_mesh_sensor_ch_str(&sensor->descriptor.tolerance.negative));
+	printk("\ttolerance: { positive: %d", sensor->descriptor.tolerance.positive);
+	printk(" negative: %d }\n", sensor->descriptor.tolerance.negative);
```
Balaklaka, line 77: Why is this not put into one printk, instead of 2?

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -579,16 +667,12 @@ static struct bt_mesh_sensor_srv chip_temp_sensor_srv =
 
 static struct k_work_delayable presence_detected_work;
 
+#define BOOLEAN_INIT(_bool) { .format = &bt_mesh_sensor_format_boolean, .raw = { !!(_bool) } }
```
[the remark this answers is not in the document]
ludvigsj: Yes. A C boolean can contain any number. The `raw` field can only legally contain 1 or 0. So it's there because the format spec is stricter than C is.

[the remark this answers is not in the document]
PavelVPV: > A C boolean can contain any number.

This I didn't get. Could you point me to the C99 standard where it would say this? You use `BOOLEAN_INIT` with `true` and `false` which are either 1 or 0, not any other number by standard. If you were thinking of using this macro with any number then, probably, fine (though the argument name is confusing), but from what I see how the macro is used in the sample `!!` is not needed.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -376,13 +455,19 @@ static void amb_light_level_gain_get(struct bt_mesh_sensor_srv *srv,
...
+
+	if (err) {
+		printk("Error encoding ambient light level gain (%d)\n", err);
+	} else {
+		printk("Ambient light level gain: %s\n", bt_mesh_sensor_ch_str(rsp));
+	}
 };
 
-static void amb_light_level_gain_store(double gain)
+static void amb_light_level_gain_store(float gain)
```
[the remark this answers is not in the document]
PavelVPV: image
Those that we support officially support FPU. Light Ctrl Spec Reg forces `CONFIG_FPU` option:
https://github.com/nrfconnect/sdk-nrf/blob/baba31095e92f9fae8f3b0ff1a58cd4932644843/subsys/bluetooth/mesh/Kconfig.models#L275-L277

samples/bluetooth/mesh/sensor_client/src/model_handler.c
ludvigsj: Good question. I just modified the existing lines above, but since we don't call `bt_mesh_sensor_ch_str` I can join them, true.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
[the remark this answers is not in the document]
ludvigsj: Right, I was confused, I thought the C99 boolean type still did the old thing where it just was an `int`. But you are correct, it will only ever be 1 or 0. I'll remove the `!!`.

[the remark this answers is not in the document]
PavelVPV: This is what I found (though I'm not an expert in the standard and may treat it incorrectly):
> 6.3.1.2 Boolean type
When any scalar value is converted to _Bool, the result is 0 if the value compares equal to 0; otherwise, the result is 1.

samples/bluetooth/mesh/sensor_client/src/model_handler.c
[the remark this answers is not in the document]
HaavardRei: Makes sense!

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -415,25 +505,34 @@ static int amb_light_level_ref_set(struct bt_mesh_sensor_srv *srv,
...
+		amb_light_level_gain_store(ref_float / dummy_ambient_light_value);
 	} else {
 		amb_light_level_gain_store(FLT_MAX);
 	}
 
-	sensor_value_from_double(&gain_value_tmp, amb_light_level_gain);
-
 	printk("Ambient light level ref(%s) ", bt_mesh_sensor_ch_str(value));
-	printk("gain(%s)\n", bt_mesh_sensor_ch_str(&gain_value_tmp));
+	printk("gain(%f)\n", amb_light_level_gain);
```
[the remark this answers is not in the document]
ludvigsj: It prints the correct value. I believe the CBPRINTF_FP_SUPPORT is no longer needed with picolibc

PavelVPV (approved):

HaavardRei (approved):

subsys/bluetooth/mesh/shell/shell_sensor_cli.c
```diff
@@ -387,7 +425,24 @@ static int cmd_series_entry_get(const struct shell *shell, size_t argc, char *ar
...
 	int err = 0;
 	uint32_t sensor_id = shell_strtoul(argv[1], 0, &err);
+	const struct bt_mesh_sensor_type *sensor_type = bt_mesh_sensor_type_get(sensor_id);
+
+	if (sensor_type == NULL) {
+		return -ENOENT;
+	}
+
+#ifdef CONFIG_BT_MESH_SENSOR_USE_LEGACY_SENSOR_VALUE
 	struct bt_mesh_sensor_column column = {.start = shell_model_strtosensorval(argv[2], &err)};
```
m-alperen-sener, line 435:
```suggestion
	struct bt_mesh_sensor_column column = {.start = shell_model_strtosensorval(NULL, argv[2], &err)};
```

m-alperen-sener (changes requested):

Balaklaka (changes requested):
I still wonder where the "Range" values in the comments for bt_mesh_sensor_format_... in sensor_types.h comes from. I looked in the "GATT Specification Supplement" and there a lot of these values has min/max values, and special values. Should we describe them in this file. Earlier this values was in a internal h file without comments, now these values are in a public h files that user are suppose to see, and then the comment should be correct. I just do not understand where these Range values come from?

include/bluetooth/mesh/sensor_types.h
```diff
@@ -44,6 +44,336 @@ extern "C" {
...
+ *  - Unit: Celsius
+ *  - Encoding: 8 bit signed scalar (Resolution: 0.5)
+ *  - Range: -64.0 to 63.5
+ */
+extern const struct bt_mesh_sensor_format bt_mesh_sensor_format_temp_8;
+
+/** Temp
+ *  - Unit: Celsius
+ *  - Encoding: 16 bit signed scalar (Resolution: 0.01)
+ *  - Range: -327.68 to 327.67
```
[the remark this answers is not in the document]
Balaklaka: Why is this change to -327.67?

samples/bluetooth/mesh/sensor_server/src/model_handler.c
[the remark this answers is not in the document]
ludvigsj: Heh, it prints the correct value on my desk, using SDK 16.0, but this seems to be a mistake and `CBPRINTF_FP_SUPPORT` is required to link the correct picolibc in SDK 16.5. I'll add this. This means I'll have to consider what to do inside `bt_mesh_sensor_ch_str` which prints a float for most formats, currently.

include/bluetooth/mesh/sensor_types.h
```diff
@@ -44,6 +44,336 @@ extern "C" {
...
 			&_type;                                                \
 	}
+
+/** @defgroup bt_mesh_sensor_formats_percentage Percentage sensor formats
+ *  @{
+ */
+
+/** Percentage 8
+ *  - Unit: Percent
+ *  - Encoding: 8 bit unsigned scalar (Resolution: 0.5)
```
mia-ko, line 54: 8/16/24/32 bits or 8-/16-/24-/32-bit
For all percentage sensor formats added in this file.

include/bluetooth/mesh/sensor_types.h
```diff
@@ -44,6 +44,336 @@ extern "C" {
...
+
+/** Pressure
+ *  - Unit: Pascal
+ *  - Encoding: 32 bit unsigned scalar (Resolution: 0.1)
+ *  - Range: 0 to 429496729.5
+ */
+extern const struct bt_mesh_sensor_format bt_mesh_sensor_format_pressure;
+
+/** Rainfall
+ *  - Unit: Metre
```
mia-ko, line 156: Meters
(US English)

include/bluetooth/mesh/sensor_types.h
```diff
@@ -44,6 +44,336 @@ extern "C" {
...
+
+/** Temp
+ *  - Unit: Celsius
+ *  - Encoding: 16 bit signed scalar (Resolution: 0.01)
+ *  - Range: -327.67 to 327.67
+ */
+extern const struct bt_mesh_sensor_format bt_mesh_sensor_format_temp;
+
+/** Co2 concentration
+ *  - Unit: Ppm
```
mia-ko, line 93: Do some of these units need an additional explanation?
E.g. 
Unit: Ppm (parts per million)

We also have ppb, Kvah, Va, Klmh, Klxh.

mia-ko (changes requested):

ludvigsj:
> I still wonder where the "Range" values in the comments for bt_mesh_sensor_format_... in sensor_types.h comes from. I looked in the "GATT Specification Supplement" and there a lot of these values has min/max values, and special values. Should we describe them in this file. Earlier this values was in a internal h file without comments, now these values are in a public h files that user are suppose to see, and then the comment should be correct. I just do not understand where these Range values come from?

They are computed as:

```math
\text{Range}_\text{min} = \begin{cases}
  \text{ MUL\_SCALAR}(\text{format\_min}) & \text{if HAS\_MIN} \\
  \text{MUL\_SCALAR}(\text{type\_min} + 1) & \text{if HAS\_UNDEFINED\_MIN} \\
  \text{MUL\_SCALAR}(\text{type\_min}) & \text{otherwise}
\end{cases}
```

```math
\text{Range}_\text{max} = \begin{cases}
  \text{ MUL\_SCALAR}(\text{format\_max}) & \text{if HAS\_MAX} \\
  \text{MUL\_SCALAR}(\text{type\_max} - 2) & \text{if HAS\_UNDEFINED and HAS\_INVALID} \\
  \text{MUL\_SCALAR}(\text{type\_max} - 1) & \text{if HAS\_UNDEFINED and } \neg \text{HAS\_INVALID} \\
  \text{MUL\_SCALAR}(\text{type\_max}) & \text{otherwise}
\end{cases}
```

based on the definitions in `sensor_types.c`. The idea was to have a first attempt at documenting this which can be improved in the future, so I took inspiration from the docs for the sensor types further down the file. I don't want to start going through everything and adding information about special values in this PR, that sounds like a separate improvement task.

include/bluetooth/mesh/sensor_types.h
ludvigsj: This is based on the `unit` in `sensor_types.c`, where this is line 80:

```
UNIT(metre) = { "Metre", "m" };
```

Should I change it there as well? There are several units there, all spelled in UK spelling.

include/bluetooth/mesh/sensor_types.h
ludvigsj: I did this to match the docs for the sensor types below. Suggest a separate PR to fix this across the file?

include/bluetooth/mesh/sensor_types.h
Balaklaka: If changed, they should be changed in both places. But maybe we could do that as a cleanup task of the values in sensor_types.c.

Balaklaka (approved):

include/bluetooth/mesh/sensor_types.h
mia-ko: I'm ok with this being fixed in a separate cleanup task.

include/bluetooth/mesh/sensor_types.h
mia-ko: I guess this will also be part of the cleanup task :)

ludvigsj:
Sample test is passing with changes in separate system test PR. Removing DNM.

m-alperen-sener (approved):

PR #13798, opened 2024-01-30
subsys: dfu: dfu_target: fix reset of uninitialized targets

kacperradoszewski:
* Unset the current DFU target on successful reset.
* Do not trigger a flash erase if stream reset is triggered on an unspecified flash device. This can happen after reset is done on a DFU target that failed to initialize properly.

### DIFF
--- a/subsys/dfu/dfu_target/src/dfu_target.c
+++ b/subsys/dfu/dfu_target/src/dfu_target.c
@@ -177,6 +177,8 @@ int dfu_target_reset(void)
 		return err;
 	}
 
+	current_target = NULL;
+
 	return 0;
 }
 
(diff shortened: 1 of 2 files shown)


kacperradoszewski:
The use case is if `dfu_target_mcuboot_init()` fails anywhere before `dfu_target_stream_init()`, for example because the file size is too big, then the context will not be initialised properly. Resetting the target would cause a crash because the flash device `stream_flash_ctx::fdex` remains NULL and so the flash driver ends up dereferencing it. It is also not possible to directly reinitialise the MCUBoot target, because it seems like reinitialising is currently only allowed for modem delta or SMP targets for some reason. 
Naturally this can be solved in various ways and at various levels, I simply wanted to avoid making intrusive API changes or adding NULL checks everywhere, so suggestions are welcome.

SeppoTakalo (approved):

lucasvdp (approved):

sigvartmh (approved):

PR #13804, opened 2024-01-31
Bluetooth: Controller: add vs tx timestamp support

henrla:
Add support for vendor specific hci command iso_read_tx_timestamp.

### DIFF
--- a/subsys/bluetooth/controller/hci_internal.c
+++ b/subsys/bluetooth/controller/hci_internal.c
@@ -663,6 +663,9 @@ static void vs_supported_commands(sdc_hci_vs_supported_vs_commands_t *cmds)
 #if defined(CONFIG_BT_CONN)
 	cmds->min_val_of_max_acl_tx_payload_set = 1;
 #endif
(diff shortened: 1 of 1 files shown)


Thalley:
Can I make a suggestion/enhancement?

The HCI version only returns a single timestamp+PSN when the command is called (and it is per Core 5.4 a bit unclear whether it is a scheduled transmitted SDU). 

Would it make sense to support requesting `N` number of SDUs and get N responses, that also indicates whether it has been transmitted, or just enqueue? If the controller has e.g. 5 ISO buffers, then the host may send 5 SDUs to the controller in a single SDU interval (with appropriate TS/PSN of course). It could be useful to be able to get information the last `N` transmitted or enqueued SDUs. This could also reduce the number of commands the host/application sends, if it can just do it once every 5 TX instead of each TX and get the same (or better) information.

rugeGerritsen (approved):

ivaniushkov (approved):

yuxCai (approved):

PR #13807, opened 2024-01-31
Add DT definitions for Wi-Fi device

rado17:
Fixes SHEL-875

### DIFF
--- a/boards/arm/nrf7002dk_nrf5340/nrf5340_cpuapp_common.dts
+++ b/boards/arm/nrf7002dk_nrf5340/nrf5340_cpuapp_common.dts
@@ -107,6 +107,11 @@
 		btrf-switch-gpios = <&gpio1 10 GPIO_ACTIVE_HIGH>;
 	};
 
+	nordic_wlan0: nordic_wlan0 {
+		compatible = "nordic,wlan0";
+		status = "okay";
+	};
+
 	/* These aliases are provided for compatibility with samples */
 	aliases {
 		led0 = &led0;
--- a/drivers/wifi/nrf700x/src/fmac_main.c
+++ b/drivers/wifi/nrf700x/src/fmac_main.c
@@ -38,6 +38,7 @@
 
 #endif /* !CONFIG_NRF700X_RADIO_TEST */
 
+#define DT_DRV_COMPAT nordic_wlan0
 LOG_MODULE_DECLARE(wifi_nrf, CONFIG_WIFI_NRF700X_LOG_LEVEL);
 
 struct nrf_wifi_drv_priv_zep rpu_drv_priv_zep;
@@ -872,8 +873,7 @@ static const struct zep_wpa_supp_dev_ops wpa_supp_ops = {
 
 
 #ifdef CONFIG_NET_L2_ETHERNET
-ETH_NET_DEVICE_INIT(wlan0, /* name - token */
-		    "wlan0", /* driver name - dev->name */
+ETH_NET_DEVICE_DT_INST_DEFINE(0,
 		    nrf_wifi_drv_main_zep, /* init_fn */
 		    NULL, /* pm_action_cb */
 		    &rpu_drv_priv_zep.rpu_ctx_zep.vif_ctx_zep[0], /* data */
@@ -886,8 +886,7 @@ ETH_NET_DEVICE_INIT(wlan0, /* name - token */
 		    &wifi_offload_ops, /* api */
 		    CONFIG_NRF_WIFI_IFACE_MTU); /*mtu */
 #else
-DEVICE_DEFINE(wlan0, /* name - token */
-	      "wlan0", /* driver name - dev->name */
+DEVICE_DT_INST_DEFINE(0,
 	      nrf_wifi_drv_main_zep, /* init_fn */
 	      NULL, /* pm_action_cb */
 #ifndef CONFIG_NRF700X_RADIO_TEST
@@ -902,5 +901,5 @@ DEVICE_DEFINE(wlan0, /* name - token */
 #endif /* CONFIG_NRF700X_STA_MODE */
 
 #ifdef CONFIG_L2_WIFI_CONNECTIVITY
-CONN_MGR_BIND_CONN(wlan0, L2_CONN_WLAN0);
+CONN_MGR_BIND_CONN(Z_DEVICE_DT_DEV_ID(DT_DRV_INST(0)), L2_CONN_WLAN0);
 #endif /* CONFIG_L2_WIFI_CONNECTIVITY */
--- a/modules/hostap/Kconfig
+++ b/modules/hostap/Kconfig
@@ -60,6 +60,10 @@ choice LIBC_IMPLEMENTATION
 	default NEWLIB_LIBC
 endchoice
 
+config WPA_SUPP_MGD_IFACES_PREFIXES
+	string "WPA supplicant managed interfaces prefixes"
+	default "wlan,esp,nordic_wlan"
+
 config WPA_SUPP_THREAD_STACK_SIZE
 	int "Stack size for wpa_supplicant thread"
 	default 8192
--- a/modules/hostap/src/supp_main.c
+++ b/modules/hostap/src/supp_main.c
@@ -306,9 +306,25 @@ static void iface_event_handler(struct net_mgmt_event_callback *cb,
 							uint32_t mgmt_event, struct net_if *iface)
 {
 	const char *ifname = iface->if_dev->dev->name;
+	char *prefix;
+	char mgd_if_prefix[] = CONFIG_WPA_SUPP_MGD_IFACES_PREFIXES;
+	bool found = false;
 
-	if (strncmp(ifname, IFACE_MATCHING_PREFIX, sizeof(IFACE_MATCHING_PREFIX) - 1) != 0) {
-		return;
+	if (sizeof(mgd_if_prefix) > 0) {
+		prefix = strtok(mgd_if_prefix, ",");
+
+		while (prefix) {
+			if (strncmp(ifname, prefix, strlen(prefix)) == 0) {
+				found = true;
+				break;
+			}
+			prefix = strtok(NULL, ",");
+		}
+
+		if (!found) {
+			wpa_printf(MSG_ERROR, "No matching prefix found");
+			return;
+		}
 	}
 
 	wpa_printf(MSG_DEBUG, "Event: %d", mgmt_event);
@@ -340,12 +356,19 @@ static void wait_for_interface_up(const char *iface_name)
 static void iface_cb(struct net_if *iface, void *user_data)
 {
 	const char *ifname = iface->if_dev->dev->name;
+	const struct device *wifi_dev = NULL;
+
+	wifi_dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_wifi));
+	if (!wifi_dev) {
+		wpa_printf(MSG_ERROR, "Device instance not found");
+		return;
+	}
 
 	if (ifname == NULL) {
 		return;
 	}
 
-	if (strncmp(ifname, DEFAULT_IFACE_NAME, strlen(ifname)) != 0) {
+	if (strncmp(ifname, wifi_dev->name, strlen(ifname)) != 0) {
 		return;
 	}
 
@@ -361,6 +384,13 @@ static void iface_cb(struct net_if *iface, void *user_data)
 static void z_wpas_iface_work_handler(struct k_work *item)
 {
 	ARG_UNUSED(item);
+	const struct device *wifi_dev = NULL;
+
+	wifi_dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_wifi));
+	if (!wifi_dev) {
+		wpa_printf(MSG_ERROR, "Device instance not found");
+		return;
+	}
 
 	int ret = k_sem_take(&z_wpas_ready_sem, K_SECONDS(5));
 
@@ -370,7 +400,7 @@ static void z_wpas_iface_work_handler(struct k_work *item)
 	}
 
 	net_if_foreach(iface_cb, NULL);
-	wait_for_interface_up(DEFAULT_IFACE_NAME);
+	wait_for_interface_up(wifi_dev->name);
 
 	k_sem_give(&z_wpas_ready_sem);
 }
@@ -475,6 +505,13 @@ static void z_wpas_start(void)
 {
 	struct wpa_params params;
 	int exitcode = -1;
+	const struct device *wifi_dev = NULL;
+
+	wifi_dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_wifi));
+	if (!wifi_dev) {
+		wpa_printf(MSG_ERROR, "Device instance not found");
+		return;
+	}
 
 #if !defined(CONFIG_WPA_SUPP_CRYPTO_NONE) && !defined(CONFIG_MBEDTLS_ENABLE_HEAP)
 	/* Needed for crypto operation as default is no-op and fails */
@@ -531,7 +568,7 @@ static void z_wpas_start(void)
 		exitcode = wpa_supplicant_run(global);
 	}
 
-	generate_supp_state_event(DEFAULT_IFACE_NAME, NET_EVENT_WPA_SUPP_CMD_NOT_READY, 0);
+	generate_supp_state_event(wifi_dev->name, NET_EVENT_WPA_SUPP_CMD_NOT_READY, 0);
 	eloop_unregister_read_sock(z_wpas_event_sockpair[0]);
 
 	z_global_wpa_ctrl_deinit();
--- a/modules/hostap/src/supp_main.c
+++ b/modules/hostap/src/supp_main.c
@@ -42,10 +42,6 @@ LOG_MODULE_REGISTER(wpa_supplicant, LOG_LEVEL_DBG);
 K_SEM_DEFINE(z_wpas_ready_sem, 0, 1);
 #include <l2_packet/l2_packet.h>
 
-/* Should match with the driver name */
-#define DEFAULT_IFACE_NAME "wlan0"
-#define IFACE_MATCHING_PREFIX "wlan"
-
 static struct net_mgmt_event_callback cb;
 struct k_mutex iface_up_mutex;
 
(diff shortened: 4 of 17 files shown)


drivers/wifi/nrf700x/src/fmac_main.c
```diff
@@ -38,10 +38,12 @@
 
 #endif /* !CONFIG_NRF700X_RADIO_TEST */
 
+#define DT_DRV_COMPAT nrf_wlan0
 LOG_MODULE_DECLARE(wifi_nrf, CONFIG_WIFI_NRF700X_LOG_LEVEL);
 
 struct nrf_wifi_drv_priv_zep rpu_drv_priv_zep;
 
+#define NRF_NODE(n) DT_NODELABEL(nrf_wlan##n)
```
sachinthegreen, line 46: `NRF_WLAN_NODE`  ?

drivers/wifi/nrf700x/src/fmac_main.c
```diff
@@ -801,10 +802,9 @@ ETH_NET_DEVICE_INIT(wlan0, /* name - token */
 #endif /* !CONFIG_NRF700X_STA_MODE */
 		    CONFIG_WIFI_INIT_PRIORITY, /* prio */
 		    &wifi_offload_ops, /* api */
-		    CONFIG_NRF_WIFI_IFACE_MTU); /*mtu */
+		    CONFIG_NRF_WIFI_IFACE_MTU); /*mtu */;
```
sachinthegreen, line 805: Why is this needed?

sachinthegreen (changes requested):

drivers/wifi/nrf700x/src/fmac_main.c
rado17: Done

drivers/wifi/nrf700x/src/net_if.c
```diff
@@ -577,8 +577,8 @@ int nrf_wifi_if_start_zep(const struct device *dev)
 	vif_info.state = NRF_WIFI_FMAC_IF_OP_STATE_UP;
 
 	memcpy(vif_info.ifacename,
-	       dev->name,
-	       strlen(dev->name));
+	       "wlan0",
```
krish2718, line 580: why is this now hardcoded?

boards/arm/nrf7002dk_nrf5340/nrf5340_cpuapp_common.dts
```diff
@@ -150,6 +151,10 @@ arduino_i2c: &i2c1 {
 	pinctrl-0 = <&uart0_default>;
 	pinctrl-1 = <&uart0_sleep>;
 	pinctrl-names = "default", "sleep";
+	nrf_wlan0: nrf_wlan0 {
```
krish2718, line 154: why is this under UART? Shouldn't this be a separate node? And the convention seems to be using `nordic`?

krish2718 (changes requested):

sachinthegreen (approved):

drivers/wifi/nrf700x/src/net_if.c
rado17: RPU identifies the dev name by wlan0.

jukkar: You can change the name of the network interface by calling
```
int net_if_set_name(struct net_if *iface, const char *buf);
```
if needed. Requires that `CONFIG_NET_INTERFACE_NAME` is enabled.
By default the first wifi interface is wlan0, but that can be changed if needed.

krish2718: > RPU identifies the dev name by wlan0.

This should be taken care in the FMAC as its specific to UMAC.

boards/arm/nrf7002dk_nrf5340/nrf5340_cpuapp_common.dts
rado17: Moved to the root node.. Vendor name convention is limited to compatible property [(ref).](https://docs.zephyrproject.org/2.7.5/guides/dts/intro.html#important-properties) .  `nrf_wlan0` should be fine for node label. But let me know if you want me to change this.

krish2718: Yeah, using both `nrf` and `nordic` seems a bit weird, but I don't have strong opinions.

drivers/wifi/nrf700x/src/net_if.c
rado17: Done

boards/arm/nrf7002dk_nrf5340/nrf5340_cpuapp_common.dts
rado17: Changed to `nordic` instead of using `nrf`

drivers/wifi/nrf700x/src/fmac_main.c
```diff
@@ -789,8 +791,7 @@ static const struct zep_wpa_supp_dev_ops wpa_supp_ops = {
 
 
 #ifdef CONFIG_NET_L2_ETHERNET
-ETH_NET_DEVICE_INIT(wlan0, /* name - token */
-		    "wlan0", /* driver name - dev->name */
+ETH_NET_DEVICE_DT_DEFINE(NRF_WLAN_NODE(0),
```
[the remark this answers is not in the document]
rado17: Done

drivers/wifi/nrf700x/src/fmac_main.c
```diff
@@ -803,8 +804,7 @@ ETH_NET_DEVICE_INIT(wlan0, /* name - token */
 		    &wifi_offload_ops, /* api */
 		    CONFIG_NRF_WIFI_IFACE_MTU); /*mtu */
 #else
-DEVICE_DEFINE(wlan0, /* name - token */
-	      "wlan0", /* driver name - dev->name */
+DEVICE_DT_DEFINE(NRF_WLAN_NODE(0),
```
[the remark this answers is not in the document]
rado17: Done

drivers/wifi/nrf700x/src/net_if.c
```diff
@@ -577,8 +577,8 @@ int nrf_wifi_if_start_zep(const struct device *dev)
 	vif_info.state = NRF_WIFI_FMAC_IF_OP_STATE_UP;
 
 	memcpy(vif_info.ifacename,
-	       dev->name,
-	       strlen(dev->name));
+		   dev->name,
+		   strlen(dev->name));
```
krish2718, line 581: unnecessary whitespace change

modules/hostap/src/supp_main.c
```diff
@@ -43,8 +43,8 @@ K_SEM_DEFINE(z_wpas_ready_sem, 0, 1);
 #include <l2_packet/l2_packet.h>
 
 /* Should match with the driver name */
-#define DEFAULT_IFACE_NAME "wlan0"
-#define IFACE_MATCHING_PREFIX "wlan"
+#define DEFAULT_IFACE_NAME "nordic_wlan0"
```
krish2718, line 46: How does the DTS entry map to iface name? I don't see anywhere we are setting the iface name actively?

drivers/wifi/nrf700x/src/net_if.c
rado17: Removed.

modules/hostap/src/supp_main.c
rado17: Created an entry in chosen node for `zephyr,wifi` which can be referred for fetching the device info.

modules/hostap/src/supp_main.c
```diff
@@ -306,8 +302,15 @@ static void iface_event_handler(struct net_mgmt_event_callback *cb,
...
+	const struct device *wifi_dev = NULL;
+
+	wifi_dev =  DEVICE_DT_GET(DT_CHOSEN(zephyr_wifi));
+	if (!wifi_dev) {
+		wpa_printf(MSG_ERROR, "Device instance not found");
+		return;
+	}
 
-	if (strncmp(ifname, IFACE_MATCHING_PREFIX, sizeof(IFACE_MATCHING_PREFIX) - 1) != 0) {
+	if (strncmp(ifname, wifi_dev->name, sizeof(wifi_dev->name) - 1) != 0) {
```
krish2718, line 313: This won't work, as it needs to match multiple interfaces.

rado17: Introduced a Kconfig option for prefix string.

modules/hostap/src/supp_main.c
```diff
@@ -307,7 +303,8 @@ static void iface_event_handler(struct net_mgmt_event_callback *cb,
 {
 	const char *ifname = iface->if_dev->dev->name;
 
-	if (strncmp(ifname, IFACE_MATCHING_PREFIX, sizeof(IFACE_MATCHING_PREFIX) - 1) != 0) {
+	if (strncmp(ifname, CONFIG_NRF700X_WLAN_IF_PREFIX,
+		    sizeof(CONFIG_NRF700X_WLAN_IF_PREFIX)) != 0) {
```
jukkar, line 307: This is big no-no, you are now introducing NRF specific checks into generic code. Beside zephyr upstream is already using interface name and not the device name to match the needed interface. It would be better if the downstream code would use the interface name too. If one knows network interface name, it is always possible to figure out what is the device, but it is not always possible to do the reverse (if VLAN is enabled).

jukkar (changes requested):

modules/hostap/src/supp_main.c
rado17: Moved the prefix definition to WPA supplicant module.

modules/hostap/src/supp_main.c
```diff
@@ -42,10 +42,6 @@ LOG_MODULE_REGISTER(wpa_supplicant, LOG_LEVEL_DBG);
 K_SEM_DEFINE(z_wpas_ready_sem, 0, 1);
 #include <l2_packet/l2_packet.h>
 
-/* Should match with the driver name */
-#define DEFAULT_IFACE_NAME "wlan0"
```
krish2718, line 46: Please put this changes in a separate commit, a commit shouldn't span multiple modules, makes is tough to cherry pick or revert.

drivers/wifi/nrf700x/src/net_if.c
```diff
@@ -606,10 +606,7 @@ int nrf_wifi_if_start_zep(const struct device *dev)
 	       sizeof(vif_info));
 
 	vif_info.state = NRF_WIFI_FMAC_IF_OP_STATE_UP;
-
-	memcpy(vif_info.ifacename,
```
krish2718, line 610: these changes are unrelated to commit log

drivers/wifi/nrf700x/src/net_if.c
```diff
@@ -715,10 +712,7 @@ int nrf_wifi_if_stop_zep(const struct device *dev)
 	       sizeof(vif_info));
 
 	vif_info.state = NRF_WIFI_FMAC_IF_OP_STATE_DOWN;
-
-	memcpy(vif_info.ifacename,
-	       dev->name,
-	       strlen(dev->name));
+	vif_info.if_index = vif_ctx_zep->vif_idx;
```
krish2718, line 715: these changes are unrelated to commit log

drivers/wifi/nrf700x/src/wpa_supp_if.c
```diff
@@ -1921,9 +1921,7 @@ static int nrf_wifi_vif_state_change(struct nrf_wifi_vif_ctx_zep *vif_ctx_zep,
 	vif_ctx = def_dev_ctx->vif_ctx[vif_ctx_zep->vif_idx];
 
 	vif_state_info.state = state;
-	memcpy(vif_state_info.ifacename,
-	       vif_ctx_zep->ifname,
-	       strlen(vif_ctx_zep->ifname));
+	vif_state_info.if_index = vif_ctx_zep->vif_idx;
```
krish2718, line 1924: these changes are unrelated to commit log

krish2718 (changes requested):

modules/hostap/src/supp_main.c
rado17: Added a separate commit

drivers/wifi/nrf700x/src/net_if.c
rado17: Included in a separate commit

drivers/wifi/nrf700x/src/wpa_supp_if.c
rado17: Included in a separate commit

drivers/wifi/nrf700x/src/net_if.c
rado17: Included in a separate commit

modules/hostap/src/supp_main.c
```diff
@@ -340,12 +337,19 @@ static void wait_for_interface_up(const char *iface_name)
...
+		wpa_printf(MSG_ERROR, "Device instance not found");
+		return;
+	}
 
 	if (ifname == NULL) {
 		return;
 	}
 
-	if (strncmp(ifname, DEFAULT_IFACE_NAME, strlen(ifname)) != 0) {
+	if (strncmp(ifname, wifi_dev->name, strlen(ifname)) != 0) {
```
krish2718, line 352: Even though its seldom used, we need a multi-interface solution for WPA supplicant, esp. as this is all upstream code.
And as changing interface manager isn't a trivial task, IMHO, we should stick a Kconfig of allowed prefixes and then use that to select managed interfaces in CSV form, something like below. WDYT?

```
config WPA_SUPP_MGD_IFACES_PREFIX
       string "WPA supplicant managed interfaces prefix"
       default "wlan"

CONFIG_WPA_SUPP_MGD_IFACES_PREFIX="wlan,esp,nrf"
```

rado17: >CONFIG_WPA_SUPP_MGD_IFACES_PREFIX="wlan,esp,nrf"

Board Kconfig seems to be a good place to have this configuration. Would you agree?

modules/hostap/src/supp_main.c
jukkar: Eventually the code in supp_main.c will be moved to upstream so we cannot have here NRF specific Kconfig symbols. So this cannot be done like this. Also wifi driver specific changes need to be separated to different commit than the hostap related ones.

jukkar (changes requested):

modules/hostap/src/supp_main.c
rado17: Done. Instead of NRF specific config name, using a WPA specific config. Separated the commits as well.

modules/hostap/src/supp_main.c
rado17: Done.

modules/hostap/src/supp_main.c
```diff
@@ -306,9 +306,32 @@ static void iface_event_handler(struct net_mgmt_event_callback *cb,
 							uint32_t mgmt_event, struct net_if *iface)
 {
 	const char *ifname = iface->if_dev->dev->name;
+	char *tmp_csv;
+	char *prefix;
+	char *mgd_if_prefix = CONFIG_WPA_SUPP_MGD_IFACES_PREFIX;
```
jukkar, line 307: We can do many things in built time with following changes.
Instead of using a pointer, we could do this (as the string is small the impact to stack is negligible)
```
char mgd_if_prefix[] = CONFIG_WPA_SUPP_MGD_IFACES_PREFIX;
```

modules/hostap/src/supp_main.c
```diff
@@ -306,9 +306,32 @@ static void iface_event_handler(struct net_mgmt_event_callback *cb,
...
+	bool found = false;
+
+	if (strlen(mgd_if_prefix) > 0) {
```
jukkar, line 310: and here we can use sizeof
```
if (sizeof(mgd_if_prefix) > 0) {
```

modules/hostap/src/supp_main.c
```diff
@@ -306,9 +306,32 @@ static void iface_event_handler(struct net_mgmt_event_callback *cb,
...
+		tmp_csv = os_zalloc(strlen(mgd_if_prefix));
```
jukkar, line 311: We do not need this heap allocation anymore, as we can manipulate the `mgd_if_prefix` variable as we want.

modules/hostap/src/supp_main.c
```diff
@@ -306,9 +306,32 @@ static void iface_event_handler(struct net_mgmt_event_callback *cb,
...
+			wpa_printf(MSG_ERROR, "Failed to allocate memory for prefix string");
+			return;
+		}
+		strcpy(tmp_csv, mgd_if_prefix);
+		prefix = strtok(tmp_csv, ",");
 
-	if (strncmp(ifname, IFACE_MATCHING_PREFIX, sizeof(IFACE_MATCHING_PREFIX) - 1) != 0) {
-		return;
+		while (prefix) {
+			if (strncmp(ifname, prefix, strlen(prefix)) == 0) {
```
jukkar, line 320: We can either write '\0` where the `,` is or use proper len (as we know the start and stop of the prefix) when doing the compare.

modules/hostap/Kconfig
```diff
@@ -60,6 +60,10 @@ choice LIBC_IMPLEMENTATION
 	default NEWLIB_LIBC
 endchoice
 
+config WPA_SUPP_MGD_IFACES_PREFIX
```
krish2718, line 63: `IFACE_PREFIXES`

modules/hostap/Kconfig
```diff
@@ -60,6 +60,10 @@ choice LIBC_IMPLEMENTATION
...
+	string "WPA supplicant managed interfaces prefix"
```
krish2718, line 64:
```suggestion
	string "WPA supplicant managed interface prefixes"
```

modules/hostap/src/supp_main.c
rado17: Will use `strtok` which returns a null-terminated string.

modules/hostap/Kconfig
rado17: Done

modules/hostap/Kconfig
rado17: Done

modules/hostap/src/supp_main.c
rado17: Done

modules/hostap/src/supp_main.c
rado17: Done.

modules/hostap/src/supp_main.c
rado17: Done

VivekUppunda (approved):

krish2718 (approved):

PR #13812, opened 2024-01-31
samples: net: Reduce TFM RAM and Flash usage and support ns build

balaji-nordic:
Changed TFM profile type from medium to small.
Reduced SRAM and Partition size of TFM to minimal required.
Change net samples to build for non-secure target.

Testing checklist.
- [x] download client
- [x] http_client
- [x] http_server
- [x] aws_iot
- [x] azure_iot
- [x] udp
- [x] mqtt
- [x] coap_client


balaji-nordic:
@frkv  Can you confirm if there are any short comings of using small TFM profile instead of medium? Most importantly we want to ensure that PSA storage continues to work.

gregersrygg (approved):

maxd-nordic:
> @frkv Can you confirm if there are any short comings of using small TFM profile instead of medium? Most importantly we want to ensure that PSA storage continues to work.

"TF-M Profile Small assumes that extremely constrained devices only contain basic
on-chip storage, without external or removable storage.
As a result, TF-M Profile Small includes ITS service and **disables Protected
Storage service.**"

frkv:
> "TF-M Profile Small assumes that extremely constrained devices only contain basic
> on-chip storage, without external or removable storage.
> As a result, TF-M Profile Small includes ITS service and **disables Protected Storage service.**"

I will be making a case for using CONFIG_TFM_PROFILE_TYPE_NOT_SET and handling crypto and security features in a way where you can combine this with other requirements.

The Small profile disables Protected Storage as you say... Enabling Protected Storage as an additional configuration makes this a customization of the profile, not in sync with the documentation.

CONFIG_TFM_PROFILE_TYPE_NOT_SET is a better choice, as it lets you enable and disable features as you will, including in some cases being able to do further optimizations

simensrostad:
Similar optimizations for MSS by @glarsennordic https://github.com/nrfconnect/sdk-nrf/pull/13841

Its tempting to make something general for all wifi samples since we should use PSA storage for wifi and tls creds for all relevant samples.

samples/net/azure_iot_hub/boards/nrf7002dk_nrf5340_cpuapp_ns.conf
```diff
@@ -0,0 +1,99 @@
...
+CONFIG_DOWNLOAD_CLIENT_STACK_SIZE=4096
+
+# Enable external flash to host MCUBoot secondary partition
+CONFIG_SPI=y
+CONFIG_SPI_NOR=y
+CONFIG_SPI_NOR_SFDP_DEVICETREE=y
+CONFIG_PM_OVERRIDE_EXTERNAL_DRIVER_CHECK=y
+
+# TFM
+CONFIG_TFM_PROFILE_TYPE_SMALL=y
```
hellesvik-nordic, line 97: The nRF Connect SDK only supports CONFIG_TFM_PROFILE_TYPE_NOT_SET and CONFIG_TFM_PROFILE_TYPE_MINIMAL.
See https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/security/tfm.html#minimal-build

balaji-nordic: As discussed offline, using `CONFIG_TFM_PROFILE_TYPE_NOT_SET` and `CONFIG_TFM_PROFILE_TYPE_MINIMAL` will result in several KConfig options in the sample. Due to this we prefer to use `CONFIG_TFM_PROFILE_TYPE_SMALL` and be happy with the configurations it provides for now and fine tune this later (if need be).

gregersrygg (approved):
The release notes should mention this change.

ioannisg:
Thanks for this work @balaji-nordic , some generic comments/thoughts/questions from my side
- Switching to TF-M small for Non-Secure builds on nRF7002 DK (5340 +7002) seems to be generic for all networking samples, and so are the respective settings (TF-M RAM and partition size); would it make sense to move this configuration in some common place, and it is somehow inherited into the sample project configuration? So any future adjustments to the TF-M profile, image size and RAM footprint won't need to be duplicated in any desired sample
- Are the Wi-Fi stack savings generic enough to be also considered as common practice for all samples? For example I see the RX buffer count and TX aggregation settings to be similar across many examples, so I wonder if the same principle could apply here (i.e. moving these settings to a common place for, at least, the Wi-Fi Networking examples)
- You changed the Networking samples, to build for non-secure nRF5340 (nRF7002 DK) targets, which is legitimate. I wonder, however, whether we should also keep the original (secure-only) configurations and overlays as well, for developers that do not require TF-M on their Wi-Fi networking applications. Several RAM reduction optimizations would also make sense for secure-only builds. Probably OK to use NS builds for integration testing, but the .conf and .overlay files could also be present for secure-only builds, thoughts?

I would also appreciate if the Wi-Fi team can add some thoughts and reviews here @sachinthegreen @krish2718

ioannisg:
> The release notes should mention this change.

Agreed. Also, If we are making the TF-M Small profile the default profile for Wi-Fi Networking samples Non-Secure builds, could we think of adding a corresponding documentation update somewhere in the NCS docs?

frkv (approved):
It has been discussed offline that we need to come up with Kconfig patterns (read: profiles) that is tied to the specific modules in use, so we can achieve the following:

1. Don't need to dictate "small" profile (as it explicitly states PSA protected storage is not enabled)
2. Improve the ability to make copy of the samples with or without additions on crypto and additional features.

These improvements will not be addressed in this PR, which is mainly about optimizing for size for _ns images as well as better TF-M support for samples interacting with cloud

LGTM as long as all tests pass

simensrostad (approved):

divipillai:
> > The release notes should mention this change.
> 
> Agreed. Also, If we are making the TF-M Small profile the default profile for Wi-Fi Networking samples Non-Secure builds, could we think of adding a corresponding documentation update somewhere in the NCS docs?

We now have the requirements table updated with _ns build target for the networking samples that builds with the nRF7002 DK for example - https://developer.nordicsemi.com/nRF_Connect_SDK_dev/doc/PR-13812/nrf/samples/net/aws_iot/README.html#requirements (Doc preview of the PR). Hence, this would show that TF-M is the default profile for Wi-Fi networking samples non-Secure builds.

balaji-nordic:
@ioannisg 

> * Switching to TF-M small for Non-Secure builds on nRF7002 DK (5340 +7002) seems to be generic for all networking samples, and so are the respective settings (TF-M RAM and partition size); would it make sense to move this configuration in some common place, and it is somehow inherited into the sample project configuration? So any future adjustments to the TF-M profile, image size and RAM footprint won't need to be duplicated in any desired sample

That would definitely be worth taking a look in the near future. Will consider along with other initiatives. (CIA-1170 created for this)

> * Are the Wi-Fi stack savings generic enough to be also considered as common practice for all samples? For example I see the RX buffer count and TX aggregation settings to be similar across many examples, so I wonder if the same principle could apply here (i.e. moving these settings to a common place for, at least, the Wi-Fi Networking examples)

We did consider a snippet for this purpose. But ideally it would be better if we get a list of optimization settings for the wifi stack that, for example, gives different profiles for size optimizations, speed optimizations, and no optimizations (more like `CONFIG_NO_OPTIMIZATIONS` or TFM profile types. (CIA-1171 created for this)

> * You changed the Networking samples, to build for non-secure nRF5340 (nRF7002 DK) targets, which is legitimate. I wonder, however, whether we should also keep the original (secure-only) configurations and overlays as well, for developers that do not require TF-M on their Wi-Fi networking applications. Several RAM reduction optimizations would also make sense for secure-only builds. Probably OK to use NS builds for integration testing, but the .conf and .overlay files could also be present for secure-only builds, thoughts?

This was a conscious decision. We wanted customers to follow best practices and we consider building for non-secure by default is one of them. As a positive side effect, this also reduces maintenance effort for us.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -568,6 +568,8 @@ Multicore samples
 Networking samples
 ------------------
 
+* Updated the samples to build for non-secure target when building for the nRF7002DK. Used small profile type for Trusted Firmware-M (TFM) and optimized its memory foot print.
```
divipillai, line 571:
```suggestion
* Updated the networking samples to build using a non-secure build target when building for the nRF7002DK. 
  A small profile type for Trusted Firmware-M (TF-M) is used to optimize its memory footprint.
```
This needs to be after the line no: 574 (After added).

divipillai (approved):

ioannisg:
> @ioannisg
> 
> > * Switching to TF-M small for Non-Secure builds on nRF7002 DK (5340 +7002) seems to be generic for all networking samples, and so are the respective settings (TF-M RAM and partition size); would it make sense to move this configuration in some common place, and it is somehow inherited into the sample project configuration? So any future adjustments to the TF-M profile, image size and RAM footprint won't need to be duplicated in any desired sample
> 
> That would definitely be worth taking a look in the near future. Will consider along with other initiatives. (CIA-1170 created for this)
> 
> > * Are the Wi-Fi stack savings generic enough to be also considered as common practice for all samples? For example I see the RX buffer count and TX aggregation settings to be similar across many examples, so I wonder if the same principle could apply here (i.e. moving these settings to a common place for, at least, the Wi-Fi Networking examples)
> 
> We did consider a snippet for this purpose. But ideally it would be better if we get a list of optimization settings for the wifi stack that, for example, gives different profiles for size optimizations, speed optimizations, and no optimizations (more like `CONFIG_NO_OPTIMIZATIONS` or TFM profile types. (CIA-1171 created for this)
> 
> > * You changed the Networking samples, to build for non-secure nRF5340 (nRF7002 DK) targets, which is legitimate. I wonder, however, whether we should also keep the original (secure-only) configurations and overlays as well, for developers that do not require TF-M on their Wi-Fi networking applications. Several RAM reduction optimizations would also make sense for secure-only builds. Probably OK to use NS builds for integration testing, but the .conf and .overlay files could also be present for secure-only builds, thoughts?
> 
> This was a conscious decision. We wanted customers to follow best practices and we consider building for non-secure by default is one of them. As a positive side effect, this also reduces maintenance effort for us.

@balaji-nordic thanks for reporting CIA-1117, I seriously think we need a centralized, generic approach, where we align the memory optimizations for Wi-Fi IoT applications with the proposed guidelines (see Usage Profiles in https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/protocols/wifi/stack_configuration.html#usage-profiles).

For NS-builds, I agree with the decision to move to having NS-builds as default. I am saying it would be nice to somehow assist developers in getting a secure-only build with the same optimizations for memory footprint. What would be the easiest path for users to profile the examples in terms of footprint for secure-only builds? Duplicate the .conf or .overlay file for secure build? Would it make sense to have that configuration available for easy pick-up?

balaji-nordic:
> > @ioannisg
> > > * Switching to TF-M small for Non-Secure builds on nRF7002 DK (5340 +7002) seems to be generic for all networking samples, and so are the respective settings (TF-M RAM and partition size); would it make sense to move this configuration in some common place, and it is somehow inherited into the sample project configuration? So any future adjustments to the TF-M profile, image size and RAM footprint won't need to be duplicated in any desired sample
> > 
> > 
> > That would definitely be worth taking a look in the near future. Will consider along with other initiatives. (CIA-1170 created for this)
> > > * Are the Wi-Fi stack savings generic enough to be also considered as common practice for all samples? For example I see the RX buffer count and TX aggregation settings to be similar across many examples, so I wonder if the same principle could apply here (i.e. moving these settings to a common place for, at least, the Wi-Fi Networking examples)
> > 
> > 
> > We did consider a snippet for this purpose. But ideally it would be better if we get a list of optimization settings for the wifi stack that, for example, gives different profiles for size optimizations, speed optimizations, and no optimizations (more like `CONFIG_NO_OPTIMIZATIONS` or TFM profile types. (CIA-1171 created for this)
> > > * You changed the Networking samples, to build for non-secure nRF5340 (nRF7002 DK) targets, which is legitimate. I wonder, however, whether we should also keep the original (secure-only) configurations and overlays as well, for developers that do not require TF-M on their Wi-Fi networking applications. Several RAM reduction optimizations would also make sense for secure-only builds. Probably OK to use NS builds for integration testing, but the .conf and .overlay files could also be present for secure-only builds, thoughts?
> > 
> > 
> > This was a conscious decision. We wanted customers to follow best practices and we consider building for non-secure by default is one of them. As a positive side effect, this also reduces maintenance effort for us.
> 
> @balaji-nordic thanks for reporting CIA-1117, I seriously think we need a centralized, generic approach, where we align the memory optimizations for Wi-Fi IoT applications with the proposed guidelines (see Usage Profiles in https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/protocols/wifi/stack_configuration.html#usage-profiles).
> 
> For NS-builds, I agree with the decision to move to having NS-builds as default. I am saying it would be nice to somehow assist developers in getting a secure-only build with the same optimizations for memory footprint. What would be the easiest path for users to profile the examples in terms of footprint for secure-only builds? Duplicate the .conf or .overlay file for secure build? Would it make sense to have that configuration available for easy pick-up?

Discussed offline. This will be considered for future improvements.

PR #13815, opened 2024-01-31
doc: Zigbee: orphan ZCL clusters page

wiba-nordic:
The Adding ZCL clusters to application page in the Zigbee documentation has been outdated for some while now. This orphans the page so it isn't visible in the table of contents until the page can be updated.


greg-fer:
@bpienk , we had a conversation about updating this page last year, but then things happen. You told me back then the Zigbee team will look into this and update the page. Is this still the plan? If not, I agree to @wiba-nordic 's proposal of orphaning the page.

greg-fer (approved):
No answer from @bpienk , but I'm approving.

milewr (approved):

PR #13817, opened 2024-01-31
applications: Matter Bridge: fixed handling of the provider object

markaj-nordic:
* Fixed ownership violation of the provider object
* Fixed races occurring when the provider cannot be added to the
  BridgeManager
* Secured the edge cases when the BLE discovery fails
* Enhanced InitializeBridgedDevice API and its documentation

### DIFF
--- a/applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
+++ b/applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
@@ -124,6 +124,7 @@ CHIP_ERROR AddMatterDevices(MatterBridgedDevice::DeviceType deviceTypes[], uint8
 			    uint16_t endpointIds[] = nullptr)
 {
 	VerifyOrReturnError(provider != nullptr, CHIP_ERROR_INVALID_ARGUMENT, LOG_ERR("No valid data provider!"));
+	chip::Platform::UniquePtr<BridgedDeviceDataProvider> providerPtr(provider);
 	VerifyOrReturnError(count <= BridgeManager::kMaxBridgedDevicesPerProvider, CHIP_ERROR_BUFFER_TOO_SMALL,
 			    LOG_ERR("Trying to add too many endpoints for single provider device."));
 
@@ -152,20 +153,24 @@ CHIP_ERROR AddMatterDevices(MatterBridgedDevice::DeviceType deviceTypes[], uint8
 	}
 
 	if (forceIndexesAndEndpoints) {
-		err = BridgeManager::Instance().AddBridgedDevices(newBridgedDevices, provider, count, deviceIndexes,
-								  endpointIds);
+		err = BridgeManager::Instance().AddBridgedDevices(newBridgedDevices, providerPtr.get(), count,
+								  deviceIndexes, endpointIds);
 	} else {
-		err = BridgeManager::Instance().AddBridgedDevices(newBridgedDevices, provider, count, deviceIndexes);
+		err = BridgeManager::Instance().AddBridgedDevices(newBridgedDevices, providerPtr.get(), count,
+								  deviceIndexes);
 	}
 
 	if (err == CHIP_NO_ERROR) {
 		for (uint8_t i = 0; i < count; i++) {
-			err = StoreDevice(newBridgedDevices[i], provider, deviceIndexes[i]);
+			err = StoreDevice(newBridgedDevices[i], providerPtr.get(), deviceIndexes[i]);
 			if (err != CHIP_NO_ERROR) {
 				break;
 			}
 		}
 	}
+
+	/* The ownership was transferred unconditionally to the BridgeManager, release the pointer. */
+	providerPtr.release();
 
 	if (err != CHIP_NO_ERROR) {
 		LOG_ERR("Adding Matter bridged device failed: %s", ErrorStr(err));
@@ -186,36 +191,29 @@ struct BluetoothConnectionContext {
 	bt_addr_le_t address;
 };
 
-void BluetoothDeviceConnected(bool discoverySucceeded, void *context)
+CHIP_ERROR BluetoothDeviceConnected(bool success, void *context)
 {
 	if (!context) {
-		return;
+		return CHIP_ERROR_INVALID_ARGUMENT;
 	}
 
 	BluetoothConnectionContext *ctx = reinterpret_cast<BluetoothConnectionContext *>(context);
 
-	VerifyOrExit(discoverySucceeded, );
+	if (!success) {
+		chip::Platform::Delete(ctx->provider);
+		chip::Platform::Delete(ctx);
+		return CHIP_ERROR_INTERNAL;
+	}
 
-	/* Schedule adding device to main thread to avoid overflowing the BT stack. */
-	VerifyOrExit(chip::DeviceLayer::PlatformMgr().ScheduleWork(
-			     [](intptr_t context) {
-				     BluetoothConnectionContext *ctx =
-					     reinterpret_cast<BluetoothConnectionContext *>(context);
-				     chip::Optional<uint8_t> indexes[BridgeManager::kMaxBridgedDevicesPerProvider];
-				     chip::Optional<uint16_t> endpointIds[BridgeManager::kMaxBridgedDevicesPerProvider];
-				     if (CHIP_NO_ERROR != AddMatterDevices(ctx->deviceTypes, ctx->count, ctx->nodeLabel,
-									   ctx->provider)) {
-					     BLEConnectivityManager::Instance().RemoveBLEProvider(ctx->address);
-					     chip::Platform::Delete(ctx->provider);
-				     }
-				     chip::Platform::Delete(ctx);
-			     },
-			     reinterpret_cast<intptr_t>(ctx)) == CHIP_NO_ERROR, );
-	return;
-
-exit:
-	BLEConnectivityManager::Instance().RemoveBLEProvider(ctx->address);
+	/* Discovery was successful, try to bridge connected BLE device with the Matter counterpart. */
+	chip::Optional<uint8_t> indexes[BridgeManager::kMaxBridgedDevicesPerProvider];
+	chip::Optional<uint16_t> endpointIds[BridgeManager::kMaxBridgedDevicesPerProvider];
+	/* AddMatterDevices takes the ownership of the passed provider object and will
+	   delete it in case the BridgeManager fails to accept this object. */
+	CHIP_ERROR err = AddMatterDevices(ctx->deviceTypes, ctx->count, ctx->nodeLabel, ctx->provider);
 	chip::Platform::Delete(ctx);
+
+	return err;
 }
 } // namespace
 
@@ -331,13 +329,17 @@ CHIP_ERROR BleBridgedDeviceFactory::CreateDevice(int deviceType, bt_addr_le_t bt
 	/* Confirm that the first connection was done and this will be only the device recovery. */
 	provider->ConfirmInitialConnection();
 	provider->InitializeBridgedDevice(btAddress, nullptr, nullptr);
-	BLEConnectivityManager::Instance().Recover(provider);
 	err = AddMatterDevices(reinterpret_cast<MatterBridgedDevice::DeviceType *>(&deviceType), 1, nodeLabel, provider,
 			       &index, &endpointId);
-exit:
 
 	if (err != CHIP_NO_ERROR) {
-		BLEConnectivityManager::Instance().RemoveBLEProvider(btAddress);
+		return err;
+	}
+
+	BLEConnectivityManager::Instance().Recover(provider);
+
+exit:
+	if (err != CHIP_NO_ERROR) {
 		chip::Platform::Delete(provider);
 	}
 
@@ -396,9 +398,7 @@ CHIP_ERROR BleBridgedDeviceFactory::CreateDevice(uint16_t uuid, bt_addr_le_t btA
 	}
 
 exit:
-
 	if (err != CHIP_NO_ERROR) {
-		BLEConnectivityManager::Instance().RemoveBLEProvider(btAddress);
 		chip::Platform::Delete(provider);
 	}
 
--- a/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
+++ b/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
@@ -99,6 +99,8 @@ void BLEConnectivityManager::ConnectionHandler(bt_conn *conn, uint8_t conn_err)
 {
 	const bt_addr_le_t *dstAddr = bt_conn_get_dst(conn);
 	BLEBridgedDeviceProvider *provider = nullptr;
+	int err = 0;
+	bool securityFailed = false;
 
 	/* Find the created device instance based on address. */
 	for (int i = 0; i < kMaxConnectedDevices; i++) {
@@ -117,6 +119,10 @@ void BLEConnectivityManager::ConnectionHandler(bt_conn *conn, uint8_t conn_err)
 	if (!provider) {
 		return;
 	}
+
+	/* If there was an error during the initial connection, we should notify the application */
+	bool firstConnFailed = (conn_err && !provider->IsInitiallyConnected());
+	VerifyOrExit(!firstConnFailed, err = conn_err);
 
 	/* The device was successfully recovered. */
 	if (conn_err == 0 && provider->IsInitiallyConnected()) {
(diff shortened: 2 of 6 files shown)


kkasperczyk-no (approved):

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -307,30 +316,39 @@ CHIP_ERROR BridgeManager::AddDevices(MatterBridgedDevice *devices[], BridgedDevi
 				     uint8_t deviceListSize, chip::Optional<uint8_t> devicesPairIndexes[],
 				     uint16_t endpointIds[])
 {
-	VerifyOrReturnError(devices && dataProvider, CHIP_ERROR_INTERNAL);
+	CHIP_ERROR err{ CHIP_NO_ERROR };
+	CHIP_ERROR addingStatus{ CHIP_NO_ERROR };
```
ArekBalysNordic, line 320: Do we need to use two separate CHIP_ERROR indicators?

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
```diff
@@ -186,36 +191,32 @@ struct BluetoothConnectionContext {
 	bt_addr_le_t address;
 };
 
-void BluetoothDeviceConnected(bool discoverySucceeded, void *context)
+CHIP_ERROR BluetoothDeviceConnected(bool success, void *context)
 {
 	if (!context) {
-		return;
+		return CHIP_ERROR_INTERNAL;
```
kkasperczyk-no, line 197: Nit: Maybe it would be better to use CHIP_ERROR_INVALID_ARGUMENT;

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
```diff
@@ -334,10 +335,14 @@ CHIP_ERROR BleBridgedDeviceFactory::CreateDevice(int deviceType, bt_addr_le_t bt
 	BLEConnectivityManager::Instance().Recover(provider);
 	err = AddMatterDevices(reinterpret_cast<MatterBridgedDevice::DeviceType *>(&deviceType), 1, nodeLabel, provider,
 			       &index, &endpointId);
-exit:
 
 	if (err != CHIP_NO_ERROR) {
 		BLEConnectivityManager::Instance().RemoveBLEProvider(btAddress);
+		return err;
```
ArekBalysNordic, line 341: Previously, we call also delete on the provider's memory if there was an error, now it returns failure but we don't call delete anymore. Is it ok?

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
```diff
@@ -186,36 +191,32 @@ struct BluetoothConnectionContext {
...
-									   ctx->provider)) {
-					     BLEConnectivityManager::Instance().RemoveBLEProvider(ctx->address);
-					     chip::Platform::Delete(ctx->provider);
-				     }
-				     chip::Platform::Delete(ctx);
-			     },
-			     reinterpret_cast<intptr_t>(ctx)) == CHIP_NO_ERROR, );
-	return;
+	if (!success) {
+		BLEConnectivityManager::Instance().RemoveBLEProvider(ctx->address);
```
kkasperczyk-no, line 203: As we discussed, I think that it would be safer to move removing BLE provider to the BLE connectivity manager side.

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
```diff
@@ -334,10 +335,14 @@ CHIP_ERROR BleBridgedDeviceFactory::CreateDevice(int deviceType, bt_addr_le_t bt
 	BLEConnectivityManager::Instance().Recover(provider);
 	err = AddMatterDevices(reinterpret_cast<MatterBridgedDevice::DeviceType *>(&deviceType), 1, nodeLabel, provider,
 			       &index, &endpointId);
-exit:
 
 	if (err != CHIP_NO_ERROR) {
 		BLEConnectivityManager::Instance().RemoveBLEProvider(btAddress);
```
kkasperczyk-no, line 340: Ditto

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
```diff
@@ -186,36 +191,32 @@ struct BluetoothConnectionContext {
...
-exit:
-	BLEConnectivityManager::Instance().RemoveBLEProvider(ctx->address);
+	/* Discovery was successful, try to bridge connected BLE device with the Matter counterpart. */
+	chip::Optional<uint8_t> indexes[BridgeManager::kMaxBridgedDevicesPerProvider];
+	chip::Optional<uint16_t> endpointIds[BridgeManager::kMaxBridgedDevicesPerProvider];
+	/* AddMatterDevices takes the ownership of the passed provider object and will
+	   delete it in case the BridgeManager fails to accept this object. */
+	CHIP_ERROR err = AddMatterDevices(ctx->deviceTypes, ctx->count, ctx->nodeLabel, ctx->provider);
+	if (CHIP_NO_ERROR != err) {
+		BLEConnectivityManager::Instance().RemoveBLEProvider(ctx->address);
```
kkasperczyk-no, line 216: Ditto

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -143,20 +148,27 @@ void BLEConnectivityManager::ConnectionHandler(bt_conn *conn, uint8_t conn_err)
 	LOG_INF("Connected: %s", addrStr);
 
 #if defined(CONFIG_BT_SMP)
-	int err = bt_conn_set_security(conn, static_cast<bt_security_t>(CONFIG_BRIDGE_BT_MINIMUM_SECURITY_LEVEL));
-	if (err) {
-		LOG_ERR("Failed to set security (%d)", err);
```
kkasperczyk-no, line 148: Could we leave this log? It was quite useful in the past.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -143,20 +148,27 @@ void BLEConnectivityManager::ConnectionHandler(bt_conn *conn, uint8_t conn_err)
...
-		return;
-	}
+	err = bt_conn_set_security(conn, static_cast<bt_security_t>(CONFIG_BRIDGE_BT_MINIMUM_SECURITY_LEVEL));
+	VerifyOrExit(err == 0, );
```
kkasperczyk-no, line 152:
```suggestion
	VerifyOrExit(err == 0, LOG_ERR("Failed to set security (%d)", err));
```

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
markaj-nordic: Yes, the object will be deleted in the `AddMatterDevices()`. The fact that we deleted the provider's memory after the aforementioned method returned an error was the direct reason for one of the observed crashes.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -288,22 +300,43 @@ void BLEConnectivityManager::DiscoveryCompletedHandler(bt_gatt_dm *dm, void *con
...
 
-	bt_gatt_dm_data_release(dm);
+	discoveryCtx->mProvider = provider;
+	discoveryCtx->mDiscoveryData = dm;
+
+	CHIP_ERROR err = chip::DeviceLayer::PlatformMgr().ScheduleWork(
+		[](intptr_t context) {
+			auto *ctx = reinterpret_cast<DiscoveryHandlerCtx *>(context);
+			if (!ctx->mProvider->IsInitiallyConnected()) {
+				/* Provider is not initalized it so we need to call the first connection callback */
```
kkasperczyk-no, line 316:
```suggestion
				/* Provider is not initalized, so we need to call the first connection callback. */
```

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -288,22 +300,43 @@ void BLEConnectivityManager::DiscoveryCompletedHandler(bt_gatt_dm *dm, void *con
...
+				CHIP_ERROR err = ctx->mProvider->GetBLEBridgedDevice().mFirstConnectionCallback(
+					ctx->mDiscoveryData,
+					ctx->mProvider->GetBLEBridgedDevice().mFirstConnectionCallbackContext);
+				ctx->mProvider->ConfirmInitialConnection();
+				VerifyOrReturn(CHIP_NO_ERROR == err, bt_gatt_dm_data_release(ctx->mDiscoveryData););
+			}
+
+			VerifyOrReturn(CHIP_NO_ERROR == ctx->mProvider->NotifyReachableStatusChange(true),
```
kkasperczyk-no, line 324: I think we cannot return without releasing gatt dm data and deleting context, right?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -288,22 +300,43 @@ void BLEConnectivityManager::DiscoveryCompletedHandler(bt_gatt_dm *dm, void *con
...
+				       LOG_WRN("The device has not been notified about the status change."));
+
+			VerifyOrReturn(ctx->mProvider->ParseDiscoveredData(ctx->mDiscoveryData) == 0,
```
kkasperczyk-no, line 327: Ditto, I think we should always release gatt data and delete context, no matter if we failed or not.

samples/matter/common/src/bridge/bridge_manager.cpp
markaj-nordic: Right, it's a leftover, thanks for the catch.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -307,30 +316,39 @@ CHIP_ERROR BridgeManager::AddDevices(MatterBridgedDevice *devices[], BridgedDevi
...
+
+	VerifyOrReturnError(devices || dataProvider, CHIP_ERROR_INTERNAL);
```
kkasperczyk-no, line 322: CHIP_ERROR_INTERNAL -> CHIP_ERROR_INVALID_ARGUMENT? And the same to the below ones.

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
markaj-nordic: Yep, I will change that.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
markaj-nordic: I think logging won't work with `VerifyOrExit()` macro.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
markaj-nordic: Right.

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
kkasperczyk-no: It seems like we have one more `BLEConnectivityManager::Instance().RemoveBLEProvider(btAddress);` here.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -288,22 +305,44 @@ void BLEConnectivityManager::DiscoveryCompletedHandler(bt_gatt_dm *dm, void *con
...
+			}
+
+			if (CHIP_NO_ERROR != ctx->mProvider->NotifyReachableStatusChange(true)) {
+				LOG_WRN("The device has not been notified about the status change.");
+			}
+
+			if (0 != ctx->mProvider->ParseDiscoveredData(ctx->mDiscoveryData)) {
+				LOG_ERR("Cannot parse the GATT discovered data.");
+			}
+			bt_gatt_dm_data_release(ctx->mDiscoveryData);
```
kkasperczyk-no, line 337: Shouldn't we delete `ctx` here?

markaj-nordic: `ctx` is now an unique_ptr, so it will be automatically deleted.

kkasperczyk-no: Ah ok, then it's ok.

kkasperczyk-no (approved):
LGTM, great job!

applications/matter_bridge/src/ble_providers/ble_bridged_device_factory.cpp
markaj-nordic: Yes, but this one is a bit more tricky. I was not sure if we really end up with discovery callbacks being triggered in the recovery path unconditionally.

markaj-nordic: Discussed f2f.

ArekBalysNordic (approved):

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -60,30 +61,38 @@ CHIP_ERROR BridgeManager::AddBridgedDevices(MatterBridgedDevice *devices[], Brid
 {
 	chip::Optional<uint8_t> indexes[deviceListSize];
 	uint16_t endpoints[deviceListSize];
+	Platform::UniquePtr<BridgedDeviceDataProvider> providerPtr(dataProvider);
 
 	VerifyOrReturnError(devicesPairIndexes, CHIP_ERROR_INTERNAL);
-
-	return AddBridgedDevices(devices, dataProvider, deviceListSize, devicesPairIndexes, endpoints, indexes);
+	CHIP_ERROR err =
+		DoAddBridgedDevices(devices, providerPtr.get(), deviceListSize, devicesPairIndexes, endpoints, indexes);
```
Damian-Nordic, line 68: nit: I guess `DoAddBridgedDevices` could take `UniquePtr` argument to make it clear that it takes the owhership.

samples/matter/common/src/bridge/bridge_manager.cpp
```diff
@@ -307,30 +316,38 @@ CHIP_ERROR BridgeManager::AddDevices(MatterBridgedDevice *devices[], BridgedDevi
...
 
-	CHIP_ERROR status{ CHIP_NO_ERROR };
 	for (auto i = 0; i < deviceListSize; ++i) {
-		status = AddSingleDevice(devices[i], dataProvider, devicesPairIndexes[i], endpointIds[i]);
+		err = AddSingleDevice(devices[i], dataProvider, devicesPairIndexes[i], endpointIds[i]);
 
-		if (status != CHIP_NO_ERROR) {
-			return status;
+		if (err != CHIP_NO_ERROR) {
+			return err;
```
Damian-Nordic, line 334: In this case, shouldn't we release the devices not added so far? Or maybe remove those added and release all?

Damian-Nordic (approved):

samples/matter/common/src/bridge/bridge_manager.cpp
markaj-nordic: Hmm yes, definitely the second option makes sense to me (though it's quite unluckily to hit it). @kkasperczyk-no any thoughts? I guess it's pointless to keep only the "part" of the bridged device when we fail to add the rest to the bridge (which would be the first Damian's suggestion).

kkasperczyk-no: Yes, definitely. We should change it to `VerifyOrExit(err == CHIP_NO_ERROR)`

markaj-nordic: It won't be that easy :) We need to take care about already added device and remove it explicitly from the `mDevicesMap`. Anyway, we are on the same page, so I'll try to implement the fix.

markaj-nordic: Ok, I discovered some further limitations in the Bridge Manager, so I it's even more difficult that I initially thought. The good thing is that the case when we fail to add devices here is currently prevented by `VerifyOrExit(mDevicesMap.FreeSlots() >= deviceListSize, err = CHIP_ERROR_NO_MEMORY);` precondition. Let's close this PR and I will open up a follow up if I find the proper fix.

PR #13835, opened 2024-02-01
net: lwm2m_client_utils: use modem_info to find HW version

SeppoTakalo:
In LwM2M Device object, as well as Advanced firmware object, we have modem hardware type visible.
Query this information from modem_info library, instead of using build-in values.
This allows same FW to run on 9161 and 9151 and we still see from server side which HW is used.

### DIFF
--- a/subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
+++ b/subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
@@ -15,6 +15,7 @@
 #include <zephyr/net/lwm2m.h>
 #include <modem/nrf_modem_lib.h>
 #include <modem/modem_key_mgmt.h>
+#include <modem/modem_info.h>
 #include <net/fota_download.h>
 #include <fota_download_util.h>
 #include <lwm2m_util.h>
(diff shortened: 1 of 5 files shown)


juhaylinen (approved):

subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
```diff
@@ -1266,8 +1267,13 @@ int lwm2m_init_firmware_cb(lwm2m_firmware_event_cb_t cb)
 		"application", firmware_block_received_cb, firmware_update_cb);
 	lwm2m_firmware_object_setup_init(application_obj_id);
 
-	modem_obj_id = lwm2m_adv_firmware_create_inst(
-		"modem:" CONFIG_SOC, firmware_block_received_cb, firmware_update_cb);
+	static char hw_buf[sizeof("modem:nRF91__ ____ ___ ")];
+
+	strcat(hw_buf, "modem:");
+	modem_info_get_hw_version(hw_buf + strlen("modem:"), strlen(hw_buf) - strlen("modem:"));
```
MarkusLassila, line 1273: modem_info_get_hw_version seems to want the size of the target buffer. As this is using strlen to calculate, it will ignore the null character and send the remaining buffer space without it?

SeppoTakalo: Good catch.
Looks like I made an error on last refactoring. It should have been `sizeof() - strlen()`

MarkusLassila (approved):

PR #13836, opened 2024-02-01
applications: Fixed onoff light switch in Matter bridge

kkasperczyk-no:
The binding support was refactored for the onoff light switch in a wrong way and CI didn't caught it, because the build was not in twister.

Fixed the implementation and added configuration including onoff switch to sample.yaml.

### DIFF
--- a/applications/matter_bridge/src/ble_providers/ble_lbs_data_provider.cpp
+++ b/applications/matter_bridge/src/ble_providers/ble_lbs_data_provider.cpp
@@ -28,26 +28,18 @@ static bt_uuid *sUuidButton = BT_UUID_LBS_BUTTON;
 static bt_uuid *sUuidCcc = BT_UUID_GATT_CCC;
 
 #ifdef CONFIG_BRIDGE_ONOFF_LIGHT_SWITCH_BRIDGED_DEVICE
-void ProcessCommand(CommandId aCommandId, const EmberBindingTableEntry &aBinding, OperationalDeviceProxy *aDevice,
-		    void *aContext)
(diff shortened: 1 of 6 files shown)


kkasperczyk-no:
Putting DNM before performing smoke tests on desk.

kkasperczyk-no:
Both simulated and bt light switches seems to work correctly.

ArekBalysNordic (approved):
LGTM

applications/matter_bridge/src/ble_providers/ble_lbs_data_provider.cpp
```diff
@@ -28,26 +28,18 @@ static bt_uuid *sUuidButton = BT_UUID_LBS_BUTTON;
 static bt_uuid *sUuidCcc = BT_UUID_GATT_CCC;
 
 #ifdef CONFIG_BRIDGE_ONOFF_LIGHT_SWITCH_BRIDGED_DEVICE
-void ProcessCommand(CommandId aCommandId, const EmberBindingTableEntry &aBinding, OperationalDeviceProxy *aDevice,
-		    void *aContext)
+void ProcessCommand(const EmberBindingTableEntry &aBinding, OperationalDeviceProxy *aDevice, Nrf::Matter::BindingHandler::BindingData &aData)
```
markaj-nordic, line 31: I guess we don't need `Nrf` namespace here and there in this file, maybe the same for `Matter`.

markaj-nordic (approved):

applications/matter_bridge/src/ble_providers/ble_lbs_data_provider.cpp
kkasperczyk-no: I think that Matter is required, but indeed Nrf should not be needed.

LuDuda (approved):

PR #13840, opened 2024-02-01
lib: date_time: Add local time API function

ewalt1:
This provides a substitute for localtime() since it isn't implemented in Zephyr's minimal libc. The resulting timestamp can be used with <ctime> or <chrono> for manipulation and logging.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -985,6 +985,10 @@ Other libraries
   * Added more default LTE metrics, such as band, operator, RSRP, and kilobytes sent and received.
   * Updated the default metric names to follow the standard |NCS| variable name convention.
 
+* :ref:`lib_date_time` library:
+
+  * Added the :c:func:`date_time_now_local()` function to the API.
+
 Common Application Framework (CAF)
 ----------------------------------
 
--- a/include/date_time.h
+++ b/include/date_time.h
@@ -91,6 +91,21 @@ int date_time_uptime_to_unix_time_ms(int64_t *uptime);
  */
 int date_time_now(int64_t *unix_time_ms);
 
+/** @brief Get the current date time in local time.
+ *
+ *  @note If the function fails, the passed in variable retains its
+ *        old value.
+ *
+ *  @param[out] local_time_ms Pointer to a variable to store the current date
+ *                            time.
+ *
+ *  @return 0        If the operation was successful.
+ *  @return -ENODATA If the library does not have a valid date time.
+ *  @return -EAGAIN  If the library has the date time, but not a valid timezone.
+ *  @return -EINVAL  If the passed in pointer is NULL.
+ */
+int date_time_now_local(int64_t *local_time_ms);
+
 /** @brief Convenience function that checks if the library has obtained
  *	   an initial valid date time.
  *
@@ -103,6 +118,17 @@ int date_time_now(int64_t *unix_time_ms);
  *  @return false The library has not obtained an initial date time.
  */
 bool date_time_is_valid(void);
+
+/** @brief Check if the library has obtained
+ *	   an initial valid date time and timezone.
+ *
+ *  @note If this function returns false there is no point of
+ *	  subsequent calls to date_time_now_local().
+ *
+ *  @return true  The library has obtained a local date time.
+ *  @return false The library has not obtained a local date time.
+ */
+bool date_time_is_valid_local(void);
 
 /** @brief Register an event handler for Date time library events.
  *
--- a/lib/date_time/date_time.c
+++ b/lib/date_time/date_time.c
@@ -126,9 +126,28 @@ int date_time_now(int64_t *unix_time_ms)
 	return err;
 }
 
+int date_time_now_local(int64_t *local_time_ms)
+{
+	if (local_time_ms == NULL) {
+		LOG_ERR("The passed in pointer cannot be NULL");
+		return -EINVAL;
+	}
+	if (!date_time_is_valid()) {
+		LOG_WRN("Valid time not currently available");
+		return -ENODATA;
+	}
+
+	return date_time_core_now_local(local_time_ms);
+}
+
 bool date_time_is_valid(void)
 {
 	return date_time_core_is_valid();
+}
+
+bool date_time_is_valid_local(void)
+{
+	return date_time_core_is_valid_local();
 }
 
 void date_time_register_handler(date_time_evt_handler_t evt_handler)
--- a/lib/date_time/date_time_core.c
+++ b/lib/date_time/date_time_core.c
@@ -33,6 +33,9 @@ static K_WORK_DELAYABLE_DEFINE(time_work, date_time_handler);
 static int64_t date_time_last_update_uptime;
 static date_time_evt_handler_t app_evt_handler;
 
+/* In units of quarters of hours, same as used by AT+CCLK. */
+static int date_time_tz = DATE_TIME_TZ_INVALID;
+
 static void date_time_core_notify_event(enum date_time_evt_type time_source)
 {
 	static struct date_time_evt evt;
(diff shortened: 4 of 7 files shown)


lib/date_time/date_time_modem.c
```diff
@@ -53,17 +54,18 @@ int date_time_modem_get(int64_t *date_time_ms)
...
 		&date_time.tm_hour,
 		&date_time.tm_min,
-		&date_time.tm_sec
+		&date_time.tm_sec,
+		&tz
 	);
 
 	/* Want to match 6 args */
-	if (rc != 6) {
+	if (rc != 7) {
```
tokangas, line 68: Modem time can be set using `%CCLK` without specifying the time zone (value `+99` means that the time zone is "not set" or "unknown"). This results in `+CCLK?` returning a string without the time zone. I'm not sure if the network can give time without specifying the time zone, but I think it would be good to accept both 6 and 7 parameters here. If the time zone is missing, `tz` should be `0`.

lib/date_time/date_time_modem.c
```diff
@@ -195,9 +204,9 @@ void date_time_modem_store(struct tm *ltm)
 		 * with modulo 100.
```
tokangas, line 204: The time zone related comment above is obsolete and should be removed.

tokangas (commented):
Thanks for your contribution!

lib/date_time/date_time_modem.c
ewalt1: done

include/date_time.h
```diff
@@ -91,6 +91,20 @@ int date_time_uptime_to_unix_time_ms(int64_t *uptime);
...
+ *        old value.
+ *
+ *  @param[out] local_time_ms Pointer to a variable to store the current date
+ *                            time.
+ *
+ *  @return 0        If the operation was successful.
+ *  @return -ENODATA If the library does not have a valid date time.
+ *  @return -EINVAL  If the passed in pointer is NULL.
+ */
+int date_time_now_local(int64_t *local_time_ms);
```
trantanen, line 106: If the timezone is not available within the library when modem doesn't give it, or NTP is used, this function will basically return UTC time (0 time zone). I'm thinking if that's what we want without knowing about not really necessarily being at UTC.
We could return some error code and still fill the `local_time_ms` with the UTC time and application could determine what to do.
This is just a thought and feel free to comment? Also interested in comments from @tokangas.

lib/date_time/date_time.c
```diff
@@ -126,6 +126,24 @@ int date_time_now(int64_t *unix_time_ms)
 	return err;
 }
 
+int date_time_now_local(int64_t *local_time_ms)
```
trantanen, line 129: Maybe add a unit test for this function in `tests/lib/date_time/src/main.c`.

trantanen (changes requested):
Looks like a good addition to the API.
Please add changelog entry `doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst` mentioning that a new `date_time_now_local()` function has been added.
A new entry for date time library should be added under `Other libraries`.

include/date_time.h
tokangas: I like the idea of returning a specific error in case the timezone is not known. I agree, that the user of the library should somehow be made aware of this.

lib/date_time/date_time_core.c
```diff
@@ -32,6 +32,7 @@ static K_WORK_DELAYABLE_DEFINE(time_work, date_time_handler);
 
 static int64_t date_time_last_update_uptime;
 static date_time_evt_handler_t app_evt_handler;
+static int date_time_tz = DATE_TIME_TZ_INVALID;
```
trantanen, line 35: Add a comment that this is in quarters of an hour like in the AT command

lib/date_time/date_time_core.c
```diff
@@ -174,6 +176,23 @@ int date_time_core_now(int64_t *unix_time_ms)
...
+	if (date_time_tz == DATE_TIME_TZ_INVALID) {
+		return -EAGAIN;
+	}
+
+	err = date_time_core_now(local_time_ms);
+	if (err) {
+		return err;
+	}
+
+	*local_time_ms += date_time_tz * 15 * 60 * 1000;
```
trantanen, line 192:
```suggestion
	*local_time_ms += date_time_tz * 15 * MIN_PER_HOUR * MSEC_PER_SEC;
```
You can add the use of those constants to line 174 if you want but not necessary as it's old code.

lib/date_time/date_time_core.c
ewalt1: done

lib/date_time/date_time_core.c
ewalt1: It should be SEC_PER_MIN, but I'm not going to add it since the version of the sdk that I'm on doesn't have those defines.

trantanen: > It should be SEC_PER_MIN

Of course 😄 

> but I'm not going to add it since the version of the sdk that I'm on doesn't have those defines.

Ok, I'll update those at some point.

include/date_time.h
```diff
@@ -104,6 +119,17 @@ int date_time_now(int64_t *unix_time_ms);
...
+/** @brief Convenience function that checks if the library has obtained
+ *	   an initial valid date time and timezone.
+ *
+ *  @note If this function returns false there is no point of
+ *	  subsequent calls to date_time_now_local().
+ *
+ *  @return true  The library has obtained a local date time.
+ *  @return false The library has not obtained a local date time.
+ */
+bool date_time_is_valid_local(void);
```
trantanen, line 131: Are you using this function at the moment without needing the time value? If not, I wouldn't add this function to the API until there is a need for it.

trantanen: Ok, that's fine then. Apart from the unit test comment, I don't think I have anything more.

ewalt1: Unfortunately I use the SDK out of tree and don't have the unit test framework set up. Feel free to close the PR if that's a deal-breaker.

trantanen: Ok, I'll add the test later.

lib/date_time/date_time.c
trantanen: I'll add the test later as author doesn't have setup.

trantanen:
Please rebase and handle conflicts in the changelog.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -863,6 +863,10 @@ Other libraries
 
   * Fixed issue where the adp536x driver was included in the immutable bootloader on Thingy:91 when :kconfig:option:`CONFIG_SECURE_BOOT` was enabled.
 
+* :ref:`lib_date_time` library:
+
+  * Added date_time_now_local() to API.
```
peknis, line 868:
```suggestion
  * Added the :c:func:`date_time_now_local()` function to the API.
```

include/date_time.h
```diff
@@ -104,6 +119,17 @@ int date_time_now(int64_t *unix_time_ms);
  */
 bool date_time_is_valid(void);
 
+/** @brief Convenience function that checks if the library has obtained
```
peknis, line 122:
```suggestion
/** @brief Check if the library has obtained
```

peknis (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
trantanen: @peknis
Should this be without `()`? That's how I've seen it usually when using `:c:func:` referencing.

peknis: Good point, probably without, yes.

lib/date_time/date_time.c
```diff
@@ -126,11 +126,34 @@ int date_time_now(int64_t *unix_time_ms)
...
+		return -EINVAL;
+	}
+	if (!date_time_is_valid()) {
+		LOG_WRN("Valid time not currently available");
+		return -ENODATA;
+	}
+
+	err = date_time_core_now_local(local_time_ms);
+
+	return err;
```
tokangas, line 144: Just a suggestion, `err` is not really needed.
```suggestion
	if (local_time_ms == NULL) {
		LOG_ERR("The passed in pointer cannot be NULL");
		return -EINVAL;
	}
	if (!date_time_is_valid()) {
		LOG_WRN("Valid time not currently available");
		return -ENODATA;
	}

	return date_time_core_now_local(local_time_ms);
```

tokangas (approved):
Looks good to me, thanks!

trantanen:
@ewalt1 If you can get this rebased and look at those few nits, I could run the CI and get this merged.

trantanen:
@ewalt1 I can see the content is fine now. However, merge commits are not allowed in sdk-nrf but you should use rebase instead, which the compliance run also points out. Also, squash all your update commits into the main commit.

ewalt1:
squashed and rebased

peknis (approved):

trantanen (approved):
Thanks @ewalt1 for the effort!

PR #13844, opened 2024-02-02
samples: net: http_server: Improvements for cellular

stig-bjorlykke:
* Configure maximum Zephyr sockets because they are also used in cellular context.
* Do not trust the received client_addr sin_family from accept() to create thread.
* Remove use of poll() before accept().


simensrostad (approved):
LGTM!

gregersrygg (approved):

PR #13846, opened 2024-02-02
Asset tracker: Fix LwM2M device data

SeppoTakalo:
Two issues were found on Asset Tracker when LwM2M protocol is used:
* Device data is initialized only after we have registered to server. This might lead to situation that server fetch empty strings from Device object. Move few initialization calls to setup function.
* String terminator was not counted in the buffer length. This is a bug as `lwm2m_set_res_buf()` requires full buffer, not just `strlen()`. Some payload codecs have been ignoring this, but current CBOR codec seem to strip last character from strings.

This does not completely fix Asset Tracker v2 as I am not familiar with the state machines. Someone should go through the state machine and ensure that following data is filled up before we register to server:
* lwm2m_codec_helpers_set_modem_dynamic_data
* lwm2m_codec_helpers_set_modem_static_data
* lwm2m_codec_helpers_set_battery_data
* lwm2m_codec_helpers_set_sensor_data
* lwm2m_codec_helpers_set_user_interface_data
* lwm2m_codec_helpers_set_neighbor_cell_data

All that data will appear visible on Coiote after it runs full device data refresh on registration. If you write that data later, it only shows on server, if you do SEND or if  the data is marked as observed.

### DIFF
--- a/applications/asset_tracker_v2/src/cloud/cloud_codec/lwm2m/lwm2m_codec_helpers.c
+++ b/applications/asset_tracker_v2/src/cloud/cloud_codec/lwm2m/lwm2m_codec_helpers.c
@@ -263,23 +263,23 @@ int lwm2m_codec_helpers_setup_resources(void)
 	}
 
 	err = lwm2m_set_res_buf(&LWM2M_OBJ(IPSO_OBJECT_TEMP_SENSOR_ID, 0, SENSOR_UNITS_RID),
-				BME680_TEMP_UNIT, (uint16_t)strlen(BME680_TEMP_UNIT),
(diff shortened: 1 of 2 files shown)


simensrostad (approved):

simensrostad:
@SeppoTakalo 
_This does not completely fix Asset Tracker v2 as I am not familiar with the state machines. Someone should go through the state machine and ensure that following data is filled up before we register to server._

ATM we don't fetch all of those values before registering. Only a subset where the defaults are static. Can look into providing a way to do an inital sample cycle before registering is triggered.

jtguggedal (approved):

PR #13853, opened 2024-02-02
samples: cellular: at_client: fix kconfig

maxd-nordic:
This patch fixes a bug where the roam notifications need too much heap and the at_client would crash when trying to attach. ~Also, it enables modem tracing which is needed for the Cellular Monitor.~


maxd-nordic:
I got this error message because of heap:
```
+CSCON: 1
%CESQ: 46,2,21,3
+CGEV: MASSERTION FAIL [at_notif] @ WEST_TOPDIR/nrf/lib/at_monitor/at_monitor.c:79
        No heap space for incoming notification: +CIND: "roam",1
*** Booting nRF Connect SDK zephyr-v3.5.0-2991-g9c754a80d072 ***
The AT host sample started
Ready
```

bjda (approved):
Makes sense to me

samples/cellular/at_client/prj.conf
```diff
@@ -23,3 +23,5 @@ CONFIG_UART_INTERRUPT_DRIVEN=y
 # Stacks and heaps
 CONFIG_MAIN_STACK_SIZE=3072
 CONFIG_HEAP_MEM_POOL_SIZE=16384
+CONFIG_NRF_MODEM_LIB_TRACE=y
```
eivindj-nordic, line 26: Traces should be enabled with the `nrf91-modem-trace-uart` snippet. This enables the deprecated UART trace backend.

maxd-nordic: Should the default behavior of this sample be incompatible with Cellular Monitor?
Or should we add the contents of the snippet directly?

maxd-nordic: I changed it to using the new backend now.

eivindj-nordic: I agree that we should keep the traces disabled by default and document how to enable them. Ideally we might want to do this for other samples as well to keep it consistent.

maxd-nordic: Alright, I changed the PR to only include the heap size option. Still, I think this will be confusing for customers.

eivindj-nordic (approved):

jtguggedal (approved):

PR #13857, opened 2024-02-03
samples: cellular: modem_shell: Make sec_tag unsigned

stig-bjorlykke:
Allow security tag values up to 4294967295.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -427,6 +427,7 @@ Cellular samples
     * The ``send`` subcommand to use the :c:func:`send` function for non-secure datagram sockets that are connected and have a peer address set.
     * The ``modem_trace`` command has been moved to :ref:`nrf_modem_lib_readme`.
       See :ref:`modem_trace_shell_command` for information about modem trace commands.
+    * The sample to allow TLS/DTLS security tag values up to ``4294967295``.
 
 * :ref:`nrf_cloud_multi_service` sample:
 
--- a/samples/cellular/modem_shell/src/sock/sock_shell.c
+++ b/samples/cellular/modem_shell/src/sock/sock_shell.c
@@ -341,7 +341,7 @@ static int cmd_sock_connect(const struct shell *shell, size_t argc, char **argv)
 	int arg_bind_port = 0;
 	int arg_pdn_cid = 0;
 	bool arg_secure = false;
-	int arg_sec_tag = -1;
+	uint32_t arg_sec_tag = 0;
 	bool arg_session_cache = false;
 	int arg_peer_verify = 2;
 	char arg_peer_hostname[SOCK_MAX_ADDR_LEN + 1];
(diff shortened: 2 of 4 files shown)


samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -314,11 +314,11 @@ static int cmd_sock_connect(const struct shell *shell, size_t argc, char **argv)
 			arg_secure = true;
 			break;
 		case 'T': /* Security tag */
-			arg_sec_tag = atoi(optarg);
-			if (arg_sec_tag < 0) {
+			arg_sec_tag = shell_strtoul(optarg, 10, &err);
```
tokangas, line 317: I was unaware of `shell_strtoul()` with error checks, nice. It seems, that the function does not set the `err` reference to zero in the success case. Now `err` happens to be zero, but just wondering if it should anyway be initialized to zero before the function call, just in case.

stig-bjorlykke: I don't know if I like the pattern of always setting `err = 0` first. If `err` is not zero the code will always return, and the extra init should not be needed.

We should use `shell_strtol()` and `shell_strtoul()` instead of `atoi()` in all shell code (~ 80 places), and it's good to determine this pattern first.

tokangas: I don't like it either, I would have preferred that `shell_strto(u)l()` would have done it. If `err` is not explicitly initialized before the function call, you always need to check the code to make sure it's zero before the function is called. Even though it works now, somebody may later modify one of the cases so, that `err` may not always be zero. There can be cases, where `err` would be set to a non-zero value, but it would not be considered an error and processing of the parameters would be continued. I don't like that we have to make assumptions about how unrelated code works. @trantanen?

trantanen: > I don't know if I like the pattern of always setting `err = 0` first. If `err` is not zero the code will always return, and the extra init should not be needed.

So the code will return when an integer is successfully parsed in cases when `err` happens to be non-zero in the memory before calling the function. I don't think this is what we want or what am I missing?

trantanen: So is there a better alternative than simply setting `err = 0` before `shell_strto(u)l()` calls?
An own wrapper function could do it but I'd prefer `err = 0`.

stig-bjorlykke: Ok, I'll do `err = 0` for now.

tokangas (approved):

trantanen (approved):

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -429,11 +429,12 @@ static int cmd_sock_connect(const struct shell *shell, size_t argc, char **argv)
...
-			arg_sec_tag = atoi(optarg);
-			if (arg_sec_tag < 0) {
+			err = 0;
+			arg_sec_tag = shell_strtoul(optarg, 10, &err);
+			if (err) {
 				mosh_error(
-					"Valid range for security tag (%d) is 0 ... 2147483647.",
-					arg_sec_tag);
+					"Valid range for security tag (%s) is 0 .. %u.",
+					optarg, ULONG_MAX);
```
MirkoCovizzi, line 437: I think that `UINT32_MAX` fits better and is more readable.
```suggestion
					optarg, UINT32_MAX);
```

stig-bjorlykke: Ok

MirkoCovizzi (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -403,6 +403,7 @@ Cellular samples
       The ``rai_no_data`` option requires the socket to be connected and have a peer address set.
       This bug is caused by the non-secure datagram socket not being connected when using the ``connect`` subcommand.
     * The ``send`` subcommand to use the :c:func:`send` function for non-secure datagram sockets that are connected and have a peer address set.
+    * Allow TLS/DTLS security tag values up to 4294967295.
```
divipillai, line 406:
```suggestion
    * The sample to allow TLS/DTLS security tag values up to ``4294967295``.
```

divipillai (approved):

PR #13861, opened 2024-02-05
net: lwm2m_client_utils: Don't expect result from inform event

SeppoTakalo:
net: lwm2m_client_utils: Don't expect result from inform event
      
Do not always expect the result from Ground Fix events.
When we only do "inform" event, where 'Send location back' is
set to false, the Coiote server is not sending any response back
towards us.
      
When we receive acknoledgement from send operation, assume the
whole ground fix event is OK.

FYI: @simensrostad   I found the root cause for ground fix retrying.


simensrostad (approved):
lgtm

juhaylinen (approved):

PR #13863, opened 2024-02-05
samples: cellular: modem_shell: Add sock send --wait_ack

stig-bjorlykke:
Add an option to sock send to specify MSG_WAITACK. This will request a blocking send operation until the data has been sent on-air and acknowledged by the peer if required by the network protocol.

Set requested blocking mode for a send with --data option because this will affect --wait_ack.

Increased time precision from 2 to 3 decimals in throughput summary to get milliseconds.


samples/cellular/modem_shell/src/sock/sock.c
```diff
@@ -992,8 +999,14 @@ int sock_send_data(
 		}
 
 	} else if (data_out != NULL && data_out_length > 0) {
+		/* Set requested blocking mode for duration of data sending */
+		set_sock_blocking_mode(socket_info->fd, blocking);
+
 		/* Send data if it's given and is not zero length */
-		sock_send(socket_info, data_out, data_out_length, true, data_format_hex);
+		sock_send(socket_info, data_out, data_out_length, flags, true, data_format_hex);
```
trantanen, line 1006: Do you need to pass the flags as a separate argument when they are stored already in `socket_info->send_flags`?

samples/cellular/modem_shell/src/sock/sock.c
```diff
@@ -743,7 +745,7 @@ static int sock_send(
 		}
 	}
 
-	bytes = send(socket_info->fd, data, length, 0);
+	bytes = send(socket_info->fd, data, length, flags);
```
trantanen: I'm thinking if setting the blocking mode could be handled here but haven't really been able to think through all possible scenarios.

samples/cellular/modem_shell/src/sock/sock.c
stig-bjorlykke: No, I will remove this extra argument.

trantanen (changes requested):
One comment was sent earlier as I pressed the wrong button.

samples/cellular/modem_shell/src/sock/sock.c
```diff
@@ -992,8 +999,14 @@ int sock_send_data(
 		}
 
 	} else if (data_out != NULL && data_out_length > 0) {
+		/* Set requested blocking mode for duration of data sending */
+		set_sock_blocking_mode(socket_info->fd, blocking);
```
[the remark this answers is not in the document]
stig-bjorlykke: For periodic sending (`-e` option) it's possible to subtract the time used in `send()` before scheduling next send. A small corner case if setting `-e 1` and the send uses more than 1 second, but I have only seen this happen on wakeup from PSM.

For sending a chunk of data (`-d` option) I think it's good to be able to specify blocking or non-blocking. Maybe also printing the time used in send when having the `-W` option.

samples/cellular/modem_shell/src/sock/sock.c
trantanen: Maybe this is not the most optimal place as we call this in a loop in `sock_send_random_data_length()`.

samples/cellular/modem_shell/src/sock/sock.c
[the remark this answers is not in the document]
trantanen: This seems to be my misunderstanding. So adding blocking mode handling for `-d` option was not directly related to `-W` option, I guess. I somehow thought it was needed to make `-W` option work at all.

> For periodic sending (`-e` option) it's possible to subtract the time used in `send()` before scheduling next send. A small corner case if setting `-e 1` and the send uses more than 1 second, but I have only seen this happen on wakeup from PSM.

I'm not sure if we need to subtract anything or just let the interval start when sending is complete, i.e., how it should work with the current code.

> For sending a chunk of data (`-d` option) I think it's good to be able to specify blocking or non-blocking.

True.

> Maybe also printing the time used in send when having the `-W` option.

Could be done but not necessary.

samples/cellular/modem_shell/src/sock/sock.c
trantanen: This comment is outdated as I've understood few things incorrectly at the beginning.

samples/cellular/modem_shell/src/sock/sock.c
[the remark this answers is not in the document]
trantanen: > I made blocking mode configurable for `-d` to be able to test different combinations, but maybe we should print a warning in case `-B 0 -W` is used?

I guess it would be good to avoid unintentional misuse.

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -111,6 +111,9 @@ static const char sock_send_usage_str[] =
 	"                                010203040506070809101112\n"
 	"                                01 02 03 04 05 06 07 08 09 10 11 12\n"
 	"                                01020304 05060708 09101112\n"
+	"  -W, --wait_ack            Request a blocking send operation until the data\n"
+	"                            has been sent on-air and acknowledged by the peer,\n"
+	"                            if required by the network protocol.\n"
```
trantanen, line 116: Now that I understand the changes better, I guess we should mention here that you should set `-B 1` whenever this `-W` is used.

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -96,8 +96,8 @@ static const char sock_send_usage_str[] =
 	"                            The number is between 0001..9999 and it can roll over.\n"
 	"                            Must be used with -d and -e options and without\n"
 	"                            -l and -x options.\n"
-	"  -B, --blocking, [int]     Blocking (1) or non-blocking (0) mode. This is only\n"
-	"                            valid when -l is given. Default value is 1.\n"
+	"  -B, --blocking, [int]     Blocking (1) or non-blocking (0) mode.\n"
+	"                            Default value is 1.\n"
```
trantanen, line 100: I guess this doesn't work with `-e`. Unless we change it, we should mentioned that this is not supported with `-e` option.

stig-bjorlykke: I don't see why `-B` shouldn't work with `-e`. Blocking or non-blocking shouldn't affect multiple send?

samples/cellular/modem_shell/src/sock/sock_shell.c
stig-bjorlykke: `-B 1` is default, but maybe add a comment that `-B 0` should not be used?

What about the issue with not logging "socket send failed" for non-blocking sockets in `sock_send()`? If this (TCP related) issue is fixed now I think that restriction should be removed, and the send failure will indicate that the `-B 0 -W` combination is not supported (here it will be `EOPNOTSUPP`).

samples/cellular/modem_shell/src/sock/sock_shell.c
trantanen: `set_sock_blocking_mode()` is not used for `-e` if I'm not mistaken.
Earlier `set_sock_blocking_mode()` was only used with `-l`, and now it's done also with `-d` but not with `-d` and `-e`.

stig-bjorlykke: Right. Do we want to keep this restriction, or do we add support for using `-B` with `-e`?

samples/cellular/modem_shell/src/sock/sock.c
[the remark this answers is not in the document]
stig-bjorlykke: I documented this restriction.

samples/cellular/modem_shell/src/sock/sock_shell.c
stig-bjorlykke: I documented this for now.

samples/cellular/modem_shell/src/sock/sock_shell.c
stig-bjorlykke: I will create a separate PR for enabling "socket send failed" for non-blocking sockets.

samples/cellular/modem_shell/src/sock/sock_shell.c
trantanen: Ok. I don't see a problem for adding `-B 1 -e` support too if you want.

samples/cellular/modem_shell/src/sock/sock_shell.c
trantanen: I don't know if this issue has been fixes or not. Or do you know it?
I recall it was something harder to manage and I don't recall that there would have been attempts to fix it but I'm not sure.

stig-bjorlykke: I will do some testing and check around, the reference page does not exist.
Let's move this discussion to PR #13902.

trantanen (approved):

tokangas (approved):

samples/cellular/modem_shell/src/sock/sock_shell.c
```diff
@@ -111,6 +111,10 @@ static const char sock_send_usage_str[] =
 	"                                010203040506070809101112\n"
 	"                                01 02 03 04 05 06 07 08 09 10 11 12\n"
 	"                                01020304 05060708 09101112\n"
+	"  -W, --wait_ack            Request a blocking send operation until the data\n"
+	"                            has been sent on-air and acknowledged by the peer,\n"
+	"                            if required by the network protocol. This is not valid\n"
+	"                            when -B0 (non-blocking mode) is used.\n"
```
stig-bjorlykke, line 117: It turns out that `-B0 -W` correctly gives an error in send(), but this is camouflaged by not printing error from send in non-blocking mode and **gives an infinite loop** when using the `-l` option. I think other errors from send also could give the same loop without being noticed.

I can either add a check for `-B0 -W` or fix the infinite loop issue (which can happen in other cases).

stig-bjorlykke: I did the first option for this PR.

samples/cellular/modem_shell/src/sock/sock.c
stig-bjorlykke: > I'm thinking if setting the blocking mode could be handled here but haven't really been able to think through all possible scenarios.

Actually I think it's a good idea to use `MSG_DONTWAIT` here instead of `fcntl(fd, F_SETFL, O_NONBLOCK)`. When sending a large random data block the `recv()` may happen during the `send()` loop, and in this case the `recv()` will be either blocking or non-blocking, depending on if the `send()` loop is done or not. I observe this when testing with `sock send -i0 -l65535`.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -380,6 +380,7 @@ Cellular samples
     * Support for DNS query using the ``sock getaddrinfo`` command.
     * Support for PDN CID 0 in the ``-I`` argument for the ``sock connect`` command.
     * Support for listing interface addresses using the ``link ifaddrs`` command.
+    * Support for send MSG_WAITACK flag using the ``sock send`` command.
```
divipillai, line 383:
```suggestion
    * Support for the ``MSG_WAITACK`` send flag using the ``sock send`` command.
```
Please check if the meaning is maintained.

divipillai (approved):

PR #13866, opened 2024-02-05
manifest: updating nRF RPC image in nrfxlib

peknis:
Manifest for https://github.com/nrfconnect/sdk-nrfxlib/pull/1193


peknis:
Something was not right. I'll try again.

PR #13870, opened 2024-02-05
Disable legacy mbedtls by default

SebastianBoe:
I don't know why we are enabling legacy mbedtls by default for TF-M builds.

This PR changes the defaults to default n.

I have observed that we save 4kB of RAM in the sha256 sample with this PR.


vili-nordic (approved):

maciejbaczmanski:
Internal tests on Thread samples have been scheduled for this PR as well, so please wait for results before merging

frkv:
Note that this PR is exploratory work. Likely there are still internal and external dependencies for legacy crypto toolbox configurations and legacy Mbed TLS APIs

krish2718 (commented):
For Wi-Fi parts.

krish2718 (approved):

maciejbaczmanski (approved):
OpenThread's side looks good

mswarowsky (approved):
Should run a crypto nightly on this as well, else looks fine for me

SebastianBoe:
I'm giving up on this.

We are getting much closer to be able to deprecate legacy mbedtls, and at that time it should be much easier to disable it.

PR #13872, opened 2024-02-05
Bluetooth: Controller: Define BT_CTLR_SDC_CONN_EVENT_EXTEND_DEFAULT and disable it for 802.15.4

rugeGerritsen:
Define CONFIG_BT_CTLR_SDC_CONN_EVENT_EXTEND_DEFAULT to allow
users to enable and disable connection event extension in a simpler
way.

We do not want to enable connection event extension when
802.15.4 protocols are enabled as we would like to reserve as much time
as possible for those protocols.

Disabling the setting has no practical consequence when connected
to a single Bluetooth device, but it improves performance
significantly when connected to two or more Bluetooth devices.

Ref: DRGN-21305

### DIFF
--- a/subsys/bluetooth/controller/Kconfig
+++ b/subsys/bluetooth/controller/Kconfig
@@ -143,10 +143,23 @@ config BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT
 	  The time set aside for connections on every connection interval in
 	  microseconds. The event length and the connection interval are the
 	  primary parameters for setting the throughput of a connection.
+	  When connection event extension is enabled, more time may be used.
 	  The event length may be set to a value that is shorter than the time needed
 	  for a single packet pair on a given PHY.
 	  In that case the controller will reserve time for receiving 27 bytes and transmitting
 	  the number of bytes configured with BT_CTLR_MIN_VAL_OF_MAX_ACL_TX_PAYLOAD_DEFAULT.
+
+config BT_CTLR_SDC_CONN_EVENT_EXTEND_DEFAULT
+	bool "Enable connection event extension by default"
+	default y
+	help
+	  When Extended Connection Events are disabled, the maximum connection event length is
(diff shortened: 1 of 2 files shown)


subsys/bluetooth/controller/Kconfig
```diff
@@ -143,11 +143,24 @@ config BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT
 	  The time set aside for connections on every connection interval in
 	  microseconds. The event length and the connection interval are the
 	  primary parameters for setting the throughput of a connection.
+	  When BT_CTLR_SDC_CONN_EVENT_EXTEND_DEFAULT is set, more time may be used.
```
Tschet1, line 146: sdc_hci_cmd_vs_conn_event_extend may be called by the application, as well.
I think it is not only about the BT_CTLR_SDC_CONN_EVENT_EXTEND_DEFAULT, but the question is if connecion event extension is enabled.

subsys/bluetooth/controller/Kconfig
```diff
@@ -143,11 +143,24 @@ config BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT
...
+	bool "Enable connection event extension by default"
+	default y if !NRF_802154_RADIO_DRIVER
+	help
+	  When Extended Connection Events are disabled, the maximum connection event length is
+	  configured with BT_CTLR_SDC_MAX_CONN_EVENT_LEN_DEFAULT.
+	  When Extended Connection Events are enabled, the controller
+	  will extend the connection event as much as possible, if
+	  - Either of the peers has more data to send.
+	    See also: Core v5.1, Vol 6, Part B, Section 4.5.6
+	  - There are no conflicts with other concurrent links.
```
Tschet1, line 162:
```suggestion
	  - There are no conflicts with other roles of equal or higher priority.
```

Tschet1 (changes requested):

Tschet1 (approved):

henrla (approved):

PR #13883, opened 2024-02-06
subsys: debug: etb_trace: Start trace and debug clock on idle exit

jtguggedal:
This patch fixes an issue where the trace and debug clock was not restarted when exiting idle state, which caused the ETB trace collection to hang indefinitely.
The issue could be observed when CONFIG_ETB_TRACE_LOW_POWER was enabled and after CONFIG_PM was no longer enabled for nRF targets.

Fixes CIA-1064


gmarull (approved):
ok as a temporary solution

DematteisGiacomo (approved):

PR #13885, opened 2024-02-06
Applications: nrf5340_audio: Use VS TX sync command when possible

rugeGerritsen:
The Read TX sync command is used to synchronize data transferred from the application to the network core. It is used to figure out if the data was sent to early or too late.

In this context, it is the assigned sync ref that matters, it is the sync ref that determines when a packet will be sent on air.

The HCI spec defined TX sync command returns a CIG/BIG reference anchor point. This may or may not be aligned with the corresponding sync ref. Therefore, we use a vendor specific command instead.


rugeGerritsen:
> I just discussed with Henrla. It would be practical if this VS command could directly return if we were too late for the next interval or not. We need to talk about the details.

This was discussed offline. It is more complex to make the command return if the packet was sent or not when also using FT>1, PTO != 0. So we have decided to keep it as is

applications/nrf5340_audio/src/bluetooth/bt_stream/bt_le_audio_tx/bt_le_audio_tx.c
```diff
@@ -127,6 +128,54 @@ static int iso_stream_send(uint8_t const *const data, size_t size, struct bt_bap
...
+	struct net_buf *rsp;
+	sdc_hci_cmd_vs_iso_read_tx_timestamp_t *cmd_read_tx_timestamp;
+	uint16_t conn_handle;
+
+	int err = bt_hci_get_conn_handle(stream->conn, &conn_handle);
+
+	if (err) {
+		LOG_ERR("Failed obtaining conn_handle (err:%d)", err);
+		return err;
+	}
```
alexsven, line 144: Nit:
```suggestion
	if (ret) {
		LOG_ERR("Failed obtaining conn_handle (err:%d)", ret);
		return ret;
	}
```
We use `ret` instead of `err`, and we usually declare it as the first variable in the function

alexsven (changes requested):

alexsven (approved):

applications/nrf5340_audio/src/bluetooth/bt_stream/bt_le_audio_tx/bt_le_audio_tx.c
alexsven: Did you change it back to using err?

rugeGerritsen:
This PR will be merged after the release

rugeGerritsen:
Rebased on main. No changes introduced

andvib (approved):

koffes:
@rugeGerritsen is this still DNM?

rugeGerritsen:
> @rugeGerritsen is this still DNM?

Yes. It is. It requires the use of https://github.com/zephyrproject-rtos/zephyr/pull/71247 to be able to obtain the connection handle for BIS

applications/nrf5340_audio/src/bluetooth/bt_stream/bt_le_audio_tx/bt_le_audio_tx.c
```diff
@@ -218,11 +258,29 @@ int bt_le_audio_tx_send(struct bt_bap_stream **bap_streams, struct le_audio_enco
...
-		/* Strictly, it is only required to call get_tx_sync on the first streaming channel
-		 * to get the timestamp which is sent to all other channels. However, to be able to
-		 * detect errors, this is called on each TX.
+		if (tx_info_arr[i].iso_conn_handle == HANDLE_INVALID) {
+			struct bt_bap_ep_info ep_info;
+
+			ret = bt_bap_ep_get_info(bap_streams[i]->ep, &ep_info);
+			if (ret) {
+				LOG_WRN("Unable to get info for ep");
+				return -EACCES;
```
koffes, line 267: Why return here? I may need a walkthrough of the latest addition.

koffes (approved):

PR #13886, opened 2024-02-06
Buffer Overflow fix

kapbh:
The samples are showing junk character in SSID because of the length of ssid is maximum(32 character) which leads to buffer overflow.


sachinthegreen (approved):
Please add Jira reference(s) in the commit log(s) as appropriate.

krish2718 (approved):

krga2022 (approved):
Tested with SSID of 32 characters and the display is as expected without any junk characters.

PR #13888, opened 2024-02-06
modules: memfault: Report unkown operator if modem doesn't give one 

jtguggedal:
Set the Memfault metric `ncs_lte_operator` to "Not reported" when
the modem does not provide an operator name to the application.

### DIFF
--- a/lib/modem_info/modem_info.c
+++ b/lib/modem_info/modem_info.c
@@ -974,8 +974,8 @@ int modem_info_get_operator(char *buf, size_t buf_size)
 				buf);
 
 	if (ret != 1) {
-		/* Warning instead of error because it is not always reported */
-		LOG_WRN("No valid operator");
+		/* Debug instead of error because it is not always reported */
+		LOG_DBG("No valid operator name found");
 		return map_nrf_modem_at_scanf_error(ret);
 	}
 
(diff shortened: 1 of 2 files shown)


lib/modem_info/modem_info.c
```diff
@@ -975,7 +975,7 @@ int modem_info_get_operator(char *buf, size_t buf_size)
 
 	if (ret != 1) {
 		/* Warning instead of error because it is not always reported */
```
jhn-nordic, line 977: It is now DBG

jhn-nordic (approved):

DematteisGiacomo (approved):

modules/memfault-firmware-sdk/memfault_lte_metrics.c
```diff
@@ -221,7 +221,12 @@ void memfault_lte_metrics_update(void)
 
 	err = modem_info_get_operator(operator_name, sizeof(operator_name));
 	if (err != 0) {
-		LOG_WRN("Network operator collection failed, error: %d", err);
+		LOG_DBG("Network operator name was not reported by the modem");
+
+		err = MEMFAULT_METRIC_SET_STRING(ncs_lte_operator, "Not reported");
```
trantanen, line 226: "Not available" would probably be a little more "standard" in many context but feel free to choose what best fits into this context.

jtguggedal: That's true, changed it to "Not available". 
FYI @gminn

PR #13889, opened 2024-02-06
drivers: wifi: Add VIF locking to public APIs

krish2718:
To fix crashes during interface de-init, add a lock + NULL check before de-referencing FMAC handle and invoking FMAC APIs. As this is expected during de-init only debug logs are used.

Ideally a locking mechanism like RCU is perfect for this and low overhead, but we don't have that in Zephyr.

Fixes SHEL-2436.

### DIFF
--- a/drivers/wifi/nrf700x/src/wifi_mgmt.c
+++ b/drivers/wifi/nrf700x/src/wifi_mgmt.c
@@ -36,20 +36,26 @@ int nrf_wifi_set_power_save(const struct device *dev,
 
 	if (!dev || !params) {
 		LOG_ERR("%s: dev or params is NULL", __func__);
-		goto out;
+		return ret;
 	}
 
 	vif_ctx_zep = dev->data;
 
 	if (!vif_ctx_zep) {
 		LOG_ERR("%s: vif_ctx_zep is NULL", __func__);
-		goto out;
+		return ret;
 	}
 
 	rpu_ctx_zep = vif_ctx_zep->rpu_ctx_zep;
 
 	if (!rpu_ctx_zep) {
 		LOG_ERR("%s: rpu_ctx_zep is NULL", __func__);
+		return ret;
+	}
+
+	k_mutex_lock(&vif_ctx_zep->vif_lock, K_FOREVER);
+	if (!rpu_ctx_zep->rpu_ctx) {
+		LOG_DBG("%s: RPU context not initialized", __func__);
 		goto out;
 	}
 
@@ -123,6 +129,7 @@ int nrf_wifi_set_power_save(const struct device *dev,
 
 	ret = 0;
 out:
+	k_mutex_unlock(&vif_ctx_zep->vif_lock);
 	return ret;
 }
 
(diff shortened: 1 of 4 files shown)


drivers/wifi/nrf700x/src/wifi_mgmt.c
```diff
@@ -158,6 +165,13 @@ int nrf_wifi_get_power_save_config(const struct device *dev,
 	}
 
 	rpu_ctx_zep = vif_ctx_zep->rpu_ctx_zep;
+
+	k_mutex_lock(&vif_ctx_zep->vif_lock, K_FOREVER);
+	if (!rpu_ctx_zep->rpu_ctx) {
+		LOG_DBG("%s: RPU context not initialized", __func__);
+		goto out;
+	}
+
```
VivekUppunda, line 174: I think the above check should be moved below the check for rpu_ctx_zep. If rpu_ctx_zep is not initialized, this will cause a NULL dereference

drivers/wifi/nrf700x/src/wifi_mgmt.c
```diff
@@ -708,6 +745,13 @@ int nrf_wifi_mode(const struct device *dev,
 	}
 
 	rpu_ctx_zep = vif_ctx_zep->rpu_ctx_zep;
+
+	k_mutex_lock(&vif_ctx_zep->vif_lock, K_FOREVER);
+	if (!rpu_ctx_zep->rpu_ctx) {
+		LOG_DBG("%s: RPU context not initialized", __func__);
```
VivekUppunda, line 751: Should these new log messages be LOG_ERR or LOG_DBG ? I lean towards LOG_DBG as it is not an error case but the format used elsewhere in this document uses LOG_ERR for the same type of cases

drivers/wifi/nrf700x/src/wifi_mgmt.c
```diff
@@ -774,6 +819,13 @@ int nrf_wifi_channel(const struct device *dev,
 	}
 
 	rpu_ctx_zep = vif_ctx_zep->rpu_ctx_zep;
+
```
VivekUppunda, line 822: i think i have failed to sanitize rpu_ctx_zep here i guess as well. Can you kindly add it?

rado17 (approved):

drivers/wifi/nrf700x/src/wpa_supp_if.c
```diff
@@ -2370,6 +2555,7 @@ int nrf_wifi_wpa_supp_sta_set_flags(void *if_priv, const u8 *addr,
 	ret = 0;
 
 out:
+	k_mutex_unlock(&vif_ctx_zep->vif_lock);
```
rlubos, line 2558: I can't comment at that particular line, but at line 2529 you jump to `out` w/o locking mutex.

drivers/wifi/nrf700x/src/wpa_supp_if.c
```diff
@@ -2333,6 +2510,7 @@ int nrf_wifi_wpa_supp_sta_remove(void *if_priv, const u8 *addr)
...
```
rlubos, line 2513: Ditto, at line 2490. It seems to be the case for other functions as well, it seems that the argument check needs to be revised (maybe just return early?).

rlubos (approved):

PR #13891, opened 2024-02-06
Bluetooth: Mesh: Make prop_format_get public

HaavardRei:
Moves the `prop_format_get` function to public light control header to allow access to the sensor formats for
use with the new sensor API.

### DIFF
--- a/include/bluetooth/mesh/light_ctrl.h
+++ b/include/bluetooth/mesh/light_ctrl.h
@@ -14,6 +14,7 @@
 #define BT_MESH_LIGHT_CTRL_H__
 
 #include <bluetooth/mesh/properties.h>
+#include <bluetooth/mesh/sensor_types.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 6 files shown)


tests/subsys/bluetooth/mesh/light_ctrl/src/main.c
```diff
@@ -10,7 +10,6 @@
 #include <bluetooth/mesh/models.h>
 #include <bluetooth/mesh/light_ctrl_srv.h>
 #include <bluetooth/mesh/light_ctrl_reg_spec.h>
-#include "light_ctrl_internal.h"
```
ludvigsj, line 13: Was this just never used? Or is there a `prop_format_get` call I can't find that you changed as well?

HaavardRei: Never used! Or it was used at some time then the functionality was removed.

HaavardRei:
Added DNM until https://github.com/nrfconnect/sdk-nrf/pull/13793 is merged. (Util improvements commit is needed)

include/bluetooth/mesh/light_ctrl.h
```diff
@@ -14,6 +14,7 @@
 #define BT_MESH_LIGHT_CTRL_H__
 
 #include <bluetooth/mesh/properties.h>
+#include <bluetooth/mesh/sensor.h>
```
Balaklaka, line 17: the include of this file probably need to be replaced by include of sensor_types.h. The define in sensor.h will probably be duplicate of what will be in sensor_types.h.

include/bluetooth/mesh/sensor.h
```diff
@@ -1160,6 +1160,116 @@ const struct bt_mesh_sensor_type *bt_mesh_sensor_type_get(uint16_t id);
...
 bt_mesh_sensor_column_format_get(const struct bt_mesh_sensor_type *type);
 
+/**
+ * @defgroup BT_MESH_SENSOR_FORMATS Sensor channel formats
+ * @brief All available sensor channel formats in the mesh device properties
+ * Specification.
+ * @{
+ */
+
+/* Percentage formats */
```
PavelVPV, line 1170: One comment I'll post now anyway. If we make these types available in the API, they need to have a description. Single comment for a set of formats can't be parsed by Doxygen (unless I'm mistaken). Though these formats can be still grouped in Doxygen.

PavelVPV: Grouped by format I mean (`percentage formats`, environmental formats`, etc).

ludvigsj: This will be replaced by my commit. I already made these public and documented them.

PavelVPV: Right, sorry, I forgot this.

ludvigsj:
Sample test is passing with changes in separate system test PR. Removing DNM.

tests/subsys/bluetooth/mesh/light_ctrl/src/main.c
HaavardRei: Found that light_ctrl_internal was included to get subsys/.../sensor.h. Changed to directly including sensor.

include/bluetooth/mesh/light_ctrl.h
```diff
@@ -183,6 +184,44 @@ enum bt_mesh_light_ctrl_coeff {
...
 };
 
+/** @brief Get the sensor format for the given Light Lightness Control property ID.
+ *
+ *  @param[in] id Light Lightness Control property or coefficient ID.
+ *
+ *  @return The associated sensor format, or NULL if the ID is unknown.
+ */
+static inline const struct bt_mesh_sensor_format *
+bt_mesh_lc_prop_format_get(int id)
```
ludvigsj, line 194: `uint16_t`? Makes it more obvious that this is the Property ID from the spec/Device Properties document.

include/bluetooth/mesh/light_ctrl.h
```diff
@@ -183,6 +184,44 @@ enum bt_mesh_light_ctrl_coeff {
 	BT_MESH_LIGHT_CTRL_COEFF_KPU = BT_MESH_PROP_ID_LIGHT_CTRL_REG_KPU,
 };
 
+/** @brief Get the sensor format for the given Light Lightness Control property ID.
+ *
+ *  @param[in] id Light Lightness Control property or coefficient ID.
```
ludvigsj, line 189: The coefficients are properties, no need to doubly specify. (As in, the are all listed under "6.2.4 Light LC Property states", no distinction is made in the spec. I see no reason to make a somewhat arbitrary distinction here)

ludvigsj (changes requested):

PavelVPV (approved):

PR #13893, opened 2024-02-06
secure_storage: Store data in one file

vili-nordic:
Instead of spreading nonce, size, flags and data over several files these are now stored in one blob. This should solve issues related to cleanup when one of the storage operation fails.

Ref: NCSDK-25493

test-sdk-nrf: test-pr-13893


frkv (approved):
LGTM, pending deep testing

magnev (approved):

PR #13894, opened 2024-02-06
lib: location: Fallback event and more data details

trantanen:
Adding LOCATION_EVT_FALLBACK to be sent when a methods times out or fails and a new method is tried.
Adding `satellites_used` and `search_time` for GNSS details. Also creating Wi-Fi and cellular details with number of APs and cells.
Adding a new method type `LOCATION_METHOD_WIFI_CELLULAR` that cannot be used in the method list for location request but may occur in the events.

Jira: NCSDK-25618

### DIFF
--- a/doc/nrf/libraries/modem/location.rst
+++ b/doc/nrf/libraries/modem/location.rst
@@ -122,6 +122,8 @@ The following diagram shows successful cellular positioning.
 The following diagram depicts a failure to find the GNSS fix, followed by a fallback to cloud positioning.
 Since the Wi-Fi and cellular positioning methods are one after another, they are combined to a single cloud positioning request.
 Both Wi-Fi APs and LTE cells are given to the application with a single :c:enum:`LOCATION_EVT_CLOUD_LOCATION_EXT_REQUEST` event.
+The :c:enum:`LOCATION_EVT_STARTED` and the :c:enum:`LOCATION_EVT_FALLBACK` events are sent,
+if the :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
 
 .. msc::
   hscale="1.3";
@@ -132,6 +134,7 @@ Both Wi-Fi APs and LTE cells are given to the application with a single :c:enum:
 
   Application => Loclib [label="location_request(&config)\nmethod list: GNSS, Wi-Fi, Cellular"];
   |||;
+  Application << Loclib [label="LOCATION_EVT_STARTED"];
   Application << Loclib [label="LOCATION_EVT_GNSS_ASSISTANCE_REQUEST"];
   Application => Cloud [label="Request A-GNSS data"];
   Application << Cloud [label="A-GNSS data"];
@@ -141,6 +144,7 @@ Both Wi-Fi APs and LTE cells are given to the application with a single :c:enum:
   Loclib rbox Loclib [label="GNSS timeout occurs"];
 
   Loclib rbox Loclib [label="Fallback to cloud positioning"];
+  Application << Loclib [label="LOCATION_EVT_FALLBACK"];
   Loclib rbox Loclib [label="Scan Wi-Fi and LTE networks"];
   |||;
   Application << Loclib [label="LOCATION_EVT_CLOUD_LOCATION_EXT_REQUEST"];
--- a/doc/nrf/libraries/modem/location.rst
+++ b/doc/nrf/libraries/modem/location.rst
@@ -60,6 +60,14 @@ The supported location methods are as follows:
 
 The ``cloud location`` method handles the location methods (cellular and Wi-Fi positioning)
 that scan for technology-specific information and sends it over to the cloud service for location resolution.
+If the following conditions are met, Wi-Fi and cellular scan results are combined into a single cloud request:
+
+* Methods are one after the other in the location request method list.
+* Location request mode is :c:enum:`LOCATION_REQ_MODE_FALLBACK`.
+* Requested cloud service for Wi-Fi and cellular is the same.
+
+A special :c:enum:`LOCATION_METHOD_WIFI_CELLULAR` method can appear within the :c:struct:`location_event_data` structure,
+but it cannot be added into the location configuration passed to the :c:func:`location_request` function.
 
 The default priority order of location methods is GNSS positioning, Wi-Fi positioning and Cellular positioning.
 If any of these methods are disabled, the method is simply omitted from the list.
--- a/doc/nrf/libraries/modem/location.rst
+++ b/doc/nrf/libraries/modem/location.rst
@@ -292,6 +292,10 @@ when :c:func:`location_config_defaults_set` function is called:
 * :kconfig:option:`CONFIG_LOCATION_REQUEST_DEFAULT_CELLULAR_CELL_COUNT`
 * :kconfig:option:`CONFIG_LOCATION_REQUEST_DEFAULT_WIFI_TIMEOUT`
 
+The following option adds more details to the :c:struct:`location_event_data` structure:
+
+* :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS`
+
 Usage
 *****
 
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -754,6 +754,7 @@ Modem libraries
     * The :c:enumerator:`LOCATION_EVT_STARTED` event to indicate that location request has been started.
       This is for metrics collection purposes and sent only if the :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
     * Support for multiple event handlers.
+    * Additional location data details into the :c:struct:`location_data_details` structure hierarchy.
 
   * Updated:
 
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -756,6 +756,8 @@ Modem libraries
       These are for metrics collection purposes and sent only if the :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
     * Support for multiple event handlers.
     * Additional location data details into the :c:struct:`location_data_details` structure hierarchy.
+    * The :c:enumerator:`LOCATION_METHOD_WIFI_CELLULAR` method that cannot be added into the location configuration passed to the :c:func:`location_request()` function,
+      but may occur within the :c:struct:`location_event_data` structure.
 
   * Updated:
 
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -751,8 +751,9 @@ Modem libraries
 
   * Added:
 
-    * The :c:enumerator:`LOCATION_EVT_STARTED` event to indicate that location request has been started.
-      This is for metrics collection purposes and sent only if the :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
+    * The :c:enumerator:`LOCATION_EVT_STARTED` event to indicate that a location request has been started and
+      the :c:enumerator:`LOCATION_EVT_FALLBACK` event to indicate that a fallback from one location method to another has occurred.
+      These are for metrics collection purposes and sent only if the :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
     * Support for multiple event handlers.
     * Additional location data details into the :c:struct:`location_data_details` structure hierarchy.
 
--- a/include/modem/location.h
+++ b/include/modem/location.h
@@ -43,6 +43,19 @@ enum location_method {
 	LOCATION_METHOD_GNSS,
 	/** Wi-Fi positioning. */
 	LOCATION_METHOD_WIFI,
+	/**
+	 * Wi-Fi and cellular positioning combined into one cloud request.
+	 *
+	 * Cannot be used in the method list passed to location_request(),
+	 * but this method can appear in the location event (@ref location_event_data.id)
+	 * to indicate that Wi-Fi and cellular positioning methods have been combined
+	 * into a single cloud location method.
+	 *
+	 * If the combined method is desired, set Wi-Fi and cellular methods next to each other
+	 * in the method list passed to location_request(). This also happens by default if
+	 * NULL configuration is passed to location_request().
+	 */
+	LOCATION_METHOD_WIFI_CELLULAR,
 };
 
 /** Location acquisition mode. */
@@ -204,8 +217,8 @@ struct location_data_details_wifi {
  *
  * Only one of the child structures is filled most of the time depending on the method
  * found in @ref location_event_data.method member of the parent structure.
- * There are cases when cellular and Wi-Fi data are combined into a single cloud request
- * in which case data details for both methods are filled.
+ * For a combined method @ref LOCATION_METHOD_WIFI_CELLULAR both @ref cellular and
+ * @ref wifi are filled.
  */
 struct location_data_details {
 	/**
@@ -285,13 +298,7 @@ struct location_data_cloud {
 struct location_event_data {
 	/** Event ID. */
 	enum location_event_id id;
-	/**
-	 * Used location method.
-	 *
-	 * When cellular and Wi-Fi positioning are used and they are combined into a single
-	 * cloud request by the library, the method is not known so there is some uncertainty
-	 * on the reported location method.
-	 */
+	/** Used location method. */
 	enum location_method method;
 
 	/** Event specific data. */
--- a/include/modem/location.h
+++ b/include/modem/location.h
@@ -161,14 +161,64 @@ struct location_datetime {
 struct location_data_details_gnss {
 	/** Number of satellites tracked at the time of event. */
 	uint8_t satellites_tracked;
+	/** Number of satellites used at the time of event. */
+	uint8_t satellites_used;
 	/** PVT data. */
 	struct nrf_modem_gnss_pvt_data_frame pvt_data;
+	/**
+	 * Elapsed GNSS time in milliseconds.
+	 *
+	 * This is the time since last start of GNSS operation until the fix or timeout
+	 * including any time LTE has blocked GNSS.
+	 *
+	 * @ref pvt_data member has ``execution_time``, which indicates cumulative GNSS
+	 * execution time since last start excluding any time LTE has blocked GNSS.
+	 */
+	uint32_t elapsed_time_gnss;
 };
 
-/** Location details. */
+/** Location details for cellular. */
+struct location_data_details_cellular {
+	/** Number of neighbor cells. */
+	uint8_t ncells_count;
+	/** Number of GCI (surrounding) cells. */
+	uint8_t gci_cells_count;
+};
+
+/** Location details for Wi-Fi. */
+struct location_data_details_wifi {
+	/** Number of Wi-Fi APs. */
+	uint16_t ap_count;
+};
+
+/**
+ * Location details.
+ *
+ * Only one of the child structures is filled most of the time depending on the method
+ * found in @ref location_event_data.method member of the parent structure.
+ * There are cases when cellular and Wi-Fi data are combined into a single cloud request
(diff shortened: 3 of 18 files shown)


include/modem/location.h
```diff
@@ -161,14 +161,56 @@ struct location_datetime {
...
+	 * GNSS search time in milliseconds.
+	 *
+	 * This is the time since last start of GNSS operation to the fix or timeout
+	 * including any time LTE has blocked GNSS.
+	 *
+	 * @ref pvt_data has nrf_modem_gnss_pvt_data_frame.execution_time which indicates
+	 * cumulative GNSS execution time since last start excluding any time LTE has blocked
+	 * GNSS.
+	 */
+	uint32_t search_time;
```
trantanen, line 178: @tokangas, @hiltunent, how should we name this? TTF is not good because we use it also for search time when timeout occurs.

include/modem/location.h
```diff
@@ -161,14 +161,56 @@ struct location_datetime {
...
+
+/**
+ * Location details.
+ *
+ * Only one of the child structures is filled most of the time depending on the method
+ * found in @ref location_event_data.method member of the parent structure.
+ * There are cases when cellular and Wi-Fi data are combined into a single cloud request
+ * in which case data details for both methods are filled.
+ */
 struct location_data_details {
```
trantanen, line 203: There is no time member at the moment. `location_data_details_gnss.search_time` exists which is not the whole time method implementation took but only from `nrf_modem_gnss_start()` function call.
I don't have strong opinion whether other time values are needed as memfault will calculate some session time values based on the events that we send.

include/modem/location.h
```diff
@@ -161,14 +161,56 @@ struct location_datetime {
...
+	uint16_t ap_count;
+};
+
+/**
+ * Location details.
+ *
+ * Only one of the child structures is filled most of the time depending on the method
+ * found in @ref location_event_data.method member of the parent structure.
+ * There are cases when cellular and Wi-Fi data are combined into a single cloud request
+ * in which case data details for both methods are filled.
```
tokangas, line 201: Is there any way to determine when cellular and Wi-Fi have been combined, I guess not?

include/modem/location.h
trantanen: "search" is not particularly good but I had to put something 😄
Somewhere "used_time" was used which I feel is worse.
"fix_time" or "ttf" cannot be used as this is for timeout cases too.
I hope @hiltunent will help us 😄

include/modem/location.h
trantanen: Not from the event data, except knowing the config you used for `location_request()`, i.e., if cellular and Wi-Fi were next to each other in the method list. So the app calling `location_request()` knows it but Memfault monitoring only the events won't know it with the current API.

trantanen: There's now special treatment for the method fields of `LOCATION_EVT_STARTED` and `LOCATION_EVT_FALLBACK` when combined method is used.
`LOCATION_EVT_STARTED` has it in the current NCS/main and that's a bug I didn't realize earlier.

Please check the way those evens are composed, and understand that when both Wi-Fi and cellular are used, we do report just one of them in the events as follows:

- `LOCATION_EVT_STARTED`: The first method in the given config
- `LOCATION_EVT_LOCATION`: If accuracy > 100m, use cellular. Otherwise Wi-Fi.
- `LOCATION_EVT_TIMEOUT`/`LOCATION_EVT_ERROR`: Wi-Fi
- `LOCATION_EVT_FALLBACK`: Whichever is the first one on the list.

Note: Above applies when we would set method to `LOCATION_METHOD_INTERNAL_WIFI_CELLULAR`, which is not in public API. It doesn't apply if methods were not combined due to conditions mentioned in the API.

lib/location/location_core.c
```diff
@@ -375,7 +375,10 @@ static int location_core_location_get_pos(void)
 	if (IS_ENABLED(CONFIG_LOCATION_DATA_DETAILS)) {
 		struct location_event_data request_started = {
 			.id = LOCATION_EVT_STARTED,
-			.method = requested_method
+			/* Use the first method from the config to avoid putting internal conbined
```
tokangas, line 378:
```suggestion
			/* Use the first method from the config to avoid putting internal combined
```

samples/cellular/modem_shell/src/location/location_shell.c
```diff
@@ -325,10 +366,39 @@ void location_ctrl_event_handler(const struct location_event_data *event_data)
...
+			location_method_str(event_data->fallback.next_method),
+			event_data->fallback.next_method);
+		mosh_print(
+			"  cause: %s",
+			(event_data->fallback.cause == LOCATION_EVT_TIMEOUT) ?
+			 "timeout" :
+			 (event_data->fallback.cause == LOCATION_EVT_ERROR) ?
+			  "error" :
+			  "unknown");
+#if defined(CONFIG_LOCATION_DATA_DETAILS)
```
tokangas, line 399: `LOCATION_EVT_STARTED` and `LOCATION_EVT_FALLBACK` events can only be received when `CONFIG_LOCATION_DATA_DETAILS` is enabled. Cases for those events could be flagged to save memory when data details are not enabled. If the events would be unexpectedly received, a warning would be printed anyway.

lib/location/location_core.c
```diff
@@ -617,6 +620,42 @@ static void location_core_event_details_get(struct location_event_data *event)
...
 
+static void location_core_event_cb_fallback(
+	enum location_method current_method,
+	enum location_method next_method,
+	enum location_event_id current_event,
+	struct location_data_details *details)
+{
+#if defined(CONFIG_LOCATION_DATA_DETAILS)
+	if (current_method == LOCATION_METHOD_INTERNAL_WIFI_CELLULAR) {
+		/* When fallback happens, current method is always the 1st one whenever
```
tokangas, line 631: As already discussed, I had hard time understanding why this works. 😄 It could be only me, but you could consider explaining in even more detail why indexes 0 and 1 can be used here.

And maybe we should consider limiting the maximum value of `CONFIG_LOCATION_METHODS_LIST_SIZE` to 3. There should be no reason to increase it past 3, but increasing it could break this implementation.

include/modem/location.h
trantanen: I've added a commit on how it would look like if there was a combined LOCATION_METHOD_WIFI_CELLULAR method. This simplifies many things in the implementation. We can discuss and see if we keep this or not.
You cannot give this method in the method list for `location_request()`.

lib/location/location_core.c
```diff
@@ -741,6 +741,21 @@ static void location_core_event_cb_fn(struct k_work *work)
 				 * also for failure events
 				 */
 				location_utils_event_dispatch(&loc_req_info.current_event_data);
+#if defined(CONFIG_LOCATION_DATA_DETAILS)
+			} else {
+				struct location_event_data fallback = {
+					.id = LOCATION_EVT_FALLBACK,
+					.method = loc_req_info.current_method,
```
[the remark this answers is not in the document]
trantanen: See https://github.com/nrfconnect/sdk-nrf/pull/13894#discussion_r1485754327

include/modem/location.h
trantanen: I've added `method_search_time` here

samples/cellular/modem_shell/src/location/location_shell.c
```diff
@@ -325,11 +370,40 @@ void location_ctrl_event_handler(const struct location_event_data *event_data)
...
+		mosh_print(
+			"  new method: %s (%d)",
+			location_method_str(event_data->fallback.next_method),
+			event_data->fallback.next_method);
+		mosh_print(
+			"  cause: %s",
+			(event_data->fallback.cause == LOCATION_EVT_TIMEOUT) ?
+			 "timeout" :
+			 (event_data->fallback.cause == LOCATION_EVT_ERROR) ?
+			  "error" :
```
tokangas, line 402: `timeout` and `error` could be on the same line with the condition, but I leave this up to you.
```
			(event_data->fallback.cause == LOCATION_EVT_TIMEOUT) ? "timeout" :
			(event_data->fallback.cause == LOCATION_EVT_ERROR) ? "error" :
			"unknown");
```

tokangas (commented):
Looks good to me. Only a couple of minor comments. Additionally, we need to agree the naming of the different time fields.

samples/cellular/modem_shell/src/location/location_shell.c
trantanen: I don't know how these are supposed to be intended and in some sense your proposal is "a bit wrong" as there is if-else structure without indentation. But I agree that's more readable so I've changed into it.

include/modem/location.h
jhirsi: I'd vote for elapsed_method_time and elapsed_gnss_time

jhirsi (approved):

include/modem/location.h
trantanen: Changed `gnss_search_time` to `elapsed_time_gnss` and `method_search_time` to `elapsed_time_method`.

doc/nrf/libraries/modem/location.rst
```diff
@@ -60,6 +60,14 @@ The supported location methods are as follows:
 
 The ``cloud location`` method handles the location methods (cellular and Wi-Fi positioning)
 that scan for technology-specific information and sends it over to the cloud service for location resolution.
+If the following conditions are met, Wi-Fi and cellular scan results are combined into single cloud request:
```
peknis, line 63:
```suggestion
If the following conditions are met, Wi-Fi and cellular scan results are combined into a single cloud request:
```

doc/nrf/libraries/modem/location.rst
```diff
@@ -60,6 +60,14 @@ The supported location methods are as follows:
...
+
+* Methods are one after the other in location request method list
+* Location request mode is :c:enum:`LOCATION_REQ_MODE_FALLBACK`
+* Requested cloud service for Wi-Fi and cellular is the same
```
peknis, line 67:
```suggestion
* Methods are one after the other in the location request method list.
* Location request mode is :c:enum:`LOCATION_REQ_MODE_FALLBACK`.
* Requested cloud service for Wi-Fi and cellular is the same.
```

doc/nrf/libraries/modem/location.rst
```diff
@@ -60,6 +60,14 @@ The supported location methods are as follows:
...
+
+A special :c:enum:`LOCATION_METHOD_WIFI_CELLULAR` method can appear within the :c:struct:`location_event_data` structure
+but it cannot be added into the location configuration given to :c:func:`location_request` function.
```
peknis, line 70:
```suggestion
A special :c:enum:`LOCATION_METHOD_WIFI_CELLULAR` method can appear within the :c:struct:`location_event_data` structure,
but it cannot be added into the location configuration passed to the :c:func:`location_request` function.
```

doc/nrf/libraries/modem/location.rst
```diff
@@ -122,6 +130,8 @@ The following diagram shows successful cellular positioning.
 The following diagram depicts a failure to find the GNSS fix, followed by a fallback to cloud positioning.
 Since the Wi-Fi and cellular positioning methods are one after another, they are combined to a single cloud positioning request.
 Both Wi-Fi APs and LTE cells are given to the application with a single :c:enum:`LOCATION_EVT_CLOUD_LOCATION_EXT_REQUEST` event.
+The :c:enum:`LOCATION_EVT_STARTED` and the :c:enum:`LOCATION_EVT_FALLBACK` events are sent
+if :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
```
peknis, line 134:
```suggestion
The :c:enum:`LOCATION_EVT_STARTED` and the :c:enum:`LOCATION_EVT_FALLBACK` events are sent,
if the :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -751,9 +751,13 @@ Modem libraries
 
   * Added:
 
-    * The :c:enumerator:`LOCATION_EVT_STARTED` event to indicate that location request has been started.
-      This is for metrics collection purposes and sent only if the :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
+    * The :c:enumerator:`LOCATION_EVT_STARTED` event to indicate that location request has been started, and
+      the :c:enumerator:`LOCATION_EVT_FALLBACK` event to indicate that a fallback from one location method to another has occurred.
```
peknis, line 755:
```suggestion
    * The :c:enumerator:`LOCATION_EVT_STARTED` event to indicate that a location request has been started and
      the :c:enumerator:`LOCATION_EVT_FALLBACK` event to indicate that a fallback from one location method to another has occurred.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -751,9 +751,13 @@ Modem libraries
...
+      These are for metrics collection purposes and sent only if the :kconfig:option:`CONFIG_LOCATION_DATA_DETAILS` Kconfig option is set.
     * Support for multiple event handlers.
+    * Additional location data details into the :c:struct:`location_data_details` structure hierarchy.
+    * The :c:enumerator:`LOCATION_METHOD_WIFI_CELLULAR` method, which cannot be added into the location configuration given to location_request() function,
+      but may occur within the :c:struct:`location_event_data` structure.
```
peknis, line 760:
```suggestion
    * The :c:enumerator:`LOCATION_METHOD_WIFI_CELLULAR` method that cannot be added into the location configuration passed to the :c:func:`location_request()` function,
      but may occur within the :c:struct:`location_event_data` structure.
```

include/modem/location.h
```diff
@@ -43,6 +43,19 @@ enum location_method {
 	LOCATION_METHOD_GNSS,
 	/** Wi-Fi positioning. */
 	LOCATION_METHOD_WIFI,
+	/**
+	 * Wi-Fi and cellular positioning combined into one cloud request.
+	 *
+	 * Cannot be used in method list given to location_request(),
```
peknis, line 49:
```suggestion
	 * Cannot be used in the method list passed to location_request(),
```

include/modem/location.h
```diff
@@ -43,6 +43,19 @@ enum location_method {
...
+	 * but this method can appear in the location event (@ref location_event_data.id)
+	 * to indicate that Wi-Fi and cellular positioning methods have been combined
+	 * into a single cloud location method.
+	 *
+	 * If combined method is desired, set Wi-Fi and cellular methods next to each other
+	 * in the method list given to location_request(). This also happens by default if
+	 * NULL configuration is given to location_request().
```
peknis, line 56:
```suggestion
	 * If the combined method is desired, set Wi-Fi and cellular methods next to each other
	 * in the method list passed to location_request(). This also happens by default if
	 * NULL configuration is passed to location_request().
```

include/modem/location.h
```diff
@@ -94,6 +107,13 @@ enum location_event_id {
 	 * This event is only sent if @kconfig{CONFIG_LOCATION_DATA_DETAILS} is set.
 	 */
 	LOCATION_EVT_STARTED,
+	/**
+	 * A fallback has occurred from one method to another,
+	 * and positioning procedure will continue.
```
peknis, line 112:
```suggestion
	 * A fallback from one method to another has occurred,
	 * and the positioning procedure continues.
```

include/modem/location.h
```diff
@@ -161,14 +181,64 @@ struct location_datetime {
...
+ * Only one of the child structures is filled most of the time depending on the method
+ * found in @ref location_event_data.method member of the parent structure.
+ * For a combined method @ref LOCATION_METHOD_WIFI_CELLULAR both @ref cellular and
+ * @ref wifi are filled.
+ */
 struct location_data_details {
+	/**
+	 * Elapsed method time in milliseconds.
+	 *
+	 * This is the time from method start until it finishes and includes any time
```
peknis, line 226:
```suggestion
	 * This is the time from method start until it completes and includes any time
```

include/modem/location.h
```diff
@@ -195,6 +265,20 @@ struct location_data_error {
 	/** Data details at the time of error. */
 	struct location_data_details details;
 };
+
+/** Location fallback information. */
+struct location_data_fallback {
+	/** New location method, which is tried next. */
```
peknis, line 271:
```suggestion
	/** New location method that is tried next. */
```

include/modem/location.h
```diff
@@ -161,14 +181,64 @@ struct location_datetime {
...
+	uint8_t satellites_used;
 	/** PVT data. */
 	struct nrf_modem_gnss_pvt_data_frame pvt_data;
+	/**
+	 * Elapsed GNSS time in milliseconds.
+	 *
+	 * This is the time since last start of GNSS operation until the fix or timeout
+	 * including any time LTE has blocked GNSS.
+	 *
+	 * @ref pvt_data member has ``execution_time`` which indicates cumulative GNSS
```
peknis, line 194:
```suggestion
	 * @ref pvt_data member has ``execution_time``, which indicates cumulative GNSS
```

peknis (changes requested):

peknis (approved):

lib/location/location_core.c
```diff
@@ -281,6 +281,10 @@ int location_core_validate_params(const struct location_config *config)
 	}
 
 	for (int i = 0; i < config->methods_count; i++) {
+		if (config->methods[i].method == LOCATION_METHOD_WIFI_CELLULAR) {
```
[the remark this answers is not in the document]
trantanen: Method bitmask was something I thought but I admit that not for too long. Maybe I should have. I'm probably more reluctant to breaking changes than many others which is why I didn't think about it seriously enough. If I would do this from scratch, bitmask would be an obvious choice.

Bitmask has also its drawbacks as API user has some uncertainty which combinations can be used in input (method configuration) and output (event structure).

@tokangas What do you think?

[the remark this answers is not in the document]
tokangas: It did also cross my mind when reviewing the changes. It would have some benefits and it would allow flexibility in the future if there are more similar combined methods. On the other hand, there would be some ambiguity which combinations are valid.

Probably it wouldn't be a problem on the method configuration side, because even if the method IDs would be different, the overall structure of the configuration would remain the same. So, only one method bit would be allowed in one `location_method_config` and combining methods would still happen internally, right? And of course we could still use an enum in input and a bitmask in output, IDs for each method would just have to be declared twice.

With the current approach it's obvious which method combination(s) are valid, I guess that's the main benefit. I don't have a very strong opinion either way.

maxd-nordic (approved):

lib/location/location_core.c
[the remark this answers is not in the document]
trantanen: After discussions today, we'd probably keep the enum. If we ever add a third method that could be combined to Wi-Fi and cellular, we'd need to move over to bitmask. However, we don't see the need for that in even in the next couple of years, but who knows...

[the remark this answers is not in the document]
trantanen: Any further comments @glarsennordic?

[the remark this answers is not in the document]
plskeggs: @trantanen he might not respond anytime soon -- he is out sick.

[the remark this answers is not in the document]
trantanen: Ok, I'll request this PR to be merged so we get forward. You can add further comments here and we can continue discussion, and create another PR if necessary.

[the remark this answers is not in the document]
glarsennordic: Yep, sorry for my lack of response, thanks Pete for letting them know what happened to me.

Fair point, enum seems just fine for now. If you ever do add that third event, I would say this:

A third option would be to still use an enum for specifying desired request modes, but then use a separate set of bitmask definitions for specifying event sources, if you really want to avoid the bitmask for specifying request modes. Though IMO using a bitmask for everythign and just validating incoming bitmasks probably makes the most sense, yeah.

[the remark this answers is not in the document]
trantanen: > Though IMO using a bitmask for everythign and just validating incoming bitmasks probably makes the most sense, yeah.

Thinking only about the method specification, bitmask for the request mode would fine.
However, we need to take the method specific configuration into account and that will become a bit more unintuitive, or require changes to the structure that won't be perfect. So you'd need to make a compromise either in the configuration structures or not using the bitmask.
But this can be re-evaluated if there is a new method.

PR #13899, opened 2024-02-07
Bluetooth: Mesh: Fix threshold value overflow

PavelVPV:
After removing `HAS_UNDEFINED` flag from `percentage_delta_trigger` sensor format, MMDL/SR/SNRS/BV-09-C PTS test started failing, because the sensor server started publishing with the fast cadence instead of the normal cadence.

The investigation showed that the multiplication of the threshold and the previous sensor state could cause int64_t overflow. This is because both are scaled preliminary by the `SENSOR_MILL` macro. Due to the overflow, the `bt_mesh_sensor_delta_threshold` function was always returning `true`, which made the sensor incorrectly use the fast cadence.

This change does the multiplication of both parts of the `sensor_value` separately before scaling it by the `SENSOR_MILL` macro to avoid the overflow.

### DIFF
--- a/subsys/bluetooth/mesh/sensor.c
+++ b/subsys/bluetooth/mesh/sensor.c
@@ -91,9 +91,17 @@ bool bt_mesh_sensor_delta_threshold(const struct bt_mesh_sensor *sensor,
 	 */
 	if (sensor->state.threshold.delta.type ==
 	    BT_MESH_SENSOR_DELTA_PERCENT) {
-		int64_t prev_mill = llabs(SENSOR_MILL(&sensor->state.prev));
+		/* Store the Status Delta Trigger threshold in sensor_value first to avoid int64_t
+		 * overflow caused by multiplication of values converted by the SENSOR_MILL() macro.
+		 */
+		struct sensor_value thrsh_delta = delta_mill >= 0 ?
+			sensor->state.threshold.delta.up : sensor->state.threshold.delta.down;
+		struct sensor_value thrsh = {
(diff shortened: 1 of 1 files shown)


subsys/bluetooth/mesh/sensor.c
```diff
@@ -91,9 +91,17 @@ bool bt_mesh_sensor_delta_threshold(const struct bt_mesh_sensor *sensor,
 	 */
 	if (sensor->state.threshold.delta.type ==
 	    BT_MESH_SENSOR_DELTA_PERCENT) {
-		int64_t prev_mill = llabs(SENSOR_MILL(&sensor->state.prev));
-
-		thrsh_mill = (prev_mill * thrsh_mill) / (100LL * 1000000LL);
+		/* Store the Status Delta Trigger threshold in sensor_value first to avoid int64_t
+		 * overflow caused by multiplication of values converted by the SENSOR_MILL() macro.
+		 */
+		struct sensor_value thrsh_delta = delta_mill >= 0 ?
```
ludvigsj, line 97: Use a pointer here, perhaps? I realize the struct is quite tiny so it wouldn't make a big difference

PavelVPV: Yeah, pointer will cause dereferencing, so not sure it is better.

Andrewpini (approved):

PavelVPV:
Now it should work, sorry for confusion.

ludvigsj (approved):

PR #13915, opened 2024-02-08
tests: bluetooth: tester: migrate mesh tester to new Sensor API

m-alperen-sener:
Migrate the PTS tester to use new sensor API
- Disable the legacy sensor value usage
- Update tester to be able to use new Sensor API 
 
Fix the light lc server and sensor for failing PTS test cases.


tests/bluetooth/tester/src/model_handler.c
```diff
@@ -302,16 +303,24 @@ static struct bt_mesh_sensor_srv sensor_srv =
...
 	for (i = 0; i < ARRAY_SIZE(sensors); ++i) {
 		if (sensor->type->id != sensors[i]->type->id) {
 			continue;
 		}
-
+#if !defined(CONFIG_BT_MESH_SENSOR_USE_LEGACY_SENSOR_VALUE)
+		int err = bt_mesh_sensor_value_from_sensor_value(sensor->type->channels[0].format,
+								 &values[i], rsp);
+		if (err) {
+			LOG_DBG("err: %d", err);
```
PavelVPV, line 318:
```suggestion
			LOG_ERR("Failed to convert sensor value (err: %d)", err);
```

m-alperen-sener:
I found failing test cases for light lc server properties set/get invalid and valid test cases. So I also added the fix commit.

subsys/bluetooth/mesh/light_ctrl_srv.c
```diff
@@ -1273,33 +1273,24 @@ static int prop_set(struct net_buf_simple *buf,
...
+	/* Regulator coefficients are raw IEEE-754 floats, status will always
+	 * be BT_MESH_SENSOR_VALUE_NUMBER:
 	 */
 	if (srv->reg) {
 		switch (id) {
 		case BT_MESH_LIGHT_CTRL_COEFF_KID:
-			status = bt_mesh_sensor_value_to_float(
-				&val, &srv->reg->cfg.ki.down);
-			break;
+			(void)bt_mesh_sensor_value_to_float(&val, &srv->reg->cfg.ki.down);
```
PavelVPV, line 1282: Why did you change it to ignore error code? What if a client (not PTS but another 3rd-party one) sent wrong data?

ludvigsj: This was my suggestion, and is explained in the comment above: `val->format` will always be `&bt_mesh_sensor_format_coefficient` here (because it is looked up from the ID above). `bt_mesh_sensor_format_coefficient.cb->to_float` always returns BT_MESH_SENSOR_VALUE_NUMBER, no matter what `val->raw` is (because it simply reinterprets the bytes as a float).

PavelVPV: Why do we need return code there then? :/

ludvigsj: Because other formats might have special `val->raw` values for which they return things like "Value is invalid", "Value is not know", etc. But since we know the format here we know that we can ignore it, it will always be a number _in this case with this format_. But `bt_mesh_sensor_value_to_float` needs to be general for all formats.

PavelVPV (approved):

subsys/bluetooth/mesh/light_ctrl_internal.h
```diff
@@ -36,6 +36,13 @@ prop_format_get(enum bt_mesh_light_ctrl_prop id)
 		return &bt_mesh_sensor_format_perceived_lightness;
 	case BT_MESH_LIGHT_CTRL_PROP_REG_ACCURACY:
 		return &bt_mesh_sensor_format_percentage_8;
+#if CONFIG_BT_MESH_LIGHT_CTRL_SRV_REG
```
ludvigsj, line 39: Should not be in #if. This function will eventually be used in client applications as well, they will need to be able to construct and send these values without enabling the regulator.

subsys/bluetooth/mesh/light_ctrl_internal.h
```diff
@@ -23,7 +23,7 @@ extern "C" {
 #endif
 
 static inline const struct bt_mesh_sensor_format *
-prop_format_get(enum bt_mesh_light_ctrl_prop id)
+prop_format_get(int id)
```
ludvigsj, line 26: Why not just extend the enum instead?

m-alperen-sener: bt_mesh_light_ctrl_prop and bt_mesh_light_ctrl_coeff are used in stack generally. Extending the bt_mesh_light_ctrl_prop seems likely to create a lot changes. And I don't see any issue that way the api is internal enough to be not confused.

ludvigsj: This API becomes public in #13899 . But I see that PR makes the same change. However, I don't see how adding values to an enum definition will cause any changes at all? As long as old values stay the same and no code uses the new definitions, nothing will "magically" change just because extra values are added.

Then again, we have `properties.h` which defines constants for all of these. I think I agree to change this to an `unit16_t`, and then in a new task just remove the `enum bt_mesh_light_ctrl_prop` entirely and use constants from `properties.h` instead in the Light Ctrl model implementations. It's weird that this is doubly defined.

tests/bluetooth/tester/src/model_handler.c
```diff
@@ -302,16 +303,24 @@ static struct bt_mesh_sensor_srv sensor_srv =
...
 	int i;
 
 	for (i = 0; i < ARRAY_SIZE(sensors); ++i) {
 		if (sensor->type->id != sensors[i]->type->id) {
 			continue;
 		}
-
+#if !defined(CONFIG_BT_MESH_SENSOR_USE_LEGACY_SENSOR_VALUE)
+		int err = bt_mesh_sensor_value_from_sensor_value(sensor->type->channels[0].format,
+								 &values[i], rsp);
```
ludvigsj, line 316: This could be its own task, but in the future the use of `sensor_value` for passing `values` should probably be changed to just passing the `raw` value. So we support everything with full precision in the tester.

ludvigsj (approved):

PR #13917, opened 2024-02-08
driver: flash: Network core flash driver

sigvartmh:
Flash driver for interacting with network core's flash from the application core. This utilizes IPC and NRF_RPC to do flash commands in the network core.


drivers/flash/Kconfig
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config  FLASH_RPC
```
nordicjm, line 7:
```suggestion
config FLASH_RPC
```

drivers/flash/Kconfig
```diff
@@ -0,0 +1,61 @@
...
+	must be higher than remote core boot priority.
+endif
+
+config FLASH_RPC_SYS_INIT_PRIORITY
+	int "Init priority"
+	default 48 if FLASH_RPC_HOST
+	default 51
+	help
+	  Device driver initialization priority.
+
```
nordicjm, line 49:
```suggestion
```

drivers/flash/Kconfig
```diff
@@ -0,0 +1,61 @@
...
+
+config FLASH_RPC_SYS_INIT
+	bool "Enable at sys init"
+	help
+	  SYS_INIT nrf_rpc
+
```
nordicjm, line 55:
```suggestion
```

drivers/flash/flash_rpc/flash_rpc_common.c
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+#include <zephyr/init.h>
+#include <zephyr/logging/log.h>
+#include <nrf_rpc/nrf_rpc_ipc.h>
+#include <nrf_rpc_cbor.h>
+
+LOG_MODULE_REGISTER(FLASH_RPC, CONFIG_FLASH_RPC_LOG_LEVEL);
+
+#ifdef CONFIG_FLASH_RPC_SYS_INIT
```
nordicjm, line 14: rather than doing this, exclude the file if the Kconfig is not enabled from cmake

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,205 @@
...
+}
+
+static const struct flash_parameters *flash_rpc_get_parameters(const struct device *dev)
+{
+	return NULL;
+}
+
+#ifdef CONFIG_FLASH_PAGE_LAYOUT
+static const struct flash_pages_layout flash_rpc_pages_layout = {
+	/* TODO Use DTS to define address and region, as in flash simulator */
```
nordicjm, line 180: this should probably be implemented

nordicjm:
I do wonder if we really need this here, or could it go upstream to zephyr instead?

drivers/flash/flash_rpc/flash_rpc_common.c
sigvartmh: Yeah makes sens I just copied how bt_rpmsg_common had done it.

sigvartmh:
> I do wonder if we really need this here, or could it go upstream to Zephyr instead?

Does it make sense for anything upstream to use this? I wouldn't mind putting it upstream; I wonder if it would be accepted. All this is just a workaround because the application core cannot access network core flash while being the driver of updates.

de-nordic:
> > I do wonder if we really need this here, or could it go upstream to Zephyr instead?
> 
> Does it make sense for anything upstream to use this? I wouldn't mind putting it upstream; I wonder if it would be accepted. All this is just a workaround because the application core cannot access network core flash while being the driver of updates.

That could be added as nrf5340 support code. This could improve the nrf5340 support upstream, so I guess upstreaming makes sens.
And regarding acceptance: we own the platform. If you want to place it in flash drivers, that would probably require some work to make it NRF PCD/nrf5340 specific.

drivers/flash/flash_rpc/flash_rpc_controller.c
sigvartmh: Implemented :)

drivers/flash/flash_rpc/flash_rpc_host.c
```diff
@@ -0,0 +1,152 @@
...
+
+LOG_MODULE_DECLARE(FLASH_RPC, CONFIG_FLASH_RPC_LOG_LEVEL);
+
+#define CBOR_BUF_FLASH_MSG_SIZE (sizeof(void *) + sizeof(size_t) + sizeof(off_t))
+
+NRF_RPC_IPC_TRANSPORT(flash_rpc_api_tr, DEVICE_DT_GET(DT_NODELABEL(ipc0)), "flash_rpc_api_ept");
+NRF_RPC_GROUP_DEFINE(flash_rpc_api, "flash_rpc_api", &flash_rpc_api_tr, NULL, NULL, NULL);
+
+static const struct device *const flash_controller =
+	DEVICE_DT_GET_OR_NULL(DT_CHOSEN(zephyr_flash_controller));
```
de-nordic, line 28: I am not sure about the layout of memory for the NET code but shouldn't we rather limit access to specific partition? We seem to allow write anywhere on that thing.

drivers/flash/flash_rpc/flash_rpc_host.c
```diff
@@ -0,0 +1,152 @@
...
+		return false;
+	}
+	if (zcbor_uint32_decode(ctx->zs, &ptr_addr)) {
+		*ptr = (void *)ptr_addr;
+	} else {
+		return false;
+	}
+
+	return zcbor_uint32_decode(ctx->zs, (uint32_t *)len);
+}
```
de-nordic, line 44: Since the function decodes the offset and size, shouldn't we here check whether these fit the flash layout? I know that flash driver will also do that, and will return -EINVAL.

drivers/flash/flash_rpc/flash_rpc_host.c
```diff
@@ -0,0 +1,152 @@
...
+
+static void flash_rpc_init_handler(const struct nrf_rpc_group *group,
+					   struct nrf_rpc_cbor_ctx *ctx, void *handler_data)
+{
+	ARG_UNUSED(handler_data);
+	nrf_rpc_cbor_decoding_done(group, ctx);
+
+	if (!device_is_ready(flash_controller)) {
+		LOG_ERR("Flash Init failure");
+		rsp_error_code_send(group, -EINVAL);
```
de-nordic, line 65: The -EINVAL is usually reserved for bad params, probably -ENODEV, maybe -ENXIO?

drivers/flash/flash_rpc/flash_rpc_host.c
```diff
@@ -0,0 +1,152 @@
...
+	struct nrf_rpc_cbor_ctx ctx;
+
+	NRF_RPC_CBOR_ALLOC(group, ctx, sizeof(int));
+
+	zcbor_int32_put(ctx.zs, err_code);
+
+	nrf_rpc_cbor_rsp_no_err(group, &ctx);
+}
+
+static void flash_rpc_init_handler(const struct nrf_rpc_group *group,
```
de-nordic, line 57: No other function checks whether this has been called so I guess this is optional.

drivers/flash/flash_rpc/flash_rpc_host.c
```diff
@@ -0,0 +1,152 @@
...
+	if (!decode_flash_msg(ctx, &offset, &buffer, &size)) {
+		LOG_INF("Decode failure");
+		err = -EBADMSG;
+		goto error_exit;
+	}
+
+	LOG_DBG("offset: 0x%x, size: %d", (uint32_t)offset, (uint32_t)size);
+	err = flash_erase(flash_controller, offset, size);
+	if (err) {
+		LOG_ERR("Erase failed");
```
de-nordic, line 131: This is only function that does the log on error, is it even needed?

drivers/flash/flash_rpc/flash_rpc_host.c
```diff
@@ -0,0 +1,152 @@
...
+static void flash_erase_handler(const struct nrf_rpc_group *group, struct nrf_rpc_cbor_ctx *ctx,
+		       void *handler_data)
+{
+	int err;
+	size_t size;
+	off_t offset;
+	void *buffer = NULL;
+
+	if (!decode_flash_msg(ctx, &offset, &buffer, &size)) {
+		LOG_INF("Decode failure");
```
de-nordic, line 123: Other functions do not do that? And shouldn't this be LOG_ERR?

drivers/flash/flash_rpc/flash_rpc_host.c
```diff
@@ -0,0 +1,152 @@
...
+		err = -EBADMSG;
+		goto error_exit;
+	}
+
+	LOG_DBG("offset: 0x%x, size: %d", (uint32_t)offset, (uint32_t)size);
```
de-nordic, line 128: From ./lib/libc/minimal/include/inttypes.h `"Offset: 0x%"PRIx32", size: %"PRIu32. `

include/drivers/flash/flash_rpc.h
```diff
@@ -0,0 +1,25 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum flash_rpc_command {
+	RPC_COMMAND_FLASH_INIT = 0x01,
+	RPC_COMMAND_FLASH_READ = 0x02,
+	RPC_COMMAND_FLASH_WRITE = 0x03,
+	RPC_COMMAND_FLASH_ERASE = 0x04,
+};
```
de-nordic, line 19: Question: is there some RPC namespace and commands we can collide with?

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+#define FLASH_RPC_FLASH_SIZE DT_REG_SIZE(FLASH_RPC_NODE)
+
+#define FLASH_RPC_PAGE_COUNT (FLASH_RPC_FLASH_SIZE/FLASH_RPC_ERASE_UNIT)
+
+static const struct flash_parameters flash_rpc_parameters = {
+	.write_block_size = FLASH_RPC_PROG_UNIT,
+	.erase_value = 0xff,
+};
+
+void rsp_error_code_handle(const struct nrf_rpc_group *group, struct nrf_rpc_cbor_ctx *ctx,
```
de-nordic, line 49: Missing `static` ?

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+			   void *handler_data)
+{
+	int32_t val;
+
+	if (zcbor_int32_decode(ctx->zs, &val)) {
+		*(int *)handler_data = (int)val;
+	} else {
+		*(int *)handler_data = -NRF_EINVAL;
```
de-nordic, line 57: I do not think this is the right code. Seems that we can only get here if failed to decode CBOR stream. Maybe ENOENT?

de-nordic: NOEND does not seem to be proper either. I have just remembered that the zcbor function can return false when different type than expected is found in stream.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+
+	if (!zcbor_uint32_put(ctx->zs, (uint32_t)*offset)) {
+		return false;
+	}
+
+	if (!zcbor_uint32_put(ctx->zs, (uint32_t)ptr)) {
+		return false;
+	}
+
+	return zcbor_uint32_put(ctx->zs, (uint32_t)*len);
```
de-nordic, line 84: You can probably just use the && between calls, the way we do in MCUmgr while decoding zcbor.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+
+	if (!encode_flash_msg(&ctx, &offset, NULL, &size)) {
+		return -EINVAL;
+	}
+
+	err = nrf_rpc_cbor_cmd(&flash_rpc_api, RPC_COMMAND_FLASH_ERASE, &ctx,
+				flash_rpc_get_rsp, &result);
+	if (err) {
+		return -EINVAL;
+	}
```
de-nordic, line 200: I think that EIO should be returned here and LOG_DBG/LOG_ERR with message and the real value of err.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+
+	LOG_DBG("offset: %x, buffer_ptr: %p, len: %d", (uint32_t)offset, buffer, (uint32_t)len);
+
+	err = nrf_rpc_cbor_cmd(&flash_rpc_api, RPC_COMMAND_FLASH_READ, &ctx,
+				flash_rpc_get_rsp, &result);
+	if (err) {
+		return -EINVAL;
+	}
```
de-nordic, line 147: Return EIO and add LOG_ERR/LOG_DBG with real value of err.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+{
+	ARG_UNUSED(dev);
+	int err;
+	int result;
+
+	struct nrf_rpc_cbor_ctx ctx;
+
+	if (buffer == NULL || len == 0) {
+		return -EINVAL;
+	}
```
de-nordic, line 135: This is disputable, but some of our drivers just return 0 and do nothing when `len == 0`. I think that this is good approach as there is no need to error if we are not going to do any work.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+
+	if (!encode_flash_msg(&ctx, &offset, buffer, &len)) {
+		return -EINVAL;
+	}
```
de-nordic, line 139: We probably need LOG_DBG/LOG_ERR here. Drivers return EINVAL everywhere and it is hard to track where the stuff comes from.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+{
+	ARG_UNUSED(dev);
+	int err;
+	int result;
+
+	struct nrf_rpc_cbor_ctx ctx;
+
+	if (data == NULL || len == 0) {
+		return -EINVAL;
+	}
```
de-nordic, line 164: Same as with read.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+
+	if (!encode_flash_msg(&ctx, &offset, (void *)data, &len)) {
+		return -EINVAL;
+	}
```
de-nordic, line 168: Same as with read.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+
+	err = nrf_rpc_cbor_cmd(&flash_rpc_api, RPC_COMMAND_FLASH_WRITE, &ctx,
+			flash_rpc_get_rsp, &result);
+
+	LOG_DBG("offset: %x, data_ptr: %p, len: %d", (uint32_t)offset, data, (uint32_t)len);
+	if (err) {
+		LOG_ERR("Failed to send command: %d", err);
+		return -EINVAL;
```
de-nordic, line 176: Change -EINVAL to -EIO.

drivers/flash/flash_rpc/flash_rpc_controller.c
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	LOG_DBG("Write command sent: %d", result);
```
de-nordic, line 179: Probably unneeded, user will have own error handling anyway.

drivers/flash/flash_rpc/flash_rpc_controller.c
de-nordic: Can we find some other E for that. As far as I can tell the zcbor, callled by encode_flash_msg, can only fail at this point if you run out of buffer to encode data.

include/drivers/flash/flash_rpc.h
de-nordic: I guess not. Getting familiar with NRF_RPC_GROUP_DEFINE

drivers/flash/flash_rpc/flash_rpc_host.c
de-nordic: OK, now I see that this is called on init by the controller side.

drivers/flash/flash_rpc/flash_rpc_host.c
sigvartmh: Ah yes, this should probably be -ENODEV

drivers/flash/flash_rpc/flash_rpc_host.c
sigvartmh: So, since the bootloader protects itself, doing a flash write to it would be a nop/cause a failure. 
We should not limit writing to the rest of the device.


This is the layout for the Netcore flash controller. Are you suggesting that we target `flash1`? Or did I misunderstand? If you wanted it to use the part of the partition here(which doesn't make sense in my mind as none of these partitions is used on the network core so far).
```DTS
	partitions {
			flash_controller: flash-controller@41080000 {
			compatible = "nordic,nrf53-flash-controller";
			reg = < 0x41080000 0x1000 >;
			partial-erase;
			#address-cells = < 0x1 >;
			#size-cells = < 0x1 >;
			flash1: flash@1000000 {
				compatible = "soc-nv-flash";
				erase-block-size = < 0x800 >;
				write-block-size = < 0x4 >;
				reg = < 0x1000000 0x40000 >;
				partitions {
					compatible = "fixed-partitions";
					#address-cells = < 0x1 >;
					#size-cells = < 0x1 >;
					boot_partition: partition@0 {
						label = "mcuboot";
						reg = < 0x0 0xc000 >;
					};
					slot0_partition: partition@c000 {
						label = "image-0";
						reg = < 0xc000 0x17000 >;
					};
					slot1_partition: partition@23000 {
						label = "image-1";
						reg = < 0x23000 0x17000 >;
					};
					storage_partition: partition@3a000 {
						label = "storage";
						reg = < 0x3a000 0x6000 >;
					};
				};
			};
		};
```

drivers/flash/flash_rpc/flash_rpc_controller.c
sigvartmh: ok, I just don't prefer that style and I assumed the compiler would optimize it away anyways :) but I've changed it to your suggestion.

drivers/flash/flash_rpc/flash_rpc_host.c
sigvartmh: it will eventually be checked by the flash driver and reported back to the caller. It would give a speed up to do a pre-check but would also add additional unnecessary code.

drivers/flash/flash_rpc/flash_rpc_host.c
sigvartmh: Not needed

drivers/flash/flash_rpc/flash_rpc_host.c
sigvartmh: moved it and changed it

drivers/flash/flash_rpc/flash_rpc_host.c
sigvartmh: added

include/drivers/flash/flash_rpc.h
sigvartmh: I Could maybe do FLASH_RPC instead will change.

drivers/flash/flash_rpc/flash_rpc_controller.c
sigvartmh: yes

drivers/flash/flash_rpc/flash_rpc_controller.c
sigvartmh: -EMSGSIZE sounded fitting.

drivers/flash/flash_rpc/flash_rpc_host.c
de-nordic: I was asking because as far as I can tell we will be using the flash driver also in MCUmgr right?

carlescufi (changes requested):
I think you should consider using a light-weight combination:

- Stop using nrf_ipc
- Use icmsg as a backend for the IPC service
- Use something super light-weight like nrfs does (check with @Rafal-Nordic)

If that doesn't work to reduce flash footprint then we should just use this PR as-is.

Great work by the way!

nvlsianpu:
@carlescufi
Thank for your comments.

We are pursuit to optimize the footprint. This will be done (most probably) by using own serialization layer instead of zcbor and tweaks the configuration. As  the result I expected something which will be possible in boot-time and in application-running time (two separate instances, separated in execution time, each with own configuration).

SeppoTakalo:
Dropping from 2.6.0 as there was no approvals before RC1.

drivers/flash/flash_rpc/flash_rpc_host.c
nvlsianpu: I need to go back there - VNFD should allow full access in boot stage (B0N's VNFD instance) and limits access in application run-time stage ( networking stack VNFD instance). Maybe we should do that in follow-up patch.

nvlsianpu (commented):
I will

nvlsianpu (approved):
I really appreciate how this code legible is.
Just one nit for help info.
I think access control might be introduced in follow-up PR.

drivers/flash/Kconfig
```diff
@@ -0,0 +1,59 @@
...
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config FLASH_RPC
+	bool "Flash over RPC [EXPERIMENTAL]"
+	select NRF_RPC
+	select NRF_RPC_CBOR
+	help
```
de-nordic, line 11: Shouldn't this depend on `FLASH`?

de-nordic (approved):
Looks OK. Once comment left.

drivers/flash/Kconfig
sigvartmh: It should depend on flash only if it's compiled for the network core I'll try to fix this.

drivers/flash/Kconfig
```diff
@@ -0,0 +1,59 @@
...
+choice
+	prompt "Flash RPC role selection"
+	default FLASH_RPC_HOST if SOC_NRF5340_CPUNET
+	default FLASH_RPC_CONTROLLER
+
+config FLASH_RPC_CONTROLLER
+	bool "Flash RPC Controller"
+	help
+	  Flash RPC API will use the nRF RPC library to access the flash of a host
+	  on the remote core.
```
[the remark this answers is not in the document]
sigvartmh: Yes that's a good note thanks.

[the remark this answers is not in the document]
sigvartmh: Added this

drivers/flash/Kconfig
```diff
@@ -0,0 +1,62 @@
...
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config FLASH_RPC
+	bool "Flash over RPC [EXPERIMENTAL]"
+	select NRF_RPC
+	select NRF_RPC_CBOR
+	select FLASH
+
```
de-nordic, line 12: Unneeded empty line?

PR #13922, opened 2024-02-08
Bluetooth: Controller: tx power radio default defines 54

KyraLengfeld:
Due to publicizing the controllers nRF54L15 and nRF54H20, and as part of the effort to include needed commits from the private fork, the controllers previously lived off of, to the main fork, this PR introduces Tx Power defines for nrf54Lx and nrf54Hx series controllers.


KyraLengfeld:
Renamed it, as I had internal names in branch name. See https://github.com/nrfconnect/sdk-nrf/pull/13929 for new PR.

PR #13923, opened 2024-02-08
samples: cellular: modem_shell: Add sock connect --keep_open

stig-bjorlykke:
Add an option to sock connect to keep socket open when its PDN connection is lost. This is useful to keep DTLS connections when network is lost for a short period.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -485,6 +485,7 @@ Cellular samples
     * Support for PDN CID 0 in the ``-I`` argument for the ``sock connect`` command.
     * Support for listing interface addresses using the ``link ifaddrs`` command.
     * Support for the ``MSG_WAITACK`` send flag using the ``sock send`` command.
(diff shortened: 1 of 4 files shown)


jfischer-no:
@stig-bjorlykke Does this draft still target v2.6.0?

stig-bjorlykke:
> @stig-bjorlykke Does this draft still target v2.6.0?

Yes, the new socket option will be added in libmodem v2.6.0 and I'm waiting for that PR to be created.

jfischer-no:
https://github.com/nrfconnect/sdk-nrf/pull/14195 is merged

trantanen (approved):

tokangas (approved):
Looks good to me.

I started wondering if we should have a separate command(s) for setting and reading options for sockets. There are several options for which we do not have support. Some could be implemented in `connect` like this, but then again most of the options are such, that those can be changed at runtime. It's a bit problematic that we do not have separate commands for opening and connecting sockets, though.

Let's keep this PR as it is, but just something to consider in the future.

peknis (approved):

PR #13926, opened 2024-02-08
Bluetooth: set configuration for HCI ISO data timestamps

Tschet1:
If the host uses timestamps for HCI ISO data packets which are not based on the controller's clock, the controller needs to be configured accordingly.

Added a KConfig which exposes this possibility.

### DIFF
--- a/subsys/bluetooth/controller/Kconfig
+++ b/subsys/bluetooth/controller/Kconfig
@@ -497,5 +497,16 @@ config BT_CTLR_SDC_SILENCE_UNEXPECTED_MSG_TYPE
 	help
 	  Convenience config to silence unexpected incoming msg type
 
+config BT_CTLR_SDC_IGNORE_HCI_ISO_DATA_TS_FROM_HOST
+	bool "Configure the controller to ignore timestamps in HCI ISO data packets."
+	depends on BT_CTLR_PERIPHERAL_ISO || BT_CTLR_CENTRAL_ISO || BT_CTLR_ADV_ISO
+	select DEPRECATED
+	help
+	  Configures the SoftDevice Controller to ignore timestamps on HCI ISO data packets.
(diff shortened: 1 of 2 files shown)


subsys/bluetooth/controller/Kconfig
```diff
@@ -484,5 +484,11 @@ config BT_CTLR_SDC_SILENCE_UNEXPECTED_MSG_TYPE
 	help
 	  Convenience config to silence unexpected incoming msg type
 
+config BT_CTLR_SDC_IGNORE_HCI_ISO_DATA_TS_FROM_HOST
+	bool
+	default 0
```
rugeGerritsen, line 489: `default n` shall be omitted according to the Kconfig style guidelines: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/build/kconfig/tips.html#redundant-defaults

rugeGerritsen (changes requested):

subsys/bluetooth/controller/Kconfig
```diff
@@ -484,5 +484,10 @@ config BT_CTLR_SDC_SILENCE_UNEXPECTED_MSG_TYPE
 	help
 	  Convenience config to silence unexpected incoming msg type
 
+config BT_CTLR_SDC_IGNORE_HCI_ISO_DATA_TS_FROM_HOST
+	bool
```
rugeGerritsen, line 488: Should probably have a prompt: 
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/build/kconfig/tips.html#what-to-turn-into-kconfig-options

rugeGerritsen (changes requested):

ivaniushkov (approved):

PR #13927, opened 2024-02-08
doc: nrf5340audio: minor additions and fixes

greg-fer:
Added link to SD Controller LE ISO Channels.
Moved the table about differences.
Fixed naming of HCI RPMsg.

### DIFF
--- a/applications/nrf5340_audio/index.rst
+++ b/applications/nrf5340_audio/index.rst
@@ -8,6 +8,32 @@ The nRF5340 Audio applications demonstrate audio playback over isochronous chann
 .. note::
    nRF5340 Audio applications and their DFU/FOTA functionality are marked as :ref:`experimental <software_maturity>`.
 
+The following table summarizes the differences between the available nRF5340 Audio applications.
+
+.. list-table:: Differences between nRF5340 Audio applications
+   :header-rows: 1
+
+   * - Application
+     - LE Audio mode and role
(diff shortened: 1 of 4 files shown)


wiba-nordic (approved):

applications/nrf5340_audio/index.rst
```diff
@@ -8,6 +8,32 @@ The nRF5340 Audio applications demonstrate audio playback over isochronous chann
...
+     - 3
+     - ✔
+   * - :ref:`Broadcast sink<nrf53_audio_broadcast_sink_app>`
+     - BIS gateway
+     - 2
+     -
+   * - :ref:`Broadcast source<nrf53_audio_broadcast_source_app>`
+     - BIS headset
+     - 2
+     - ✔
```
greg-fer, line 35: @alexsven , @koffes , @Krrrl , anything else you would add to this table?

alexsven: It would make sense to rather point out that you need 2-3 DKs to be able to test out unicast, and you need unicast_client and unicast_server applications
In all cases it makes most sense to have 3 DKs, one gateway and two headsets/earbuds

koffes (approved):

applications/nrf5340_audio/index.rst
greg-fer: Changed to 3 for both unicasts.

PR #13931, opened 2024-02-08
modules: hostap: Reserve the heap

krish2718:
By default (-1) libc uses all the available heap, this is good to efficiently use the RAM, but for scenarios which there isn't enough RAM the build will be success, but we see runtime failures.

Reserved the necessary RAM based on tests ahead to catch such scenarios, of course this would mean that we might be no using full RAM in case more is left, but that's never the case the with Wi-Fi :).

And this also adds the constant evaluation and fine tuneing like we do for kernel heap and other parameters.

Implements SHEL-2443.


krish2718:
@krga2022 FYI, Now we don't need to remember the magic RAM number to see if Wi-Fi works or not, if it compiles WPA supplicant should work.

krish2718:
@markaj-nordic @LuDuda  FYI.

sachinthegreen (approved):

richabp (approved):

rlubos (approved):

modules/hostap/Kconfig
```diff
@@ -16,11 +16,16 @@ menuconfig WPA_SUPP
...
 	select WIFI_NM
 	select EXPERIMENTAL if !SOC_SERIES_NRF53X && !SOC_SERIES_NRF91X
+	select COMMON_LIBC_MALLOC
 	help
 	  WPA supplicant implements 802.1X related functions.
 
 if WPA_SUPP
 
+config COMMON_LIBC_MALLOC_ARENA_SIZE
+	# 30K is mandatory, but might need more for long duration use cases
```
ioannisg, line 26: What is the meaning of "long duration" use cases , here @krish2718 ?

markaj-nordic:
> @markaj-nordic @LuDuda FYI.

As discussed f2f: In Matter we use sys_heap based allocators and thus we need to keep the configuration `CONFIG_COMMON_LIBC_MALLOC_ARENA_SIZE`=-1 to avoid duplicated heap being reserved (in practice we would just have 30k or RAM reserved but not used at all in Matter).

modules/hostap/Kconfig
krish2718: 12hours stress tests that we have nightly and we are seeing occasional OOM messages randomly (4/6/9hrs), but running with 40K is helping tests pass. But we still don't know the root cause of the heap crunch.

bama-nordic (approved):

modules/hostap/Kconfig
ioannisg: Thanks. @krish2718 it would be nice if the help text here explains these points
- "30 kB is mandatory" (mandatory for what? Should we say this is minimum requirement/baseline?)
- "A higher amount of heap may be required for long duration stress-tests" Although I've go doubts is this is the right thing to do: I can definitely think of customer scenarios that would require long uptime for the Wi-Fi stack :)

Wouldn't it be safer to determine the right size of the MALLOC arena that is OK for all mainline cases?

ioannisg: Great. Just improve a bit the help text, for the Kconfig, if possible

carlescufi:
@jukkar PTAL

rado17 (approved):

jukkar (approved):

jfischer-no:
@krish2718 please rebase

bjda (approved):

PR #13935, opened 2024-02-09
manifest: sdk-zephyr: Fix print of ssid overflow

kapbh:
WIFI scan result shows junk character in SSID because of buffer overflow. Fix the print of ssid in wifi scan.


jfischer-no (changes requested):
Manifest is updated in #14006

jfischer-no:
#14006 is merged.

PR #13939, opened 2024-02-09
samples: matter: Fix memory leak in Window Covering

markaj-nordic:
We were leaking 8 bytes each time the new attribute change was reported by the WindowCovering Server. It was resulting in running out of heap when stressing the device (by sending multiple up-or-open/down-or-close commands).

More specifically the `ChipCertificateSet::Init()` was failing to allocate the ~1k certificate buffer during CASE session
establishment after ~8 up-or-open/down-or-close sequences having been sent before.

### DIFF
--- a/samples/matter/window_covering/src/window_covering.cpp
+++ b/samples/matter/window_covering/src/window_covering.cpp
@@ -287,4 +287,6 @@ void WindowCovering::DoPostAttributeChange(intptr_t aArg)
 	VerifyOrReturn(data != nullptr);
 
 	PostAttributeChange(data->mEndpoint, data->mAttributeId);
+
+	chip::Platform::Delete(data);
 }


kkasperczyk-no (approved):

ArekBalysNordic (approved):

doublemis1:
`More specifically the ChipCertificateSet::Init() was failing to allocate the ~1k certificate buffer during CASE session`

Does it mean that the one certifcate buffer for CASE is 1k of heap?

markaj-nordic:
> `More specifically the ChipCertificateSet::Init() was failing to allocate the ~1k certificate buffer during CASE session`
> 
> Does it mean that the one certifcate buffer for CASE is 1k of heap?

Yep, this is what gets allocated in ChipCertificateSet::Init. In fact we allocate an  array of 3 certificates:
Ref:
https://github.com/nrfconnect/sdk-connectedhomeip/blob/4936f9fa2be9e6cd16b02e4fcb9a7aff611f4981/src/credentials/FabricTable.cpp#L361

PR #13945, opened 2024-02-09
samples: net: download: Add default config to CI

SebastianBoe:
see commit message

### DIFF
--- a/samples/net/download/sample.yaml
+++ b/samples/net/download/sample.yaml
@@ -3,6 +3,13 @@ sample:
 tests:
   sample.net.download_client:
     build_only: true
+    integration_platforms:
+      - nrf9161dk_nrf9161_ns
+    platform_allow:
+      - nrf9161dk_nrf9161_ns
+    tags: ci_build
+  sample.net.download_client.ci:
+    build_only: true
     extra_configs:
       - CONFIG_SHELL=y
       - CONFIG_COAP=y


samples/net/download/sample.yaml
```diff
@@ -3,11 +3,6 @@ sample:
 tests:
   sample.net.download_client:
     build_only: true
-    extra_configs:
-      - CONFIG_SHELL=y
-      - CONFIG_COAP=y
-      - CONFIG_DOWNLOAD_CLIENT_SHELL=y
-      - CONFIG_SAMPLE_COMPUTE_HASH=y
```
lemrey, line 10: Aren't these option set when building the sample in a test configuration?

This PR moves those option to the default sample build, which is perhaps not intended.

lemrey: The sample is not configured to use CoAP, nor it makes sense to have a shell (it just downloads a file and exits); that stuff is used when the sample is built for CI testing.

SebastianBoe: Usually, if a sample supports both a test configuration and a default build.

It will have a sample.yaml entry for the default build to prevent bit rot of the default build.

Shall I add a default build to the sample.yaml to make it clear it is supported and prevent bit rot?

SebastianBoe: discussed offline

balaji-nordic (approved):

lemrey (approved):
Thank you

PR #13947, opened 2024-02-09
quarantine: Add native_sim to quarantine for Windows

PerMac:
Native_sim is an enhanced implementation of native_posix and is not available on Windows, causing several tests to fail, instead of being skipped.

### DIFF
--- a/scripts/quarantine_windows_mac.yaml
+++ b/scripts/quarantine_windows_mac.yaml
@@ -9,6 +9,7 @@
 
 - platforms:
     - native_posix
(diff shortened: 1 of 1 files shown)


aescolar (approved):
I do not know how this list is used, but the same would apply to native_posix_64 , native_sim_64, nrf52_bsim, nrf5340bsim_nrf5340_cpuapp ,  nrf5340bsim_nrf5340_cpunet

shanthanordic (approved):

PR #13948, opened 2024-02-09
applications: matter: Repair Weather Station BLE advertising

ArekBalysNordic:
We need to initialize advertising and BLE SMP service before starting the "DFU over SMP" server. Schedule starting the server beyond the Application's init and do it in the Matter's thread.

### DIFF
--- a/applications/matter_weather_station/src/app_task.cpp
+++ b/applications/matter_weather_station/src/app_task.cpp
@@ -331,7 +331,15 @@ void AppTask::UpdateLedState()
 CHIP_ERROR AppTask::Init()
 {
 	/* Initialize Matter stack */
+#ifdef CONFIG_MCUMGR_TRANSPORT_BT
+	ReturnErrorOnFailure(
+		Nrf::Matter::PrepareServer(Nrf::Matter::InitData{ .mPostServerInitClbk = []() -> CHIP_ERROR {
+			Nrf::GetDFUOverSMP().StartServer();
+			return CHIP_NO_ERROR;
+		} }));
+#else
 	ReturnErrorOnFailure(Nrf::Matter::PrepareServer());
+#endif
 
 	/* Set references for RGB LED */
 	mRedLED = &Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED1);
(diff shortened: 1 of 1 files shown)


LuDuda (approved):

applications/matter_weather_station/src/app_task.cpp
```diff
@@ -384,7 +384,7 @@ CHIP_ERROR AppTask::Init()
 	k_timer_start(&sMeasurementsTimer, K_MSEC(kMeasurementsIntervalMs), K_MSEC(kMeasurementsIntervalMs));
 
 #ifdef CONFIG_MCUMGR_TRANSPORT_BT
-	Nrf::GetDFUOverSMP().StartServer();
+	PlatformMgr().ScheduleWork([](intptr_t) { Nrf::GetDFUOverSMP().StartServer(); });
```
kkasperczyk-no, line 387: Why to do this in the Matter thread? If we need to do that after `StartServer` maybe let's move it after this call, e.g. to the `StartApp` after `Init`.

markaj-nordic: I guess it's because the initialization of the SMP server takes place in the `Nrf::Matter::PrepareServer` that schedules all operation to the CHIP Thread. That being said, I'd suggest to place `Nrf::GetDFUOverSMP().StartServer()` in mPostServerInitClbk and feed it to the `Nrf::Matter::PrepareServer` call.

markaj-nordic:
```suggestion
		ReturnErrorOnFailure(Nrf::Matter::PrepareServer(Nrf::Matter::InitData{ .mPostServerInitClbk = [] {
#ifdef CONFIG_MCUMGR_TRANSPORT_BT
		Nrf::GetDFUOverSMP().StartServer();
#endif
		return CHIP_NO_ERROR;
	} }));
```

ArekBalysNordic: Ok, I agree that we can do it this way, but I hope that we documented InitData properly because assigning the lambda function to a strange field of the InitData struct is crazy, especially for users who do not use C++.

markaj-nordic: Each field of IniData is documented:
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/common/src/app/matter_init.h#L36
But we can always extend the docs if needed.

markaj-nordic (approved):

kkasperczyk-no (approved):

PR #13956, opened 2024-02-12
Add Monitor mode sample.

chiranjeevi2776:
This sample sets up WiFi in monitor mode, parses incoming packets, and displays statistics at regular intervals.


chiranjeevi2776:
@VivekUppunda, can you please review the Monitor sample once. It will set the mode into monitor and analyze the incoming wifi packets on raw socket and print the statistics at fixed interval. Please do let me know if anything else needed.

chiranjeevi2776:
This needs more cleanup so raised new PR.

PR #13958, opened 2024-02-12
Bluetooth: Mesh: Add metadata extraction for DFU

Andrewpini:
Adds script that can parse Mesh metadata from application .elf and .config file. The script is automatically run if the extract Mesh metadata config option is enabled. The output from this script is appended to dfu_application.zip in the project build folder.

### DIFF
--- a/cmake/mesh_dfu_metadata.cmake
+++ b/cmake/mesh_dfu_metadata.cmake
@@ -1,0 +1,25 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+find_package(Python3 REQUIRED)
+
+function(mesh_dfu_metadata)
+    add_custom_command(
+        OUTPUT ${PROJECT_BINARY_DIR}/dfu_application.zip_ble_mesh_metadata.json
+        COMMAND
+        ${PYTHON_EXECUTABLE}
+        ${ZEPHYR_NRF_MODULE_DIR}/scripts/bluetooth/mesh/mesh_dfu_metadata.py
+        ${PROJECT_BINARY_DIR}
+        DEPENDS
+        ${PROJECT_BINARY_DIR}/dfu_application.zip
+    )
+
+    add_custom_target(
+        parse_mesh_metadata
+        ALL
+        DEPENDS ${PROJECT_BINARY_DIR}/dfu_application.zip_ble_mesh_metadata.json
+    )
+endfunction()
--- a/scripts/bluetooth/mesh/mesh_dfu_metadata.py
+++ b/scripts/bluetooth/mesh/mesh_dfu_metadata.py
@@ -1,0 +1,412 @@
+#!/usr/bin/env python3
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+'''
+This script is used to extract the Bluetooth Mesh composition data from the build folder of
+an application. This information is used to generate a hash that can verify if the composition
+of a device will change after a Mesh device firmware upgrade has completed.
+The input to this script is the following content of the build folder of an application:
+    - build/zephyr/zephyr.elf
+    - build/zephyr/.config
+The script produces the following output to the build folder:
+	- build/zephyr/dfu_application.zip_mesh_metadata.json
+This output is also appended to the archive located at build/zephyr/dfu_application.zip.
+'''
+
+import struct
+import sys
+import os
+from elftools.elf.elffile import ELFFile
+import json
+from cryptography.hazmat.primitives import cmac
+from cryptography.hazmat.primitives.ciphers import algorithms
+from zipfile import ZipFile
+import traceback
+
+FILE_NAME_IN_ZIP = 'ble_mesh_metadata.json'
+FILE_NAME = 'dfu_application.zip_ble_mesh_metadata.json'
+
+
+def exit_with_error_msg():
+    traceback.print_exc()
+    print("Extracting BLE Mesh metadata failed")
+    print("You can bypass this script by disabling the CONFIG_BT_MESH_DFU_METADATA_ON_BUILD option in your project config")
+    sys.exit(0)
+
+
+class Elem:
+    def __init__(self, loc):
+        self.loc = loc
+        self.vnd_list = []
+        self.sig_list = []
+
+    def vnd_model_add(self, cid, vid):
+        self.vnd_list.append((cid << 16) + vid)
+
+    def sig_model_add(self, id):
+        self.sig_list.append(id)
+
+    def bytestring_generate(self):
+        bytestring = bytearray()
+        bytestring.extend(self.loc.to_bytes(2, 'little'))
+        bytestring.append(len(self.sig_list))
+        bytestring.append(len(self.vnd_list))
+
+        for sig in self.sig_list:
+            bytestring.extend(sig.to_bytes(2, 'little'))
+        for vnd in self.vnd_list:
+            bytestring.extend(vnd.to_bytes(4, 'little'))
+
+        return bytestring
+
+    def dict_generate(self):
+        return {
+            "location": self.loc,
+            "sig_models": self.sig_list,
+            "vendor_models": self.vnd_list,
+        }
+
+
+class Comp0:
+    # Must stay in order
+    FEATURE_KCONF_OPTS = [
+        'CONFIG_BT_MESH_RELAY',
+        'CONFIG_BT_MESH_GATT_PROXY',
+        'CONFIG_BT_MESH_FRIEND',
+        'CONFIG_BT_MESH_LOW_POWER',
+    ]
+
+    def __init__(self, cid, pid, vid, kconfig):
+        if 'CONFIG_BT_MESH_CRPL' not in kconfig.keys():
+            raise Exception("Could not find CONFIG_BT_MESH_CRPL Kconfig option")
+        self.elems = []
+        self.cid = cid
+        self.pid = pid
+        self.vid = vid
+        self.__features_add(kconfig)
+        self.hash = None
+
+    def elem_add(self, loc):
+        new_elem = Elem(loc)
+        self.elems.append(new_elem)
+        return new_elem
+
+    def __features_add(self, kconfig):
+
+        self.feat = 0
+        self.crpl = int(kconfig['CONFIG_BT_MESH_CRPL'])
+
+        for i, opt in enumerate(self.FEATURE_KCONF_OPTS):
+            self.feat += (1 if kconfig.get(opt) == 'y' else 0) << i
+
+    def __bytestring_generate(self):
+        bytestring = bytearray()
+        bytestring.extend(self.cid.to_bytes(2, 'little'))
+        bytestring.extend(self.pid.to_bytes(2, 'little'))
+        bytestring.extend(self.vid.to_bytes(2, 'little'))
+        bytestring.extend(self.crpl.to_bytes(2, 'little'))
+        bytestring.extend(self.feat.to_bytes(2, 'little'))
+
+        for elem in self.elems:
+            bytestring.extend(elem.bytestring_generate())
+
+        return bytestring
+
+    def dict_generate(self):
+        return {
+            "cid": self.cid,
+            "pid": self.pid,
+            "vid": self.vid,
+            "crpl": self.crpl,
+            "features": self.feat,
+            "elements": [e.dict_generate() for e in self.elems]
+        }
+
+    def hash_generate(self):
+        # Uses 16-byte zero key
+        crypto = cmac.CMAC(algorithms.AES(bytes(16)))
+        crypto.update(bytes(self.__bytestring_generate()))
+        self.hash, *_ = struct.unpack('<L', crypto.finalize()[:4])
+        return self.hash
+
+
+class KConfig(dict):
+
+    @classmethod
+    def from_file(cls, filename):
+        """
+        Extracts content of '.config' file into a dictionary
+
+        Parameters:
+            config_path (str): Path to '.config' file used by a firmware
+        Returns:
+            cls: A KConfig object where keys are Kconfig option names and values
+            are option values, parsed from the config file at config_path.
+        """
+        configs = cls()
+        try:
+            with open(filename, 'r') as config:
+                for line in config:
+                    if not line.startswith("CONFIG_"):
+                        continue
+                    kconfig, value = line.split("=")
+                    configs[kconfig] = value.strip()
+                return configs
+        except Exception as err :
+            raise Exception("Unable to parse .config file") from err
+
+    def version_parse(self):
+        try:
+            clean_str = self['CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION'].replace("+", ".").replace("\"", "")
+            version_list = [int(s) for s in clean_str.split(".") if s.isdigit()]
+            return {
+                "major": version_list[0],
+                "minor": version_list[1],
+                "revision": version_list[2],
+                "build_number": version_list[3],
+            }
+        except Exception as err :
+            raise Exception("Unable to parse CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION Kconfig option") from err
+
+
+def read_symbol_data(elf, symbol_addr):
+    """
+    Reads variable data from the '.symtab' section of the .elf file.
+
+    Parameters:
+        elf (ELFFile): ELFFile instance
+        symbol_addr (int): Address of the variable to read
+
+    Returns:
+        bytearray: Data read from the specified address
+    """
+    section = elf.get_section_by_name('.symtab')
+    if section is None:
+        raise Exception('Unable to find .symtab section')
+    symbol = None
+    for s in section.iter_symbols():
+        if (s.entry.st_value == symbol_addr) and\
+            (len(s.name) > 0) and\
+            ("$" not in s.name) and\
+            s.entry.st_size > 0:
+            symbol = s
+            break
+    else:
+        raise Exception('Unable to find symbol at address {symbol_addr}')
+    file_offset = None
+    for segment in elf.iter_segments():
+        if segment.header['p_type'] != 'PT_LOAD':
+            continue
+        if (symbol['st_value'] >= segment['p_vaddr']) and\
+            (symbol['st_value'] < segment['p_vaddr'] + segment['p_filesz']):
+            file_offset = symbol['st_value'] - segment['p_vaddr'] + segment['p_offset']
+            break
+    else:
+        raise Exception('Error getting file offset from ELF data')
+    elf.stream.seek(file_offset)
+    sz = symbol['st_size']
+
+    return elf.stream.read(sz)
+
+def find_comp_data_from_dwarf(elf_path):
+    """
+    Find all occurrences of the `bt_mesh_comp` variable in the .elf file
+
+    The composition data declaration must have the const-qualifier. It can also be declared as an array. Example:
+        ```
+        static const struct bt_mesh_comp comp;
+        const struct bt_mesh_comp comp;
+        const struct bt_mesh_comp comp[2];
+        ```
+    Parameters:
+        elf_path (ELFFile): ELFFile instance
+
+    Returns:
+        addr: Address of the found composition data instances in the firmware.
+    """
+    DW_OP_addr = 0x3
+
+    with open(elf_path, 'rb') as file:
+        elf_file = ELFFile(file)
+        dwarf_info = elf_file.get_dwarf_info()
+        comp_data_arr = []
+        for cu_header in dwarf_info.iter_CUs():
+            for die in cu_header.iter_DIEs():
+                if die.tag != 'DW_TAG_variable':
+                    continue
+                location = die.attributes.get('DW_AT_location')
+                if location is None:
+                    continue
+                if location.form not in ("DW_FORM_exprloc"):
+                    continue
+                opcode = location.value[0]
+                if opcode != DW_OP_addr:
+                    continue
+                # Check that the variable type is either `const struct bt_mesh_comp` or
+                # `const struct bt_mesh_comp[]`.
+                exp_tags = [
+                    ['DW_TAG_const_type', 'DW_TAG_structure_type'],
+                    ['DW_TAG_const_type', 'DW_TAG_array_type', 'DW_TAG_const_type', 'DW_TAG_structure_type'],
+                ]
+                type_die = die
+                max_length = max(len(arr) for arr in exp_tags)
+                try:
+                    for i in range(max_length):
+                        type_die_type = type_die.attributes.get('DW_AT_type')
+                        if type_die_type is None:
+                            raise Exception('DW_AT_type is missing')
+                        type_die = type_die.get_DIE_from_attribute('DW_AT_type')
+                        for j, _ in enumerate(exp_tags):
+                            if len(exp_tags[j]) > i and\
+                                type_die.tag == exp_tags[j][i]:
+                                break
+                        else:
+                            raise Exception('Wrong DW_AT_type')
+                        name = type_die.attributes.get('DW_AT_name')
+                        if name and name.value == b'bt_mesh_comp':
+                            break
+                    else:
+                        continue
+                except Exception:
+                    continue
+
+                addr = int.from_bytes(die.attributes.get('DW_AT_location').value[1:5], 'little')
+                comp_data_arr.append(addr)
+
+        if comp_data_arr is None or len(comp_data_arr) == 0:
+            raise Exception("Could not find composition data in .elf file")
+
+        if len(comp_data_arr) > 1:
+            raise Exception("Multiple instances of composition data found in .elf file")
+
+        return comp_data_arr[0]
+
+def read_comp_data(elf_path, addr, kconfigs):
+    """
+    Reads content of the composition data variable from .elf file.
+
+    Parameters:
+        elf_path (ELFFile): ELFFile instance
+        addr (int): Address of the composition data variable
+        kconfigs (KConfig): A KConfig object representing Kconfig options used for the firmware to compile with
+    Returns:
+        Parsed Composition data
+    """
+
+    label_cnt = int(kconfigs['CONFIG_BT_MESH_LABEL_COUNT']) if 'CONFIG_BT_MESH_LABEL_COUNT' in kconfigs.keys() else 0
+    lcd_srv = (kconfigs['CONFIG_BT_MESH_LARGE_COMP_DATA_SRV'] == 'y') if 'CONFIG_BT_MESH_LARGE_COMP_DATA_SRV' in kconfigs.keys() else False
+
+    with open(elf_path, 'rb') as elf_file:
+        elf = ELFFile(elf_file)
+        cdp0_value = read_symbol_data(elf, addr)
+
+        # The format of the composition data is defined by `struct bt_mesh_comp` type.
+        # The format of an element is defined by `struct bt_mesh_elem` type.
+        # The format of a model is defined by `struct bt_mesh_model` type.
+        # All types are declared in `zephyr/include/zephyr/bluetooth/mesh/access.h`.
+        cid, pid, vid, _, elems_count, elems_ptr = struct.unpack('HHHHII', cdp0_value)
+
+        comp = Comp0(cid, pid, vid, kconfigs)
+
+        elems_value = read_symbol_data(elf, elems_ptr)
+        elems_iter = struct.iter_unpack('IHBBII', elems_value)
+        i = 0
+
+        for elem in elems_iter:
+            i += 1
+            if i > elems_count:
+                raise Exception('Extracted more elems than \'elem_count\'')
+            __rt, loc, sig_count, vnd_count, sig_ptr, vnd_ptr = elem
+            elem_item = comp.elem_add(loc)
+
+            def models_unpack(ptr, elem_item, vnd):
+                models_value = read_symbol_data(elf, ptr)
+                model_format = 'HHIIIHHIHH' + ('I' if label_cnt > 0  else '') + 'II' + ('I' if lcd_srv else '')
+                models_iter = struct.iter_unpack(model_format, models_value)
+
+                for model in models_iter:
+                    id1, id2, __rt, __pub, __keys, __keys_cnt, _, __groups, __groups_cnt, _, __uuids, __op, __cb = model
+                    if not vnd:
+                        elem_item.sig_model_add(id1)
+                    else:
+                        elem_item.vnd_model_add(id1, id2)
+
+            if sig_count > 0:
+                models_unpack(sig_ptr, elem_item, False)
+            if vnd_count > 0:
+                models_unpack(vnd_ptr, elem_item, True)
+        return comp
+
+def parse_comp_data(elf_path, kconfigs):
+    """
+    Extract composition data from .elf and .config file.
+
+    Parameters:
+        kconfigs (KConfig): A KCoonfig object representing Kconfig options used for the firmware to compile with
+    Returns:
+        Parsed Composition data
+    """
+    try:
+        addr = find_comp_data_from_dwarf(elf_path)
+        return read_comp_data(elf_path, addr, kconfigs)
+    except Exception as err :
+        raise Exception("Failed to extract composition data from .elf file") from err
+
+def encoded_metadata_get(version, comp):
+    core_type = 0 # FIX ME: For now, hardcoded to application core
+    binary_size = os.path.getsize(os.path.join(bin_path, 'app_update.bin'))
+    elem_cnt = len(comp.elems)
+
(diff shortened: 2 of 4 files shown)


Andrewpini:
This is not 100% complete yet. Im still trying to iron out some issues with Cmake. The issue is that I'm append something to an output target, namely the dfu_application.zip, which also is an dependency of the custom command. This causes the script to be ran redundantly on secnd build since it has altered one of its dependencies. On the second build it bail out early due to the fact that the file already exists on in the zip, meaning it will not try to re-run the script on the consecutive build.

I'm trying to see if there are a way around this, but im open to input from others. Have not worked alot with Cmake before.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,314 @@
...
+        read_comp_data(elf_path, addr, kconfigs, comp)
+
+    j_dict = {
+        "sign_version": version_parse(kconfigs),
+        "composition_hash": str(hex(comp.hash_generate())),
+        "composition_data": comp.dict_generate(),
+    }
+
+    with open(metadata_path, "w") as outfile:
+        outfile.write(json.dumps(j_dict, indent=4))
```
omkar3141, line 312: Could you also print the hash and the metadata into console? That would make it easy for users to do manual commands by copy-pasting metadata directly from console instead of having to extract the zip and manifest etc.


```
Hash: 0937465829
Metadata: 0010303030040494954865755757577365622
...
...
```

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,314 @@
...
+                location = die.attributes.get('DW_AT_location')
+                if location is None:
+                    continue
+                if location.form not in ("DW_FORM_exprloc"):
+                    continue
+                opcode = location.value[0]
+                if opcode != DW_OP_addr:
+                    continue
+                exp_tags = [
+                    ['DW_TAG_const_type', 'DW_TAG_structure_type'],
```
PavelVPV, line 146:
```suggestion

                # Check that the variable type is either `const struct bt_mesh_comp` or
                # `const struct bt_mesh_comp[]`.
                exp_tags = [
                    ['DW_TAG_const_type', 'DW_TAG_structure_type'],
```

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,314 @@
...
+    return elf.stream.read(sz)
+
+def read_comp_data(elf_path, addr, kconfigs, comp):
+    label_cnt = int(kconfigs['CONFIG_BT_MESH_LABEL_COUNT']) if 'CONFIG_BT_MESH_LABEL_COUNT' in kconfigs.keys() else 0
+    lcd_srv = (kconfigs['CONFIG_BT_MESH_LARGE_COMP_DATA_SRV'] == 'y') if 'CONFIG_BT_MESH_LARGE_COMP_DATA_SRV' in kconfigs.keys() else False
+
+    with open(elf_path, 'rb') as elf_file:
+        elf = ELFFile(elf_file)
+        cdp0_value = read_symbol_data(elf, addr)
+        cid, pid, vid, _, elems_count, elems_ptr = struct.unpack('HHHHII', cdp0_value)
```
PavelVPV, line 229:
```suggestion
        cdp0_value = read_symbol_data(elf, addr)
        
        # The format of the composition data is defined by `struct bt_mesh_comp` type.
        # The format of an element is defined by `struct bt_mesh_elem` type.
        # The format of a model is defined by `struct bt_mesh_model` type.
        # All types are declared in `zephyr/include/zephyr/bluetooth/mesh/access.h`.
        cid, pid, vid, _, elems_count, elems_ptr = struct.unpack('HHHHII', cdp0_value)
```

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,314 @@
...
+    def hash_generate(self):
+        # Uses 16-byte zero key
+        crypto = cmac.CMAC(algorithms.AES(bytes(16)))
+        crypto.update(bytes(self.__bytestring_generate()))
+        self.hash, *_ = struct.unpack('<L', crypto.finalize()[:4])
+        return self.hash
+
+
+def get_die_var_addr(die):
+    location = die.attributes.get('DW_AT_location')
```
PavelVPV, line 120:
```suggestion
def get_die_var_addr(die):
    """
    Extracts variable address from `DW_AT_location` attribute.
    
    Parameters:
        die (DIE): A die containing the variable
        
    Return:
        int: address of the variable
    """
    location = die.attributes.get('DW_AT_location')
```

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,314 @@
...
+
+    with open(elf_path, 'rb') as file:
+        elf_file = ELFFile(file)
+        dwarf_info = elf_file.get_dwarf_info()
+        comp_data_arr = []
+        for cu_header in dwarf_info.iter_CUs():
+            for die in cu_header.iter_DIEs():
+                if die.tag != 'DW_TAG_variable':
+                    continue
+                name = die.attributes.get('DW_AT_name')
```
PavelVPV, line 136: `name` assignment here can be removed. It is not used anywhere.

PavelVPV (commented):
Looks good! The comments are mostly about adding comments to the python script.

I like appending the manifest file to the archive instead of changing the script that generates the archive.

We need a proper way of handling error to not disrupt build process.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,314 @@
...
+        new_elem = Elem(loc)
+        self.elems.append(new_elem)
+        return new_elem
+
+    def features_add(self, kconfig):
+        if 'CONFIG_BT_MESH_CRPL' not in kconfig.keys():
+            print("Could not find CONFIG_BT_MESH_CRPL Kconfig option")
+            sys.exit(1)
+
+        self.crpl = int(kconfigs['CONFIG_BT_MESH_CRPL'])
```
omkar3141, line 72: `kconfigs` or `kconfig`?

omkar3141 (changes requested):

cmake/mesh_dfu_metadata.cmake
```diff
@@ -0,0 +1,20 @@
...
+function(mesh_dfu_metadata)
+    add_custom_command(
+        OUTPUT ${PROJECT_BINARY_DIR}/dfu_application.zip_mesh_metadata.json
+        COMMAND
+        ${PYTHON_EXECUTABLE}
+        ${ZEPHYR_NRF_MODULE_DIR}/scripts/bluetooth/mesh/mesh_dfu_metadata.py
+        ${PROJECT_BINARY_DIR}
+        DEPENDS
+        ${PROJECT_BINARY_DIR}/dfu_application.zip
+        ${PROJECT_BINARY_DIR}/app_update.bin
```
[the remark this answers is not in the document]
Andrewpini: Yes, if the Cmake stuff stays 'as is' then you are right.

cmake/mesh_dfu_metadata.cmake
```diff
@@ -0,0 +1,20 @@
+find_package(Python3 REQUIRED)
+
+function(mesh_dfu_metadata)
+    add_custom_command(
+        OUTPUT ${PROJECT_BINARY_DIR}/dfu_application.zip_mesh_metadata.json
```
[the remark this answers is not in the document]
Andrewpini: I choose the name to mirror what they did for the original file. They produce the dfu_application.zip_manifest.json first, then puts it in the zip renamed as manifest.json. I do not think it is possible to have the mesh_manifest inside the zip as the OUTPUT directly, which has the name mesh_metadata.json. On another note, I think that the output here can be symbolic, e.g. whatever name we want to call it that does not have to be an existing file in the build. It was at first, but then I experimented with it when I tried to resolve the Cmake build issue.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def elem_add(self, loc):
+        new_elem = Elem(loc)
+        self.elems.append(new_elem)
+        return new_elem
+
+    def features_add(self, kconfig):
+        if 'CONFIG_BT_MESH_CRPL' not in kconfig.keys():
+            print("Could not find CONFIG_BT_MESH_CRPL Kconfig option")
+            sys.exit(1)
```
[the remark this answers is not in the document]
Andrewpini: Do you have any good idea how the script should behave if it fails to extract metadata? Of course, if it cant find anything  to parse then the output will not be produced, but how to handle if some of the info is missing? Should it still generate the JSON if lets say the image version is misssing?

scripts/bluetooth/mesh/mesh_dfu_metadata.py
Andrewpini: Well, they do not need to extract it. It is also located outside the zip as dfu_application.zip_mesh_metadata.json, just like for the existing manifest. I can add a print for the Hash, but I would like to avoid printing too much since it kind of polutes the build output if you just pile on line after line of information.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+    def bytestring_generate(self):
+        bytestring = bytearray()
+        bytestring.extend(self.loc.to_bytes(2, 'little'))
+        bytestring.append(len(self.sig_list))
+        bytestring.append(len(self.vnd_list))
+
+        for sig in self.sig_list:
+            bytestring.extend(sig.to_bytes(2, 'little'))
+        for vnd in self.vnd_list:
+            bytestring.extend(vnd.to_bytes(4, 'little'))
```
m-alperen-sener, line 50: Should we check system endianness or little endianness is always the case?

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+
+            self.feat += (1 if opt_set else 0) << i
+
+    def __bytestring_generate(self):
+        bytestring = bytearray()
+        bytestring.extend(self.cid.to_bytes(2, 'little'))
+        bytestring.extend(self.pid.to_bytes(2, 'little'))
+        bytestring.extend(self.vid.to_bytes(2, 'little'))
+        bytestring.extend(self.crpl.to_bytes(2, 'little'))
+        bytestring.extend(self.feat.to_bytes(2, 'little'))
```
m-alperen-sener, line 107: the same here; is endianness always little?

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        Reads content of the composition data variable from .elf file.
+
+        Parameters:
+            addr (int): Address of the composition data variable
+            kconfigs (dict): A dictionary representing Kconfig options used for the firmware to compile with
+            comp (dict): A Composition Data dictionary class object to be filled in
+
+        Returns:
+            TBA
```
m-alperen-sener, line 288: ?

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+    def extract_comp_data(self, kconfigs, comp):
+        """
+        Extract composition data from .elf file.
+
+        Parameters:
+            kconfigs (dict): A dictionary representing Kconfig options used for the firmware to compile with
+            comp (dict): A Composition Data dictionary class object to be filled in
+
+        Returns:
+            TBA
```
m-alperen-sener, line 345: ?

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+
+    Parameters:
+        config_path (str): Path to '.config' file used by a firmware
+    Returns:
+        dict: A dictionary where key is Kconfig option name and value is the value of the corresponding Kconfig option.
+    """
+    with open(config_path, 'r') as config:
+        configs = {}
+        for line in config:
+            if not line.startswith("CONFIG_"):
```
Balaklaka, line 363: Could this be CONFIG_BT_MESH_? To reduce the number of extracted sets.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
Andrewpini: It is always little I think.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
Andrewpini: It is not only BT_MESH options that are extracted here.

Andrewpini: It has minimal impact on execution time with and without.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
m-alperen-sener: Seems like it is possible to get this info: https://github.com/eliben/pyelftools/blob/main/elftools/elf/elffile.py#L65

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+            "elements": []
+        }
+
+        for elem in self.elems:
+            j_dict["elements"].append(elem.dict_generate())
+
+        return j_dict
+
+    def hash_generate(self):
+        # Uses 16-byte zero key
```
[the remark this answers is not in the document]
PavelVPV: Created NCSDK-26026

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+    Returns:
+        dict: A dictionary where key is Kconfig option name and value is the value of the corresponding Kconfig option.
+    """
+    with open(config_path, 'r') as config:
+        configs = {}
+        for line in config:
+            if not line.startswith("CONFIG_"):
+                continue
+            kconfig, value = line.split("=")
+            configs[kconfig] = value.split("\n")[0]
```
ludvigsj, line 366:
```suggestion
            configs[kconfig] = value.strip()
```
?

cmake/mesh_dfu_metadata.cmake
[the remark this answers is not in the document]
Andrewpini: Yes, that would be nice. I have no idea how to do that right now, but it could be usefull.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+        "composition_data": comp.dict_generate(),
+    }
+
+    with open(metadata_path, "w") as outfile:
+        outfile.write(json.dumps(j_dict, indent=4))
+    zip.write(metadata_path, file_name_in_zip)
+    zip.close()
+
+    print("Bluetooth Mesh Composition metadata generated:")
+    print("\tMetadata hash: {}".format(j_dict["composition_hash"]))
```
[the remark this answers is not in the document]
Andrewpini: You mean the composition+version+size etc as a string?

cmake/mesh_dfu_metadata.cmake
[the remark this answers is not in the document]
PavelVPV: Can be done in a separate task.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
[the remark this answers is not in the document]
PavelVPV: Yes, which is the output of `metadata encode` shell command: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/connectivity/bluetooth/api/mesh/shell.html#mesh-models-dfu-metadata-encode-major-minor-rev-buildnum-size-coretype-hash-elems-userdata

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+
+        Parameters:
+            die (DIE): A die containing the variable
+
+        Return:
+            int: address of the variable
+        """
+        location = die.attributes.get('DW_AT_location')
+        addr = ((location.value[1] << 0 ) | (location.value[2] << 8)  |
+                        (location.value[3] << 16) | (location.value[4] << 24))
```
ludvigsj, line 162:
```suggestion
        addr = int.from_bytes(location.value[1:5], 'little')
```

scripts/bluetooth/mesh/mesh_dfu_metadata.py
m-alperen-sener: Don't bother I thought this part is about reading the raw data.

cmake/mesh_dfu_metadata.cmake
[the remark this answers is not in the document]
Andrewpini: Maybe this should be a separate task as well? it might be easy, but if this thing should go in by friday I think it would be wise to do it later.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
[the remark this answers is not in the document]
Andrewpini: What if I add a warning print that is issued at the end of the script if something fails, with cause of failure? The json is not generated in this instance.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+        Parameters:
+            kconfigs (dict): A dictionary representing Kconfig options used for the firmware to compile with
+            comp (dict): A Composition Data dictionary class object to be filled in
+
+        Returns:
+            TBA
+        """
+        for addr in self.addrs:
+            self.__read_comp_data(addr, kconfigs, comp)
+
```
ludvigsj, line 349: I might be misreading, but `__read_comp_data` will _overwrite_ `cid`, `pid` and `vid`, and then just append the elements to what is already there in `comp`. Meaning if there are multiple `addrs`, `comp` will contain the `cid` etc.from the last one parsed, and all the elements from all the composition data instances?

In general it is a bit weird for this function to take in `comp` at all, looks like writing C in Python. I suggest to rewrite `__read_comp_data` to return a `Comp0` directly, and do

```python
return [self.__read_comp_data(addr, kconfigs) for addr in self.addrs]
```

or something here. Then the caller will have to handle potentially multiple instances.

ludvigsj (changes requested):
Most of these are more or less small comments on Python coding style, however there is a potential issue with multiple comp datas.

In general, the structure of functions vs classes vs dict and passing return values by parameter vs return value is somewhat all over the place, I'll write up a high-level comment about the general structure later, but submitting these for now.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,314 @@
...
+    def hash_generate(self):
+        # Uses 16-byte zero key
+        crypto = cmac.CMAC(algorithms.AES(bytes(16)))
+        crypto.update(bytes(self.__bytestring_generate()))
+        self.hash, *_ = struct.unpack('<L', crypto.finalize()[:4])
+        return self.hash
+
+
+def get_die_var_addr(die):
```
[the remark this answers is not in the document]
ludvigsj: I disagree with this. The class has no state (or, it doesn't need to have state, it has now because of the arbitrary split that `find_comp_data_from_dwarf` is called in `__init__` and `read_comp_data` is not.) and makes no sense as a class just to group a lot of (things that should all be) `@staticmethod`s.

My suggestion:
  * inline `get_die_var_addr`, it is used only once and with my suggested rewrite it could be one or two lines long.
  * either:
      * create a `parse_comp_data` which calls both `find_comp_data_from_dwarf` and `read_comp_data` (potentially inlining the latter)
      * The one I prefer: create `@classmethod Comp0::from_elf() -> [Comp0]` which does the same.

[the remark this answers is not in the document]
PavelVPV: Fine by me

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+        self.crpl = int(kconfig['CONFIG_BT_MESH_CRPL'])
+
+        for i, opt in enumerate(self.FEATURE_KCONF_OPTS):
+            opt_set = False
+            try:
+                opt_set = kconfig[opt] == 'y'
+            except Exception:
+                pass
+
+            self.feat += (1 if opt_set else 0) << i
```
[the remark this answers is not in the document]
Andrewpini: Nice, did not know of the "get".

scripts/bluetooth/mesh/mesh_dfu_metadata.py
Andrewpini: Yes... I think you are right. I had a feeling that something was a bit off.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+            comp (dict): A Composition Data dictionary class object to be filled in
+
+        Returns:
+            TBA
+        """
+        for addr in self.addrs:
+            self.__read_comp_data(addr, kconfigs, comp)
+
+
+def parse_kconfig(config_path):
```
ludvigsj, line 351: Suggestion for more readable structure:

```python
class KConfig(dict):

    @classmethod
    def from_file(cls, filename: str) -> cls:
        # This is `parse_kconfig` moved here

    def parse_version(self) -> {}:
        # This is `version_parse` moved here
```

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,425 @@
...
+    zip = ZipFile(zip_path, "a")
+    if file_name_in_zip in zip.namelist():
+        # Mesh metadata already present in zip file
+        sys.exit(0)
+
+    comp = Comp0()
+    kconfigs = parse_kconfig(config_path)
+    elf_parser = CompElfParser(elf_path)
+
+    comp.features_add(kconfigs)
```
ludvigsj, line 409: I'm thinking this should perhaps be called in the parse-elf-function that parses the Elf. That function necessarily already has the `kconfigs` available, it makes more sense for that parse-function to return a fully specified `Comp0` (or, list of `Comp0`s like mentioned yesterday) I think.

ludvigsj (commented):
Small comments on improved structure, not blockers from my side but nice to have.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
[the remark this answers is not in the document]
Balaklaka: After I tried this I see that either remove this line as it is not needed, or better do as Pavel suggest and print the full encoded metadata string. If it is needed in the file it is optional as the file be read by some machine/script and that can create the encoded string. But I do not think it hurt either.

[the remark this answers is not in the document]
ludvigsj: I agree, the bytstring should be the entire string you need to include in the message. I also think it should be included in the json, this will make it easier for customers wanting to make their own initiators for use with our DFU solution (on platforms where the Mobile App lib is not availble) and is easy to do.

Andrewpini:
@m-alperen-sener @Balaklaka @omkar3141 @ludvigsj please take a round and resolve your comments if you feel they are finished. A little hard to keep track of all comments now.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,389 @@
...
+        "composition_hash": str(hex(comp.hash_generate())),
+        "composition_data": comp.dict_generate(),
+    }
+
+    with open(metadata_path, "w") as outfile:
+        outfile.write(json.dumps(j_dict, indent=4))
+    zip.write(metadata_path, FILE_NAME_IN_ZIP)
+    zip.close()
+
+    print(f"Bluetooth Mesh Composition metadata generated:")
```
ludvigsj, line 387: No need for f-string without parameters

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,389 @@
...
+            if vnd_count > 0:
+                models_unpack(vnd_ptr, elem_item, True)
+        return comp
+
+def parse_comp_data(elf_path, kconfigs):
+    """
+    Extract composition data from .elf and .config file.
+
+    Parameters:
+        kconfigs (dict): A dictionary representing Kconfig options used for the firmware to compile with
```
ludvigsj, line 348:
```suggestion
        kconfigs (KConfig): A KConfig object representing Kconfig options used for the firmware to compile with
```

scripts/bluetooth/mesh/mesh_dfu_metadata.py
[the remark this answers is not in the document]
ludvigsj: @PavelVPV  @Andrewpini  resolving this now, since it seems we all agree now and the class is gone.

cmake/mesh_dfu_metadata.cmake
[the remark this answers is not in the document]
PavelVPV: Created NCSDK-26078

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,388 @@
...
+    Parameters:
+        elf (ELFFile): ELFFile instance
+        symbol_addr (int): Address of the variable to read
+
+    Returns:
+        bytearray: Data read from the specified address
+    """
+    section = elf.get_section_by_name('.symtab')
+    if section is None:
+        raise Exception('Unable to find section')
```
PavelVPV, line 187:
```suggestion
        raise Exception('Unable to find .symbtab section')
```

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,388 @@
...
+        try:
+            with open(filename, 'r') as config:
+                for line in config:
+                    if not line.startswith("CONFIG_"):
+                        continue
+                    kconfig, value = line.split("=")
+                    configs[kconfig] = value.strip()
+                return configs
+        except Exception:
+            exit_with_error_msg("Unable to parse .config file")
```
PavelVPV, line 158: Maybe worth printing exception message in all try-except cases to know exactly what happened?

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,388 @@
...
+        self.hash = None
+
+    def elem_add(self, loc):
+        new_elem = Elem(loc)
+        self.elems.append(new_elem)
+        return new_elem
+
+    def __features_add(self, kconfig):
+        if 'CONFIG_BT_MESH_CRPL' not in kconfig.keys():
+            exit_with_error_msg("Could not find CONFIG_BT_MESH_CRPL Kconfig option")
```
PavelVPV, line 96: Could it be better to raise the exception up to the main function and catch it there?

PavelVPV (commented):
Last few comments from my side

ludvigsj (commented):
Looks good now, will approve once the full encoded metadata (with version and size, not just the hash) is included in the json and printout

scripts/bluetooth/mesh/mesh_dfu_metadata.py
Andrewpini: Im trying something like this:
```
except Exception as err:
exit_with_error_msg(f"Failed to extract composition data from .elf file: \n\t{err}")
```
That will add the subreason for the fail as well.

PavelVPV: Sounds good! Please try it once (manually raise exception at any place) just to see how it prints the message and the exception.

ludvigsj: I agree with Pavel that exceptions should be raised up to main. Then, for exceptions where can't add any more info than the original exception (which IMO are most of them), remove the `try ... except` and let main handle the original exception. If you really need to report more info, do 
```python
try:
    do_thing_that_can_fail()
except Exception as err:
    raise Exception("Thing that can fail has failed") from err
```

ludvigsj: When doing this, `f"{err}"` returns `"Thing that can fail has failed"`. To get the string for the sub-exception, do `f"{err.__cause__}"` (after checking for `err.__cause__ is None`)

ludvigsj: Or, easier, `import traceback` and `print_exception`/`print_exc` and let Python do the work of printing the relevant messages, line numbers etc.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -0,0 +1,413 @@
...
+        "encoded_metadata": str(encoded_metadata.hex()),
+    }
+
+    with open(metadata_path, "w") as outfile:
+        outfile.write(json.dumps(j_dict, indent=4))
+    zip.write(metadata_path, FILE_NAME_IN_ZIP)
+    zip.close()
+
+    print("Bluetooth Mesh Composition metadata generated:")
+    print(f"\tMetadata hash: {j_dict['composition_hash']}")
```
PavelVPV, line 411: Sorry for nitpicking, but this hash has no sense on its own without other parameters. The encoded metadata is what we really need to be printed.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
PavelVPV: ? :(

scripts/bluetooth/mesh/mesh_dfu_metadata.py
ludvigsj: @Andrewpini you seem to have misunderstood me. What I meant was: `exit_with_error_msg` should be called from exactly one place, an `except` clause in main. All other errors should just be left to propagate, or chained using `raise Exception() from err`.

Meaning, for this piece of code specifically (but also every other place currently calling `exit_with_error_msg` from deep within some function), you'd do

```python
try:
    with open(filename, 'r') as config:
        for line in config:
            if not line.startswith("CONFIG_"):
                continue
            kconfig, value = line.split("=")
            configs[kconfig] = value.strip()
        return configs
except Exception as err:
    raise Exception("Unable to parse .config file") from err
```

and then main would be

```python
try:
    kconfigs = KConfig.from_file(config_path)
    comp = parse_comp_data(elf_path, kconfigs)
    version = kconfigs.version_parse()
    encoded_metadata = encoded_metadata_get(version, comp)
except Exception as err:
    # Do something here, either traceback.print_exc + sys.exit(0), print(err) + sys.exit etc.
```

but really, unless you really care about returning 0 in sys.exit, just letting any exception propagate out of main will accomplish the same thing: The error is printed and the script exits (but with exit code 1). This is currently what will happen in case something goes wrong with the zip-file operations, for instance.

scripts/bluetooth/mesh/mesh_dfu_metadata.py
Andrewpini: Sorry, got a bit confused. It should be better now.

ludvigsj (approved):

scripts/bluetooth/mesh/mesh_dfu_metadata.py
ludvigsj: @PavelVPV @Andrewpini resolving this since I think it is handled fully now

subsys/bluetooth/mesh/Kconfig
```diff
@@ -128,6 +128,17 @@ choice TFM_PROFILE_TYPE
 	default TFM_PROFILE_TYPE_NOT_SET
 endchoice
 
+config BT_MESH_DFU_METADATA_ON_BUILD
+	bool "Generate DFU metadata on build"
+	default y if BT_MESH_DFU_SRV || BT_MESH_DFU_CLI
+	depends on SIGN_IMAGES
+	depends on BT_MESH_DFU_METADATA
```
Balaklaka, line 135: Minor: Then BT_MESH_DFU_METADATA should also be "default y if (BT_MESH_DFU_SRV || BT_MESH_DFU_CLI)". Why should BT_MESH_DFU_METADATA explicit be set to y if this is enabled by default and we are dependent on it.

Balaklaka (approved):

PavelVPV (approved):

m-alperen-sener (approved):

omkar3141 (approved):

PR #13964, opened 2024-02-12
Enable UART runtime configuration

tokangas:
The default value for `CONFIG_UART_USE_RUNTIME_CONFIGURE` has been changed, so it needs to be explicitly enabled for those who need `uart_config_get()`.

### DIFF
--- a/lib/nrf_modem_lib/trace_backends/uart/Kconfig
+++ b/lib/nrf_modem_lib/trace_backends/uart/Kconfig
@@ -35,6 +35,7 @@ config NRF_MODEM_LIB_TRACE_BACKEND_UART_ZEPHYR
 	bool "UART"
 	depends on SERIAL
 	depends on UART_ASYNC_API
+	select UART_USE_RUNTIME_CONFIGURE
 	help
 	  Modem tracing with UART require the chosen `nordic,modem-trace-uart`
 	  to be set in the Device tree.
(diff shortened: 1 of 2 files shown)


tokangas:
@eivindj-nordic @lemrey I added the dependency to the nrf_modem_lib UART trace backend `Kconfig` file. Is this ok, or should it be in the UART trace snippet instead?

eivindj-nordic:
> @eivindj-nordic @lemrey I added the dependency to the nrf_modem_lib UART trace backend `Kconfig` file. Is this ok, or should it be in the UART trace snippet instead?

It should be in the snippet for now I think as it is not needed for the deprecated uart trace backends (using nrfx directly). There are more Kconfigs that we can move to the uart trace backend kconfig once we remove the deprecated backends.

tokangas:
> If it is not adding anything to the application when using the deprecated zephyr driver I'm ok with keeping it in the `Kconfig` file. If not it should be in the snippet for now I think as it is not needed for the deprecated uart trace backends (using nrfx directly). There are more Kconfigs that we can move to the uart trace backend kconfig once we remove the deprecated backends.

The change only affects `CONFIG_NRF_MODEM_LIB_TRACE_BACKEND_UART_ZEPHYR`, so it should be fine.

eivindj-nordic:
@tokangas Where was this changed? It seems to be enabled in zephyr upstream(?).
https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/serial/Kconfig#L51

eivindj-nordic (approved):

trantanen (approved):

moksanen (approved):

PR #13966, opened 2024-02-12
Bluetooth: controller: Add support for nrf54H/L

hermabe:
- drivers/mpsl support
- mpsl library support
- softdevice controller support

`zephyr,entropy` configuration and driver is missing so bluetooth does not compile. Keeping as draft for now.


hermabe:
https://github.com/nrfconnect/sdk-nrfxlib/pull/1207 needed to find the libraries.

However, they changed the value of `CONFIG_SOC` so we will have to move the libraries to match that.

subsys/mpsl/init/mpsl_init.c
```diff
@@ -267,12 +358,24 @@ static int32_t mpsl_lib_init_internal(void)
 	clock_cfg.rc_ctiv = 0;
 	clock_cfg.rc_temp_ctiv = 0;
 #endif
+#else
```
olivier-le-sage, line 361: Isn't it fine to leave the clock_cfg as-is? Atm it just gets ignored in MPSL on 54h

Otherwise: we can use less ifdef logic by initializing clock_cfg as:

`mpsl_clock_lfclk_cfg_t clock_cfg = {};`

subsys/bluetooth/controller/Kconfig
```diff
@@ -107,8 +108,8 @@ config SYSTEM_WORKQUEUE_STACK_SIZE
 
 config BT_CTLR_SDC_LLPM
 	bool "Enable Low Latency Packet Mode support"
-	depends on SOC_SERIES_NRF52X
 	select BT_CONN_PARAM_ANY if !BT_HCI_RAW
+	depends on (SOC_SERIES_NRF52X || SOC_SERIES_NRF54LX || SOC_SERIES_NRF54HX)
```
olivier-le-sage, line 112: I wonder if it would be more future-proof to have `depends on (!SOC_NRF5340_CPUNET)` or similar

subsys/mpsl/init/mpsl_init.c
hermabe: Maybe that is simpler. The clock handling is anyway being refactored now so this will have to change in the future anyway.

subsys/bluetooth/controller/Kconfig
```diff
@@ -301,7 +302,9 @@ choice BT_LL_SOFTDEVICE_VARIANT
 						     BT_CTLR_SDC_PAWR_ADV || \
 						     BT_CTLR_SDC_PAWR_SYNC || \
 						     BT_ISO || \
-						     SOC_COMPATIBLE_NRF5340_CPUNET)
```
vegu-nordic, line 304: Will this change still select softdevice for 53 babblesim?

subsys/mpsl/init/mpsl_init.c
hermabe: The `CLOCK_CONTROL_*` configs are not defined since `CONFIG_CLOCK_CONTROL` is not set so there will be some ifdeffery anyway. Propose to leave it as is for now.

vegu-nordic (approved):

olivier-le-sage (approved):

rugeGerritsen (approved):

PR #13973, opened 2024-02-12
Bluetooth: Mesh: Add a note about fixed passkey and force LESC in LE Pairing Responder model

PavelVPV:
This PR changes the following:
- forces an application to only use the LE Secure Connection pairing when using the LE Pairing Responder model as it is currently the most secure way to establish pairing between LE devices,
- adds a note regarding a warning generated by `CONFIG_BT_FIXED_PASSKEY`. When `CONFIG_BT_FIXED_PASSKEY` is enabled, it generates the warning that this option should not be used for production. This should not be a problem as far as the passkey is generated using secure RNG, generated on every pairing request and the passkey is transferred over mesh. These all correct for the LE Pairing Responder model use-case. Therefore, adding a note that this warning can be disregarded.
- adds a diagram of interaction between an initiator and the LE Pairing Responder model


mia-ko (approved):

omkar3141 (commented):
The SVG diagram is bit unclear to me. Specially, sequence of events after "The Initiator initiates ..."

1. When "A passkey is requested" triggers? Is there a callback that implies this? May be better to specify that.
2. Same goes for "setting the received passkey".
3. Also, it is not clear, which side is the mesh node and which side is mobile app. It will be nice to label that in the top boxes as well. For example: Initiator (Mobile App), LE Pairing Responder Model (Mesh node).

omkar3141:
> > * When "A passkey is requested" triggers? Is there a callback that implies this? May be better to specify that.
> > * Same goes for "setting the received passkey".
> 
> We can't know how this is implemented on the other side. It can be a mobile phone or anything else.

Yeah, but here, we are intending this to be a mobile phone for our use cases. I am lookin for something that will clearly distinguish the actual devices on which this is running. For example:
Initiator (SMP client), Responder (SMP Server); or
Initiator (DFU initiator), Responder (DFU Distributor); or something else.

PavelVPV:
> > > * When "A passkey is requested" triggers? Is there a callback that implies this? May be better to specify that.
> > > * Same goes for "setting the received passkey".
> > 
> > 
> > We can't know how this is implemented on the other side. It can be a mobile phone or anything else.
> 
> Yeah, but here, we are intending this to be a mobile phone for our use cases. I am lookin for something that will clearly distinguish the actual devices on which this is running. For example: Initiator (SMP client), Responder (SMP Server); or Initiator (DFU initiator), Responder (DFU Distributor); or something else.

How does this relate to "A passkey is requested` and "Setting the received passkey" triggers? Also note that LE Pairing Responder model page has no connection with DFU/SMP or anything. This is a model documentation, not a sample. The use-case can be explained in the sample documentation, which is already done here: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/bluetooth/mesh/dfu/distributor/README.html#smp-over-bluetooth-authentication.

doc/nrf/libraries/bluetooth_services/mesh/vnd/le_pair_resp.rst
```diff
@@ -12,11 +12,22 @@ The LE Pairing Responder model can be used to hand over a passkey that will auth
 Before the pairing is initiated, an initiator should send an LE Pairing message with Passkey Reset sub-opcode to set a new passkey for the next pairing request.
 The passkey returned in the LE Pairing message with the Passkey Status sub-opcode should be used for the next pairing.
 
+The passkey is generated using the :c:func:`bt_rand` function and set to the host using the :c:func:`bt_passkey_set` function.
```
omkar3141, line 15: Just realized that we don't mention anywhere that this is a vendor model. Can't add comments to the unchanged text. So adding it here.

> The LE Pairing Responder model **is a vendor model. This model** can be used to hand over a passkey that will authenticate a Bluetooth LE connection over a mesh network when it is not possible to use other pairing methods.

omkar3141:
> How does this relate to "A passkey is requested` and "Setting the received passkey" triggers? Also note that LE Pairing Responder model page has no connection with DFU/SMP or anything. This is a model documentation, not a sample. The use-case can be explained in the sample documentation, which is already done here: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/bluetooth/mesh/dfu/distributor/README.html#smp-over-bluetooth-authentication.

When you hit reply, everything is copied, that's why. 

The labels on the diagram look fine now. I'm still a little bit wondering about sequence. Does the "A passkey is requested" event happens exactly at the same time as "bt_conn_auth_cb.pairing_display()"? It is probably before that. However, diagram is bit misleading in this aspect.

PavelVPV:
> When you hit reply, everything is copied, that's why.

I replied to those first 2 bullet points.

> The labels on the diagram look fine now. I'm still a little bit wondering about sequence. Does the "A passkey is requested" event happens exactly at the same time as "bt_conn_auth_cb.pairing_display()"? It is probably before that. However, diagram is bit misleading in this aspect.

From the user level it happens "at the same time". From the code level it doesn't matter either because the firmware blindly accepts the pairing request. For this kind of diagram it doesn't matter as it doesn't explain the pairing protocol. What matter is what each side does.

omkar3141 (approved):

mia-ko (commented):
The only comment that I have is if we need to state what mobile app we are talking about..? Will this be clear enough to the user that we mean nRF Mesh mobile app?

Oh, and if we are not adding any more details, change Mobile App to Mobile app :)

PavelVPV:
> The only comment that I have is if we need to state what mobile app we are talking about..? Will this be clear enough to the user that we mean nRF Mesh mobile app?
> 
> Oh, and if we are not adding any more details, change Mobile App to Mobile app :)

Changed `Mobile App` to `Mobile app`.

PR #13975, opened 2024-02-12
applications: matter: Add Bluetooth LE state indications

ArekBalysNordic:
- BLE connectivity manager now supports registering callbacks for state change. The callback can be defined in the application to indicate the current state using for example LEDs, LCDs, etc.

- There are several states defined:
	- Idle
	- Scanning
	- Pairing
	- LostDevice

- Implemented Indication for Bridge sample utilizing LED 2 located on the nRF7002DK board.

### DIFF
--- a/applications/matter_bridge/doc/matter_bridge_description.rst
+++ b/applications/matter_bridge/doc/matter_bridge_description.rst
@@ -156,6 +156,20 @@ Button 1:
       Releasing the button within a 3-second window of the initiation cancels the factory reset procedure.
 
 .. include:: ../../../samples/matter/lock/README.rst
+    :start-after: matter_door_lock_sample_led1_start
+    :end-before: matter_door_lock_sample_led1_end
+
+LED 2:
+   If the :kconfig:option:`CONFIG_BRIDGED_DEVICE_BT` kconfig option is set to ``y``, shows the current state of Bridge's Bluetooth LE connectivity.
+   The following states are possible:
+
+   * Turned Off - The Bridge device is in the idle state and has no Bluetooth LE devices paired.
+   * Solid On - The Bridge device is in the idle state and all connections to the Bluetooth LE bridged devices are stable.
+   * Slow Even Flashing (1000 ms on / 1000 ms off) - The Bridge device lost connection to at least one Bluetooth LE bridged device.
+   * Even Flashing (300 ms on / 300 ms off) - The scan for Bluetooth LE devices is in progress.
+   * Fast Even Flashing (100 ms on / 100 ms off) - The Bridge device is connecting to the Bluetooth LE device and waiting for the Bluetooth LE authentication PIN code.
+
+.. include:: ../../../samples/matter/lock/README.rst
     :start-after: matter_door_lock_sample_jlink_start
     :end-before: matter_door_lock_sample_jlink_end
 
--- a/applications/matter_bridge/src/app_task.cpp
+++ b/applications/matter_bridge/src/app_task.cpp
@@ -48,6 +48,42 @@ static bt_uuid *sUuidLbs = BT_UUID_LBS;
 static bt_uuid *sUuidEs = BT_UUID_ESS;
 static bt_uuid *sUuidServices[] = { sUuidLbs, sUuidEs };
 static constexpr uint8_t kUuidServicesNumber = ARRAY_SIZE(sUuidServices);
+/**
+ * @brief Blink rates for indication the BLE Connectivity Manager state.
+ *
+ */
+constexpr static uint32_t kPairingBlinkRate{ 100 };
+constexpr static uint32_t kScanningBlinkRate_ms{ 300 };
+constexpr static uint32_t kLostBlinkRate_ms{ 1000 };
+
+void BLEStateChangeCallback(Nrf::BLEConnectivityManager::State state)
+{
+	switch (state) {
+	case Nrf::BLEConnectivityManager::State::Connected:
+		Nrf::PostTask([] { Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED2).Set(true); });
+		break;
+	case Nrf::BLEConnectivityManager::State::Scanning:
+		Nrf::PostTask([] {
+			Nrf::GetBoard()
+				.GetLED(Nrf::DeviceLeds::LED2)
+				.Blink(kScanningBlinkRate_ms, kScanningBlinkRate_ms);
+		});
+		break;
+	case Nrf::BLEConnectivityManager::State::LostDevice:
+		Nrf::PostTask([] {
+			Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED2).Blink(kLostBlinkRate_ms, kLostBlinkRate_ms);
+		});
+		break;
+	case Nrf::BLEConnectivityManager::State::Pairing:
+		Nrf::PostTask([] {
+			Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED2).Blink(kPairingBlinkRate, kPairingBlinkRate);
+		});
+		break;
+	default:
+		Nrf::PostTask([] { Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED2).Set(false); });
+	}
+}
+
 #endif /* CONFIG_BRIDGED_DEVICE_BT */
 
 } /* namespace */
@@ -110,6 +146,7 @@ CHIP_ERROR AppTask::Init()
 	/* Initialize Matter stack */
 	ReturnErrorOnFailure(Nrf::Matter::PrepareServer(Nrf::Matter::InitData{ .mPostServerInitClbk = [] {
 #ifdef CONFIG_BRIDGED_DEVICE_BT
+		Nrf::BLEConnectivityManager::Instance().RegisterStateCallback(BLEStateChangeCallback);
 		/* Initialize BLE Connectivity Manager before the Bridge Manager, as it must be ready to recover
 		 * devices loaded from persistent storage during the bridge init. */
 		CHIP_ERROR bleInitError =
--- a/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
+++ b/samples/matter/common/src/bridge/ble_connectivity_manager.cpp
@@ -139,14 +139,21 @@ void BLEConnectivityManager::ConnectionHandler(bt_conn *conn, uint8_t conn_err)
 				Instance().Reconnect(reinterpret_cast<BLEBridgedDeviceProvider *>(context));
 			},
 			reinterpret_cast<intptr_t>(providerToRecover));
+		/* We have still a device to recover, keep the LostDevice state active */
+		Instance().UpdateStateFlag(State::LostDevice, true);
 	} else if (!sys_slist_is_empty(&Instance().mRecovery.mListToRecover)) {
 		/* There are pending providers to recover and no more scanned ones, schedule next scan operation. */
 		Instance().mRecovery.StartTimer();
+	} else {
+		/* All devices have been recovered, disable LostDevice state */
+		Instance().UpdateStateFlag(State::LostDevice, false);
 	}
 
 	char addrStr[BT_ADDR_LE_STR_LEN];
 	bt_addr_le_to_str(dstAddr, addrStr, sizeof(addrStr));
 	LOG_INF("Connected: %s", addrStr);
+	/* A new devices has been connected, activate the Connected state */
+	Instance().UpdateStateFlag(State::Connected, true);
 
 #if defined(CONFIG_BT_SMP)
 	err = bt_conn_set_security(conn, static_cast<bt_security_t>(CONFIG_BRIDGE_BT_MINIMUM_SECURITY_LEVEL));
@@ -170,6 +177,10 @@ exit:
 		LOG_ERR("Failed to set the connection security (%d)", err);
 	} else {
 		LOG_ERR("The connection failed (%d)", err);
+#if defined(CONFIG_BT_SMP)
+		/* In case when the connection failed, disable the Pairing state */
+		Instance().UpdateStateFlag(State::Pairing, false);
+#endif
 	}
 	/* Trigger the connection callback to inform the application that the connection procedure failed. */
 	provider->GetBLEBridgedDevice().mFirstConnectionCallback(
@@ -257,6 +268,9 @@ void BLEConnectivityManager::PairingComplete(struct bt_conn *conn, bool bonded)
 		return;
 	}
 
+	/* Pairing finished, disable the Pairing state */
+	Instance().UpdateStateFlag(State::Pairing, false);
+
 	/* Once pairing completed successfully, start GATT discovery procedure. */
 	StartGattDiscovery(conn, provider);
 }
@@ -268,6 +282,9 @@ void BLEConnectivityManager::PairingFailed(struct bt_conn *conn, enum bt_securit
 	bt_addr_le_to_str(bt_conn_get_dst(conn), addr, sizeof(addr));
 
 	LOG_ERR("Pairing failed conn: %s, reason %d\n", addr, reason);
+
+	/* Pairing finished, disable the Pairing state */
+	Instance().UpdateStateFlag(State::Pairing, false);
 
 	Instance().RemoveBLEProvider(*bt_conn_get_dst(conn));
 }
@@ -378,6 +395,9 @@ CHIP_ERROR BLEConnectivityManager::Init(bt_uuid **serviceUuids, uint8_t serviceU
 		return CHIP_ERROR_INVALID_ARGUMENT;
 	}
 
+	/* Initialize the BLEConnectivity manage's state to Idle */
+	UpdateStateFlag(State::Idle, true);
+
 	memcpy(mServicesUuid, serviceUuids, serviceUuidsCount * sizeof(serviceUuids));
 	mServicesUuidCount = serviceUuidsCount;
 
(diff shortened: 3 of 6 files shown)


samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -32,6 +32,23 @@ class BLEConnectivityManager {
 	static constexpr uint16_t kMaxConnectedDevices = CONFIG_BT_MAX_CONN - 1;
 	static constexpr uint8_t kMaxServiceUuids = CONFIG_BT_SCAN_UUID_CNT;
 
+	/**
+	 * States for indication the most important BLE Connectivity Manager states.
+	 * If the new state occurs the previous one is automatically disabled until the next occurrence.
```
wiba-nordic, line 37: This is an odd sentence. I'm guessing it tries to say something like "Only the latest state is indicated" or "A new state disables the previous state"?

wiba-nordic (changes requested):

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -257,6 +261,9 @@ void BLEConnectivityManager::PairingComplete(struct bt_conn *conn, bool bonded)
 		return;
 	}
 
+	/* Indicate that pairing has been finished */
+	Instance().NotifyUpdateState(State::Idle);
```
kkasperczyk-no, line 265: Generally I'm not sure if this notify-callback architecture will work correctly. My understanding is that this is quite generic mechanism that could be used for anything (in our particular case for blinking LEDs).

In the current implementation transition between the states is done in a way that sometimes give false information to the user (for LED use case it does not matter, but generally for someone it could matter). E.g. let's assume we have one device and we start adding new one, it will go like this:

LostDevice->Scanning->Pairing->**Idle**->LostDevice

So at some point the user will be notified that bridge is in the idle state, while it's not true, because it should be in `LostDevice`.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -269,6 +276,9 @@ void BLEConnectivityManager::PairingFailed(struct bt_conn *conn, enum bt_securit
 
 	LOG_ERR("Pairing failed conn: %s, reason %d\n", addr, reason);
 
+	/* Indicate that pairing has been finished */
+	Instance().NotifyUpdateState(State::Idle);
```
kkasperczyk-no, line 280: Ditto, I would make sure that we should not go back to LostDevice state.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -155,6 +155,15 @@ Button 1:
     * If pressed for more than three seconds, it initiates the factory reset of the device.
       Releasing the button within a 3-second window of the initiation cancels the factory reset procedure.
 
+LED 2:
+   If the :kconfig:option:`CONFIG_BRIDGED_DEVICE_BT` kconfig option is set to ``y`` shows the current state of Bluetooth LE Connectivity Manager state.
```
kkasperczyk-no, line 159: Could we not mention BLE Connectivity Manager name here? I think it's quite technically detailed information that is not relevant for UI description. I would suggest just calling it e.g. "current state of bridge's Bluetooth LE connectivity".

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -155,6 +155,15 @@ Button 1:
...
+   The following states are possible:
+
+   * Turned Off - The Bridge device is in the idle state and all connections to the Bluetooth LE bridged devices are stable.
```
kkasperczyk-no, line 162: Could we add one additional state? I'm missing here an information that bridge has any device paired. We could split it to:
* Turned Off - The Bridge device is in the idle state and has no Bluetooth LE devices paired.
* Turned On - The Bridge device is in the idle state and all connections to the Bluetooth LE bridged devices are stable.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -155,6 +155,15 @@ Button 1:
...
+   * Slow Even Flashing (1000 ms on / 100 ms off) - The Bridge device lost connection to at least one Bluetooth LE bridged device.
```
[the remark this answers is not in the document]
ArekBalysNordic: Yes, I forgot one "0" :)

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -32,6 +32,29 @@ class BLEConnectivityManager {
 	static constexpr uint16_t kMaxConnectedDevices = CONFIG_BT_MAX_CONN - 1;
 	static constexpr uint8_t kMaxServiceUuids = CONFIG_BT_SCAN_UUID_CNT;
 
+	/**
+	 * States for indicating the most important BLE Connectivity Manager states.
+	 * The priority is defined by the value in descending order, so Scanning (0) state has the highest one.
+	 * Only the active state with the highest priority is signalizing.
```
markaj-nordic, line 38:
```suggestion
	 * Only the active state with the highest priority is being signalized.
```

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -32,6 +32,29 @@ class BLEConnectivityManager {
...
+	 * Scanning - The Bluetooth scan is in progress.
+	 * Pairing - The pairing to the new device has been requested and Bridge is awaiting for PIN code.
+	 * LostDevice - The Bridge device lost connection to at least one bridged device.
+	 * Connected - All connections to bridged devices are stable.
+	 * Idle - There is no connections to bridged devices.
+	 * Unknown - None of those states are set.
+	 */
+	enum State : uint8_t { Scanning = 0, Pairing, LostDevice, Connected, Idle, Unknown };
+	/**
+	 * @brief Callback to indicate the highest state of the bridge's Bluetooth LE connectivity.
```
markaj-nordic, line 49:
```suggestion
	 * @brief Callback to indicate the bridge's Bluetooth LE connectivity state of the highest priority.
```

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -881,4 +919,23 @@ void BLEConnectivityManager::Recovery::StartTimer()
 	k_timer_start(&mRecoveryTimer, K_SECONDS(time), K_NO_WAIT);
 }
 
+BLEConnectivityManager::State BLEConnectivityManager::GetCurrentState()
+{
+	for (uint8_t s = State::Scanning; s <= State::Idle; s++) {
```
markaj-nordic, line 924: nit: we can extend State enum with State::Max

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -237,7 +270,11 @@ class BLEConnectivityManager {
 
 private:
 	bt_le_conn_param *GetScannedDeviceConnParams(bt_addr_le_t address);
+	State GetCurrentState();
+	void UpdateStateFlag(State newStateFlag, bool enabled);
 
+	StateChangedCallback mStateChanged = nullptr;
```
markaj-nordic, line 276:
```suggestion
	StateChangedCallback mStateChangedCb = nullptr;
```

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -881,4 +919,23 @@ void BLEConnectivityManager::Recovery::StartTimer()
...
+		}
+	}
+	return State::Unknown;
+}
+
+void BLEConnectivityManager::UpdateStateFlag(State newState, bool enabled)
+{
+	uint8_t newStateBitmask = enabled ? mStateBitmask | (1 << newState) : mStateBitmask & ~(1 << newState);
+	if (newStateBitmask != mStateBitmask && mStateChanged) {
+		mStateBitmask = newStateBitmask;
```
markaj-nordic, line 936: I guess we should update the `mStateChanged` regardless of the callback being nullptr or not.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -881,4 +919,23 @@ void BLEConnectivityManager::Recovery::StartTimer()
...
+{
+	for (uint8_t s = State::Scanning; s <= State::Idle; s++) {
+		if (mStateBitmask & (1 << s)) {
+			return static_cast<State>(s);
+		}
+	}
+	return State::Unknown;
+}
+
+void BLEConnectivityManager::UpdateStateFlag(State newState, bool enabled)
```
markaj-nordic, line 932: Should we deliberately treat some states as mutually exclusive?

ArekBalysNordic: I would like to do it as easily as possible, so I decided to support only the current state. Moreover, we can manage only the current action in the BLE Connectivity Manager, so we don't need to signalize more states simultaneously. We can discuss it f2f.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
ArekBalysNordic: I guess, you meant `mStateBitmask` instead of `mStateChanged` update. :)

wiba-nordic (approved):

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
markaj-nordic: Right :)

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -139,14 +139,21 @@ void BLEConnectivityManager::ConnectionHandler(bt_conn *conn, uint8_t conn_err)
 				Instance().Reconnect(reinterpret_cast<BLEBridgedDeviceProvider *>(context));
 			},
 			reinterpret_cast<intptr_t>(providerToRecover));
+		/* We have still a device to recover, keep the LostDevice state active */
+		Instance().UpdateStateFlag(State::LostDevice, true);
 	} else if (!sys_slist_is_empty(&Instance().mRecovery.mListToRecover)) {
 		/* There are pending providers to recover and no more scanned ones, schedule next scan operation. */
 		Instance().mRecovery.StartTimer();
+	} else if (provider->IsInitiallyConnected()) {
```
kkasperczyk-no, line 147: Why do we need `provider->IsInitiallyConnected()` check here? Wouldn't `else` be enough?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -460,9 +480,14 @@ CHIP_ERROR BLEConnectivityManager::Scan(ScanDoneCallback callback, void *context
 		mRecovery.CancelTimer();
 		ret = StopScan();
 		VerifyOrExit(ret == CHIP_NO_ERROR, );
+		/* A scan has been requested, activate the Scanning state */
+		Instance().UpdateStateFlag(State::Scanning, true);
 	} else if (mScanActive) {
 		LOG_ERR("Scan is already in progress");
 		return CHIP_ERROR_INCORRECT_STATE;
+	} else {
+		/* Th Bridge lost connection to the device, activate the LostDevice state */
```
kkasperczyk-no, line 489:
```suggestion
		/* The Bridge lost connection to the device, activate the LostDevice state */
```

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -722,6 +755,11 @@ CHIP_ERROR BLEConnectivityManager::RemoveBLEProvider(bt_addr_le_t address)
 		return CHIP_ERROR_INTERNAL;
 	}
 
+	if (0 == mConnectedProvidersCounter) {
+		/* There are no active providers anymore, so disable the Connected state */
+		UpdateStateFlag(State::Connected, false);
```
kkasperczyk-no, line 760: Maybe it's a minor comment, but according to description:

> Connected - All connections to bridged devices are stable.

shouldn't we then set `Connected` state to false on the first `Disconnect` event, not in the situation if all providers were removed?

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -578,6 +603,11 @@ CHIP_ERROR BLEConnectivityManager::StopScan()
 		return System::MapErrorZephyr(err);
 	}
 
+	if (mScanDoneCallback != ReScanCallback) {
+		/* Scanning has been finished, disable the Scanning state */
+		Instance().UpdateStateFlag(State::Scanning, false);
```
kkasperczyk-no, line 608: Shouldn't we disable Scanning in the scan callback? I think this is the moment that actually scanning is over.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
ArekBalysNordic: Why? Some of the existing providers can be still stable. If the Bridge loses connection, then the DeviceLost flag will be set, and since it has a bigger priority than the `Connected` state it will be signalized, If we remove this device the `DeviceLost` state will be disabled. If it still has some bridged device the `Connected` state will be signalized automatically after that. Otherwise, if we clear the `Connected` on the first `Disconnect` event, the Bridge won't signalize that other bridged devices are connected when all states with higher priority are disabled. So the description is also correct, because we can signalize the `Connected` state only if all devices are stable, and there is no scanning or pairing in progress.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
ArekBalysNordic: But, the scanning can be disabled before the scan callback is called, and what in that situation? Moreover, scanning is over when we call `bt_scan_stop`, not in the scan callback, so I guess the `StopScan` method is the best option to disable the `Scanning` state..

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
ArekBalysNordic: It isn't enough because without that we are going to clear the `LostDevice` state on each connection. So when we lose a device, the state is active, and then we connect another one, the state will be cleared, but we still haven't recovered the previous one.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: > But, the scanning can be disabled before the scan callback is called, and what in that situation?

Right, I actually mean we should disable it in both cases. 

> Moreover, scanning is over when we call bt_scan_stop, not in the scan callback

I don't think so, it's over when it was stopped manually or the scan timeout was reached, so we call `mScanDoneCallback`. I think we proactively always call `StopScan` in the `Connect` method just to not care if the scan is in progress or not.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
kkasperczyk-no: But why? Above we have two checks that make sure the list of devices to reconnect and the list of devices to recover are both empty. The else condition simply means there is nothing to be recovered, so why don't have to care if this device was previously connected or it is connected for the first time.

kkasperczyk-no (approved):

applications/matter_bridge/src/app_task.cpp
```diff
@@ -48,6 +48,42 @@ static bt_uuid *sUuidLbs = BT_UUID_LBS;
...
 static constexpr uint8_t kUuidServicesNumber = ARRAY_SIZE(sUuidServices);
+/**
+ * @brief Blink rates for indication the BLE Connectivity Manager state.
+ *
+ */
+constexpr static uint32_t kPairingBlinkRate{ 100 };
+constexpr static uint32_t kScanningBlinkRate_ms{ 300 };
+constexpr static uint32_t kLostBlinkRate_ms{ 1000 };
+
+void BLEStateChangeCallback(Nrf::BLEConnectivityManager::State state)
```
LuDuda, line 59: nit: Shouldn't be part of AppTask class? Or at least be a static function.

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -881,4 +922,26 @@ void BLEConnectivityManager::Recovery::StartTimer()
 	k_timer_start(&mRecoveryTimer, K_SECONDS(time), K_NO_WAIT);
 }
 
+BLEConnectivityManager::State BLEConnectivityManager::GetCurrentState()
+{
+	/* Iterate through states from the maximum priority to the minimum one */
+	for (uint8_t s = State::Max; s < State::Min; s++) {
```
LuDuda, line 928: nit: A bit counter intuitive, that Max is smaller than Min ;)

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -881,4 +922,26 @@ void BLEConnectivityManager::Recovery::StartTimer()
...
+		if (mStateBitmask & (1 << s)) {
+			return static_cast<State>(s);
+		}
+	}
+	return State::Unknown;
+}
+
+void BLEConnectivityManager::UpdateStateFlag(State newState, bool enabled)
```
LuDuda, line 936: If function updates the given state, i would not use `newState` name but rather just a `state` name.

samples/matter/common/src/bridge/ble_connectivity_manager.h
```diff
@@ -237,7 +270,11 @@ class BLEConnectivityManager {
 
 private:
 	bt_le_conn_param *GetScannedDeviceConnParams(bt_addr_le_t address);
+	State GetCurrentState();
+	void UpdateStateFlag(State newStateFlag, bool enabled);
```
LuDuda, line 274: nit: same comment here, it is not a `new` state, right?

LuDuda (approved):

applications/matter_bridge/src/app_task.cpp
ArekBalysNordic: It's already "static" since it is in the anonymous namespace.

PR #13980, opened 2024-02-13
shields: wifi: Modify default SPI clock frequency

bama-nordic:
Make the default SPI clock frequency as 16MHz on all nRF7002 EK shields

### DIFF
--- a/boards/shields/nrf7002ek/nrf7002ek.overlay
+++ b/boards/shields/nrf7002ek/nrf7002ek.overlay
@@ -12,7 +12,7 @@
 		compatible = "nordic,nrf700x-spi";
 		status = "okay";
 		reg = <0>;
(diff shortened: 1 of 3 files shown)


ioannisg:
Are you still working on this @bama-nordic ?

bama-nordic:
> Are you still working on this @bama-nordic ?

Looks like only 8MHz SPI clock is guaranteed to work across different chipsets 5340/52840/9160 and my thinking is to better keep 8MHz as default. If any customers want higher throughputs, we can use up to 16MHz based on the host SoC they are using and also the correct SPI# - only one of the SPIMs devices support higher than 8MHz clocks in our hosts (and I think 9160 only support up to 8MHz).

bama-nordic:
Keep default SPI clock to 8MHz for compatibility reasons

PR #13984, opened 2024-02-13
applications: nrf5340_audio: Move bitrate calculation into le_audio

andvib:
Move calculation of the codec bitrate into le_audio module to reduce code duplication.

The integration of sample rate conversion will introduce the bitrate calculation in three more places, so the gain from creating a function for this will increase when that is merged.

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
@@ -109,6 +109,31 @@ int le_audio_frame_blocks_per_sdu_get(const struct bt_audio_codec_cfg *codec,
 	return 0;
 }
 
+int le_audio_bitrate_get(const struct bt_audio_codec_cfg *const codec, uint32_t *bitrate)
+{
+	int ret;
+	int dur_us;
+
+	ret = le_audio_duration_us_get(codec, &dur_us);
+	if (ret) {
+		*bitrate = 0;
+		return ret;
+	}
+
+	int frames_per_sec = 1000000 / dur_us;
+	int octets_per_sdu;
+
+	ret = le_audio_octets_per_frame_get(codec, &octets_per_sdu);
+	if (ret) {
+		*bitrate = 0;
+		return ret;
+	}
+
+	*bitrate = frames_per_sec * (octets_per_sdu * 8);
+
+	return 0;
+}
+
 int le_audio_stream_dir_get(struct bt_bap_stream const *const stream)
 {
 	int ret;
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.h
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.h
@@ -111,7 +111,19 @@ int le_audio_octets_per_frame_get(const struct bt_audio_codec_cfg *codec, uint32
  */
 int le_audio_frame_blocks_per_sdu_get(const struct bt_audio_codec_cfg *codec,
 				      uint32_t *frame_blks_per_sdu);
-/*
+
+/**
+ * @brief	Get the bitrate for the codec configuration.
+ *
+ * @details	Decodes the audio frame duration and the number of octets per fram from the codec
+ *		configuration, and calculates the bitrate.
+ *
+ * @param[in]	codec	The audio codec structure.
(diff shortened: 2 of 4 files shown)


applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.h
```diff
@@ -111,7 +111,17 @@ int le_audio_octets_per_frame_get(const struct bt_audio_codec_cfg *codec, uint32
  */
 int le_audio_frame_blocks_per_sdu_get(const struct bt_audio_codec_cfg *codec,
 				      uint32_t *frame_blks_per_sdu);
-/*
+
+/**
+ * @brief	Decode the audio frame duration and number of octets per frame in the codec
```
koffes, line 116: "Decode" can confuse the reader. Suggest replacing with "get".

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.h
```diff
@@ -111,7 +111,17 @@ int le_audio_octets_per_frame_get(const struct bt_audio_codec_cfg *codec, uint32
...
+ *		configuration and return the corresponding bitrate.
+ *
+ * @param[in]	codec	The audio codec structure.
+ * @param[out]	bitrate	Pointer to the bitrate in bps.
+ */
+int le_audio_bitrate_get(const struct bt_audio_codec_cfg *codec, int *bitrate);
```
koffes, line 122:
```suggestion
int le_audio_bitrate_get(const struct bt_audio_codec_cfg const * const codec, int *bitrate);
```

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
```diff
@@ -109,6 +109,31 @@ int le_audio_frame_blocks_per_sdu_get(const struct bt_audio_codec_cfg *codec,
 	return 0;
 }
 
+int le_audio_bitrate_get(const struct bt_audio_codec_cfg *codec, int *bitrate)
```
koffes, line 112: Bitrate can only be positive, so can it be changed to uint32_t?

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
```diff
@@ -551,25 +555,11 @@ int unicast_server_config_get(uint32_t *bitrate, uint32_t *sampling_rate_hz,
 	}
 
 	if (bitrate != NULL) {
-		int dur_us;
-
-		ret = le_audio_duration_us_get(audio_streams[0].codec_cfg, &dur_us);
+		ret = le_audio_bitrate_get(audio_streams[0].codec_cfg, bitrate);
 		if (ret) {
-			LOG_ERR("Invalid frame duration: %d", ret);
+			LOG_ERR("Unable to get bitrate: %d", ret);
```
koffes, line 560: Nitpick: Do we need to translate the error here? Can we just do a `return ret`?

Also, the error prints should be the same on line 211 and here.

koffes (changes requested):

alexsven (approved):
Approved given that the comments from Kristoffer are fixed

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.h
andvib: I re-wrote it a bit so the brief says "get bitrate", and the details contain the "decode other stuff"

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
andvib: I'm not sure how valuable the error translation is actually. However, all the other `_get()` calls in this function does, and return the same error code, so I think it's better to keep the log message to not make the bitrate a special case within the function.

Updated the log message so it's the same as the other bitrate call.

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.h
koffes: Great.I still believe the use of the word "decode" can confuse

koffes (approved):

PR #13994, opened 2024-02-13
samples: matter: Matter simulated bridge with SMP DFU sample

HubertHugo:
Commit adds additional matter simulated bridge sample with smp dfu enabled.


doublemis1 (approved):
Check yaml linting issues.
Expect that, everything looks good to me

kkasperczyk-no (approved):

PR #13995, opened 2024-02-13
Bluetooth: Mesh: restore ttl in Time model after unsolicited statuses

alxelax:
Commit fixes the known issue when the Time model changes the configured publishing ttl after sending any
unsolicited Time status.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -85,6 +85,7 @@ Bluetooth Mesh
   * :ref:`bt_mesh_ug_reserved_ids` with model ID and opcodes for the new :ref:`bt_mesh_le_pair_resp_readme` model.
   * :ref:`bt_mesh_light_ctrl_readme` APIs to match new Sensor APIs.
   * :ref:`ug_bt_mesh_configuring` with the recommendation on how to configure persistent storage to increase performance.
+  * Fixed an issue when the Time Server model rewrites TTL to zero forever during the unsolicited Time Status publication.
 
 Matter
 ------
--- a/subsys/bluetooth/mesh/time_srv.c
+++ b/subsys/bluetooth/mesh/time_srv.c
@@ -416,6 +416,8 @@ static int bt_mesh_time_srv_init(const struct bt_mesh_model *model)
 	srv->model = model;
 	srv->data.timestamp = -STATUS_INTERVAL_MIN;
 	net_buf_simple_init(srv->pub.msg, 0);
+	srv->is_unsolicited = false;
+	srv->cached_ttl = 0;
 
 	k_work_init_delayable(&srv->status_delay, time_status_send_after_delay);
 
@@ -429,6 +431,8 @@ static void bt_mesh_time_srv_reset(const struct bt_mesh_model *model)
 
 	srv->data = data;
 	net_buf_simple_reset(srv->pub.msg);
+	srv->is_unsolicited = false;
+	srv->cached_ttl = 0;
 	(void)k_work_cancel_delayable(&srv->status_delay);
 
 	if (IS_ENABLED(CONFIG_BT_SETTINGS)) {
@@ -489,6 +493,7 @@ const struct bt_mesh_model_cb _bt_mesh_time_setup_srv_cb = {
 int _bt_mesh_time_srv_update_handler(const struct bt_mesh_model *model)
 {
 	struct bt_mesh_time_srv *srv = model->rt->user_data;
+	struct bt_mesh_model_pub *pub = srv->model->pub;
 	struct bt_mesh_time_status status;
 	int64_t uptime;
 	int err;
@@ -504,6 +509,15 @@ int _bt_mesh_time_srv_update_handler(const struct bt_mesh_model *model)
 		return err;
 	}
 
+	/* If sent as an unsolicited message, the Time Status message shall be sent
+	 * with TTL=0 to avoid building up cumulative time errors resulting from delays
+	 * in processing the messages by relays.
+	 */
+	if (!bt_mesh_model_pub_is_retransmission(pub->mod) && srv->is_unsolicited) {
+		pub->ttl = srv->cached_ttl;
+		srv->is_unsolicited = false;
+	}
+
 	srv->data.timestamp = uptime;
 	/* Account for delay in TX processing: */
 	status.uncertainty += CONFIG_BT_MESH_TIME_MESH_HOP_UNCERTAINTY;
@@ -517,6 +531,7 @@ int _bt_mesh_time_srv_update_handler(const struct bt_mesh_model *model)
 int bt_mesh_time_srv_time_status_send(struct bt_mesh_time_srv *srv,
 				      struct bt_mesh_msg_ctx *ctx)
 {
+	struct bt_mesh_model_pub *pub = srv->model->pub;
 	int64_t uptime = k_uptime_get();
 	int err;
 
@@ -528,7 +543,14 @@ int bt_mesh_time_srv_time_status_send(struct bt_mesh_time_srv *srv,
 		return -EOPNOTSUPP;
 	}
 
-	srv->model->pub->ttl = 0;
+	if (ctx) {
+		ctx->send_ttl = 0;
+		ctx->rnd_delay = false;
+	} else {
+		srv->cached_ttl = srv->is_unsolicited ? srv->cached_ttl : pub->ttl;
+		srv->is_unsolicited = true;
+		pub->ttl = 0;
+	}
 
 	err = send_time_status(srv->model, ctx, uptime);
 	if (!err) {
--- a/include/bluetooth/mesh/time_srv.h
+++ b/include/bluetooth/mesh/time_srv.h
@@ -107,6 +107,11 @@ struct bt_mesh_time_srv {
 	struct bt_mesh_time_srv_data data;
 	/** Delayable work to randomize status relaying. */
 	struct k_work_delayable status_delay;
+	/** Property whether the Time status was unsolicited. */
+	bool is_unsolicited;
+	/** Cached publishing TTL while server is sending the unsolicited Time Status with zero TTL.
+	 */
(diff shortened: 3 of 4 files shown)


m-alperen-sener (approved):

m-alperen-sener (approved):

m-alperen-sener:
Can not we embed 'is_unsolicited' and 'chached ttl into' into time server and use them on the moment the actual message is send out ?

subsys/bluetooth/mesh/time_srv.c
```diff
@@ -20,6 +20,9 @@ struct bt_mesh_time_srv_settings_data {
 	uint64_t tai_of_delta_change;
 };
 
+static bool is_unsolicited;
+static uint8_t cached_ttl;
```
PavelVPV, line 24: How does this work for multiple time server model instances?

subsys/bluetooth/mesh/time_srv.c
```diff
@@ -504,6 +512,16 @@ int _bt_mesh_time_srv_update_handler(const struct bt_mesh_model *model)
 		return err;
 	}
 
+	/* If sent as an unsolicited message, the Time Status message shall be sent
+	 * with TTL=0 to avoid building up cumulative time errors resulting from delays
+	 * in processing the messages by relays.
+	 */
+	if (!bt_mesh_model_pub_is_retransmission(pub->mod)) {
+		cached_ttl = pub->ttl ? pub->ttl : cached_ttl;
+		pub->ttl = is_unsolicited ? 0 : cached_ttl;
```
PavelVPV, line 521: `bt_mesh_time_srv_time_status_send` calls `send_time_status` which calls `bt_mesh_msg_send`. `bt_mesh_msg_send` uses `bt_mesh_model_send` instead of `bt_mesh_model_publish` when message context is provided:
https://github.com/zephyrproject-rtos/zephyr/blob/080d51bb367836b73021fdd7ade3e2c4c54a0435/subsys/bluetooth/mesh/msg.c#L97-L99
It may happen that a user calls `bt_mesh_time_srv_time_status_send` with non-NULL `ctx`, which will set `is_unsolicited` to `true`. In this case the message will be passed to the transport layer directly bypassing the publication mechanism, and this callback won't be triggered. Then later if periodic publication is configured this callback will be called, but `is_unsolicited` will be set to `true` which in turns will set `ttl` to zero. I guess this is not what we want here.

PavelVPV (changes requested):

subsys/bluetooth/mesh/time_srv.c
```diff
@@ -504,6 +509,16 @@ int _bt_mesh_time_srv_update_handler(const struct bt_mesh_model *model)
...
 
+	/* If sent as an unsolicited message, the Time Status message shall be sent
+	 * with TTL=0 to avoid building up cumulative time errors resulting from delays
+	 * in processing the messages by relays.
+	 */
+	if (!bt_mesh_model_pub_is_retransmission(pub->mod)) {
+		srv->cached_ttl = pub->ttl ? pub->ttl : srv->cached_ttl;
+		pub->ttl = srv->is_unsolicited ? 0 : srv->cached_ttl;
+		srv->is_unsolicited = false;
+	}
```
PavelVPV, line 520: I'm still uncertain if this is correct. When `bt_mesh_model_publish` is called by a model, `pub->count` is set to a non-zero value:
https://github.com/zephyrproject-rtos/zephyr/blob/f9778472105d756fff7d1e5b54353421d356ed43/subsys/bluetooth/mesh/access.c#L1625
https://github.com/zephyrproject-rtos/zephyr/blob/f9778472105d756fff7d1e5b54353421d356ed43/include/zephyr/bluetooth/mesh/access.h#L680
In this case, the update callback is only called for retransmissions:
https://github.com/zephyrproject-rtos/zephyr/blob/f9778472105d756fff7d1e5b54353421d356ed43/subsys/bluetooth/mesh/access.c#L927-L953

What we probably what here is in the `bt_mesh_time_srv_time_status_send` function store the current publication ttl value, set `srv->model->pub->ttl` to zero like it was before and set the `is_unsolicited` flag. Then, when the update callback is called, `bt_mesh_model_pub_is_retransmission` returns false and `is_unsolicited` is true, we restore the old ttl and drop the `is_unsolicited` flag.

PavelVPV: > The update callback is unconditionally called for the first publication, always here:
> https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/bluetooth/mesh/access.c#L861
> over this call:
> https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/bluetooth/mesh/access.c#L939-L940
> exactly in `pub_period_start` function counter is configured if retransmission was configured:
> https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/bluetooth/mesh/access.c#L855

No, this is not correct because as I said, `pub->count` is already set. The idea here is that a model feels up the publication buffer in advance before calling `bt_mesh_model_publish`. Look at how `bt_mesh_msg_send` does this: https://github.com/zephyrproject-rtos/zephyr/blob/793c507209eb508c1c29441b56d60303df4187a7/subsys/bluetooth/mesh/msg.c#L101-L104

alxelax: Hm... I guess the line that you pointed here: https://github.com/zephyrproject-rtos/zephyr/blob/f9778472105d756fff7d1e5b54353421d356ed43/subsys/bluetooth/mesh/access.c#L1625
is access layer bug. 
If publication goes over API (not over timer) it skips the first attempt and goes directly to retransmissions.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -85,6 +85,7 @@ Bluetooth Mesh
   * :ref:`bt_mesh_ug_reserved_ids` with model ID and opcodes for the new :ref:`bt_mesh_le_pair_resp_readme` model.
   * :ref:`bt_mesh_light_ctrl_readme` APIs to match new Sensor APIs.
   * :ref:`ug_bt_mesh_configuring` with the recommendation on how to configure persistent storage to increase performance.
+  * Fixed known issue when Time model rewrote TTL to zero forever during unsolicited Time Status publication.
```
mia-ko, line 88:
```suggestion

* Fixed an issue when the Time Server model rewrites TTL to zero forever during the unsolicited Time Status publication.
```

mia-ko (changes requested):

PavelVPV (approved):

m-alperen-sener (approved):

ludvigsj (commented):
Note: Changing `srv->mod->pub->ttl` state like this could cause issues with some third party config clients. We had a request to avoid similar behavior in the EnOcean server a bit over a year back, where a third party config client would get confused when it configured a value (such as TTL in this case) and then got a different value back when it sent a Model Publication Get some time later. Since this model already is rewriting this `pub->ttl` even without this PR, I think it can go in as is, but we should probably make a separate issue about this fact.

alxelax:
> Note: Changing `srv->mod->pub->ttl` state like this could cause issues with some third party config clients. We had a request to avoid similar behavior in the EnOcean server a bit over a year back, where a third party config client would get confused when it configured a value (such as TTL in this case) and then got a different value back when it sent a Model Publication Get some time later. Since this model already is rewriting this `pub->ttl` even without this PR, I think it can go in as is, but we should probably make a separate issue about this fact.

I do not think you are right. It rewrites only for sending. My PR adds restoring ttl value, exactly to not confuse third-party config clients too. Config client cannot get a response until publishing is in progress

ludvigsj:
> I do not think you are right. It rewrites only for sending. My PR adds restoring ttl value, exactly to not confuse third-party config clients too. Config client cannot get a response until publishing is in progress

Can you explain how this works? The way I can tell, TTL is restored in the pub update handler, and it is set to 0 when `bt_mesh_time_srv_time_status_send`. The pub update handler is called on the periodic publication schedule(?). Meaning, in the time between a call to `bt_mesh_time_srv_time_status_send` and the next periodic publication (which could be more than enough time to receive a Model Publication Get in between), a Model Publication Status message would report TTL=0?

PavelVPV:
> Scratch this, I missed the fact that `bt_mesh_model_msg_send` eventually will call the update handler as well. Will approve now.

What? You are right that until the next publication period starts, ttl will be zero and Config Client will get this value. However, does this really a problem? If so, then the access layer behavior should be changed.

ludvigsj:
> > Scratch this, I missed the fact that `bt_mesh_model_msg_send` eventually will call the update handler as well. Will approve now.
> 
> What? You are right that until the next publication period starts, ttl will be zero and Config Client will get this value. However, does this really a problem? If so, then the access layer behavior should be changed.

Yes, but when "next publication starts" is almost immediately, since `bt_mesh_model_msg_send` calls `bt_mesh_model_publish` internally, which schedules `mod_publish` with no wait, and `mod_publish` calls the update handler. Which is not a problem, I thought the next call to `mod_publish` (and the update handler) would be at the next _periodic_ publication, tens of seconds later, but this was wrong.

alxelax:
Yes, Config client might receive ttl 0 in some time range between unsolicited and periodic publishing. However, you need to consider this is a minor issue(not sure this is the issue at all) comparing with the current behavior. At the moment Time model stops working as Time Authority and Time Relay after the first unsolicited publication. In general, Time model and as consequence Scheduler model are not functional as network functionality.

PavelVPV:
>  `mod_publish` calls the update handler. 

No, it is not like that. When `bt_mesh_model_publish` is called, the update handler is not called immediately (except the retransmission case). I explained this in this PR in this comment: https://github.com/nrfconnect/sdk-nrf/pull/13995#discussion_r1490659395

ludvigsj:
Right, seems the issue I noted with "wrong" Model Publication Status is actually there then. But I agree, it does not relate to this PR. Will approve now and create a separate issue for this, then we can decide what to do with it.

ludvigsj (approved):

mia-ko (approved):

PR #13997, opened 2024-02-13
lib: sample_rate_converter: Small filters for conversion

andvib:
Add small filters for sample rate conversion. These filters are meant to use little space while still having some low-pass filter capabilites.


lib/sample_rate_converter/sample_rate_converter_filter.c
```diff
@@ -71,6 +164,23 @@ int sample_rate_converter_filter_get(enum sample_rate_converter_filter filter_ty
...
+			*filter_ptr = filter_24khz_to_48khz_16bit_small;
+			*filter_size = ARRAY_SIZE(filter_24khz_to_48khz_16bit_small);
+		} else if (conversion_ratio == -3) {
+			*filter_ptr = filter_48khz_to_16khz_16bit_small;
+			*filter_size = ARRAY_SIZE(filter_48khz_to_16khz_16bit_small);
+		} else if (conversion_ratio == 3) {
+			*filter_ptr = filter_16khz_to_48khz_16bit_small;
+			*filter_size = ARRAY_SIZE(filter_16khz_to_48khz_16bit_small);
+		}
+		break;
```
alexsven, line 182: Should be a switch case, and you could create an enum for the values -3, -2, 2 ,and 3

alexsven (changes requested):

include/sample_rate_converter.h
```diff
@@ -39,7 +39,8 @@
 
 /** Filter types supported by the sample rate converter */
 enum sample_rate_converter_filter {
-	SAMPLE_RATE_FILTER_TEST = 1
+	SAMPLE_RATE_FILTER_TEST = 1,
+	SAMPLE_RATE_FILTER_SMALL
```
koffes, line 43: Consideration: Can/should these be named "simple"? Since they reduce both size and time, "small" is not that descriptive.

lib/sample_rate_converter/sample_rate_converter_filter.c
```diff
@@ -32,6 +32,99 @@ static q31_t filter_16khz_to_48khz_32bit_test[] = {0x7fffffff, 0x7fffffff, 0x7ff
 #endif
 #endif /* CONFIG_SAMPLE_RATE_CONVERTER_FILTER_TEST */
 
+#ifdef CONFIG_SAMPLE_RATE_CONVERTER_FILTER_SMALL
+#ifdef CONFIG_SAMPLE_RATE_CONVERTER_BIT_DEPTH_16
+static q15_t filter_48khz_to_16khz_16bit_small[] = {
```
koffes, line 37: I think all filters should be `static const`

lib/sample_rate_converter/Kconfig
```diff
@@ -27,8 +27,16 @@ config SAMPLE_RATE_CONVERTER_FILTER_TEST
...
 	  output.
 
+config SAMPLE_RATE_CONVERTER_FILTER_SMALL
+	bool "Include the small sample rate converter filters"
+	help
+	  Includes the small filters for the sample rate converter. The small filters uses a small
+	  amount of space and time for the conversion, while also giving some low-pass filter
+	  capabilities.
+
 config SAMPLE_RATE_CONVERTER_MAX_FILTER_SIZE
```
koffes, line 37: Do we need this KConfig at all?
The preprocessor knows the filter sizes so we may use` MAX(ARRAY_SIZE(filter_A), ARRAY_SIZE(filter_B), ...) `

tests/lib/sample_rate_converter/src/main.c
```diff
@@ -899,6 +899,41 @@ ZTEST(suite_sample_rate_converter, test_init_invalid_sample_rates_equal)
...
+	uint32_t conversion_ratio = input_sample_rate / output_sample_rate;
+
+	uint16_t input_samples[] = {1000, 2000, 3000, 4000,  5000,  6000,
+				    7000, 8000, 9000, 10000, 11000, 12000};
+	size_t num_samples = ARRAY_SIZE(input_samples);
+	size_t expected_output_samples = num_samples / conversion_ratio;
+	uint16_t output_samples[expected_output_samples];
+	size_t output_written;
+
+	enum sample_rate_converter_filter original_filter = SAMPLE_RATE_FILTER_SMALL;
```
koffes, line 917: Would appreciate a walkthrough of this test.

koffes (changes requested):

include/sample_rate_converter.h
andvib: Good point, agreed!

lib/sample_rate_converter/Kconfig
andvib: Using `MAX()` would be great, but I couldn't find a way to get this to work as filters can be `#ifdef`ed away. If you have some ideas we could sit down and discuss offline!

alexsven (approved):

koffes (approved):
Approved, but two things I would like to discuss.

PR #14001, opened 2024-02-14
samples: wifi: softap: Modify default values for SoftAP configuration

D-Triveni:
Update default settings for SoftAP to use WPA2 security by default.


divyagona (commented):
Default settings as wpa2-psk and ssid/password- Myssid/Mypassword settings are working.

divyagona (approved):

divyagona (commented):
Default regulatory in sample is "world" band-2.4GHz which is a working combination.
For 5GHz "world" regulatory will not work.
 
 May we add a note in the documentation about mandate regulatory domain set to specific country  for 5GHz?

krga2022 (approved):

richabp (approved):

PR #14009, opened 2024-02-14
samples: matter: Removed _NUS suffix from BT device name

kkasperczyk-no:
For some reason the BT NUS module adds _NUS suffix to BT name, while it creates own advertising data. It is sort of confusing, because the device changes BT name after Matter BLE advertising timeout and switching to advertising NUS.


Damian-Nordic (approved):

doublemis1:
Do we need to update that in the documentation.

PR #14016, opened 2024-02-15
Promiscuous mode checkin

VivekUppunda:
This set of changes adds promiscuous mode support to nrf Wi-Fi driver.

### DIFF
--- a/drivers/wifi/nrf700x/Kconfig
+++ b/drivers/wifi/nrf700x/Kconfig
@@ -84,6 +84,11 @@ config NRF700X_RAW_DATA_RX
 	bool "Enable RAW RX sniffer operation in the driver"
 	depends on NRF700X_SYSTEM_MODE
 
+config NRF700X_PROMISC_DATA_RX
+	bool "Enable promiscuous RX sniffer operation in the driver"
+	select EXPERIMENTAL
+	depends on NRF700X_SYSTEM_MODE
+
 config NRF_WIFI_IF_AUTO_START
 	bool "Enable Wi-Fi interface auto start on boot"
 	default y
--- a/drivers/wifi/nrf700x/src/wifi_mgmt.c
+++ b/drivers/wifi/nrf700x/src/wifi_mgmt.c
@@ -79,6 +79,9 @@ int nrf_wifi_set_power_save(const struct device *dev,
 #ifdef CONFIG_NRF700X_RAW_DATA_TX
 		    && (vif_ctx_zep->if_type != NRF_WIFI_STA_TX_INJECTOR)
 #endif /* CONFIG_NRF700X_RAW_DATA_TX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+		    && (vif_ctx_zep->if_type != NRF_WIFI_STA_PROMISC_TX_INJECTOR)
+#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
 		   ) {
 			LOG_ERR("%s: Operation supported only in STA enabled mode",
 				__func__);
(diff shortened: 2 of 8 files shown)


drivers/wifi/nrf700x/src/wifi_mgmt.c
```diff
@@ -791,11 +797,19 @@ int nrf_wifi_mode(const struct device *dev,
 
 	} else {
 		mode->mode = def_dev_ctx->vif_ctx[vif_ctx_zep->vif_idx]->mode;
-
+		/**
+		 * This is a work-around to handle current UMAC mode handling.
+		 * This might be removed in future versions when UMAC has more space.
+		 */
 #if defined(CONFIG_NRF700X_RAW_DATA_TX) || defined(CONFIG_NRF700X_RAW_DATA_RX)
```
sachinthegreen, line 804:
```suggestion
#ifdef(CONFIG_NRF700X_RAW_DATA_TX)
```

drivers/wifi/nrf700x/src/wifi_mgmt.c
```diff
@@ -791,11 +797,19 @@ int nrf_wifi_mode(const struct device *dev,
...
 		if (def_dev_ctx->vif_ctx[vif_ctx_zep->vif_idx]->txinjection_mode == true) {
 			mode->mode ^= NRF_WIFI_TX_INJECTION_MODE;
 		}
+#endif
```
sachinthegreen, line 808:
```suggestion
#endif /* CONFIG_NRF700X_RAW_DATA_TX */
```

drivers/wifi/nrf700x/src/wifi_mgmt.c
```diff
@@ -791,11 +797,19 @@ int nrf_wifi_mode(const struct device *dev,
...
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+		if (def_dev_ctx->vif_ctx[vif_ctx_zep->vif_idx]->promisc_mode == true) {
+			mode->mode ^= NRF_WIFI_PROMISCUOUS_MODE;
+		}
 #endif
```
sachinthegreen, line 813:
```suggestion
#endif /* CONFIG_NRF700X_PROMISC_DATA_RX */
```

sachinthegreen (changes requested):

sachinthegreen (approved):

drivers/wifi/nrf700x/Kconfig
```diff
@@ -84,6 +84,10 @@ config NRF700X_RAW_DATA_RX
 	bool "Enable RAW RX sniffer operation in the driver"
 	depends on NRF700X_SYSTEM_MODE
 
+config NRF700X_PROMISC_DATA_RX
+	bool "Enable RAW promiscuous RX sniffer operation in the driver"
```
krish2718, line 88:
```suggestion
	bool "Enable promiscuous RX sniffer operation in the driver"
```

drivers/wifi/nrf700x/src/wifi_mgmt.c
```diff
@@ -791,19 +797,27 @@ int nrf_wifi_mode(const struct device *dev,
...
+		 */
+#ifdef CONFIG_NRF700X_RAW_DATA_TX
 		if (def_dev_ctx->vif_ctx[vif_ctx_zep->vif_idx]->txinjection_mode == true) {
 			mode->mode ^= NRF_WIFI_TX_INJECTION_MODE;
 		}
-#endif
+#endif /* CONFIG_NRF700X_RAW_DATA_TX */
+#ifdef CONFIG_NRF700X_PROMISC_DATA_RX
+		if (def_dev_ctx->vif_ctx[vif_ctx_zep->vif_idx]->promisc_mode == true) {
+			mode->mode ^= NRF_WIFI_PROMISCUOUS_MODE;
```
krish2718, line 811: This is the only place where the define is used as standalone, so, do we really need a separate define for promiscuous ? Can't we use RAW_DATA_RX for both and rely on `mode` runtime setting?

VivekUppunda: It has other checks specifically for Promiscuous mode. One of the ask initially was to keep different implementations completely code configurable during build. It was as per that discussion that Monitor, Promiscuous and TX Injection modes are kept separate.

krish2718: > It has other checks specifically for Promiscuous mode

Ok, in the driver?

> It was as per that discussion that Monitor, Promiscuous and TX Injection modes are kept separate.

Anyways, I see a lot of duplication b/w them, so, we should re-discuss, but we can proceed with the PR.

kapbh (approved):

PR #14022, opened 2024-02-15
applications: nrf5340_audio: Extended codec cfg max metadata size

rick1082:
Enlarge codec cfg max metadata size to prevent compatibility issue. 
OCT-NONE

### DIFF
--- a/applications/nrf5340_audio/broadcast_sink/Kconfig.defaults
+++ b/applications/nrf5340_audio/broadcast_sink/Kconfig.defaults
@@ -50,6 +50,8 @@ config BT_PAC_SNK
 config BT_AUDIO_RX
 	default y
 
+config BT_AUDIO_CODEC_CFG_MAX_METADATA_SIZE
+	default 80
 
 ## Audio related configs ##
 config AUDIO_MUTE


alexsven:
We should increase this for all applications, so may want to set it in the Kconfig for bt_stream (have to be created)

alexsven (changes requested):
We should increase this for all applications, so may want to set it in the Kconfig for bt_stream (have to be created)

rick1082:
> We should increase this for all applications, so may want to set it in the Kconfig for bt_stream (have to be created)
This part is dedicated for broadcast sink.
The needed of metadata size would be different between BIS and CIS I think.
Since there's no handshaking between broadcast sink and source, so the broadcast source has to expose all the information from different level to metadata at once. So we have to set BT_AUDIO_CODEC_CFG_MAX_METADATA_SIZE bigger for broadcast sink, otherwise we cannot decode it properly.

For CIS, setting BT_AUDIO_CODEC_CFG_MAX_METADATA_SIZE to a higher number will also need to change CONFIG_BT_ATT_PREPARE_COUNT, which is the buffer for long write.

alexsven (approved):

PR #14028, opened 2024-02-15
applications: nrf5340_audio: Change default controller in buildprog t…

gWacey:
OCT-2881 and OCT-2880

Change the default controller to the SDC in buildprog and Kconfig.defaults. Also deprecate BT_LL_ACS_NRF43.

Added comment:
* The SoftDevice Controller is the default controller for the build script.
to the file /nrf/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst

BT_LL_ACS_NRF53 has been marked as deprecated, will need to remove all code within #ifdef CONFIG_CONFIG_BT_LL_ACS_NRF53 for release 2.7.


applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -113,10 +113,10 @@ def __build_cmd_get(core: Core, device: AudioDevice, build: BuildType, pristine,
             device_flag += " -DCONFIG_AUDIO_DFU=2"
         if options.min_b0n:
             device_flag += " -DCONFIG_B0N_MINIMAL=y"
-        if options.controller == Controller.sdc:
-            device_flag += " -DCONFIG_BT_LL_ACS_NRF53=n"
-            if not child_image:
-                device_flag += " -DCONFIG_NCS_INCLUDE_RPMSG_CHILD_IMAGE=n"
+        if options.controller == Controller.acs_nrf53:
```
koffes, line 116: Nice. Have you run tests with this and tried with mcuboot flags? These must be disabled for SDC.

koffes:
Regarding the changelog. Yes. We could also handle that in a different PR, but IMO we can gather it here. We need to add it to the changelog and update the documentation. When this is done we add Gregfer as a reviwer.

applications/nrf5340_audio/tools/buildprog/buildprog.py
gWacey: Yes, the code builds. When the SDC is default there is no change to the script in regard to the mcuboot flags, with the result that the code will not build if mcuboot is set to anything for a default build. These flags, including the b0n flags, can only be set if the controller is ACS_nRF53.

alexsven (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -255,6 +255,7 @@ Serial LTE modem
 nRF5340 Audio
 -------------
 
+* The SoftDevice Controller is the default controller for the build script.
```
koffes, line 258:
```suggestion
* The Nordic Semiconductor SoftDevice Controller (:ref:`ug_ble_controller_softdevice`) has now replaced the LE Audio Controller Subsystem for nRF5340 (The :ref:`_lib_bt_ll_acs_nrf53_readme`) as the default controller. For NCS 2.6.0 and future NCS releases, users must transition to and use the SoftDevice Controller. There should be no negative impact on performance for users of the nRF5340 Audio application. This change will enable use of standard tools for building and configuring the controller. Note that the SoftDevice controller will for some time not have a QDID for ISO operation.
```

koffes (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
greg-fer: Most of the information from @koffes edit should go to https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/releases_and_maturity/migration/migration_guide_2.6.html and be added there in an entry about the Audio applications.

I suggest keeping it short and simple here in the changelog:
```suggestion
* Changed the default controller from the :ref:`_lib_bt_ll_acs_nrf53_readme` to the :ref:`ug_ble_controller_softdevice`. Read :ref:`migration_2.6` for information about how this affects your application.
```

And then in the migration guide doc:
```
* The :ref:`nrf53_audio_app` have changed the default controller from the :ref:`_lib_bt_ll_acs_nrf53_readme` to Nordic Semiconductor's :ref:`ug_ble_controller_softdevice`. Make sure to transition to using the SoftDevice Controller. There should be no negative impact on performance for the nRF5340 Audio applications. This change will enable use of standard tools for building and configuring the controller. The SoftDevice Controller will for some time not have a QDID for the ISO operation. 
```

greg-fer (changes requested):

rick1082 (changes requested):
Excuse me for one request on this:
It would be great if we can print a log for showing which controller we are going to build for the code. And for SDC, we need to add "experimental" for reminding the user.

applications/nrf5340_audio/src/bluetooth/bt_management/controller_config/Kconfig.defaults
```diff
@@ -1,22 +0,0 @@
-#
```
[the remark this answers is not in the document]
gWacey: This is also for OCT-2880 which is to deprecate the BT_LL_ACS_NRF53.

gWacey:
> Excuse me for one request on this: It would be great if we can print a log for showing which controller we are going to build for the code. And for SDC, we need to add "experimental" for reminding the user.

By default now that SDC is used there is a yellow print that the controller is experimental.

koffes:
Replaced by: https://github.com/nrfconnect/sdk-nrf/pull/14204

PR #14034, opened 2024-02-15
modules: trusted-firmware-m: HMAC DRBG for minimal

Vge0rge:
Enable HMAC DRBG when we build TFM with minimal profile since it reduces the ROM usage.

Ref: NCSDK-25891


vili-nordic (approved):

mswarowsky (approved):
Nice catch

We should also move the builds that was quarantined due to FLASH issues out of quarantine. 
But I think this is not done in sdk-nrf

frkv (approved):
LGTM

PR #14036, opened 2024-02-15
modules: hostap: Fix key management for SAP

krish2718:
Key management for SAP should be retrieved from the network block rather than using the per-interface global structure.

Fixes SHEL-2466.


sachinthegreen (approved):

divyagona (approved):
Working as expected. wifi status display is correct

krga2022 (approved):

PR #14043, opened 2024-02-16
Increase the default TF-M crypto partition stack size when 4096 bit RSA keys are enabled

vili-nordic:


samples/crypto/rsa/src/key.h
```diff
@@ -0,0 +1,207 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
```
mswarowsky, line 2:
```suggestion
 * Copyright (c) 2024 Nordic Semiconductor ASA
```

mswarowsky (approved):
One small nit, but rest lgtm

samples/crypto/rsa/README.rst
```diff
@@ -26,13 +26,13 @@ The sample performs the following operations:
 1. Initialization:
 
    a. The Platform Security Architecture (PSA) API is initialized.
-   #. A random RSA key pair is generated and imported into the PSA crypto keystore.
-   #. The public key of the RSA key pair is imported into the PSA crypto keystore.
+   #. A RSA key pair is imported into the PSA crypto keystore.
```
frkv, line 29: It is unfortunate that we have to rely on a pre-generated key, but I think it should be mentioned that this key is not usable in any end-products. Maybe this text should be given in italics

```suggestion
   #. A RSA key pair is imported into the PSA crypto keystore.
   #. *Note that this is a test key and it should never be used in a product!*
```

SebastianBoe (commented):
The title might mislead a customer into thinking using 4096 bit keys was not supported with TF-M previously.

The customer could have 4096 bit keys in his TF-M application by doing

CONFIG_TFM_CRYPTO_PARTITION_STACK_SIZE=0x4000

maybe:

Increase the default TF-M crypto partition stack size when 4096 bit RSA keys are enabled

SebastianBoe (approved):

samples/crypto/rsa/README.rst
```diff
@@ -26,13 +26,15 @@ The sample performs the following operations:
...
```
mia-ko, line 29:
```suggestion
   #. An RSA key pair is imported into the PSA crypto keystore.
```

samples/crypto/rsa/README.rst
```diff
@@ -26,13 +26,15 @@ The sample performs the following operations:
...
+   #. The public key is imported into the PSA crypto keystore.
+
+   *Note that the keys used in the sample are for demonstration purposes only and never be used in a product.*
```
mia-ko, line 32:
```suggestion
   #. The public key is imported into the PSA crypto keystore.

      .. note::
         The keys used in the sample are for demonstration purposes only and must never be used in a product.
         ```

**Review by mia-ko — CHANGES_REQUESTED** [review]

File: samples/crypto/rsa/README.rst [hunk 5]
```diff
@@ -26,13 +26,17 @@ The sample performs the following operations:
...
 
    a. The Platform Security Architecture (PSA) API is initialized.
-   #. A random RSA key pair is generated and imported into the PSA crypto keystore.
-   #. The public key of the RSA key pair is imported into the PSA crypto keystore.
+   #. An RSA key pair is imported into the PSA crypto keystore.
+   #. The public key is imported into the PSA crypto keystore.
+
+      .. note::
+         The keys used in the sample are for demonstration purposes only and must never be used in a product.
+         ```
```
mia-ko, line 34: These sneaked in, you probably want to remove them :)

mia-ko (commented):
Do these changes require a changelog update?

vili-nordic:
I don't think updating the default stack-size for tf-m require changelog entry. The sample is showing something slightly different now (but probably a more "common" case), but I'm not sure how relevant that is for a changelog.

PR #14050, opened 2024-02-16
tfm: Delete deprecated symbol PSA_INITIAL_ATTEST_TOKEN_MAX_SIZE

SebastianBoe:
PSA_INITIAL_ATTEST_TOKEN_MAX_SIZE was renamed in TF-M, but we didn’t
update our symbol name.

It appears that this symbol is now provided by TF-M's
initial_attestation.h.in and that we should therefore delete, not
rename, this symbol.

Interestingly, it’s documented that region_defs.h should define this,
but no-one is doing so for the new symbol name. The docs are probably
wrong.


frkv (approved):
LGTM

vili-nordic (approved):

PR #13662, opened 2024-01-18
TF-M 2.0.0 + Mbed TLS 3.5.2 upmerge+ Oberon PSA core 1.2.1 + ocrypto 3.0.14 + cc3xx library 0.9.19 update

mswarowsky:
test-sdk-nrf: test-nrf-pr-13662

PR #14053, opened 2024-02-16
actions: oss: disable tf-m oss-history check

thst-nordic:
This does the same as commit
20e50c32958465916240b167a07bee9e82af02bc
which was reverted by
8042ef1ed8a2d128f7ad1683e1ea2efd4d59b371

Pull request #13662 is bringing in a lot of commits from upstream tf-m
that are not present in the revision of tf-m in zephyr/west.yml, and
the script is getting confused.

In particular, it thinks that all of the new upstream commits are
actually downstream out-of-tree commits. Then it tries to reapply them
onto the old upstream. This fails, and the action fails.

Signed-off-by: Markus Swarowsky <markus.swarowsky@nordicsemi.no>

### DIFF
--- a/.github/workflows/oss-history.yml
+++ b/.github/workflows/oss-history.yml
@@ -31,4 +31,4 @@ jobs:
         uses: nrfconnect/action-oss-history@main
         with:
           workspace: 'ncs'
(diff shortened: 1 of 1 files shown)


mswarowsky:
Got merged as part of #13662

PR #14065, opened 2024-02-18
modules: hostap: SAP mode needs more heap

krish2718:
SAP mode adds more functionality that increases the heap usage, based on experiments, 40000 works well to avoid startup failures.


divyagona (approved):
working in wpa2-sha256 ,wpa3 and open mode

sachinthegreen (approved):

PR #14074, opened 2024-02-19
Bluetooth: Mesh: Add cmd to print DFU metadata

Andrewpini:
Adds utility command that prints BLE Mesh DFU metadata for the target application folder.

### DIFF
--- a/cmake/mesh_dfu_metadata.cmake
+++ b/cmake/mesh_dfu_metadata.cmake
@@ -12,7 +12,7 @@ function(mesh_dfu_metadata)
         COMMAND
         ${PYTHON_EXECUTABLE}
         ${ZEPHYR_NRF_MODULE_DIR}/scripts/bluetooth/mesh/mesh_dfu_metadata.py
-        ${PROJECT_BINARY_DIR}
+        --bin-path ${PROJECT_BINARY_DIR}
         DEPENDS
         ${PROJECT_BINARY_DIR}/dfu_application.zip
     )
@@ -22,4 +22,16 @@ function(mesh_dfu_metadata)
         ALL
         DEPENDS ${PROJECT_BINARY_DIR}/dfu_application.zip_ble_mesh_metadata.json
     )
+
+    add_custom_target(
+        # Prints already generated metadata
+        ble_mesh_dfu_metadata
+        COMMAND
+        ${PYTHON_EXECUTABLE}
+        ${ZEPHYR_NRF_MODULE_DIR}/scripts/bluetooth/mesh/mesh_dfu_metadata.py
+        --bin-path ${PROJECT_BINARY_DIR}
+        --print-metadata
+        COMMAND_EXPAND_LISTS
+        )
+
 endfunction()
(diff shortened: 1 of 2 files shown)


scripts/west-commands.yml
```diff
@@ -16,3 +16,8 @@ west-commands:
       - name: ncs-sbom
         class: NcsSbom
         help: generate Software Bill Of Materials including license information
+  - file: scripts/west_commands/bluetooth/mesh/dfu_metadata.py
```
PavelVPV, line 19: Hmm, I was actually thinking about something like this:
https://github.com/nrfconnect/sdk-nrf/blob/7d209a57228062b2453d8ba25b172e8d09a26d49/cmake/partition_manager.cmake#L708-L715
And then user calls `west build -t ble_mesh_metadata_report` or something. Script also knows about the path to the json and as I understand won't work if firmware is not build.

PavelVPV: And perhaps the same script which generates metadata can be used.

Balaklaka (commented):
Added this to the existing script sound like a good idea.

scripts/west-commands.yml
Andrewpini: Ok, I'm on it 👍

cmake/mesh_dfu_metadata.cmake
```diff
@@ -22,4 +22,16 @@ function(mesh_dfu_metadata)
         ALL
         DEPENDS ${PROJECT_BINARY_DIR}/dfu_application.zip_ble_mesh_metadata.json
     )
+
+    add_custom_target(
+        # Prints already generated metadata
+        ble-mesh-dfu-metadata
```
PavelVPV, line 28: From what I see, names for custom commands use underscore instead of dash (except docs).

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -372,16 +373,32 @@ def encoded_metadata_get(version, comp):
...
+def input_parse():
+    parser = argparse.ArgumentParser(allow_abbrev=False)
+    parser.add_argument('--bin-path', required=True, type=str)
+    parser.add_argument('--print-metadata', action='store_true')
+    return parser.parse_known_args()[0]
+
+def existing_metadata_print(path):
+    try:
+        metadata_file = open(path, 'r')
+        print(f'Encoded metadata: {json.load(metadata_file).get("encoded_metadata")}')
```
PavelVPV, line 385: Now I feel that Ingar's suggestion about printing the entire JSON could make sense for this command at least. So that you get the complete metadata, not just the encoded part. This is just a thought, not a request (unless you and/or others feel the same).

PavelVPV (commented):
One minor comment

scripts/bluetooth/mesh/mesh_dfu_metadata.py
Andrewpini: Well let me try it out. One thing comes to mind: What to do in the docs https://github.com/nrfconnect/sdk-nrf/pull/14078 with regard to command output (If we are doing this)? It will bloat the docs quite a bit to have an example of the entire composition.

PavelVPV: For example, use `..toggle::` https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/templates/cheat_sheet.html#toggle
Or leave as is, or ask @mia-ko .

PR #14080, opened 2024-02-19
wifi: Fix stats crash deinit

krish2718:


drivers/wifi/nrf700x/src/net_if.c
```diff
@@ -806,27 +837,38 @@ int nrf_wifi_if_set_config_zep(const struct device *dev,
...
 	}
 
-	/**
-	 * If vif_ctx_zep exists, rpu_ctx_zep and subsequent priv structures
-	 * should exist and cannot be NULL. Derive the pointers for further
-	 * structures
-	 */
+	ret = k_mutex_lock(&vif_ctx_zep->vif_lock, K_FOREVER);
+	if (ret != 0) {
+		LOG_ERR("%s: Failed to lock vif_lock", __func__);
```
jfischer-no, line 848: This is redundant and can be confusing if LOG_FUNC_NAME_PREFIX_ERR is used.

krish2718: Right, but this is prevalent across the driver, I have raised SHEL-2501Jira  to work on this outside this PR. This becomes tricky for OS agnostic code.

jfischer-no: > This becomes tricky for OS agnostic code.

This driver is for Zephyr RTOS in NCS, using Zephyr RTOS logging subsystem, or what was it implemented for?

krish2718: Yes, but this is only part of the driver, the other part is in `nrfxlib` which is OS agnostic. And this change need to be done for both, hence the Jira.

PR #13775, opened 2024-01-27
net: lib: nrf_cloud_coap: enable non-modem build

jayteemo:
Enable experimental Wi-Fi build option for nRF Cloud CoAP.

PR #14082, opened 2024-02-20
net: lib: nrf_cloud: Fix P-GPS non-MQTT download retries

plskeggs:
The function nrf_cloud_pgps_request_reset() resets the state to one that nrf_cloud_pgps_loading() considered a loading state. This prevented the next opportunity to download again from issuing another PGPS_EVT_REQUEST.

Only set the state to PGPS_NONE on unrecoverable errors.

Jira: NCSDK-25969

### DIFF
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
@@ -110,7 +110,8 @@ static uint8_t prediction_cache[PGPS_PREDICTION_STORAGE_SIZE];
 
 static uint8_t prediction_buf[PGPS_PREDICTION_STORAGE_SIZE];
 static atomic_t accept_packets;
-static atomic_t pgps_need_assistance;
+static atomic_t loading_in_progress;
+static atomic_t notified;
 
 static int validate_stored_predictions(uint16_t *bad_day, uint32_t *bad_time);
 static void log_pgps_header(const char *msg, const struct nrf_cloud_pgps_header *header);
@@ -506,15 +507,16 @@ int nrf_cloud_pgps_notify_prediction(void)
 	LOG_INF("Searching for prediction");
 	err = nrf_cloud_pgps_find_prediction(&prediction);
 	if (err == -ELOADING) {
-		pgps_need_assistance = true;
+		loading_in_progress = true;
 		err = 0;
 	} else if (err < 0) {
-		if (!pgps_need_assistance) {
-			pgps_need_assistance = true;
+		if (!loading_in_progress) {
+			loading_in_progress = true;
+			notified = false;
 			err = pgps_request_all();
 			if (err) {
 				LOG_ERR("Error while requesting pgps set: %d", err);
-				pgps_need_assistance = false; /* try again next time */
+				loading_in_progress = false; /* try again next time */
 			}
 		} else {
 			err = 0;
@@ -523,7 +525,7 @@ int nrf_cloud_pgps_notify_prediction(void)
 		pnum = err;
 		err = 0;
 		LOG_DBG("Found P-GPS prediction %d", pnum);
-		pgps_need_assistance = false;
+		loading_in_progress = false;
 		/* the application should now send data to modem with
 		 * nrf_cloud_pgps_inject()
 		 */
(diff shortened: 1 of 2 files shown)


subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
```diff
@@ -662,6 +663,9 @@ int nrf_cloud_pgps_find_prediction(struct nrf_cloud_pgps_prediction **prediction
 		return -ETIMEUNKNOWN;
 	} else if (cur_gps_sec > end_sec) {
 		if ((cur_gps_sec - end_sec) > PGPS_MARGIN_SEC) {
+			index.cur_pnum = 0xff;
+			state = PGPS_EXPIRED;
+			pgps_need_assistance = false; /* make sure we request it */
```
jayteemo, line 668: should the `atomic_` functions be used to modify this?

jayteemo: `pgps_need_assistance = false;` reads like assistance is not needed.
and the comment reads like assistance IS needed.

subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
```diff
@@ -1778,6 +1782,8 @@ int nrf_cloud_pgps_init(struct nrf_cloud_pgps_init_param *param)
 		}
 	}
 
+	state = PGPS_INITIALIZED;
```
jayteemo, line 1785: what is the purpose of this new state?
`PGPS_INITIALIZING` and `PGPS_INITIALIZED` are set but never checked for.
would a single state of `PGPS_INIT` cover both?

plskeggs: Removed and went back to previous.

subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
plskeggs: changed the name; also fixed an issue that caused double ephemeris injections

peknis (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
jayteemo: are the `atomic_` functions not required when modifying the `atomic_t` variables?

plskeggs: No, not when they are only set to a fixed value and compared to being non-zero -- when they are used as a simple boolean flag.  In fact, using `atomic_t` is overkill for such a usecase. It is definitely needed when manipulating bits within an integer from multiple threads, or when needing to to test-and-set or other interruptable operations.

Setting an atomic_t to a single value (false or true) is an atomic operation already. Checking if an atomic_t value is 0 or not is also an atomic operation.

`typedef long atomic_t;`

jayteemo: ok so using `atomic_t` has no benefit in this case.

plskeggs: I've changed them to be volatile bool variables, but in  the PR stacked on top of this one.

jayteemo (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_pgps.c
```diff
@@ -1593,34 +1599,35 @@ int nrf_cloud_pgps_begin_update(void)
 	/* assume cache is no longer valid */
 	discard_prediction_buffer();
 
+	index.loading_count = 0;
+	index.store_block = npgps_alloc_block();
+	if (index.store_block == NO_BLOCK) {
+		LOG_ERR("No free flash space!");
+		return -ENOMEM;
```
tokangas, line 1606: @plskeggs Looks like there should be a npgps_download_unlock() call here?

jayteemo: pete is out this week.
yes it looks like an unlock is needed here.
curious that this block of code was just relocated but the unlock line was removed.
i have an open PR, i'll add a commit.

jayteemo: https://github.com/nrfconnect/sdk-nrf/pull/13775/commits
it belongs to that PR

PR #14089, opened 2024-02-20
lib: modem_key_mgmt: Return errors from _mgmt_exists()

SeppoTakalo:
After adding a mutext to protect the internal buffer, the modem_key_mgmt_exists() contained a mistake that it always returned zero, instead of erro code from AT command.


SeppoTakalo:
I'll drop the changelog requirement, as this was not broken in the last release. I'll instead add 2.6.0 label, so it should be merged in before the release.

lemrey (approved):

PR #14090, opened 2024-02-20
Enable nfct-pins-as-gpios for 5340 non secure

Vge0rge:
Allows the NRF5340 non-secure target to configure configure the nfct pins through device tree, like the secure only target.

### DIFF
--- a/modules/trusted-firmware-m/CMakeLists.txt
+++ b/modules/trusted-firmware-m/CMakeLists.txt
@@ -187,7 +187,10 @@ if (CONFIG_TFM_PSA_FRAMEWORK_HAS_MM_IOVEC)
     )
 endif()
 
-if (CONFIG_NFCT_PINS_AS_GPIOS)
+# Get the nfct-pins-as-gpios property from the device tree
+dt_prop(nfct_pins_as_gpios_dt PATH "/uicr" PROPERTY "nfct-pins-as-gpios")
+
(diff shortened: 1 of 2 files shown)


modules/trusted-firmware-m/CMakeLists.txt
```diff
@@ -187,7 +187,10 @@ if (CONFIG_TFM_PSA_FRAMEWORK_HAS_MM_IOVEC)
     )
 endif()
 
-if (CONFIG_NFCT_PINS_AS_GPIOS)
+# Get the nfct-pins-as-gpios property from the device tree
+dt_prop(nfct_pins_as_gpios_dt PATH "/uicr" PROPERTY "nfct-pins-as-gpios")
+
+if (CONFIG_NFCT_PINS_AS_GPIOS OR ${nfct_pins_as_gpios_dt})
```
SebastianBoe, line 193: I don't suppose it would be possible to have CONFIG_NFCT_PINS_AS_GPIOS be true when the DT node is set like that?

That would fix all usage of CONFIG_NFCT_PINS_AS_GPIOS instead of just this one.

SebastianBoe: discussed offline

mswarowsky (approved):

nordicjm (changes requested):
No, this was rejected upstream, you should read the comments and fix them. This change needs to be reverted from sdk-zephyr

PR #14095, opened 2024-02-20
ext: oberon: Add a client view of the multipart contexts

vili-nordic:
This is similar to https://github.com/Mbed-TLS/mbedtls/pull/8729


mswarowsky (approved):
I assume this is coordinated with Oberon, else @Vge0rge will not be happy

Vge0rge:
Discussed offline, that's ok from my side.

PR #14101, opened 2024-02-21
net: lib: nrf_cloud: Prevent simultaneous FOTA and PGPS DL

plskeggs:
The FOTA poll thread was using fota_download_start_with_image_type() directly rather than nrf_cloud_download_start() which  coordinates downloads to prevent multiple simultaneous ones.

Switch to nrf_cloud_download_start(). Debug handling of aborting the P-GPS download at the start -- the first allocated block was
not freed. 

Jira: IRIS-8255

**NOTE** this is stacked on top of https://github.com/nrfconnect/sdk-nrf/pull/14082. Merge that one first, then I can rebase and push this one.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -976,6 +976,7 @@ Libraries for networking
 
   * Fixed a bug in prediction set update when the :kconfig:option:`CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD` Kconfig option was set to non-zero value.
   * Fixed a bug in handling errors when downloading P-GPS data that prevented retries until after a reboot.
+  * Fixed a bug in handling errors when downloading P-GPS data does not begin due to active FOTA download.
 
 * :ref:`lib_nrf_provisioning` library:
 
--- a/subsys/net/lib/nrf_cloud/CMakeLists.txt
+++ b/subsys/net/lib/nrf_cloud/CMakeLists.txt
@@ -45,6 +45,7 @@ zephyr_library_sources_ifdef(
 	src/nrf_cloud_download.c)
 zephyr_library_sources_ifdef(
   CONFIG_NRF_CLOUD_FOTA_POLL
+  src/nrf_cloud_download.c
   src/nrf_cloud_fota_poll.c)
 zephyr_library_sources_ifdef(
 	CONFIG_NRF_CLOUD_REST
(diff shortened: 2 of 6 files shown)


subsys/net/lib/nrf_cloud/src/nrf_cloud_fota.c
```diff
@@ -795,6 +795,7 @@ static int start_job(struct nrf_cloud_fota_job *const job, const bool send_evt)
 			.sec_tag_count = (sec_tag < 0 ? 0 : 1),
 			.pdn_id = 0,
 			.frag_size_override = CONFIG_NRF_CLOUD_FOTA_DOWNLOAD_FRAGMENT_SIZE,
+			.family = AF_INET
```
jayteemo, line 798: i don't think this is used/needed.

jayteemo (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota.c
plskeggs: It is. It is a newly added field that controls whether IPv6 or IPv4 or both are used. Since we see the IPv6 download fail every time, I added this to select IPv4 only.

plskeggs: See https://github.com/nrfconnect/sdk-nrf/blob/c191b6f6c7aeaa48ce842b4ede8238d4d81e7932/subsys/net/lib/download_client/src/download_client.c#L444

jayteemo: https://github.com/nrfconnect/sdk-nrf/blob/c191b6f6c7aeaa48ce842b4ede8238d4d81e7932/subsys/net/lib/nrf_cloud/src/nrf_cloud_download.c#L64

do you want to instead add it to `npgps_download_start`?

plskeggs: Instead? I was changing it for FOTA, not PGPS.  Or maybe you meant to type `nrf_cloud_download_start` instead of `npgps_download_start`?

jayteemo: unless im missing something, the family flag is not used for fota downloads

plskeggs: You're right -- I was following the code into `download_client_get` not `fota_download_start_with_image_type`, which eventually calls `fota_download` which creates its own `struct download_client_cfg`.  I think we want to be able to control the family for that as well.

jfischer-no:
@plskeggs Please rebase

PR #14108, opened 2024-02-21
lib: lte_link_control/date_time: No AT%XMONITOR or AT%XTIME=1 from AT monitor event handler

trantanen:
Components shouldn't be sending AT commands from system work queue because other AT commands might be ongoing, and the 2nd command will be stuck until the first one completes.

If there is an AT command ongoing by other apps/libraries, and CEREG is received from AT monitor:

- lte_lc will query for PSM parameters using `AT%XMONITOR` when we go to connected state (status 1 or 5). This was done in AT monitor event handler, which is executed in the system workqueue.
- date_time library is sending `AT%XTIME=1` whenever modem goes to search state (CEREG: 2). This happens in lte_lc event handler which is called from at_monitor in system work queue.

As the event handlers are not executed, AT monitor heap will run out and assert.

AT%XMONITOR is moved to an own work queue within lte_lc.
`AT%XTIME=1` is moved to an own work queue within date_time.

We are now also able to revert increase of CONFIG_AT_MONITOR_HEAP_SIZE from 1kB to 2kB in modem shell.

Jira: MOSH-551

### DIFF
--- a/lib/date_time/date_time_core.c
+++ b/lib/date_time/date_time_core.c
@@ -25,10 +25,11 @@ BUILD_ASSERT(CONFIG_DATE_TIME_TOO_OLD_SECONDS <= CONFIG_DATE_TIME_UPDATE_INTERVA
 
 #define DATE_TIME_EVT_TYPE_PREVIOUS 0xFF
 
-static K_SEM_DEFINE(time_fetch_sem, 0, 1);
+K_THREAD_STACK_DEFINE(date_time_stack, CONFIG_DATE_TIME_THREAD_STACK_SIZE);
+struct k_work_q date_time_work_q;
 
-static void date_time_handler(struct k_work *work);
-static K_WORK_DELAYABLE_DEFINE(time_work, date_time_handler);
+static void date_time_update_work_fn(struct k_work *work);
+static K_WORK_DELAYABLE_DEFINE(date_time_update_work, date_time_update_work_fn);
 
 static int64_t date_time_last_update_uptime;
 static date_time_evt_handler_t app_evt_handler;
@@ -59,69 +60,59 @@ static void date_time_core_schedule_update(bool check_pending)
 		/* Don't reschedule time update work in some cases,
 		 * e.g. if time is not obtained and the work is already pending.
 		 */
-		if (check_pending && k_work_delayable_is_pending(&time_work)) {
+		if (check_pending && k_work_delayable_is_pending(&date_time_update_work)) {
 			return;
 		}
 
 		LOG_DBG("New periodic date time update in: %d seconds",
 			CONFIG_DATE_TIME_UPDATE_INTERVAL_SECONDS);
 
-		k_work_reschedule(&time_work, K_SECONDS(CONFIG_DATE_TIME_UPDATE_INTERVAL_SECONDS));
+		k_work_reschedule_for_queue(
+			&date_time_work_q,
+			&date_time_update_work,
+			K_SECONDS(CONFIG_DATE_TIME_UPDATE_INTERVAL_SECONDS));
 	}
 }
 
-static void date_time_update_thread(void)
+static void date_time_update_work_fn(struct k_work *work)
 {
 	int err;
 
-	while (true) {
-		k_sem_take(&time_fetch_sem, K_FOREVER);
+	LOG_DBG("Updating date time UTC...");
 
-		LOG_DBG("Updating date time UTC...");
-
-		err = date_time_core_current_check();
-		if (err == 0) {
-			LOG_DBG("Using previously obtained time");
-			date_time_core_schedule_update(true);
-			date_time_core_notify_event(DATE_TIME_EVT_TYPE_PREVIOUS);
-			continue;
-		}
+	err = date_time_core_current_check();
+	if (err == 0) {
+		LOG_DBG("Using previously obtained time");
+		date_time_core_schedule_update(true);
+		date_time_core_notify_event(DATE_TIME_EVT_TYPE_PREVIOUS);
+		return;
+	}
 
 #if defined(CONFIG_DATE_TIME_MODEM)
-		LOG_DBG("Getting time from cellular network");
-		int tz = DATE_TIME_TZ_INVALID;
-		int64_t date_time_ms_modem = 0;
+	LOG_DBG("Getting time from cellular network");
+	int tz = DATE_TIME_TZ_INVALID;
+	int64_t date_time_ms_modem = 0;
 
-		err = date_time_modem_get(&date_time_ms_modem, &tz);
-		if (err == 0) {
-			date_time_core_store_tz(date_time_ms_modem, DATE_TIME_OBTAINED_MODEM, tz);
-			continue;
-		}
+	err = date_time_modem_get(&date_time_ms_modem, &tz);
+	if (err == 0) {
+		date_time_core_store_tz(date_time_ms_modem, DATE_TIME_OBTAINED_MODEM, tz);
+		return;
+	}
 #endif
 #if defined(CONFIG_DATE_TIME_NTP)
-		LOG_DBG("Getting time from NTP server");
-		int64_t date_time_ms_ntp = 0;
+	LOG_DBG("Getting time from NTP server");
+	int64_t date_time_ms_ntp = 0;
 
-		err = date_time_ntp_get(&date_time_ms_ntp);
-		if (err == 0) {
-			date_time_core_store(date_time_ms_ntp, DATE_TIME_OBTAINED_NTP);
-			continue;
-		}
-#endif
-		LOG_DBG("Did not get time from any time source");
-
-		date_time_core_schedule_update(true);
-		date_time_core_notify_event(DATE_TIME_NOT_OBTAINED);
+	err = date_time_ntp_get(&date_time_ms_ntp);
+	if (err == 0) {
+		date_time_core_store(date_time_ms_ntp, DATE_TIME_OBTAINED_NTP);
+		return;
 	}
-}
+#endif
+	LOG_DBG("Did not get time from any time source");
 
-K_THREAD_DEFINE(time_thread, CONFIG_DATE_TIME_THREAD_STACK_SIZE,
-		date_time_update_thread, NULL, NULL, NULL,
-		K_LOWEST_APPLICATION_THREAD_PRIO, 0, 0);
-
-static void date_time_handler(struct k_work *work)
-{
-	k_sem_give(&time_fetch_sem);
+	date_time_core_schedule_update(true);
+	date_time_core_notify_event(DATE_TIME_NOT_OBTAINED);
 }
 
 void date_time_lte_ind_handler(const struct lte_lc_evt *const evt)
@@ -134,14 +125,12 @@ void date_time_lte_ind_handler(const struct lte_lc_evt *const evt)
 		case LTE_LC_NW_REG_REGISTERED_HOME:
 		case LTE_LC_NW_REG_REGISTERED_ROAMING:
 			if (!date_time_is_valid()) {
-				k_work_reschedule(&time_work, K_SECONDS(1));
+				k_work_reschedule_for_queue(
+					&date_time_work_q,
+					&date_time_update_work,
+					K_SECONDS(1));
 			}
 			break;
-#if defined(CONFIG_DATE_TIME_MODEM)
-		case LTE_LC_NW_REG_SEARCHING:
-			date_time_modem_xtime_subscribe();
-			break;
-#endif
 		default:
 			break;
 		}
@@ -154,6 +143,17 @@ void date_time_lte_ind_handler(const struct lte_lc_evt *const evt)
 
 void date_time_core_init(void)
 {
+	struct k_work_queue_config cfg = {
+		.name = "date_time_work_q",
+	};
+
+	k_work_queue_start(
+		&date_time_work_q,
+		date_time_stack,
+		CONFIG_DATE_TIME_THREAD_STACK_SIZE,
+		K_LOWEST_APPLICATION_THREAD_PRIO,
+		&cfg);
+
 	if (IS_ENABLED(CONFIG_DATE_TIME_AUTO_UPDATE) && IS_ENABLED(CONFIG_LTE_LINK_CONTROL)) {
 		lte_lc_register_handler(date_time_lte_ind_handler);
 	}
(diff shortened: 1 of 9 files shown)


lib/lte_link_control/lte_lc.c
```diff
@@ -123,6 +123,15 @@ static const char system_mode_preference[] = {
...
 };
 
+static void lte_lc_psm_get_work_fn(struct k_work *work_item);
+K_WORK_DEFINE(lte_lc_psm_get_work, lte_lc_psm_get_work_fn);
+
+#define LTE_LC_PSM_GET_STACK_SIZE 1024
+#define LTE_LC_PSM_GET_PRIORITY 5
+K_THREAD_STACK_DEFINE(lte_lc_psm_get_stack, LTE_LC_PSM_GET_STACK_SIZE);
+
+static struct k_work_q lte_lc_psm_get_work_q;
```
trantanen, line 133: Right now I'm adding this work queue just for doing AT%XMONITOR when CEREG notification had been received. It's a bit awkward that we need to make a work queue and reserve memory for it for such a small task that occurs in a corner case but there's really no other way out.

We could think of sending all events from the work queue and even parsing AT notifications in the future.

We need to think if we want to delay this feature beyond 2.6.0 to mitigate the risks.

lib/lte_link_control/lte_lc.c
```diff
@@ -250,23 +300,23 @@ static void at_handler_cereg(const char *response)
...
-	if (err) {
-		if (err != -EBADMSG) {
-			LOG_ERR("Failed to get PSM information");
-		}
+	if (psm_cfg.tau == -1) {
+		/* Need to get legacy TAU using AT%XMONITOR.
+		 *
+		 * As we are in an AT notification handler that is run from the system work queue,
+		 * we shall not send AT commands here because another AT command might be ongoing,
+		 * and the second command will be stuck until the first one completes.
```
tokangas, line 308:
```suggestion
		 * and the second command will be blocked until the first one completes.
```

lib/lte_link_control/lte_lc.c
```diff
@@ -250,23 +300,23 @@ static void at_handler_cereg(const char *response)
...
+		 * Sending any further AT notifications will gradually exhaust AT monitor
+		 * library's heap and eventually it will run out causing assert or fatal errors.
```
tokangas, line 310: What do you refer to with "fatal errors"? To the case when asserts are disabled? If asserts are disabled and the heap runs out, notifications will simply not be delivered, right? You could just say that in the comment, I think, for example:
```suggestion
		 * Further AT notifications from the modem will gradually exhaust AT monitor
		 * library's heap and eventually it will run out causing an assert or AT notifications
		 * not being dispatched.
```

lib/lte_link_control/lte_lc.c
```diff
@@ -716,6 +766,22 @@ int lte_lc_init(void)
...
+		struct k_work_queue_config cfg = {
+			.name = "lte_lc_psm_get_work_q",
+		};
+
+		k_work_queue_start(
+			&lte_lc_psm_get_work_q,
+			lte_lc_psm_get_stack,
+			K_THREAD_STACK_SIZEOF(lte_lc_psm_get_stack),
+			LTE_LC_PSM_GET_PRIORITY,
+			&cfg);
```
tokangas, line 781: Maybe the work_q could be initialized using `SYS_INIT()` instead? You could also get rid of couple of comments above.

tokangas (commented):
The added check if `%XMONITOR` is needed to get the PSM parameters is good. As far as I know, most of the networks use T3412_ext timer, so it's not even necessary to do `%XMONITOR`. For the remaining cases, we have to use the `%XMONITOR` command to get the T3412 legacy timer value. The downside of this change is, that we need to have a new work queue for it, but as @trantanen said, maybe we should consider dispatching all events from somewhere else than the system work queue.

lib/lte_link_control/lte_lc.c
trantanen: Yep, we should just say that AT notifications are not dispatched.

lib/lte_link_control/lte_lc.c
trantanen: Yep, makes sense.

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -815,6 +818,60 @@ int parse_cereg(const char *at_response,
...
+			/* Legacy TAU is not requested because we do not get it from CEREG.
+			 * If extended TAU is not set, TAU will be set to inactive so
+			 * caller can then make its conclusions.
+			 */
+			err = parse_psm(active_time_str, tau_ext_str, NULL, psm_cfg);
+			if (err) {
+				LOG_ERR("Failed to parse PSM configuration, error: %d", err);
+			}
+		}
+		/* Even if a failure occurred, it's normal sometimes so we clear the error */
```
tokangas, line 871:
```suggestion
		/* The notification does not always contain PSM parameters, so this is not considered */
		 * an error
		 */
```

lib/lte_link_control/lte_lc.c
tokangas: IMHO, the work queue could be already named for example `lte_lc_work_q`, even though it's (for now) used only for this one use case.

trantanen: I would rename the work queue when we use it for something else. Because it may be that for the next two years, we won't be using it for anything else 😄

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -815,6 +818,62 @@ int parse_cereg(const char *at_response,
...
+		psm_cfg->tau = -1;
+
+		/* Get active time */
+		err_active_time = at_params_string_get(
+				&resp_list,
+				is_notif ? AT_CEREG_ACTIVE_TIME_INDEX :
+					   AT_CEREG_READ_ACTIVE_TIME_INDEX,
+				active_time_str, &str_len);
+		if (err_active_time) {
+			LOG_DBG("Active time not found, error code: %d", err_active_time);
```
tokangas, line 844: To be consistent with other error traces.
```suggestion
			LOG_DBG("Active time not found, error: %d", err_active_time);
```

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -815,6 +818,62 @@ int parse_cereg(const char *at_response,
...
+		}
+
+		/* Get TAU */
+		err_tau = at_params_string_get(
+				&resp_list,
+				is_notif ? AT_CEREG_TAU_INDEX :
+					   AT_CEREG_READ_TAU_INDEX,
+				tau_ext_str, &str_len);
+		if (err_tau) {
+			LOG_DBG("TAU not found, error code: %d", err_tau);
```
tokangas, line 856:
```suggestion
			LOG_DBG("TAU not found, error: %d", err_tau);
```

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -815,6 +818,62 @@ int parse_cereg(const char *at_response,
...
+				is_notif ? AT_CEREG_ACTIVE_TIME_INDEX :
+					   AT_CEREG_READ_ACTIVE_TIME_INDEX,
+				active_time_str, &str_len);
+		if (err_active_time) {
+			LOG_DBG("Active time not found, error code: %d", err_active_time);
+		} else {
+			LOG_DBG("Active time: %s", active_time_str);
+		}
+
+		/* Get TAU */
```
tokangas, line 849: I would make it explicit that this is "extended periodic TAU value", "T3412-ext" or "Periodic-TAU-ext". Or even "TAU-ext". And other logs below should be updated accordingly.
```suggestion
		/* Get Periodic-TAU-ext */
```

lib/lte_link_control/lte_lc.c
tokangas: I would still prefer the generic naming, but I'll leave it up to you to decide. 😄

lib/lte_link_control/lte_lc.c
```diff
@@ -132,6 +141,43 @@ static bool is_cellid_valid(uint32_t cellid)
...
+
+static void lte_lc_psm_get_work_fn(struct k_work *work_item)
+{
+	int err;
+	struct lte_lc_psm_cfg psm_cfg = {
+		.active_time = -1,
+		.tau = -1
+	};
+
+	LOG_DBG("lte_lc_psm_get_work started");
```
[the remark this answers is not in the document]
trantanen: I'd like to have this if it was printed but in MOSH case, logging is not working when `AT+COPS=?` is pending so not sure what to do.

lib/date_time/date_time_modem.c
```diff
@@ -215,7 +215,13 @@ void date_time_modem_store(struct tm *ltm, int tz)
 	}
 }
 
-void date_time_modem_xtime_subscribe(void)
+#define DATE_TIME_XTIME_SUBSCRIBE_STACK_SIZE 1024
+#define DATE_TIME_XTIME_SUBSCRIBE_PRIORITY 5
+K_THREAD_STACK_DEFINE(date_time_modem_xtime_subscribe_stack, DATE_TIME_XTIME_SUBSCRIBE_STACK_SIZE);
+
+static struct k_work_q date_time_modem_xtime_subscribe_work_q;
```
trantanen, line 222: Need to check if `AT+CCLK=?` should be handled as well

lib/lte_link_control/lte_lc.c
```diff
@@ -123,6 +123,15 @@ static const char system_mode_preference[] = {
 	[LTE_LC_SYSTEM_MODE_PREFER_NBIOT_PLMN_PRIO]	= '4',
 };
 
+static void lte_lc_psm_get_work_fn(struct k_work *work_item);
+K_WORK_DEFINE(lte_lc_psm_get_work, lte_lc_psm_get_work_fn);
+
+#define LTE_LC_PSM_GET_STACK_SIZE 1024
+#define LTE_LC_PSM_GET_PRIORITY 5
```
lemrey, line 130: Why `5`? :) I think the default should work.

lib/lte_link_control/lte_lc.c
```diff
@@ -250,23 +297,21 @@ static void at_handler_cereg(const char *response)
 		return;
 	}
 
-	err = lte_lc_psm_get(&psm_cfg.tau, &psm_cfg.active_time);
-	if (err) {
-		if (err != -EBADMSG) {
-			LOG_ERR("Failed to get PSM information");
-		}
+	if (psm_cfg.tau == -1) {
+		/* Need to get legacy TAU using AT%XMONITOR.
```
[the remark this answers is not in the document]
trantanen: I'd like to have TAU mentioned too so modifying:
`Need to get legacy T3412 value as TAU using AT%XMONITOR.`

lib/date_time/date_time_modem.c
tokangas: Having a work queue for `%XTIME` and a thread for the `+CCLK` query seems excessive. Is there any reason why `date_time_update_thread()` couldn't be replaced with a delayable work running in the new work queue?

trantanen: Converted existing thread to work queue and combined them

lib/date_time/date_time_core.c
```diff
@@ -25,10 +25,11 @@ BUILD_ASSERT(CONFIG_DATE_TIME_TOO_OLD_SECONDS <= CONFIG_DATE_TIME_UPDATE_INTERVA
...
 #define DATE_TIME_EVT_TYPE_PREVIOUS 0xFF
 
-static K_SEM_DEFINE(time_fetch_sem, 0, 1);
+K_THREAD_STACK_DEFINE(date_time_stack, CONFIG_DATE_TIME_THREAD_STACK_SIZE);
+struct k_work_q date_time_work_q;
 
-static void date_time_handler(struct k_work *work);
-static K_WORK_DELAYABLE_DEFINE(time_work, date_time_handler);
+static void date_time_work_fn(struct k_work *work);
+static K_WORK_DELAYABLE_DEFINE(date_time_work, date_time_work_fn);
```
tokangas, line 32:
```suggestion
static K_WORK_DELAYABLE_DEFINE(date_time_update_work, date_time_update_work_fn);
```

tokangas (approved):

lib/lte_link_control/lte_lc.c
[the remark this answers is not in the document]
trantanen: Decided to remove this.

trantanen:
Added one more bug fix which was also in the old code:
`lib: date_time: Manual update stops periodic updates`

lib/date_time/date_time_core.c
```diff
@@ -25,10 +25,14 @@ BUILD_ASSERT(CONFIG_DATE_TIME_TOO_OLD_SECONDS <= CONFIG_DATE_TIME_UPDATE_INTERVA
...
+K_THREAD_STACK_DEFINE(date_time_stack, CONFIG_DATE_TIME_THREAD_STACK_SIZE);
+struct k_work_q date_time_work_q;
 
-static void date_time_handler(struct k_work *work);
-static K_WORK_DELAYABLE_DEFINE(time_work, date_time_handler);
+static void date_time_update_work_fn(struct k_work *work);
+static K_WORK_DELAYABLE_DEFINE(date_time_update_work, date_time_update_work_fn);
+
+static void date_time_update_manual_work_fn(struct k_work *work);
+static K_WORK_DEFINE(date_time_update_manual_work, date_time_update_manual_work_fn);
```
tokangas, line 35: Not a big deal, but it came to my mind if it would be possible to define two works using the same `date_time_update_work_fn` function.

trantanen: Good point! Seems to work.

jhirsi (approved):

gchwier:
I've restarted test-sdk-mcuboot CI regression, because were some problems in lab

lib/date_time/date_time_modem.c
```diff
@@ -224,3 +226,12 @@ void date_time_modem_xtime_subscribe(void)
 		LOG_ERR("Subscribing to modem AT%%XTIME notifications failed, err=%d", err);
 	}
 }
+
+K_WORK_DEFINE(date_time_modem_xtime_subscribe_work, date_time_modem_xtime_subscribe_work_fn);
+
+void date_time_modem_xtime_subscribe(void)
```
trantanen, line 232: We could call this from `NRF_MODEM_LIB_ON_CFUN` callback instead of when `CEREG: 2` is received.

tokangas: Yes, most likely that was the intention of the current implementation, but the callback wasn't available at the time. CFUN callback is a much better way to do it.

trantanen: This is now done

lemrey (approved):

PR #14110, opened 2024-02-21
doc: config_and_build: Note on how to pass Kconfig to child image

de-nordic:
The commit adds note on how to pass Kconfig options to child image, here MCUboot specifically.

Ref: NCSIDB-1020

### DIFF
--- a/doc/nrf/device_guides/working_with_nrf/nrf53/nrf5340.rst
+++ b/doc/nrf/device_guides/working_with_nrf/nrf53/nrf5340.rst
@@ -593,6 +593,9 @@ To enable the simultaneous update of multiple images in the MCUboot, set the fol
 * :kconfig:option:`CONFIG_PCD_APP` - Enable commands exchange with the network core.
 * :kconfig:option:`CONFIG_UPDATEABLE_IMAGE_NUMBER` - Enable support for multiple update partitions by setting this option to ``2``.
 
+As described in :ref:`multi-image build guide <ug_multi_image_variables>`, make sure to add the child image prefix to the name of Kconfig options that are used when building MCUboot as a child image (for example, ``child_image_CONFIG_PCD_APP``).
+If not changed, then the default child image prefix for MCUboot is ``mcuboot_`` (for example, ``mcuboot_CONFIG_PCD_APP``).
+
 .. note::
 
    The application core can be reverted, but doing so bricks the network core upon revertion, as the reversion process fills the network core with the content currently in the RAM that pcd uses.


greg-fer:
Adding @kittydepa , as this is Pluto's area.

doc/nrf/device_guides/working_with_nrf/nrf53/nrf5340.rst
```diff
@@ -584,6 +584,13 @@ Bluetooth buffers configuration introduced by the :kconfig:option:`CONFIG_NCS_SA
 Simultaneous multi-image DFU
 ****************************
 
+.. note::
+
+   When building MCUboot as child image, within a multi-image project, the Kconfig options provided
+   via command line or general project file need to be prefixed with a child image name.
+   If a user has not changed anything the MCUboot will be built as mcuboot child image
+   and the prefix will be mcuboot_, for example mcuboot_CONFIG_PCD_APP.
```
greg-fer, line 592:
```suggestion
   As described in :ref:`multi-image build guide <ug_multi_image_variables>`, make sure to the child image name prefix to the name of the Kconfig options that are used when building MCUboot as child image (for example, ``child_image_CONFIG_PCD_APP``.
   If you don't do this, the MCUboot will be built as the standard child image with the ``mcuboot`` prefix (for example, ``mcuboot_CONFIG_PCD_APP``).
```

greg-fer (changes requested):
I think adding the ref to the multi-image build guide makes sense here.

doc/nrf/device_guides/working_with_nrf/nrf53/nrf5340.rst
```diff
@@ -584,6 +584,11 @@ Bluetooth buffers configuration introduced by the :kconfig:option:`CONFIG_NCS_SA
 Simultaneous multi-image DFU
 ****************************
 
+.. note::
+
+   As described in :ref:`multi-image build guide <ug_multi_image_variables>`, make sure to the child image name prefix to the name of the Kconfig options that are used when building MCUboot as child image (for example, ``child_image_CONFIG_PCD_APP``.
```
kittydepa, line 589:
```suggestion
   As described in :ref:`multi-image build guide <ug_multi_image_variables>`, make sure to name the child image prefix to the name of the Kconfig options that are used when building MCUboot as a child image (for example, ``child_image_CONFIG_PCD_APP``.
```

doc/nrf/device_guides/working_with_nrf/nrf53/nrf5340.rst
```diff
@@ -584,6 +584,11 @@ Bluetooth buffers configuration introduced by the :kconfig:option:`CONFIG_NCS_SA
...
+   If you don't do this, the MCUboot will be built as the standard child image with the ``mcuboot`` prefix (for example, ``mcuboot_CONFIG_PCD_APP``).
```
kittydepa, line 590:
```suggestion
   If you don't do this, MCUboot will be built as the standard child image with the ``mcuboot`` prefix (for example, ``mcuboot_CONFIG_PCD_APP``).
```

doc/nrf/device_guides/working_with_nrf/nrf53/nrf5340.rst
```diff
@@ -584,6 +584,11 @@ Bluetooth buffers configuration introduced by the :kconfig:option:`CONFIG_NCS_SA
 Simultaneous multi-image DFU
 ****************************
 
+.. note::
```
kittydepa, line 587: We don't usually have "Notes" immediately after a heading title (I could be wrong about this?)
I would suggest to move this to  after line 594.

Or, if this info is very important to have it here in the beginning I would take it out of the "note".

doc/nrf/device_guides/working_with_nrf/nrf53/nrf5340.rst
annwoj: maybe: "to add the child image prefix (...)"?

greg-fer (approved):

doc/nrf/device_guides/working_with_nrf/nrf53/nrf5340.rst
greg-fer: This suggestion is OK, but now we have two notes one after another, which sometimes is a bit of an overkill.

de-nordic: > This suggestion is OK, but now we have two notes one after another, which sometimes is a bit of an overkill.

Yeah I agree. I wonder, whether instead of the note there should be rather some introduction chapter, on the page, that would remind on how the Kconfig options work with child-images and how to address child images, instead of doing remark here.

It kinda reminds me what Larry Trusk has written about footnotes in "Punctuation" (https://www.sussex.ac.uk/informatics/punctuation/essaysandletters/footnotes); in short:
```
 There are two main rules in the use of footnotes. First:

    Do not use a footnote if you can possibly avoid it. 
```

greg-fer: That is a good point, as too many notes lead to users ignoring them. I don't oppose the idea of the introduction, but in this particular case I also think that the ``.. note::`` is not needed and this information can leave as normal text in the new location after the list of Kconfigs.

de-nordic: OK, I will remove the `.. note::` and make it part of text.

annwoj (approved):

PR #14116, opened 2024-02-21
Bluetooth: Mesh: Add new Sensor API support to Light Ctrl shell

akredalen:
* Light Ctrl shell can be used in an application with new sensor API.
* Small fix: updated shell_ctrl_cli.c to comply with split model structure.

### DIFF
--- a/subsys/bluetooth/mesh/shell/shell_ctrl_cli.c
+++ b/subsys/bluetooth/mesh/shell/shell_ctrl_cli.c
@@ -27,7 +27,7 @@ static int cmd_mode_get(const struct shell *shell, size_t argc, char *argv[])
 		return -ENODEV;
 	}
 
-	struct bt_mesh_light_ctrl_cli *cli = mod->user_data;
+	struct bt_mesh_light_ctrl_cli *cli = mod->rt->user_data;
 	bool rsp;
 
 	int err = bt_mesh_light_ctrl_cli_mode_get(cli, NULL, &rsp);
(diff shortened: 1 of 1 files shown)


ludvigsj (approved):

HaavardRei (approved):

subsys/bluetooth/mesh/shell/shell_ctrl_cli.c
```diff
@@ -27,7 +27,7 @@ static int cmd_mode_get(const struct shell *shell, size_t argc, char *argv[])
 		return -ENODEV;
 	}
 
-	struct bt_mesh_light_ctrl_cli *cli = mod->user_data;
```
alxelax, line 30: This is not a unique light control client issue. I took a quick look over shell clients, looks, this is a general problem. Currently, almost all clients cannot be built. I'm not sure if it should be fixed in this PR, but obviously should.

Andrewpini (approved):

alxelax (commented):
LGTM, but I have question. Will it be built with `BT_MESH_SENSOR_USE_LEGACY_SENSOR_VALUE=y`? I guess, it should be while we support legacy API,

subsys/bluetooth/mesh/shell/shell_ctrl_cli.c
alxelax: Created JIRA ticket for this. Let's do it in another PR.

akredalen:
> LGTM, but I have question. Will it be built with `BT_MESH_SENSOR_USE_LEGACY_SENSOR_VALUE=y`? I guess, it should be while we support legacy sensor API.

Yes, the light ctrl shell was tested in a custom sample with and without the legacy API enabled.

subsys/bluetooth/mesh/shell/shell_ctrl_cli.c
akredalen: I will remove this part from this PR then.

alxelax: No, please, leave it as is. Otherwise, the client cannot be built

alxelax (approved):

PR #14120, opened 2024-02-21
drivers: wifi: Fix duplicated logger object names

markaj-nordic:
Fixed Wi-Fi driver bus layer logger registering (duplicated logger symbol with different log levels)

@krish2718 the problem with Wi-Fi driver is not visible in Wi-Fi samples because they use the same default
value for both `CONFIG_WIFI_NRF700X_LOG_LEVEL` and `CONFIG_WIFI_NRF700X_BUS_LOG_LEVEL`.
Note that it was not reproducible with LOG_MODE_MINIMAL, because this mode does not really use log objects.

### DIFF
--- a/drivers/wifi/nrf700x/src/qspi/src/qspi_if.c
+++ b/drivers/wifi/nrf700x/src/qspi/src/qspi_if.c
@@ -102,7 +102,7 @@ static void qspi_device_uninit(const struct device *dev);
 #define QSPI_SCK_DELAY 0
 #define WORD_SIZE 4
 
-LOG_MODULE_DECLARE(wifi_nrf, CONFIG_WIFI_NRF700X_BUS_LOG_LEVEL);
+LOG_MODULE_DECLARE(wifi_nrf_bus, CONFIG_WIFI_NRF700X_BUS_LOG_LEVEL);
 
 /**
  * @brief QSPI buffer structure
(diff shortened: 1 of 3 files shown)


ArekBalysNordic (commented):
I can see some inconsistencies here. I know that this is a workaround for radio problems on nRF52840, but why do we want to add it to the Matter Bridge which is Wi-Fi only I guess for now it is not required, right? I can see that you changed only prj.conf files, but what about proprietary conf files such as `prj_thread_wifi_switched.conf`? Moreover, what about the Weather Station? I know that workaraound is dedicated to nRF52, but if you're considering also nRF70 so why not nRF53?

markaj-nordic:
> I can see some inconsistencies here. I know that this is a workaround for radio problems on nRF52840, but why do we want to add it to the Matter Bridge which is Wi-Fi only I guess for now it is not required, right? I can see that you changed only prj.conf files, but what about proprietary conf files such as `prj_thread_wifi_switched.conf`? Moreover, what about the Weather Station? I know that workaraound is dedicated to nRF52, but if you're considering also nRF70 so why not nRF53?

The decision was to turn it on for all samples regardless of the networking transport layer. The Weather Station has deferred mode already enabled. I guess I just missed lock switched sample. To sum up, it'd be better to introduce this change in the Matter SDK to avoid discrepancies. But let's wait until there is a final decision because we may have just found more reliable workaround but it needs to be tested more thoroughly:
https://github.com/nrfconnect/sdk-nrf/pull/14147

Damian-Nordic (approved):

krish2718 (approved):
Thanks @markaj-nordic  for fixing this.

PR #14122, opened 2024-02-21
net: lib: fota_download: Accept family parameter

plskeggs:
Now that the download_client library has changed control over use of IPv6 vs. IPv4 to be done at runtime, control family with optional Kconfigs.

Use IPv4 for nrfcloud samples.


include/net/fota_download.h
```diff
@@ -135,6 +135,8 @@ int fota_download_init(fota_download_callback_t client_callback);
  * @param fragment_size Fragment size to be used for the download.
  *			If 0, @kconfig{CONFIG_DOWNLOAD_CLIENT_HTTP_FRAG_SIZE} is used.
  * @param expected_type Type of firmware file to be downloaded and installed.
+ * @param family Address family to be used for the download, AF_INET6 or AF_INET.
```
jayteemo, line 138: should this be a new kconfig option instead of a function parameter?

plskeggs: Well, this change is in response to someone changing the download_client library to set it at runtime instead of with a Kconfig option, so I was maintaining the same spirit.  Seems odd to simply move the goalpost.

jayteemo (approved):

include/net/fota_download.h
plskeggs: That's true, we do want to minimize that. We will see what other feedback comes.

MarkusLassila (approved):
Okay for SLM

include/net/fota_download.h
tomi-font: Agreed that minimizing API breaking changes should be avoided where feasible.

SeppoTakalo:
This feels like going into wrong direction.

Why would we need to define address family for DNS addresses? Is that the whole point of DNS so that we don't need to care about IPv4/6? Pushing this to high level API feels a bit wrong.

If you must hard code the family, why don't you resolve the name into address first, then feed it into download client.

SeppoTakalo (changes requested):
Util I understand what problem this fixes, I would propose not to accept API changes.

plskeggs:
@SeppoTakalo with the change in `download_client` every download (FOTA, P-GPS) is always attempted with a name lookup using IPv6 followed by IPv4.  Without modifying the fota_download library to set the family parameter, we have no control over this, other than as you implied (look up the address and then pass that as the host parameter).

We get errors like this, at least with the carriers iBasis uses in Portland:
```
[00:00:19.709,930] <inf> download_client: Setting up TLS credentials, sec tag count 1
[00:00:19.710,083] <inf> download_client: Connecting to 2600:9000:2549:8e00:3:e9a6:59c0:93a1
[00:00:19.710,266] <err> download_client: Unable to connect, errno 114 <-- NOTE ERROR
[00:00:19.849,945] <inf> download_client: Setting up TLS credentials, sec tag count 1
[00:00:19.850,067] <inf> download_client: Connecting to 18.160.156.125
...
[00:05:18.390,502] <inf> nrf_cloud_fota_poll: Starting FOTA download of static.api.nrfcloud.com/v1/firmwares/modem/MODEM*9e08401c*mfw_nrf9160_1.3.6-FOTA-TEST/mfw_nrf9160_update_from_1.3.6_to_1.3.6-FOTA-TEST.bin
[00:05:18.390,655] <inf> download_client: Downloading: v1/firmwares/modem/MODEM*9e08401c*mfw_nrf9160_1.3.6-FOTA-TEST/mfw_nrf9160_update_from_1.3.6_to_1.3.6-FOTA-TEST.bin [0]
[00:05:18.580,963] <wrn> download_client: Failed to resolve hostname static.api.nrfcloud.com on IPv6 <-- NOTE THE FAILURE
[00:05:18.774,414] <inf> download_client: Setting up TLS credentials, sec tag count 1

**Comment by tomi-font** [comment]
Could the code that is responsible for doing the DNS lookup be changed to resolve the hostname to some address family that is actually supported rather than always going first with IPv6?
We had something like this in SLM, that I think we fixed by using AF_UNSPEC as the address family.

**Comment by SeppoTakalo** [comment]
DNS responses are cached so it is not that wasteful to do both DNS queries.

As there is limitation in modem that it only responds with one address, there is a slight possibility that AF_UNSPEC would render the client unable to connect anymore.

In a standard POSIX, using AF_UNSPEC and going through list of responses would be the right way to do it, but we only receive one response.

@stig-bjorlykke changed the download_client operation from hard coded Kconfig value to runtime change, so I believe there was the reasoning why it was not AF_UNSPEC. The client still allows runtime changing the family on its configuration structure.

However, we are now proposing API break on a FOTA client and justifying that by optimizing one packet from a FOTA process. That process does not happen very often. So if I have to choose whether to break public API for this, or keep developer happy by not touching the API, I would prefer to keep the API intact.

**Comment by plskeggs** [comment]
@SeppoTakalo it defaults to AF_UNSPEC (==0) for fota_download() by virtue of default initialization of members when it partially initializes `struct download_client_cfg`. 

> The client still allows runtime changing the family on its configuration structure.

Yes, and that is what I was doing here -- using that change in one of the libraries that uses download_client().

> As there is limitation in modem that it only responds with one address, there is a slight possibility that AF_UNSPEC would render the client unable to connect anymore.

This would be disaster if this occurred on a FOTA download. A device would not be able to be field-upgraded.

**Comment by stig-bjorlykke** [comment]
The changes in commit 57c5706f8f fixed two issues where using `AF_UNSPEC` also would fail. One when the network does not give the device an IPv6 address (happens in many networks and when using MVNO SIMs), and another when the network does not correctly route IPv6 traffic (this happens for one of the big US carriers).

The changes in commit fff3039545 added runtime support for not trying both IP versions when the application knows which IP version it support/uses. This also made it default to try both, but as Seppo write this should not be a big issue because a FOTA does not happen that often in a device lifetime.

I was thinking about explicit configure IPv4 only for all users of the `download_client` (to keep that code unchanged), but I ended up with the conclusion that it's added feature and will not fail a download. If you don't agree we can easily add the `.family = AF_INET` where needed.

File: subsys/net/lib/fota_download/src/fota_download.c [hunk 2]
```diff
@@ -431,6 +431,7 @@ int fota_download(const char *host, const char *file,
 	struct download_client_cfg config = {
 		.pdn_id = pdn_id,
 		.frag_size_override = fragment_size,
+		.family = family
```
jayteemo, line 434: question for the `fota_download` codeowners:
what is the reason for this separate config struct?  why not use `dlc.config`?

jayteemo, line 434: suggest adding the following to `fota_download`'s Kconfig:

```
choice
	prompt "Address family to be used for the FOTA download"
	default FOTA_DOWNLOAD_ADDR_FAM_UNSPEC

config FOTA_DOWNLOAD_ADDR_FAM_IPV4
	bool "IPv4 (AF_INET)"
config FOTA_DOWNLOAD_ADDR_FAM_IPV6
	bool "IPv6 (AF_INET6)"
config FOTA_DOWNLOAD_ADDR_FAM_UNSPEC
	bool "Unspecified (AF_UNSPEC)"

endchoice
```

and just set the family here:

```
	if (IS_ENABLED(CONFIG_FOTA_DOWNLOAD_ADDR_FAM_IPV4)) {
		config.family = AF_INET;
	} else if (IS_ENABLED(CONFIG_FOTA_DOWNLOAD_ADDR_FAM_IPV6)) {
		config.family = AF_INET6;
	} else {
		config.family = AF_UNSPEC;
	}
```

plskeggs: I will go ahead and make these changes. Better a Kconfig than no control.

stig-bjorlykke:
> A failure of FOTA to function is terrible _because_, though it does not get used often, when it does, a failure of it to work could have bad consequences and so we should do what we can to avoid failures for known issues.

I don't understand your concerns about not having control. The `download_client` will always try both IPv6 and IPv4 if configured with `.family = AF_UNSPEC`. If you have a use case that does not work then I would like to have a look. The logs you provided here have some error/warning messages, but the fallback works fine.

Using IPv4-only for `fota_download` to avoid trying IPv6 first is fine, but be aware that this will fail in a IPv6-only network.

The issues described in [IRIS-8317](https://nordicsemi.atlassian.net/browse/IRIS-8317) are different.

peknis (approved):

plskeggs:
> I don't understand your concerns about not having control.

@stig-bjorlykke I was reacting to this comment from Seppo:

> As there is limitation in modem that it only responds with one address, there is a slight possibility that AF_UNSPEC would render the client unable to connect anymore.

It is that case we need to avoid.

samples/cellular/nrf_cloud_multi_service/prj.conf
```diff
@@ -94,6 +94,8 @@ CONFIG_MODEM_INFO_ADD_NETWORK=y
 CONFIG_NRF_CLOUD_PGPS=y
 CONFIG_NRF_CLOUD_PGPS_REPLACEMENT_THRESHOLD=4
 CONFIG_NRF_CLOUD_PGPS_REQUEST_UPON_INIT=y
+CONFIG_FOTA_DOWNLOAD_ADDR_FAM_IPV4=y
```
glarsennordic, line 97: This will make the default build of this sample incompatible with IPv6, right?

If we have specific test hardware that needs IPV4, I think we should set the builds for that hardware to use IPV4 rather than apply this as the sample default

But I have another concern: The `_UNSPEC` option should, at least in theory, try both IPv6 and IPv4, right?

So I do not see how this will solve the issue we've been seeing with IPv6 connection failures -- In theory those failed IPv6 downloads should have fallen back to IPv4 on their own.

glarsennordic (changes requested):
At minimum I recommend removing defaults changes from sample configs -- but I also think this may not actually address the issue it is intended to solve

samples/cellular/nrf_cloud_multi_service/prj.conf
plskeggs: I was unware that we are now seeing carriers with IPv6-only networks. I think in the past (an NCS rev or two in the past) our samples were set to IPv4 only.

If IPv6-only networks are now common, I wonder why we have not been getting complaints about not being able to use nrfcloud libraries in older NCS versions?

glarsennordic: > I was unware that we are now seeing carriers with IPv6-only networks.

It was not my intent to imply this, just to be clear. I don't know how common IPv6 is as a requirement.

But it is true that networks can, at any time, impose IPv6 on the PDN. So even if they are not doing it today, it seems within the realm of possibility to me that networks might soon start doing so.

That, coupled with the fact that we already have an IPv4/IPv6 fallback mechanism that seems to work 99% of the time, makes me hesitate to approve changing the defaults for our samples.

Have we reproduced the issue with IPv6 at any point? Last I checked in on it, we had attempted to reproduce it, but couldn't, because the cell networks we kept trying were enforcing IPv4. So until we can reproduce on an IPv6-capable network, I am furthermore hesitant to disable IPv6

plskeggs: Maybe this logic from Stig:

> The changes in commit https://github.com/nrfconnect/sdk-nrf/commit/57c5706f8f2ac13d5c1e0d064fdad27f9f9e2e04 fixed two issues where using AF_UNSPEC also would fail. One when the network does not give the device an IPv6 address (happens in many networks and when using MVNO SIMs), and another when the network does not correctly route IPv6 traffic (this happens for one of the big US carriers).

needs to be repeated everywhere we try to connect to a socket, not just in the download client.

glarsennordic: But I also don't want to steamroll -- I will retract my request for changes if you are still not convinced by the above

glarsennordic: > Maybe this logic from Stig needs to be repeated everywhere we try to connect to a socket, not just in the download client.

Yes, or perhaps integrated directly into the nrf91 offloaded socket interface (I ***think*** that would be possible)

plskeggs:
Closing -- seems to be unnecessary.

PR #14126, opened 2024-02-22
Remove option for user-provided Mbed TLS configuration header

mswarowsky:


SeppoTakalo:
Dropping from 2.6.0 as there was no approvals before RC1.

Vge0rge:
I would propose to not immediately remove MBEDTLS_USER_CONFIG_FILE but deprecate it first. I see some references to it form different projects (matter etc). The generate one I find no issue to remove it.

mswarowsky:
> I would propose to not immediately remove MBEDTLS_USER_CONFIG_FILE but deprecate it first. I see some references to it form different projects (matter etc). The generate one I find no issue to remove it.

Good point changed it to DEPRECATED

PR #14137, opened 2024-02-22
applications: nrf5340_audio: Fix CIS TWS MIC stream disappeared

rick1082:
Fixed the issue that CIS gateway might not setup MIC stream properly. There could be a chance the sink stream setup failed, but before the retry, the source stream from same ISO setup properly which also setup the sink. This cause the state out of sync for retrying, and leads to retry mechanism failed.
OCT-NONE

### DIFF
--- a/applications/nrf5340_audio/unicast_client/main.c
+++ b/applications/nrf5340_audio/unicast_client/main.c
@@ -255,6 +255,17 @@ static void le_audio_msg_sub_thread(void)
 			ERR_CHK(ret);
 			break;
 
+		case LE_AUDIO_EVT_NO_VALID_CFG:
+			LOG_WRN("No valid configurations found or CIS establishment failed, will "
+				"disconnect");
+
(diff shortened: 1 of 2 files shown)


koffes (approved):

applications/nrf5340_audio/unicast_client/main.c
```diff
@@ -255,6 +255,17 @@ static void le_audio_msg_sub_thread(void)
 			ERR_CHK(ret);
 			break;
 
+		case LE_AUDIO_EVT_NO_VALID_CFG:
+			LOG_WRN("No valid configurations found or CIS establishment failed, will "
+				"disconnect");
+
+			ret = bt_mgmt_conn_disconnect(msg.conn, BT_HCI_ERR_REMOTE_USER_TERM_CONN);
```
koffes, line 262: This is very reasonable. Needs a bit of manual testing to verify that all is OK

rick1082: Yes, I did some tests for using PCFT to trigger the CIS created failed, and we can trigger disconnection and then start everything from scratch.

alexsven (approved):

PR #14069, opened 2024-02-19
Bluetooth: Controller: Use chosen entropy device

hermabe:
NRF_RNG is not available on all platform, use the chosen device instead.

PR #14154, opened 2024-02-23
samples: Bluetooth: Mesh: add nrf54l support for mesh light sample

alxelax:
Add support of nrf54l15pdk_nrf54l15_cpuapp in mesh light
sample.

### DIFF
--- a/samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
+++ b/samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
@@ -1,0 +1,9 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+################################################################################
+# Application overlay - nrf54l15
+
+CONFIG_SOC_FLASH_NRF_PARTIAL_ERASE=n
--- a/samples/bluetooth/mesh/light/sample.yaml
+++ b/samples/bluetooth/mesh/light/sample.yaml
@@ -12,9 +12,10 @@ tests:
       - nrf5340dk_nrf5340_cpuapp_ns
       - thingy53_nrf5340_cpuapp
       - nrf21540dk_nrf52840
+      - nrf54l15pdk_nrf54l15_cpuapp
     platform_allow: nrf52dk_nrf52832 nrf52840dk_nrf52840 nrf5340dk_nrf5340_cpuapp
       nrf5340dk_nrf5340_cpuapp_ns thingy53_nrf5340_cpuapp
-      nrf21540dk_nrf52840 nrf52833dk_nrf52833
+      nrf21540dk_nrf52840 nrf52833dk_nrf52833 nrf54l15pdk_nrf54l15_cpuapp
     tags: bluetooth ci_build
   sample.bluetooth.mesh.light.dfu:
     build_only: true
--- a/samples/bluetooth/mesh/light/src/main.c
+++ b/samples/bluetooth/mesh/light/src/main.c
@@ -23,8 +23,17 @@ static void bt_ready(int err)
 
 	printk("Bluetooth initialized\n");
 
-	dk_leds_init();
-	dk_buttons_init(NULL);
+	err = dk_leds_init();
+	if (err) {
+		printk("Initializing LEDs failed (err %d)\n", err);
+		return;
+	}
+
+	err = dk_buttons_init(NULL);
+	if (err) {
+		printk("Initializing buttons failed (err %d)\n", err);
+		return;
+	}
 
 	err = bt_mesh_init(bt_mesh_dk_prov_init(), model_handler_init());
 	if (err) {
--- a/samples/bluetooth/mesh/ble_peripheral_lbs_coex/src/main.c
+++ b/samples/bluetooth/mesh/ble_peripheral_lbs_coex/src/main.c
@@ -24,8 +24,17 @@ static void bt_ready(int err)
 
 	printk("Bluetooth initialized\n");
 
-	dk_leds_init();
-	dk_buttons_init(NULL);
+	err = dk_leds_init();
+	if (err) {
+		printk("Initializing LEDs failed (err %d)\n", err);
+		return;
+	}
+
+	err = dk_buttons_init(NULL);
+	if (err) {
+		printk("Initializing buttons failed (err %d)\n", err);
+		return;
+	}
 
 	err = bt_mesh_init(bt_mesh_dk_prov_init(), model_handler_init());
 	if (err) {
--- a/samples/bluetooth/mesh/chat/src/main.c
+++ b/samples/bluetooth/mesh/chat/src/main.c
@@ -25,8 +25,17 @@ static void bt_ready(int err)
 
 	printk("Bluetooth initialized\n");
 
-	dk_leds_init();
-	dk_buttons_init(NULL);
+	err = dk_leds_init();
+	if (err) {
+		printk("Initializing LEDs failed (err %d)\n", err);
+		return;
+	}
+
+	err = dk_buttons_init(NULL);
+	if (err) {
+		printk("Initializing buttons failed (err %d)\n", err);
+		return;
+	}
 
 	err = bt_mesh_init(bt_mesh_dk_prov_init(), model_handler_init());
 	if (err) {
--- a/samples/bluetooth/mesh/dfu/distributor/src/main.c
+++ b/samples/bluetooth/mesh/dfu/distributor/src/main.c
@@ -112,7 +112,11 @@ static void bt_ready(int err)
 
 	printk("Bluetooth initialized\n");
 
-	dk_leds_init();
+	err = dk_leds_init();
+	if (err) {
+		printk("Initializing LEDs failed (err %d)\n", err);
+		return;
+	}
 
 	err = bt_mesh_init(bt_mesh_dk_prov_init(), &comp);
 	if (err) {
--- a/samples/bluetooth/mesh/dfu/target/src/main.c
+++ b/samples/bluetooth/mesh/dfu/target/src/main.c
@@ -97,7 +97,11 @@ static void bt_ready(int err)
 
 	printk("Bluetooth initialized\n");
 
-	dk_leds_init();
+	err = dk_leds_init();
+	if (err) {
+		printk("Initializing LEDs failed (err %d)\n", err);
+		return;
+	}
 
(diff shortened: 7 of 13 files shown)


alxelax:
These PRs should be merged first:
https://github.com/nrfconnect/sdk-nrf/pull/14060
https://github.com/nrfconnect/sdk-nrf/pull/14069

Andrewpini (commented):
Seems to work fine-ish at my desk. Did you experience issues/wierdness with the LEDs? In my setup, only `LED3` seemed to map correctly, and was the only one that worked as expected. The model instances seemed to do what the where supposed to do, so I think it is most likely an I/O issue. I got this log err: `<err> dk_buttons_and_leds: Cannot disable callbacks()`, which might be related.

Also, should the platform be added to the `sample.yaml` now, or is that a part of a later docs task you think?

Andrewpini:
FYI: I tried building the docs, both with and without the added platform in `sample.yaml`. Neither of the build seemed to be able to add the autogenerated list of supported targets correctly for some reason.

alxelax:
> Seems to work fine-ish at my desk. Did you experience issues/wierdness with the LEDs? In my setup, only `LED3` seemed to map correctly, and was the only one that worked as expected. The model instances seemed to do what the where supposed to do, so I think it is most likely an I/O issue. I got this log err: `<err> dk_buttons_and_leds: Cannot disable callbacks()`, which might be related.

There are tasks to add nrf54l platform to the samples(including the task for the light sample). I guess the correct assignment of LEDs, buttons, and probably something else should be done in appropriate tasks. There was no purpose for me to make the sample 100% ready. I just provided basic functionality to be able to run mesh in ncs.

> 
> Also, should the platform be added to the `sample.yaml` now, or is that a part of a later docs task you think?

As I mentioned on stand up there is a bug in the documentation generator. The platform wasn't added correctly yet. If add nrf54l in sample.yaml then this table cannot be built: [bug
](https://github.com/nrfconnect/sdk-nrf/blob/main/samples/bluetooth/mesh/light/README.rst?plain=1#L23)

Andrewpini (approved):
Ok, then I guess this is good to go 👍

ludvigsj:
Why do we seemingly arbitrary add this to the "light" sample? Why can't this be done when updating the sample to work properly when using nRF54L15?

alxelax:
> Why do we seemingly arbitrary add this to the "light" sample? Why can't this be done when updating the sample to work properly when using nRF54L15?

which sample should be chosen for this work from your point of view?

ludvigsj:
> > Why do we seemingly arbitrary add this to the "light" sample? Why can't this be done when updating the sample to work properly when using nRF54L15?
> 
> which sample should be chosen for this work from your point of view?

None of them. I'd prefer not doing this change until we actually create support (including GPIO mapping etc.) to a sample. Then we add it to that/those samples.

ludvigsj:
> I guess, the next task to run mesh on nrf54l will cause the same investigation that I did for people who pick it up. I'd prefer to keep the task value as an iterative result in repo.

No, because you can add your ~15 lines of code as a comment to that issue. But let's discuss this after daily tomorrow.

alxelax:
> > I guess, the next task to run mesh on nrf54l will cause the same investigation that I did for people who pick it up. I'd prefer to keep the task value as an iterative result in repo.
> 
> No, because you can add your ~15 lines of code as a comment to that issue. But let's discuss this after daily tomorrow.

The code should be kept in the repository. This is the place for it. Even more, the next iterative step for EMDS on nrf54l is not possible without this. Will you suggest adding emds related changes to JIRA either?

PavelVPV (changes requested):
Requesting change as I need time to look at this before it is merged.

alxelax:
Check if dtc overlay is required after this PR: https://github.com/nrfconnect/sdk-zephyr/pull/1518

alxelax:
Repurposed PR. This is adding nrf54l support to Mesh Light Sample.

samples/bluetooth/mesh/light/sample.yaml
ludvigsj, line 1: This will add this to the list of "Supported boards" in the documentation. Do we really want to claim the nRF54L15 PDK as a "supported board" for anything yet?

alxelax: This line is mandatory to add platform to twister.

ludvigsj: Yes I know but it also adds it to the docs which I'm a bit concerned about. I'm not so sure we want that yet, at the very least we should discuss it before adding text claiming formal support. If that means holding out on twister runs for now, or finding a workaround, we'll have to see.

ludvigsj: I agree, we should add it to CI. I just want to discuss if we want it in the docs (and then we have to figure out how to untie these from each other if we want CI but not doc update). I'm not a fan of the "remember to remove it before tagging 2.7.0"-approach, easy to forget and then suddenly we have angry customers having unstable nodes running on a "supported" board. But if this is the only way, then I guess this has to be the way to do it.

alxelax: We do not need to remember anything. There is JIRA and 2.7.0 tag in it.

samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+################################################################################
+# Application overlay - nrf54l15
+
+CONFIG_PM_SINGLE_IMAGE=n
```
PavelVPV, line 9: The sample works fine (probably after PM fix) without this option. Why should we touch it now?

samples/bluetooth/mesh/light/src/main.c
```diff
@@ -23,8 +23,17 @@ static void bt_ready(int err)
 
 	printk("Bluetooth initialized\n");
 
-	dk_leds_init();
-	dk_buttons_init(NULL);
+	err = dk_leds_init();
```
PavelVPV, line 26: Could you please apply this change for all samples (and in a separate commit)? **Yes, I know**, this is not part of the task, **but** since you are touching this part, it would be nice to have the same code in all samples. It will also help others if they also meet this problem with LEDs and buttons.

samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
```diff
@@ -0,0 +1,51 @@
...
+			label = "Push button 2";
+			zephyr,code = <INPUT_KEY_2>;
+		};
+		button3: button_3 {
+			gpios = <&gpio0 4 (GPIO_PULL_UP | GPIO_ACTIVE_LOW)>;
+			label = "Push button 3";
+			zephyr,code = <INPUT_KEY_3>;
+		};
+	};
+};
```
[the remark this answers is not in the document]
alxelax: Our team has different to the current official PDK version of boards. The ongoing dtc files are done according to the current official version. I asked @Balaklaka already about that, will our version become official or will not? I guess the process of figuring out is in progress. If you want that mesh worked on the boards you have this overlay must be.

samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
alxelax: Hm... It always shows a settings initialization error. Did you run it on my branch?

samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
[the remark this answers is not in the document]
PavelVPV: Then it should be <board_name>_<revision>.overlay, otherwise the next board will fail to work because of this overlay and we will have to change this configuration again. But then if someone will use old revision, they will have problems.

samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
PavelVPV: I pulled main branch and build it. No warning except the partial erase option.

alxelax: It is built normally - yes. However, when you run device it cannot pass settings initialization.

samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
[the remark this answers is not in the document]
PavelVPV: you pass @<revision> to select specific revision from what I know

samples/bluetooth/mesh/light/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
PavelVPV: It worked as well on my desk. Please try again.

PavelVPV:
Could you now explain why did you remove the overlay file?

alxelax:
> Could you now explain why did you remove the overlay file?

Supported revision should be added here first: https://github.com/nrfconnect/sdk-zephyr/blob/main/boards/arm/nrf54l15pdk_nrf54l15/revision.cmake#L8
Otherwise build system rejects all revision related overlays.

PavelVPV:
> > Could you now explain why did you remove the overlay file?
> 
> Supported revision should be added here first: https://github.com/nrfconnect/sdk-zephyr/blob/main/boards/arm/nrf54l15pdk_nrf54l15/revision.cmake#L8 Otherwise build system rejects all revision related overlays.

Ok, I see, then I'm fine with keeping the overlay for now. Please create a follow-up task to remove/fix it later (for 2.7.0 or earlier if it gets fixed in upstream). Please also add a note why the overlay is added (so that reviewers don't need to guess), you can use commit message or comment in the overlay for that.

omkar3141:
I checked-out the PR and tried it this does not work (does not advertise unprovisioned device beacon) without overlay. After manually adding overlay in the previous commit, I'm able to provision the devices, but LEDs connected to Generic OnOff server don't light-up after sending mesh messages.

PavelVPV (approved):

ludvigsj (approved):

PR #14196, opened 2024-02-26
trusted_storage: Fix Return code errors

mswarowsky:
Align return codes and some behavior with the PSA ITS tests provided by the TF-M project.

Ref: NCSDK-26234


vili-nordic (approved):

subsys/trusted_storage/src/aead/trusted_backend_aead.c
```diff
@@ -70,19 +72,24 @@ psa_status_t trusted_get_info(const psa_storage_uid_t uid, const char *prefix,
...
+			 size_t data_length, void *p_data, size_t *p_data_length)
 {
 	psa_status_t status = PSA_ERROR_CORRUPTION_DETECTED;
 	uint8_t key_buf[AEAD_KEY_SIZE + 1];
 	size_t out_length;
 	stored_object object_data;
 
-	if (data_length == 0 || p_data == NULL || p_data_length == NULL) {
+	if (data_length == 0) {
+		*p_data_length = 0;
```
Vge0rge, line 83: Please move this code after the if statement that we check if p_data_length is NULL.

Vge0rge (changes requested):

subsys/trusted_storage/src/aead/trusted_backend_aead.c
mswarowsky: fixed

Vge0rge (approved):

PR #14201, opened 2024-02-26
Experiment with svg font polyfill

doki-nordic:
This is just an experiment. It is not indented to be merge. It is here just to see docs build results on the server.


doki-nordic:
Declined

PR #14202, opened 2024-02-26
applications: nrf5340_audio: change SMP log level

rick1082:
Update the SMP log level, since SMP failed could happen during power on/off testing, which should not block us from continuing the test.
OCT-NONE


koffes (approved):

asbjornsabo (approved):
LGTM

PR #14207, opened 2024-02-26
nrf5340: use minimal configuration for b0n by default

LuDuda:
This commit configures NCS samples on nRF5340 to use the minimal configuration for b0 network core bootloader.

Currently, all samples use debug variant of b0 network core bootloader e.g. with serial and console being enabled - which increases the boot time and power consumption on boot. Additionally, with this change, we save around 24kB of flash memory. Any reason to keep the debug variant by default in our samples?


LuDuda:
It seems that this is a breaking change and DFU from previous versions won't work. So it requires more discussion whether we want to do it.

PR #14210, opened 2024-02-27
samples: wifi: throughput: Add profile for memory footprints.

visinor:
Add samples for memory footprints for these profile
1. IoT Deviceis - overlay-iot-devices.conf
2. Memory Optimized - overlay-memory-optimized.conf
3. High Performance - overlay-high-performance.conf
4. TX prioritized - overlay-tx-prio.conf
5. RX prioritized - overlay-rx-prio.conf

### DIFF
--- a/samples/wifi/throughput/sample.yaml
+++ b/samples/wifi/throughput/sample.yaml
@@ -9,3 +9,39 @@ tests:
       - nrf7002dk_nrf5340_cpuapp
     platform_allow: nrf7002dk_nrf5340_cpuapp
     tags: ci_build
+  # Used by QA to measure memory footprints
+  sample.nrf7002.iot_devices:
+    build_only: true
+    extra_args: OVERLAY_CONFIG=overlay-iot-devices.conf
+    integration_platforms:
+      - nrf7002dk_nrf5340_cpuapp
+    platform_allow: nrf7002dk_nrf5340_cpuapp
+    tags: ci_build
+  sample.nrf7002.high_performance:
+    build_only: true
+    extra_args: OVERLAY_CONFIG=overlay-high-performance.conf
+    integration_platforms:
+      - nrf7002dk_nrf5340_cpuapp
+    platform_allow: nrf7002dk_nrf5340_cpuapp
+    tags: ci_build
+  sample.nrf7002.memory_optimized:
+    build_only: true
+    extra_args: OVERLAY_CONFIG=overlay-memory-optimized.conf
+    integration_platforms:
+      - nrf7002dk_nrf5340_cpuapp
+    platform_allow: nrf7002dk_nrf5340_cpuapp
+    tags: ci_build
+  sample.nrf7002.rx_prioritized:
+    build_only: true
+    extra_args: OVERLAY_CONFIG=overlay-rx-prio.conf
+    integration_platforms:
+      - nrf7002dk_nrf5340_cpuapp
+    platform_allow: nrf7002dk_nrf5340_cpuapp
+    tags: ci_build
+  sample.nrf7002.tx_prioritized:
+    build_only: true
+    extra_args: OVERLAY_CONFIG=overlay-tx-prio.conf
+    integration_platforms:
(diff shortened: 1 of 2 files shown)


D-Triveni (approved):

samples/wifi/throughput/sample.yaml
```diff
@@ -9,3 +9,43 @@ tests:
...
     platform_allow: nrf7002dk_nrf5340_cpuapp
     tags: ci_build
+  sample.nrf7002.iot_devices:
+    build_only: true
+    extra_args: OVERLAY_CONFIG=overlay-iot-devices.conf
+    integration_platforms:
+      - nrf7002dk_nrf5340_cpuapp
+    platform_allow: nrf7002dk_nrf5340_cpuapp
+    tags: ci_build
+    skip: true
```
krish2718, line 19: Can you please add a comment explaining why?

krish2718 (approved):

samples/wifi/throughput/sample.yaml
visinor: We added `skip: true` to ensure that CI time is not impacted, allowing us to run Twister on these samples to measure memory footprint.

krish2718: I know :), I was referring to adding a comment in the YAML file above `skip: true`

PerMac: The explanation makes no sens to me. **Have you tried running twister on anything which has `skip: true`?** 
Such entry shouldn't be used at all in samples. It just messes with the CI an documentation. It tell twister to ALWAYS skip the build sample and makes it untestable in the CI. It is a way how upstream wanted to skip a broken builds to not break CI, but we have a quarantine for this now.

PerMac:
I am not sure I understand the need for getting experimental status without building. We do build everything at night, so you can have it from there.

PerMac:
the best way to don't have a build on PRs but still get in on nightly run:
Make sure that every configuration that you support have integration_platforms field with all the platform that should work during nightly build. Having platform under integration_platforms make sure it won't be skipped, unless it is quarantined.
Then, to limit the amount of configurations that are build on PRs please add the ones you don't want to be built on PR here https://github.com/nrfconnect/sdk-nrf/blob/main/scripts/quarantine_integration.yaml
This way those configuration will be quarantined (hence legally skipped) on PR, but will be still build on nightly (quarantine_integration is only used on PRs)

samples/wifi/throughput/sample.yaml
krish2718: As discussed https://github.com/nrfconnect/sdk-nrf/pull/14210#issuecomment-1970779858 we can:

* add an entry to sample.yaml without `skip: true`
* Add that case to quarantine-integration
* It will be part of nightly

SeppoTakalo:
This is just CI build configurations, so I will remove the 2.6.0 tag as this is not relevant to release itself.
Thanks.

samples/wifi/throughput/sample.yaml
visinor: @PerMac Updated changes as per comments, please review.

krish2718 (approved):

PerMac (approved):

jfischer-no:
> This is just CI build configurations, so I will remove the 2.6.0 tag as this is not relevant to release itself. Thanks.

It does not seem to be a bugfix, why was it added back to the milestone?

SeppoTakalo:
Still no justification have been given why this should go into release, so I'll drop the milestone marker.

krish2718:
> Still no justification have been given why this should go into release, so I'll drop the milestone marker.

This doesn't impact any existing samples, and this helps in extending documentation at https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/protocols/wifi/station_mode/mem_requirements_sta.html where our customer's are asking for more details.

IMHO, this is just a doc update and eligible to be merged as part of RC2, WDYT?

SeppoTakalo (approved):

PR #14218, opened 2024-02-27
applications: serial_lte_modem: add reference overlays for Zephyr modem driver

tomi-font:
Documentation will follow.

### DIFF
--- a/applications/serial_lte_modem/overlay-zephyr-modem.conf
+++ b/applications/serial_lte_modem/overlay-zephyr-modem.conf
@@ -1,0 +1,10 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# The nRF91 is power-managed by Zephyr's modem_cellular driver running on the other chip.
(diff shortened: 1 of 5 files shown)


applications/serial_lte_modem/overlay-zephyr-modem.conf
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# The nRF91 is power-managed by Zephyr's modem_cellular driver running on the other chip.
```
MarkusLassila, line 7: If the chip is always the 52840, that could be mentioned.

MarkusLassila (approved):

SeppoTakalo (approved):

applications/serial_lte_modem/overlay-zephyr-modem.conf
tomi-font: No it's not, that's why the comment is generic.
This one is to be included in any build that is making SLM work with Zephyr's modem driver, be it on the 52840, an external MCU...

PR #14220, opened 2024-02-27
manifest: update sdk-zephyr

maxd-nordic:
This patch updates sdk-zephyr to include a doc pr.


jfischer-no (changes requested):
Please do not update sdk-zephyr revision here.

PR #14228, opened 2024-02-27
modules: trusted-firmware-m: Add NFC pins option

Vge0rge:
Add option for configuring the NFC pins in TF-M.

The option CONFIG_NFCT_PINS_AS_GPIOS in Zephyr
is being deprecated and it will removed soon in
favor of DTS configuration for this.

TF-m doesn't support DTS and thus we cannot use
the DTS option.

Add a separate option for TF-M which will not be
deprecated.

Ref: NCSDK-24437


modules/trusted-firmware-m/Kconfig
```diff
@@ -295,6 +295,18 @@ config TFM_PROFILE_TYPE_MINIMAL
 	  and reproduce the desired configuration through kconfig fragments.
 endchoice
 
+config TFM_NFCT_PINS_AS_GPIOS
+	bool "TF-M configure NFCT pins as GPIOs"
+	depends on SOC_NRF5340_CPUAPP
```
SebastianBoe, line 300:
```suggestion
	bool
	depends on $(dt_has_compat,$(DT_COMPAT_NORDIC_NRF_NFCT))
```

Let's leave this as a promptless option so that the user must configure this through DT in the same way for TF-M and non-TF-M builds.

Unless I'm misunderstanding something?

vili-nordic: Both nrf54h and nrf54l has this option in the mdk, so depends on SOC_NRF5340_CPUAPP is not very future proof.

Note that it's randomly called NRF_CONFIG_NFCT_PINS_AS_GPIOS and CONFIG_NFCT_PINS_AS_GPIOS in these mdks, so there's probably more issues to be handled there.

Vge0rge: Your depends actually is great. But we cannot make it promptless because this is configured though the UICR node in DTS. And we don't have a UICR node in the non-secure targets.

Vge0rge: The usage of NRF_CONFIG_NFCT_PINS_AS_GPIOS seems like a hack to me. So hopefully we can revert this and use only CONFIG_NFCT_PINS_AS_GPIOS, right? 
Are you suggesting to enable this here as well @vili-nordic ?

vili-nordic (approved):

vili-nordic:
Are there documentation we need to update for this?

Vge0rge:
I added documentation in the Zephyr Kconfig option so that the users can see the new option.

SebastianBoe (approved):

gchwier (approved):
One sample in test-fw-nrfconnect-boot had to be updated, because with this PR the NFC peripheral are disabled by default in DTS, that caused some problems in one testcase for  nrf5340dk_nrf5340_cpuapp_ns.
After updating sample, tests from test-fw-nrfconnect-boot are PASSED

miha-nordic (approved):

PR #14237, opened 2024-02-27
Bluetooth: Controller: Integrate Zephyr PM calls in MPSL

KyraLengfeld:
This approach enables us to call appropriate Power Management policies
from zephyr Power Management.

Within MPSL a global struct is defined and modified from high priority
context whenever an event is scheduled/rescheduled/...

Within nrf-sdk this global struct is read and appropriate zephyr
functions are called.

Note: If High Prio changed the parameters already before we could
process them, do nothing, we will get back here with next workqueue
item.

Unit Tests are added accordingly.

### DIFF
--- a/include/mpsl/mpsl_pm_utils.h
+++ b/include/mpsl/mpsl_pm_utils.h
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef MPSL_PM_UTILS_H__
+#define MPSL_PM_UTILS_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** @brief Initialize MPSL Power Management
+ *
+ * This routine initializes MPSL PM (via `mpsl_pm_init`).
+ */
+void mpsl_pm_utils_init(void);
+
+/** @brief Handles MPSL Power Management work
+ *
+ * This calls Zephyr Power Management policy functions according
+ * to MPSL PM requirements.
+ *
+ * @pre mpsl_pm_utils_init() needs to have been called before.
+ */
+void mpsl_pm_utils_work_handler(void);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* MPSL_PM_UTILS_H__ */
--- a/include/mpsl/mpsl_work.h
+++ b/include/mpsl/mpsl_work.h
@@ -41,12 +41,37 @@ extern struct k_work_q mpsl_work_q;
  * @note Can be called by ISRs.
  *
  * @param work Address of work item.
- *
- * @return N/A
  */
-static inline int mpsl_work_submit(struct k_work *work)
+static inline void mpsl_work_submit(struct k_work *work)
 {
-	return k_work_submit_to_queue(&mpsl_work_q, work);
+	if (k_work_submit_to_queue(&mpsl_work_q, work) < 0) {
+		__ASSERT(false, "k_work_submit_to_queue() failed.");
+	}
+}
+
+/** @brief Submit an idle work item to the MPSL workqueue after a delay.
+ *
+ * @note Can be called by ISRs.
+ *
+ * @note
+ * Work items submitted to the MPSL workqueue should avoid using handlers
+ * that block or yield since this may prevent the MPSL workqueue from
+ * processing other work items in a timely manner.
+ *
+ * @note This is a no-op if the work item is already scheduled or submitted,
+ * even if @p delay is @c K_NO_WAIT.
+ *
+ * @param dwork Address of delayable work item.
+ *
+ * @param delay the time to wait before submitting the work item.  If @c
+ * K_NO_WAIT and the work is not pending this is equivalent to
+ * mpsl_work_submit().
+ */
+static inline void mpsl_work_schedule(struct k_work_delayable *dwork, k_timeout_t delay)
+{
+	if (k_work_schedule_for_queue(&mpsl_work_q, dwork, delay) < 0) {
+		__ASSERT(false, "k_work_schedule_for_queue() failed.");
+	}
 }
 
 #ifdef __cplusplus
--- a/subsys/mpsl/init/mpsl_init.c
+++ b/subsys/mpsl/init/mpsl_init.c
@@ -22,6 +22,10 @@
 #include <hal/nrf_ipc.h>
 #endif
 
+#if IS_ENABLED(CONFIG_MPSL_USE_ZEPHYR_PM)
+#include <pm/mpsl_pm_utils.h>
+#endif
+
 LOG_MODULE_REGISTER(mpsl_init, CONFIG_MPSL_LOG_LEVEL);
 
 /* The following two constants are used in nrfx_glue.h for marking these PPI
@@ -165,7 +169,7 @@ static uint8_t __aligned(4) timeslot_context[TIMESLOT_MEM_SIZE];
 
 static void mpsl_low_prio_irq_handler(const void *arg)
 {
-	k_work_submit_to_queue(&mpsl_work_q, &mpsl_low_prio_work);
+	mpsl_work_submit(&mpsl_low_prio_work);
 }
 
 static void mpsl_low_prio_work_handler(struct k_work *item)
@@ -176,7 +180,13 @@ static void mpsl_low_prio_work_handler(struct k_work *item)
 
 	errcode = MULTITHREADING_LOCK_ACQUIRE();
 	__ASSERT_NO_MSG(errcode == 0);
+
 	mpsl_low_priority_process();
+
+#if IS_ENABLED(CONFIG_MPSL_USE_ZEPHYR_PM)
+	mpsl_pm_utils_work_handler();
+#endif
+
 	MULTITHREADING_LOCK_RELEASE();
 }
 
@@ -329,8 +339,8 @@ static void mpsl_calibration_work_handler(struct k_work *work)
 
 	mpsl_calibration_timer_handle();
 
-	k_work_schedule_for_queue(&mpsl_work_q, &calibration_work,
-				  K_MSEC(CONFIG_CLOCK_CONTROL_NRF_CALIBRATION_PERIOD));
+	mpsl_work_schedule(&calibration_work,
+			   K_MSEC(CONFIG_CLOCK_CONTROL_NRF_CALIBRATION_PERIOD));
 }
 #endif /* CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC && !CONFIG_SOC_SERIES_NRF54HX */
 
@@ -406,6 +416,10 @@ static int mpsl_lib_init_sys(void)
 		return err;
 	}
 
+#if IS_ENABLED(CONFIG_MPSL_USE_ZEPHYR_PM)
+	mpsl_pm_utils_init();
+#endif
+
 #if IS_ENABLED(CONFIG_MPSL_DYNAMIC_INTERRUPTS)
 	/* Ensure IRQs are disabled before attaching. */
 	mpsl_lib_irq_disable();
@@ -449,8 +463,8 @@ static int mpsl_low_prio_init(void)
 		    mpsl_low_prio_irq_handler, NULL, 0);
 
 #if defined(CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC) && !defined(CONFIG_SOC_SERIES_NRF54HX)
-	k_work_schedule_for_queue(&mpsl_work_q, &calibration_work,
-				  K_MSEC(CONFIG_CLOCK_CONTROL_NRF_CALIBRATION_PERIOD));
+	mpsl_work_schedule(&calibration_work,
+			   K_MSEC(CONFIG_CLOCK_CONTROL_NRF_CALIBRATION_PERIOD));
 #endif /* CONFIG_CLOCK_CONTROL_NRF_K32SRC_RC && !CONFIG_SOC_SERIES_NRF54HX */
 
 	return 0;
--- a/subsys/mpsl/pm/CMakeLists.txt
+++ b/subsys/mpsl/pm/CMakeLists.txt
@@ -1,0 +1,9 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+zephyr_library()
+
+zephyr_library_sources(mpsl_pm_utils.c)
--- a/subsys/mpsl/pm/Kconfig
+++ b/subsys/mpsl/pm/Kconfig
@@ -1,0 +1,13 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config MPSL_USE_ZEPHYR_PM
+	bool "Use Zephyr's Power Management API"
+	depends on SOC_SERIES_NRF54HX
+	depends on MPSL
+	depends on PM
+	help
+	  This option configures MPSL to use Zephyr's Power Management.
--- a/subsys/mpsl/pm/mpsl_pm_utils.c
+++ b/subsys/mpsl/pm/mpsl_pm_utils.c
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/kernel.h>
+#include <mpsl_pm.h>
+#include <mpsl_pm_config.h>
+#include <zephyr/pm/policy.h>
+#include <zephyr/logging/log.h>
+
+#include <mpsl/mpsl_work.h>
+#include <mpsl/mpsl_pm_utils.h>
+
+LOG_MODULE_REGISTER(mpsl_pm_utils, CONFIG_MPSL_LOG_LEVEL);
+
+/* These constants must be updated once the Zephyr PM Policy API is updated
+ * to handle low latency events. Ideally, the Policy API should be changed to use
+ * absolute time instead of relative time. This would remove the need for safety
+ * margins and allow optimal power savings.
+ */
+#define MAX_DELAY_SINCE_READING_PARAMS_US 50
+#define TIME_TO_REGISTER_EVENT_IN_ZEPHYR_US 1000
+#define PM_MAX_LATENCY_HCI_COMMANDS_US 4999999
+
+static void m_work_handler(struct k_work *work);
+static K_WORK_DELAYABLE_DEFINE(pm_work, m_work_handler);
+
+#define RETRY_TIME_MAX_US (UINT32_MAX - TIME_TO_REGISTER_EVENT_IN_ZEPHYR_US)
+
+static uint8_t                          m_pm_prev_flag_value;
+static bool                             m_pm_event_is_registered;
+static uint32_t                         m_prev_lat_value_us;
+static struct pm_policy_latency_request m_latency_req;
+static struct pm_policy_event           m_evt;
+
+
+static void m_update_latency_request(uint32_t lat_value_us)
+{
+	if (m_prev_lat_value_us != lat_value_us) {
+		pm_policy_latency_request_update(&m_latency_req, lat_value_us);
+		m_prev_lat_value_us = lat_value_us;
+	}
+}
+
+void mpsl_pm_utils_work_handler(void)
+{
+	mpsl_pm_params_t params	= {0};
+	bool pm_param_valid = mpsl_pm_params_get(&params);
+
+	if (m_pm_prev_flag_value == params.cnt_flag) {
+		/* We have no new info to process.*/
+		return;
+	}
+	if (!pm_param_valid) {
+		/* High prio did change mpsl_pm_params, while we read the params. */
+		m_pm_prev_flag_value = params.cnt_flag;
+		return;
+	}
+	switch (params.event_state) {
+	case MPSL_PM_EVENT_STATE_NO_EVENTS_LEFT:
+	{
+		/* No event scheduled, so set latency to restrict deepest sleep states*/
+		m_update_latency_request(PM_MAX_LATENCY_HCI_COMMANDS_US);
+		if (m_pm_event_is_registered) {
+			pm_policy_event_unregister(&m_evt);
+			m_pm_event_is_registered = false;
+		}
+		break;
+	}
+	case MPSL_PM_EVENT_STATE_BEFORE_EVENT:
+	{
+		/* Event scheduled */
+		uint64_t event_time_us = params.event_time_rel_us -
+						MAX_DELAY_SINCE_READING_PARAMS_US;
+
+		/* In case we missed a state and are in zero-latency, set low-latency.*/
+		m_update_latency_request(PM_MAX_LATENCY_HCI_COMMANDS_US);
+
+		if (event_time_us > UINT32_MAX) {
+			mpsl_work_schedule(&pm_work, K_USEC(RETRY_TIME_MAX_US));
+			return;
+		}
+
+		if (m_pm_event_is_registered) {
+			pm_policy_event_update(&m_evt, event_time_us);
+		} else {
+			pm_policy_event_register(&m_evt, event_time_us);
+			m_pm_event_is_registered = true;
+		}
+		break;
+	}
+	case MPSL_PM_EVENT_STATE_IN_EVENT:
+	{
+		m_update_latency_request(0);
+		break;
+	}
+	default:
+		__ASSERT(false, "MPSL PM is in an undefined state.");
+	}
+	m_pm_prev_flag_value = params.cnt_flag;
+}
+
+static void m_work_handler(struct k_work *work)
+{
+	ARG_UNUSED(work);
+	mpsl_pm_utils_work_handler();
+}
+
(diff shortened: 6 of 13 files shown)


subsys/mpsl/init/mpsl_init.c
```diff
@@ -153,9 +175,86 @@ BUILD_ASSERT(MPSL_TIMESLOT_SESSION_COUNT <= MPSL_TIMESLOT_CONTEXT_COUNT_MAX,
...
+			/* Event scheduled */
+			uint64_t wakeup_latency_us = params.event_time_rel_us -
+			                             MAX_DELAY_SINCE_READING_PARAMS;
+
+			if (wakeup_latency_us > UINT32_MAX) {
+				uint32_t retry_time_us = UINT32_MAX -
+							 TIME_TO_REGISTER_EVENT_IN_ZEPHYR;
+
+				k_work_schedule_for_queue(&mpsl_work_q, &pm_work,
+							  K_MSEC(retry_time_us));
```
knutel-nordic, line 234:
```suggestion
							  K_USEC(retry_time_us));
```

subsys/mpsl/init/mpsl_init.c
```diff
@@ -153,9 +175,86 @@ BUILD_ASSERT(MPSL_TIMESLOT_SESSION_COUNT <= MPSL_TIMESLOT_CONTEXT_COUNT_MAX,
...
+				pm_policy_event_update(&pm_policy_event, wakeup_latency_us);
+			} else {
+				pm_policy_event_register(&pm_policy_event, wakeup_latency_us);
+			}
+			m_pm_event_is_scheduled = true;
+		} else {
+			/* Event started */
+			/* we are in and event, set latency 0*/
+			lat_value_us = 0;
+			m_update_latency_pm(lat_value_us);
```
knutel-nordic, line 248:
```suggestion
			m_update_latency_pm(0);
```

subsys/mpsl/init/mpsl_init.c
```diff
@@ -324,6 +430,29 @@ static void mpsl_calibration_work_handler(struct k_work *work)
...
+	ARG_UNUSED(work);
+	m_pm_work();
+}
+
+static void m_pm_init(void)
+{
+	mpsl_pm_params_t params = {0};
+	/* So that PM doesn't shut us off power when we are idle.*/
+	/* see min. residency times: https://github.com/nrfconnect/sdk-sysctrl/blob/master/sysctrl/conf/pm.overlay#L19*/
+	/* we want max sleep?? -> PM_MAX_LATENCY_IDLE_US < 5000000us*/
```
knutel-nordic, line 445: Maybe this is not needed and should rather be determined by the max latency for starting to process an HCI command? We will not go to system off automatically - that has to be done explicitly by the application.

knutel-nordic (changes requested):

subsys/mpsl/init/mpsl_init.c
```diff
@@ -153,9 +175,86 @@ BUILD_ASSERT(MPSL_TIMESLOT_SESSION_COUNT <= MPSL_TIMESLOT_CONTEXT_COUNT_MAX,
...
+			pm_policy_event_unregister(&pm_policy_events);
+			m_pm_event_is_scheduled = false;
+		} else if (params.event_state == MPSL_PM_EVENT_STATE_BEFORE_EVENT) {
+			/* Event scheduled */
+			uint64_t wakeup_latency_us = params.event_time_rel_us -
+			                             MAX_DELAY_SINCE_READING_PARAMS;
+
+			if (wakeup_latency_us > UINT32_MAX) {
+				uint32_t retry_time_us = UINT32_MAX -
+							 TIME_TO_REGISTER_EVENT_IN_ZEPHYR;
```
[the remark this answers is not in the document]
KyraLengfeld: Looking what is behind `K_MSEC`, it takes a uint32_t, and in this case `wakeup_latency_us - TIME_TO_REGISTER_EVENT_IN_ZEPHYR` is likely to exceed that.

knutel-nordic, line 231:
```
#define RETRY_TIME_MAX_US (UINT32_MAX - TIME_TO_REGISTER_EVENT_IN_ZEPHYR)
```
```suggestion
				uint32_t retry_time_us = RETRY_TIME_MAX_US;
```

tests/subsys/mpsl/pm/pm_test.c
```diff
@@ -0,0 +1,216 @@
...
+	m_pm_init(&test_params);
+
+	/* After init we expect to have set non-zero-low-latency,
+	 * latency update to zero-latency is expected.
+	 */
+	test_params.event_state = MPSL_PM_EVENT_STATE_IN_EVENT;
+	m_pm_work_handling_update_latency(&test_params, true);
+
+	/* Since zero-latency is set now, no latency update is expected.*/
+	m_pm_work_handling_update_latency(&test_params, false);
```
knutel-nordic, line 134: Please re-initialize `test_params` completely first.

knutel-nordic (changes requested):
Consider a table based approach: https://gist.github.com/knutel-nordic/4207c6d606d4b6e6d94e7cc03e86e396

tests/subsys/mpsl/pm/pm_test.c
```diff
@@ -0,0 +1,216 @@
...
+
+	p_test_params->cnt_flag++;
+	p_test_params->event_time_rel_us = test_rel_evt_time_us;
+
+	while ((test_rel_evt_time_us - MAX_DELAY_SINCE_READING_PARAMS) > UINT32_MAX) {
+		__cmock_mpsl_pm_params_get_ExpectAnyArgsAndReturn(true);
+		__cmock_mpsl_pm_params_get_ReturnThruPtr_p_params(p_test_params);
+		p_test_params->event_time_rel_us = p_test_params->event_time_rel_us -
+						   MAX_DELAY_SINCE_READING_PARAMS;
+	}
```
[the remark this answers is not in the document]
KyraLengfeld: Because I am testing PM and using the real k_work stuffs, it would reenter (or rather call again) pm_work_handling in case of delayed handling. Since the UT would fail that expectations aren't met I assumed that this is actually how it is also working.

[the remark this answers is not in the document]
knutel-nordic: The tests pass if these lines are deleted, too.

tests/subsys/mpsl/pm/pm_test.c
```diff
@@ -0,0 +1,227 @@
...
+	bool new_test;
+	bool pm_params_get_retval;
+	mpsl_pm_params_t params;
+	event_func_t event_func;
+	uint64_t event_time_us;
+	latency_func_t latency_func;
+	uint64_t latency_us;
+} test_vector_t;
+
+void test_pm_with_table(void)
```
knutel-nordic, line 62:
```suggestion
void test_pm(void)
```
or break up in smaller functions. See below.

knutel-nordic (changes requested):

tests/subsys/mpsl/pm/pm_test.c
```diff
@@ -0,0 +1,227 @@
...
+				break;
+			case EVENT_FUNC_UPDATE:
+				__cmock_pm_policy_event_update_Expect(0, v.event_time_us);
+				__cmock_pm_policy_event_update_IgnoreArg_evt();
+				break;
+			case EVENT_FUNC_UNREGISTER:
+				__cmock_pm_policy_event_unregister_ExpectAnyArgs();
+				break;
+			case EVENT_FUNC_DELAY_SCHEDULING:
+				__cmock_mpsl_work_schedule_ExpectAnyArgsAndReturn(1);
```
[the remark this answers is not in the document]
KyraLengfeld: I think there was some trouble regarding expansion of K_USEC and then missing includes from that, but I will try. Will discuss in chat if there are issues.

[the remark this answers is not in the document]
knutel-nordic: If that does not work, then we can change `mpsl_work_schedule` to take a value in microseconds instead.

subsys/mpsl/pm/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+zephyr_library()
+
+zephyr_library_sources_ifdef(CONFIG_MPSL_USE_ZEPHYR_PM mpsl_pm_utils.c)
```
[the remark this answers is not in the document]
KyraLengfeld: true, as there's just one config option for PM, it's better to remove this.

subsys/mpsl/pm/mpsl_pm_utils.c
```diff
@@ -0,0 +1,123 @@
...
+static K_WORK_DELAYABLE_DEFINE(pm_work, m_work_handler);
+
+#define RETRY_TIME_MAX_US (UINT32_MAX - TIME_TO_REGISTER_EVENT_IN_ZEPHYR)
+
+static uint8_t				m_pm_prev_flag_value;
+static bool				m_pm_latency_idle_set;
+static bool				m_pm_event_is_registered;
+static uint32_t			m_prev_lat_value_us;
+static struct pm_policy_latency_request	m_latency_req;
+static struct pm_policy_event			m_evt;
```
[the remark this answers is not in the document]
KyraLengfeld: I do, just that my vs code is not set to the same tab size (display wise) and event the normal text editor presents it sometimes different. (I mean most times the alignment is the same as can be seen in these diffs on github, but sometimes, like this case, they aren't, which is strange, probably should investigate why.)

tests/subsys/mpsl/pm/pm_test.c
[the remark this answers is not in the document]
KyraLengfeld: It does work, probably had some other issues causing strange behaviour previously.

include/mpsl/mpsl_pm_utils.h
```diff
@@ -0,0 +1,17 @@
...
+
+#ifndef MPSL_PM_UTILS_H__
+#define MPSL_PM_UTILS_H__
+
+#include <stdint.h>
+#include <stddef.h>
+
+void mpsl_pm_utils_init(void);
+
+void mpsl_pm_utils_work_handler(void);
```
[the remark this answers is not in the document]
KyraLengfeld: Please see if the phrasing is ok.

subsys/mpsl/pm/mpsl_pm_utils.c
```diff
@@ -0,0 +1,123 @@
...
+/* Max latency for starting to process an HCI command. */
+#define PM_MAX_LATENCY_HCI_COMMANDS_US 4999999
+
+static void m_work_handler(struct k_work *work);
+static K_WORK_DELAYABLE_DEFINE(pm_work, m_work_handler);
+
+#define RETRY_TIME_MAX_US (UINT32_MAX - TIME_TO_REGISTER_EVENT_IN_ZEPHYR)
+
+static uint8_t					m_pm_prev_flag_value;
+static bool					m_pm_latency_idle_set;
```
knutel-nordic, line 30:
```suggestion
```
Not used for anything.

subsys/mpsl/pm/mpsl_pm_utils.c
```diff
@@ -0,0 +1,123 @@
...
+	mpsl_pm_utils_work_handler();
+}
+
+void mpsl_pm_utils_init(void)
+{
+	mpsl_pm_params_t params = {0};
+
+	pm_policy_latency_request_add(&m_latency_req, PM_MAX_LATENCY_HCI_COMMANDS_US);
+	m_prev_lat_value_us = PM_MAX_LATENCY_HCI_COMMANDS_US;
+	m_pm_latency_idle_set = true;
```
knutel-nordic, line 117:
```suggestion
```
This was the only usage.

knutel-nordic (changes requested):

include/mpsl/mpsl_work.h
```diff
@@ -49,6 +49,32 @@ static inline int mpsl_work_submit(struct k_work *work)
...
+ * @note This is a no-op if the work item is already scheduled or submitted,
+ * even if @p delay is @c K_NO_WAIT.
+ *
+ * @param dwork Address of delayable work item.
+ *
+ * @param delay the time to wait before submitting the work item.  If @c
+ * K_NO_WAIT and the work is not pending this is equivalent to
+ * mpsl_work_submit().
+ *
+ * @return N/A
```
knutel-nordic, line 70: I'm not sure why the existing function had this. If the return value from `k_work_*` is negative, we should assert. The Zephyr docs make a point about the importance of checking the return values: https://docs.zephyrproject.org/latest/kernel/services/threads/workqueue.html#check-return-values.

subsys/mpsl/pm/mpsl_pm_utils.c
KyraLengfeld: Leftover from initial approach, will remove :)

subsys/mpsl/pm/mpsl_pm_utils.c
```diff
@@ -0,0 +1,123 @@
...
+#include <mpsl_pm_config.h>
+#include <zephyr/pm/policy.h>
+#include <zephyr/logging/log.h>
+
+#include <mpsl/mpsl_work.h>
+#include <mpsl/mpsl_pm_utils.h>
+
+LOG_MODULE_REGISTER(mpsl_pm_utils, CONFIG_MPSL_LOG_LEVEL);
+
+/* These values are still arbritrary*/
```
KyraLengfeld, line 18: As we cannot predict what gets executed in between in high prio, nor really how many non-PM items are in the workqueue beforehand, this is very hard to determine. Testing it is currently also not viable, as the zephyr implementation is non functional. (will be worked on soon--I hope)

katgiadla (dismissed):
Please specify the responsible person for this test in the file `CODEOWNERS`.

include/mpsl/mpsl_work.h
```diff
@@ -42,13 +42,55 @@ extern struct k_work_q mpsl_work_q;
...
+ * @retval 2 if work was running and has been queued to the queue that was
+ * running it
+ * @retval -EBUSY
+ * * if work submission was rejected because the work item is cancelling; or
+ * * @p queue is draining; or
+ * * @p queue is plugged.
+ * @retval -EINVAL if @p queue is null and the work item has never been run.
+ * @retval -ENODEV if @p queue has not been started.
  */
 static inline int mpsl_work_submit(struct k_work *work)
```
olivier-le-sage, line 56: It's maybe a bit simpler if this and `mpsl_work_schedule` return void and assert inside the functions themselves. Then we don't need to repeat the zephyr docs for `k_work_submit_to_queue()` and `k_work_schedule_for_queue()`

subsys/mpsl/pm/mpsl_pm_utils.c
```diff
@@ -0,0 +1,123 @@
...
+		break;
+	}
+	case MPSL_PM_EVENT_STATE_BEFORE_EVENT:
+	{
+		/* Event scheduled */
+		uint64_t event_time_us = params.event_time_rel_us -
+						MAX_DELAY_SINCE_READING_PARAMS;
+
+		/* In case we missed a state and are in zero-latency, set low-latency.*/
+		m_update_latency_request(PM_MAX_LATENCY_HCI_COMMANDS_US);
```
olivier-le-sage, line 76: I'm not sure I understand this one. Here we know there is an BLE event upcoming, potentially very soon -- isn't this latency way to high?

tests/subsys/mpsl/pm/pm_test.c
```diff
@@ -0,0 +1,314 @@
...
+
+#include "cmock_policy.h"
+#include "cmock_mpsl_pm.h"
+#include "cmock_mpsl_pm_config.h"
+#include "cmock_mpsl_work.h"
+
+#include <mpsl/mpsl_pm_utils.h>
+#include <zephyr/kernel.h>
+
+#define NUMELTS(X) (sizeof((X)) / sizeof(X[0]))
```
olivier-le-sage, line 21: would this work? https://docs.zephyrproject.org/latest/kernel/util/index.html#c.ARRAY_SIZE

include/mpsl/mpsl_work.h
hermabe: Or do as in zephyr does with `k_work_submit`, just refer to `k_work_submit_to_queue`. https://github.com/zephyrproject-rtos/zephyr/blob/1cfe90d825a1e07d9aa77bf389338dc72dfda6fe/include/zephyr/kernel.h#L3336-L3344

KyraLengfeld: I would prefer to keep asserting on the return value, can shorten the description like Herman's suggestion. Like this we take responsibility of it here, and can easily handle different return values if needed in the future. 
(If there are strong opinions I don't mind going with other solutions as well though.)

tests/subsys/mpsl/pm/pm_test.c
KyraLengfeld: Will try :)

subsys/mpsl/pm/mpsl_pm_utils.c
KyraLengfeld: This I added while testing, as potentially we have been last here in case of MPSL_PM_EVENT_STATE_IN_EVENT (skipping/not reaching MPSL_PM_EVENT_STATE_NO_EVENTS_LEFT as high prio scheduled the next before we get here). In the case that event_time_rel_us is far into the future, especially >UINT32_MAX in the future, it would be bad for power saving to have zero latency still set.
I see your point though, simplest would be to call this function on condition event_time_rel_us is further than X(let me see if there is a constant defined above that fits).
If you have another suggestion or see a flaw with my explanation, let me know.

bencefr:
I suggest to remove `@trond-snekvik` from the CODEOWNERS without any replacement. I don't want to be a code owner in the SDK and I also don't know if anyone else would want to be.

include/mpsl/mpsl_work.h
KyraLengfeld: I see, then I misunderstood it. Seems then I am overvoted in my weak opinion. Will change it next time I update the PR.

olivier-le-sage: If we find there's a need to implement some kind of error code handling we can always change it later

include/mpsl/mpsl_work.h
```diff
@@ -41,12 +41,35 @@ extern struct k_work_q mpsl_work_q;
...
+ * @param delay the time to wait before submitting the work item.  If @c
+ * K_NO_WAIT and the work is not pending this is equivalent to
+ * mpsl_work_submit().
  */
-static inline int mpsl_work_submit(struct k_work *work)
+static inline void mpsl_work_schedule(struct k_work_delayable *dwork, k_timeout_t delay)
 {
-	return k_work_submit_to_queue(&mpsl_work_q, work);
+	__ASSERT((k_work_schedule_for_queue(&mpsl_work_q, dwork, delay) >= 0),
+		 "k_work_submit_to_queue() failed.");
```
knutel-nordic, line 72:
```suggestion
	__ASSERT((k_work_schedule_for_queue(&mpsl_work_q, dwork, delay) >= 0),
		 "k_work_schedule_for_queue() failed.");
```

tests/subsys/mpsl/pm/pm_test.c
```diff
@@ -0,0 +1,312 @@
...
+			case EVENT_FUNC_UPDATE:
+				__cmock_pm_policy_event_update_Expect(0, v.event_time_us);
+				__cmock_pm_policy_event_update_IgnoreArg_evt();
+				break;
+			case EVENT_FUNC_UNREGISTER:
+				__cmock_pm_policy_event_unregister_ExpectAnyArgs();
+				break;
+			case EVENT_FUNC_DELAY_SCHEDULING:
+				__cmock_mpsl_work_schedule_Expect(0,
+									   K_USEC(v.event_time_us));
```
knutel-nordic, line 95:
```suggestion
				__cmock_mpsl_work_schedule_Expect(0, K_USEC(v.event_time_us));
```

include/mpsl/mpsl_work.h
KyraLengfeld: Will change.
FYI took it from description of [k_work_schedule_for_queue](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/kernel.h#L3630), will look if there it may have been copied and pasted wrong as well.

include/mpsl/mpsl_work.h
```diff
@@ -41,12 +41,35 @@ extern struct k_work_q mpsl_work_q;
...
+ * @param delay the time to wait before submitting the work item.  If @c
+ * K_NO_WAIT and the work is not pending this is equivalent to
+ * mpsl_work_submit().
  */
-static inline int mpsl_work_submit(struct k_work *work)
+static inline void mpsl_work_schedule(struct k_work_delayable *dwork, k_timeout_t delay)
 {
-	return k_work_submit_to_queue(&mpsl_work_q, work);
+	__ASSERT((k_work_schedule_for_queue(&mpsl_work_q, dwork, delay) >= 0),
+		 "k_work_schedule_for_queue() failed.");
```
knutel-nordic, line 72: This depends on `CONFIG_ASSERT=y` to work. Please move assert like this, or use macro that does not get compiled away.
```suggestion
	if (k_work_schedule_for_queue(&mpsl_work_q, dwork, delay) < 0) {
		__ASSERT(false, "k_work_schedule_for_queue() failed.");
	}
```
Or replace with

subsys/mpsl/pm/Kconfig
```diff
@@ -0,0 +1,13 @@
+#
```
hermabe, line 1: This file isnt sourced anywhere. Something like `rsource "pm/Kconfig"` should be added to `subsys/mpsl/Kconfig`.

subsys/mpsl/pm/Kconfig
```diff
@@ -0,0 +1,13 @@
...
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config MPSL_USE_ZEPHYR_PM
+	bool "Use Zephyr's Management API"
```
knutel-nordic, line 8:
```suggestion
	bool "Use Zephyr's Power Management API"
```

subsys/mpsl/pm/mpsl_pm_utils.c
```diff
@@ -0,0 +1,121 @@
...
+#include <zephyr/logging/log.h>
+
+#include <mpsl/mpsl_work.h>
+#include <mpsl/mpsl_pm_utils.h>
+
+LOG_MODULE_REGISTER(mpsl_pm_utils, CONFIG_MPSL_LOG_LEVEL);
+
+/* These values are still arbritrary*/
+#define MAX_DELAY_SINCE_READING_PARAMS 50
+#define TIME_TO_REGISTER_EVENT_IN_ZEPHYR 1000
```
knutel-nordic, line 20:
```suggestion
/* These constants must be updated once the Zephyr PM Policy API is updated
 * to handle low latency events. Ideally, the Policy API should be changed to use
 * absolute time instead of relative time. This would remove the need for safety
 * margins and allow optimal power savings. */
#define MAX_DELAY_SINCE_READING_PARAMS_US 50
#define TIME_TO_REGISTER_EVENT_IN_ZEPHYR_US 1000
```

subsys/mpsl/pm/mpsl_pm_utils.c
```diff
@@ -0,0 +1,121 @@
...
+		uint64_t event_time_us = params.event_time_rel_us -
+						MAX_DELAY_SINCE_READING_PARAMS;
+
+		/* In case we missed a state and are in zero-latency, set low-latency.*/
+		m_update_latency_request(PM_MAX_LATENCY_HCI_COMMANDS_US);
+
+		if (event_time_us > UINT32_MAX) {
+			uint32_t retry_time_us = RETRY_TIME_MAX_US;
+
+			mpsl_work_schedule(&pm_work, K_USEC(retry_time_us));
```
knutel-nordic, line 81:
```suggestion
			mpsl_work_schedule(&pm_work, K_USEC(RETRY_TIME_MAX_US));
```

subsys/mpsl/pm/mpsl_pm_utils.c
```diff
@@ -0,0 +1,121 @@
...
+			return;
+		}
+
+		if (m_pm_event_is_registered) {
+			pm_policy_event_update(&m_evt, event_time_us);
+		} else {
+			pm_policy_event_register(&m_evt, event_time_us);
+		}
+		m_pm_event_is_registered = true;
```
knutel-nordic, line 90:
```suggestion
			m_pm_event_is_registered = true;
		}
```

tests/subsys/mpsl/pm/pm_test.c
```diff
@@ -0,0 +1,311 @@
...
+#include "cmock_mpsl_work.h"
+
+#include <mpsl/mpsl_pm_utils.h>
+#include <zephyr/kernel.h>
+
+
+#define PM_MAX_LATENCY_HCI_COMMANDS_US 4999999
+#define MAX_DELAY_SINCE_READING_PARAMS 50
+
+#define TIME_TO_REGISTER_EVENT_IN_ZEPHYR 1000
```
knutel-nordic, line 25:
```suggestion
#define MAX_DELAY_SINCE_READING_PARAMS_US 50

#define TIME_TO_REGISTER_EVENT_IN_ZEPHYR_US 1000
```

katgiadla (approved):
LGTM

knutel-nordic (approved):

PR #14238, opened 2024-02-27
Bluetooth: Mesh: Allow multiple model appkey bindings

omkar3141:
NLC profile requirements mandate the minimum value of "3" for BT_MESH_APP_KEY_COUNT setting. However, these multiple appkeys cannot be used with single model because corresponding BT_MESH_MODEL_KEY_COUNT config setting is still set to "1".

This is quite limiting from use-case perspective. As multiple appkeys are supported on a node, there is no reason why it should not support multiple appkey bindings per model out-of-the-box for NLC samples.


PavelVPV (approved):

Andrewpini (approved):

ludvigsj (changes requested):
This is not a formal requirement in any NLC profile. If we need a more sensible default for this option in general, that should be done in Zephyr, but it does not belong inside this NLC option since it's not a NLC-specific requirement.

omkar3141:
This is a downstream update, driven by feedback from field. By adding option in Zephyr we expand the scope for this default setting where nobody is asking for this. 

I would strongly recommend to have this, the alternative would be to add this option manually in every single sample and set it to "3".

ludvigsj:
> This is a downstream update, driven by feedback from field. By adding option in Zephyr we expand the scope for this default setting where nobody is asking for this.
> 
> I would strongly recommend to have this, the alternative would be to add this option manually in every single sample and set it to "3".

I like the alternative of updating the samples much better. The sample at least has a well defined use-case. This is also where we would have made the change for any other sample which is not implementing any NLC profiles.

samples/bluetooth/mesh/light_ctrl/prj.conf
```diff
@@ -52,6 +52,7 @@ CONFIG_BT_MESH_DK_PROV=y
 CONFIG_BT_MESH_NLC_PERF_CONF=y
 CONFIG_BT_MESH_NLC_PERF_LIGHTNESS_CTRL=y
 CONFIG_BT_MESH_MODEL_EXTENSION_LIST_SIZE=18
+CONFIG_BT_MESH_MODEL_KEY_COUNT=3
```
PavelVPV, line 55: Add a note why it is increased?

ludvigsj:
`sensor_client` also needs this change, it also sets up 3 app keys

samples/bluetooth/mesh/light_ctrl/prj.conf
omkar3141: Done.

samples/bluetooth/mesh/light_ctrl/prj.conf
```diff
@@ -52,6 +52,9 @@ CONFIG_BT_MESH_DK_PROV=y
 CONFIG_BT_MESH_NLC_PERF_CONF=y
 CONFIG_BT_MESH_NLC_PERF_LIGHTNESS_CTRL=y
 CONFIG_BT_MESH_MODEL_EXTENSION_LIST_SIZE=18
+# By virtue of BT_MESH_NLC_PERF_CONF defaults, the sample would support
+# 3 appkeys.Therefore, allow upto 3 appkey bindings per model instance.
```
ludvigsj, line 56:
```suggestion
# Enabling BT_MESH_NLC_PERF_CONF enables support for 3 application keys by default.
# Therefore, allow up to 3 application key bindings per model instance.
```

ludvigsj (approved):

ludvigsj (changes requested):
Pressed wrong button, see comment above (and `sensor_client`)

omkar3141:
> Pressed wrong button, see comment above (and `sensor_client`)

Ok, I don't think sensor client would need any additional explanation since `BT_MESH_APP_KEY_COUNT` is added without any explanation anyway. Also, if anyone sees commit message, they would get the reason.

PavelVPV (approved):

ludvigsj (approved):

PR #14240, opened 2024-02-27
samples: wifi: Increase polling sockets

krish2718:
Recent changes to Zperf caused an increase in maximum polling sockets as it now uses socket services (TCP-6, UDP-2) for it's receiver, and due to a but in the socket services, there is a array overflow causing a crash and corruption in BLE Co-ex sample.

Increase the sockets to fix the warning at the boot and the corruption, the proper fix to Zephyr socket services will be done separately.

Fixes SHEL-2530.


krish2718:
Discussion of root cause fix is in progress with @jukkar and will be taken up in a separate PR.

krish2718:
> Discussion of root cause fix is in progress with @jukkar and will be taken up in a separate PR.

Already fixed upstream last week: https://github.com/zephyrproject-rtos/zephyr/pull/69234

bama-nordic (approved):

muraliThokala (approved):

PR #14246, opened 2024-02-28
partition_manager: NRF54L support for MCUboot added

michalek-no:
Changes to Kconfig and CMake that allow MCUboot to work with NRF54L. The Kconfig has been changed to change size of MCUboot header and CMake to properly invoke partition manager script with RRAM as NVM.

### DIFF
--- a/modules/mcuboot/boot/zephyr/Kconfig
+++ b/modules/mcuboot/boot/zephyr/Kconfig
@@ -15,8 +15,14 @@ partition-size=0x1e000
 source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_config"
 
 partition=MCUBOOT_PAD
-partition-size=0x200
+if SOC_PLATFORM_NRF54L
+	partition-size=0x800
 source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_config"
+endif
+if !SOC_PLATFORM_NRF54L
+	partition-size=0x200
+source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_config"
+endif
 
 config PM_PARTITION_SIZE_MCUBOOT
 	hex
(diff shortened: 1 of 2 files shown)


modules/mcuboot/boot/zephyr/Kconfig
```diff
@@ -15,8 +15,14 @@ partition-size=0x1e000
 source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_config"
 
 partition=MCUBOOT_PAD
-partition-size=0x200
+if SOC_PLATFORM_NRF54L
+	partition-size=0x800
```
de-nordic, line 19: Note for the future: this could probably be taken from the `CONFIG_ROM_START_OFFSET`

nordicjm, line 19: why 0x800?

nvlsianpu: nRF54L15 has so big vector table.

nvlsianpu (approved):

modules/mcuboot/boot/zephyr/Kconfig
nordicjm: 0x800 seems a bit many

de-nordic: image
Gives 271 IRQS.
With
image
so there is mandatory 32 words (4 bytes each that cover the base 16 interrupts), this takes 128 bytes already.
271-16 = 255 additional interrupt vectors, one word each is 4 * 255 = 1020, so 1148 bytes total.
This rounded up to next power of two is 2048, which is 0x800.
Am I correct here?

de-nordic (approved):

modules/mcuboot/boot/zephyr/Kconfig
nvlsianpu: Ref.: https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/linker_script/arm/linker.cmake#L153
Vector alignment is 0x800.

SeppoTakalo (approved):

PR #14248, opened 2024-02-28
doc: update bluetooth features support

yuxCai:
Adding LE power control, periodic advertising with responses, CIS central/peripheral, and BIS broadcaster/receiver to the bluetooth features support table

### DIFF
--- a/doc/_scripts/software_maturity/software_maturity_features.yaml
+++ b/doc/_scripts/software_maturity/software_maturity_features.yaml
@@ -22,10 +22,27 @@ features:
     Sidewalk over LORA: SIDEWALK && SIDEWALK_LINK_MASK_LORA
     Sidewalk - OTA DFU over Bluetooth LE: SIDEWALK && SIDEWALK_DFU_SERVICE_BLE
   bluetooth:
-    Bluetooth LE Peripheral/Central: BT_CTLR && (BT_CENTRAL || BT_PERIPHERAL)
-    LLPM: BT_CTLR_SDC_LLPM
-    LE Coded PHY: BT_CTLR_PHY_CODED_SUPPORT
-    Connectionless/Connected CTE Transmitter: BT_CTLR_DF_CTE_TX_SUPPORT
+    LE 2M PHY: BT_CTLR_PHY_2M_SUPPORT
+    Peripheral: BT_CTLR && BT_PERIPHERAL
+    Central: BT_CTLR && BT_CENTRAL
+    Data Length Extensions: BT_CTLR && BT_DATA_LEN_UPDATE
+    Advertising Extensions: BT_CTLR_ADV_EXT
+    Periodic Advertising: BT_CTLR_ADV_PERIODIC
+    Connectionless CTE Advertising: CONFIG_BT_CTLR_DF_CTE_TX
+    Connection CTE Response: CONFIG_BT_CTLR_DF_CTE_TX
+    LE Coded PHY: BT_CTLR_PHY_CODED
+    LE Power Control: BT_CTLR_LE_POWER_CONTROL
+    Periodic Advertising Sync Transfer - Sender: BT_CTLR_SYNC_TRANSFER_SENDER
+    Periodic Advertising Sync Transfer - Receiver: BT_CTLR_SYNC_TRANSFER_RECEIVER
+    Sleep Clock Accuracy Updates: BT_CTLR_SCA_UPDATE
+    Periodic Advertising with Responses - Advertiser: BT_CTLR_SDC_PAWR_ADV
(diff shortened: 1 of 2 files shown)


doc/nrf/releases_and_maturity/software_maturity.rst
```diff
@@ -199,10 +199,10 @@ The following table indicates the software maturity levels of the support for ea
 
   .. sml-table:: sidewalk
 
-Bluetooth features support
+Bluetooth Low Energy features support
 **************************
```
wiba-nordic, line 203:
```suggestion
************************************
```
Should be the same character length as the header text itself.

doc/nrf/releases_and_maturity/software_maturity.rst
```diff
@@ -199,10 +199,10 @@ The following table indicates the software maturity levels of the support for ea
...
 
-The following table indicates the software maturity levels of the support for each Bluetooth feature:
+The following table indicates the software maturity levels of the support for each Bluetooth Low Energy feature with * :ref:`softdevice_controller`:
```
wiba-nordic, line 205:
```suggestion
The following table indicates the software maturity levels of the support for each Bluetooth Low Energy feature.
See the :ref:`softdevice_controller` documentation for additional information.
```

wiba-nordic (changes requested):

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -22,10 +22,27 @@ features:
...
     Sidewalk - OTA DFU over Bluetooth LE: SIDEWALK && SIDEWALK_DFU_SERVICE_BLE
   bluetooth:
-    Bluetooth LE Peripheral/Central: BT_CTLR && (BT_CENTRAL || BT_PERIPHERAL)
-    LLPM: BT_CTLR_SDC_LLPM
+    LE 2M PHY: BT_CTLR_PHY_2M_SUPPORT
+    Peripheral: BT_CTLR && BT_PERIPHERAL
+    Central: BT_CTLR && BT_CENTRAL
+    Data Length Extensions: BT_CTLR_DATA_LEN_UPDATE_SUPPORT
+    Advertising Extensions: BT_CTLR_ADV_EXT_SUPPORT
+    Periodic Advertising: BT_CTLR_ADV_PERIODIC_SUPPORT
```
rugeGerritsen, line 30: I would assume we need to use `BT_CTLR_ADV_PERIODIC` instead of `BT_CTLR_ADV_PERIODIC_SUPPORT`. The latter is only a dummy symbol used to check if the feature can be enabled at all

yuxCai: should I remove the `SUPPORT` suffix for other features as well?

rugeGerritsen: I would assume so. I would recommend having a look at how they are defined here: https://github.com/nrfconnect/sdk-zephyr/blob/main/subsys/bluetooth/controller/Kconfig

wiba-nordic (approved):

jori-nordic (approved):

alwa-nordic (changes requested):
I can't see the new data in the [rendered output](https://developer.nordicsemi.com/nRF_Connect_SDK_dev/doc/PR-14248/nrf/releases_and_maturity/software_maturity.html#bluetooth-low-energy-features-support). Is it supposed to visible there?

Also, the support for PAwR in the host is not mature. At least some of the host kconfigs are `EXPERIMENTAL`. The maturity status should describe the whole system, not just the SDC. I think that's what our customers would want. @rugeGerritsen agrees.

There are at least some known issues with PAwR and Privacy.
    https://github.com/zephyrproject-rtos/zephyr/issues/59653
    https://github.com/zephyrproject-rtos/zephyr/issues/59364

yuxCai:
@alwa-nordic I was told that the changes in the table will be visible only after it is merged because of AWS involvement. I can do a follow up if it doesn't work

About the maturity levels - is there a list of features for the host? We can put the link here and do some manual changes to the table as well

alwa-nordic:
> About the maturity levels - is there a list of features for the host? We can put the link here and do some manual changes to the table as well

In the rendered output, I see some table cells say 'experimental'. I'm guessing the documentation script looks at the supplied kconfigs in the yaml to determine this. So if we just `&&`-add the host kconfigs to the features, the script should just figure it out.

Or do you mean that this table represents only the controller support and we should have a separate table for the host?

Upstream, we have [the Bluetooth overview document](https://github.com/zephyrproject-rtos/zephyr/blob/main/doc/connectivity/bluetooth/overview.rst). But this is manually updated, and I would not trust it as a data source.

PR #14208, opened 2024-02-26
Bring IPC support for nRF54H20

gmarull:

PR #14250, opened 2024-02-28
manifest: sdk-zephyr: ensure the sh is valid before call shell_printf

kapbh:
It is possible that the `sh` was not set before use. This change adds a NULL check for `sh` in the following macros: PR, PR_SHELL, PR_ERROR, PR_INFO, and PR_WARNING.


jfischer-no (changes requested):
Please do not update sdk-zephyr revision here.

kapbh:
PR is merged and manifest is updated in https://github.com/nrfconnect/sdk-nrf/pull/14208.

PR #14252, opened 2024-02-28
modules/mcuboot: Fix incorrect size used for signing external image

de-nordic:
The image signing has been taking, as --slot-size parameter, size of primary internal slot rather than expected external slot it has been signing image for.


nvlsianpu (approved):
Awesome!
Thank you @de-nordic !

michalek-no (approved):

PR #14259, opened 2024-02-28
applications: nrf_desktop: Update event propagation arrays

MarekPieta:
Change updates event propagation arrays in documentation.

Jira: NCSDK-26286


annwoj (approved):

zycz (approved):
The table looks bad in the web browser - it should expand more if we expand window. Right now all the information is placed tighly.

annwoj:
> The table looks bad in the web browser - it should expand more if we expand window. Right now all the information is placed tighly.

I think the issue is caused by the contents menu. Decreasing the depth to ``1`` might solve it.
The other solution could be to force blank lines by adding `|` (though that won't look great in the source), or simply to expand the first section by adding some information to it. 
However, please note that the same table render is present in the Bluetooth LE bond module, HID forward module, Wheel module and Motion module, so as you can see it's not uncommon. 
The question here is if this should be addressed in any way, because if it is, it should probably be done collectively. 
Please also note, that this does not occur on our new docs platform (that is our target doc platform): https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/applications/nrf_desktop/doc/ble_adv.html . From this perspective we should probably avoid non-standard rst solutions that could affect the text or structure.

zycz:
Yes, my comment was general and it should be done collectively. I just raise a small issue that can be done so it looks better :)

PR #14260, opened 2024-02-28
Promiscuous mode shell command

VivekUppunda:
This set of changes adds a promiscuous mode shell command to shell sample to configure promiscuous mode for the nRF Wi-Fi driver.

### DIFF
--- a/samples/wifi/shell/src/wifi_promiscuous_shell.c
+++ b/samples/wifi/shell/src/wifi_promiscuous_shell.c
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/** @file
+ * @brief Wi-Fi Promiscuous mode shell
+ */
+
+#include <stdlib.h>
+#include <zephyr/shell/shell.h>
+#include <zephyr/posix/unistd.h>
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(promiscuous_mode, CONFIG_LOG_DEFAULT_LEVEL);
+
+#include "net_private.h"
+
+static int cmd_wifi_promisc(const struct shell *sh, size_t argc, char *argv[])
+{
+	struct net_if *iface = NULL;
+	bool mode_val = 0;
+
+	if (argc != 2) {
+		shell_help(sh);
+		return -ENOEXEC;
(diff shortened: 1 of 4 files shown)


samples/wifi/shell/src/wifi_promiscuous_shell.c
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
```
krish2718, line 2: 2024

samples/wifi/shell/src/wifi_promiscuous_shell.c
```diff
@@ -0,0 +1,67 @@
...
+		LOG_ERR("Promiscuous mode %s failed", mode_val ? "enabling" : "disabling");
+	}
+	return 0;
+}
+
+SHELL_STATIC_SUBCMD_SET_CREATE(
+	promisc_cmd,
+	SHELL_CMD_ARG(mode, NULL,
+		      "This command may be used to enable or disable promiscuous mode\n"
+		      "[enable: 1, disable: 0]\n"
```
krish2718, line 56: what about taking the interface index as an arg?

krish2718 (commented):
shell code and doc can be in a single commit itself.

samples/wifi/shell/src/wifi_promiscuous_shell.c
VivekUppunda: currently - net_eth_promisc_mode API does not have an ability to configure interface index to the driver. Once that support comes in, we can add the same in the shell sample. It is basically defined as - 
`int net_eth_promisc_mode(struct net_if *iface, bool enable)`

krish2718: You don't need to pass iface index, you can get iface handle from index and pass it: `iface = net_if_get_by_index`

krish2718 (approved):

samples/wifi/shell/src/wifi_promiscuous_shell.c
VivekUppunda: Thanks Chaitanya for that input. In the interest of time, I will take that up as a new PR task to update the code.

VivekUppunda:
> shell code and doc can be in a single commit itself.

Merged the same

richabp (approved):

rlubos:
CLA seems just stuck

PR #14261, opened 2024-02-28
Bluetooth: Mesh: add model metadata

alxelax:
PR adds metadata to:
            - Light Lightness Server
            - Light CTL Temperature Server
            - Light HSL Server
            - Sensor server
            - Time Server
as well as health and lightness server metadata usage in light_ctrl sample

### DIFF
--- a/include/bluetooth/mesh/lightness_srv.h
+++ b/include/bluetooth/mesh/lightness_srv.h
@@ -65,6 +65,53 @@ struct bt_mesh_lightness_srv;
 						 _srv),                        \
 			 &_bt_mesh_lightness_setup_srv_cb)
 
+#ifdef CONFIG_BT_MESH_LARGE_COMP_DATA_SRV
+/** @def BT_MESH_MODEL_LIGHTNESS_SRV_WITH_METADATA
+ *
+ * @brief Light Lightness model entry with metadata.
+ *
+ * @param[in] _srv Pointer to a @ref bt_mesh_lightness_srv instance.
+ * @param[in] _metadata Pointer to a @ref bt_mesh_models_metadata_entry.
+ */
+#define BT_MESH_MODEL_LIGHTNESS_SRV_WITH_METADATA(_srv, _metadata)                                 \
+	BT_MESH_MODEL_LVL_SRV(&(_srv)->lvl),                                                       \
+	BT_MESH_MODEL_PONOFF_SRV(&(_srv)->ponoff),                                                 \
+	BT_MESH_MODEL_METADATA_CB(                                                                 \
+			BT_MESH_MODEL_ID_LIGHT_LIGHTNESS_SRV, _bt_mesh_lightness_srv_op,           \
+			&(_srv)->pub, BT_MESH_MODEL_USER_DATA(struct bt_mesh_lightness_srv, _srv), \
+			&_bt_mesh_lightness_srv_cb, &_metadata),                                   \
+	BT_MESH_MODEL_CB(                                                                          \
+			BT_MESH_MODEL_ID_LIGHT_LIGHTNESS_SETUP_SRV,                                \
+			_bt_mesh_lightness_setup_srv_op, NULL,                                     \
+			BT_MESH_MODEL_USER_DATA(struct bt_mesh_lightness_srv, _srv),               \
+			&_bt_mesh_lightness_setup_srv_cb)
+#endif
+
+/** The Light Purpose Metadata ID. */
+#define BT_MESH_LIGHT_PURPOSE_METADATA_ID 0x0002
+
+/** Light Lightness Range Metadata ID. */
+#define BT_MESH_LIGHT_LIGHTNESS_RANGE_METADATA_ID 0x0003
+
+/**
+ *  Define the Light Purpose Metadata entry.
+ *
+ *  @param light_purpose 16-bit light purpose value.
+ */
+#define BT_MESH_LIGHT_PURPOSE_METADATA(light_purpose)                                              \
+	BT_MESH_MODELS_METADATA_ENTRY(2, BT_MESH_LIGHT_PURPOSE_METADATA_ID,                        \
+				      ((uint16_t[]){(light_purpose)}))
+
+/**
+ *  Light Lightness Range Metadata entry.
+ *
+ *  @param range_min Minimum boundary of range (16-bit)
+ *  @param range_max Maximum boundary of range (16-bit)
+ */
+#define BT_MESH_LIGHT_LIGHTNESS_RANGE_METADATA(range_min, range_max)                               \
+	BT_MESH_MODELS_METADATA_ENTRY(4, BT_MESH_LIGHT_LIGHTNESS_RANGE_METADATA_ID,                \
+				      ((uint16_t[]){(range_min), (range_max)}))
+
 /** Collection of handler callbacks for the Light Lightness Server. */
 struct bt_mesh_lightness_srv_handlers {
 	/** @brief Set the Light state.
--- a/include/bluetooth/mesh/sensor_srv.h
+++ b/include/bluetooth/mesh/sensor_srv.h
@@ -58,6 +58,38 @@ struct bt_mesh_sensor_srv;
 					      _srv),                           \
 			 &_bt_mesh_sensor_setup_srv_cb)
 
+#ifdef CONFIG_BT_MESH_LARGE_COMP_DATA_SRV
+/** @def BT_MESH_MODEL_SENSOR_SRV_WITH_METADATA
+ *
+ *  @brief Sensor Server model composition data entry with metadata.
+ *
+ *  @param[in] _srv Pointer to a @ref bt_mesh_sensor_srv instance.
+ *  @param[in] _metadata Pointer to a @ref bt_mesh_models_metadata_entry.
+ */
+#define BT_MESH_MODEL_SENSOR_SRV_WITH_METADATA(_srv, _metadata)                                    \
+	BT_MESH_MODEL_METADATA_CB(BT_MESH_MODEL_ID_SENSOR_SRV, _bt_mesh_sensor_srv_op,             \
+				  &(_srv)->pub,                                                    \
+				  BT_MESH_MODEL_USER_DATA(struct bt_mesh_sensor_srv, _srv),        \
+				  &_bt_mesh_sensor_srv_cb, &_metadata),                     \
+	BT_MESH_MODEL_CB(BT_MESH_MODEL_ID_SENSOR_SETUP_SRV, _bt_mesh_sensor_setup_srv_op,          \
+			 &(_srv)->setup_pub,                                                       \
+			 BT_MESH_MODEL_USER_DATA(struct bt_mesh_sensor_srv, _srv),                 \
+			 &_bt_mesh_sensor_setup_srv_cb)
+#endif
+
+/** Sensor Properties Metadata ID. */
+#define BT_MESH_SENSOR_PROP_METADATA_ID 0x0001
+
+/**
+ *  Sensor Properties Metadata entry.
+ *
+ *  @param[in] ... list of 16-bit sensor property IDs split by comma.
+ */
+#define BT_MESH_SENSOR_PROP_METADATA(...)                                                          \
+	BT_MESH_MODELS_METADATA_ENTRY(2 * ARRAY_SIZE(((uint16_t[]){__VA_ARGS__})),                 \
+				      BT_MESH_SENSOR_PROP_METADATA_ID,                             \
+				      ((uint16_t[]){__VA_ARGS__}))
+
 /** Sensor server instance. */
 struct bt_mesh_sensor_srv {
 	/** Sensors owned by this server. */
--- a/samples/bluetooth/mesh/light_ctrl/src/model_handler.c
+++ b/samples/bluetooth/mesh/light_ctrl/src/model_handler.c
@@ -11,6 +11,7 @@
 #include "lc_pwm_led.h"
 
 #define PWM_SIZE_STEP 512
+#define BT_MESH_LIGHT_PURPOSE_INDICATOR_LIGHT 0x0013
 
 struct lightness_ctx {
 	struct bt_mesh_lightness_srv lightness_srv;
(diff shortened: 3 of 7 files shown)


ludvigsj (changes requested):

PavelVPV (commented):
I wonder why do we need extra pointer in each server struct to store metadata? Why not to extend `BT_MESH_MODEL_<MODEL_NAME>_SRV` macros to take metadata as extra argument and pass it straight to `struct bt_mesh_model`?

alxelax:
> I wonder why do we need extra pointer in each server struct to store metadata? Why not to extend `BT_MESH_MODEL_<MODEL_NAME>_SRV` macros to take metadata as extra argument and pass it straight to `struct bt_mesh_model`?

@PavelVPV, I didn't understand what you meant. Could you clarify or provide even a sample of what you mean.

PavelVPV:
> > I wonder why do we need extra pointer in each server struct to store metadata? Why not to extend `BT_MESH_MODEL_<MODEL_NAME>_SRV` macros to take metadata as extra argument and pass it straight to `struct bt_mesh_model`?
> 
> @PavelVPV, I didn't understand what you meant. Could you clarify or provide even a sample of what you mean?

https://github.com/PavelVPV/sdk-nrf/commit/e3c67195c0884dc92b293d21e5c166b0bf749197

PavelVPV:
> Not each server has extra metadata pointer. The only limited list of servers that I pointed in PR comment + Health server.

I should have said that I gave Gen OnOff Server as an example as it was easier to show the suggestion using it. **Of course**, it doesn't support metadata.

> The one more point is that the model metadata is optional functionality and exist only if LCD server is present.
It causes the situation we have to have 2 API versions with and without metadata or mandate customers to pass metadata even it is not required.

Same you have with the suggestion you use. But, please have a look how it is solved in upstream:
https://github.com/zephyrproject-rtos/zephyr/blob/99707246527ab2df9212a29f01fbef87c6cce44c/include/zephyr/bluetooth/mesh/access.h#L511-L529

> I'd suggest to use macro with VA_ARG list and give customer to provide during initialization whatever he\she wants.

This has a risk if we add more parameters that need to be initialized at the model definition.

> I'd like to refuse to start refactoring in this task and my PR. If you want to suggest some improvements, please do it as a separate task.

Sure, lets bring this to the daily meeting. So far I don't see point doing it the way it is implemented in this PR.

PavelVPV (changes requested):
Needs change.. Explained above.

PavelVPV:
Ok, I see why my suggestion won't work. In the `_SRV` macros we also instantiate base models, which may also require metadata. Taking metadata pointers for all instantiated models through `_SRV` macro will of course be a horrible solution. Therefore, `_INIT` macros with extra pointer to metadata entries is the only possible way to initialise metadata in model's struct. Please disregard my suggestion.

PavelVPV (approved):

PavelVPV:
> Ok, I see why my suggestion won't work. In the `_SRV` macros we also instantiate base models, which may also require metadata. Taking metadata pointers for all instantiated models through `_SRV` macro will of course be a horrible solution. Therefore, `_INIT` macros with extra pointer to metadata entries is the only possible way to initialise metadata in model's struct. Please disregard my suggestion.

Actually, no, it is the same for both implementation. And `BT_MESH_LIGHT_CTL_SRV_INIT_WITH_METADATA` should also take metadata for the lightness server model. So my point is still valid.

alxelax:
WIP

alxelax:
> WIP

@PavelVPV, @ludvigsj PR is ready for review

include/bluetooth/mesh/lightness_srv.h
```diff
@@ -65,6 +65,52 @@ struct bt_mesh_lightness_srv;
...
+/** Light Lightness Range Metadata ID. */
+#define BT_MESH_LIGHT_LIGHTNESS_RANGE_METADATA_ID 0x0003
+
+/**
+ *  Define the Light Purpose Metadata entry.
+ *
+ *  @param light_purpose 16-bit light purpose value.
+ */
+#define BT_MESH_LIGHT_PURPOSE_METADATA(light_purpose)   \
+	BT_MESH_MODELS_METADATA_ENTRY(2, BT_MESH_LIGHT_PURPOSE_METADATA_ID, &light_purpose)
```
ludvigsj, line 102: Is there a reason while this passes a pointer to an existing variable, while the `RANGE` macro allocates a new compound literal? I feel like these macros should be design around one or the other paradigm, currently it looks weird that `BT_MESH_LIGHT_LIGHTNESS_RANGE_METADATA(0, 1000)` works while `BT_MESH_LIGHT_PURPOSE_METADATA(0x0012)` does not.

include/bluetooth/mesh/sensor_srv.h
```diff
@@ -58,6 +58,38 @@ struct bt_mesh_sensor_srv;
...
+ *  @brief Sensor Server model composition data entry with metadata.
+ *
+ *  @param[in] _srv Pointer to a @ref bt_mesh_sensor_srv instance.
+ *  @param[in] _metadata Pointer to a @ref bt_mesh_models_metadata_entry.
+ */
+#define BT_MESH_MODEL_SENSOR_SRV_WITH_METADATA(_srv, _metadata)                                    \
+	BT_MESH_MODEL_METADATA_CB(BT_MESH_MODEL_ID_SENSOR_SRV, _bt_mesh_sensor_srv_op,             \
+				  &(_srv)->pub,                                                    \
+				  BT_MESH_MODEL_USER_DATA(struct bt_mesh_sensor_srv, _srv),        \
+				  &_bt_mesh_sensor_srv_cb, &_metadata),                     \
```
PavelVPV, line 73: Parenthesis everywhere and remove `&` so it gives more flexibility for a user:
```suggestion
				  &_bt_mesh_sensor_srv_cb, (_metadata)),                     \
```

include/bluetooth/mesh/sensor_srv.h
```diff
@@ -58,6 +58,38 @@ struct bt_mesh_sensor_srv;
...
+/** Sensor Properties Metadata ID. */
+#define BT_MESH_SENSOR_PROP_METADATA_ID 0x0001
+
+/**
+ *  Sensor Properties Metadata entry.
+ *
+ *  @param[in] ... list of 16-bit sensor property IDs split by comma.
+ */
+#define BT_MESH_SENSOR_PROP_METADATA(...)                                                          \
+	BT_MESH_MODELS_METADATA_ENTRY(2 * ARRAY_SIZE(((uint16_t[]){__VA_ARGS__})),                 \
```
PavelVPV, line 89: Could probably be simply `sizeof` to reduce mathematics:
```suggestion
	BT_MESH_MODELS_METADATA_ENTRY(sizeof(((uint16_t[]){__VA_ARGS__})),                 \
```

samples/bluetooth/mesh/light_ctrl/src/model_handler.c
```diff
@@ -11,6 +11,7 @@
 #include "lc_pwm_led.h"
 
 #define PWM_SIZE_STEP 512
+#define BT_MESH_LIGHT_PURPOSE_INDICATOR_LIGHT 0x0013
```
PavelVPV, line 14: Shouldn't this be in `lightness.h` together with other Light Purpose metadata?

PavelVPV (changes requested):
I think the lightness srv metadata should be possible to set through light ctl srv model macro instantiation:
https://github.com/nrfconnect/sdk-nrf/blob/80d8123f958f45f43d47996421d7e5c4c567f4d7/include/bluetooth/mesh/light_ctl_srv.h#L48-L49

Should we provide macros for encoding clock accuracy, timekeeping, etc? I think so far our approach was to provide such macros.

Please add release notes.

include/bluetooth/mesh/sensor_srv.h
alxelax: The design of access layer metadata handler implementation assumed the metadata pointer will be stored somewhere in the model. Like it has been done in Health server: https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/bluetooth/mesh/health_srv.h#L160-L163
Since the same idea for NCS models was disagreed, then it requires providing pointer to pointer directly to access layer.
I do not think this is convenient for user to provide pointer to pointer directly in macros. It looks cumbersome.
Even now, it is difficult to explain why user needs to have both metadata and pointer to it. 
Nevertheless, the current implementation is more convenient if we rely on customer responsibility of managing both metadata and its pointer.

include/bluetooth/mesh/sensor_srv.h
alxelax: this is preprocessor mathematic that doesn't impact on anything but Ok

include/bluetooth/mesh/sensor_srv.h
PavelVPV: The problem really in the double pointer. It should've been single pointer, then it would work like for opcode handlers. But now we need to have the extra pointer for unknown reasons.

I still lean towads removing `&` as the customer could do:
```
static struct bt_mesh_models_metadata_entry **entries = {
        &(struct bt_mesh_models_metadata_entry){...},
        &(struct bt_mesh_models_metadata_entry){...},
        ...
};
```

alxelax: I guess your example is against of Mesh specification. 
```
4.2.50.1 
Models Metadata Page 0 

Composition Metadata Page 0 shall not change during a term of a node on the network
```
Technically, your suggestion is possible but doesn't have sense.
I do not see any reason to have an implementation that breaks specification.

PavelVPV: Though I don't see contradiction in my suggestion with the spec, I will address only one point in your comment: you can change metadata between terms. For example, you can change the light purpose while device is unprovisioned.

alxelax: This is obvious. Why does the model require an array of metadata entries if specification mandates to use only one entry?
For changing of light purpose, you do not need to provide metadata that is not used. Just provide actual entry. That's it.

PavelVPV: I think the argue went away from the initial comment. I'm tired arguing here. I gave my reasons. Please remove `&`. Thanks.

alxelax: I provided my counterarguments against it. The reason of changes `because I want` doesn't seem correct to me. Thanks

PavelVPV: It is not `I want` it is about more flexibility. Please read my comment correctly.

alxelax: Yep, the same from me. I provided my counterarguments against of it. Those were ignored. Please read them.

PavelVPV: You get the same code as before, just pass `p_lightness_srv_metadata` with `&`. But then you also able to instantiate metadata entries differently (I gave example). I don't think we should limit the customers in how they should instantiate metadata.

PavelVPV: > Customer is able to instantiate metadata in any way (even how you provided in example). It will work in both implementations.

With your current changes I can't do this:

```
diff --git a/samples/bluetooth/mesh/light_ctrl/src/model_handler.c b/samples/bluetooth/mesh/light_ctrl/src/model_handler.c
index cbba6743bc2..508da22479c 100644
--- a/samples/bluetooth/mesh/light_ctrl/src/model_handler.c
+++ b/samples/bluetooth/mesh/light_ctrl/src/model_handler.c
@@ -107,6 +107,7 @@ static uint8_t health_tests[] = {
        BT_MESH_HEALTH_TEST_INFO(CONFIG_BT_COMPANY_ID, 3, 0x11, 0x22, 0x33),
 };
 
+#if 0
 static struct bt_mesh_models_metadata_entry lightness_srv_metadata[] = {
        BT_MESH_LIGHT_PURPOSE_METADATA(BT_MESH_LIGHT_PURPOSE_INDICATOR_LIGHT),
        BT_MESH_LIGHT_LIGHTNESS_RANGE_METADATA(0, UINT16_MAX),
@@ -114,6 +115,13 @@ static struct bt_mesh_models_metadata_entry lightness_srv_metadata[] = {
 };
 
 static struct bt_mesh_models_metadata_entry *p_lightness_srv_metadata = lightness_srv_metadata;
+#else
+static struct bt_mesh_models_metadata_entry *p_lightness_srv_metadata[] = {
+       &(struct bt_mesh_models_metadata_entry)BT_MESH_LIGHT_PURPOSE_METADATA(BT_MESH_LIGHT_PURPOSE_INDICATOR_LIGHT),
+       &(struct bt_mesh_models_metadata_entry)BT_MESH_LIGHT_LIGHTNESS_RANGE_METADATA(0, UINT16_MAX),
+       &(struct bt_mesh_models_metadata_entry)BT_MESH_MODELS_METADATA_END,
+};
+#endif
 
 static struct bt_mesh_models_metadata_entry health_srv_metadata[] = {
        BT_MESH_HEALTH_TEST_INFO_METADATA(health_tests),
```

```
[1/8] Building C object CMakeFiles/app.dir/src/model_handler.c.obj                                                                                                                                 [161/161]
In file included from /home/pavel/nordic/ncs/git-repos/ncs/zephyr/include/zephyr/bluetooth/bluetooth.h:24,                                                                                                  
                 from ../src/model_handler.c:7:                                                                                                                                                             
/home/pavel/nordic/ncs/git-repos/ncs/nrf/include/bluetooth/mesh/lightness_srv.h:82:53: warning: initialization of 'struct bt_mesh_models_metadata_entry **' from incompatible pointer type 'struct bt_mesh_m
odels_metadata_entry * (*)[3]' [-Wincompatible-pointer-types]                                                                                                                                               
   82 |                         &_bt_mesh_lightness_srv_cb, &_metadata),                 
```

> I try to convey idea to you about useless of providing pointer to pointer since the only single entry is used in access....

Just see how `BT_MESH_MODEL_OP_END` macro works:
https://github.com/zephyrproject-rtos/zephyr/blob/35176b993d4e06cc840b8dff61c905ba8437de50/include/zephyr/bluetooth/mesh/access.h#L394-L395
https://github.com/zephyrproject-rtos/zephyr/blob/35176b993d4e06cc840b8dff61c905ba8437de50/samples/bluetooth/mesh/src/main.c#L215-L219

PavelVPV: > With your current changes I can't do this:

With my suggestion I can initialize metadata in both ways.

alxelax:
Seems there is unresolvable disagreement regarding API. Close PR.

PR #14266, opened 2024-02-28
samples: wifi: Fix SML experimental update for nRF7000 and nRF7001

krish2718:
The idea was to skip twister for these variants to save CI time, but still add an entry for SML. But SML in current state needs access to "build.log" and ".config" build artifacts, so, the experimental SML isn't updated properly.

Till we figure out a way to get experimental status without build artifacts, enable these experimental features for boards nRF7000 and nRF7001 in the twister.

### DIFF
--- a/samples/wifi/monitor/sample.yaml
+++ b/samples/wifi/monitor/sample.yaml
@@ -17,8 +17,6 @@ tests:
     platform_allow: nrf5340dk_nrf5340_cpuapp
     tags: ci_build
   sample.nrf7000.monitor:
-    # For SML only
-    skip: true
     build_only: true
     extra_args:
       SHIELD=nrf7002ek_nrf7000
@@ -27,8 +25,6 @@ tests:
     platform_allow: nrf5340dk_nrf5340_cpuapp
     tags: ci_build
   sample.nrf7001.monitor:
-    # For SML only
-    skip: true
     build_only: true
     integration_platforms:
       - nrf7002dk_nrf7001_nrf5340_cpuapp
--- a/samples/wifi/raw_tx_packet/sample.yaml
+++ b/samples/wifi/raw_tx_packet/sample.yaml
@@ -17,8 +17,6 @@ tests:
     platform_allow: nrf5340dk_nrf5340_cpuapp
     tags: ci_build
   sample.nrf7000.raw_tx_packet:
-    # For SML only
-    skip: true
     build_only: true
     extra_args:
       SHIELD=nrf7002ek_nrf7000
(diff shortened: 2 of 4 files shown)


VivekUppunda (approved):

bama-nordic (approved):

sachinthegreen (approved):

PerMac:
I am not sure I understand the need for getting experimental status without building. We do build everything at night, so you can have it from there. Anyway, this PR is good since it removes "skip: true". Such entry shouldn't be used at all in samples. It just messes with the CI an documentation. It tell twister to ALWAYS skip the build sample and makes it untestable in the CI. It is a way how upstream wanted to skip a broken builds to not break CI, but we have a quarantine for this now.

PerMac (approved):

PerMac:
"The idea was to skip twister for these variants to save CI time, but still add an entry for SML." This is a wrong approach. If you skip the build always and entirely, how do you want to assure it even builds? Beside marking what's experimental, the table should give some warranty to the users, that a given configuration is at least buildable

krish2718:
> If you skip the build always and entirely, how do you want to assure it even builds? Beside marking what's experimental, the table should give some warranty to the users, that a given configuration is at least buildable

These are for the cases where the code is same for variants eg., 7002 and 7001, there is not special cases and code is exactly same, so, verifying build for 7002 should suffice but as 7001 is newly releases we want to force it to be `experimental`, so, for those cases to save CI time we don't need to run the twister, we can add support for conveying experimental status directly from `sample.yaml` itself, I had a `WIP` branch that uses `extra_args: CONFIG_EXPERIMENTAL=y` but we still need the `.config`, so, for now parked that approach.

> This way those configuration will be quarantined (hence legally skipped) on PR, but will be still build on nightly (quarantine_integration is only used on PRs)

Yeah, we have used the quarantine approach before, and SML is run `nighlty` and not in `integration`, but I need to check if it works.

PerMac:
>These are for the cases where the code is same for variants eg., 7002 and 7001, there is not special cases and code is exactly same

This is a tricky statement. There must be some differences ;) even in the way how overlay works, maybe you change it in one but not in the other and there is your discrepancy. That's why IMO it is good to build everything on nightly bases, but pick only the most interesting for the PRs

>but I need to check if it works.` 

I will also check, maybe it wasn't turned on...

krish2718:
> > These are for the cases where the code is same for variants eg., 7002 and 7001, there is not special cases and code is exactly same
> 
> This is a tricky statement. There must be some differences ;) even in the way how overlay works, maybe you change it in one but not in the other and there is your discrepancy. That's why IMO it is good to build everything on nightly bases, but pick only the most interesting for the PRs

Not really, there is a HW difference b/w the two chips, but if it builds for 7002, it will build for 7001 (During testing there might be differences though). Hence this was a candidate for `CI time optimization`.

But yeah, `Quarantine nightly` sounds good to me. I will also explore `forcing experimental SoC` without twister a bit more.

PR #14268, opened 2024-02-28
net: lib: nrf_cloud: add support for anchor name list

jayteemo:
Add support for a list of anchor names to be received by the application when a location request response is parsed. 
IRIS-8036

### DIFF
--- a/include/net/nrf_cloud_location.h
+++ b/include/net/nrf_cloud_location.h
@@ -12,6 +12,7 @@
  */
 
 #include <zephyr/kernel.h>
+#include <zephyr/sys/slist.h>
 #include <modem/lte_lc.h>
 #include <net/nrf_cloud.h>
 #include <net/wifi_location_common.h>
@@ -32,6 +33,9 @@ extern "C" {
 /** Minimum number of access points required by nRF Cloud */
 #define NRF_CLOUD_LOCATION_WIFI_AP_CNT_MIN 2
 
+/** Maximum string length of an anchor name */
+#define NRF_CLOUD_LOCATION_ANCHOR_NAME_MAX 32
+
 /** @brief Location request type */
 enum nrf_cloud_location_type {
 	LOCATION_TYPE_SINGLE_CELL,
@@ -40,6 +44,15 @@ enum nrf_cloud_location_type {
 
 	LOCATION_TYPE__INVALID
 };
+
+/** @brief Anchor name list node, to be used with sys_slist_t APIs */
+struct nrf_cloud_anchor_list_node {
+	sys_snode_t node;
+	char name[];
+};
+
+#define NRF_CLOUD_ANCHOR_LIST_BUF_MIN_SZ	(sizeof(struct nrf_cloud_anchor_list_node) + \
+						 NRF_CLOUD_LOCATION_ANCHOR_NAME_MAX + 1)
 
 /** @brief Location request result */
 struct nrf_cloud_location_result {
@@ -55,6 +68,21 @@ struct nrf_cloud_location_result {
 	uint32_t unc;
 	/** Error value received from nRF Cloud. NRF_CLOUD_ERROR_NONE on success. */
 	enum nrf_cloud_error err;
+
+	/** The number of anchors received */
+	uint32_t anchor_cnt;
+
+	/** List of received anchor names contained as @ref nrf_cloud_anchor_list_node.
+	 *  If all anchor names cannot fit in the list, the number of items in the list
+	 *  will be less than @ref anchor_cnt.
+	 */
+	sys_slist_t anchor_list;
+
+	/** User provided buffer to contain the @ref anchor_list */
+	char *anchor_buf;
+
+	/** Size of provided buffer */
+	size_t anchor_buf_sz;
 };
 
 /** @brief Location request config */
@@ -109,6 +137,9 @@ typedef void (*nrf_cloud_location_response_t)(const struct nrf_cloud_location_re
  *                 which is do_reply = true, hi_conf = false, and fallback = true.
  * @param cb Callback function to receive parsed location result. Only used when
  *           config->do_reply is true or config is NULL.
+ *           If @kconfig{CONFIG_NRF_CLOUD_LOCATION_ANCHOR_LIST} is enabled, the application
+ *           should not access the anchor list data after exiting the callback as it may
+ *           become invalid.
  *           If cb is NULL, JSON result will be sent to the cloud event handler as
  *           an NRF_CLOUD_EVT_RX_DATA_LOCATION event.
  * @retval 0       Request sent successfully.
--- a/subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_location
+++ b/subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_location
@@ -8,3 +8,17 @@ config NRF_CLOUD_LOCATION
 	imply FPU
 	depends on NRF_CLOUD_MQTT
 	select CJSON_LIB
+
+config NRF_CLOUD_LOCATION_ANCHOR_LIST
+	bool "Include the list of anchor names when receiving location results from the callback"
+	depends on NRF_CLOUD_LOCATION
+	depends on WIFI
+
+config NRF_CLOUD_LOCATION_ANCHOR_LIST_BUFFER_SIZE
+	int "Size of the buffer to contain the anchor name list"
+	depends on NRF_CLOUD_LOCATION_ANCHOR_LIST
+	default 185
+	help
+	  The maximum length of an anchor name is 32 characters.
+	  Anchor names are returned to the user in a singly-linked list,
+	  which requires 37 bytes for each anchor name on a 32-bit system.
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
@@ -2815,6 +2815,69 @@ static bool json_item_string_exists(const cJSON *const obj, const char *const ke
 	return (strcmp(str_val, val) == 0);
 }
 
+static void nrf_cloud_parse_location_anchors_json(const cJSON * const loc_obj,
+	struct nrf_cloud_location_result * const location_out)
+{
+	cJSON *anchors, *mac, *name;
+	size_t buf_idx = 0;
+	size_t node_sz;
+	size_t name_sz;
+	bool buf_exists = location_out->anchor_buf_sz && location_out->anchor_buf;
+
+	/* Init anchor output */
+	sys_slist_init(&location_out->anchor_list);
+	if (buf_exists) {
+		location_out->anchor_buf[0] = '\0';
+	}
+
+	/* Anchor info is provided in an array of objects */
+	anchors	= cJSON_GetObjectItem(loc_obj, NRF_CLOUD_LOCATION_JSON_KEY_ANCHORS);
+	location_out->anchor_cnt = (uint32_t)cJSON_GetArraySize(anchors);
+
+	for (int idx = 0; idx < location_out->anchor_cnt; ++idx) {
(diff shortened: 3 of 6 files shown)


peknis (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -2880,9 +2882,28 @@ static int nrf_cloud_parse_location_json(const cJSON *const loc_obj,
...
+			/* Copy the anchor array to the user provided buffer */
+			if (!cJSON_PrintPreallocated(name_array, location_out->anchor_buf,
+						     location_out->anchor_buf_sz, false)) {
+				LOG_WRN("Failed to copy %u anchor names, check buffer size",
+					location_out->anchor_cnt);
+				location_out->anchor_buf[0] = '\0';
+			}
+		}
+
+		cJSON_Delete(name_array);
```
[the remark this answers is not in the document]
glarsennordic: Something else I don't fully understand: In, say, MSS, who is responsible for initializing `anchor_buf`? 

Like, for instance, with MQTT, if I understand correctly, location lib calls `nrf_cloud_location_request`.

Then, when the response is received, `cloud_service_nrf_location_ready_cb` is called with a pointer to a `nrf_cloud_location_result` struct which was stack-allocated in directly in `location_process`.

So it doesn't seem like there is an opportunity for the location lib to allocate its response container. But possibly I misunderstand how this all works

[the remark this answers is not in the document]
jayteemo: > Then, when the response is received, `cloud_service_nrf_location_ready_cb` is called with a pointer to a `nrf_cloud_location_result` struct which was stack-allocated in directly in `location_process`.

ah, i did not notice that the location lib uses a separate struct to pass the data to the application.
it looks like a separate location lib PR would be needed.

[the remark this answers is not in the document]
jayteemo: created `IRIS-8530` for location lib update.

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_location
```diff
@@ -8,3 +8,18 @@ config NRF_CLOUD_LOCATION
...
+config NRF_CLOUD_LOCATION_ANCHOR_LIST
+	bool "Include the list of anchor names when receiving location results from the callback"
+	depends on NRF_CLOUD_LOCATION
+	depends on WIFI
+
+config NRF_CLOUD_LOCATION_ANCHOR_LIST_BUFFER_SIZE
+	int "Size of the buffer to contain the anchor name list"
+	depends on NRF_CLOUD_LOCATION_ANCHOR_LIST
+	range 37 702
+	default 142
```
plskeggs, line 21: These are rather specific numbers. I'm curious where they come from.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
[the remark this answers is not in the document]
plskeggs: I think using a Zephyr `sys_slist_t` singly-linked list for the names would be better than JSON, though maybe not that much better that it is worth re-writing this now.

plskeggs (approved):

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_location
jayteemo: the size of a JSON array with one anchor = 37.
`["<32_chars>"]<NULL>`

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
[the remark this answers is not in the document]
jayteemo: Using a sys list would result in the `struct nrf_cloud_location_result` looking something like: 
```
        /** The number of anchors received */
	uint32_t anchor_cnt;
        
        /* List of anchor names */
        sys_slist_t anchor_list;
        
        /** @ref anchor_list buffer */
	char *anchor_buf;
	
	/** Size of provided buffer */
	size_t anchor_buf_sz;
```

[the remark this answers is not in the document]
plskeggs: Yes, of course.

If we used an slist, the nodes would be inside a wrapper struct that would point to the start of each string. The user could use the slist macros to iterate through the slist and access each name directly.  They would not need to parse the JSON and iterate through that.

plskeggs (approved):
Looks great. It would be nice to see this used in one of the samples.

jayteemo:
> It would be nice to see this used in one of the samples.


IRIS-8530 will add location library support so MSS can get the anchor names.

We don't have any other samples that support Wi-Fi location.

plskeggs:
> > It would be nice to see this used in one of the samples.
> 
> IRIS-8530 will add location library support so MSS can get the anchor names.
> 
> We don't have any other samples that support Wi-Fi location.

True, the location sample does, but is not an nrfcloud sample per se. MSS is fine.

peknis (approved):

include/net/nrf_cloud_location.h
```diff
@@ -41,6 +45,12 @@ enum nrf_cloud_location_type {
 	LOCATION_TYPE__INVALID
 };
 
+/** @brief Anchor name list node, to be used with @ref single-linked-list_apis */
```
peknis, line 48: This @ref makes the doc build fail.

jayteemo: removed the ref.
is there a proper way to do such a reference?

peknis: Where exactly is this link supposed to link to? Seems that links within the API doc to structures and members work ok.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -2815,6 +2815,75 @@ static bool json_item_string_exists(const cJSON *const obj, const char *const ke
...
+
+		/* Get the size of the anchor name and node container */
+		name_sz = strlen(name->valuestring) + 1;
+		node_sz = sizeof(*node) + name_sz;
+
+		/* Ensure node will fit in buffer */
+		if ((buf_idx + node_sz) > location_out->anchor_buf_sz) {
+			LOG_WRN("Anchor does not fit in provided buffer");
+			/* Reset the list and buffer */
+			sys_slist_init(&location_out->anchor_list);
```
glarsennordic, line 2867: Instead of resetting the list, could we keep the list and update `anchor_cnt`?

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -2815,6 +2815,75 @@ static bool json_item_string_exists(const cJSON *const obj, const char *const ke
...
+			location_out->anchor_buf[0] = '\0';
+			ret = -ENOBUFS;
+			break;
+		}
+
+		/* Get a node pointer from the anchor buffer */
+		node = (struct nrf_cloud_anchor_list_node *)&location_out->anchor_buf[buf_idx];
```
glarsennordic, line 2874: Do we have any kind of alignment restrictions on our systems? I need to play around with this and find out.

But supposing the previous entry has an anchor name whose length is not `4n-1` for some integer `n`, then this would result in a non-aligned `node->node.next` pointer, right?

But maybe that doesn't matter on our systems? This is an area I am not deeply familiar with, sorry.

subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
```diff
@@ -668,9 +671,16 @@ static int location_process(const char * const buf)
...
-		struct nrf_cloud_location_result res;
-		int ret = nrf_cloud_location_process(buf, &res);
 
+		int ret;
+		struct nrf_cloud_location_result res = {0};
+
+#if defined(CONFIG_NRF_CLOUD_LOCATION_ANCHOR_LIST)
+		res.anchor_buf = anchor_list_buf;
+		res.anchor_buf_sz = sizeof(anchor_list_buf);
+#endif
```
glarsennordic, line 681: One potential issue: What happens when we get a new location result while the user is still processing the old one?

We should probably include an explicit warning in the struct comments and the KConfig comments that the provided location buffer is ONLY valid during the callback, and that any long-term use should be done by copying the results somewhere else.

And then I think we should also use a semaphore to protect this buffer while `location_cb` is being evaluated

And then, finally, we should probably wipe the buffer immediately after `location_cb` is called just to help enforce this.

glarsennordic (commented):
Looks great! Most of my concerns are minor, but I probably shouldn't approve until we at least do the semaphore thing

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_location
```diff
@@ -8,3 +8,18 @@ config NRF_CLOUD_LOCATION
...
+
+config NRF_CLOUD_LOCATION_ANCHOR_LIST
+	bool "Include the list of anchor names when receiving location results from the callback"
+	depends on NRF_CLOUD_LOCATION
+	depends on WIFI
+
+config NRF_CLOUD_LOCATION_ANCHOR_LIST_BUFFER_SIZE
+	int "Size of the buffer to contain the anchor name list"
+	depends on NRF_CLOUD_LOCATION_ANCHOR_LIST
+	range 37 740
```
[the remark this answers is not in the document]
jayteemo: 20 anchors seemed sufficient.
i imagine most use cases will have... one anchor.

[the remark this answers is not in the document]
glarsennordic: I agree, though I'd argue we should just let the end-user decide on this, rather than force any specific limitation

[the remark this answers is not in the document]
jayteemo: that's fine.
what is the kconfig syntax for this?

subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
jayteemo: this is done in the context of the mqtt event handler

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
jayteemo: i'd prefer to keep `anchor_cnt` unchanged.
how about keeping the list data and updating the comments?

```
	/** List of received anchor names contained as @ref nrf_cloud_anchor_list_node.
	 *  If all anchor names cannot fit in the list, the number of items in the list
	 *  will be less than @ref anchor_cnt.
	 */
	sys_slist_t anchor_list;
```

subsys/net/lib/nrf_cloud/src/nrf_cloud_fsm.c
glarsennordic: I'm more just fearful that the end-user attempts to retain a copy of what they were given. But you're right, the semaphore isn't needed. But explicit warnings not do that / some mild enforcement by wiping the static buffer both seem potentially wise

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
glarsennordic: Perfect. This avoids any confusion, I like it.

glarsennordic (approved):
LGTM! But do consider still adding the warning about not retaining any of the provided anchor names

peknis:
> LGTM! But do consider still adding the warning about not retaining any of the provided anchor names

Not a warning, but maybe a note or caution.

PR #14278, opened 2024-02-29
Misc. fixes for Bugs found during NCS 2.6.0 release Testing

SebastianBoe:
see commit messages

test-sdk-nrf: test_cracen_ncs

### DIFF
--- a/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
+++ b/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
@@ -28,8 +28,6 @@
 #include <zephyr/sys/__assert.h>
 #include <zephyr/sys/byteorder.h>
 
-/* Entropy is recommended to be max 32 (in trng.h) */
-#define ENTROPY_SIZE (32)
 enum asn1_tags {
 	ASN1_SEQUENCE = 0x10,
 	ASN1_INTEGER = 0x2,
--- a/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
+++ b/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
@@ -7,9 +7,6 @@
 #include "common.h"
 #include <cracen/mem_helpers.h>
 #include "cracen_psa.h"
-#include "psa/crypto.h"
-#include "psa/crypto_values.h"
-#include <psa/crypto_sizes.h>
 
 #include <sicrypto/drbghash.h>
 #include <sicrypto/ecc.h>
@@ -18,6 +15,7 @@
 #include <sicrypto/ed448.h>
 #include <sicrypto/montgomery.h>
 #include <sicrypto/rsa_keygen.h>
+#include <sicrypto/util.h>
 #include <silexpk/ed448.h>
 #include <silexpk/sxops/rsa.h>
 #include <sicrypto/ik.h>
@@ -33,6 +31,8 @@ enum asn1_tags {
 	ASN1_INTEGER = 0x2,
 	ASN1_CONSTRUCTED = 0x20
 };
+
+extern const uint8_t cracen_N3072[384];
 
 static psa_status_t check_brainpool_alg_and_key_bits(psa_algorithm_t alg, size_t key_bits)
 {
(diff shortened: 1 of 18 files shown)


subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
```diff
@@ -429,6 +426,89 @@ static psa_status_t import_rsa_key(const psa_key_attributes_t *attributes, const
...
+		}
+		if (bits != 0 && (bits != 256)) {
+			return PSA_ERROR_INVALID_ARGUMENT;
+		}
+		if (constant_memcmp_is_zero(data, 32)) {
+			return PSA_ERROR_INVALID_ARGUMENT;
+		}
+		if (constant_memcmp_is_zero(data + 32, 32)) {
+			return PSA_ERROR_INVALID_ARGUMENT;
+		}
```
SebastianBoe, line 450: I assume we do constant_memcmp to avoid timing analysis.

Do we care that attackers can know which of the two 32bit fields in the data are zero?

vili-nordic: No, we can compare the "normal" way here

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/spake2p.c
```diff
@@ -224,10 +222,33 @@ static psa_status_t cracen_get_confirmation_keys(cracen_spake2p_operation_t *ope
...
+			&kdf_op, PSA_KEY_DERIVATION_INPUT_SECRET, V, hash_len);
+		if (status) {
+			goto exit;
+		}
+		status = psa_driver_wrapper_key_derivation_output_bytes(&kdf_op, operation->shared,
+									operation->shared_len);
+		if (status) {
+			goto exit;
+		}
+		psa_key_derivation_abort(&kdf_op);
```
SebastianBoe, line 250: was it intentional to use PSA here and the wrappers elsewhere?

This inconsistency would be good to document if so.

I see the exit: uses the wrapper.

vili-nordic: right, it should probably be the wrapper here

SebastianBoe, line 250: Someone might not know that PSA is inconsistent here and that abort is required on success.

Let's add a comment saying that abort is usually not required on success, but for KDF it is.

vili-nordic: the abort is to be able to run setup again on the next line. also there's no finish() for kdf, so you'd always have to call abort to end the operation.

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/pake.c
```diff
@@ -110,56 +43,43 @@ psa_status_t cracen_pake_set_user(cracen_pake_operation_t *operation, const uint
...
+					       peer_id_len);
+	} else {
+		return PSA_ERROR_NOT_SUPPORTED;
+	}
+}
+
+psa_status_t cracen_pake_set_context(cracen_pake_operation_t *operation, const uint8_t *context,
+				     size_t context_length)
 {
 	psa_status_t status = PSA_ERROR_BAD_STATE;
```
mswarowsky, line 74: Should this be `BAD_STATE` or rather `NOT_SUPPORTED` as we don't have have 
`else {
		return PSA_ERROR_NOT_SUPPORTED;
	}`   in this function at the end of the if / else cascade

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/pake.c
```diff
@@ -169,24 +89,12 @@ psa_status_t cracen_pake_set_role(cracen_pake_operation_t *operation, psa_pake_r
...
```
mswarowsky, line 90: Same here as in `cracen_pake_set_context`
and the follwoing functions?

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
```diff
@@ -428,6 +428,101 @@ static psa_status_t import_rsa_key(const psa_key_attributes_t *attributes, const
...
```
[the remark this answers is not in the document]
SebastianBoe: presumably we only want to test the first and 32'nd byte, and not bytes 0 to 32.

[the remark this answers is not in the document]
vili-nordic: because it is an error if either w0 or w1 is 0. If we just do one call w0 or w1 can be 0

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
```diff
@@ -428,6 +428,101 @@ static psa_status_t import_rsa_key(const psa_key_attributes_t *attributes, const
...
+		if (bits != 0 && (bits != 256)) {
+			return PSA_ERROR_INVALID_ARGUMENT;
+		}
+		if (constant_memcmp_is_zero(data, 32)) {
+			return PSA_ERROR_INVALID_ARGUMENT;
+		}
+
+		if (check_wstr_pub_key_data(PSA_ALG_ECDH, PSA_ECC_FAMILY_SECP_R1, bits, &data[32],
+					    65)) {
```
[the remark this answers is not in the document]
vili-nordic: 65 is correct. we can of course add comments

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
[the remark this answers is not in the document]
SebastianBoe: two can be 0, but not exactly one?

I didn't understand that from the code either, maybe add a comment? :)

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/pake.c
vili-nordic: These would have failed in cracen_pake_setup, so bad state seem appropriate

mswarowsky (approved):

SeppoTakalo (approved):

PR #14280, opened 2024-02-29
applications: nrf5340_audio: Update error check for scanning

rick1082:
Update the error check for scanning for bypassing the scan start/stop already return code.
OCT-2858

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_management/bt_mgmt.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_management/bt_mgmt.c
@@ -103,7 +103,7 @@ static void connected_cb(struct bt_conn *conn, uint8_t err)
 		if (IS_ENABLED(CONFIG_BT_CENTRAL)) {
 			ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_CONN, NULL,
 						 BRDCAST_ID_NOT_USED);
-			if (ret) {
+			if (ret && ret != -EALREADY) {
 				LOG_ERR("Failed to restart scanning: %d", ret);
 			}
 		}
@@ -170,6 +170,8 @@ static void connected_cb(struct bt_conn *conn, uint8_t err)
 	}
 }
 
+K_MUTEX_DEFINE(mtx_duplicate_scan);
+
 static void disconnected_cb(struct bt_conn *conn, uint8_t reason)
 {
 	int ret;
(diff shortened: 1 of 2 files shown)


applications/nrf5340_audio/src/bluetooth/bt_management/bt_mgmt.c
```diff
@@ -199,8 +204,13 @@ static void disconnected_cb(struct bt_conn *conn, uint8_t reason)
 
 	if (IS_ENABLED(CONFIG_BT_CENTRAL)) {
 		ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_CONN, NULL, BRDCAST_ID_NOT_USED);
-		if (ret) {
-			LOG_ERR("Failed to restart scanning: %d", ret);
+		if (ret && ret != -EALREADY) {
+			if (ret == -EIO) {
+				LOG_WRN("Controller might already in scanning state, check the HCI "
```
koffes, line 209:
```suggestion
				LOG_WRN("Controller might already be scanning, check the HCI "
```

koffes: What is meant by check HCI?

applications/nrf5340_audio/src/bluetooth/bt_management/bt_mgmt.c
```diff
@@ -103,8 +103,13 @@ static void connected_cb(struct bt_conn *conn, uint8_t err)
 		if (IS_ENABLED(CONFIG_BT_CENTRAL)) {
 			ret = bt_mgmt_scan_start(0, 0, BT_MGMT_SCAN_TYPE_CONN, NULL,
 						 BRDCAST_ID_NOT_USED);
-			if (ret) {
-				LOG_ERR("Failed to restart scanning: %d", ret);
+			if (ret && ret != -EALREADY) {
+				if (ret == -EIO) {
+					LOG_WRN("Controller might already in scanning state, check "
```
koffes, line 108:
```suggestion
					LOG_WRN("Controller might already be scanning, check "
```

applications/nrf5340_audio/src/bluetooth/bt_management/bt_mgmt.c
koffes: Not sure we should mention HCI at all. This is a common debugging step.

koffes (changes requested):

koffes (approved):

applications/nrf5340_audio/src/bluetooth/bt_management/bt_mgmt.c
rick1082: There's a chance the log will print like this:
<wrn> bt_hci_core: opcode 0x2041 status 0x0c
0x2041 means scanning, 0x0c means already.
The host doesn't print low level response from controller, instead, the host only return -EIO in this case.

alexsven: #define BT_HCI_ERR_CMD_DISALLOWED   0x0c
But we need to figure out why we are trying to get the scanner to scan while it's already scanning

rick1082: Just figured out the root cause:
If two headsets are reset at a very close time, the disconnect_cb() from gateway will be triggered twice all most at the ver y close time. So bt_mgmt_scan_start() might be executed twice closely, So the check mechanism might failed leads to duplicate scanning.

applications/nrf5340_audio/src/bluetooth/bt_management/bt_mgmt.c
```diff
@@ -196,13 +198,14 @@ static void disconnected_cb(struct bt_conn *conn, uint8_t reason)
 		ret = bt_mgmt_adv_start(NULL, 0, NULL, 0, true);
 		ERR_CHK(ret);
 	}
-
+	k_mutex_lock(&mtx_duplicate_scan, K_FOREVER);
```
alexsven, line 201: Very nice, just add a comment above to explain why we have a mutex here

alexsven (changes requested):

alexsven (approved):

SeppoTakalo (approved):

jfischer-no (approved):

PR #14285, opened 2024-02-29
lib: dk_buttons_and_leds: Add no interrupts mode for buttons

mkapala-nordic:
Add no interrupts mode for buttons (CONFIG_DK_LIBRARY_BUTTON_NO_ISR). This option avoids relying on GPIO interrupts at all and unconditionally periodically scan all of the buttons.

It can be used on the nRF54L15 PDK revision 0.2.X, where buttons 3 and 4 are connected to GPIO PORT2 which does not support GPIOTE. Because of that, the PDK cannot rely on GPIO interrupts for the mentioned buttons.

Jira: NCSDK-23210

PR #14286, opened 2024-02-29
samples: bluetooth: fast_pair: input_device: Add init nRF54L support 

mkapala-nordic:
Depends on: 
* https://github.com/nrfconnect/sdk-nrf/pull/14285 -> buttons workaround
* https://github.com/nrfconnect/sdk-nrf/pull/14368 -> partition manager for nRF54L fix
* https://github.com/nrfconnect/sdk-nrf/pull/14409 -> partition manager for nRF54L support in sysbuild

Jira: NCSDK-26123

### DIFF
--- a/samples/bluetooth/fast_pair/input_device/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
+++ b/samples/bluetooth/fast_pair/input_device/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
@@ -1,0 +1,11 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Workaround:
+# The nRF54L15 PDK (PCA10156) revisions v0.2.0 AA0-ES2, v0.2.0 AA0-ES3,
+# and v0.2.1 AB0-ES5 have Buttons 3 and 4 connected to the GPIO port
+# which does not support interrupts.
+CONFIG_DK_LIBRARY_BUTTON_NO_ISR=y
--- a/samples/bluetooth/fast_pair/input_device/sample.yaml
+++ b/samples/bluetooth/fast_pair/input_device/sample.yaml
@@ -9,9 +9,10 @@ tests:
       - nrf52840dk_nrf52840
       - nrf5340dk_nrf5340_cpuapp
       - nrf5340dk_nrf5340_cpuapp_ns
+      - nrf54l15pdk_nrf54l15_cpuapp
     tags: bluetooth ci_build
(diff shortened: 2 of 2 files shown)


subsys/bluetooth/services/fast_pair/CMakeLists.txt
```diff
@@ -51,41 +51,69 @@ if(NOT SYSBUILD)
...
+                          -m ${FP_MODEL_ID} -k ${FP_ANTI_SPOOFING_KEY}
+      COMMAND_ECHO STDOUT
+      )
+
+    set_property(
+      GLOBAL APPEND PROPERTY
+      HEX_FILES_TO_MERGE
+      ${FP_PROVISIONING_DATA_HEX}
+      ${CMAKE_BINARY_DIR}/zephyr/zephyr.hex
+      )
```
nordicjm, line 109: I don't really have time at present to go through this in detail or check it but at first glance this does not seem right, it assumes that the output is names zephyr.hex, which is not the case. The output filename is set with a Kconfig, and there's also the possibility of using MCUboot signing which generates a signed image

mkapala-nordic: > The output filename is set with a Kconfig

Alright, I was not aware of the `CONFIG_KERNEL_BIN_NAME` Kconfig option. Thanks, fixed.

> there's also the possibility of using MCUboot signing which generates a signed image

I think that MCUboot forces to use Partition Manager so should not be a problem here as this path is only valid for the `CONFIG_PARTITION_MANAGER_ENABLED=n`

SeppoTakalo (approved):

MarekPieta (changes requested):

subsys/bluetooth/services/fast_pair/CMakeLists.txt
```diff
@@ -51,41 +51,69 @@ if(NOT SYSBUILD)
...
+      ${CMAKE_BINARY_DIR}/zephyr/${CONFIG_KERNEL_BIN_NAME}.hex
+      )
+
+    set_property(
+      TARGET
+      runners_yaml_props_target
+      PROPERTY
+      hex_file
+      ${CMAKE_BINARY_DIR}/zephyr/merged.hex
+      )
```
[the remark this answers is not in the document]
MarekPieta: E.g. try to add Zephyr hex file again to replicate a build error

```
    set_property(                                                                                    
      GLOBAL APPEND PROPERTY                                                                         
      HEX_FILES_TO_MERGE                                                                             
      ${CMAKE_BINARY_DIR}/zephyr/${CONFIG_KERNEL_BIN_NAME}.hex                                       
      )
```

alstrzebonski (commented):
No comments from me apart from Marek's comments. I will rereview once they will be addressed.

nordicjm (approved):

MarekPieta (approved):

PR #14292, opened 2024-02-29
rpc_flash test

michalek-no:


tests/drivers/rpc_flash/src/main.c
```diff
@@ -0,0 +1,116 @@
...
+	const struct flash_parameters *flash_params =
+			flash_get_parameters(flash_dev);
+
+	flash_get_page_info_by_offs(flash_dev, TEST_AREA_OFFSET,
+				    &page_info);
+
+	/* Check if test region is not empty */
+	uint8_t buf[EXPECTED_SIZE];
+
+	rc = flash_read(flash_dev, TEST_AREA_OFFSET,
```
de-nordic, line 39: I do not think this is correct. Here the output of the `flash_get_page_info_by_offs` should rather be used as it is not warranted that `TEST_AREA_OFFSET` will be `page_info.start_offset`.

tests/drivers/rpc_flash/src/main.c
```diff
@@ -0,0 +1,116 @@
...
+
+	if (!is_buf_clear) {
+		/* Erase a nb of pages aligned to the EXPECTED_SIZE */
+		rc = flash_erase(flash_dev, page_info.start_offset,
+				(page_info.size *
+				((EXPECTED_SIZE + page_info.size - 1)
+				/ page_info.size)));
+
+		zassert_equal(rc, 0, "Flash memory not properly erased");
+	}
```
de-nordic, line 70: We are saving flash endurance here or time, because I do not think that we should bother doing that with tests, 90% of time we will have to clean pre-erase the device anyway, because there will be some leftovers of previous test runs.

tests/drivers/rpc_flash/src/main.c
```diff
@@ -0,0 +1,116 @@
...
+	uint8_t buf[EXPECTED_SIZE];
+
+	rc = flash_read(flash_dev, TEST_AREA_OFFSET,
+			buf, EXPECTED_SIZE);
+	zassert_equal(rc, 0, "Cannot read flash");
+
+	/* Fill test buffer with random data */
+	for (int i = 0; i < EXPECTED_SIZE; i++) {
+		expected[i] = i;
+	}
```
de-nordic, line 46: This should skip putting CANARY in the buffer.

tests/drivers/rpc_flash/src/main.c
```diff
@@ -0,0 +1,116 @@
...
+	/* read buffer length*/
+	for (off_t len = 0; len < 25; len++) {
+		/* address offset */
+		for (off_t ad_o = 0; ad_o < 4; ad_o++) {
+			/* buffer offset; leave space for buffer guard */
+			for (off_t buf_o = 1; buf_o < 5; buf_o++) {
+				/* buffer overflow protection */
+				buf[buf_o - 1] = CANARY;
+				buf[buf_o + len] = CANARY;
+				memset(buf + buf_o, 0, len);
```
de-nordic, line 94: IMHO you could just fill in the buf with CANARY; you know where your real data will start and end, and, assuming that there are no CANARies in `expected`, you can figure out whether boundaries of buffer have been crossed in either way.

PR #14308, opened 2024-03-01
Add missing Carrier library event logs

nordic-mare:
Add logs for the events `LWM2M_CARRIER_EVENT_MODEM_DOMAIN` and `LWM2M_CARRIER_EVENT_APP_DATA` to the samples and applications where these were missing.

### DIFF
--- a/samples/cellular/modem_shell/src/shell.c
+++ b/samples/cellular/modem_shell/src/shell.c
@@ -144,6 +144,12 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 	case LWM2M_CARRIER_EVENT_REBOOT:
 		mosh_print("LwM2M carrier event: reboot");
 		break;
+	case LWM2M_CARRIER_EVENT_MODEM_DOMAIN:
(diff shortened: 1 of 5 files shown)


trantanen (approved):

lucasvdp (approved):

samples/cellular/modem_shell/src/shell.c
```diff
@@ -144,6 +144,12 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 	case LWM2M_CARRIER_EVENT_REBOOT:
 		mosh_print("LwM2M carrier event: reboot");
 		break;
+	case LWM2M_CARRIER_EVENT_MODEM_DOMAIN:
+		mosh_print("LwM2M carrier event: modem domain");
```
tokangas, line 148: This is ok as such, but also the actual modem domain event could be printed here. It's a single integer.

tokangas (approved):

SeppoTakalo (approved):

PR #14314, opened 2024-03-01
modules: trusted-firmware-m: Fix include guard

Vge0rge:
CONFIG_CRYPTOCELL_USABLE is now removed and cannot be used. Replace it with CONFIG_HAS_HW_NRF_CC3XX which is supported.


vili-nordic:
Same change is in https://github.com/nrfconnect/sdk-nrf/pull/14278 ?

Vge0rge:
True, I was not aware of this, sorry. I will close this PR then.

PR #14316, opened 2024-03-01
samples: lwm2m_client: Workaround WiFi ground fix problem

SeppoTakalo:
AVSystem server assumes incorrect object version for Connection Monitoring object.

Force object version reporting by using different
Connection Monitor object version.

Reported as CDM-665


SeppoTakalo:
Does not need changelog, because this is already reported and fixed for Cell, A-GPS/P-GPS. Just the WiFi was missing.

MarkusLassila (approved):

PR #14321, opened 2024-03-01
Upmerge 01.03.2024

rlubos:
test_iot_libmodem: PR-4439
test_crypto: PR-608
test_ble: DRGN-21609_upmerge_test_ble

Upmerge 01.03.2024:
- [x] https://github.com/nrfconnect/sdk-zephyr/pull/1555
- [x] https://github.com/nrfconnect/sdk-mcuboot/pull/301

PRs with fixes:
- [x] https://github.com/nrfconnect/sdk-hostap/pull/158
- [x] https://github.com/nrfconnect/sdk-connectedhomeip/pull/402
- [x] https://github.com/nrfconnect/sdk-nrfxlib/pull/1242
- [x] https://github.com/nrfconnect/sdk-sidewalk/pull/512


tomi-font (approved):
Approving SLM.

MechanicV1:
testing downstream propagation.

test_iot_libmodem: PR-4439

thst-nordic:
> test_iot_libmodem: PR-4439
This has to be in the PR description to be parsed.

modules/hostap/Kconfig
```diff
@@ -8,8 +8,9 @@
 menuconfig WPA_SUPP
 	bool "WPA supplicant support"
 	# Need full POSIX from libc, Zephyr's POSIX support is only partial
-	depends on !POSIX_API
 	select POSIX_CLOCK
+	select POSIX_SIGNAL
```
Damian-Nordic, line 12: Can this be avoided? In Matter we use `NET_SOCKETS_POSIX_NAMES` which seems mutually exclusive with POSIX_API.

SeppoTakalo: This is the reason why I always try to use `zsock_` calls and leave the `POSIX_NAMES` or `POSIX_API` for application to choose.

This is constant source of conflicts.

rlubos: Here's some reference: https://github.com/zephyrproject-rtos/zephyr/issues/56294

But to answer your question - I didn't really find an alternative to mitigate the issues with POSIX signal API usage, other than finally enabling the POSIX_SIGNAL (and POSIX_API as a consequence) for the supplicant. The recent reworks upstream made the libc headers/definitions incompatible with rest, causing build errors.

As for `NET_SOCKETS_POSIX_NAMES` - the transition to `POSIX_API` isn't that bad, just need to add a few extra headers, see other commits like the one with `https_client`. Or just stick to `zsock_` as suggested.

applications/machine_learning/src/events/ml_result_event.c
```diff
@@ -14,7 +14,7 @@ static void log_ml_result_event(const struct app_event_header *aeh)
 	const struct ml_result_event *event = cast_ml_result_event(aeh);
 
 	APP_EVENT_MANAGER_LOG(aeh, "%s val: %0.2f anomaly: %0.2f",
-			event->label, event->value, event->anomaly);
+			event->label, (double)event->value, (double)event->anomaly);
```
pdunaj, line 17: why is this needed - aren't all float arguments promoted to double in such function calls anyway?

modules/hostap/Kconfig
krish2718: > I think it's quite unfortunate that a low-level component forces POSIX while this should be an application's decision, but if it's unavoidable we will have to deal with it.

Technically WPA supplicant is an application :) which was turned in to a library.

applications/machine_learning/src/events/ml_result_event.c
rlubos: Yes, they are. This results in a warning however (since https://github.com/zephyrproject-rtos/zephyr/pull/57154) as Zephyr enabled `-Wdouble-promotion` compiler flag, which gives a warning in such cases (implict promotion to double). Therefore, we either need to make the promotion explicit to avoid the warning, or find some other way to avoid implicit promotion.

pdunaj: thanks. zephyr should have taken on java instead...

gmarull (approved):

MarkusLassila (approved):

maxd-nordic (approved):

nordicjm (approved):

jfischer-no (approved):
LGTM

jfischer-no (approved):

PR #14323, opened 2024-03-01
nrf_security: cracen: Fix configuration system

mswarowsky:
The CRACEN configuration system wasn't working with oberon core 1.2.1.1 As the key management changed, this fixed that for CRACEN


subsys/nrf_security/src/drivers/cracen/cracenpsa/src/common.c
```diff
@@ -125,27 +125,27 @@ static psa_status_t get_sx_brainpool_curve(size_t curve_bits, const struct sx_pk
...
 		break;
 	case 384:
-		IF_ENABLED(PSA_NEED_CRACEN_KEY_MANAGEMENT_BRAINPOOL_P_R1_384,
+		IF_ENABLED(PSA_NEED_CRACEN_KEY_TYPE_ECC_ANY,
 			   (selected_curve = &sx_curve_brainpoolP384r1));
 		break;
 	case 512:
-		IF_ENABLED(PSA_NEED_CRACEN_KEY_MANAGEMENT_BRAINPOOL_P_R1_512,
+		IF_ENABLED(PSA_NEED_CRACEN_KEY_TYPE_ECC_ANY,
 			   (selected_curve = &sx_curve_brainpoolP512r1));
```
SebastianBoe, line 149: won't this new system put all or none of the curves in the binary whereas before we were able to just put the curves we wanted into the binary?

SebastianBoe, line 149: Let's not repeat ourselves so much.

if(IS_ENABLED(PSA_NEED_CRACEN_KEY_TYPE_ECC_ANY)) {
switch(curve_bits) bla bla
} else {
  return not supported;
}

SebastianBoe: I compared the map file of HCS and NCS.

This new config system pulls in all the curves whereas the old did not.

For the ecjpake sample we are consuming an extra 11kB of FLASH.

old:
           FLASH:       63000 B      1524 KB      4.04%
             RAM:       32912 B       256 KB     12.55%


new:
           FLASH:       74032 B      1524 KB      4.74%
             RAM:       24696 B       256 KB      9.42%

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
```diff
@@ -12,380 +12,719 @@ visible if 0
...
+	depends on PSA_NEED_CRACEN_HMAC || PSA_NEED_CRACEN_CMAC
 
 # CRACEN Entropy Driver
 
 config PSA_NEED_CRACEN_CTR_DRBG_DRIVER
 	bool
 	default y
+	select PSA_ACCEL_GENERATE_RANDOM
+	# Is needed for AES counter measurements
+	depends on (PSA_WANT_ALG_CTR_DRBG		&& \
```
SebastianBoe, line 723: Don't we want to be using CTR_DRBG for AES countermeasures independently of whether the user has asked for CTR_DRBG?

SebastianBoe (changes requested):
I've done a first review pass, but if you don't mind, I would like to review after we have a green CI :)

subsys/nrf_security/src/drivers/Kconfig
```diff
@@ -202,11 +205,6 @@ config PSA_USE_CRACEN_CTR_DRBG_DRIVER
 	bool
 	default y
 
-# CTR DRBG is needed for AES counter measurements
-config PSA_NEED_CRACEN_CTR_DRBG_DRIVER
-	bool
-	default y
-
```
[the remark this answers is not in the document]
mswarowsky: By adding the dependencies when it is needed here:
https://github.com/nrfconnect/sdk-nrf/pull/14323/files#diff-90f0fe40d5fc3bd7321de95b0334ac7ef05a12b9cf2422cee238b484219f7a05R724

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/common.c
mswarowsky: Yes it does that is due to new PSA Core that is more detailed on the Key types than the old one we have follow up task (NCSDK-26237) tho add a config that separates the curves, it is planned for the next sprint after this gets in

mswarowsky: I would leave the code here referring to NCSDK-26237 as then we have to add it again

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
mswarowsky: Yes it gets selected if the user wants it `(PSA_WANT_ALG_CTR_DRBG && PSA_USE_CRACEN_CTR_DRBG_DRIVER)`
or if we use cryptomaster
`PSA_USE_CRACEN_CIPHER_DRIVER || PSA_USE_CRACEN_AEAD_DRIVER || PSA_USE_CRACEN_MAC_DRIVER`

SebastianBoe: ah, now I got it, nicely done :)

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/common.c
SebastianBoe: ah, I was not aware of this follow-up task. Good to know!

SebastianBoe: true

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
```diff
@@ -12,2688 +12,821 @@ visible if 0
 config PSA_NEED_CRACEN_CCM_AES
 	bool
 	default y
+	select PSA_ACCEL_CCM_AES_128
+	select PSA_ACCEL_CCM_AES_192
+	select PSA_ACCEL_CCM_AES_256
+	depends on PSA_WANT_AES_KEY_SIZE_128
```
SebastianBoe, line 18: I see that for Cryptocell, we document in driver_config.rst:

The :ref:`nrf_security_drivers_cc3xx` is limited to AES key sizes of 128 bits on devices with Arm CryptoCell cc310.

But not for CRACEN.

Did we accidentally copy this dependency from Cryptocell to CRACEN?

Is CRACEN limited to AES128? If so we should update the docs accordingly, if not we should remove this depends (or maybe expand it to depend on all the key sizes).

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
```diff
@@ -12,2688 +12,821 @@ visible if 0
...
 
 # CRACEN Cipher Driver
 
 config PSA_NEED_CRACEN_CTR_AES
 	bool
 	default y
+	select PSA_ACCEL_CTR_AES_128
+	select PSA_ACCEL_CTR_AES_192
+	select PSA_ACCEL_CTR_AES_256
+	depends on PSA_WANT_AES_KEY_SIZE_128 || PSA_WANT_AES_KEY_SIZE_192 || PSA_WANT_AES_KEY_SIZE_256
```
SebastianBoe, line 53: Why does CTR_AES depend on a key size being enabled, but GCM does not?

SebastianBoe (approved):

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
mswarowsky: It is aligned with the oberon configs, I thinking behind that is to make sure it is only enabled if the supported key size is set

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
mswarowsky: Good catch, started with taking cryptocell as reference and this limitation is not there for CRACEN, so added the other key sizes as well

SebastianBoe (approved):

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
```diff
@@ -12,2688 +12,1741 @@ visible if 0
...
+	depends on PSA_NEED_CRACEN_KEY_TYPE_SRP_6_PUBLIC_KEY_3072 	|| \
+		   PSA_NEED_CRACEN_KEY_TYPE_SRP_6_KEY_PAIR_IMPORT_3072 	|| \
+		   PSA_NEED_CRACEN_KEY_TYPE_SRP_6_KEY_PAIR_EXPORT_3072
 
-config PSA_ACCEL_KEY_TYPE_ECC_KEY_PAIR_DERIVE_SECP_R1_192
+# RSA key types
+config PSA_NEED_CRACEN_KEY_TYPE_RSA_PUBLIC_KEY
 	bool
 	default y
+	select PSA_ACCEL_KEY_TYPE_RSA_1536_PUBLIC_KEY
```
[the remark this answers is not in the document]
mswarowsky: very good catch, I even added this to the documentation as not supported... removed from RSA

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/pake.c
```diff
@@ -43,13 +45,19 @@ psa_status_t cracen_pake_set_user(cracen_pake_operation_t *operation, const uint
 {
 	psa_status_t status = PSA_ERROR_BAD_STATE;
 
-	if (PSA_ALG_IS_JPAKE(operation->alg) && IS_ENABLED(PSA_NEED_CRACEN_ECJPAKE_SECP_R1_256)) {
+	if (PSA_ALG_IS_JPAKE(operation->alg)) {
+#ifdef CONFIG_PSA_NEED_CRACEN_ECJPAKE
```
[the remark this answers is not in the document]
mswarowsky: Fair point change it to `PSA_NEED_XXXX`

Vge0rge (approved):

magnev (approved):

PR #14324, opened 2024-03-04
sample: wifi_provisioning: fix stack overflow bug

wentong-li:
In commit https://github.com/nrfconnect/sdk-hostap/commit/3f3b7c45620b6f9710c884ea84d3e1811362117d, the supplicant buffer no longer requests memory from heap. It leads to higher stack usage of which calls the wifi connection api, and causes stack overflow of BT RX thread in wifi provisioning sample given its stack size of 4096B. Increasing the stack size by 1024B accordingly can solve the issue.


lats1980 (approved):
Looks good to me!

krish2718 (approved):
Thanks, Wentong. You just beat me to it, I had the same fix in my local :).

markaj-nordic:
@krish2718 I expect other Wi-Fi based apps can be impacted as well, because the related change in wpa_cli_cmd is quite generic?

krish2718:
@markaj-nordic  yeah, I had already fixed this in the shell sample, need to check others, but haven't seen any issues reported except for provisioning, will keep an eye.

SeppoTakalo (approved):

PR #14326, opened 2024-03-04
manifest: sdk-nrfxlib: Pull command status check

ajayparida:
Pull command status check phase1 changes.


markaj-nordic:
@ajayparida the Matter failure is caused by the net_mgmt API returning error when requesting for low power mode:
https://github.com/nrfconnect/sdk-connectedhomeip/blob/89eb37b47ca8e4b9d309c431bde30bdb19d07577/src/platform/nrfconnect/wifi/WiFiManager.cpp#L696

I believe the nrfxlib related change refines commands handling between host and RPU, so it looks like a regression to me. Is there any test in WiFi suite that checks the `net_mgmt(NET_REQUEST_WIFI_PS...)` call?

ajayparida:
> @ajayparida the Matter failure is caused by the net_mgmt API returning error when requesting for low power mode: https://github.com/nrfconnect/sdk-connectedhomeip/blob/89eb37b47ca8e4b9d309c431bde30bdb19d07577/src/platform/nrfconnect/wifi/WiFiManager.cpp#L696
> 
> I believe the nrfxlib related change refines commands handling between host and RPU, so it looks like a regression to me. Is there any test in WiFi suite that checks the `net_mgmt(NET_REQUEST_WIFI_PS...)` call?
> 
> Does application code require alignments after your change?

Hi, 

Yes, there is change in nrfxlib. Host passes the commands to rpu and will wait for command status event. In a way host will know, command has been indeed received by rpu and processed. On timeout, in case no event received it will return failure. This way it helps to know that command has been received by fw and processed.

There is no separate test suite,  we can use existing "shell" sample to set ps on/off("cmd_wifi_ps on/off"). It will invoke net_mgmt(NET_REQUEST_WIFI_PS...) with params.enabled=on/off as being done in your code.

No, I don't think application code require alignment. As in case host driver doesn't receive status event, it will return failure . It will be same as currenetly expected at your application for failure of command. It will  call "ChipLogError(DeviceLayer, "Set low power mode request failed");"

I have also asked UMAC firmware team to check they are giving event status for the commad  NRF_WIFI_UMAC_CMD_SET_POWER_SAVE".

doublemis1 (changes requested):
Due to upmerge issues, the nRF7002 testing on Matter over WiFI is disabled on integration, however we found regression on that PR (current state).
After disabling the nRF7002DK in Matter testing the Integration tests will be green, however it does not mean that the regression has been fixed. Please contact me if the issue will be fixed, I can verify it on desk and give result.

PR #14329, opened 2024-03-04
applications: serial_lte_modem: set range for data buffer length

kacperradoszewski:
Defined a maximum range for SLM_CARRIER_APP_DATA_BUFFER_LEN to prevent stack overflow on the system work queue, whose size is configured independently, and to conform with a requirement from one of the operators.

### DIFF
--- a/applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
+++ b/applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
@@ -12,7 +12,8 @@ if SLM_CARRIER
 
 config SLM_CARRIER_APP_DATA_BUFFER_LEN
 	int "Size of the buffer for setting application data"
-	default 512
+	default 768
+	range 128 768
 	help
 	  Specifies maximum application data size to be set in the App Data Container,
 	  the Binary App Data Container or the Event Log objects.


MarkusLassila (approved):

applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
```diff
@@ -12,7 +12,8 @@ if SLM_CARRIER
 
 config SLM_CARRIER_APP_DATA_BUFFER_LEN
 	int "Size of the buffer for setting application data"
-	default 512
+	default 768
+	range 0 768
```
tomi-font, line 16: Just a wild idea: Could the value be made to depend on `CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE`?

tomi-font (approved):

kacperradoszewski:
@SeppoTakalo this is a requirement from one of the operators, please consider it for the v2.6.0 milestone.

nordic-mare (approved):

SeppoTakalo (approved):

applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
kacperradoszewski: It would probably make more sense, however in this case, the maximum range is dictated by a requirement from one of the operators.

SeppoTakalo: Unfortunately it cannot. Kconfig does not resolve symbols when parsing these.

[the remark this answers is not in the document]
kacperradoszewski: Done as suggested

junqingzou (approved):
Thanks

PR #14333, opened 2024-03-04
net: lib: download_client: Change log level

simensrostad:
Change log level for failure to resolve hostname
on a specific ip familiy, from warning to debug.

This is quite typical for services to not support
a specific family. AWS S3 for instance doesnt support IPv6.

### DIFF
--- a/subsys/net/lib/download_client/src/download_client.c
+++ b/subsys/net/lib/download_client/src/download_client.c
@@ -264,7 +264,7 @@ static int host_lookup(const char *host, int family, uint8_t pdn_id,
 	}
 
 	if (err) {
-		LOG_WRN("Failed to resolve hostname %s on %s",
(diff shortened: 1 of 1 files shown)


SeppoTakalo (approved):

SeppoTakalo:
If name resolution fails, we now don't receive any error messages anymore.

So I don't think this is a bug fix. Just a nice cleanup for when things work OK, but might hide the issue when there are problem.

I would consider dropping this away from release, as this does not really fix any known issue.

simensrostad:
> If name resolution fails, we now don't receive any error messages anymore.
> 
> So I don't think this is a bug fix. Just a nice cleanup for when things work OK, but might hide the issue when there are problem.
> 
> I would consider dropping this away from release, as this does not really fix any known issue.

Wasn't the intention to add the bug label. True, not critical at all. Ill leave it up to you if you want to drop it or not.

PR #14335, opened 2024-03-04
drivers: wifi: Fix setting monitor mode when station is connected

VivekUppunda:
The change fixes an issue where in monitor mode was able to be configured when station was in connected state. Disallow monitor mode to be set when station is in connected state.

Fixes SHEL-2555

### DIFF
--- a/drivers/wifi/nrf700x/src/wifi_mgmt.c
+++ b/drivers/wifi/nrf700x/src/wifi_mgmt.c
@@ -782,6 +782,12 @@ int nrf_wifi_mode(const struct device *dev,
 			goto out;
 		}
 
+		if (vif_ctx_zep->authorized && (mode->mode == NRF_WIFI_MONITOR_MODE)) {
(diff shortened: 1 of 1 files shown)


krish2718 (commented):
This needs a lock to synchronize with setting of `authorized` flag, as both can come from different threads user vs wpa_s.

VivekUppunda:
> This needs a lock to synchronize with setting of `authorized` flag, as both can come from different threads user vs wpa_s.

Both pieces of code are already under `k_mutex_lock(&vif_ctx_zep->vif_lock, K_FOREVER)` . So both code blocks cannot interfere with one another

kapbh (approved):

krish2718 (approved):

rado17 (approved):

sachinthegreen (approved):

SeppoTakalo (approved):

PR #14351, opened 2024-03-04
modules: hostap: Reduce WPA supplicant heap when MbedTLS heap is enabled

krish2718:
If MbedTLS uses its own heap which is a static heap (not libc heap), then WPA supplicant heap usage will come down, so, reduce 8K (minimum MbedTLS heap for Wi-Fi) from the libc heap size.

Fixes SHEL-2560.

### DIFF
--- a/modules/hostap/Kconfig
+++ b/modules/hostap/Kconfig
@@ -24,6 +24,8 @@ if WPA_SUPP
 
 config COMMON_LIBC_MALLOC_ARENA_SIZE
 	default 40000 if WPA_SUPP_AP
+	# 8192 for MbedTLS heap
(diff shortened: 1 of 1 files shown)


ioannisg:
@balaji-nordic  @simensrostad  maybe do a sanity-check on the networking samples, that currently enable the MBEDTLS own heap, confirming this works :)

sachinthegreen (approved):

jorgenmk (approved):

simensrostad (approved):
Passes CI for samples/net

krish2718:
Commit merged as part of https://github.com/nrfconnect/sdk-nrf/pull/14353/commits

PR #14354, opened 2024-03-05
samples: cellular: nrf_cloud_multi_service: Move ZCBOR_CANONICAL

plskeggs:
Move the config setting from the overlay-coap_nrf_provisioning.conf to overlay_coap.conf. This ensures ground-fix will work when using CoAP for nrfcloud communication regardless of which nrf_provisioning overlay is used, or even if any is used at all.

Jira: IRIS-8378


plskeggs:
> this solution is application specific. some enforcement of this dependency via kconfig or a build assert at the nrf_cloud library level, if possible, would be helpful.

Will address post-release.

glarsennordic (approved):

SeppoTakalo (approved):

PR #14357, opened 2024-03-05
manifest: Update Sidewalk revision

totyz:
Update sdk-sidewalk revision in manifest

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -196,7 +196,7 @@ manifest:
           compare-by-default: false
     - name: sidewalk
       repo-path: sdk-sidewalk
-      revision: v2.6.0-rc1
+      revision: 53627888ed64c7b9443f1d08dfe1f41a14486d24
       groups:
         - sidewalk
     - name: find-my


wojciechpluszynski (approved):

RobertGalatNordic (approved):

jfischer-no (changes requested):
There is no justification for this being part of RC1. There is no proper commit message describing the change. In fact, you should not just merge something into the sidewalk repository and open PR to check the manifest change in CI afterwards.

totyz:
@jfischer-no , If you go into Sidewalk repo you will see the changes. Their are small but essential and required to have software maturity PR#14332 update the docs as it should.

 In fact, you should not just merge something into the sidewalk repository and open PR to check the manifest change in CI afterwards. - Realy unfair statement - Why do you think I'm doing like this? Did you see too much PR update Sidewalk in nrf? Am I flood CI with my PRs - NO!

So, what should I do in your opinion to update Sidewalk in nrf now?

SeppoTakalo:
@totyz By justification, we mean that we should understand the reason why this should go into RC2.
We are already in phase where we should merge in just bugfixes.

Your commit message is just `Update sidewalk revision`  which does not indicate what has been changed.

And by looking at the changes, I'm still unsure why:
https://github.com/nrfconnect/sdk-sidewalk/compare/v2.6.0-rc1...main

If it because you want to lose the EXPERIMENTAL or is it just to fix the documentation? Rest of the changes seem to be just related to internal manifest, which should not concern the release.

Is this a bugfix, or is this a documentation update?

totyz:
@SeppoTakalo ,
The most important change in this update is to lose the EXPERIMENTAL which allows me to create good SW maturity level with PR mentioned earlier.
So, this is documentation update.

Anyhow, I expect one more sidewalk update, next week or end of this week, which will fix bugs we found.

I can decline this PR and do everything in doc+bugfix PR latter on. Up to you

totyz:
There will be other PR which introduce wider fixes in documentation and code

PR #14368, opened 2024-03-05
partition_manager: Fix nRF54L support

MarekPieta:
Change fixes references to a Kconfig option that does not exist.

Jira: NCSDK-26123

### DIFF
--- a/cmake/partition_manager.cmake
+++ b/cmake/partition_manager.cmake
@@ -206,7 +206,7 @@ elseif (DEFINED CONFIG_SOC_NRF5340_CPUAPP)
   set(otp_size 764)  # 191 * 4
 endif()
 
-if (DEFINED CONFIG_SOC_PLATFORM_NRF54L)
+if (DEFINED CONFIG_SOC_SERIES_NRF54LX)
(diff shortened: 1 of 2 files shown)


MarekPieta:
Hi @carlescufi, @nordicjm, @nvlsianpu, @michalek-no
The `CONFIG_SOC_PLATFORM_*` Kconfig options seem not available in the nRF Connect SDK. Are there plans to add them?
If the options would not be added, we need to use another Kconfig option with similar purpose I think (I would  suggest `CONFIG_SOC_SERIES_NRF54LX` here).

The PR is necessary to run Fast Pair sample (actually nearly any sample with Partition Manager) on nRF54L. Without this commit, settings initialization fails when Partition Manager is enabled - wrong FLASH driver configuration is used.

MarekPieta:
Added `DNM` label to ensure PR is not merged before we confirm it's change in good direction

michalek-no (approved):

nvlsianpu (approved):

SeppoTakalo (approved):

PR #14374, opened 2024-03-05
applications: nrf5340_audio: Include hci_ipc conf file for release builds

alexsven:
- Fix missing conf file for hci_ipc when building release


rick1082 (approved):

koffes (approved):
Approved, but we should also consider renaming the ordinary file to *_debug*

SeppoTakalo (approved):

PR #14375, opened 2024-03-05
subsy/bootloader/bl_boot: uninitialized UARTE pins

nvlsianpu:
These pins might be source of the current leakage, which actually is observed on PCA 10095 2.0.1

Added procedure for setting these pins to default state.

Ref.: NCSDK-26253


LuDuda (approved):
This is great and works! Thank you for you prompt help 🚀 

Should we consider adding a known-issue for previous releases?

nvlsianpu:
> Should we consider adding a known-issue for previous releases?

Yes

SeppoTakalo (approved):

PR #14379, opened 2024-03-05
doc: nrf_security: Add RSA key size support table

mswarowsky:
Extend the table with RSA key sizes to list the support by different crypto  backends

Ref: NCSDK-22536


mia-ko (commented):
Not sure if this doc update should be added to the changelog? Otherwise, LGTM :)

mia-ko (approved):

Vge0rge (approved):

PR #14406, opened 2024-03-06
Mesh update sample configs

akredalen:


subsys/bluetooth/mesh/Kconfig
```diff
@@ -51,7 +51,7 @@ endif
 config BT_MESH_NLC_PERF_CONF
 	bool "Enable NLC performance configurations"
 	select BT_MESH_COMP_PAGE_2
-	select BT_MESH_MODEL_EXTENSIONS
+	imply BT_MESH_MODEL_EXTENSIONS
```
ludvigsj, line 54: Hm. If this is not a "this is always required for any node claiming NLC compliance" (which it doesn't look like, seeing as it is made optional now with the `imply`), I don't think it should be here at all, but in the individual samples that want to use it. I general, I like KConfig to be explicit and not magical with options being `imply`ed all over the place, it makes it hard to track what options are actually enabled or disabled after all the config is parsed.

ludvigsj (changes requested):

samples/bluetooth/mesh/light_dimmer/prj.conf
```diff
@@ -49,10 +48,9 @@ CONFIG_BT_MESH_GATT_PROXY=y
 CONFIG_BT_MESH_PROXY_USE_DEVICE_NAME=y
 CONFIG_BT_MESH_DK_PROV=y
 CONFIG_BT_MESH_NLC_PERF_CONF=y
-# Disable model extension API and set list size to 0 as the models in this sample don't have any
-# relation between each other.
+# Disable model extension API as the models in this sample don't have any relation between
+# each other.
```
PavelVPV, line 52: `CONFIG_BT_MESH_MODEL_EXTENSION_LIST_SIZE` defaults to 10. I think `CONFIG_BT_MESH_MODEL_EXTENSION_LIST_SIZE=0` should stay.

subsys/bluetooth/mesh/Kconfig
ludvigsj: If we want to `imply` this anywhere, it should be under `BT_MESH_COMP_PAGE_1` in Zephyr. Not here.

PavelVPV: Hmm, makes sense to me.

PavelVPV: > Hmm, makes sense to me.

This:
> I general, I like KConfig to be explicit and not magical with options being implyed all over the place, it makes it hard to track what options are actually enabled or disabled after all the config is parsed.

samples/bluetooth/mesh/light_dimmer/prj.conf
akredalen: It defaults to 0 if `!BT_MESH_MODEL_EXTENSIONS`?

subsys/bluetooth/mesh/Kconfig
akredalen: Adding `BT_MESH_MODEL_EXTENSIONS=y` to the relevant samples then. Should it be set to `n` by default? (since its value is checked in BT_MESH_MODEL_EXTENSION_LIST_SIZE)

ludvigsj:
General thought: Why does CDP1 have a KConfig option at all, if it's always required by the Mesh spec? Not for this PR, but I think I'd like to remove the option entirely.

samples/bluetooth/mesh/light_ctrl/prj.conf
```diff
@@ -50,6 +49,7 @@ CONFIG_BT_MESH_GATT_PROXY=y
 CONFIG_BT_MESH_ADV_EXT_GATT_SEPARATE=y
 CONFIG_BT_MESH_DK_PROV=y
 CONFIG_BT_MESH_NLC_PERF_CONF=y
+CONFIG_BT_MESH_MODEL_EXTENSIONS=y
 CONFIG_BT_MESH_NLC_PERF_LIGHTNESS_CTRL=y
```
ludvigsj, line 53: Extreme nitpick, but:

```suggestion
CONFIG_BT_MESH_NLC_PERF_LIGHTNESS_CTRL=y
CONFIG_BT_MESH_MODEL_EXTENSIONS=y
```

ludvigsj (approved):

PavelVPV:
> General thought: Why does CDP1 have a KConfig option at all, if it's always required by the Mesh spec? Not for this PR, but I think I'd like to remove the option entirely.

It allows to build 1.0.1 device.

ludvigsj:
> Might change-log be required now? (after the latest change based on discussions)

Hm, good question. I guess people who depended upon the `select` will need to add it to their prj.conf now. I'd say yes, add something simple like "`CONFIG_BT_MESH_NLC_PERF` no longer selects `CONFIG_BT_MESH_MODEL_EXTENSIONS`", people will know what to do.

mia-ko (approved):

ludvigsj (approved):

PR #14412, opened 2024-03-06
samples: caf: Add nrf54h20pdk configuration

pdunaj:
Add configuration for nRF54H20PDK.

### DIFF
--- a/samples/caf/boards/nrf54h20pdk_nrf54h20_cpuapp.overlay
+++ b/samples/caf/boards/nrf54h20pdk_nrf54h20_cpuapp.overlay
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+ /{
+	/delete-node/ leds;
+
+	leds0 {
+		compatible = "gpio-leds";
+		led0: led_0 {
+			gpios = <&gpio9 0 GPIO_ACTIVE_HIGH>;
+			label = "Green LED 0";
+		};
+	};
+
+	leds1 {
+		compatible = "gpio-leds";
+		led1: led_1 {
+			gpios = <&gpio9 1 GPIO_ACTIVE_HIGH>;
+			label = "Green LED 1";
+		};
+	};
+
+	leds2 {
+		compatible = "gpio-leds";
+		led2: led_2 {
+			gpios = <&gpio9 2 GPIO_ACTIVE_HIGH>;
+			label = "Green LED 2";
+		};
+	};
+
+	leds3 {
+		compatible = "gpio-leds";
+		led3: led_3 {
+			gpios = <&gpio9 3 GPIO_ACTIVE_HIGH>;
+			label = "Green LED 3";
+		};
+	};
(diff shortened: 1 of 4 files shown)


rakons (approved):

jfischer-no:
@pdunaj please get some more +1

samples/caf/boards/nrf54h20pdk_nrf54h20_cpuapp.overlay
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+ /{
+	/delete-node/ leds;
+
+	leds0 {
+		compatible = "gpio-leds";
+		led0: led_0 {
+			gpios = < &gpio9 0 GPIO_ACTIVE_HIGH >;
```
gmarull, line 13:
```suggestion
			gpios = <&gpio9 0 GPIO_ACTIVE_HIGH>;
```

in all places

samples/caf/boards/nrf54h20pdk_nrf54h20_cpuapp.overlay
```diff
@@ -0,0 +1,41 @@
...
+		};
+	};
+
+	leds3 {
+		compatible = "gpio-leds";
+		led3: led_3 {
+			gpios = < &gpio9 3 GPIO_ACTIVE_HIGH >;
+			label = "Green LED 3";
+		};
+	};
```
gmarull, line 40: question: why is this repeating information already upstream? https://github.com/zephyrproject-rtos/zephyr/blob/main/boards/nordic/nrf54h20pdk/nrf54h20pdk_nrf54h20_cpuapp.dts

There's no need to have one node per led, `gpio-leds` doesn't work like that for cases like this.

gmarull (changes requested):

samples/caf/boards/nrf54h20pdk_nrf54h20_cpuapp.overlay
pdunaj: will do

samples/caf/boards/nrf54h20pdk_nrf54h20_cpuapp.overlay
pdunaj: Our application framework use dts leds grouping to handle physical LEDs into logical (RGB or monochromatic) LEDs.
Upstream does not match what we want.

gmarull: Then this needs to be fixed upstream, it makes no sense to keep copies of stuff like this downstream.

KAGA164 (approved):

samples/caf/boards/nrf54h20pdk_nrf54h20_cpuapp.overlay
pdunaj: This is application specific but to be honest I don't mind if DKs follow the configuration that we set up.

pdunaj:
the first force push is a rebase, the second one is requested formatting fix

shanthanordic (approved):

PR #14413, opened 2024-03-06
tfm: cmake: Link with mbedcrypto_common and tfm_sprt

SebastianBoe:
For some strange configurations it has been observed that the sources in TF-M's platform_crypto_keys library use TF-M headers instead of Oberon headers as they should.

It's also observed that they are missing the include path to tfm_sp_log.h.

Link this library with the two CMake libraries that provide the correct include directories.


mswarowsky (approved):

Vge0rge (approved):

SebastianBoe:
already applied in a different PR

PR #13545, opened 2024-01-05
samples: net: cloud: Increase AT_HOST heap/stack 

glarsennordic:
Make sure AT_HOST heap and stack are at least 2048 for all nrf_cloud
samples.

NCSDK-22966

PR #14422, opened 2024-03-06
net: lib: nrf_cloud: fix build issue with FOTA poll support

jayteemo:
The nrf_cloud_rest_fota sample failed to build when JITP
was enabled.

Update the nrf_cloud_fota_poll module to allow for the
pending job type to be accessed by the application.
Moved fota_download_init into nrf_cloud_fota_poll_init.
Added function comments and moved into nrf_cloud_fota_poll.h

Increased AT host stack size for the nrf_cloud_rest_fota
sample.

Also, corrected changelog entries from https://github.com/nrfconnect/sdk-nrf/pull/13545.

IRIS-8444

### DIFF
--- a/include/net/nrf_cloud_fota_poll.h
+++ b/include/net/nrf_cloud_fota_poll.h
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef NRF_CLOUD_FOTA_POLL_H_
+#define NRF_CLOUD_FOTA_POLL_H_
+
+/** @file nrf_cloud_fota_poll.h
+ * @brief Module to provide nRF Cloud FOTA assistance for applications
+ *        connecting to nRF Cloud using REST or CoAP.
+ */
+
+#include <net/nrf_cloud_rest.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** @defgroup nrf_cloud_fota_poll nRF Cloud FOTA poll
+ * @{
+ */
+
+enum nrf_cloud_fota_reboot_status {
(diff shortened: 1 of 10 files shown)


plskeggs (approved):

include/net/nrf_cloud_fota_poll.h
```diff
@@ -0,0 +1,94 @@
...
+ *                 NRF_CLOUD_FOTA_TYPE__INVALID indicates no pending FOTA job.
+ */
+int nrf_cloud_fota_poll_process_pending(struct nrf_cloud_fota_poll_ctx *ctx);
+
+/**
+ * @brief Perform the following FOTA tasks:
+ *        Report the status of an in progress FOTA job.
+ *        Check for a queued FOTA job.
+ *        Execute FOTA job.
+ *        Save status and request reboot.
```
peknis, line 76: Want to have these as a bullet list? You can use - or *

peknis (approved):

samples/cellular/nrf_cloud_multi_service/src/cloud_connection.c
```diff
@@ -527,11 +527,6 @@ static int setup_cloud(void)
 		LOG_ERR("Error initializing FOTA: %d", err);
 		return err;
 	}
-	err = coap_fota_begin();
```
glarsennordic, line 530: This will cause `nrf_cloud_fota_poll_process_pending` to be called before `nrf_cloud_init` won't it?

I don't remember why that is a problem, I am pretty sure it has caused us issues in the past

jayteemo: this is for CoAP and REST FOTA.
nrf_cloud_init is MQTT.

glarsennordic: Ah, you're right, my bad

glarsennordic (approved):

SeppoTakalo (approved):

PR #14423, opened 2024-03-06
manifest: sdk-hostap: Pull error logs for security

krish2718:
These logs are crucial to debug and fix any configuration issues that affect Wi-Fi security handshakes.

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -112,7 +112,7 @@ manifest:
     - name: hostap
       repo-path: sdk-hostap
       path: modules/lib/hostap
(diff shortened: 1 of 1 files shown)


sachinthegreen (approved):

SeppoTakalo (changes requested):
The proposed changes in just `[nrf noup]` commits that add more debug logs.

I don't understand how this justifies as a bugfix for RC.

If this is improvement, it should be first proposed upstream.

krish2718:
Will revisit after submitting upstream.

PR #14432, opened 2024-03-07
doc: cc3xx: Add limitation of rng driver

mswarowsky:
The cryptocell driver can only produce 1024 random bytes per request, so we should list tis limitation

Ref: NCSDK-24840


endre-nordic (approved):
LGTM

mia-ko (approved):

PR #14440, opened 2024-03-07
bluetooth: controller: remove usage of sdc_soc.h in crypto.c

olivier-le-sage:
It has been replaced by mpsl_ecb.h

### DIFF
--- a/subsys/bluetooth/controller/crypto.c
+++ b/subsys/bluetooth/controller/crypto.c
@@ -9,7 +9,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <soc.h>
-#include <sdc_soc.h>
+#include <mpsl_ecb.h>
 #include <zephyr/drivers/entropy.h>
 
 #include "nrf_errno.h"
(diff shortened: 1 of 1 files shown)


subsys/bluetooth/controller/crypto.c
```diff
@@ -36,25 +36,19 @@ int bt_encrypt_le(const uint8_t key[BT_ECB_BLOCK_SIZE],
...
-	uint8_t plaintext_le[BT_ECB_BLOCK_SIZE];
-	uint8_t enc_data_le[BT_ECB_BLOCK_SIZE];
-
 	LOG_HEXDUMP_DBG(key, BT_ECB_BLOCK_SIZE, "key");
 	LOG_HEXDUMP_DBG(plaintext, BT_ECB_BLOCK_SIZE, "plaintext");
 
-	sys_memcpy_swap(key_le, key, BT_ECB_BLOCK_SIZE);
-	sys_memcpy_swap(plaintext_le, plaintext, BT_ECB_BLOCK_SIZE);
-
 	int errcode = MULTITHREADING_LOCK_ACQUIRE();
```
seko-nordic, line 42: do we need this?

olivier-le-sage: I think the case that this guards against is the one where this thread and mpsl have different priorities. Without this mutex, if this is running in a preemptive thread for instance, the MPSL thread can preempt it.

I don't think this is different from being preempted by interrupts really. Ideally we would have a test with all three priorities and remove this

rugeGerritsen (approved):
Can we use plaintext without the memcpy if we make `mpsl_ecb_block_encrypt_extended()` take in a const cleartext?

olivier-le-sage:
> Can we use plaintext without the memcpy if we make `mpsl_ecb_block_encrypt_extended()` take in a const cleartext?

Already discussed offline with Sergey, I'm going to add a flag to remove (or rather, move) the memcpy(). The challenge is that the hardware isn't able to read directly from NVM.

seko-nordic (approved):

PR #14444, opened 2024-03-07
net: Add Wi-Fi provisioning library and sample

simensrostad:
Add softap support, will be added later:
 - library tests
 - Documentation
 - Integration with upstream HTTP server library https://github.com/zephyrproject-rtos/zephyr/pull/64465


subsys/net/lib/wifi_provision/wifi_provision.c
```diff
@@ -0,0 +1,995 @@
...
+		ret = smf_run_state(SMF_CTX(&state_object));
+		if (ret) {
+			LOG_ERR("smf_run_state, error: %d", ret);
+			notify_app(WIFI_PROVISION_EVT_FATAL_ERROR);
+			return;
+		}
+	}
+}
+
+K_THREAD_DEFINE(wifi_provision_task_id, 8192,
```
rlubos, line 994: This stack seems quite heavy. Though personally, I'd reconsider if we need a state machine at all, it seems that all those actions (scan (if needed), AP enable,  DHCP enable, HTTP server processing) could just take place sequentially, w/o a fancy state machine behind it. Separate thread here is IMO expensive choice.

subsys/net/lib/wifi_provision/wifi_provision.c
```diff
@@ -0,0 +1,995 @@
...
+	int ret;
+	struct net_if *iface = net_if_get_first_wifi();
+	struct wifi_scan_params params = { 0 };
+
+	LOG_DBG("Scanning for Wi-Fi networks...");
+
+	/* Clear scan result buffer */
+	memset(scan_result_buffer, 0, sizeof(scan_result_buffer));
+
+	ret = net_mgmt(NET_REQUEST_WIFI_SCAN, iface, &params, sizeof(params));
```
rlubos, line 309: Side question - what's the intent of scanning in the provisioning library?

subsys/net/lib/wifi_provision/wifi_provision.c
```diff
@@ -0,0 +1,995 @@
...
+	parser_settings.on_url = on_url;
+
+	/* Start processing incoming IPv4 clients */
+	ret = process_tcp4();
+	if (ret < 0) {
+		LOG_ERR("Failed to start TCP server %d", ret);
+		return ret;
+	}
+
+	k_sem_take(&wifi_provision_sem, K_FOREVER);
```
rlubos, line 954: Would likely make sense to remove the registered net_mgm handler before leaving, it should no longer need to process events (`net_mgmt_del_event_callback()`).

subsys/net/lib/wifi_provision/wifi_provision.c
```diff
@@ -0,0 +1,995 @@
...
+static struct s_object {
+	/* This must be first */
+	struct smf_ctx ctx;
+
+	/* State machine variables */
+	enum module_event event_next;
+
+	/* Provisioning parameters */
+	char ssid[50];
+	char psk[50];
```
rlubos, line 115: ssid/psk have predefined maximum lengths in `wifi.h` (`WIFI_SSID_MAX_LEN`/`WIFI_PSK_MAX_LEN`), could use them instead of magic numbers.

samples/wifi/wifi_provision/src/main.c
```diff
@@ -0,0 +1,277 @@
...
+	if (ret) {
+		LOG_ERR("Failed to disable power save, error: %d", ret);
+		FATAL_ERROR();
+		return ret;
+	}
+
+	return 0;
+}
+
+int socket_setup(void)
```
balaji-nordic, line 170:
```suggestion
static int socket_setup(void)
```

samples/wifi/wifi_provision/src/main.c
```diff
@@ -0,0 +1,277 @@
...
+		FATAL_ERROR();
+		return ret;
+	}
+
+	/* Setup the socket for the mDNS-SD service to work after provisioning.
+	 * The registered service is the same as configured with the CONFIG_NET_HOSTNAME option.
+	 *
+	 * This is a way for the provisioning client to find the device on the network and confirm
+	 * that it has been succsessfully provisioned to the network.
+	 */
```
balaji-nordic, line 244:
```suggestion
	/* Setup the socket for the mDNS-SD service to work after provisioning.
	 * The registered service is the same as configured with the CONFIG_NET_HOSTNAME option.
	 *
	 * This is a way for the provisioning client to find the device on the network and confirm
	 * that it has been successfully provisioned to the network.
         * This is kept in this sample for demonstration purposes only.
	 * Ideally this should be replaced by application code that uses networking for its specific use case.
	 */
```

subsys/net/lib/wifi_provision/wifi_provision.c
simensrostad: A state machine is always nice. Making the library more robust and easier to maintain. Things tend to get messy without some sort of structure keeping order. You also protect against out-of-order behavior without having flags everywhere to check if some functionality is triggered when it shouldn't.

I guess an argument can be made for that _nothing else should happen when provisioning is ongoing_ but I don't want to make that assumption.

I initially did something much simpler but experience a lot of issues that were typically solved by scheduling actions in a separate wq. Instead of calling the respective functionality in the net mgmt handlers. Not sure if it was a prio thing or not.

With the additional wq I found the overhead of the current structure acceptable as memory wise it was not far away from what I already had.

I'm ofc open to optimizing the library if it doesn't impose too many drawbacks in terms of stability and maintainability.

samples/wifi/wifi_provision/prj.conf
```diff
@@ -0,0 +1,95 @@
...
+
+# General
+CONFIG_LOG=y
+CONFIG_LOG_BUFFER_SIZE=4096
+CONFIG_ASSERT=y
+CONFIG_ASSERT_VERBOSE=n
+
+## Temporarily enable FS support so that linking succeeds
+CONFIG_FILE_SYSTEM=y
+CONFIG_MAIN_STACK_SIZE=16384
```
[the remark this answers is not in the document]
simensrostad: Ill make sure to go over the stack sizes and optimize them, I haven't done that yet.

subsys/net/lib/wifi_provision/certs/server_certificate.pem
```diff
@@ -0,0 +1,12 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
simensrostad: hmm, not sure. I guess its more convenient if the application owns the certificates. I can create a Kconfig that allows the user to specify an app folder that the credentials needs to be included in.

'CONFIG_MQTT_HELPER_CERTIFICATES_FOLDER'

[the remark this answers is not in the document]
jukkar: Better is to application load the certificates to store, and then just load them when creating the socket. Then you could just pass the tag list to the library instead of the certificate folder etc.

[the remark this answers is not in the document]
simensrostad: That means that all samples that include the wifi provision library needs to also include logic to convert and load certs. Instead of the library doing this.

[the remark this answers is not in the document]
rlubos: I don't think there's anything wrong in expecting that application deals with its own sensitive data (keys/certs).

But on the other hand, the MQTT helper library approach seems as a good tradeoff here, as it still allows the app to handle certificates while offering some convenience functions to provision certs on the app behalf, IMO we could go that way.

My major concern here was that the credentials were hardcoded within the library, which shouldn't be the case no matter what.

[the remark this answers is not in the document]
simensrostad: Yeah, in a production scenario creds are provisioned as a separate stage to protected storage and referenced using a sec tag in the actual app. I plan to add docs for this, there is also a cmake warning that warns users about the cert use in its current state.

subsys/net/lib/wifi_provision/wifi_provision.c
simensrostad: Apparently on iOS application doesnt have access to APIs that lists available wifi networks or something like that.
@roshanrajaratnam can you elaborate?

Also it might be different what the device sees and the client that wants to provisioned it, although I suspect this is quite the rare case.

subsys/net/lib/wifi_provision/certs/server_certificate.pem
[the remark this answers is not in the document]
simensrostad: So it makes sense to be able to disable automatic cert provisioning actually.

samples/wifi/wifi_provision/prj.conf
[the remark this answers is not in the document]
simensrostad: Reduced to 6144

subsys/net/lib/wifi_provision/wifi_provision.c
```diff
@@ -0,0 +1,995 @@
...
+
+static int process_tcp(sa_family_t family)
+{
+	int client;
+	struct sockaddr_in6 client_addr;
+	socklen_t client_addr_len = sizeof(client_addr);
+	char addr_str[INET6_ADDRSTRLEN];
+	int received;
+	int ret = 0;
+	char buf[2048];
```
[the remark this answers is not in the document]
simensrostad: Added kconfig to be able to configure this size

subsys/net/lib/wifi_provision/wifi_provision.c
simensrostad: didnt need these, removed

subsys/net/lib/wifi_provision/wifi_provision.c
roshanrajaratnam: This is more or less to identify what Wi-Fi networks are visible to the DK and not necessarily what's supported by the Android/iOS aps.

rlubos: Should this be a part of the library though? It doesn't seem to be essential to the provisioning process, and I imagine getting console logs isn't really useful for actual products.

subsys/net/lib/wifi_provision/certs/server_certificate.pem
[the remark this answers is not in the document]
simensrostad: Added options:

CONFIG_WIFI_PROVISION_SERVER_CERTIFICATE_REGISTER
CONFIG_WIFI_PROVISION_SERVER_CERTIFICATE_FOLDER

subsys/net/lib/wifi_provision/proto/common.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+
+enum Band {
+  BAND_UNSPECIFIED = 0;
```
roshanrajaratnam, line 4: This needs to be renamed to BAND_ANY based on the Wifi provisioner over BLE. In  addition SoftAP adds support to 6_GHZ so may be this needs to be reflected on the BLE example as well?

subsys/net/lib/wifi_provision/proto/common.proto
```diff
@@ -0,0 +1,39 @@
...
+  WPA_PSK = 2;
+  WPA2_PSK = 3;
+  WPA_WPA2_PSK = 4;
+  WPA2_ENTERPRISE = 5;
+  WPA3_PSK = 6;
+}
+
+message WifiInfo {
+  string ssid = 1;
+  Band band = 3;
```
roshanrajaratnam, line 22: seems like we are missing `bssid` here

subsys/net/lib/wifi_provision/proto/common.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+
+enum Band {
+  BAND_UNSPECIFIED = 0;
+  BAND_2_4_GHZ = 1;
+  BAND_5_GHZ = 2;
+  BAND_6_GHZ = 3;
```
roshanrajaratnam, line 7: I see that SoftAp adds support for 6 ghz, but should this be reflected on the ble example as well?

subsys/net/lib/wifi_provision/wifi_provision.c
```diff
@@ -0,0 +1,1058 @@
...
+		LOG_ERR("Failed to create IPv4 socket %d", ret);
+		return ret;
+	}
+
+	LOG_DBG("Waiting for IPv4 HTTP connections on port %d", CONFIG_WIFI_PROVISION_TCP_PORT);
+
+	while (true) {
+
+		/* Process incoming IPv4 clients */
+		ret = process_tcp(AF_INET);
```
balaji-nordic, line 821: I believe we should handle the ECONNRESET as a special case and continue staying in the loop. I managed to crash the provisioning sample by sending a bad request from httpie desktop client. It was because of the ECONNRESET returned by `process_tcp()` function.

subsys/net/lib/wifi_provision/proto/common.proto
simensrostad: Probably yes.

subsys/net/lib/wifi_provision/proto/common.proto
```diff
@@ -0,0 +1,39 @@
...
+message WifiInfo {
+  string ssid = 1;
+  Band band = 3;
+  uint32 channel = 4;
+  AuthMode authMode = 5;
+}
+
+message WifiConfig {
+  WifiInfo info = 1;
+  string passphrase = 2;
```
roshanrajaratnam, line 29: This passphrase must be nullable in the case of an open network.

subsys/net/lib/wifi_provision/proto/common.proto
```diff
@@ -0,0 +1,39 @@
...
+  WPA_WPA2_PSK = 4;
+  WPA2_ENTERPRISE = 5;
+  WPA3_PSK = 6;
+}
+
+message WifiInfo {
+  string ssid = 1;
+  Band band = 3;
+  uint32 channel = 4;
+  AuthMode authMode = 5;
```
roshanrajaratnam, line 24: AuthMode is also optional

subsys/net/lib/wifi_provision/proto/common.proto
simensrostad: Im in the process of updating the schema to be _identical_ to the BLE counterpart :+1:

subsys/net/lib/wifi_provision/proto/common.proto
simensrostad: Removed 6ghz entry

samples/wifi/wifi_provision/prj.conf
```diff
@@ -0,0 +1,98 @@
...
+CONFIG_NET_L2_ETHERNET=y
+CONFIG_WIFI_CREDENTIALS=y
+
+# General
+CONFIG_LOG=y
+CONFIG_LOG_BUFFER_SIZE=4096
+CONFIG_ASSERT=y
+CONFIG_ASSERT_VERBOSE=n
+
+## Temporarily enable FS support so that linking succeeds
```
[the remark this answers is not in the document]
simensrostad: Yes

subsys/net/lib/wifi_provision/wifi_provision.c
```diff
@@ -0,0 +1,1058 @@
...
+/* Forward declarations */
+static int ap_enable(void);
+
+/* Internal variables */
+static ScanResults scan = ScanResults_init_zero;
+static uint8_t scan_result_buffer[CONFIG_WIFI_PROVISION_SCAN_RESULT_BUFFER_SIZE];
+static size_t scan_result_buffer_len;
+static const struct smf_state state[];
+static struct http_parser_settings parser_settings;
+static char linkaddr_string[sizeof("XXXXXXXXXXXX")];
```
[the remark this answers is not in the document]
simensrostad: Not sure, in that case we would need to multiply with 2 and add 1 to make space for null termination.
Its a bit more direct like this.

balaji-nordic (approved):
Just a couple of minor comments. Looks good :)

gregersrygg (approved):

rlubos (approved):

samples/wifi/ble_provision/README.rst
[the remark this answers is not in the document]
divipillai: We could have the sample name as "Wi-Fi: Bluetooth LE provision sample". Apologies, I forgot to add the LE (low energy with the Bluetooth term).

subsys/net/lib/softap_wifi_provision/softap_wifi_provision.c
```diff
@@ -0,0 +1,1187 @@
...
+		"ssid: %.*s, bssid: %s, passphrase: xxxxxx, sectype: %d, channel: %d, band: %d",
+		creds.wifi.ssid.size, creds.wifi.ssid.bytes,
+		bssid_str,
+		creds.wifi.auth,
+		creds.wifi.channel,
+		creds.wifi.band);
+
+	enum wifi_security_type sec_type =
+		creds.wifi.auth == AuthMode_WPA_WPA2_PSK ? WIFI_SECURITY_TYPE_PSK :
+		creds.wifi.auth == AuthMode_WPA2_PSK ? WIFI_SECURITY_TYPE_PSK_SHA256 :
```
krish2718, line 412: WPA2-PSK and WPA2-PSK-SHA256 are two different options, so, might need to rename it accordingly

WPA_WPA2_PSK->WPA2_PSK
WPA2_PSK->WPA2_PSK_SHA256

samples/wifi/provisioning/softap/src/main.c
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	LOG_INF("Network interface brought up");
+
+	ret = softap_wifi_provision_start();
+	switch (ret) {
+	case 0:
+		provisioning_completed = true;
+		break;
+	case -EALREADY:
```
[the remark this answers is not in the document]
simensrostad: Not really, if the device has already been provisioned after boot (stored credentials). There is no point of disabling psm for a while because there will be no open the wifiprov SD service has been registered too.

samples/wifi/ble_provision/README.rst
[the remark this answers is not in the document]
simensrostad: I think @divipillai is doing some updates here atm?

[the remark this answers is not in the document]
divipillai: yeah, will fix this in my commit.

doc/nrf/samples/wifi_provisioning.rst
```diff
@@ -0,0 +1,13 @@
+.. _wifi_provisioning_samples:
+
+Wi-Fi: provisioning samples
+###########################
+
+The following samples are used to provision Wi-Fi® credentials to a Wi-Fi enabled device (nRF70 Series).
```
ioannisg, line 6:
```suggestion
The below samples demonstrate ways to provision Wi-Fi® credentials to a Wi-Fi-capable Nordic device.
```

include/net/softap_wifi_provision.h
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/**
+ * @file softap_wifi_provision.h
+ * @brief Header file for the softAP Wi-Fi Provision Library.
```
ioannisg, line 9: SoftAP

include/net/softap_wifi_provision.h
```diff
@@ -0,0 +1,106 @@
...
+	SOFTAP_WIFI_PROVISION_EVT_STARTED,
+
+	/** A client (station) has connected to the softAP Wi-Fi network. */
+	SOFTAP_WIFI_PROVISION_EVT_CLIENT_CONNECTED,
+
+	/** A client (station) has disconnected from the softAP Wi-Fi network. */
+	SOFTAP_WIFI_PROVISION_EVT_CLIENT_DISCONNECTED,
+
+	/** Wi-Fi credentials received from the client. */
+	SOFTAP_WIFI_PROVISION_EVT_CREDENTIALS_RECEIVED,
```
ioannisg, line 39: Not so important, but it is more logical to me that the CREDENTIALS RECEIVED state precedes the CONNECTED, or DISCONNECTED states

include/net/softap_wifi_provision.h
```diff
@@ -0,0 +1,106 @@
...
+
+	/** The provisioning process has completed. */
+	SOFTAP_WIFI_PROVISION_EVT_COMPLETED,
+
+	/** Wi-Fi credentials deleted, a reboot is required to enter an unprovisioned state. */
+	SOFTAP_WIFI_PROVISION_EVT_RESET_REBOOT_REQUEST,
```
ioannisg, line 45: Credentials-deleted is not reflected in the enum's name (neither does the resulting state: ie. the device is becoming un-provisioned). Reboot request is essentially the outcome of the device becoming un-provisioned and could change in the future. Again, that's reflecting only the enum naming, so not so important.

include/net/softap_wifi_provision.h
```diff
@@ -0,0 +1,106 @@
...
+
+	/** The provisioning process has failed, this is an irrecoverable error. */
```
ioannisg, line 47:
```suggestion
	/** The provisioning process has failed, this is an unrecoverable error. */
```

samples/wifi/provisioning/ble/README.rst
```diff
@@ -1,15 +1,17 @@
 .. _wifi_provisioning:
+.. _ble_wifi_provision:
 
-Wi-Fi: Provisioning Service
-###########################
+Wi-Fi: Bluetooth LE provision
```
ioannisg, line 4:
```suggestion
Wi-Fi: Bluetooth LE based provisioning
```

samples/wifi/provisioning/softap/README.rst
```diff
@@ -0,0 +1,51 @@
+.. _softap_wifi_provision_sample:
+
+Wi-Fi: SoftAP provision
```
ioannisg, line 3:
```suggestion
Wi-Fi: SoftAP based provisioning
```

samples/wifi/provisioning/softap/README.rst
```diff
@@ -0,0 +1,51 @@
...
+#######################
+
+.. contents::
+   :local:
+   :depth: 2
+
+This sample demonstrates how to use the :ref:`lib_softap_wifi_provision` library to provision credentials to a Wi-Fi® device (nRF70 Series).
```
ioannisg, line 10: Can we remove references to nRF70 Series? I believe the library and the sample are agnostic to the underlying hardware. I vote for Nordic Wi-Fi device, instead.

samples/wifi/provisioning/softap/README.rst
```diff
@@ -0,0 +1,51 @@
...
+|config|
+
+Configuration files
+===================
+
+The sample includes the following pre-configured configuration file for the development kits that are supported:
+
+* :file:`prj.conf` - Configuration file for all build targets.
+
+Building and running
```
ioannisg, line 37: Does the sample require a corresponding mobile app counterpart? how do we suggest developers testing it? It would be nice to add a couple of sentences about this here.

samples/wifi/provisioning/softap/prj.conf
```diff
@@ -0,0 +1,98 @@
...
+CONFIG_WIFI_CREDENTIALS=y
+
+# General
+CONFIG_LOG=y
+CONFIG_LOG_MODE_IMMEDIATE=y
+CONFIG_ASSERT=y
+CONFIG_ASSERT_VERBOSE=n
+
+# Link-time optimizations
+CONFIG_LTO=y
```
ioannisg, line 33: Do we have reasons not to have the LTO generically enabled? E.g. make it default for Networking/Wi-Fi samples on nRF7002-enabled boards? Is it an agreed convention to only enable this at the top (user) configuration layer, i.e. in samples' prj.conf?

samples/wifi/provisioning/softap/prj.conf
```diff
@@ -0,0 +1,98 @@
...
+CONFIG_NET_HOSTNAME="wifiprov"
+# Disable POSIX uname support to suppress build warning that hostname is too long.
+CONFIG_POSIX_UNAME=n
+
+# DHCPv4 client and server
+CONFIG_NET_DHCPV4=y
+CONFIG_NET_DHCPV4_SERVER=y
+
+# MBed TLS heap size, required by WPA supplicant and MBed TLS.
+CONFIG_MBEDTLS_HEAP_SIZE=81920
```
ioannisg, line 89: Do we really need that much? Are we using TLS connections in the sample? No objection, of course, if we do have the memory. Just wondering.

ioannisg (commented):
I left some comments and minor questions @simensrostad , otherwise this is fine by me.

ioannisg:
@krish2718 please ensure the Wi-Fi team has looked at and gave their 'go' for this PR :)

include/net/softap_wifi_provision.h
simensrostad: ^^ @divipillai

include/net/softap_wifi_provision.h
simensrostad: Ill see if I can find a bit more suiting name.

samples/wifi/provisioning/ble/README.rst
simensrostad: ^^ @divipillai

samples/wifi/provisioning/softap/README.rst
simensrostad: ^^ @divipillai

samples/wifi/provisioning/softap/README.rst
simensrostad: Sure, I dont mind this. ^^ @divipillai

samples/wifi/provisioning/softap/README.rst
simensrostad: I plan to extend the docs after the initial commit has beeen made. Good idea to mention this.  We are not dependent on the mobile app.

samples/wifi/provisioning/softap/prj.conf
simensrostad: I think we can enable LTO for all net and wifi samples once we know that it is supported. IMO this is separate effort but worth doing.

samples/wifi/provisioning/softap/prj.conf
simensrostad: Last time I checked, yes. But Ill have another go just to be sure. I Don't remember if I tested again after starting to use ECDSA server cert.

samples/wifi/provisioning/softap/README.rst
divipillai: Agree with the above comment. I have updated the sentence as : _This sample demonstrates how to use the :ref:`lib_softap_wifi_provision` library to provision credentials to Nordic Semiconductor's Wi-Fi® device._

samples/wifi/provisioning/softap/prj.conf
simensrostad: I suggest we discuss this further in:
https://github.com/nrfconnect/sdk-nrf/pull/14558

samples/wifi/provisioning/softap/README.rst
ioannisg: That's ok

subsys/net/lib/softap_wifi_provision/softap_wifi_provision.c
simensrostad: The naming schemas has been adopted from the BLE wifi prov lib/proto. We decided to keep them as similar as possible.
The naming is a bit bland yes.

krish2718 (approved):

divipillai (approved):
Discussed offline and more doc contents will be added in a separate PR, also covering the testing steps for the SoftAP based provision sample.

ioannisg (approved):

nordicjm (changes requested):
Needs to wait for sysbuild pr to be merged and updated to use sysbuild

simensrostad:
> Needs to wait for sysbuild pr to be merged and updated to use sysbuild

Updated sample.yaml

simensrostad:
@nordicjm Can this go in now?

PR #14445, opened 2024-03-07
manifest: pull hostap image size optimizaton

markaj-nordic:
This pulls in a patch that strips off helper
strings from the wpa commands that occupy FLASH,
but are not really used in some configurations.

It saves 730 bytes for Matter that will allow to unblock the nightly CI plans (Matter Bridge image stopped fitting into FLASH slot).

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -112,7 +112,7 @@ manifest:
     - name: hostap
       repo-path: sdk-hostap
       path: modules/lib/hostap
-      revision: dda5457ad2cfce99e333980c7764c8d480ae4010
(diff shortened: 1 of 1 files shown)


thst-nordic (approved):

thst-nordic:
If this doesn't get merged today we have to add it to `scripts/quarantine_windows_mac.yaml`
before release.

Damian-Nordic (approved):

sachinthegreen (approved):

LuDuda (approved):

PR #14446, opened 2024-03-07
nrf_security: Fix crypto include path issue

SebastianBoe:
Fix crypto include path issue.

See comment in code for details.

Fixes an issue I had when building for the non-secure image.


Vge0rge (approved):

frkv (changes requested):
Please delay taking this in until I have run CI on a proposal for a bugfix

(will remove the change request-statement, depending on first run)

mswarowsky (approved):

SebastianBoe:
already applied by a different PR

PR #14447, opened 2024-03-07
applications: matter: Fix Data Model for Bridge

ArekBalysNordic:
The Data Model is set wrongly for all bridged devices.

This PR repairs a critical bug for Matter Bridge DataModel, found during certification testing.

REF BUG: KRKNWK-18600


applications/matter_bridge/src/bridged_device_types/generic_switch.cpp
```diff
@@ -68,6 +68,8 @@ CHIP_ERROR GenericSwitchDevice::HandleRead(ClusterId clusterId, AttributeId attr
 	switch (clusterId) {
 	case Clusters::Switch::Id:
 		return HandleReadSwitch(attributeId, buffer, maxReadLength);
+	case Clusters::Descriptor::Id:
+		return HandleReadDescriptor(attributeId, buffer, maxReadLength);
```
Damian-Nordic, line 72: Does this work? Descriptor cluster proves `AttributeAccesInterface` that overrides read regular read handlers. I think the right solution is to update `descriptor.cpp` in Matter (in the upstream the fix seems to be already done).

applications/matter_bridge/src/bridged_device_types/onoff_light.cpp
```diff
@@ -46,6 +61,10 @@ constexpr CommandId groupsIncomingCommands[] = {
 	app::Clusters::Groups::Commands::RemoveGroup::Id,
 	app::Clusters::Groups::Commands::RemoveAllGroups::Id,
 	app::Clusters::Groups::Commands::AddGroupIfIdentifying::Id,
+	app::Clusters::Groups::Commands::AddGroupResponse::Id,
```
Damian-Nordic, line 64: Those don't seem like incoming commands.

applications/matter_bridge/src/bridged_device_types/generic_switch.cpp
ArekBalysNordic: Oh, Indeed it works this way. In Matter Bridge, there is no declaration of Descriptor's feature map attribute, so for now, it is enough to add it only, without changing the value because I can see that after enabling it the value read by chiptool is correct.

applications/matter_bridge/src/bridge.zap
```diff
@@ -1,6 +1,6 @@
 {
   "fileFormat": 2,
-  "featureLevel": 99,
+  "featureLevel": 100,
```
LuDuda, line 3: This might be problematic, can you use ZAP tool from this release?

Damian-Nordic (approved):

applications/matter_bridge/src/bridge.zap
ArekBalysNordic: Fixed

ArekBalysNordic:
This PR contains important certification fixes for Matter Bridge.

markaj-nordic (approved):

doublemis1:
Light Switch On Off does not support Identify cluster (client side): which based on the spec i mandatory.
Can we add last this last thing?

doublemis1 (approved):
LGTM!

SeppoTakalo (approved):

PR #14453, opened 2024-03-08
application: serial_lte_modem: Support nRF9131 EK

junqingzou:
Add the overlay files to support nRF9131 EK.
NOTE UART_2 cannot be used as I2C_2 is in use on the EK.

### DIFF
--- a/applications/serial_lte_modem/boards/nrf9131ek_nrf9131_ns.conf
+++ b/applications/serial_lte_modem/boards/nrf9131ek_nrf9131_ns.conf
@@ -1,0 +1,14 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Configuration file for nRF9131EK.
+# This file is merged with prj.conf in the application folder, and options
+# set here will take precedence if they are present in both files.
+
+CONFIG_UART_0_NRF_HW_ASYNC_TIMER=2
+CONFIG_UART_0_NRF_HW_ASYNC=y
+CONFIG_SLM_POWER_PIN=28
+CONFIG_SLM_INDICATE_PIN=0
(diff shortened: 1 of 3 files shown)


applications/serial_lte_modem/boards/nrf9131ek_nrf9131_ns.conf
```diff
@@ -0,0 +1,16 @@
...
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Configuration file for nRF9131EK.
+# This file is merged with prj.conf in the application folder, and options
+# set here will take precedence if they are present in both files.
+
+# Use UART_0 (when working with PC terminal)
```
MarkusLassila, line 11: These comments should be removed as they apply to a scenario where UART2 is used. Which is, as far as I understand, not a scenario with this board.

applications/serial_lte_modem/boards/nrf9131ek_nrf9131_ns.overlay
tomi-font, line 1: I'm pondering whether it would make sense that we use a common overlay file for this. Not sure.

junqingzou: how?

tomi-font: By having the contents of this file be in a generic `*.overlay` file, that others like this one `#include`. @MarkusLassila An opinion on this?

MarkusLassila: You mean that the chosen and uart0 parts from the overlay would be in common file as they are repeated over multiple files? I mean, maybe. It is repetitive to have them multiple times, but it is a bit simpler for someone who is looking at these for the first time if they are duplicated. I do not think that it needs to be solved in this PR though.

MarkusLassila (approved):

applications/serial_lte_modem/boards/nrf9131ek_nrf9131_ns.overlay
tomi-font: Yes. Agreed.

tomi-font (approved):

junqingzou:
@SeppoTakalo pls help to review in time.

tomi-font:
I think that you can ask for this PR to be merged already as it has two code owner approvals.

SeppoTakalo (approved):

PR #1255, opened 2019-10-08
LTE link controller extensions

jtguggedal:
### drivers: lte_lc: Add API to get network registration status

This patch adds an API to get the current network registration
status.
The LTE link controller is changed to use the new parsing
function for registration status internally as well.

---

### drivers: lte_lc: Add API for setting and getting system mode

Adds APIs for getting and setting system mode.
Also adds utility function for checking if a returned AT
response is the same as expected.

---

### drivers: lte_lc: Add API for getting functional mode

Adds API for getting the modem's current functional mode.

PR #406, opened 2019-01-23
lib: gps_sim: Reduce GPS simulator stack usage

jtguggedal:
Updated description:

This patch reduces the GPS simulator stack usage.

PR #14465, opened 2024-03-08
manifest: nrf_security: Fix legacy + PSA crypto regression for WIFI

frkv:
-Pulls in matter PR #406 and nrfxlib PR #1255 as well as Mbed TLS PR #33
 which are dependancies for this
-Done as a single commit for bisectability. Note that the
 entirety of this commit can be reverted depending on state of CI
 and the intent to possible use multiple files for configurations.
 Splitting these up lead to some regressions on legacy APIs that only
 were shown in CI runs.
-Reverting back to using nrf-config and nrf-config-user as this
 supports the resolvement of legacy configurations in case PSA crypto
 is used, but the projects still require legacy APIs for the time
 being...
-Moved platform.c and platform_util.c to mbedcrypto_base as these
 are needed for memory mangement (heap) and other basic operations
 This makes mbedcrypto_base a more platform-oriented library which
 exists at the tail-end of nrf_cc3xx and nrf_oberon libraries that
 provide PSA and legacy support
-Removing psa_crypto_want_config.h.template, but this can be added
 back later if/when we gravitate to the newer mechanism of crypto
 configurations
-Adding fix for enabling MBEDTLS_PK_USE_PSA_EC_DATA even when
 MBEDTLS_ECP_C is enabled. Please see pk.h for details, but note that
 we have chosen to always use PSA when PSA is enabled.
-Ensuring that all TLS/DTLS and X.509 is wrapped with a check to see
 that we aren't inside a secure image (MBEDTLS_PSA_CRYPTO_SPM) in which
 case we don't want to compile TLS/DTLS and X.509 support
-Adding back the PSA_WANT_XXXX functionality to
 psa_crypto_config.h.template (was split out before)
-Removing psa_crypto_want_config.h.template and
 psa_crypto_want_config.cmake
-Copied Mbed TLS check_config.h into generated folder which is higher
 in include order. This is done to prevent using Oberon PSA core version
 of the same file, that isn't compatible with the new way of resolving
 if PSA is enabled (and shall be used).
-Changed resolvement of adding include-folders for nrf_security
 and Mbed TLS for situations where you don't enable a local build of
 PSA crypto core. The use-cases for this is TF-M in ns build and
 full legacy solution.
-Minor code convention fixes for CMakeLists

Note: This fix was primarily for WIFI, but it may also solve additional issues in other cases...

ref: NCSDK-26412

Signed-off-by: Frank Audun Kvamtrø <frank.kvamtro@nordicsemi.no>


mswarowsky:
Got fixed with #14854

PR #14468, opened 2024-03-11
manifest: update Find My revision

kapi-no:
Aligned the Find My tree with the name change for configuration files related to the HCI IPC image for nRF53 Network Core target.


mkapala-nordic (approved):

MarekPieta (approved):

kapi-no:
nRF53 runtime error that makes it impossible to initialize the application. This bug results from an unaligned HCI IPC child image configurations that were missed during the January upmerge.

SeppoTakalo (approved):

PR #14474, opened 2024-03-11
lwm2m: Fix FOTA cancellation issues

SeppoTakalo:
As explained in https://nordicsemi.atlassian.net/browse/NCSDK-26534

Cancelling the FOTA process is now broken.
Two issues found:
1. Empty string might be `"\0"` which is not same as zero length data. This should be handled similarly.
2. LwM2M FOTA utils wrote 16bit value using `sizeof(int)` which caused value to wrap over on certain cases.


juhaylinen (approved):

SeppoTakalo:
Ping @rlubos can you review?

Almost equivalent fix already merged in upstream https://github.com/zephyrproject-rtos/zephyr/pull/70056
We just have to deal a bit extra in FOTA handling.

MarkusLassila (approved):

rlubos (approved):

PR #14475, opened 2024-03-11
smp svr for nrf54l15

michalek-no:


nordicjm (changes requested):
Why?

nvlsianpu:
@nordicjm We have objective (form Projects management team) to expose DFU sample for NCS which will build right away. It was decided that code should base on the zephyr smp_svr. We will support (mainly configuration) for more platform there.

nordicjm:
> @nordicjm We have objective (form Projects management team) to expose DFU sample for NCS which will build right away. It was decided that code should base on the zephyr smp_svr. We will support (mainly configuration) for more platform there.

Just build smp_svr? What value does this even add?

nvlsianpu:
- This aim to build with our NSC flavoured build system, including Partition manager
- This is place where we can add our configuration (if needed)
- we will add NCS specific README there

nordicjm:
Partition manager is on it's way out so it doesn't make sense for that also note that communication was sent out some time ago that no new features are to be added to partition manager and it is bug fixes and maintenance only.

west.yml
```diff
@@ -61,7 +61,7 @@ manifest:
     # https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/modules.html
     - name: zephyr
       repo-path: sdk-zephyr
-      revision: v3.5.99-ncs1-rc2
+      revision: d0e66720ac8532b0d7f7377b576bace366638586
```
nvlsianpu, line 64: this points to commit which is out NCS reach. Please correct.

samples/smp_svr/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.20.0)
+
+# Standard Zephyr application boilerplate.
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(smp_svr)
+
+target_sources(app PRIVATE ../../../zephyr/samples/subsys/mgmt/mcumgr/smp_svr/src/main.c)
+target_sources(app PRIVATE ../../../zephyr/samples/subsys/mgmt/mcumgr/smp_svr/src/bluetooth.c)
```
gchwier, line 8:
```suggestion
set(smp_srv_dir $ENV{ZEPHYR_BASE}/samples/subsys/mgmt/mcumgr/smp_svr)
target_sources(app PRIVATE ${smp_srv_dir}/src/main.c)
target_sources_ifdef(CONFIG_MCUMGR_TRANSPORT_BT app PRIVATE ${smp_srv_dir}/src/bluetooth.c)
```
you can use ZEPHYR_BASE to point the original code and use variable to not repeat it.
Also include bluetooth.c conditionally, as it is done in the original config

samples/smp_svr/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
gchwier, line 1: It should be moved to prj.conf, this is common config, not only for nrf54l15

samples/smp_svr/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
gchwier, line 1: do we really need it? Why to change default uart20, that is used in the Upstream?

gchwier (changes requested):
missing sample.yaml file for Twister

PR #14476, opened 2024-03-11
hw_unique_key: cracen: Add cracen support

vili-nordic:
Implements the hw_unique_key library for Cracen.


subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.c
```diff
@@ -342,6 +347,26 @@ psa_status_t convert_from_psa_attributes(const psa_key_attributes_t *key_attr,
 	memset(metadata, 0, sizeof(*metadata));
 	metadata->metadata_version = 0;
 
+	metadata->key_usage_scheme = CRACEN_PSA_GET_KEY_USAGE_SCHEME(psa_get_key_id(key_attr));
+	if (metadata->key_usage_scheme > KMU_METADATA_SCHEME_RAW) {
```
Vge0rge, line 351: It's kind of a nit but maybe it makes sense to have explicit check on the specific keys that we support rather than implicit like this. 
Maybe a switch statement or something which the key types that we support and the default option can return invalid or something.

That's more of a suggestion though, we can also keep it like that if you think that it is better.

samples/keys/hw_unique_key/src/derive_key.c
```diff
@@ -29,7 +29,7 @@ LOG_MODULE_DECLARE(app, LOG_LEVEL_DBG);
 psa_status_t derive_key(psa_key_attributes_t *attributes, uint8_t *key_label,
 			uint32_t label_size, psa_key_id_t *key_id_out)
 {
-	uint8_t key_out[HUK_SIZE_BYTES];
+	uint8_t key_out[PSA_BITS_TO_BYTES(256)];
```
mswarowsky, line 32: Why do we hardcode the number here now instead of using the definition?

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.c
```diff
@@ -607,6 +626,20 @@ psa_status_t cracen_kmu_get_builtin_key(psa_drv_slot_number_t slot_number,
...
+		case METADATA_ALG_KEY_BITS_256:
+			key->number_of_slots = 2;
+			break;
+		case METADATA_ALG_KEY_BITS_384_SEED:
+			key->number_of_slots = 3;
+			break;
+		default:
+			return PSA_ERROR_DATA_INVALID;
+		}
 		key->number_of_slots = metadata.size > METADATA_ALG_KEY_BITS_128 ? 2 : 1;
```
mswarowsky, line 643: I don't get why we set `key->number_of_slots` in the switch case before just to set it here to either two or one.

samples/keys/hw_unique_key/src/derive_key.c
vili-nordic: it's the wrong definition to use. the length of the derived key is unrelated to the "huk-size" (which is 48 bytes on cracen. It just happened to align with aes key sizes on cryptocell)

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.c
vili-nordic: this line should be removed :)

mswarowsky (approved):
Approving with the assumption that @Vge0rge comment about the key size for CC310 will be handled

Vge0rge (approved):

PR #14478, opened 2024-03-12
manifest: Fix defined in monitor mode support

kapbh:
Add missing defined into condition in  monitor mode support.


kapbh:
manifest is updated with https://github.com/nrfconnect/sdk-nrf/pull/14514

PR #14480, opened 2024-03-12
nrf_rpc_ipc: Lazy waiting for endpoint bond

rakons:
This commit removes the waiting for the remote cpu while initialization is done.
The waiting is moved to a sending function.
This way we would be able to initialize multiple IPC endpoints also in other modules and wait for the bond to happen only if we need to transfer anything.

JIRA: NCSDK-26208

This is proof of concept for the time being. DO NOT MERGE until agreed with @hubertmis


grochu (approved):
Looks good.

include/nrf_rpc/nrf_rpc_ipc.h
```diff
@@ -31,6 +31,14 @@ extern "C" {
  */
 extern const struct nrf_rpc_tr_api nrf_rpc_ipc_service_api;
 
+/** @brief Endpoint state */
+enum nrf_rpc_ipc_state {
+	NRF_RPC_IPC_STATE_UNINITIALIZED, /**< Endpoint is uninitialized */
```
KAGA164, line 36:
```suggestion
        /** Endpoint is uninitialized */
	NRF_RPC_IPC_STATE_UNINITIALIZED, /**
```

Minor comment about documentation style. It is different now than anywhere else

include/nrf_rpc/nrf_rpc_ipc.h
```diff
@@ -57,7 +68,7 @@ struct nrf_rpc_ipc {
 	void *context;
 
 	/** Indicates if transport is already initialized. */
-	bool used;
+	enum nrf_rpc_ipc_state state;
```
hubertmis, line 71: nit: I suggest avoiding enumeration in header files because using different compiler options for building a library exposing this header than the code using the library, the size of this structure field could differ.

How about `typedef uint8_t nrf_rpc_ipc_state;` to avoid this potential issue?
```suggestion
	nrf_rpc_ipc_state state;
```

hubertmis (approved):

include/nrf_rpc/nrf_rpc_ipc.h
KAGA164: I forgot to add this comment. I have strong believe that `init` and `sent` can be called from different context. Doesn't we need to make state atomic?

rakons: We do not need the typedef here. I can put uint8_t directly into the structure. It is used internally only.
Moreover - with that change I can move enumerations into *.c file.

include/nrf_rpc/nrf_rpc_ipc.h
rakons: This would be moved to *.c file and the comments would be changed to non-doxygen ones.

rakons: Done

include/nrf_rpc/nrf_rpc_ipc.h
rakons: Done

KAGA164: Discussed offline

KAGA164 (approved):

include/nrf_rpc/nrf_rpc_ipc.h
rakons: > I forgot to add this comment. I have strong believe that `init` and `sent` can be called from different context. Doesn't we need to make state atomic?

For the time being and how it is implemented right now we do not need synchronization. The initialization looks only if the state is UNINITIALIZED. If it is not the function would exit. In the same time if the state is UNINITIALIZED the send function would exit immediately. The place where the state is changed in initialization function it taken carefully to be changed only when we are ready for the send function to wait. It means that even if the send function would be called before init function leaves but it already changed the state - the further processing would be as expected.

I suppose the special care would be needed when we would allow the communication to reinitialize. But lets keep it simple if we do not need it right now.

hubertmis (approved):

PR #14482, opened 2024-03-12
applications: nrf5340_audio: We should improve how the SIRK is set

gWacey:
OCT-2723

Set the default SIRK and test this against the SIRK to be used. A warning will be logged if these are the same.

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
@@ -29,6 +29,9 @@ LOG_MODULE_REGISTER(unicast_server, CONFIG_UNICAST_SERVER_LOG_LEVEL);
 BUILD_ASSERT(CONFIG_BT_ASCS_ASE_SRC_COUNT <= 1,
 	     "A maximum of one source stream is currently supported");
 
+BUILD_ASSERT(strlen(CONFIG_BT_SET_IDENTITY_RESOLVING_KEY) == BT_CSIP_SET_SIRK_SIZE,
+	     "SIRK incorrect size, must be 16 bytes");
+
 ZBUS_CHAN_DEFINE(le_audio_chan, struct le_audio_msg, NULL, NULL, ZBUS_OBSERVERS_EMPTY,
 		 ZBUS_MSG_INIT(0));
 
@@ -115,13 +118,6 @@ static struct bt_csip_set_member_cb csip_callbacks = {
 struct bt_csip_set_member_register_param csip_param = {
 	.set_size = CSIP_SET_SIZE,
 	.lockable = true,
-#if !CONFIG_BT_CSIP_SET_MEMBER_TEST_SAMPLE_DATA
-	/* CSIP SIRK for demo is used, must be changed before production */
-	.set_sirk = {'N', 'R', 'F', '5', '3', '4', '0', '_', 'T', 'W', 'S', '_', 'D', 'E', 'M',
-		     'O'},
-#else
-#warning "CSIP test sample data is used, must be changed before production"
-#endif
 	.cb = &csip_callbacks,
 };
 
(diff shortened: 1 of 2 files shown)


applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
```diff
@@ -718,6 +711,19 @@ int unicast_server_enable(le_audio_receive_cb recv_cb)
 
 	channel_assignment_get(&channel);
 
+	if (!IS_ENABLED(CONFIG_BT_CSIP_SET_MEMBER_TEST_SAMPLE_DATA)) {
+		if (!strcmp(CONFIG_BT_SET_IDENTITY_RESOLVING_KEY_DEFAULT,
+			    CONFIG_BT_SET_IDENTITY_RESOLVING_KEY)) {
```
alexsven, line 716: Might be easier to read if it was:
```suggestion
		if (strcmp(CONFIG_BT_SET_IDENTITY_RESOLVING_KEY_DEFAULT,
			    CONFIG_BT_SET_IDENTITY_RESOLVING_KEY) == 0) {
```

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
```diff
@@ -718,6 +711,19 @@ int unicast_server_enable(le_audio_receive_cb recv_cb)
...
+			LOG_WRN("CSIP using the default SIRK, must be changed before production");
+		}
+
+		memcpy(csip_param.set_sirk, CONFIG_BT_SET_IDENTITY_RESOLVING_KEY,
+		       MIN(sizeof(CONFIG_BT_SET_IDENTITY_RESOLVING_KEY),
+			   sizeof(csip_param.set_sirk)));
```
alexsven, line 722:
```suggestion
		memcpy(csip_param.set_sirk, CONFIG_BT_SET_IDENTITY_RESOLVING_KEY,
		       MIN(sizeof(CONFIG_BT_SET_IDENTITY_RESOLVING_KEY),
			   BT_CSIP_SET_SIRK_SIZE));
```
Have you checked what happens if the key is shorter than 16 characters? Do we then also get the '\0'?

I see that broadcast key should use BT_AUDIO_BROADCAST_CODE_SIZE in the MIN as well, but that can probably be done in a separate issue

alexsven (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
```diff
@@ -718,6 +711,19 @@ int unicast_server_enable(le_audio_receive_cb recv_cb)
...
+	} else {
+		LOG_WRN("CSIP test sample data is used, must be changed before production");
+	}
+
```
koffes, line 726: For readability:
```suggestion
	if (IS_ENABLED(CONFIG_BT_CSIP_SET_MEMBER_TEST_SAMPLE_DATA)) {
		LOG_WRN("CSIP test sample data is used, must be changed before production");		
	} else {
	        if (!strcmp(CONFIG_BT_SET_IDENTITY_RESOLVING_KEY_DEFAULT,
			    CONFIG_BT_SET_IDENTITY_RESOLVING_KEY)) {
			LOG_WRN("CSIP using the default SIRK, must be changed before production");
		}

		memcpy(csip_param.set_sirk, CONFIG_BT_SET_IDENTITY_RESOLVING_KEY,
		       MIN(sizeof(CONFIG_BT_SET_IDENTITY_RESOLVING_KEY),
			   sizeof(csip_param.set_sirk)));

	}

```

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
koffes: It is demanded that the SIRK is 128 bits -> 16 bytes. This should be documented in the Kconfig and we need a check for it.

koffes (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
gWacey: Yes that looks better.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
gWacey: I had it like that but it looked a bit messy. I will go with == 0.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_server.c
gWacey: Yes I have checked for less than 16 characters and the '\0' is included in the key.
So do we only allow SIRKs of 16 bytes?

alexsven (approved):

koffes (approved):

PR #14484, opened 2024-03-12
samples: matter: Align the thermostat feature map

markaj-nordic:
We support Thermostat AUTO mode but it was not indicated by the cluster feature map (bit 5 in the feature map).


doublemis1 (changes requested):
If we want to support AUTO Feature we shall enable:
MinSetpointDeadBand attribute

ArekBalysNordic (approved):
The regeneration of the ZAP file looks good.

doublemis1 (approved):
Thanks for update

PR #14487, opened 2024-03-12
nrf_security: kconfig: Dont enable TLS or X509 by default

SebastianBoe:
It is bad practice for features to be enabled by default.

It is not clear to me why these should be enabled when PSA and legacy ECD* are enabled.


subsys/nrf_security/Kconfig.tls
```diff
@@ -8,7 +8,7 @@ menuconfig MBEDTLS_X509_LIBRARY
 	bool
 	prompt "Create mbed TLS x509 library"
 	select MBEDTLS_PK_C
-	default y if MBEDTLS_PSA_CRYPTO_C && (MBEDTLS_ECDSA_C || MBEDTLS_ECDH_C)
+	default n
```
mswarowsky, line 11: no default == `default n` so then we can just remove the line, as we do for most other Kconfigs, same goes for the  second Kconfig further down

Vge0rge (approved):

modules/hostap/Kconfig
```diff
@@ -136,9 +136,10 @@ config WPA_SUPP_CRYPTO_LEGACY
 	select MBEDTLS_ECP_C
 	select MBEDTLS_CTR_DRBG_C
 	select MBEDTLS_PK_C
+	select MBEDTLS_PK_PARSE_C
```
krish2718, line 139: Is this an unrelated change? If so, can you please move to a new commit with the commit log explaining the reason?

SebastianBoe: Hi,

in the commit message I have written:

"Enable PK_PARSE_C as
hostap/src/crypto/crypto_mbedtls-ec.c
uses it."

does this suffice?

A different commit would make it less clear that this change is required because of the changing default values.

krish2718: Ah ok, may be we can modify that to:

```
Enable PK_PARSE_C explicitly (used by `hostap/src/crypto/crypto_mbedtls-ec.c`) now that legacy is disabled by default.
```

SebastianBoe: will do :)

SebastianBoe: done

mswarowsky (approved):

SebastianBoe:
I'm giving up on this.

I expect this to change significantly soon anyway.

PR #14497, opened 2024-03-13
samples: wifi: raw_tx_packet: Fix raw packet transmission issue

D-Triveni:
Implement semaphore to wait for DHCP completion after connection, fixing raw packet transmission issue.
Fixes SHEL-2598.

### DIFF
--- a/samples/wifi/raw_tx_packet/Kconfig
+++ b/samples/wifi/raw_tx_packet/Kconfig
@@ -8,9 +8,13 @@ source "Kconfig.zephyr"
 
 menu "Nordic Raw Tx Packet sample"
 
-config CONNECTION_TIMEOUT
+config RAW_TX_PKT_SAMPLE_CONNECTION_TIMEOUT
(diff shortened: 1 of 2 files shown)


lave30 (approved):

samples/wifi/raw_tx_packet/Kconfig
```diff
@@ -12,6 +12,10 @@ config CONNECTION_TIMEOUT
 	int "Time to wait for a station to connect"
 	default 30
 
+config DHCP_TIMEOUT_S
```
ioannisg, line 15: Any reason this Kconfig and the one above not having the "sample" prefix?

D-Triveni: There is no specific reason for not having the "sample" prefix in those particular Kconfig options.
Added now.

jukkar (approved):

ioannisg (commented):
Thanks. Looks good to me.

PR #14503, opened 2024-03-13
manifest: Matter bootstrap script fix

LipinskiPNordicSemi:
This commit cherry pick from connectedhomeip upstream fix bootsrtap.sh script.


ArekBalysNordic:
This PR is not needed, since the change is only regarding the sdk-connectedhomeip unit tests fix.

PR #14511, opened 2024-03-14
samples: matter: Add Event Trigger support for all samples

ArekBalysNordic:
Matter test events can be used by the Matter controller
to trigger specific device's behavior.

Added a mechanism to register event triggers and use their
callbacks through Matter Controller.

Added the documentation page to describe the mechanism and
create a list of the currently available default event triggers.


doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,229 @@
...
+------------------
+
+You can set the enable key to a specific value only if the :kconfig:option:`CONFIG_CHIP_FACTORY_DATA` Kconfig option is set to ``y``.
+Otherwise, the default value ``00112233445566778899AABBCCDDEEFF`` will be used.
+
+To specify the enable key through the build system, enable the :kconfig:option:`CONFIG_CHIP_FACTORY_DATA_BUILD` Kconfig option by setting it to ``y``.
+Then, set the :kconfig:option:`CONFIG_CHIP_DEVICE_ENABLE_KEY` Kconfig option to a 32-byte hexadecimal string value.
+
+If :kconfig:option:`CONFIG_CHIP_FACTORY_DATA_BUILD` is set to ``n``, you can use the Python script to generate the factory data set with the specific enable key value.
+To learn how to do this, read the :doc:`matter:nrfconnect_factory_data_configuration` guide in the Matter documentation.
```
wiba-nordic, line 93:
```suggestion
If :kconfig:option:`CONFIG_CHIP_FACTORY_DATA_BUILD` is set to ``n``, you can follow the :doc:`matter:nrfconnect_factory_data_configuration` guide in the Matter documentation to generate the factory data set with the specific enable key value.
```
Since we're not introducing the Python script on this page, I think this is a better wording.

wiba-nordic (changes requested):
Will review the rest of the document later today.

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,229 @@
...
+Setting the enable key
+======================
+
+The enable key can be provided utilizing the factory data or directly in the application code.
+
+Using factory data
+------------------
+
+You can set the enable key to a specific value only if the :kconfig:option:`CONFIG_CHIP_FACTORY_DATA` Kconfig option is set to ``y``.
+Otherwise, the default value ``00112233445566778899AABBCCDDEEFF`` will be used.
```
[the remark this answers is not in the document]
ArekBalysNordic: I guess it should be, I will propose something.

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,242 @@
...
+
+An activation code, which is 64 bits in length, and consist of two components: the event trigger ID and the event trigger value.
+
+* The event trigger ID is 32 bits long and uniquely identifies the trigger.
+  It is provided as the first 32 bits of the activation code.
+* The event trigger value is specific to the given trigger and is supplied as the last 32 bits of the activation code.
+
+The activation code follows the pattern ``0xIIIIIIIIVVVVVVVV``, where ``I`` represents the ID part and ``V`` represents the value part.
+
+For example the ``0x1000000000001234`` activation code stands for trigger ID qual to ``0x1000000000000000`` and it contains a specific value ``0x1234``.
```
wiba-nordic, line 160:
```suggestion
For example the ``0x1000000000001234`` activation code stands for a trigger ID qual to ``0x1000000000000000`` and a specific value of ``0x1234``.
```

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,242 @@
...
+The maximum number of event triggers that can be registered is configurable.
+To adjust this limit, set the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS_MAX` Kconfig option to the desired value.
+
+To register a new test event trigger, follow these steps:
+
+1. Create a function that will be executed when the device receives a valid enable key and activation code.
+
+  This function must return a ``CHIP_ERROR`` code and accept a ``Nrf::Matter::TestEventTrigger::TriggerValue`` as its argument.
+  You can utilize the provided argument within this function as needed.
+  Use ``CHIP_ERROR`` codes to communicate appropriate responses to the Matter Controller.
```
wiba-nordic, line 177:
```suggestion
  You can utilize the provided argument within this function as needed.
  
  Use ``CHIP_ERROR`` codes to communicate appropriate responses to the Matter Controller.
```

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,242 @@
...
+  For instance, you might return ``CHIP_ERROR_INVALID_ARGUMENT`` to indicate that the user has provided an incorrect value argument.
+
+    .. note::
+
+      The callback method is not thread-safe.
+      Ensure that all operations within it are executed in a thread-safe manner.
+      To perform operations within the Matter stack context, use the ``chip::DeviceLayer::SystemLayer().ScheduleLambda`` method. For operations in the application context, use the ``Nrf::PostTask`` function.
```
wiba-nordic, line 184:
```suggestion
      To perform operations within the Matter stack context, use the ``chip::DeviceLayer::SystemLayer().ScheduleLambda`` method.
      For operations in the application context, use the ``Nrf::PostTask`` function.
```

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,242 @@
...
+
+The :ref:`CHIP Tool application <ug_matter_gs_tools_chip>` is required to send ``test-event-trigger`` commands to the general diagnostics cluster.
+Ensure that it is installed before initiating Matter testing events.
+
+Your device needs to includes the general diagnostics cluster within its Matter data model database to perform Matter testing events.
+For guidance on configuring the general diagnostics cluster on your device, refer to the instructions provided on the :ref:`ug_matter_gs_adding_cluster` page.
+
+Activation of test event codes requires the provisioning of an enable key.
+To ensure secure operation, the enable key must be specific for each device.
+You can find the detailed instructions on how to establish the enable key for your device in the :ref:`matter_test_event_triggers_setting_enable_key`` section.
```
wiba-nordic, line 31:
```suggestion
You can find the detailed instructions on how to establish the enable key for your device in the :ref:`matter_test_event_triggers_setting_enable_key` section.
```
Very small error that breaks the doc build.

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,242 @@
...
+  .. code-block:: c++
+
+    CHIP_ERROR MyFunctionCallback(Nrf::Matter::TestEventTrigger::TriggerValue value)
+    {
+        /* Define the required behavior of the device here. */
+
+        return CHIP_NO_ERROR;
+    }
+
+#. Register the new event trigger:
```
wiba-nordic, line 198:
```suggestion
#. Register the new event trigger.
```

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,242 @@
...
+
+      Nrf::Matter::TestEventTrigger::EventTrigger myEventTrigger;
+      myEventTrigger.Id = 0xF000F0000000;
+      myEventTrigger.Mask = 0xFFFF;
+      myEventTrigger.Callback = MyFunctionCallback;
+
+      CHIP_ERROR err = Nrf::Matter::TestEventTrigger::Instance().RegisterTestEventTrigger(myEventTrigger);
+
+Usage
+=====
```
wiba-nordic, line 230:
```suggestion
*****
```

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,242 @@
...
+
+To trigger a specific event on the device, fill the `enable key`, `activation code`, and `node id` fields and run the following command:
```
wiba-nordic, line 232:
```suggestion
To trigger a specific event on the device, run the following command: 
```

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,242 @@
...
+
+.. code-block:: console
+
+   ./chip-tool generaldiagnostics test-event-trigger hex:<enable key> <activation code> <node id> 0
+
+An example of the Reboot activation code with 5 ms delay for Matter Template device which has a `node id` set to ``1``, using the default enable key:
```
wiba-nordic, line 238:
```suggestion
Replace the ``enable key``, ``activation code``, and ``node id`` values with the values for the event you want to trigger.
The following is an example of the Reboot activation code with a 5 ms delay for Matter Template device which has a ``node id`` set to ``1``, using the default enable key:
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -218,7 +218,9 @@ Keys samples
 Matter samples
 --------------
 
-|no_changes_yet_note|
+   * Added Test Event Triggers to all Matter samples.
+     By utilizing the Test Event Triggers, you can simulate various operational conditions and responses in your Matter device without the need for external setup.
+     To get started with using Test Event Triggers in your Matter samples and to understand the full capabilities of this feature, please refer to the guide available in the :ref:`matter_test_event_triggers` page.
```
wiba-nordic, line 223:
```suggestion
* Added test event triggers to all Matter samples.
  By utilizing the test event triggers, you can simulate various operational conditions and responses in your Matter device without the need for external setup.

  To get started with using test event triggers in your Matter samples and to understand the capabilities of this feature, refer to the :ref:`matter_test_event_triggers` page.
```
The change to indentation hides it from the diff, but there's some wording changes as well.

wiba-nordic (changes requested):

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,248 @@
...
+  Here's an example to handle the ``0xF000F00000001234`` activation code, where 1234 is the event trigger value field:
+
+    .. code-block:: c++
+
+      Nrf::Matter::TestEventTrigger::EventTrigger myEventTrigger;
+      myEventTrigger.Id = 0xF000F0000000;
+      myEventTrigger.Mask = 0xFFFF;
+      myEventTrigger.Callback = MyFunctionCallback;
+
+      CHIP_ERROR err = Nrf::Matter::TestEventTrigger::Instance().RegisterTestEventTrigger(myEventTrigger);
```
wiba-nordic, line 231:
```suggestion
1. Create a function that will be executed when the device receives a valid enable key and activation code.

   This function must return a ``CHIP_ERROR`` code and accept a ``Nrf::Matter::TestEventTrigger::TriggerValue`` as its argument.
   You can utilize the provided argument within this function as needed.

   Use ``CHIP_ERROR`` codes to communicate appropriate responses to the Matter Controller.
   For instance, you might return ``CHIP_ERROR_INVALID_ARGUMENT`` to indicate that the user has provided an incorrect value argument.

   .. note::

      The callback method is not thread-safe.
      Ensure that all operations within it are executed in a thread-safe manner.
      To perform operations within the Matter stack context, use the ``chip::DeviceLayer::SystemLayer().ScheduleLambda`` method.
      For operations in the application context, use the ``Nrf::PostTask`` function.


   Here is an example of how to create a callback function:

   .. code-block:: c++

      CHIP_ERROR MyFunctionCallback(Nrf::Matter::TestEventTrigger::TriggerValue value)
      {
          /* Define the required behavior of the device here. */

          return CHIP_NO_ERROR;
      }

#. Register the new event trigger.

   Use the following example as a guide to register a new event:

   .. code-block:: c++

      /* Create a new event */
      Nrf::Matter::TestEventTrigger::EventTrigger myEventTrigger;

      /* Assign all fields */
      myEventTrigger.Id = /* Fill the ID filed */;
      myEventTrigger.Mask = /* Fill the value mask filed */;
      myEventTrigger.Callback = MyFunctionCallback;

      /* Register the new event */
      CHIP_ERROR err = Nrf::Matter::TestEventTrigger::Instance().RegisterTestEventTrigger(myEventTrigger);

      /* Remember to check the CHIP_ERROR return code */


   Here's an example to handle the ``0xF000F00000001234`` activation code, where 1234 is the event trigger value field:

   .. code-block:: c++

      Nrf::Matter::TestEventTrigger::EventTrigger myEventTrigger;
      myEventTrigger.Id = 0xF000F0000000;
      myEventTrigger.Mask = 0xFFFF;
      myEventTrigger.Callback = MyFunctionCallback;

      CHIP_ERROR err = Nrf::Matter::TestEventTrigger::Instance().RegisterTestEventTrigger(myEventTrigger);
```
Indentation wasn't correct, so the numbered steps didn't build correctly.

wiba-nordic (changes requested):
Also need to fix the trailing white spaces to pass compliance.

samples/matter/common/src/util/finite_map.h
```diff
@@ -0,0 +1,121 @@
...
+
+	uint16_t GetFirstFreeSlot()
+	{
+		uint16_t foundIndex = 0;
+		for (auto &it : mMap) {
+			if (kInvalidKey == it.key)
+				return foundIndex;
+			foundIndex++;
+		}
+		return -1;
```
markaj-nordic, line 98: Is it done on purpose to underflow here, maybe it's better to return UINT16_MAX or kInvalidKey. Also I guess we could use `std::find_if`.

samples/matter/common/src/event_triggers/default_event_triggers.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+void FactoryResetDelayTimerTimeoutCallback(k_timer *timer)
+{
+	Server::GetInstance().ScheduleFactoryReset();
+}
+
+void RebootDelayTimerTimeoutCallback(k_timer *timer)
+{
+	DeviceLayer::SystemLayer().ScheduleLambda(
+		[] { DeviceLayer::Reboot(DeviceLayer::SoftwareRebootReason::kOther); });
```
markaj-nordic, line 33: Should we emulate the ShutDown event is this case? Adding:
`PlatformMgr().HandleServerShuttingDown();
k_msleep(CHIP_DEVICE_CONFIG_SERVER_SHUTDOWN_ACTIONS_SLEEP_MS);`

should do the job.

samples/matter/lock/src/app_task.cpp
```diff
@@ -261,6 +271,13 @@ CHIP_ERROR AppTask::Init()
 	/* Initialize lock manager */
 	BoltLockMgr().Init(LockStateChanged);
 
+	/* Register Door Lock test event trigger */
+#ifdef CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS_REGISTER_DEFAULTS
```
markaj-nordic, line 275: I'd say in the case of adding a custom event trigger the `CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS` must be enough.

samples/matter/lock/src/app_task.h
```diff
@@ -17,6 +17,10 @@ struct Identify;
 enum class SwitchButtonAction : uint8_t { Pressed, Released };
 #endif
 
+#ifdef CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS_REGISTER_DEFAULTS
```
markaj-nordic, line 20: ditto, application code should not depend on `CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS_REGISTER_DEFAULTS`

samples/matter/lock/src/app_task.h
```diff
@@ -48,4 +52,9 @@ class AppTask {
 	static void NUSLockCallback(void *context);
 	static void NUSUnlockCallback(void *context);
 #endif
+
+#ifdef CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS_REGISTER_DEFAULTS
```
markaj-nordic, line 56: ditto

samples/matter/common/src/event_triggers/event_triggers.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+	/* Check if the Enable Key is set to the "disabled" value. */
+	if (!IsEventTriggerEnabled()) {
+		return CHIP_ERROR_INCORRECT_STATE;
+	}
+
+	TriggerSlot slot = GetTriggerSlot(eventTrigger);
+
+	VerifyOrReturnError(slot != kInvalidTriggerSlot, CHIP_ERROR_INVALID_ARGUMENT);
+	VerifyOrReturnError(mTriggersMap[slot].Callback != nullptr, CHIP_ERROR_INCORRECT_STATE);
```
markaj-nordic, line 66: Before using []operator from the FiniteMap, we must first check if the element is present in the map with Contains(), that's the current FiniteMap API limitation.

samples/matter/common/src/event_triggers/event_triggers.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+	return mTriggersMap[slot].Callback(eventTrigger & mTriggersMap[slot].Mask);
+}
+
+uint16_t TestEventTrigger::GetTriggerSlot(uint64_t eventTrigger)
+{
+	for (uint16_t slot = 0; slot < kMaxEventTriggers; slot++) {
```
Damian-Nordic, line 73: What's actually the value of using `FiniteMap` in this case if you still have to loop over all entries because your _real_ key is `uint64_t` while `FiniteMap` supports only `uint16_t` keys. I think either `FiniteMap` should use keys of any type or it would be easier and more efficient to use a plain array of triggers.

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,248 @@
...
+You can find the detailed instructions on how to establish the enable key for your device in the :ref:`matter_test_event_triggers_setting_enable_key` section.
+
+.. _matter_test_event_triggers_default_test_event_triggers:
+
+Default test event triggers
+***************************
+
+You can use the pre-defined common test event triggers in your application.
+These pre-defined test triggers are enabled by default.
+To disable them, set the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS_REGISTER_DEFAULTS` Kconfig option to ``n``.
```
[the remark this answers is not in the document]
ArekBalysNordic: I do it regarding our tests, but in fact we can disable it by default - it was my first thought.

samples/matter/common/src/event_triggers/event_triggers.cpp
markaj-nordic: Yes, originally the FiniteMap was developed as a dedicated Matter Bridge utility, not with generic usability in mind. So we may try to make it more versatile or use something else here.

ArekBalysNordic: That's right I wanted to align with the current solution, but I will propose something more efficient :)

samples/matter/common/src/event_triggers/event_triggers.h
```diff
@@ -0,0 +1,116 @@
...
+#include <lib/core/CHIPError.h>
+
+#include "util/finite_map.h"
+
+#include <cstddef>
+#include <cstdint>
+
+namespace Nrf::Matter
+{
+class TestEventTrigger : public chip::TestEventTriggerDelegate {
```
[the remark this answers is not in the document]
ArekBalysNordic: Pulled :)

samples/matter/common/src/event_triggers/event_triggers.cpp
ArekBalysNordic: I added the FiniteDict, to cover our requirements and thoughts, @markaj-nordic @Damian-Nordic what do you think?

samples/matter/common/src/event_triggers/event_triggers.h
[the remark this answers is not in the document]
Damian-Nordic: What I meant is that you could probably use `SimpleTestEventTriggerDelegate` directly without implementing this class, right? At least at first glance, it provides the same functionality i.e. adding `TestEventTriggerHandler` instances that play the same role as `EventTrigger` in your solution. Thus, we would only need `TestEventTriggerHandler` implementations in `matter/common`. Either way is fine with me, but I'm just letting you know that `TestEventTrigger` doesn't need to be reimplemented in our platform's code.

ArekBalysNordic:
@wiba-nordic  I've added the new section to docs: `Register an existing test event trigger handler`

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,275 @@
...
+     CHIP_ERROR err = Nrf::Matter::TestEventTrigger::Instance().RegisterTestEventTrigger(myTriggerID, myEventTrigger);
+
+Register an existing test event trigger handler
+***********************************************
+
+In Matter SDK there are some already implemented test event trigger handlers.
+All of them inherits by `TestEventTriggerHandler` class and are implemented in various places in the Matter SDK.
+You can use them and their pre-defined behaviors using the test event triggers mechanism presented on this page.
+
+The events declared in an existing test event triggers can have the a different semantic than described in the :ref:`matter_test_event_triggers_custom_triggers` section.
```
wiba-nordic, line 238:
```suggestion
The events declared in existing test event triggers can have a different semantic than described in the :ref:`matter_test_event_triggers_custom_triggers` section.
```

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,275 @@
...
+
+.. code-block:: c++
+
+  /* Create the Trigger Handler object */
+  static TestEventTriggerHandler existingTriggerHandler;
+  CHIP_ERROR err = Nrf::Matter::TestEventTrigger::Instance().RegisterTestEventTriggerHandler(&existingTriggerHandler);
+
+  /* Remember to check the CHIP_ERROR return code */
+
+If the returning `CHIP_ERROR` code is equal to `CHIP_ERROR_NO_MEMORY`, you need to increase the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS_MAX_TRIGGERS_DELEGATES` kconfig option to the higher value.
```
wiba-nordic, line 250:
```suggestion
If the returning ``CHIP_ERROR`` code is equal to ```CHIP_ERROR_NO_MEMORY``, you need to increase the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS_MAX_TRIGGERS_DELEGATES` kconfig option to the higher value.
```

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,275 @@
...
+
+For example, you can register and use the `OTATestEventTriggerHandler` and trigger pre-defined Matter OTA DFU behaviors using the following code:
```
wiba-nordic, line 252:
```suggestion
For example, you can register and use the ``OTATestEventTriggerHandler`` handler and trigger pre-defined Matter OTA DFU behaviors using the following code:
```

wiba-nordic (changes requested):

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,274 @@
...
+Usage
+*****
+
+To trigger a specific event on the device, run the following command:
+
+.. code-block:: console
+
+  ./chip-tool generaldiagnostics test-event-trigger hex:<enable key> <activation code> <node id> 0
+
+Replace the ``enable key``, ``activation code``, and ``node id`` values with the values for the event you want to trigger.
```
wiba-nordic, line 268:
```suggestion
Replace the ``enable key`` value with your device's enable key, and the ``activation code`` and ``node id`` values with the values for the event you want to trigger.
```
Just realized that this sentence could be bit more specific.

wiba-nordic (changes requested):
One little improvement, but otherwise this is looking good.

wiba-nordic (approved):

samples/matter/common/src/util/finite_dict.h
```diff
@@ -0,0 +1,119 @@
...
+     * Erasing existing items under a given key (Erase) - this operation does not take care of the
+	   duplicated items, meaning that the pointer under a given key is released unconditionally
+	   and the responsibility for the duplications is on the application side.
+     * Retrieving values stored under provided key ([] operator)
+     * Checking if the map contains a non-null value under given key (Contains)
+	 * Retrieving a number of free slots available in the map (FreeSlots)
+	Prerequisites:
+     * T must have move semantics and bool()/==operators implemented
+*/
+template <typename T, typename K, std::size_t N> struct FiniteDict {
```
markaj-nordic, line 32: I'd rather refactor the existing FiniteMap to make it more versatile or  because now we end up with a lot of boiler plate code.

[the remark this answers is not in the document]
ArekBalysNordic: I wanted to utilize the tool that Marcin created recently, but maybe Zephyr's list will work here also, and we can leave the finite_dict only for bridge purposes, I will try.

samples/matter/common/src/app/matter_init.cpp
```diff
@@ -170,6 +178,20 @@ void DoInitChipServer(intptr_t /* unused */)
...
 	SetCommissionableDataProvider(sLocalInitData.mFactoryDataProvider);
+
+#ifdef CONFIG_NCS_SAMPLE_MATTER_TEST_EVENT_TRIGGERS
+	/* Read the enable key from the factory data set */
+	uint8_t enableKeyData[chip::TestEventTriggerDelegate::kEnableKeyLength] = {};
+	MutableByteSpan enableKey(enableKeyData);
+	sInitResult = sLocalInitData.mFactoryDataProvider->GetEnableKey(enableKey);
+	VerifyInitResultOrReturn(sInitResult, "GetEnableKey() failed");
+	Nrf::Matter::TestEventTrigger::Instance().SetEnableKey(enableKey);
+	VerifyInitResultOrReturn(sInitResult, "SetEnableKey() failed");
```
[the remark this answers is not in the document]
ArekBalysNordic: Do you mean, the part of the Matter's core?

samples/matter/common/src/util/finite_dict.h
ArekBalysNordic: Ok, I agree, apart from that a Map should translate from one typename to another, so we can extend the existing FiniteMap :)

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,274 @@
+.. _matter_test_event_triggers:
+
+Matter test event triggers
+##########################
+
+.. contents::
+   :local:
+   :depth: 2
+
+Matter test event triggers are a set of software commands designed to initiate specific test scenarios.
```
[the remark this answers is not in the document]
ArekBalysNordic: OK, Will move that information higher then :)

samples/matter/common/src/app/matter_init.cpp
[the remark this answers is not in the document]
ArekBalysNordic: Oh, you meant our default triggers implementation :) Hmm, the only reason is to avoid hiding it from the users, because they need to read the enable key and register it in the TestEventTrigger class. So if we move it to default event triggers it won't be shown explicitly in the init method.

samples/matter/common/src/util/finite_dict.h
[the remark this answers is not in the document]
ArekBalysNordic: There is no such a map either in Matter Core or Zephyr. Marcin added recently the simple one, but I guess we can extend it, make it more versatile, and use it both in Bridge and here. :)

ArekBalysNordic:
@LuDuda @wiba-nordic I've moved information regarding certification to the top of the page. Could you please verify whether it is more clear now?

@markaj-nordic I edited FiniteMap, and now the key can be various, please verify it :)

samples/matter/common/src/app/matter_init.cpp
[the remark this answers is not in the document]
markaj-nordic: discussed f2f

samples/matter/common/src/util/finite_map.h
```diff
@@ -0,0 +1,128 @@
...
+     * retrieving values stored under provided key ([] operator)
+     * checking if the map contains a non-null value under given key (Contains)
+	 * retrieving a number of free slots available in the map (FreeSlots)
+     * iterating though stored item via publicly available mMap member
+     * retrieving the first free slot (GetFirstFreeSlot)
+	Prerequisites:
+     * T1 must be trivial.
+     * T2 must have move semantics and bool()/==operators implemented
+*/
+template <typename T1, typename T2, std::size_t N> struct FiniteMap {
```
markaj-nordic, line 35: The problem with this template parameter T1 is that it must be unsigned numeric type, because it's used for accessing array elements. So, I suggest adding the following assert, or something similar:

`static_assert(std::is_integral_v<T1> && !std::numeric_limits<T1>::is_signed);`

 It'd be also good to mention this in the commentary above (T1 must be unsigned integer type).
 
 I think we should also at least do some smoke tests of the Matter Bridge application before merging this modification.

doc/nrf/protocols/matter/end_product/index.rst
```diff
@@ -36,3 +36,4 @@ Finally, we discuss topics related to Matter certification (:ref:`ug_matter_devi
    configuring_cd
    versioning
    last_fabric_removal_delegate
+   test_event_triggers
```
wiba-nordic, line 39: Should probably add a mention to Line 22 as well. Not sure if it should be with the bootloader mention, or inside the parenthesis with topics related to certification.

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -0,0 +1,274 @@
...
+This mechanism can be used during the certification process to simulate common and specific device scenarios, and during software verification events to validate the correct operation of Matter-compliant devices.
+
+When you provide the enable key using factory data, you can utilize the event trigger feature during the Matter Certification process.
+This is because, when done this way, you can turn off the event trigger functionality by disabling access to the ``generaldiagnostics`` cluster without altering the code that has already been certified.
+
+Once the certification process is complete, you can deactivate the test event trigger functionality by generating new factory data with a modified enable key value.
+This is done by setting the :kconfig:option:`CONFIG_CHIP_DEVICE_ENABLE_KEY` Kconfig option to a value consisting solely of zeros.
+
+For instance, to generate factory data with disabled event trigger functionality, set the :kconfig:option:`CONFIG_CHIP_DEVICE_ENABLE_KEY` Kconfig option to the value ``0x00000000000000000000000000000000``.
+After generating it, flash the :file:`factory_data.hex` file onto the device.
```
wiba-nordic, line 23: I think we could probably move this information to be a separate section. I'd say on the same level and just before the section for the custom event triggers section, or as the very last section. Something like "Using in Matter certification".

Having its own section gives it visibility, and it could then be linked from the mention on Line 14.
```suggestion
When you provide the enable key using factory data, you can utilize the event trigger feature during the Matter certification process.
This is because, when done this way, you can turn off the event trigger functionality by disabling access to the ``generaldiagnostics`` cluster without altering the code that has already been certified.

Once the certification process is complete, you can deactivate the test event trigger functionality by generating new factory data with a modified enable key value.
This is done by setting the :kconfig:option:`CONFIG_CHIP_DEVICE_ENABLE_KEY` Kconfig option to a value consisting solely of zeros.

For instance, to generate factory data with disabled event trigger functionality, set the :kconfig:option:`CONFIG_CHIP_DEVICE_ENABLE_KEY` Kconfig option to the value ``0x00000000000000000000000000000000``.
After generating it, flash the :file:`factory_data.hex` file onto the device.
```

wiba-nordic (changes requested):

samples/matter/common/src/util/finite_map.h
markaj-nordic: Ok, I did not notice the implementation had been adapted for storing Items on the first free slot instead of the index corresponding to the Item.key. So in theory there are no such limitations on the T1 but we need to select the `kInvalidKey` value differently (as now we can store keys of values outside of the `mMap` range) - maybe the maximum of the numeric limits of T1 would be ok (let's then add this information to the map description).

samples/matter/common/src/util/finite_map.h
```diff
@@ -0,0 +1,128 @@
...
+/*
+   FiniteMap template container allows to wrap mappings between T1-type trivial key and T2-type value.
+   The size or the container is predefined at compilation time, therefore
+   the maximum number of stored elements must be well known. FiniteMap owns
+   inserted values, meaning that once the user inserts a value it will not longer
+   be valid outside of the container, i.e. FiniteMap cannot return stored object by copy.
+   FiniteMap's API offers basic operations like:
+     * inserting new values under provided key (Insert)
+     * erasing existing items under a given key (Erase) - this operation does not take care of the
+	   duplicated items,  meaning that the pointer under a given key is released unconditionally
```
markaj-nordic, line 24:
```suggestion
     * erasing existing items under a given key (Erase)
```
I think it's no longer true. This responsibility has been moved to the application.

wiba-nordic (approved):

LuDuda (approved):
LGTM :+1:

Assuming comment is resolved :)

markaj-nordic (approved):

PR #14512, opened 2024-03-14
applications: nrf5340_audio: Only enable encoder and decoder when used

alexsven:
- OCT-2951
- Only enable LC3 encoder if bitrate is set
- Only enable LC3 decoder if decoder sample rate is set


rick1082 (approved):
Confirm works with Android phone.

koffes (approved):

PR #14515, opened 2024-03-14
modules: wfa-qt: Refactor response sending code

D-Triveni:
Refactor the de-duplication of response sending code.

### DIFF
--- a/modules/wfa-qt/src/indigo_api_callback_dut.c
+++ b/modules/wfa-qt/src/indigo_api_callback_dut.c
@@ -169,6 +169,31 @@ done:
 		} \
 	} while (0)
 
+static int fill_resp_to_ctrl(struct packet_wrapper *req, struct packet_wrapper *resp,
+			     int status, char *message, char *data, int tlv_type)
+{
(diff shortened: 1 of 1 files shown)


modules/wfa-qt/src/indigo_api_callback_dut.c
```diff
@@ -1740,13 +1736,7 @@ static int get_ip_addr_handler(struct packet_wrapper *req, struct packet_wrapper
...
 	}
 
-	fill_wrapper_message_hdr(resp, API_CMD_RESPONSE, req->hdr.seq);
-	fill_wrapper_tlv_byte(resp, TLV_STATUS, status);
-	fill_wrapper_tlv_bytes(resp, TLV_MESSAGE, strlen(message), message);
-	if (status == TLV_VALUE_STATUS_OK) {
-		fill_wrapper_tlv_bytes(resp, TLV_DUT_WLAN_IP_ADDR, strlen(buffer), buffer);
-	}
-	return 0;
+	return FILL_RESP_SUCCESS(buffer, TLV_DUT_WLAN_IP_ADDR);
```
krish2718, line 1739: Are we not checking status here? this would always return success?

PR #14536, opened 2024-03-15
cracen: Fix code issues

vili-nordic:
Fixes for issues found using static analysis.

### DIFF
--- a/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/common.c
+++ b/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/common.c
@@ -413,11 +413,15 @@ int cracen_signature_asn1_get_operand(unsigned char **p, const unsigned char *en
 				      struct sx_buf *op)
 {
 	int ret;
-	size_t len;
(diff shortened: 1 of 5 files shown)


SebastianBoe (approved):

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.h
```diff
@@ -19,7 +19,7 @@ enum kmu_metadata_key_bits {
 
 typedef struct {
 	uint8_t key_usage_scheme: 2; /* value of @ref kmu_metadata_key_usage_scheme. */
-	uint8_t number_of_slots: 2;  /* Number of slots to push. */
+	uint8_t number_of_slots: 3;  /* Number of slots to push. */
```
SebastianBoe, line 22: I'm not used to this syntax.

We are encoding number_of_slots into a 3 bit field instead of a uint8_t?

Shall we just use a regular uint8_t to be future proof perhaps?

mswarowsky (approved):

PR #14541, opened 2024-03-18
samples: cellular: modem_shell: ppp: fixes

jhirsi:
Following fixes to mosh PPP functionality:
- Consecutive `ppp down` and _ppp up_ again wasn't working always correctly. Will work even better when Zephyr PR is included in ncs: https://github.com/zephyrproject-rtos/zephyr/pull/67731
With that also PC side notices `ppp down`.
- mosh ppp side is notified about the PDN detach.


trantanen:
The first commit and PR name has `cellula` in `samples/cellula/modem_shell`

tokangas (approved):

PR #14544, opened 2024-03-18
ext: iperf3: improving bi-dectional transfer

jhirsi:
Update patch for improving bi-directional transfer with iperf3.


jhirsi:
ext/iperf3 got external/mixed code style. PR got problems with compliance check.  I'll try to solve those.

ext/iperf3/iperf_client_api.c
```diff
@@ -104,9 +104,10 @@ iperf_create_streams(struct iperf_test *test, int sender)
 #endif /* HAVE_TCP_CONGESTION */
 
 	if (sender)
-	    FD_SET(s, &test->write_set);
+		FD_SET(s, &test->write_set);
 	else
-	    FD_SET(s, &test->read_set);
+		FD_SET(s, &test->read_set);
+	FD_SET(s, &test->err_set);
```
trantanen, line 110: If we cannot use curly braces in above if-statement, maybe even an empty line before this line to have a separation to the if-else.

trantanen (approved):

ext/iperf3/iperf_tcp.c
```diff
@@ -317,7 +317,7 @@ iperf_tcp_listen(struct iperf_test *test)
 		close(s);
 		freeaddrinfo(res);
 		errno = saved_errno;
-                i_errno = IESETBUF;
+                test->test->i_errno = IESETBUF;
```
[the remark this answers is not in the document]
jhirsi: Yes. That isn't in nrf compilation config and wasn't noticed. Fixing.

jhirsi:
All checks and tests passed now.

ext/iperf3/iperf_client_api.c
```diff
@@ -718,23 +769,80 @@ iperf_run_client(struct iperf_test * test)
...
+        }
+#endif
+
+#if defined(CONFIG_NRF_IPERF3_INTEGRATION)
+		iperf_time_now(&test_end_started_time);
+#endif
 	}
 	} /* test->state == TEST_RUNNING */
-	// If we're in reverse mode, continue draining the data
+	  // If we're in reverse or bidirectional mode, continue draining the data
```
tokangas, line 802: *nit* Extra spaces.

ext/iperf3/iperf_udp.c
```diff
@@ -593,14 +727,29 @@ iperf_udp_connect(struct iperf_test *test)
     }
 
 #if defined(CONFIG_NRF_IPERF3_INTEGRATION)
+    struct timeval tv;
+
 #ifdef SO_RCVTIMEO
     /* 30 sec timeout for a case when there is a network problem. */
```
tokangas, line 733: It would be better to set the timeout value before the `SO_RCVTIMEO` flagging. If `SO_SNDTIMEO` would be defined and `SO_RCVTIMEO` not, the timeout value would be uninitialized when setting the `SO_SNDTIMEO`. Although probably that will never be the case. 😃

tokangas (approved):
Not much to comment, only a couple of nits.

ext/iperf3/iperf_client_api.c
jhirsi: Fixed

ext/iperf3/iperf_udp.c
jhirsi: Changed.

jukkar (approved):

rlubos (approved):

PR #14546, opened 2024-03-18
app_event_manager: Improve OOM error handling

MarekPieta:
Change improves out of memory error handling for default application event allocator. The CONFIG_APP_EVENT_MANAGER_REBOOT_ON_EVENT_ALLOC_FAIL Kconfig option was introduced to allow user to select between system reboot or kernel panic.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -371,6 +371,11 @@ Other libraries
 
 * Added the :ref:`lib_uart_async_adapter` library.
 
+* :ref:`app_event_manager`:
+
+  * Added the :kconfig:option:`CONFIG_APP_EVENT_MANAGER_REBOOT_ON_EVENT_ALLOC_FAIL` Kconfig option.
+    The option allows to select between system reboot or kernel panic on event allocation failure for default event allocator.
+
 Common Application Framework (CAF)
 ----------------------------------
 
--- a/subsys/app_event_manager/Kconfig
+++ b/subsys/app_event_manager/Kconfig
@@ -12,6 +12,16 @@ menuconfig APP_EVENT_MANAGER
 
 if APP_EVENT_MANAGER
 
+config APP_EVENT_MANAGER_REBOOT_ON_EVENT_ALLOC_FAIL
+	bool "Reboot on event allocation failure"
+	depends on REBOOT
+	default y
+	help
+	  The default event allocator triggers assertion failure on event
+	  allocation failure. Then, depending on the value of this Kconfig
+	  option, the default allocator either triggers a system reboot or
+	  kernel panic.
+
 config APP_EVENT_MANAGER_SHOW_EVENTS
 	bool "Show events"
 	depends on LOG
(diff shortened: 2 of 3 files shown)


subsys/app_event_manager/Kconfig
```diff
@@ -12,6 +12,15 @@ menuconfig APP_EVENT_MANAGER
 
 if APP_EVENT_MANAGER
 
+config APP_EVENT_MANAGER_REBOOT_ON_EVENT_ALLOC_FAIL
+	bool "Reboot on event allocation failure"
+	depends on REBOOT
+	default y
```
ball-hayden, line 18: Given that `sys_reboot` does not flush logs, the `LOG_ERR` message will likely be lost before it is printed.

`k_panic` does flush logs, meaning the message will be visible before we reboot.

As such, I'd default this to `n`, as it's the least surprising option.
If an application wishes to "just" reboot, it can opt in.

MarekPieta: I prefer to keep it enabled by default to keep backwards compatibility (keep in mind that for debug configurations assertion failure would lead to log flush anyway), I can change it if needed, @pdunaj what's your opinion?

pdunaj: I would go with backward compatibility in this case.

pdunaj (approved):

subsys/app_event_manager/Kconfig
MarekPieta: @ball-hayden, is the approach acceptable for you?

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -357,7 +357,10 @@ Security libraries
 Other libraries
 ---------------
 
-|no_changes_yet_note|
+* :ref:`app_event_manager`:
+
+  * Added :kconfig:option:`CONFIG_APP_EVENT_MANAGER_REBOOT_ON_EVENT_ALLOC_FAIL` Kconfig option.
```
peknis, line 362:
```suggestion
  * Added the :kconfig:option:`CONFIG_APP_EVENT_MANAGER_REBOOT_ON_EVENT_ALLOC_FAIL` Kconfig option.
```

peknis (approved):
Approved with a minor nit.

subsys/app_event_manager/Kconfig
```diff
@@ -12,6 +12,15 @@ menuconfig APP_EVENT_MANAGER
...
+	help
+	  Default event allocator triggers assertion failure on event allocation
+	  failure. Then, depending on value of this Kconfig option, the default
+	  allocator either triggers system reboot or kernel panic.
```
kapi-no, line 22:
```suggestion
	  The default event allocator triggers assertion failure on event allocation
	  failure. Then, depending on the value of this Kconfig option, the default
	  allocator either triggers a system reboot or kernel panic.
```

kapi-no (approved):

subsys/app_event_manager/Kconfig
MarekPieta: @ball-hayden, please approve the PR and we are ready to merge :)

ball-hayden (approved):

PR #14555, opened 2024-03-18
samples: Bluetooth: mesh: Add nrf54l15 support in sensor samples

akredalen:
Add mesh sensor client and mesh sensor server sample support for the nrf54l15 board.


ludvigsj:
Have you checked that the (hardware) `temp` sensor actually reports sensible values in the server sample?

mia-ko (approved):

akredalen:
> Have you checked that the (hardware) `temp` sensor actually reports sensible values in the server sample?

Relative runtime in 0 to 20 degrees: 0 percent
Relative runtime in 20 to 25 degrees: 0 percent
Relative runtime in 25 to 30 degrees: 0 percent
Relative runtime in 30 to 100 degrees: 58 percent
366 second(s) since last presence detected
Ambient light level: 825
Chip temperature: 62.25

ludvigsj:
> > Have you checked that the (hardware) `temp` sensor actually reports sensible values in the server sample?
> 
> Relative runtime in 0 to 20 degrees: 0 percent Relative runtime in 20 to 25 degrees: 0 percent Relative runtime in 25 to 30 degrees: 0 percent Relative runtime in 30 to 100 degrees: 58 percent 366 second(s) since last presence detected Ambient light level: 825 Chip temperature: 62.25

Right, so it reports a sensible-ish value but the columns don't make much sense if it typically reports that high. I suggest to create an issue for this, so we can decide if we want to do something about it later. It's fine for now, since there is actually a value reported. It does change over time, right? It's not a dummy that's hardcoded to 62.25 in a driver somewhere, but actually reads the sensor?

akredalen:
A separate task is created for following up on the broken sensor driver for Moonlight. This PR should **not be blocked** and can be reviewed as is.

PavelVPV (approved):

ludvigsj (approved):

PR #14557, opened 2024-03-19
application: serial_lte_modem: Carrier-based factory bootstrap

junqingzou:
Notify lwm2m_carrier of carrier-specific factory bootstrap if needed, to support auto-connect/auto-register scenario.

Disable SLM IP services in NIDD overlay.


junqingzou:
No need of review for now.
For discussing with @kacperradoszewski for SoftBank LwM2M IOT.

junqingzou:
After discussion with Kacper, the specific carrier is detected automatically inside lwm2m_carrier lib so there is no need of explicit singnaling from application in order to perform factory bootstrap.

PR #14558, opened 2024-03-19
samples: net: Enable LTO for all samples

simensrostad:
Enable LTO for all samples to save space
LTO is now supported in Zephyr.


balaji-nordic (approved):

LuDuda:
> LTO is now supported in Zephyr.

Just for the record, indeed in our case LTO seems to work well, but it is not supported (production ready) in Zephyr. It is still experimental: https://github.com/nrfconnect/sdk-zephyr/blob/c0e08266a73496e70a6b44331b79e03bc55507b6/Kconfig.zephyr#L520

jtguggedal:
LTO comes at a cost, so it should probably be enabled on a per sample/sample configuration basis and only where absolutely needed. For reference: https://interrupt.memfault.com/blog/best-and-worst-gcc-clang-compiler-flags#-flto

PR #14559, opened 2024-03-19
net: fota_download: New API to support download configuration

junqingzou:
Add new fota_download() to allow application to config download_client.
Add new CONFIG_FOTA_DOWNLOAD_NEW_API to use the new API.


subsys/net/lib/fota_download/src/fota_download.c
```diff
@@ -423,16 +423,37 @@ int fota_download(const char *host, const char *file,
...
 {
-	uint32_t host_hash = 0;
-	uint32_t file_hash = 0;
-	int err = -1;
 	static int sec_tag_list_copy[CONFIG_FOTA_DOWNLOAD_SEC_TAG_LIST_SIZE_MAX];
-
 	struct download_client_cfg config = {
 		.pdn_id = pdn_id,
 		.frag_size_override = fragment_size,
 	};
```
sigvartmh, line 430: If we have the final goal of deprecating fota_download and eventually moving over to `fota_v2` API, I think we should have `if not set`, set `pdn+fragsize`, and have these be the same. It would give the least API change if we eventually are able to deprecate `fota_download`. However this is just an opinion and not something that really needs to be considered.

```C
config->pdn_id = pdn_id;
config->frag_size_override = fragment_size;
```

sigvartmh (approved):
Overall I'm okay with this change :) I think we should come up with a deprecation + migration plan for the current `fota_download` API. As we keep rough patching issues with the current abstraction.

subsys/net/lib/fota_download/src/fota_download.c
junqingzou: The new API can replace below APIs
.fota_download()
.fota_download_any()
.fota_downlaod_start()
.fota_download_start_with_image_type()

include/net/fota_download.h
```diff
@@ -146,6 +146,28 @@ int fota_download(const char *host, const char *file, const int *sec_tag_list,
...
 		  const enum dfu_target_image_type expected_type);
 
+/**@brief Download the given file with the specified image type from the given host.
+ *
+ * Validate that the file type matches the expected type before proceeding with the download.
+ * When the download is complete, the secondary slot of MCUboot is tagged as having
+ * valid firmware inside it. The completion is reported through an event.
+ *
+ * URI parameters (host and file) are not copied, so pointers must stay valid
+ * until download is finished.
```
SeppoTakalo, line 156:
```suggestion
 * URI parameters (host and file) and sec_tag lists are not copied, so pointers must stay valid
 * until download is finished.
```

include/net/fota_download.h
```diff
@@ -146,6 +146,28 @@ int fota_download(const char *host, const char *file, const int *sec_tag_list,
...
+ *
+ * @param host Name of host to start downloading from. Can include scheme
+ *             and port number, for example https://google.com:443
+ * @param file Path to the file you wish to download. See fota_download_any()
+ *             for details on expected format.
```
SeppoTakalo, line 161:
```suggestion
 * @param host URI of the host to connect to.
 *                          Can include scheme, port number and full file path, defaults to
 *                          HTTP or HTTPS if no scheme is provided.
 * @param file   File to download or NULL if path is already provided in host parameter.
```

Follow the same documentation as `download_client_get()`. So full URI is OK to be fed in the `host` parameter.

include/net/fota_download.h
```diff
@@ -146,6 +146,28 @@ int fota_download(const char *host, const char *file, const int *sec_tag_list,
...
+ * @param config Download client configuration options.
```
SeppoTakalo, line 162:
```suggestion
 * @param config Download client configuration options. Shallow copied, so sec_tag list pointer must say valid.
```

include/net/fota_download.h
```diff
@@ -146,6 +146,28 @@ int fota_download(const char *host, const char *file, const int *sec_tag_list,
...
+ * @param expected_type Type of firmware file to be downloaded and installed.
+ *
+ * @retval 0	     If download has started successfully.
+ * @retval -EALREADY If download is already ongoing.
+ *                   Otherwise, a negative value is returned.
+ */
+int fota_download_v2(const char *host, const char *file, const struct download_client_cfg *config,
```
SeppoTakalo, line 169:
```suggestion
int fota_download_cfg(const char *host, const char *file, const struct download_client_cfg *config,
```

We don't version our APIs other than what the SDK version is.
So I recommend to use just `_cfg` postfix instead of `_v2`

subsys/net/lib/fota_download/src/fota_download.c
```diff
@@ -423,16 +423,37 @@ int fota_download(const char *host, const char *file,
...
+
+		config.sec_tag_count = sec_tag_count;
+		config.sec_tag_list = sec_tag_list_copy;
+
+		if (!is_ip_address(host)) {
+			config.set_tls_hostname = true;
+		}
+	}
+
+	return fota_download_v2(host, file, &config, expected_type);
```
SeppoTakalo, line 447: We have a bug here.

sec_tag_list_copy is overwritten without checking if that is already in use.
Why exactly do we copy, when the `fota_download_cfg()` does not copy it? The behaviour here is different.

And because we don't copy URI, I recommend that we do the same for sec_tag list. Otherwise the API is a bit mixed up because it has two different memory handling strategies.

SeppoTakalo (changes requested):
I like the approach of taking download_client configuration structure in API.
This was overlooked when the API was last time refactored and therefore address family have already been proposed to be exposed here as well.

Few changes I would like to do:
* Follow what Download client does and allow full URI in host.
* Don't claim V2, just use `_cfg` prefix.

include/net/fota_download.h
junqingzou: It has already been documented [here ](https://github.com/nrfconnect/sdk-nrf/blob/main/include/net/download_client.h#L105)but I would suggest to change it to be `const int sec_tag_list[NRF_SOCKET_TLS_MAX_SEC_TAG_LIST_SIZE];`, i.e. max 7 integers

include/net/fota_download.h
junqingzou: I would suggest below:
.define a new Kconfig, e.g. `CONFIG_FOTA_DOWNLOAD_NEW_API`, which is default 'y'. We use this flag to clean up all legacy APIs and provide a new API of `fota_download()`, apply it to all known clients in NCS;
.mark all old APIs to be deprecated in next NCS version and remove them in next next NCS version;
.Remove the Kconfig in next next NCS version.

include/net/fota_download.h
junqingzou: Notably existing fota_download() does not allow `file==NULL`

SeppoTakalo:
Please, no new Kconfig.

I think this PR is now going wrong direction.
I suggested few simple things that would have been OK.

But adding more things to configure means more things to go wrong. I cannot accept any conditions on the public headers.
How do you thing the Doxygen is going to render it?
The API should be stable and unconditional.

We cannot block anyone using the "old" API. That will live as long as it has been deprecated for few releases. And there has been no reason to deprecate it.

junqingzou:
Me not a fan of CONFIGs, either.
If no deprecation plan then no need of the new CONFIG.
Also old configs like below are also redundent:
CONFIG_FOTA_DOWNLOAD_SEC_TAG_LIST_SIZE_MAX, the max has been defined by libmodem which is 7.
CONFIG_FOTA_DOWNLOAD_HOST_NAME_LENGTH, there is CONFIG_DOWNLOAD_CLIENT_MAX_HOSTNAME_SIZE 
CONFIG_FOTA_DOWNLOAD_FILE_NAME_LENGTH, there is CONFIG_DOWNLOAD_CLIENT_MAX_FILENAME_SIZE
etc.

SeppoTakalo:
> Me not a fan of CONFIGs, either. If no deprecation plan then no need of the new CONFIG. Also old configs like below are also redundent: CONFIG_FOTA_DOWNLOAD_SEC_TAG_LIST_SIZE_MAX, the max has been defined by libmodem which is 7. CONFIG_FOTA_DOWNLOAD_HOST_NAME_LENGTH, there is CONFIG_DOWNLOAD_CLIENT_MAX_HOSTNAME_SIZE CONFIG_FOTA_DOWNLOAD_FILE_NAME_LENGTH, there is CONFIG_DOWNLOAD_CLIENT_MAX_FILENAME_SIZE etc.

Ah.. true.
That should be a follow up task for someone. I did not even know about these configs, they seem pretty irrelevant.

junqingzou:
Drop the PR as the required modification is more than original understanding.
Leave the work for others with better understanding of the need.

PR #14569, opened 2024-03-19
bluetooth: controller: prepare hci_driver to separate scan/init mem

olivier-le-sage:
Prepares the hci driver to allocate memory for the initiator

### DIFF
--- a/subsys/bluetooth/controller/hci_driver.c
+++ b/subsys/bluetooth/controller/hci_driver.c
@@ -137,6 +137,12 @@ BUILD_ASSERT(!IS_ENABLED(CONFIG_BT_PERIPHERAL) ||
 	#define SDC_SCAN_SIZE 0
 #endif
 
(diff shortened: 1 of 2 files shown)


subsys/bluetooth/controller/hci_driver.c
```diff
@@ -137,6 +137,12 @@ BUILD_ASSERT(!IS_ENABLED(CONFIG_BT_PERIPHERAL) ||
 	#define SDC_SCAN_SIZE 0
 #endif
 
+#if defined(SDC_MEM_INIT)
```
rugeGerritsen, line 140: I would be happy if we could use "INITIATOR" instead of "INIT" to avoid confusion

rugeGerritsen (approved):

olivier-le-sage:
Replaced with a kconfig option as discussed offline.

rugeGerritsen (approved):

PR #12588, opened 2023-10-09
core_fw: Add remote_radio firmware code

dchat-nordic:
Add remote_radio firmware code for network core.
This image serves purpose as universal network core image for hci_rpmsg rpc_host and IEEE 802.15.4 remote image.

PR #14576, opened 2024-03-19
sdfw_services: add sdfw_services framework

anhmolt:
* Adds `sdfw_services` framework for invoking secure domain services on `nRF54h20`.
* Adds `ssf_client` sample showing how to use services from application and radio core.
* Small fixes for nRF RPC.

_Note:_ Framework does not work with already released SecDomFW binaries. Most likely because of incompatibilities/changes in ICMsg shared memory. Address of read and write indices at the start of shared memory seems to deviate and ICMsg binding fails as a result. ~~Investigation is ongoing.~~

A new release of SecDomFW will be needed for the IPC to work.


katgiadla (dismissed):
Please specify the responsible person for this test and sample (add github nicknames in the file `CODEOWNERS`)

greg-fer (changes requested):
This looks like a doc domain of @FrancescoSer  and @annwoj , given the relation to nRF54.

Also, I think we need a mention about this in https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/security.html and an entry in the changelog.

It also looks like a candidate for one of the core firmware components discussed here: https://nordicsemi.atlassian.net/browse/NCSDK-10551 -- @carlescufi , @pdunaj , @grochu , @wbober , what do you think? The only existing core firmware component has been integrated as an application in https://github.com/nrfconnect/sdk-nrf/pull/12588 given that there were no other core firmware components at that point. Should this SDFW and IPC Radio be grouped together?

samples/sdfw/ssf_client/README.rst
```diff
@@ -0,0 +1,60 @@
...
+
+.. table-from-rows:: /includes/sample_board_rows.txt
+   :header: heading
+   :rows: nrf54h20pdk_nrf54h20_cpuapp, nrf54h20pdk_nrf54h20_cpurad
+
+Overview
+********
+
+The sample shows how to use Secure Domain services by invoking some of the available services.
+Available Secure Domain Services can be found in :file:`nrf/include/sdfw_services`.
```
FrancescoSer, line 26: Could we add a file where we describe in detail these 8 services listed there?

osaether (approved):
LGTM but content itself is not reviewed in this round.

9gcity (approved):

hakonfam (approved):
Locations of files looks good to me. Content itself is not reviewed in this round.

pdunaj:
> This looks like a doc domain of @FrancescoSer and @annwoj , given the relation to nRF54.
> 
> Also, I think we need a mention about this in https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/security.html and an entry in the changelog.
> 
> It also looks like a candidate for one of the core firmware components discussed here: https://nordicsemi.atlassian.net/browse/NCSDK-10551 -- @carlescufi , @pdunaj , @grochu , @wbober , what do you think? The only existing core firmware component has been integrated as an application in #12588 given that there were no other core firmware components at that point. Should this SDFW and IPC Radio be grouped together?

As for core components, this is framework not a firmware image. This does not fit there.
Regarding other changes - this change is on a critical path. Unless something is very wrong, I highly recommend to merge it and fix whatever is needed on the go.
Assuming the above comment could you let it in @greg-fer ? If there are some remaining activities could these be placed under a jira ticket and fixed before June?

pdunaj:
@doki-nordic please have a look at the changes after this get's in.

jonathannilsen (approved):

PR #14577, opened 2024-03-19
Bluetooth: Mesh: Make metadata script handle multiple comp data

ludvigsj:
This makes the metadata extraction script handle applications with multiple instances of composition data, where the composition data to use is picked at runtime.

When more than one instance is found, no encoded metadata is printed as part of the build output, as users would need to look at the JSON anyways to decide which composition data to use when doing the update.


ludvigsj:
DNM for now, I found a failure when the multiple composition data is in an array, will have to investigate

ludvigsj:
Fixed the array case. Open for review and merging again

PavelVPV (approved):

scripts/bluetooth/mesh/mesh_dfu_metadata.py
```diff
@@ -348,16 +346,18 @@ def parse_comp_data(elf_path, kconfigs):
...
+        addrs = find_comp_data_from_dwarf(elf_path)
+        return [comp
+                for addr in addrs
+                for comp in read_comp_data(elf_path, addr, kconfigs)]
+    except Exception as err:
         raise Exception("Failed to extract composition data from .elf file") from err
 
 def encoded_metadata_get(version, comp, binary_size):
-    core_type = 1 # FIX ME: For now, hardcoded to application core
+    core_type = 1 # FIXME: For now, hardcoded to application core
```
alxelax, line 360: If you really want to fix it, then it should have JIRA ticket. Otherwise, there is no sense in such comments. It will be permanent FIXME.

ludvigsj: This was here before this PR, I simply removed the space. @Andrewpini : since you added this, you can decide whether to add a JIRA issue or if I should remove it here.

Andrewpini: I think we can just remove the FIXME. We will most likely have to address this at some point, but the comment is not neccesary.

Andrewpini (approved):

scripts/bluetooth/mesh/mesh_dfu_metadata.py
PavelVPV: Shouldn't we include core_type into json file even though it is hardcoded now?

ludvigsj: You mean as a separate field? Or to keep it in the encoded metadata? If you mean the latter, I'm only planning to remove the comment here, keep the rest. If you mean to add a separate field to the json object, please add a separate issue for this. It's not part of making the script handle multiple comp data which is what I'm fixing here.

PavelVPV: NCSDK-26810

alxelax (approved):

doc/nrf/protocols/bt/bt_mesh/dfu_over_bt_mesh.rst
```diff
@@ -205,14 +205,15 @@ Additionally, the metadata string required by the ``mesh models dfu slot add`` c
     Full metadata written to: APPLICATION_FOLDER\build\zephyr\dfu_application.zip
 
 .. note::
-   Currently this script only supports applications that contain a single composition data structure, meaning that applications containing multiple composition data structures must rely on manual generation of the metadata string(s).
-   It is also required that the composition data is declared with the const-qualifier.
+   It is required that the composition data is declared with the const-qualifier.
+   If the application contains more than one composition data structure (for instance when the structure to use is picked at runtime), the script will not print any encoded metadata.
+   In this case, the JSON file must be used to find the encoded metadata matching the composition data to be used by the device after the update.
```
mia-ko, line 210:
```suggestion
   It is required that the Composition Data is declared with the ``const`` qualifier.
   If the application contains more than one Composition Data structure (for example, when the structure to be used is picked at runtime), the script will not print any encoded metadata.
   In this case, use the JSON file to find the encoded metadata matching the Composition Data to be used by the device after the update.
```
I see we use ``const`` qualifier in Zephyr, so probably ok to be consistent.

mia-ko (changes requested):

mia-ko (approved):

PR #14579, opened 2024-03-19
wifi: Add a new NW selection criterion

krish2718:


sachinthegreen (approved):

modules/hostap/Kconfig
```diff
@@ -247,4 +247,22 @@ endif
...
+	help
+		Select the network selection method for the supplicant.
+
+config WPA_SUPP_NW_SEL_THROUGHPUT
+	bool "Throughput based network selection"
+	help
+	  Select the network based on throughput.
+
+config WPA_SUPP_NW_SEL_RELIABILITY
+	bool "Reliability based network selection"
```
markaj-nordic, line 263: What is the reliability indicator in this case?

markaj-nordic (commented):
What would be the exemplary use case? Scanning with multiple networks with the same SSID?

modules/hostap/Kconfig
krish2718: Reliable connection i.e., no disconnections. E.g., choosing a far network over a near network.

krish2718:
> What would be the exemplary use case? Scanning with multiple networks with the same SSID?

Yeah, Enterprise or mesh.

jukkar (approved):

PR #14587, opened 2024-03-20
Bluetooth: Mesh: Remove unneeded dependency from metadata extraction

ludvigsj:
Make it possible to extract metadata at build time even if BT_MESH_DFU_METADATA is not enabled in the image. There is no need to have this enabled in the new firmware for it to make sense, as long as it is enable in the firmware currently running on the target. Removing the dependency allows things like removing Mesh DFU support by performing a Mesh DFU update, or changing metadata schemes by performing a DFU update.

Also remove the default y for BT_MESH_DFU_CLI. It makes no sense to enable it by default for distributor images, unless they support self-update, in which case they would also enable BT_MESH_DFU_SRV.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -74,7 +74,11 @@ Bluetooth® LE
 Bluetooth Mesh
 --------------
 
-|no_changes_yet_note|
+* Updated:
+
+  * The Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA_ON_BUILD` to no longer depend on the Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA`.
+  * The Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_CLI` to no longer enable the Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA_ON_BUILD` by default.
+    The Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA_ON_BUILD` can still be manually enabled.
 
 Matter
 ------
--- a/subsys/bluetooth/mesh/Kconfig
+++ b/subsys/bluetooth/mesh/Kconfig
@@ -129,9 +129,8 @@ endchoice
 
 config BT_MESH_DFU_METADATA_ON_BUILD
 	bool "Generate DFU metadata on build"
-	default y if BT_MESH_DFU_SRV || BT_MESH_DFU_CLI
+	default y if BT_MESH_DFU_SRV
 	depends on SIGN_IMAGES
-	depends on BT_MESH_DFU_METADATA
 	help
 	  Generate a JSON file containing all DFU related metadata upon
 	  application build. The JSON will be added to dfu application zip


Andrewpini (approved):

ludvigsj:
> Release notes?

I thought they perhaps were not needed, were you thinking about the changed default or the fact that a limitation has been removed? I can add a small note

PavelVPV:
> > Release notes?
> 
> I thought they perhaps were not needed, were you thinking about the changed default or the fact that a limitation has been removed? I can add a small note

Yes, seems reasonable to let know customers who already met or will meet the limitation that it is now removed.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -74,7 +74,9 @@ Bluetooth® LE
 Bluetooth Mesh
 --------------
 
-|no_changes_yet_note|
+* Updated:
+
+  * The Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA_ON_BUILD` to no longer depend on the Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA`.
```
PavelVPV, line 79: Maybe add something like this:
Now, by default the encoded metadata will be generated for every firmware that has the :kconfig:option:`CONFIG_BT_MESH_DFU_SRV` model enabled.
?

ludvigsj: That was already the case. The change it that it will no longer be generated by default if you have DFU_CLI enabled. But I can add a sentence about that.

PavelVPV: Ok, maybe it is fine now. When read this I feel I need more context to understand the change.

mia-ko: I feel tempted to say that you maybe should add a bit more text, so it is more clear to customers what this change implies. E.g.
Now, the DFU metadata will no longer be generated by default when the Firmware Update Client model is enabled (:kconfig:option:`CONFIG_BT_MESH_DFU_CLI` set), but only when the Firmware Update Server (:kconfig:option:`CONFIG_BT_MESH_DFU_SRV`) is enabled.

mia-ko (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
ludvigsj: But that is a different change. There are two customer-facing changes in this PR:

1. Enabling CONFIG_BT_MESH_DFU_CLI no longer enables metadata generation by default
2. Metadata generation can now be enabled without enabling CONFIG_BT_MESH_DFU_METADATA

Should I make two bullets for these? Or combine them into one? Only mention one or the other? It seems both of your suggestions seem to conflate the two changes, but they are completely independent from each other.

ludvigsj: With regards to "what enables the metadata generation", that hasn't changed, other than the change mentioned in the first bullet about the DFU Client no longer enabling it by default. Shouldn't the changelog be about changes since last time?

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -74,7 +74,11 @@ Bluetooth® LE
...
+  * The Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_CLI` to no longer enable the Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA_ON_BUILD` by default.
+    :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA_ON_BUILD` can still be enabled manually.
```
mia-ko, line 81:
```suggestion
    The Kconfig option :kconfig:option:`CONFIG_BT_MESH_DFU_METADATA_ON_BUILD` can still be manually enabled.
```

mia-ko (changes requested):

mia-ko (approved):

PavelVPV (approved):

PR #14564, opened 2024-03-19
lib: nrf_modem: Initialize CFUN hooks with modem library

eivindj-nordic:
This fixes an issue if the modem library is initialized on SYS_INIT at kernel level and makes calls to AT commands before the application level initialization is done.

PR #14590, opened 2024-03-20
lib: nrf_modem: Initialize CFUN hooks with modem library

eivindj-nordic:
This fixes an issue if the modem library is initialized on SYS_INIT at kernel level and makes calls to AT commands before the application level initialization is done.

Fix for v2.6-branch, same as #14564 to main branch.


shanthanordic (changes requested):
@eivindj-nordic  please close this PR and add backport label to PR that is posted for main.

PR #14599, opened 2024-03-21
samples: cellular: http_update: Support 1.3.6 on nRF9160

SeppoTakalo:
The modem FW image is already in server, so we should support it.


eivindj-nordic:
We should update the `modem_full_update` sample as well, the v1.3.6 cbor is also in the server. Will you add it here or should I open another PR?

eivindj-nordic (approved):

rlubos (approved):

PR #14600, opened 2024-03-21
lwm2m: Fix supported protocols list for FOTA

SeppoTakalo:
## net: lwm2m_client_utils: Fix supported protocols list for FOTA

FOTA object has list of supported protocols.
We assumed that when CA chain is present in certain sec_tag,
we can use HTTPS or CoAPS. This is OK.

But AVSystem uses mutual DTLS authentication on CoAPS so
the sec_tag we use, might not have CA chain at all.
So when only PSK credentials are present, claim that we
support CoAPS but not HTTPS.

## samples: lwm2m_client: Use same security tag for FOTA and DTLS

AVSystem FOTA defaults to CoAPS so use same security tag where our
DTLS credentials are stored.

### DIFF
--- a/subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
+++ b/subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
@@ -1010,12 +1010,12 @@ static void lwm2m_firmware_object_pull_protocol_init(int instance_id)
 #endif
 }
 
-static bool modem_has_credentials(int sec_tag)
+static bool modem_has_credentials(int sec_tag, enum modem_key_mgmt_cred_type cred_type)
 {
 	bool exist;
 	int ret;
 
-	ret = modem_key_mgmt_exists(sec_tag, MODEM_KEY_MGMT_CRED_TYPE_CA_CHAIN, &exist);
+	ret = modem_key_mgmt_exists(sec_tag, cred_type, &exist);
 	if (ret < 0) {
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


juhaylinen (approved):

subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
```diff
@@ -1032,9 +1032,14 @@ static void lwm2m_firware_pull_protocol_support_resource_init(int instance_id)
 		lwm2m_firmware_object_pull_protocol_init(instance_id);
 	}
 
-	if (modem_has_credentials(CONFIG_LWM2M_CLIENT_UTILS_DOWNLOADER_SEC_TAG)) {
-		/* Enable non-security &  Security protocols for download client */
+	int tag = CONFIG_LWM2M_CLIENT_UTILS_DOWNLOADER_SEC_TAG;
+
+	if (modem_has_credentials(tag, MODEM_KEY_MGMT_CRED_TYPE_CA_CHAIN)) {
+		/* CA chain means that HTTPS and CoAPS might work */
 		supported_protocol_count = 4;
```
tomi-font, line 1039: These numbers are a bit magic, I would be more comfortable seeing at least some reference to `pull_protocol_support` here (e.g. with `ARRAY_SIZE()`).

tomi-font (changes requested):
Just a small nit (in the Kconfig fragments) for readability.

subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
SeppoTakalo: Refactored a bit.. Can you re-check does it make more sense?

PR #14605, opened 2024-03-22
802154_sniffer: Add Wireshark script and update readme.

dberliner:
The current example does not work with wireshark because the shell echos too many `/r/n` characters. The behavior of the shell is to echo back the characters, including the newline, and an additional `/r/n` . The version of `nrf802154_sniffer,py` matches the shells behavior.. I would have liked to modify the program but it did not seem trivial to change that behavior.

The diff between the wireshark scripts here and the ones in https://github.com/NordicSemiconductor/nRF-Sniffer-for-802.15.4 are
```
74c74
<     SNIFFER_802154_PID = 0x154A
---
>     SNIFFER_802154_PID = 0x154B
326c326,327
<             self.serial.write(b'\r\n')
---
>             # The shell will add this \r\n for us
>             #self.serial.write(b'\r\n')
370a372,374
>
>             # Throw out the first /r/n on connect
>             self.serial.read(2)
```

### DIFF
--- a/samples/peripheral/802154_sniffer/README.rst
+++ b/samples/peripheral/802154_sniffer/README.rst
@@ -147,6 +147,10 @@ After programming the sample to your development kit, complete the following ste
 
    The **LED 1** will start blinking with longer intervals.
 
+Running with Wireshark
+==================
+The shell and USB descriptor are a little different in this application and the official 802.15.4 sniffer binary. For that purpose a slightly modified version of `nrf_802154_sniffer.py`. With the modified script the same instructions apply.
+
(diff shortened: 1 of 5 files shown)


e-rk:
Thanks for the PR. I would like to keep the scripts out of the NCS repository and have pre-built images and Wireshark scripts in the https://github.com/NordicSemiconductor/nRF-Sniffer-for-802.15.4 repository. I'll try to merge [this](https://github.com/NordicSemiconductor/nRF-Sniffer-for-802.15.4/pull/57) at the end of this week.

PR #14606, opened 2024-03-22
applications: nrf5340_audio: Update BASE decode feature

rick1082:
Check and config codec setting in both subgroup/BIS level. OCT-2953


gWacey (approved):
It would be good somewhere to trap if there is an error in the codec configuration once completed.

alexsven (approved):

rick1082:
> It would be good somewhere to trap if there is an error in the codec configuration once completed.

We will handle that in another PR, since that will need a middle level change.

PR #14607, opened 2024-03-22
Revert "applications: asset_tracker_v2: Add missing external flash re…

jorgenmk:
…ference"

This reverts commit a32b68ad10985ee80b22fab9ce00d3cc7535b0cd.

### DIFF
--- a/applications/asset_tracker_v2/boards/nrf9160dk_nrf9160_ns.overlay
+++ b/applications/asset_tracker_v2/boards/nrf9160dk_nrf9160_ns.overlay
@@ -41,17 +41,4 @@
 			label = "Green LED 4";
 		};
 	};
-
-	/* Configure partition manager to use mx25r64 as the external flash */
-	chosen {
-		nordic,pm-ext-flash = &mx25r64;
-	};
-
-	aliases {
-		ext-flash = &mx25r64;
-	};
-};
-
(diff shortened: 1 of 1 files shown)


bjda (approved):

gregersrygg (approved):
This looks correct to me, unless there is some recent change that inverts the board revision logic.

@de-nordic @rlubos The external flash only exists on board revisions 0.14.0 and newer, so we only enable the external flash in [boards/nrf9160dk_nrf9160_ns_0_14_0.overlay](https://github.com/nrfconnect/sdk-nrf/blob/main/applications/asset_tracker_v2/boards/nrf9160dk_nrf9160_ns_0_14_0.overlay). The 0.14.0 revision is now the default, so it should not be necessary to specify the board revision to enable external flash when building. Why was this change necessary in the upmerge? Has there been recent changes upstream changing this logic?

rlubos:
> Why was this change necessary in the upmerge?

You can see in the CI results why. If the proposed fix was not a correct one, we need to find an alternative.

de-nordic:
Are we pursuing this further?

jorgenmk:
This commit was cherry-picked into upmerge: https://github.com/nrfconnect/sdk-nrf/pull/14626, closing.

PR #14609, opened 2024-03-22
drivers: uart_nrf_sw_lpuart: Fix interrupt driven case

nordic-krch:
Driver was calling isr function from unknown context when uart_irq_[rx,tx]_enable was called. It could lead to failures because driver assumes that isr function won't be interrupt by the same isr function.

A solution proposed here is to use k_timer handler context which is called from RTC1 interrupt context. Assert is ensuring that RTC1 and UART interrupt priorities are the same. Timer is called from the shortest possible timeout and from that context isr function is called.

### DIFF
--- a/drivers/serial/uart_nrf_sw_lpuart.c
+++ b/drivers/serial/uart_nrf_sw_lpuart.c
@@ -56,6 +56,8 @@ struct lpuart_int_driven {
 	bool tx_enabled;
 	bool rx_enabled;
 	bool err_enabled;
+
+	struct k_timer trampoline_timer;
 };
 #endif
 
@@ -821,6 +823,24 @@ static void int_driven_rx_feed(const struct device *dev,
 	__ASSERT_NO_MSG(err >= 0);
 }
 
+static void trampoline_timeout(struct k_timer *timer)
+{
+	const struct device *dev = k_timer_user_data_get(timer);
+	struct lpuart_data *data = get_dev_data(dev);
+
+	data->int_driven.callback(dev, data->int_driven.user_data);
+
+	/* If RX was disabled and there were pending data, it may have been
+	 * processed above after re-enabling. If whole data has been processed
+	 * we must feed the buffer back to the uart to allow reception of the
+	 * next packet.
+	 */
+	if (!int_driven_rd_available(data) && data->rx_state == RX_BLOCKED) {
+		/* Whole packet read, RX can be re-enabled. */
+		int_driven_rx_feed(dev, data);
+	}
+}
+
 static void int_driven_evt_handler(const struct device *lpuart,
 				   struct uart_event *evt,
 				   void *user_data)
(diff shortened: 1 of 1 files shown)


nordic-krch:
@anangl  can you take a look?

drivers/serial/uart_nrf_sw_lpuart.c
```diff
@@ -917,7 +937,8 @@ static void api_irq_tx_enable(const struct device *dev)
 
 	data->int_driven.tx_enabled = true;
 	if (data->tx_buf == NULL) {
-		data->int_driven.callback(dev, data->int_driven.user_data);
+		/* We need to move to the interrupt context of the same priority as UARTE. */
+		k_timer_start(&data->int_driven.trampoline_timer, K_USEC(1), K_NO_WAIT);
```
anangl, line 941: Wouldn't `K_NO_WAIT` give even shorter delay than `K_USEC(1)`?

drivers/serial/uart_nrf_sw_lpuart.c
```diff
@@ -1174,6 +1189,12 @@ static const struct uart_driver_api lpuart_api = {
 		     DT_IRQ(GPIOTE_NODE(gpio_node), priority),		  \
 		     "UARTE and GPIOTE interrupt priority must match.");))
 
+#if CONFIG_NRF_SW_LPUART_INT_DRIVEN
+/* UARTE interrupt priority must be the same as system timer (RTC1) priority. */
+BUILD_ASSERT(DT_IRQ(DT_PARENT(DT_NODELABEL(lpuart)), priority) ==
+		DT_IRQ(DT_NODELABEL(rtc1), priority));
```
anangl, line 1195: And what if the system timer uses GRTC?

JordanYates:
@nordic-krch any chance of resurrecting this? This PR took my nRF9151 + nRF54L15 HCI link from crashing every 30 seconds to stable as far as I can tell.
I used `K_NO_WAIT` as the timer delay as suggested by @anangl and added the following for the interrupt validation
```
#if CONFIG_NRF_SW_LPUART_INT_DRIVEN
/* UARTE interrupt priority must be the same as system timer priority. */
#if CONFIG_NRFX_GRTC
BUILD_ASSERT(DT_IRQ(DT_PARENT(DT_NODELABEL(lpuart)), priority) ==
	     DT_IRQ(DT_NODELABEL(grtc), priority));
#else
BUILD_ASSERT(DT_IRQ(DT_PARENT(DT_NODELABEL(lpuart)), priority) ==
	     DT_IRQ(DT_NODELABEL(rtc1), priority));
#endif
#endif
```
Complete commit here: https://github.com/Embeint/sdk-nrf/commit/c8b5f3902259f530679bbb4dd4c1af7f0cfd431f

drivers/serial/uart_nrf_sw_lpuart.c
nordic-krch: comments applied in #23654 (i could not reopen that one).

drivers/serial/uart_nrf_sw_lpuart.c
nordic-krch: comments applied in #23654 (i could not reopen that one).

nordic-krch:
@JordanYates I've opened #23654

PR #14615, opened 2024-03-22
net: lib: nrf_cloud: Improve A-GNSS data parsing robustness

tokangas:
If A-GNSS data contained an invalid element count, the parser was left into invalid state and no error was logged. Because of this further parsing attempts failed. Also, the end of parsing was not detected in a successful case and the "Parsing finished" debug log was missing.

### DIFF
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_agnss.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_agnss.c
@@ -465,6 +465,13 @@ static size_t get_next_agnss_element(struct nrf_cloud_agnss_element *element,
 	 * each element.
 	 */
 	if (elements_left_to_process == 0) {
+		element_type = NRF_CLOUD_AGNSS__TYPE_INVALID;
+
+		if (buf_len == 0) {
+			/* No more data to parse. */
+			return 0;
+		}
+
 		/* Check that there's enough data for type and count. */
 		if (buf_len < NRF_CLOUD_AGNSS_BIN_TYPE_SIZE + NRF_CLOUD_AGNSS_BIN_COUNT_SIZE) {
 			LOG_ERR("Unexpected end of data");
@@ -532,6 +539,7 @@ static size_t get_next_agnss_element(struct nrf_cloud_agnss_element *element,
 	default:
 		LOG_DBG("Unhandled A-GNSS data type: %d", element->type);
 		elements_left_to_process = 0;
+		element_type = NRF_CLOUD_AGNSS__TYPE_INVALID;
 		return 0;
 	}
 
@@ -539,6 +547,7 @@ static size_t get_next_agnss_element(struct nrf_cloud_agnss_element *element,
 	if (buf_len < len) {
 		LOG_ERR("Unexpected end of data");
 		elements_left_to_process = 0;
+		element_type = NRF_CLOUD_AGNSS__TYPE_INVALID;
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


jayteemo (approved):

jayteemo:
would you consider this a bug fix?  if so, a changelog entry would be a good idea.

subsys/net/lib/nrf_cloud/src/nrf_cloud_agnss.c
```diff
@@ -465,6 +465,13 @@ static size_t get_next_agnss_element(struct nrf_cloud_agnss_element *element,
 	 * each element.
 	 */
 	if (elements_left_to_process == 0) {
+		/* If there's no more data, reset the parser state. */
+		if (buf_len == 0) {
+			elements_left_to_process = 0;
```
plskeggs, line 470: This line is redundant -- the if statement on line 467 cannot be entered unless elements_left_to_process == 0 already.

plskeggs (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_agnss.c
```diff
@@ -465,6 +465,13 @@ static size_t get_next_agnss_element(struct nrf_cloud_agnss_element *element,
...
+			element_type = NRF_CLOUD_AGNSS__TYPE_INVALID;
+			return 0;
+		}
```
glarsennordic, line 473: I don't fully follow the operation of the AGNSS/PGNSS parsers, but:

It seems like 

```c
if (buf_len < NRF_CLOUD_AGNSS_BIN_TYPE_SIZE + NRF_CLOUD_AGNSS_BIN_COUNT_SIZE) {
			LOG_ERR("Unexpected end of data");
			return 0;
}
```

Also detects this?

Would placing `element_type = NRF_CLOUD_AGNSS__TYPE_INVALID;` inside that branch accomplish the same effect?

subsys/net/lib/nrf_cloud/src/nrf_cloud_agnss.c
tokangas: Ah, yes. I was in a bit of a hurry. The main point is the change on line 606, it's the only change that is actually needed. Also, setting the `element_type` to `NRF_CLOUD_AGNSS__TYPE_INVALID` is not mandatory to fix the problem I found, but I thought it would be nice to re-initialize both values. This is the second similar problem I have ran into and setting `element_type` to `NRF_CLOUD_AGNSS__TYPE_INVALID` should prevent all similar problems and make the parsing error visible.

subsys/net/lib/nrf_cloud/src/nrf_cloud_agnss.c
tokangas: I think it would be good to also set `element_type` to invalid in the unexpected end of data case, I missed that one.

These two if's are for different cases. The first one is for the success case, where all data was parsed and there's no data left. The second one is for the case where there's still data left, but not enough for the type and count fields. In both cases the function returns 0 to indicate that no next element was found, but in the latter case there should be an error trace.

`element_type` can be set before the if's and `elements_left_to_process` is already zero, so no need to set it. I'll update the implementation.

tokangas:
> would you consider this a bug fix? if so, a changelog entry would be a good idea.

It's a bug fix, but as long as cloud sends properly formatted data, it will not be visible to the users. `modem_shell` has a command for injecting assistance data from the command line. We use that in our testing and it's easier to get the input data incorrect. It shouldn't impact regular users, so maybe a changelog entry isn't mandatory.

PR #14617, opened 2024-03-22
actions: oss: enable tf-m oss-history check

mswarowsky:
Re-enable the  trusted-firmware-m after it had to be disabled here: in commit 6f98e9d810664977a9a5eee94fbb5697b51bfcc0.

It works again after the upmerge

Ref: NCSDK-26099

### DIFF
--- a/.github/workflows/oss-history.yml
+++ b/.github/workflows/oss-history.yml
@@ -31,4 +31,4 @@ jobs:
         uses: nrfconnect/action-oss-history@main
         with:
           workspace: 'ncs'
-          args: -p zephyr -p mcuboot -p hostap -p wfa-qt-control-app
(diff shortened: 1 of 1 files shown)


rlubos (approved):

Vge0rge (approved):

rlubos (changes requested):
It doesn't seem that TFM was verified after all:
> 2024-03-22T12:37:17.2653955Z WARNING: trusted-firmware-m (modules/tee/tf-m/trusted-firmware-m): can't get loot: please fetch upstream URL https://github.com/zephyrproject-rtos/trusted-firmware-m (need revision 0b898c9b72171b0a260c0bb64a92ea4713f9e931)

Reverting my approval till we figure out what's wrong.

PR #14529, opened 2024-03-15
Replace nRF54H20 PDK with DK

anangl:

PR #14618, opened 2024-03-22
drivers: wifi: Disable workaround for anomaly 159

krish2718:
The proper workaround would be to use 64MHz CPU that will impact Wi-Fi performance, and as its never been seen so far, disable the workaround, and this can be revisited in the future.

Fixes SHEL-2636.

### DIFF
--- a/drivers/wifi/nrf700x/CMakeLists.txt
+++ b/drivers/wifi/nrf700x/CMakeLists.txt
@@ -129,11 +129,15 @@ if(CONFIG_NRF_WIFI_PATCHES_EXT_FLASH_XIP)
 zephyr_code_relocate(FILES src/fw_load.c LOCATION EXTFLASH_RODATA NOCOPY)
 endif()
 
+zephyr_compile_definitions_ifdef(CONFIG_NRF700X_ON_QSPI
 # These are XIP related anomalies and aren't applicable for nRF7002 and cause
 # throughput issues.
-zephyr_compile_definitions_ifdef(CONFIG_NRF700X_ON_QSPI
   -DNRF53_ERRATA_43_ENABLE_WORKAROUND=0
   -DNRF52_ERRATA_215_ENABLE_WORKAROUND=0
+# nRF70 QSPI doesn't use 192MHz clock and most samples use 128MHz, this can cause anomaly 159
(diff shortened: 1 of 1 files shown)


anangl (approved):

drivers/wifi/nrf700x/CMakeLists.txt
```diff
@@ -136,6 +136,13 @@ zephyr_compile_definitions_ifdef(CONFIG_NRF700X_ON_QSPI
   -DNRF52_ERRATA_215_ENABLE_WORKAROUND=0
 )
 
+# QSPI doesn't use 192MHz clock and most samples use 128MHz, this can cause Anamoly 159
+# but as its rare and not seen in most cases, we can disable it.
+# Alternative is 128MHz CPU should be disabled that impacts Wi-Fi performance.
+zephyr_compile_definitions(
+  -DNRF53_ERRATA_159_ENABLE_WORKAROUND=0
```
anangl, line 143: I think this should be disabled only if QSPI is used for communication with nRF700x, so as above, `zephyr_compile_definitions_ifdef(CONFIG_NRF700X_ON_QSPI` should be used here.

anangl (changes requested):

anangl (approved):

krish2718:
@anangl I saw that you had already fixed the issue https://github.com/nrfconnect/sdk-nrf/commit/23639c79b86ecbab2be64a75ce33c7ad164fd2f0 , as you now use 192MHz, do we still need this fix?

anangl:
> @anangl I saw that you had already fixed the issue [23639c7](https://github.com/nrfconnect/sdk-nrf/commit/23639c79b86ecbab2be64a75ce33c7ad164fd2f0) , as you now use 192MHz, do we still need this fix?

Yes, we need it. 23639c79b86ecbab2be64a75ce33c7ad164fd2f0 fixes the PCLK192M (QSPI clock) divider. But if the HCLK128M (CPU clock) divider is not set properly, only a related error log message is issued and the requested QSPI operation fails. Unless the workaround is disabled.

sachinthegreen (approved):

krish2718:
PR https://github.com/nrfconnect/sdk-nrf/pull/14529 cherry-picked this commit.

PR #14619, opened 2024-03-22
Bluetooth: Mesh: Add test to verify metadata extraction output

ludvigsj:
This adds build-only tests with targets with verious layouts of composition data, and runs a verification script as part of the build.

Currently tested:
  - Single composition data variables
  - Multiple composition data variables
  - Array containing multiple composition data

### DIFF
--- a/tests/subsys/bluetooth/mesh/metadata_extraction/prj.conf
+++ b/tests/subsys/bluetooth/mesh/metadata_extraction/prj.conf
@@ -1,0 +1,29 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Bootloader configuration
+CONFIG_BOOTLOADER_MCUBOOT=y
+CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION="1.2.3+4"
+
+# Bluetooth configuration
+CONFIG_BT=y
+CONFIG_BT_OBSERVER=y
+CONFIG_BT_PERIPHERAL=y
+
+# Bluetooth Mesh configuration
+CONFIG_BT_MESH=y
+CONFIG_BT_MESH_DFU_METADATA_ON_BUILD=y
+
+# Set to something other than default to check that it is parsed correctly
+CONFIG_BT_MESH_CRPL=64
+
+# Enable extra mesh features to check that the features field is parsed correctly
+CONFIG_BT_MESH_RELAY=y
+CONFIG_BT_MESH_FRIEND=y
+
+# Bluetooth Mesh models
+CONFIG_BT_MESH_LVL_CLI=y
+CONFIG_BT_MESH_ONOFF_CLI=y
--- a/tests/subsys/bluetooth/mesh/metadata_extraction/src/main.c
+++ b/tests/subsys/bluetooth/mesh/metadata_extraction/src/main.c
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/** @file
+ *  @brief Bluetooth Mesh DFU Target role sample
+ */
+#include <zephyr/kernel.h>
+#include <zephyr/bluetooth/bluetooth.h>
+#include <zephyr/bluetooth/mesh.h>
+#include <bluetooth/mesh/models.h>
+
+
+static struct bt_mesh_health_srv health_srv;
+
+BT_MESH_HEALTH_PUB_DEFINE(health_pub, 0);
+
+static struct bt_mesh_model models[] = {
+	BT_MESH_MODEL_CFG_SRV,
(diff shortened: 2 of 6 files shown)


ludvigsj:
This will fail for multiple comp data until #14577 is merged, but is otherwise ready for review

ludvigsj:
Just realized that this entire thing also depends on #14587 to compile.

tests/subsys/bluetooth/mesh/metadata_extraction/verify_metadata.py
```diff
@@ -0,0 +1,119 @@
...
+            'vid': 6,
+            'crpl': 64,
+            'features': 5,
+            'elements': [
+                {'location': 1, 'sig_models': [0, 2], 'vendor_models': []},
+                {'location': 2, 'sig_models': [4097, 4099], 'vendor_models': []}
+            ]
+        },
+        'composition_hash': '0x13f1a143',
+        'encoded_metadata': WildcardString('0102030004000000??????0143a1f1130200')
```
PavelVPV, line 57: I think it is better to extract the firmware size, put it here and then compare whole metadata.

tests/subsys/bluetooth/mesh/metadata_extraction/prj.conf
```diff
@@ -0,0 +1,31 @@
...
+
+# Bootloader configuration
+CONFIG_BOOTLOADER_MCUBOOT=y
+CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION="1.2.3+4"
+
+# Bluetooth configuration
+CONFIG_BT=y
+CONFIG_BT_OBSERVER=y
+CONFIG_BT_PERIPHERAL=y
+CONFIG_BT_ECC=y
```
alxelax, line 15: For which purposes do you need hardcoded ECC and Tinycrypt to extract metadata?

PavelVPV (commented):
One major comment regarding the firmware size.

tests/subsys/bluetooth/mesh/metadata_extraction/verify_metadata.py
ludvigsj: I thought about this, but the size extraction code would be literally identical to the code we're testing so I decided not to. Is there really any point to check that the same python code returns the same value?

ludvigsj: The code for encoding it would also be identical to the one we're testing

PavelVPV: But if the script change, we can catch it in the test

PavelVPV (approved):

tests/subsys/bluetooth/mesh/metadata_extraction/src/main.c
```diff
@@ -0,0 +1,117 @@
...
+#endif /* CONFIG_COMP_DATA_LAYOUT_MULTIPLE */
+#endif /* defined(COMP_DATA_LAYOUT_SINGLE) || defined(COMP_DATA_LAYOUT_MULTIPLE) */
+
+static const uint8_t dev_uuid[16];
+
+static const struct bt_mesh_prov prov = {
+	.uuid = dev_uuid,
+};
+
+volatile bool use_alt;
```
alxelax, line 86: Why is this variable volatile and not static?

alxelax (commented):
LGTM but there is still question.
Seems CI cannot pass.

tests/subsys/bluetooth/mesh/metadata_extraction/src/main.c
ludvigsj: If it's static it will be assumed (correctly) by the compiler to always be 0/false and the alternative composition data will be optimized out of the build, and not picked up by the script. It's a trick to make the compiler think it can't know which composition data to use until runtime. (LTO might break this trick, I haven't checked, but let's leave that until we potentially add LTO test to this test)

ludvigsj:
@alxelax Like mentioned above, this test will fail until #14587 and #14577 are merged.

tests/subsys/bluetooth/mesh/metadata_extraction/src/main.c
alxelax: Could you add it here as a comment?

tests/subsys/bluetooth/mesh/metadata_extraction/src/main.c
```diff
@@ -0,0 +1,120 @@
...
+#endif /* defined(COMP_DATA_LAYOUT_SINGLE) || defined(COMP_DATA_LAYOUT_MULTIPLE) */
+
+static const uint8_t dev_uuid[16];
+
+static const struct bt_mesh_prov prov = {
+	.uuid = dev_uuid,
+};
+
+#ifndef CONFIG_COMP_DATA_LAYOUT_SINGLE
+/* Declare as `volatile` to avoid it being optimized out. */
```
alxelax, line 87:
```suggestion
/* Since the variable does not change value, compiler might optimize it out
 * as well as conditions where variable is used. To prevent it, `volatile` qualifier is used. */
```
Could I suggest it? It will be more clear (at least to me) if I take a look at the code after half a year for example

alxelax (approved):

PR #14625, opened 2024-03-25
lib: location: Make wifi_scan_ready semaphore to static

trantanen:
Semaphore wifi_scan_ready has been reserved from the stack. This doesn't sound good when it's used from different threads. Moving it to static RAM.

Jira: NCSDK-26707

### DIFF
--- a/lib/location/method_cloud_location.c
+++ b/lib/location/method_cloud_location.c
@@ -36,6 +36,10 @@ struct method_cloud_location_start_work_args {
 static struct method_cloud_location_start_work_args method_cloud_location_start_work;
 static bool running;
 
+#if defined(CONFIG_LOCATION_METHOD_WIFI)
+static K_SEM_DEFINE(wifi_scan_ready, 0, 1);
+#endif
+
 static void method_cloud_location_positioning_work_fn(struct k_work *work)
 {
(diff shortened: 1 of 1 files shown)


lib/location/method_cloud_location.c
```diff
@@ -46,9 +50,7 @@ static void method_cloud_location_positioning_work_fn(struct k_work *work)
 	struct lte_lc_cells_info *scan_cellular_info = NULL;
 	int err = 0;
 #if defined(CONFIG_LOCATION_METHOD_WIFI)
-	struct k_sem wifi_scan_ready;
-
-	k_sem_init(&wifi_scan_ready, 0, 1);
+	k_sem_reset(&wifi_scan_ready);
```
tokangas, line 53: An empty line after the variable declarations would be nice.

tokangas (approved):

lib/location/method_cloud_location.c
trantanen: Added. Seems compliance was happy when there was #if statement in the next line after the declarations.

glarsennordic (approved):
LGTM, thanks!

jhirsi (approved):

PR #14630, opened 2024-03-26
net: lib: nrf_cloud_coap: add proxy download support

jayteemo:
Add proxy download support for nRF Cloud CoAP.
Enable P-GPS data to be downloaded using the
nRF Cloud CoAP connection.
IRIS-6636


plskeggs (approved):

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap_transport.c
```diff
@@ -49,6 +49,13 @@ LOG_MODULE_REGISTER(nrf_cloud_coap_transport, CONFIG_NRF_CLOUD_COAP_LOG_LEVEL);
 #define VER_STRING_FMT "mver=%s&cver=%s&dver=%s"
 #define VER_STRING_FMT2 "cver=" CDDL_VERSION "&dver=" BUILD_VERSION_STR
 #define BUILD_VERSION_STR STRINGIFY(BUILD_VERSION)
+#define COAP_PROXY_RSC "proxy"
+
+/* CoAP extended options: When multiple options are used,
+ * for simplicity, always put COAP_OPTION_ACCEPT first
+ */
+#define ACPT_IDX 0 /* COAP_OPTION_ACCEPT */
+#define PRXY_IDX 1 /* COAP_OPTION_PROXY_URI */
```
glarsennordic, line 58: I fear this will lead to confusion since the underlying options don't seem to care about the order they are provided in

glarsennordic (approved):
Approving since I'm not 100% sure anything I've pointed out is actually a problem

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap_transport.c
glarsennordic: As a compromise, we could assert that `opt->value[ACPT_IDX].code == COAP_OPTION_ACCEPT` in client_transfer_opt

Though I see now that it is a static function, so I guess we only stand to confuse our future selves

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap.c
```diff
@@ -144,7 +144,7 @@ int nrf_cloud_coap_agnss_data_get(struct nrf_cloud_rest_agnss_request const *con
 	} else if (agnss_err > 0) {
 		LOG_RESULT_CODE_ERR("Unexpected result code:", agnss_err);
 		err = agnss_err;
-	} else {
+	} else if (agnss_err < 0) {
```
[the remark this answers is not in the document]
jayteemo: > What about the case that `err != 0` and `agnss_err == 0`? Is that case possible?

sure, if that happens this function will correctly return `err`.
i'll update so the log is also printed in that case.

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap_transport.c
jayteemo: the only reason i added this was for the `LOG_DBG` on line 552.

plskeggs: CoAP options need to be specified in numerical order or else you'll get a message (when `CONFIG_LOG_NET=y` and CONFIG_COAP_LOG_LEVEL_DBG=y`) indicating the "Option is not added in ascending order".

It turns out this is a requirement for encoding the option number.  See:
https://datatracker.ietf.org/doc/html/rfc7252#section-3.1

The Zephyr CoAP stack does the right thing and inserts the out-of-order option in the correct position, but it is best to provide them in order.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -454,7 +454,17 @@ Libraries for networking
...
+
+    * Support for downloads using nRF Cloud CoAP's proxy resource.
+
+  * Updated:
+
+    * Request proprietary PSM mode for ``SOC_NRF9151_LACA`` and ``SOC_NRF9131_LACA`` in addition to ``SOC_NRF9161_LACA``.
+
+* :ref:`lib_nrf_cloud_pgps` library:
+
+  * Added Kconfig option :kconfig:option:`NRF_CLOUD_PGPS_DOWNLOAD_TRANSPORT_COAP` to enable using the nRF Cloud CoAP connection to download P-GPS data.
```
peknis, line 467:
```suggestion
  * Added the :kconfig:option:`NRF_CLOUD_PGPS_DOWNLOAD_TRANSPORT_COAP` Kconfig option to enable using the nRF Cloud CoAP connection to download P-GPS data.
```

peknis (approved):
Approved with a minor nit.

jayteemo:
closing, using https://github.com/nrfconnect/sdk-nrf/pull/15431 instead

PR #14631, opened 2024-03-26
Dwell time range check, Add channel validity check

kapbh:


drivers/wifi/nrf700x/src/wifi_mgmt_scan.c
```diff
@@ -137,10 +137,30 @@ int nrf_wifi_disp_scan_zep(const struct device *dev, struct wifi_scan_params *pa
 		}
 
 		scan_info->scan_params.bands = params->bands;
-		scan_info->scan_params.dwell_time_active = params->dwell_time_active;
-		scan_info->scan_params.dwell_time_passive = params->dwell_time_passive;
 
-		vif_ctx_zep->max_bss_cnt = params->max_bss_cnt;
+		if ((params->dwell_time_active < 5) || (params->dwell_time_active > 1000)) {
```
sachinthegreen, line 141: Generic comment: Please create defines for all the magic numbers used (such as `5`, `1000` etc.) maybe in wifi_mgmt.h and then use those defines.

sachinthegreen (changes requested):
Please give Jira references in `37a7bfee8ed2e61e56b743e1de5ae33e5a8deb62` commit rather than in the manifest commit ?

drivers/wifi/nrf700x/src/wifi_mgmt_scan.c
kapbh: Ok. we can create define all these magic numbers. Will it be approved in upstream wifi_mgmt.h?

sachinthegreen (approved):

krish2718 (approved):

PR #14645, opened 2024-03-26
scripts: cert_tool: Add argument to set common-name

simensrostad:
Add argument to set common-name


shanthanordic (approved):

gregersrygg (changes requested):
As discussed offline, we should probably have a separate command to create server certificates instead of re-using the root ca command. The attributes should be different for a server than a root ca.

jangalda-nsc:
@shanthanordic I have no experience with changed file. So, I do not think I'm the best reviewer.

PR #14646, opened 2024-03-26
net: lib: wifi_credentials: Add new functions and params

simensrostad:
Add function to clear all stored credentials and to check if there are any valid entries in storage.

Add storing of channel.

Make wifi exts auto connect feature pass in preferred channel to the wifi stack.

### DIFF
--- a/include/net/wifi_credentials.h
+++ b/include/net/wifi_credentials.h
@@ -51,6 +51,7 @@ struct wifi_credentials_header {
 	size_t ssid_len;
 	uint8_t bssid[WIFI_MAC_ADDR_LEN];
 	uint32_t flags;
+	uint8_t channel;
 };
 
 /**
@@ -94,6 +95,7 @@ struct wifi_credentials_enterprise {
  * @param[in] password_buf_len		length of password_buf
  * @param[out] password_len		length of password
  * @param[out] flags			flags
+ * @param[out] channel			channel
  *
  * @return 0		Success.
  * @return -ENOENT	No network with this SSID was found.
@@ -109,7 +111,8 @@ int wifi_credentials_get_by_ssid_personal(
 	char *password_buf,
 	size_t password_buf_len,
 	size_t *password_len,
-	uint32_t *flags
+	uint32_t *flags,
+	uint8_t *channel
 );
 
 /**
@@ -123,6 +126,7 @@ int wifi_credentials_get_by_ssid_personal(
  * @param[in] password		password
  * @param[in] password_len		length of password
  * @param[in] flags			flags
+ * @param[in] channel			Channel
  *
  * @return 0			Success. Credentials are stored in persistent storage.
  * @return -EINVAL		A required buffer was NULL or security type is not supported.
@@ -137,7 +141,8 @@ int wifi_credentials_set_personal(
 	size_t bssid_len,
 	const char *password,
 	size_t password_len,
-	uint32_t flags
+	uint32_t flags,
+	uint8_t channel
 );
 
 /**
@@ -177,6 +182,23 @@ int wifi_credentials_set_personal_struct(const struct wifi_credentials_personal 
  * @return			0 on success, otherwise a negative error code
  */
 int wifi_credentials_delete_by_ssid(const char *ssid, size_t ssid_len);
+
+/**
+ * @brief Check if credentials storage is empty.
+ *
+ * @return			true if credential storage is empty, otherwise false
+ */
+bool wifi_credentials_is_empty(void);
+
+/**
+ * @brief Deletes all stored Wi-Fi credentials.
+ *
+ * This function deletes all Wi-Fi credentials that have been stored in the system.
+ * It is typically used when you want to clear all saved networks.
+ *
+ * @return			0 on successful, otherwise a negative error code
+ */
+int wifi_credentials_delete_all(void);
 
 /**
  * @brief Callback type for wifi_credentials_for_each_ssid.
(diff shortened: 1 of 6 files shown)


ioannisg (commented):
Looks good to me.
Just a question @simensrostad : shouldn't this API extension compiled-in only if non-STATIC (or not BACKEND_NONE) configuration is selected? Ignore comment if this is already the case. (if not, I believe it is cleaner and possibly more memory-optimized if STATIC credential option does not include this api extension) Thoughts?

simensrostad:
> Looks good to me. Just a question @simensrostad : shouldn't this API extension compiled-in only if non-STATIC (or not BACKEND_NONE) configuration is selected? Ignore comment if this is already the case. (if not, I believe it is cleaner and possibly more memory-optimized if STATIC credential option does not include this api extension) Thoughts?

Note sure of the general structure of the different backends and how each parameter is affected. But I assume that if this is the case, there are also other parameters that fall under this umbrella. Its only ssid and PSK that can be set with the static backend. (Kconfig) Its a separate effort to hide all these params.

cc @maxd-nordic can maybe elaborate here?

simensrostad:
Extending the wifi cred shell module is missing

maxd-nordic:
@simensrostad - @ioannisg is right, `wifi_credentials_is_empty` and `wifi_credentials_delete_all` only make sense if `CONFIG_WIFI_CREDENTIALS_BACKEND_NONE` is not set. (It doesn't make a huge difference though.)

simensrostad:
> @simensrostad - @ioannisg is right, `wifi_credentials_is_empty` and `wifi_credentials_delete_all` only make sense if `CONFIG_WIFI_CREDENTIALS_BACKEND_NONE` is not set. (It doesn't make a huge difference though.)

Ah, right. I got mixed up with the params. Didn't consider the API 😢 Ill revisit this after vacation, I also need to update the tests. But I'm not a big fan of ifdeffing header files. An IS_ENABLED guard in the beginning of the function entries is perhaps cleaner and we avoid this. WDYT?

ioannisg:
> > @simensrostad - @ioannisg is right, `wifi_credentials_is_empty` and `wifi_credentials_delete_all` only make sense if `CONFIG_WIFI_CREDENTIALS_BACKEND_NONE` is not set. (It doesn't make a huge difference though.)
> 
> Ah, right. I got mixed up with the params. Didn't consider the API 😢 Ill revisit this after vacation, I also need to update the tests. But I'm not a big fan of ifdeffing header files. An IS_ENABLED guard in the beginning of the function entries is perhaps cleaner and we avoid this. WDYT?

Thanks @simensrostad . Yeah, whatever solution results in compiling out unwanted code; up to you to decide... By the way, the same applies to other (already defined & implemented) APIs that make sense only if non-STATIC (or NONE) credentials' solution is selected. So you might want to go through the rest of the library and compile out a bunch of things :)

subsys/net/lib/wifi_credentials/wifi_credentials.c
```diff
@@ -191,7 +191,8 @@ int wifi_credentials_set_personal_struct(const struct wifi_credentials_personal
 
 int wifi_credentials_set_personal(const char *ssid, size_t ssid_len, enum wifi_security_type type,
 				  const uint8_t *bssid, size_t bssid_len, const char *password,
-				  size_t password_len, uint32_t flags)
+				  size_t password_len, uint32_t flags,
```
balaji-nordic, line 194: I believe the new line is not needed.

subsys/net/lib/wifi_credentials/wifi_credentials.c
```diff
@@ -353,4 +355,38 @@ void wifi_credentials_for_each_ssid(wifi_credentials_ssid_cb cb, void *cb_arg)
...
+{
+	int ret = 0;
+
+	k_mutex_lock(&wifi_credentials_mutex, K_FOREVER);
+	for (size_t i = 0; i < CONFIG_WIFI_CREDENTIALS_MAX_ENTRIES; ++i) {
+		if (is_entry_used(i)) {
+			ret = wifi_credentials_delete_entry(i);
+			if (ret) {
+				LOG_ERR("Failed to delete WiFi credentials index %d, err: %d", i, ret);
+				goto exit;
```
balaji-nordic, line 381: You can replace the `goto` statement a `break` .

balaji-nordic (changes requested):

simensrostad:
> > > @simensrostad - @ioannisg is right, `wifi_credentials_is_empty` and `wifi_credentials_delete_all` only make sense if `CONFIG_WIFI_CREDENTIALS_BACKEND_NONE` is not set. (It doesn't make a huge difference though.)
> > 
> > 
> > Ah, right. I got mixed up with the params. Didn't consider the API 😢 Ill revisit this after vacation, I also need to update the tests. But I'm not a big fan of ifdeffing header files. An IS_ENABLED guard in the beginning of the function entries is perhaps cleaner and we avoid this. WDYT?
> 
> Thanks @simensrostad . Yeah, whatever solution results in compiling out unwanted code; up to you to decide... By the way, the same applies to other (already defined & implemented) APIs that make sense only if non-STATIC (or NONE) credentials' solution is selected. So you might want to go through the rest of the library and compile out a bunch of things :)

In that case its reasonable to add these guards in a separate PR (Since its a broader effort). It would also do good to investigate if its needed at all, I.E the compiler already optimizes away the unneeded code.

JIRA:
https://nordicsemi.atlassian.net/browse/CIA-1190

divipillai (approved):

include/net/wifi_credentials.h
```diff
@@ -94,6 +95,7 @@ struct wifi_credentials_enterprise {
  * @param[in] password_buf_len		length of password_buf
  * @param[out] password_len		length of password
  * @param[out] flags			flags
+ * @param[out] flags			channel
```
divipillai, line 98:
```suggestion
 * @param[out] channel			channel
```
I think this should be channel? fixes the build error.

simensrostad: thanks!

balaji-nordic (approved):
The commit message has a minor typo. `news` instead of `new`. Rest looks good

rlubos (approved):

PR #14654, opened 2024-03-27
Add support for BSSID configuration in connect call.

rado17:
test_wifi: sdk-nrf-pr-14626


sachinthegreen (approved):

modules/hostap/src/supp_api.c
```diff
@@ -398,6 +398,11 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
 		}
 	}
 
+	if (params->bssid[0] != 0) {
```
krish2718, line 401: `....00` this is still a valid BSSID, no?

rado17: It is. But here, we are checking this against NULL character (ASCII value = 0).

krish2718: Yes, so, a BSSID like `00:11:22:33:44:00` would be considered NULL and hence not valid. We should check for all zeroes  to handle the case where user hasn't set or not. And if non-zero, then do basic validation and only then accept.

VivekUppunda (approved):

kapbh (approved):

modules/hostap/src/supp_api.c
```diff
@@ -403,6 +404,19 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
 		}
 	}
 
+	memcpy((void *)&mac, params->bssid, WIFI_MAC_ADDR_LEN);
+	if (!net_eth_is_addr_broadcast(&mac) &&
```
krish2718, line 408: we should fail for invalid BSSID except for all zeroes (which can be treated as unset).

modules/hostap/src/supp_api.c
```diff
@@ -403,6 +404,25 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
 		}
 	}
 
+	memcpy((void *)&mac, params->bssid, WIFI_MAC_ADDR_LEN);
+	if (net_eth_is_addr_broadcast(&mac) ||
```
krish2718, line 408: You move this check in to the below branch (!unspecified)

krish2718 (approved):

modules/hostap/src/supp_api.c
rado17: !unspecified checks for all 0s which is equivalent to `not setting the BSSID`. That is why I am returning error only for broadcast and multicast case.

modules/hostap/src/supp_api.c
rado17: Done

jukkar (approved):

rlubos:
@rado17 Needs a rebase.

PR #14655, opened 2024-03-27
bluetooth: controller: update SDC and MPSL

olivier-le-sage:
Brings in the latest changes to SDC and MPSL


rugeGerritsen (changes requested):
Please also update the dragoon revision

PR #14659, opened 2024-03-27
samples: net: Use default TCP stack size with Wi-Fi

rlubos:
The Wi-Fi driver by default runs on a separate TX thread, therefore it shouldn't really be needed to increase the TCP stack size, just rely on the defaults.


rlubos:
For reviewers - I've tested this with `https_client` sample, and even with logging enabled and forced retransmissions we've barely exceeded half kbyte of the TCP stack usage (default size is 1k). But please let me know if you observe some abnormalities with this change.

ioannisg (approved):
Thanks!

glarsennordic (approved):
MSS Wi-Fi connectivity builds work!

plskeggs (approved):

simensrostad (approved):
Thanks! @jorgenmk Can we do a nightly run on this one?

bjda (approved):

PR #14661, opened 2024-03-27
SPAKE2P fixes

Vge0rge:
This fixes the SPAKE2P sample which used the wrong key for the server role.

It also adds support for the server role to use a private key in the CRACEN PSA driver.

And does some configuration fixes for CRACEN.


vili-nordic (approved):

ArekBalysNordic (approved):
It allows running Matter stack with Cracen or Oberon backend.

PR #14664, opened 2024-03-28
doc: wifi: Add Wi-Fi certification page

richabp:
Added Wi-Fi certification page under the Wi-Fi protocols. Moved the Wi-Fi certification section from the Overview page to the new page.
Added certification info to the nRF7002 DK gsg.

SHEL-1666


ioannisg (approved):
Looks fine to me, thanks!

PR #14665, opened 2024-03-28
net: lib: download_client: Fix address parsing

SeppoTakalo:
Fixed few address parsing issues from download client
* If unsupported protocol was used, for example telnet://example.com parser just assumed that parsing failure as no protocol and defaulted to HTTP. Now returns EPROTONOSUPPORT as documented earlier.
* If no protocol was provided, but beginning of of the host address matched some protocol, that protocol was assumed. For example: download_client_get(&c, "coap.me/index.html", &cfg , 0, 0) uses coap://, because "coap" matches, but it should have used "http://". Now it needs to have "://" in the protocol to match.

Also, converted ztests to native_sim instead of native_posix. Added few more unittests for the cases above.


rlubos (approved):

SeppoTakalo:
Changed the error code in unittests to something that is known by the Arm compiler.
Also added `qemu_cortex_m3` to allowed target, as it uses same compiler than the target HW (nRF91). Removed the `nrf9160_nrf9160_ns` from the list. Makes no sense to run unittests on HW.

stig-bjorlykke (approved):

eivindj-nordic (approved):

PR #14676, opened 2024-03-30
bluetooth: controller: deselect BT_CTLR_ECDH

olivier-le-sage:
It's rarely, if ever, used by the app. It also uses considerable memory. Places that need it can still select it.

### DIFF
--- a/samples/bluetooth/central_hids/child_image/hci_ipc.conf
+++ b/samples/bluetooth/central_hids/child_image/hci_ipc.conf
@@ -1,9 +1,7 @@
 #
-# Copyright (c) 2021 Nordic Semiconductor
+# Copyright (c) 2024 Nordic Semiconductor
 #
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
-CONFIG_BT_MAX_CONN=2
-CONFIG_BT_CENTRAL=n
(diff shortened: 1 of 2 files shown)


rugeGerritsen (changes requested):
This is a change that will change behavior for many applications. It will stop using the oberon library for ECDH by default. 

Instead of applying this change we should rather make the host use that library. That will also improve the performance for nRF53 where the cryptocell is located in the application core.

jori-nordic:
I think @seko-nordic hinted about moving all crypto to libs in the host/app core. Do you still have plans for this Sergey?

seko-nordic:
> I think @seko-nordic hinted about moving all crypto to libs in the host/app core. Do you still have plans for this Sergey?

Ideally, host should use HW accelerated crypto. SW crypto back-end should be used only with variants where HW crypto is not available (ECDH on nRF52820 for example).
For ECDH, I think in the end we should remove from SDC fake BT_HCI_OP_LE_P256_PUBLIC_KEY, BT_HCI_OP_LE_GENERATE_DHKEY, BT_HCI_OP_LE_GENERATE_DHKEY_V2 commands and its dependencies. I'm not sure if we need to move any libs for that, I though there should be support in app core already?
Anyone wants to give it a try?

PR #14678, opened 2024-04-01
wifi: Fix channel 14 handling in JP reg domain

krish2718:


jukkar (approved):

sachinthegreen (approved):

rado17 (approved):

krish2718:
@jukkar FYI, the last two commits increase `CONFIG_NET_MGMT_EVENT_STACK_SIZE` the upstream is still 4096 if you cherry-picked the OOM fixes (heap->stack), then it needs a similar fix in upstream as well.

maxd-nordic (approved):

samples/cellular/nrf_cloud_multi_service/overlay_nrf7002ek_wifi_no_lte.conf
```diff
@@ -155,7 +155,6 @@ CONFIG_HEAP_MEM_POOL_SIZE=138000
 ## Miscellaneous resource allocation tweaks needed to support Wi-Fi.
 CONFIG_MAIN_STACK_SIZE=2048
 CONFIG_SYSTEM_WORKQUEUE_STACK_SIZE=4096
-CONFIG_NET_MGMT_EVENT_STACK_SIZE=4096
```
glarsennordic, line 158: This is fine for now (though we'll need to test MSS with this setting before approving), but how should we handle this if at some point an exterior subsystem requires increases to this stack?

krish2718: > but how should we handle this if at some point an exterior subsystem requires increases to this stack?

Tough question, I don't think we have any process for this, best way is to modify the configuration at the trigger i.e, whoever needs a modified value.

bjda (approved):

PR #14699, opened 2024-04-03
NCSDK-26629: Migrate SUIT platform

tomchy:


CODEOWNERS
```diff
@@ -343,6 +344,7 @@ Kconfig*                                  @tejlmand
 /tests/subsys/nrf_profiler/               @pdunaj @MarekPieta
 /tests/subsys/sdfw_services/              @anhmolt @hakonfam @jonathannilsen
 /tests/subsys/zigbee/                     @milewr
+/tests/subsys/suit/                       @tomchy
```
ahasztag, line 347: Wouldn't we like to extend the code owners list with other developers in the SUIT team (probably @adsz-nordic, @ahasztag perhaps @e-rk if he'd like). I think it wouldn't block anyone, and the team would be informed about external changes if you'd be on vacation

subsys/mgmt/suitfu/src/suitfu_mgmt_img.c
```diff
@@ -200,12 +200,13 @@ static int suitfu_mgmt_img_state_read(struct smp_streamer *ctx)
...
 	if (ok) {
 		installed_vers_str[sizeof(installed_vers_str) - 1] = '\0';
-		ok = zcbor_tstr_put_term(zse, installed_vers_str);
+		ok = zcbor_tstr_put_term(zse, installed_vers_str, sizeof(installed_vers_str));
 		LOG_DBG("Manifest version encoded: %d", ok);
 	}
 
 	if (ok) {
-		ok = zcbor_tstr_put_term(zse, "hash") && zcbor_bstr_encode(zse, &zhash);
+		ok = zcbor_tstr_put_term(zse, "hash", sizeof("hash")) &&
```
oyvindronningstad, line 208: All these literals can be encoded using `*_put_lit()` instead, to avoid the runtime call to strnlen.

adsz-nordic (approved):

robertstypa (approved):

oyvindronningstad (approved):

tests/subsys/suit/flash_sink/boards/nrf54h20dk_nrf54h20_cpuapp.overlay
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+&mram1x {
+	erase-block-size = < 0x400 >;
```
nordicjm, line 8: Not going to request changes but might be an idea to have one of these files and then just `#include "../..."` from the rest

subsys/suit/memory_layout/Kconfig
```diff
@@ -6,7 +6,7 @@
 
 config SUIT_MEMORY_LAYOUT_EXTMEM_ADDRESS_RANGE_START
 	hex
-	default 0x60000000 if SOC_NRF54H20
+	default 0x60000000 if SOC_SERIES_NRF54HX
 	default 0x12000000 if SOC_NRF52840
 	default 0x0 if BOARD_NATIVE_POSIX
```
nordicjm, line 11: This should come from dts

subsys/suit/memory_layout/Kconfig
```diff
@@ -15,7 +15,7 @@ config SUIT_MEMORY_LAYOUT_EXTMEM_ADDRESS_RANGE_START
 
 config SUIT_MEMORY_LAYOUT_EXTMEM_ADDRESS_RANGE_SIZE
 	hex
-	default 0x40000000 if SOC_NRF54H20
+	default 0x40000000 if SOC_SERIES_NRF54HX
 	default 0x8000000 if SOC_NRF52840
 	default 0x0 if BOARD_NATIVE_POSIX
 	help
```
nordicjm, line 21: and this, this is a hardware property which should come from dts

subsys/suit/stream/Kconfig
```diff
@@ -44,14 +44,14 @@ if SUIT_STREAM_SINK_EXTMEM
...
 	  Start address of the extended memory range.
 	  This value is SOC specific and is not meant to be changed.
 
 config SUIT_STREAM_SINK_EXTMEM_ADDRESS_RANGE_SIZE
 	hex
-	default 0x40000000 if SOC_NRF54H20
+	default 0x40000000 if SOC_SERIES_NRF54HX
 	help
 	  Size of the extended memory range.
 	  This value is SOC specific and is not meant to be changed.
```
nordicjm, line 57: I don't get why this is here, you already have a Kconfig (which needs to move to dts) but why is the option duplicated already?

tomchy: Nice catch!
This is a leftover and can be safely removed.

subsys/suit/memory_layout/Kconfig
tomchy: Unfortunately the EXMIF is not yet defined in DTS. To be aligned in the future (possibly along with the falsh_companion sample migration process).

subsys/suit/memory_layout/Kconfig
tomchy: Same here - I'll add a note in the task describing flash_companion image migration.

nordicjm (approved):

PR #14701, opened 2024-04-03
net: lib_ fota_download: Return errors from URI parsing

SeppoTakalo:
Parse URI before starting the download client
and return parsing failures from _start().

Also, check that download_client parses the protocol,
which means that it knows the protocol.

NCSDK-26790

### DIFF
--- a/subsys/net/lib/fota_download/src/util/fota_download_util.c
+++ b/subsys/net/lib/fota_download/src/util/fota_download_util.c
@@ -32,6 +32,8 @@
 #include "fota_download_smp.h"
 #endif
 
+#include "download_client_internal.h"
+
 LOG_MODULE_REGISTER(fota_download_util, CONFIG_FOTA_DOWNLOAD_LOG_LEVEL);
 
 /**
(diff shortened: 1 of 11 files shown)


juhaylinen (approved):

subsys/net/lib/download_client/src/download_client_internal.h
```diff
@@ -0,0 +1,27 @@
+/*
```
stig-bjorlykke, line 1: I think this file belongs in `/include/` instead of `/src/`?

subsys/net/lib/fota_download/src/util/fota_download_util.c
```diff
@@ -32,6 +32,8 @@
 #include "fota_download_smp.h"
 #endif
 
+#include "../../download_client/src/download_client_internal.h"
```
stig-bjorlykke, line 35: Can CMake help finding this include instead of moving up the tree?

SeppoTakalo: I changed this to:
```
zephyr_include_directories_ifdef(CONFIG_DOWNLOAD_CLIENT
  ${ZEPHYR_NRF_MODULE_DIR}/subsys/net/lib/download_client/include)
```

Is that better?

stig-bjorlykke: Yes, I think this looks better.

stig-bjorlykke (approved):

PR #14706, opened 2024-04-03
manifest: Update Zephyr's net stack.

MarGasiorek:
Two commits from zephyr upstream.


rlubos (changes requested):
SHA can be updated. But please drop the `nrf fromtree` tag from the commit title, we only use tags in forks.

rlubos (approved):

PR #14720, opened 2024-04-04
tests: unity: Make unity configuration template configurable

nordic-krch:
Allow to use configurable unity configuration template file.


balaji-nordic:
This is a good improvement. :) Lets get it in

balaji-nordic (approved):

PR #14724, opened 2024-04-04
samples: matter: Introduced initial support for nRF54L15

kkasperczyk-no:
Added support for nRF54L15 platform in the following samples:
* Light Bulb
* Light Switch
* Template
* Thermostat

Additionally, modified light bulb sample to use PWM conditionally to address the fact that nRF54L does not support PWM yet.

### DIFF
--- a/samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
+++ b/samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
@@ -1,0 +1,30 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# TODO: Workaround to be removed once PWM will be supported on nRF54L.
+CONFIG_PWM=n
+
+# Multirole is the only currently supported role by SoftDevice.
+CONFIG_BT_LL_SOFTDEVICE_MULTIROLE=y
+
+# TODO: Workaround to be removed once DFU will be supported on nRF54L.
+CONFIG_CHIP_OTA_REQUESTOR=n
+CONFIG_CHIP_QSPI_NOR=n
+
+CONFIG_FPROTECT=n
+CONFIG_LOG_MODE_DEFERRED=y
+CONFIG_FPU=n
+CONFIG_PM=n
+CONFIG_HWINFO_NRF=n
+
+CONFIG_NRF_802154_TEMPERATURE_UPDATE=n
+
+# TODO: Workaround that disables factory data until it will be supported.
+CONFIG_CHIP_FACTORY_DATA=n
+CONFIG_CHIP_FACTORY_DATA_BUILD=n
+
+# TODO: Workaround for buttons on nRF54L15 PDK in revision 0.2.x. To be removed for a PDK in higher revisions.
+CONFIG_DK_LIBRARY_BUTTON_NO_ISR=y
--- a/samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
+++ b/samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/ {
+	chosen {
+		zephyr,console = &uart30;
+		zephyr,shell-uart = &uart30;
+	};
+
+	aliases {
+		factory-data = &factory_data;
+		factory-data-memory-region = &rram0;
+	};
+};
+
+/delete-node/ &rram0;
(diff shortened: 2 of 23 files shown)


greg-fer (approved):
I'm approving in the absence of @wiba-nordic .

FYI, @annwoj and @FrancescoSer .

greg-fer:
@kkasperczyk-no , please make sure to add a changelog entry for this change.

doublemis1 (approved):

samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
```diff
@@ -0,0 +1,68 @@
...
+		zephyr,shell-uart = &uart30;
+	};
+
+	aliases {
+		factory-data = &factory_data;
+		factory-data-memory-region = &rram0;
+	};
+};
+
+/* Matter needs at least 700 kB Flash size, instead of default 344 kB. */
```
ArekBalysNordic, line 19: NIT: This is a bit obsolete comment, and it can be removed or changed somehow because we use the full 1.5 MB now.

samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
```diff
@@ -0,0 +1,68 @@
...
+			};
+		};
+	};
+};
+
+&uart30 {
+	status = "okay";
+};
+
+/* Workaround for nRF54L15 PDK in revision 0.2.x
```
ArekBalysNordic, line 58: Could you please check if removing those lines and adding `CONFIG_DK_LIBRARY_BUTTON_NO_ISR=y` to the board's prj.conf file is a better workaround? It will be easier to remove it from the prj.conf file or just set this config as CMAKE args instead of manipulating the DTS file.

samples/matter/light_bulb/src/app_task.cpp
```diff
@@ -143,6 +160,9 @@ void AppTask::LightingActionEventHandler(const LightingEvent &event)
 	if (action == Nrf::PWMDevice::INVALID_ACTION || !Instance().mPWMDevice.InitiateAction(action, actor, NULL)) {
 		LOG_INF("An action could not be initiated.");
 	}
+#else
+	Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED2).Set(!Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED2).GetState());
```
ArekBalysNordic, line 164: Are you sure that here should be a toggle? I guess it depends on `static_cast<Nrf::PWMDevice::Action_t>(event.Action)`, so depending on `event.Action` without casting to PWMDevice in this case.

ArekBalysNordic (commented):
Two comments refer to all DTS files.

samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
```diff
@@ -0,0 +1,35 @@
...
+CONFIG_CHIP_FACTORY_DATA=n
+CONFIG_CHIP_FACTORY_DATA_BUILD=n
+
+# TODO: Workaround that disables CRACEN and switch to OBERON for now. To be removed once SPAKE2+ will be supported.
+CONFIG_PSA_CRYPTO_DRIVER_OBERON=y
+
+CONFIG_PSA_USE_CRACEN_KEY_MANAGEMENT_DRIVER=n # Disable ECC_SECP_R1_256 [SPAKE | PASE]
+CONFIG_PSA_USE_CRACEN_ASYMMETRIC_DRIVER=n # Disable ECDSA [Certificates]
+CONFIG_PSA_USE_CRACEN_MAC_DRIVER=n # Disable CMAC/HMAC [Secure session]
+CONFIG_PSA_USE_CRACEN_KEY_AGREEMENT_DRIVER=n # disable ECDH_SECP_R1_256 [CASE]
```
ArekBalysNordic, line 35: This PR: https://github.com/nrfconnect/sdk-nrf/pull/14661 allows using this workaround (without the fix these configs are useless here), but it also fixes Cracen's SPAKE2+ implementation, so those configs and the workaround won't be needed anymore. They can be removed.

ArekBalysNordic: And we should set `CONFIG_PSA_CRYPTO_DRIVER_OBERON=n` here

samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
kkasperczyk-no: Done

samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
kkasperczyk-no: Done

ArekBalysNordic (approved):

PR #14726, opened 2024-04-04
subsys: pcd: Fix build error 'CONFIG_FW_INFO_MAGIC_LEN' undeclared here

joerchan:
Fix build error 'CONFIG_FW_INFO_MAGIC_LEN' undeclared here. Including the fw_info_bare.h header file without CONFIG_FW_INFO enabled causes issues in struct definition array size, and in static inline functions.


sigvartmh (approved):
Thanks 👍

Vge0rge (approved):

PR #14731, opened 2024-04-05
applications: nrf5340_audio: Support any context

alexsven:
- Add CONTEXT_TYPE_ANY to both supported and available
- OCT-2621


koffes (approved):
Approved, but it would be appreciated if @Thalley could have a look as well.

Thalley (approved):
Looks fine to me

I would, however, suggest to either define a Kconfig to set the (at least supported) context at buildtime, and use that everywhere. 

Alternatively as a step towards that, I'd define a single value something (either a const value or a `#define` for your context, so that when you want to modify the context at some other point, you can just modify a single value, rather than these 16 places.

alexsven:
> Looks fine to me
> 
> I would, however, suggest to either define a Kconfig to set the (at least supported) context at buildtime, and use that everywhere.
> 
> Alternatively as a step towards that, I'd define a single value something (either a const value or a `#define` for your context, so that when you want to modify the context at some other point, you can just modify a single value, rather than these 16 places.

Good point, will create a task for it

PR #14737, opened 2024-04-05
Applications: nrf5340_audio: Remove bt ll acs nrf53

koffes:
Need to add doc before merging.


doc/nrf/libraries/bin/bt_ll_acs_nrf53/CHANGELOG.rst
koffes, line 1: @greg-fer, should we delete this file/folder?

rick1082 (approved):
This is a huge PR, thanks so much for the great work from everyone who contributed to Audio Controller Subsystem!

greg-fer (changes requested):
Shouldn't this module be removed in 2.9.0 at the earliest? The controller was replaced in 2.6.0, and we have the following guideline, provided by @tejlmand and @carlescufi and matching Zephyr:

> The Deprecated status is assigned to API that has gone through all maturity levels, but is being phased out. The deprecated API will be removed in one of future releases, **no earlier than two releases after the deprecation is announced** and only when the code has transitioned to not using the deprecated API. 

https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/releases_and_maturity/software_maturity.html#api-deprecation

doc/nrf/libraries/bin/bt_ll_acs_nrf53/CHANGELOG.rst
greg-fer: Yes, but we need to provide redirects and ensure nothing breaks. I can take care of this after we confirm the module should be indeed removed at this point.

alexsven: We have it confirmed that it can be removed now

alexsven (approved):

doc/nrf/libraries/bin/bt_ll_acs_nrf53/CHANGELOG.rst
greg-fer: OK, let me work on redirects and doc removal later today.

koffes:
@Frodevan : We need a separate PR to the test system which removes the use of the CONFIG_BT_LL_ACS_NRF53 flag.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -149,7 +149,11 @@ Serial LTE modem
 nRF5340 Audio
 -------------
 
-|no_changes_yet_note|
+* Updated:
+
+  * The BT_LL_ACS_NRF53 controller has been removed.
+    The only supported controller for LE Audio is :ref:`ug_ble_controller_softdevice`.
+    This enables use of standard tools for building, configuring and DFU.
```
greg-fer, line 156:
```suggestion
  * Removed the LE Audio controller for nRF5340 library.
    The only supported controller for LE Audio is :ref:`ug_ble_controller_softdevice`.
    This enables use of standard tools for building, configuring, and DFU.
```

greg-fer (approved):
@koffes , added commit with doc changes.
Please keep the double entry in the changelog, as we need a mention of the library removal in both areas, but kindly apply my edit.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
koffes: Done

PR #14741, opened 2024-04-05
manifest: Correct http_client edge-case behavior

glarsennordic:
Fix http_client edge-case behaviors, especially regarding NULL response from endpoint.

IRIS-8638

### DIFF
--- a/west.yml
+++ b/west.yml
@@ -63,7 +63,7 @@ manifest:
     # https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/modules.html
     - name: zephyr
       repo-path: sdk-zephyr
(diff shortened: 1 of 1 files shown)


glarsennordic:
@rlubos Since https://github.com/nrfconnect/sdk-nrf/pull/14757 updates the manifest to 76c76edcbabed3192d5cba95b66e94149d2d1879 which is later than 0f0b9f3c45c6884671a8205f9526be23f9463d15 in the sdk-nrf history, it seems to me I should close this PR, is that correct?

rlubos:
@glarsennordic Yes, you're totally right, I'm sorry I forgot to comment here.

glarsennordic:
No problem! Just wanted to confirm since I wasn't sure. Thanks for closing

PR #14742, opened 2024-04-05
feat(version): report NCS version on boot

gminn:
### Summary

The NCS version is now reported on boot. This metric is
useful when debugging issues that are dependent on the
NCS version.

### Test Plan

Enable debug logs and flashed a thingy91,
confirming correct version recorded

```
[00:00:00.563,140] <dbg> memfault_ncs_metrics: memfault_ncs_metrics_init: NCS version: 2.6.99
```


gminn:
Ignore, did not intend to open on upstream

PR #14749, opened 2024-04-08
tests: Bluetooth: Mesh: add LTO to metadata extraction test

HaavardRei:
Adds a testcase for metadata extraction with LTO enabled.

### DIFF
--- a/tests/subsys/bluetooth/mesh/metadata_extraction/testcase.yaml
+++ b/tests/subsys/bluetooth/mesh/metadata_extraction/testcase.yaml
@@ -14,3 +14,8 @@ tests:
   bluetooth.mesh.metadata_extraction_array:
     extra_configs:
       - CONFIG_COMP_DATA_LAYOUT_ARRAY=y
+  bluetooth.mesh.metadata_extraction_single_lto:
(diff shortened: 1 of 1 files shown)


tests/subsys/bluetooth/mesh/metadata_extraction/testcase.yaml
```diff
@@ -14,3 +14,7 @@ tests:
   bluetooth.mesh.metadata_extraction_array:
     extra_configs:
       - CONFIG_COMP_DATA_LAYOUT_ARRAY=y
+  bluetooth.mesh.metadata_extraction_single_lto:
+    extra_configs:
+      - CONFIG_COMP_DATA_LAYOUT_SINGLE=y
+      - CONFIG_LTO=y
```
ludvigsj, line 20: You also need `CONFIG_ISR_TABLES_LOCAL_DECLARATION` to test the case that Pavel fixed. Looks like LTO depends on this, have you checked that `CONFIG_LTO` actually gets the value `y`?

ludvigsj (approved):

PavelVPV (approved):

PR #14750, opened 2024-04-08
Bluetooth: Fix EnOcean module parsing sensor data as commissioning

ludvigsj:
This fixes an issue where the EnOcean module would incorrectly parse sensor data as if it were switch commissioning, which causes device slots to be filled up with garbage data if run in an environment where sensors are present.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -537,7 +537,9 @@ Bluetooth libraries and services
 
   * Updated the :ref:`bt_mesh_light_ctrl_srv_readme` model documentation to explicitly mention the Occupany On event.
 
-|no_changes_yet_note|
+* :ref:`bt_enocean_readme` library:
+
+  * Fixed an issue where the sensor data of a certain length was incorrectly parsed as switch commissioning.
 
 Bootloader libraries
 --------------------
--- a/subsys/bluetooth/enocean.c
+++ b/subsys/bluetooth/enocean.c
@@ -28,6 +28,9 @@ LOG_MODULE_REGISTER(bt_enocean);
 #define DATA_TYPE_LIGHT_LEVEL_SOLAR_CELL 0x04
 #define DATA_TYPE_OPTIONAL_DATA 0x3c
 
+#define DEVICE_TYPE_SWITCH 0xe215
+#define DEVICE_TYPE_SENSOR 0xe500
+
 #define FLAG_ACTIVE BIT(0)
 #define FLAG_DIRTY BIT(1)
 
@@ -142,18 +145,22 @@ static int auth(const struct bt_enocean_device *dev, uint32_t seq,
 	return 0;
 }
 
-static int handle_switch_commissioning(const struct bt_le_scan_recv_info *info,
-				       struct net_buf_simple *buf,
-				       bool has_short_name)
+static void handle_switch_commissioning(const struct bt_le_scan_recv_info *info,
+					struct net_buf_simple *buf,
+					bool has_short_name)
 {
+	if (!commissioning) {
+		return;
+	}
+
 	uint32_t seq = net_buf_simple_pull_le32(buf);
 	const uint8_t *key = net_buf_simple_pull_mem(buf, 16);
 
 	if (!has_short_name && buf->len != 6) {
-		return -EINVAL;
+		return;
 	}
 
-	return bt_enocean_commission(info->addr, key, seq);
+	(void)bt_enocean_commission(info->addr, key, seq);
 }
 
 static void handle_switch_data(const struct bt_le_scan_recv_info *info,
@@ -215,6 +222,24 @@ static void handle_switch_data(const struct bt_le_scan_recv_info *info,
 	enum bt_enocean_button_action action = status & BIT(0);
 
 	cb->button(dev, action, status >> 1, opt_data, opt_data_len);
+}
+
+static void handle_switch_payload(const struct bt_le_scan_recv_info *info,
+				  struct net_buf_simple *buf, const uint8_t *payload, uint8_t len,
+				  bool has_shortened_name)
+{
+	/* The data format is decided by a mix of lengths and bitfields */
+	if (len == 29 || (has_shortened_name && len == 23)) {
+		handle_switch_commissioning(info, buf, has_shortened_name);
+		return;
+	}
+
+	if (!has_shortened_name && !(payload[8] & (BIT_MASK(3) << 5)) && len >= 12 && len <= 16) {
+		handle_switch_data(info, buf, payload);
+		return;
+	}
+
+	LOG_DBG("Unknown switch payload");
 }
 
 struct data_entry {
@@ -378,13 +403,15 @@ static void adv_recv(const struct bt_le_scan_recv_info *info,
 	uint8_t len = net_buf_simple_pull_u8(buf);
 	uint8_t type = net_buf_simple_pull_u8(buf);
 	bool has_shortened_name = (type == BT_DATA_NAME_SHORTENED);
+	/* Upper two bytes of address indicate device type */
+	uint16_t device_type = sys_get_le16(&info->addr->a.val[4]);
 
 	/* An old version of EnOcean firmware would include the shortened name
 	 * AD type, skip this.
 	 */
 	if (has_shortened_name) {
 
-		if (buf->len < ((len - 1) + 2)) {
+		if (buf->len < ((len - 1) + 2) || device_type != DEVICE_TYPE_SWITCH) {
 			return;
 		}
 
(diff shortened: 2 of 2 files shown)


ludvigsj:
Just realized that there is an EnOcean sample in NCS, will run through that to check that it still works, DNM until then.

subsys/bluetooth/enocean.c
```diff
@@ -28,6 +28,9 @@ LOG_MODULE_REGISTER(bt_enocean);
 #define DATA_TYPE_LIGHT_LEVEL_SOLAR_CELL 0x04
 #define DATA_TYPE_OPTIONAL_DATA 0x3c
 
+#define DEVICE_TYPE_SWITCH 0xe215
```
omkar3141, line 31: May be worth to move these macros to header files if someone would like to use them in the application level. Like hooking up the scanner callback and doing some UI prompting when commissioning packets are received.

subsys/bluetooth/enocean.c
```diff
@@ -404,41 +429,16 @@ static void adv_recv(const struct bt_le_scan_recv_info *info,
...
-			 * switch commissioning, it could be sensor data, so we
-			 * should keep parsing.
-			 */
-			return;
-		}
-
-		net_buf_simple_restore(buf, &state);
+	switch (device_type) {
+	case DEVICE_TYPE_SENSOR:
+		handle_sensor_data(info, buf, payload, len + 1 - SIGNATURE_LEN);
```
omkar3141, line 434: May be also add humidity decoding in the "handle_sensor_data()"?

subsys/bluetooth/enocean.c
```diff
@@ -378,13 +401,15 @@ static void adv_recv(const struct bt_le_scan_recv_info *info,
 	uint8_t len = net_buf_simple_pull_u8(buf);
 	uint8_t type = net_buf_simple_pull_u8(buf);
 	bool has_shortened_name = (type == BT_DATA_NAME_SHORTENED);
+	/* Upper two bytes of address indicate device type */
+	uint16_t device_type = sys_get_le16(&info->addr->a.val[4]);
```
omkar3141, line 405: This is not a full proof method. This will only work for commissioning telegrams. But for data telegrams, if the devices are configured to transmit using RPA, module won't be able to infer device type.

We may have to refactor this module later to properly support RPA based data telegrams.

omkar3141 (changes requested):

subsys/bluetooth/enocean.c
ludvigsj: I thought the controller resolved the RPA and passed it pre-resolved to this? If that's not the case, we need bigger changes, since the address is used to lookup the key when switch data is received. If you'd like to check this, go ahead, and create a separate issue for supporting RPA switches if it doesn't already work as-is.

subsys/bluetooth/enocean.c
ludvigsj: Sounds like out of scope. Please create a separate issue for this.

subsys/bluetooth/enocean.c
ludvigsj: Sounds like out of scope, and they can easilly be moved later if we decide to expose this to the application. Please create a separate issue for this.

subsys/bluetooth/enocean.c
ludvigsj: (I wasn't able to check this since the PTM-215B I have available apparently can't be reconfigured using the Android App, but needs a PC-based NFC reader/writer)

subsys/bluetooth/enocean.c
```diff
@@ -217,6 +220,26 @@ static void handle_switch_data(const struct bt_le_scan_recv_info *info,
...
+static void handle_switch_payload(const struct bt_le_scan_recv_info *info,
+				  struct net_buf_simple *buf, const uint8_t *payload, uint8_t len,
+				  bool has_shortened_name)
+{
+	/* The data format is decided by a mix of lengths and bitfields */
+	if (len == 29 || (has_shortened_name && len == 23)) {
+		if (commissioning) {
+			handle_switch_commissioning(info, buf, has_shortened_name);
+		}
+		return;
```
Andrewpini, line 232: Nitpick: Could the commissioning flag check be moved inside `handle_switch_commissioning`? I feel it might make it a bit cleaner

subsys/bluetooth/enocean.c
```diff
@@ -217,6 +220,26 @@ static void handle_switch_data(const struct bt_le_scan_recv_info *info,
...
+	}
+
+	if (!has_shortened_name && !(payload[8] & (BIT_MASK(3) << 5)) && len >= 12 && len <= 16) {
```
Andrewpini, line 235: Nitpick: Could the `!(payload[8] & (BIT_MASK(3) << 5))` check be moved inside `handle_switch_data`? I feel it might make it a bit cleaner

subsys/bluetooth/enocean.c
```diff
@@ -378,13 +401,15 @@ static void adv_recv(const struct bt_le_scan_recv_info *info,
 	uint8_t len = net_buf_simple_pull_u8(buf);
```
Andrewpini, line 401: Should there not be a general check to ensure that the length field value aligns with the buf->len?

I imagine the following scenario:

1.  You receive an incoming switch data payload (!has_shortened_name) with a malformed length, E.G length_field_val is greater than buf_len.
2. The incorrect length_field_val is copied to len variable
3.  Before `handle_switch_data` is called, the check verifying that the message is between 12 and 16 passes, while in reality, the remaining buf->len is actually 3 for instance.
4. `handle_switch_data` tries to pull more than 3 bytes from the buffer, and the whole thing crashes.

Am I missing something or is this a real issue?

Andrewpini (changes requested):

subsys/bluetooth/enocean.c
ludvigsj: No, I think this is a real issue. I would, however, prefer it to be reported as a separate bug, as what I'm trying to do here is to fix the "module thinks sensor data is commissioning" problem specifically, while trying to keep all other behavior the same (even if "the same" == buggy, i.e. I haven't spent time looking at other potential bugs that are already here)

ludvigsj: (I don't want this PR to turn into a "fix every bug and problem in enocean.c"-PR)

ludvigsj: You can do it since you did the read-through of this and can probably write down the bug description better than I can

subsys/bluetooth/enocean.c
ludvigsj: No, I need it to proceed to the "LOG_DBG" in case these any of these bits of the payload are set (because that could in the future be a valid message for a new switch firmware that we don't support, and then it should log it so we (or customers) notice, whereas putting the check inside `handle_switch_data` would silently ignore it)

ludvigsj: Basically, this if is supposed to be `if (we are sure this is a switch data packet) {`, and then it needs this bit check.

Andrewpini (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -372,7 +372,9 @@ Binary libraries
 Bluetooth libraries and services
 --------------------------------
 
-|no_changes_yet_note|
+* :ref:`bt_enocean_readme`:
+
+  * Fixed an issue where sensor data of a certain length would incorrectly be parsed as switch commissioning.
```
mia-ko, line 377:
```suggestion
  * Fixed an issue where the sensor data of a certain length was incorrectly parsed as switch commissioning.
```

mia-ko: Btw, do we need to be more specific about at what length(s) this issue would be observed, instead of saying 'certain length'?

mia-ko (changes requested):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
ludvigsj: I'm not sure, but I can add it.
```suggestion
  * Fixed an issue where the sensor data of length 18 bytes was incorrectly parsed as switch commissioning.
```
?

mia-ko: of 18 bytes length :)

mia-ko (approved):

subsys/bluetooth/enocean.c
omkar3141: That is not how RPA works. After controller resolves RPA, it will inform the "identity address" corresponding to this RPA. This "identity" can be anything user defined. My impression is that, this requires some amount of refactoring to support RPA based data telegrams (commissioning telegrams are never sent using RPA). Needs a seperate task.

omkar3141 (approved):
We should capture pending items in tasks.

ludvigsj:
> We should capture pending items in tasks.

Will you create those tasks, or should I do it?

subsys/bluetooth/enocean.c
ludvigsj: NCSDK-27048

PR #14768, opened 2024-04-08
applications: nrf5340_audio: Use clock control to turn on HFClock

alexsven:
- Use clock control to enable high freq clock
- OCT-2847

### DIFF
--- a/applications/nrf5340_audio/src/utils/nrf5340_audio_dk.c
+++ b/applications/nrf5340_audio/src/utils/nrf5340_audio_dk.c
@@ -32,10 +32,6 @@ static int hfclock_config_and_start(void)
 		return ret;
 	}
 
-	nrfx_clock_hfclk_start();
-	while (!nrfx_clock_hfclk_is_running()) {
-	}
-
 	return 0;
 }
 


applications/nrf5340_audio/src/utils/nrf5340_audio_dk.c
```diff
@@ -20,6 +23,39 @@ LOG_MODULE_REGISTER(nrf5340_audio_dk, CONFIG_MODULE_NRF5340_AUDIO_DK_LOG_LEVEL);
 
 static struct board_version board_rev;
 
+static int hfclock_init(void)
```
koffes, line 26: Could we have a word on how this differs from the previous solution @rick1082 ?

koffes (changes requested):

andvib (approved):

alexsven:
Should not be merged until https://github.com/nrfconnect/sdk-nrf/pull/14737 has been merged

applications/nrf5340_audio/src/utils/nrf5340_audio_dk.c
rick1082: Now we don't need to start HFCLK all the time, since that was prepared for PCFT.
So we can remove those part now, probably we can also get some benefits for power consumption.

gWacey (approved):

koffes (approved):

PR #14769, opened 2024-04-08
emds: add rram support for emds

alxelax:
Commit adds rram support for emds for nrf54l


subsys/emds/emds_flash.c
```diff
@@ -283,22 +339,23 @@ static int entry_wrt(struct emds_fs *fs, uint16_t id, const void *data, size_t l
...
-	memset(cmp_buf, 0, sizeof(cmp_buf));
-	if (!memcmp(entry, cmp_buf, fs->ate_size)) {
+	memset(entry, 0, sizeof(struct emds_ate));
+	if (!memcmp(entry, cmp_buf, sizeof(struct emds_ate))) {
 		return ATE_TYPE_INVALIDATED;
 	}
 
-	memset(cmp_buf, fs->flash_params->erase_value, sizeof(cmp_buf));
-	if (!memcmp(entry, cmp_buf, fs->ate_size)) {
+	memset(entry, fs->flash_params->erase_value, sizeof(struct emds_ate));
```
ludvigsj, line 353: Is there a reason why you flipped the usage of `entry` and `cmp_buf` around? As far as I can see, the behavior of this function is identical before and after your changes, but now you need the final `memcpy` on line 358 since you don't read directly into `entry`. If there is a reason for this that I can't see, then I'd suggest to rename `cmp_buf` to something like `read_entry` or `stored_entry` or `flash_data`, since it's no longer a compare buffer.

ludvigsj (changes requested):

tests/subsys/emds/emds_flash/boards/nrf52840dk_nrf52840.conf
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+################################################################################
+# Application overlay - nrf52840dk_nrf52840
+
+CONFIG_SOC_FLASH_NRF_PARTIAL_ERASE=y
```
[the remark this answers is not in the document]
alxelax: Yep, it wasn't enabled before. However, ble mesh default setting mandates partial erasing where it is possible. 
I'm sure it is better to test behavior with KConfig options, those we provide to customers.
I guess it was just lost occasionally in the initial implementation.

subsys/emds/emds_flash.c
alxelax: Yes, there is. `sizeof(struct emds_ate) < fs->ate_size` for rram. The existing implementation corrupted stack memory that caused hardfaults in unexpected places.
As I said before, emds has been based on assumption the ATE structure is bigger than writing granularity of persistent memory. That's not the truth for flash emulator on rram. 

In general, my implementation didn't add more data copying. Yes, it added one more on line 358 but removed it on line 368.

ludvigsj: Right, I missed the size being different. Then I guess my only suggestion would be to rename `cmp_buf` which is a bit confusingly named now after you flipped things around

ludvigsj (approved):

subsys/emds/emds_flash.c
Andrewpini: I also thought this looked strange at first, but from a functional view it looks fine. I would claim that it is a bit more vulnerable for introducing bugs if this is is changed in the future, but im ok with it.

Andrewpini (approved):
This looks good to me 👍

PR #14776, opened 2024-04-09
applications: nrf5340_audio: Send correct disconnect reason on auth f…

alexsven:
…ailed

- Use 0x05 when disconnecting after security failed
- Don't try to disconnect MCS if it hasn't started yet
- OCT-2970

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_content_control/bt_content_ctrl.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_content_control/bt_content_ctrl.c
@@ -86,7 +86,10 @@ int bt_content_ctrl_conn_disconnected(struct bt_conn *conn)
 
 	if (IS_ENABLED(CONFIG_BT_MCC)) {
 		ret = bt_content_ctrl_media_conn_disconnected(conn);
-		if (ret) {
+		/* Try to reset MCS state. -ESRCH is returned if MCS hasn't been discovered
+		 * yet, and shouldn't cause an error print
+		 */
+		if (ret && ret != -ESRCH) {
 			LOG_ERR("bt_content_ctrl_media_conn_disconnected failed with %d", ret);
 		}
 	}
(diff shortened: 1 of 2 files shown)


applications/nrf5340_audio/src/bluetooth/bt_content_control/bt_content_ctrl.c
```diff
@@ -86,7 +86,7 @@ int bt_content_ctrl_conn_disconnected(struct bt_conn *conn)
 
 	if (IS_ENABLED(CONFIG_BT_MCC)) {
 		ret = bt_content_ctrl_media_conn_disconnected(conn);
-		if (ret) {
+		if (ret && ret != -ESRCH) {
```
rick1082, line 89: May I know in which condition would get -ESRCH?
And would that cause issue ?

gWacey: This should not cause an issue as the returned index is < 0 and hence logs a warning in the disconnect. The -ESRCH is then passed to the disconnect caller and as it is not 0 it errors. Finally the disconnect returns to the event handler where it again logs an error, the event handler does not return any value, and then the disconnect call-back finishes cleaning up.

gWacey (approved):

applications/nrf5340_audio/src/bluetooth/bt_content_control/bt_content_ctrl.c
andvib: Can you add a comment to why `-ESRCH` is not an error? gWacey explanation sounds right, but I don't think the code is self explanatory on this one

andvib (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_content_control/bt_content_ctrl.c
alexsven: -ESRCH is returned when we try to do a disconnect on a conn pointer that doesn't exist in the MCS overview since security failed before it even got to the discovery part. I will add a comment

rick1082 (approved):

koffes (approved):

andvib (approved):

PR #14780, opened 2024-04-09
Misc. TF-M+54L support

SebastianBoe:
test_crypto: PR-626

Misc. TF-M+54L support


tests/tfm/tfm_psa_test/CMakeLists.txt
```diff
@@ -100,7 +100,7 @@ ExternalProject_Add(tfm_psa_arch_test_app
                 -DCROSS_COMPILE=${TFM_TOOLCHAIN_PATH}/${TFM_TOOLCHAIN_PREFIX}
                 -DPSA_TOOLCHAIN_FILE=${TFM_BINARY_DIR}/api_ns/cmake/${TFM_TOOLCHAIN_NS_FILE}
                 -DQCBOR_PATH${QCBOR_PATH_TYPE}=${CONFIG_TFM_QCBOR_PATH}
-                -DCMAKE_BUILD_TYPE=RelWithDebInfo
+                -DCMAKE_BUILD_TYPE=Debug
```
mswarowsky, line 103: So we are always building as Debug, maybe that should be linked to the Kconfig `CONFIG_DEBUG_OPTIMIZATIONS`, if that can be used here?

tests/tfm/tfm_psa_test/testcase.yaml
```diff
@@ -16,39 +12,109 @@ tests:
     tags: tfm_lvl1
     extra_args: "CONFIG_TFM_PSA_TEST_PROTECTED_STORAGE=y CONFIG_TFM_ISOLATION_LEVEL=1"
     timeout: 120
+    platform_allow: >
+      nrf5340dk_nrf5340_cpuapp_ns
+      nrf9160dk_nrf9160_ns
+    integration_platforms:
+      - nrf5340dk_nrf5340_cpuapp_ns
+      - nrf9160dk_nrf9160_ns
+      - nrf54l15pdk_nrf54l15_cpuapp_ns
```
mswarowsky: Reading the doc (https://docs.zephyrproject.org/latest/develop/test/twister.html#test-cases) 

`platform_allow: <list of platforms>`
_Set of platforms that this test case should only be run for. Do not use this option to limit testing or building in CI due to time or resource constraints, this option should only be used if the test or sample can only be run on the allowed platform and nothing else._

`integration_platforms: <YML list of platforms/boards>`
_This option limits the scope to the listed platforms when twister is invoked with the --integration option. Use this instead of platform_allow if the goal is to limit scope due to timing or resource constraints._

`nrf54l15pdk_nrf54l15_cpuapp_ns` should be part of the platform_allow as well, maybe one of the testers could give a comment @magnev , @stephen-nordic

SebastianBoe: fixed

mswarowsky (approved):

tests/tfm/tfm_psa_test/CMakeLists.txt
SebastianBoe: done

vili-nordic (approved):

modules/trusted-firmware-m/tfm_boards/nrf54l15_cpuapp/cpuarch.cmake
```diff
@@ -1,21 +1,10 @@
-#-------------------------------------------------------------------------------
-# Copyright (c) 2024, Nordic Semiconductor ASA.
+#
+# Copyright (c) 2021 - 2024, Nordic Semiconductor ASA.
```
frkv, line 2: We shall not use ranges in copyrights. Just use first year when a feature was added

SebastianBoe: fixed

PR #14781, opened 2024-04-09
samples: crypto: Crypto samples cleanup

stephen-nordic:
- Use correct tags when running on CC3xx/CRACEN/Oberon
- Add _ns boards for ecjpake and spake2p
- Use CRACEN for ecjpake on nRF54L15

test_crypto: PR-625

### DIFF
--- a/samples/crypto/persistent_key_usage/sample.yaml
+++ b/samples/crypto/persistent_key_usage/sample.yaml
@@ -5,12 +5,11 @@ sample:
     to the PSA keystore.
   name: Persistent key example
 tests:
-  sample.persistent_key_usage:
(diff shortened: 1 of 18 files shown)


samples/crypto/persistent_key_usage/sample.yaml
```diff
@@ -6,7 +6,7 @@ sample:
   name: Persistent key example
 tests:
   sample.persistent_key_usage:
-    tags: introduction psa cc3xx
```
magnev, line 9: This sample.yaml should also be split into cracen and cc3xx

magnev (approved):

mswarowsky (approved):
Makes the config more clear, nice work

PR #14793, opened 2024-04-10
kernel: Add NCS boot banner

nordicjm:
Adds an enhanced boot banner which can be configured to show application name and version, NCS version, and underlying zephyr version.

### DIFF
--- a/cmake/gen_commit_h.cmake
+++ b/cmake/gen_commit_h.cmake
@@ -1,0 +1,43 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+cmake_minimum_required(VERSION 3.20.0)
+
+find_package(Git QUIET)
+if(GIT_FOUND)
+  execute_process(
+    COMMAND ${GIT_EXECUTABLE} rev-parse HEAD
+    WORKING_DIRECTORY                ${COMMIT_PATH}
+    OUTPUT_VARIABLE                  ${COMMIT_TYPE}_COMMIT
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    ERROR_STRIP_TRAILING_WHITESPACE
+    ERROR_VARIABLE                   stderr
+    RESULT_VARIABLE                  return_code
+  )
+
+  if(return_code)
+    message(STATUS "git rev-parse failed: ${stderr}")
+  elseif(NOT "${stderr}" STREQUAL "")
+    message(STATUS "git rev-parse warned: ${stderr}")
+  endif()
+
+  # Limit to 12 characters
+  string(SUBSTRING "${${COMMIT_TYPE}_COMMIT}" 0 12 ${COMMIT_TYPE}_COMMIT)
+endif()
+
+file(READ ${NRF_DIR}/cmake/commit.h.in commit_content)
+string(CONFIGURE "${commit_content}" commit_content)
+string(CONFIGURE "${commit_content}" commit_content)
+
+if(EXISTS ${OUT_FILE})
+  file(READ ${OUT_FILE} current_contents)
+else()
+  set(current_contents "")
+endif()
+
+if(NOT "${current_contents}" STREQUAL "${commit_content}")
+  file(WRITE ${OUT_FILE} "${commit_content}")
+endif()
--- a/cmake/version.cmake
+++ b/cmake/version.cmake
@@ -19,5 +19,28 @@ add_custom_command(
     DEPENDS ${NRF_DIR}/VERSION ${git_dependency}
 )
 add_custom_target(ncs_version_h DEPENDS ${PROJECT_BINARY_DIR}/include/generated/ncs_version.h)
-
 add_dependencies(version_h ncs_version_h)
+
+add_custom_command(
+  OUTPUT ${PROJECT_BINARY_DIR}/include/generated/ncs_commit.h
+  COMMAND ${CMAKE_COMMAND} -DZEPHYR_BASE=${ZEPHYR_BASE} -DNRF_DIR=${NRF_DIR}
+    -DOUT_FILE=${PROJECT_BINARY_DIR}/include/generated/ncs_commit.h
+    -DCOMMIT_TYPE=NCS
+    -DCOMMIT_PATH=${NRF_DIR}
+    -P ${ZEPHYR_NRF_MODULE_DIR}/cmake/gen_commit_h.cmake
+    DEPENDS ${NRF_DIR}/VERSION ${git_dependency}
+)
+add_custom_target(ncs_commit_h DEPENDS ${PROJECT_BINARY_DIR}/include/generated/ncs_commit.h)
+add_dependencies(version_h ncs_commit_h)
+
+add_custom_command(
+  OUTPUT ${PROJECT_BINARY_DIR}/include/generated/zephyr_commit.h
+  COMMAND ${CMAKE_COMMAND} -DZEPHYR_BASE=${ZEPHYR_BASE} -DNRF_DIR=${NRF_DIR}
+    -DOUT_FILE=${PROJECT_BINARY_DIR}/include/generated/zephyr_commit.h
+    -DCOMMIT_TYPE=ZEPHYR
+    -DCOMMIT_PATH=${ZEPHYR_BASE}
+    -P ${ZEPHYR_NRF_MODULE_DIR}/cmake/gen_commit_h.cmake
+    DEPENDS ${ZEPHYR_BASE}/VERSION ${git_dependency}
+)
+add_custom_target(zephyr_commit_h DEPENDS ${PROJECT_BINARY_DIR}/include/generated/zephyr_commit.h)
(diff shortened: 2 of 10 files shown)


nordicjm:
@gchwier This looks to need an update to the `test-fw-nrfconnect-boot` tests, though would hold off until people have approved the PR

kernel/banner.c
tejlmand, line 1: This file contains same function names as its upstream counter part.

But the upstream `boot_banner.c` function is not upstream.
Both versions of `banner.c` are compiled and linked, which means that the linker picks the first function which satisfies linking. This happens to work because nrf banner.c is added to zephyr lib, which is inside the whole-archive flags, and the other is outside (libkernel.a).

This is a risky approach, which I don't like.
I think upstream `banner.c` should instead be made such a way that a downstream project can easily replace it.

From build:
```

[37/145] ccache /opt/zephyr-sdk-0.16.1/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc ... -MF zephyr/CMakeFiles/zephyr.dir/projects/github/ncs/nrf/kernel/banner.c.obj.d -o zephyr/CMakeFiles/zephyr.dir/projects/github/ncs/nrf/kernel/banner.c.obj -c /projects/github/ncs/nrf/kernel/banner.c
...
[105/145] ccache /opt/zephyr-sdk-0.16.1/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc ... -MF zephyr/kernel/CMakeFiles/kernel.dir/banner.c.obj.d -o zephyr/kernel/CMakeFiles/kernel.dir/banner.c.obj -c /projects/github/ncs/zephyr/kernel/banner.c
...
[144/145] : && ccache /opt/zephyr-sdk-0.16.1/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc  ... -o zephyr/zephyr.elf  ...  -Wl,--whole-archive  app/libapp.a  zephyr/libzephyr.a  ... -Wl,--no-whole-archive  zephyr/kernel/libkernel.a ...
```

cmake/version.cmake
```diff
@@ -21,3 +21,41 @@ add_custom_command(
...
+  OUTPUT ${PROJECT_BINARY_DIR}/include/generated/zephyr_commit.h
+  COMMAND ${CMAKE_COMMAND} -DZEPHYR_BASE=${ZEPHYR_BASE} -DNRF_DIR=${NRF_DIR}
+    -DOUT_FILE=${PROJECT_BINARY_DIR}/include/generated/zephyr_commit.h
+    -DCOMMIT_TYPE=ZEPHYR
+    -DCOMMIT_PATH=${ZEPHYR_BASE}
+    -P ${ZEPHYR_NRF_MODULE_DIR}/cmake/gen_commit_h.cmake
+    DEPENDS ${ZEPHYR_BASE}/VERSION ${git_dependency}
+)
+add_custom_target(zephyr_commit_h DEPENDS ${PROJECT_BINARY_DIR}/include/generated/zephyr_commit.h)
+add_dependencies(version_h zephyr_commit_h)
```
tejlmand: A lot of duplication here.

```suggestion
set(types ncs zephyr)
set(types_upper NCS ZEPHYR)
set(path_vars NRF_DIR ZEPHYR_BASE)

if(${APP_VERSION_NUMBER})
  set(types ${types} app)
  set(types_upper ${types_upper} APP)
  set(path_vars ${path_vars} CMAKE_SOURCE_DIR)
endif()

foreach(type IN ZIP_LISTS types types_upper path_vars)
  add_custom_command(
    OUTPUT ${PROJECT_BINARY_DIR}/include/generated/${type_0}_commit.h
    COMMAND ${CMAKE_COMMAND} -DZEPHYR_BASE=${ZEPHYR_BASE} -DNRF_DIR=${NRF_DIR}
      -DOUT_FILE=${PROJECT_BINARY_DIR}/include/generated/${type_0}_commit.h
      -DCOMMIT_TYPE=${type_1}
      -DCOMMIT_PATH=${${type_2}}
      -P ${ZEPHYR_NRF_MODULE_DIR}/cmake/gen_commit_h.cmake
      DEPENDS ${${type_2}}/VERSION ${git_dependency}
  )
  add_custom_target(ncs_commit_h DEPENDS ${PROJECT_BINARY_DIR}/include/generated/ncs_commit.h)
  add_dependencies(version_h ${type_0}_commit_h)
endforeach()
```

Note, it's possible to use `foreach(type TUPPER path_var IN ZIP_LISTS types types_upper path_vars)` and then refer to those as `type` (was: type_0), `TUPPER` (was: type_1), `path_var` (was: type_2) in above code.

cmake/gen_commit_h.cmake
```diff
@@ -26,6 +26,27 @@ if(GIT_FOUND)
 
   # Limit to 12 characters
   string(SUBSTRING "${${COMMIT_TYPE}_COMMIT}" 0 12 ${COMMIT_TYPE}_COMMIT)
+
+  execute_process(
+    COMMAND ${GIT_EXECUTABLE} describe --abbrev=1 --always --dirty
```
tejlmand, line 31: this will not work as expected in all cases.
Please keep out of this PR and open separately for further discussions.

tejlmand (changes requested):

cmake/gen_commit_h.cmake
nordicjm: in what cases will it not work?

nordicjm: Has been removed

kernel/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+if(CONFIG_NCS_BOOT_BANNER)
+  set(ZEPHYR_CURRENT_LIBRARY zephyr)
+  zephyr_library_sources(banner.c)
+  unset(ZEPHYR_CURRENT_LIBRARY)
```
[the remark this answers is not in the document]
nordicjm: Fixed

LuDuda (approved):
This is great! :+1:

cmake/version_app.cmake
```diff
@@ -0,0 +1,40 @@
+if(${APP_VERSION_NUMBER})
```
tejlmand, line 1: why do we need a dedicated `version_app.cmake`.

Also, if we do need it, then it should have copyright and license header.

kernel/banner.c
```diff
@@ -29,16 +34,20 @@ void boot_banner(void)
 
 #if defined(CONFIG_NCS_APPLICATION_BOOT_BANNER_STRING)
 	printk("*** Booting " CONFIG_NCS_APPLICATION_BOOT_BANNER_STRING " v"
-	       APP_VERSION_STRING " ***\n");
+	       APP_VERSION_STRING
+#if defined(NCS_APPLICATION_BOOT_BANNER_GIT_REPO)
+	       "-" APP_COMMIT_STRING
```
tejlmand, line 39: instead of this, wouldn't it be better to indicate that there was no git repo found ?

I believe something like this:
```
*** Booting my_app v1.0.0 - unknown commit
```

will catch attention earlier that something might be wrong, compared to:
```
*** Booting my_app v1.0.0
```

(None, Null, empty, etc are also options)

nordicjm: It is omitted because it might not be in a git repo

cmake/version_app.cmake
nordicjm: Because the user application might not reside in a git repo, however zephyr and ncs do

kernel/banner.c
tejlmand: Exactly, as I wrote:
> wouldn't it be better to indicate that there was no git repo found ?

So I understand why it's omitted, what i'm asking is to indicate that no git repo was found, exactly for this reason:
> will catch attention earlier that something might be wrong, 

Feel free to decide how to indicate that the app is not in a git repo, but I think it should be indicated.

cmake/version_app.cmake
tejlmand: > Because the user application might not reside in a git repo, however zephyr and ncs do

I don't see why the question of application being in a git repo or not should decide whether the version files cannot be unified ?

If this is because you think the `git` handling doesn't belong in `version.cmake` then I prefer that upstream is updated to support a common `git` CMake function, which can then be reused downstream, and only the need for a single `version.cmake` file.

kernel/banner.c
nordicjm: Added

cmake/version_app.cmake
nordicjm: It is need to know if a commit file is generated and what message to show

nordicjm:
@gchwier Would you be able to prepare updates to the test-fw-nrfconnect-boot tests for this change?

gchwier:
> @gchwier Would you be able to prepare updates to the test-fw-nrfconnect-boot tests for this change?

I've updated it two weeks ago. Now regression test on test-fw-nrfconnect-boot are passed (previous fail was because of some problems in lab)

tejlmand (approved):

PR #14795, opened 2024-04-10
samples: fast_pair: input_device: Add nRF54L15 PDK rev 0.3.0 support

mkapala-nordic:
Added nRF54L15 PDK revision 0.3.0 support.
Workaround present in the previous board configuration was only applicable for revision 0.2.x and is no longer needed for the revision 0.3.0.

Jira: NCSDK-26698


zycz (approved):

MarekPieta (approved):

kapi-no (approved):

kapi-no:
General question:

Should we add the new target to the sample.yaml? The rationale for it is that it will appear in the Supported Development Kits table in the Readme file. Customers may be not aware that they need to pass an additional `@0.3.0` suffix to the board name.

Could we align the approach with tech writes?

annwoj:
> General question:
> 
> Should we add the new target to the sample.yaml? The rationale for it is that it will appear in the Supported Development Kits table in the Readme file. Customers may be not aware that they need to pass an additional `@0.3.0` suffix to the board name.
> 
> Could we align the approach with tech writes?

Agreed, this information will be included in the Moonlight getting started guide but at the same time users shouldn't need to refer to a separate page just to check for it, nor should they need to remember about it. I think it's better if we include all the supported build targets for the current state of development.

mkapala-nordic:
Closing as https://github.com/nrfconnect/sdk-nrf/pull/14826 introduced global workaround for 0.2.x

PR #14800, opened 2024-04-10
manifest: update zephyr - calc coverage for qdec

nordic-piks:
Enable coverage calculation for qdec sample


nordic-segl (approved):

nordic-piks:
Included in #14798

PR #14803, opened 2024-04-10
manifest: Update sdk-zephyr

de-nordic:
Brings in change where inclusion of partition manager in cmake/modules/kernel.cmake depends on
SB_CONFIG_PARTITION_MANAGER=y


tejlmand (approved):

de-nordic:
Not the right approach.

PR #14812, opened 2024-04-11

anangl:
Pull in a fix in pinctrl that prevents the nrf_qspi_nor driver from failing to initialize when the flash chip is configured to work in non-Quad mode.


de-nordic (approved):

anangl:
Rebased and updated with SHA.

nika-nordic (approved):

PR #14821, opened 2024-04-11
manifest: Update zephyr with nrf54h20 DCACHE enabling

nordic-krch:
Update zephyr to enable DCACHE by default for nrf54h20.


nordic-krch:
SHA updated.

rlubos (approved):

PR #14823, opened 2024-04-11
applications: nrf5340_audio: Only expose source PAC if BIDIR

alexsven:
- Set ASE_SRC_COUNT to 0 if not BIDIR
- OCT-2972


koffes (approved):

rick1082 (approved):
Just curious: do we face issues on this when doing IOP test with other vendor?

alexsven:
> Just curious: do we face issues on this when doing IOP test with other vendor?

Yes

PR #14826, opened 2024-04-11
lib: dk_leds_and_buttons: nrf54l15 config

KAGA164:
This adds global workaround for nRF54l15
DK in version 0.2.x. since these versions
are supported together with revision 0.3.0.


miha-nordic (approved):

lib/dk_buttons_and_leds/Kconfig
```diff
@@ -21,6 +21,8 @@ config DK_LIBRARY_DYNAMIC_BUTTON_HANDLERS
 
 config DK_LIBRARY_BUTTON_NO_ISR
 	bool "Poll buttons unconditionally (no interrupts) [EXPERIMENTAL]"
+	# Workaround for buttons on nRF54L15 PDK in revision 0.2.x.
+	default y if BOARD_NRF54L15PDK_NRF54L15_CPUAPP && (BOARD_REVISION = "0.2.0" || BOARD_REVISION = "0.2.1")
```
MarekPieta, line 25: `default y if BOARD_NRF54L15PDK_NRF54L15_CPUAPP && (BOARD_REVISION = "0.2.0")`?
Seems like Zephyr generalizes to 2 revisions: https://github.com/nrfconnect/sdk-zephyr/blob/main/boards/arm/nrf54l15pdk_nrf54l15/revision.cmake

MarekPieta (approved):

lib/dk_buttons_and_leds/Kconfig
KAGA164: Yes, but there is also physically revision 2.0.1 so added it to be safe if someone change it in upstream

kapi-no:
Could you also align Find My samples?

maje-emb (approved):

pdunaj (approved):

KAGA164:
> Could you also align Find My samples (`find-my/samples`) and the Bluetooth Fast Pair Input Device sample (`nrf/samples/bluetooth/fast_pair/input_device`)?
> 
> CC: @mkapala-nordic

@kapi-no Fast Pair sample is already updated. Find My can be done in separate PR

PR #14829, opened 2024-04-11
applications: nrf5340_audio: Disable play/pause only in walkie-talkie…

alexsven:
… mode

- Re-enable play/pause for bidirectional mode
- Reduce LOG_WRN for thrown packets to every 100 packets
- OCT-2980


rick1082:
According to the implementation in unicast_client_start/stop(), we only focus on handing sink_stream. Maybe we should also handle the source stream? I will make a test tomorrow.

rick1082:
We might still not able to have play/pause for bidirectional since we only disable sink side, and the source will still send packet but gateway is not in streaming state.

rick1082 (approved):

PR #14830, opened 2024-04-11
samples: net: http_server: support for native_sim

DematteisGiacomo:
Adding support for native_sim board.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -398,7 +398,10 @@ Networking samples
 
 * :ref:`http_server` sample:
 
-  * Added ``DNS_SD_REGISTER_TCP_SERVICE`` so that mDNS services can locate and address the server using its hostname.
+  * Added:
+
+    * ``DNS_SD_REGISTER_TCP_SERVICE`` so that mDNS services can locate and address the server using its hostname.
+    * Support for the :ref:`native simulator <zephyr:native_sim>` board.
 
   * Updated:
 
--- a/samples/net/http_server/boards/native_sim.conf
+++ b/samples/net/http_server/boards/native_sim.conf
@@ -1,0 +1,17 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# LED Control off
+CONFIG_DK_LIBRARY=n
+
+# Network address config
+CONFIG_NET_CONFIG_SETTINGS=y
+CONFIG_NET_CONFIG_NEED_IPV4=y
+CONFIG_NET_CONFIG_MY_IPV4_ADDR="192.0.2.1"
+CONFIG_NET_HOSTNAME_ENABLE=y
+CONFIG_NET_HOSTNAME_UNIQUE=n
+CONFIG_NET_HOSTNAME="httpserver"
+CONFIG_POSIX_UNAME=n
--- a/samples/net/http_server/src/main.c
+++ b/samples/net/http_server/src/main.c
@@ -25,7 +25,9 @@
 #include <zephyr/net/http/client.h>
 #include <zephyr/net/http/parser.h>
 
+#if defined(CONFIG_DK_LIBRARY)
 #include <dk_buttons_and_leds.h>
+#endif /* defined(CONFIG_DK_LIBRARY) */
 
 #if defined(CONFIG_POSIX_API)
 #include <zephyr/posix/arpa/inet.h>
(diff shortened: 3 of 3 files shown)


samples/net/http_server/boards/native_sim.conf
```diff
@@ -0,0 +1,23 @@
...
+CONFIG_NET_CONFIG_MY_IPV4_ADDR="192.0.2.1"
+CONFIG_NET_HOSTNAME_ENABLE=y
+CONFIG_NET_HOSTNAME_UNIQUE=n
+CONFIG_NET_HOSTNAME="httpserver"
+CONFIG_POSIX_UNAME=n
+
+# DNS
+CONFIG_DNS_RESOLVER=y
+CONFIG_DNS_SERVER_IP_ADDRESSES=y
+CONFIG_DNS_SERVER1="8.8.8.8"
```
simensrostad, line 22: Do we need this? AFAIK we don't do any DNS request from the server code that require this.
I also don't see mDNS nor SD being enabled. Something that we use for in the Wi-Fi config to be able to be addressed with the hostname in HTTP requests.

samples/net/http_server/src/main.c
```diff
@@ -159,12 +161,14 @@ static int led_update(uint8_t index, uint8_t state)
...
 	}
 
+#if defined(CONFIG_DK_LIBRARY)
 	ret = dk_set_led(index, led_states[index]);
 	if (ret) {
 		LOG_ERR("Failed to update LED %d state to %d", index, led_states[index]);
 		FATAL_ERROR();
 		return -EIO;
 	}
+#endif
```
jtguggedal, line 171: You could add `/* defined(CONFIG_DK_LIBRARY) */` here and other places

lemrey: Perhaps when plan to have `native_sim` board support in a sample, it might be easier to just avoid using LED/buttons altogether, so we avoid `#ifdefs` right and left, and in general having to maintain code that runs only on some of the boards that are supported by the sample.

It's nice to use LEDs for this sample when it's running on the kit, but it's not so nice to have `#ifdefs` in `main.c` files.
What if instead of a `/led/` endpoint this sample had a `/print/` endpoint? It could serve the purposes of the sample, be easier to verify in a test (instead of a LED turning on), and avoid all the LED-related code and `#ifdef` in `main.c`.

samples/net/http_server/boards/native_sim.conf
DematteisGiacomo: yes you are right. Not needed

bjda (approved):

jtguggedal (approved):

samples/net/http_server/src/main.c
jtguggedal: Yeah, agree with the ifdefs. An alternative could be to have support for `native_sim` in the DK library and just let it print something like `LED 1: ON` or whatever

lemrey:
Could have a changelog entry for this

samples/net/http_server/src/main.c
lemrey: That could work too

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -400,6 +400,8 @@ Networking samples
 
   * Added ``DNS_SD_REGISTER_TCP_SERVICE`` so that mDNS services can locate and address the server using its hostname.
 
+  * Added support for native_sim board.
```
divipillai, line 403:
```suggestion
  * Added:  

    * ``DNS_SD_REGISTER_TCP_SERVICE`` so that mDNS services can locate and address the server using its hostname.
    * Support for the :ref:`native simulator <zephyr:native_sim>` board.
```

divipillai (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -398,7 +398,11 @@ Networking samples
 
 * :ref:`http_server` sample:
 
-  * Added ``DNS_SD_REGISTER_TCP_SERVICE`` so that mDNS services can locate and address the server using its hostname.
+  * Added:
+
+    * ``DNS_SD_REGISTER_TCP_SERVICE`` so that mDNS services can locate and address the server using its hostname.
+
+    * Support for the :ref:`native simulator <zephyr:native_sim>` board.
```
divipillai, line 405:
```suggestion
    * ``DNS_SD_REGISTER_TCP_SERVICE`` so that mDNS services can locate and address the server using its hostname.
    * Support for the :ref:`native simulator <zephyr:native_sim>` board.
```
Could avoid the space between the sentences.

PR #14831, opened 2024-04-11
[backport-v2.5] manifest: Update zephyr revision

anangl:
Pull in a fix in pinctrl that prevents the nrf_qspi_nor driver from failing to initialize when the flash chip is configured to work in non-Quad mode.


anangl:
Rebased.

mariusz-nordicsemi (approved):

PR #14839, opened 2024-04-12
manifest: update sdk-zephyr

jukkar:
Include following change:

NCSIDB-1249: Build warning when disabling IPv6
https://github.com/nrfconnect/sdk-zephyr/pull/1616


jukkar:
Newer PRs will update sdk-zephyr ref so closing.

PR #14840, opened 2024-04-12
doc: install: edits from review

greg-fer:
Applied edits from the recent getting started testing. Requested by @ihansse and tracked in NCSDK-27019.


bencefr (approved):
VS Code related changes are good.

ihansse (approved):
Looks a lot better, but we should continue improving this further

gmarull (approved):

PR #14849, opened 2024-04-12
applications: nrf5340_audio: Use BAP presets for sample rate

alexsven:
- If a BAP preset is used, it will override PREF_SAMPLE_RATE
- Added more BAP presets for broadcast
- Increased LC3 bitrate range to include BAP 48_6_1
- Moved BAP presets for broadcast to header file
- OCT-2981


applications/nrf5340_audio/src/audio/Kconfig
```diff
@@ -212,7 +218,7 @@ visible if SW_CODEC_LC3
 
 config LC3_BITRATE_MAX
 	int "Max bitrate for LC3"
-	default 96000
+	default 144000
```
koffes, line 221: We need to adjust CI to do one run on this new max value.

applications/nrf5340_audio/src/audio/Kconfig
```diff
@@ -45,6 +45,12 @@ config AUDIO_MAX_PRES_DLY_US
 
 choice AUDIO_SYSTEM_SAMPLE_RATE
 	prompt "System audio sample rate"
```
koffes, line 47: We should adjust the prompts to make it more clear what is the HW and what is "over the air"

applications/nrf5340_audio/src/audio/Kconfig
koffes: Should this value be 124000?

koffes (changes requested):

applications/nrf5340_audio/src/audio/Kconfig
alexsven: Changed prompt for BT_AUDIO_SAMPLE_RATE to include BT

applications/nrf5340_audio/src/bluetooth/Kconfig
```diff
@@ -34,11 +34,19 @@ config BT_AUDIO_PREF_SAMPLE_RATE_VALUE
 	default 0x08 if BT_AUDIO_PREF_SAMPLE_RATE_48KHZ
 
 choice BT_AUDIO_PREF_SAMPLE_RATE
-	prompt "Preferred audio sample rate"
+	prompt "Preferred BT audio sample rate"
```
rick1082, line 37: Nitpick: Maybe we can use LE audio instead of BT audio :P

rick1082 (approved):

applications/nrf5340_audio/src/bluetooth/Kconfig
alexsven: It's because the config is called BT_AUDIO_

koffes (approved):

PR #14850, opened 2024-04-12
applications: asset_tracker_v2: Add check for GNSS heading validity

tokangas:
Only accurate enough GNSS heading is considered valid and sent to the nRF Cloud.

Added accuracy values for heading, altitude and speed into internal data structures. Aligned naming and order of the fields in the data structures.


jtguggedal:
Could add a release note for this

tokangas:
This originates from a question from a customer. Currently, the heading from GNSS is always sent to the cloud. Also, the heading accuracy value from GNSS is not used. When the device is not moving or is moving very slowly, GNSS can not determine the orientation of the device and the heading accuracy is usually set to +-180 deg, i.e. the heading is unknown. In such cases, the heading should not be used. This PR modifies the implementation so, that when heading is considered too inaccurate, the heading is not encoded into the data sent to the nRF Cloud. Another option would be to send both heading and heading accuracy values to the cloud. 

I did not do a similar modification for `json_common.c`, it could be added as well.

The PR also adds accuracy values for altitude and speed in the data structures. Those are not used at the moment, but we could have similar functionality at least for the speed.

trantanen:
I don't think I have any new comments that haven't been mentioned. You should add similar changes to `json_common.c` and update unit tests.

applications/asset_tracker_v2/src/cloud/cloud_codec/cloud_codec.h
```diff
@@ -38,6 +38,11 @@
 extern "C" {
 #endif
 
+/** Heading accuracy limit for the heading to be considered valid. Heading is only sent to the
+ *  cloud if it's accurate enough.
```
peknis, line 42:
```suggestion
 *  cloud if it is accurate enough.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -143,6 +143,7 @@ Asset Tracker v2
 * Updated:
 
   * The MQTT topic name for A-GNSS requests is changed to ``agnss`` for AWS and Azure backends.
+  * GNSS heading is only sent to the cloud when it's considered accurate enough.
```
peknis, line 146:
```suggestion
  * GNSS heading is only sent to the cloud when it is considered accurate enough.
```

peknis (changes requested):

tokangas:
@jtguggedal @trantanen I've now done all requested changes.

peknis (approved):

trantanen (approved):

PR #14854, opened 2024-04-12
nrf_security: Add experimental legacy and PSA crypto option

Vge0rge:
Adds option to support legacy and PSA crypto at the same time.
This feature is experimental and not well tested.


krish2718 (approved):
Do we have a final plan to handle this? I.e., When will this experimental symbol be removed?

krish2718:
> Do we have a final plan to handle this? I.e., When will this experimental symbol be removed?

Wi-Fi already has `WPA_SUPP_CRYPTO_PSA` but it's default only for TF-M  see https://github.com/nrfconnect/sdk-nrf/blob/main/modules/hostap/Kconfig#L91, If it helps we can make this as default and drop the Legacy MbedTLS support.

subsys/nrf_security/Kconfig
```diff
@@ -50,6 +50,36 @@ config NRF_SECURITY
...
+	# the trusted storage for ITS support. The depenedencies here
+	# match what we enable in the build_config.h file so if we need to
+	# modify the dependencies here we also need to modify the build_config.h.
+	depends on PSA_CRYPTO_DRIVER_OBERON && !PSA_CRYPTO_DRIVER_CC3XX
+	depends on NRF_CC3XX_PLATFORM
+	depends on TRUSTED_STORAGE
+	depends on !BUILD_WITH_TFM
+	help
+	   This is an option to support legacy mbedTLS and PSA crypto APIs
+	   at the same time. This is not recommended and it is not fully
```
frkv, line 73:
```suggestion
	   at the same time. This is not recommended as it is not fully
```

krish2718:
> Dropping legacy support is a step to the right direction and it helps us to focus on PSA support more. So if you can do it that's for sure beneficial to us.

We already have `WPA_SUPP_CRYPTO_PSA` (Uses PSA options, but still doesn't use PSA APIs), so, will making it default will render this PR moot? Or do we need this PR either way?

subsys/nrf_security/src/drivers/nrf_oberon/CMakeLists.txt
```diff
@@ -66,4 +66,43 @@ if (COMPILE_PSA_APIS)
...
+
+  # Turn off warnings that Oberon are systematically
+  # triggering. Oberon is testing this deliverable before we get it so
+  # we don't need to re-test it with gcc warnings.
+  target_compile_options(
+	${mbedcrypto_target}
+	PRIVATE
+	-Wno-uninitialized
+	-Wno-maybe-uninitialized
+  )
```
mswarowsky, line 106: This could be outside of the if/elseif block as we do the same no what branch taken,

mswarowsky (approved):
A small NIT, but no blocker.

Lets's hope feature doesn't create us too much headache...

subsys/nrf_security/Kconfig
```diff
@@ -50,6 +50,36 @@ config NRF_SECURITY
 	  Set this configuration to enable nRF Security. This provides
 	  Arm PSA cryptography APIs with RNG support (optionally).
 
+config NRF_SECURITY_LEGACY_AND_PSA
+	bool
+	default y
+	select EXPERIMENTAL
```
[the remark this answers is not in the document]
Vge0rge: I would still keep it if it is not a big issue since we don't fully perform testing for this I think.

PR #14885, opened 2024-04-16
tfm: Make TF-M support Cracen key solution

vili-nordic:
Updates TF-M and related configuration to support the Cracen design where CMAC KDF is used for key derivation with IKG generated keys.


include/tfm/tfm_builtin_key_ids.h
```diff
@@ -7,6 +7,25 @@
...
+
+#define CRACEN_BUILTIN_IDENTITY_KEY_ID ((uint32_t)0x7fffc001)
+#define CRACEN_BUILTIN_MKEK_ID	       ((uint32_t)0x7fffc002)
+#define CRACEN_BUILTIN_MEXT_ID	       ((uint32_t)0x7fffc003)
+
+#define CRACEN_IDENTITY_KEY_SLOT_NUMBER 0
+#define CRACEN_MKEK_SLOT_NUMBER		1
+#define CRACEN_MEXT_SLOT_NUMBER		2
+
+#define PSA_KEY_LOCATION_CRACEN (0x800000 | ('N' << 8))
```
SebastianBoe, line 20: Is it intentional to not cast to psa_key_location_t like the non-CRACEN driver does in the else clause below?

include/tfm/tfm_builtin_key_ids.h
```diff
@@ -7,6 +7,25 @@
 #ifndef __TFM_BUILTIN_KEY_IDS_H__
 #define __TFM_BUILTIN_KEY_IDS_H__
 
+#ifdef CONFIG_PSA_CRYPTO_DRIVER_CRACEN
+
+#define CRACEN_BUILTIN_IDENTITY_KEY_ID ((uint32_t)0x7fffc001)
+#define CRACEN_BUILTIN_MKEK_ID	       ((uint32_t)0x7fffc002)
+#define CRACEN_BUILTIN_MEXT_ID	       ((uint32_t)0x7fffc003)
```
SebastianBoe, line 14: I see these are already defined by cracen_psa.h.

Why are we redefining them here instead of #including them from there?

include/tfm/tfm_builtin_key_ids.h
vili-nordic: Not really, I'll add it when I do the fix mentioned in the above "issue"

modules/trusted-firmware-m/Kconfig
```diff
@@ -64,6 +64,7 @@ config TFM_CRYPTO_BUILTIN_KEYS
 	prompt "TF-M crypto builtin keys"
 	default y
 	depends on !TFM_PROFILE_TYPE_MINIMAL
+	depends on !CRACEN_HW_PRESENT
```
SebastianBoe, line 67: So we don't support TFM builtin keys with CRACEN, but we are modifying tfm_builtin_key_ids.h to support CRACEN?

What's the deal? :)

modules/trusted-firmware-m/tfm_boards/nrf54l15_cpuapp/config.cmake
```diff
@@ -11,3 +11,6 @@ include(${PLATFORM_PATH}/common/core/config.cmake)
 set(NRF_SOC_VARIANT nrf54l15 CACHE STRING "nRF SoC Variant")
 
 include(${OOT_54L_PLATFORM_PATH}/config.cmake)
+
+# Override PS_CRYPTO_KDF_ALG
+set(PS_CRYPTO_AEAD_ALG                  PSA_ALG_SP800_108_COUNTER_CMAC CACHE STRING    "KDF Algorithm to use")
```
SebastianBoe, line 16: Why does the comment say PS_CRYPTO_KDF_ALG and the code say PS_CRYPTO_AEAD_ALG?

samples/keys/hw_unique_key/src/derive_key_tfm.c
```diff
@@ -26,7 +26,12 @@ psa_status_t derive_key(psa_key_attributes_t *attributes, uint8_t *key_label,
 
 	*key_id_out = PSA_KEY_ID_NULL;
 
+#if CONFIG_CRACEN_HW_PRESENT
+	/* On platforms with Cracen, the HUK is an AES key. */
+	status = psa_key_derivation_setup(&op, PSA_ALG_SP800_108_COUNTER_CMAC);
+#else
 	status = psa_key_derivation_setup(&op, PSA_ALG_HKDF(PSA_ALG_SHA_256));
+#endif
```
SebastianBoe, line 34: Let's move the ALG into a define at file level scope so we don't have an ugly ifdef in the code.

Same for the label.

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
```diff
@@ -1834,8 +1834,6 @@ config PSA_NEED_CRACEN_KEY_DERIVATION_DRIVER
 
 config PSA_NEED_CRACEN_KMU_DRIVER
 	bool
-	#TODO: Add support for KMU in TF-M builds: NCSDK-26633
-	default y if !BUILD_WITH_TFM
 	depends on PSA_NEED_CRACEN_GCM_AES && \
```
SebastianBoe, line 1837: Is it intentional to diverge from the other symbols here (which have default y)?

SebastianBoe (changes requested):

modules/trusted-firmware-m/tfm_boards/CMakeLists.txt
```diff
@@ -187,3 +195,9 @@ endif()
 
 file(COPY ${CMAKE_CURRENT_LIST_DIR}/common
      DESTINATION ${INSTALL_PLATFORM_NS_DIR})
+
+
+if((NRF_SOC_VARIANT STREQUAL nrf54l15) OR (target STREQUAL nrf54l15))
```
vili-nordic, line 200: This is probably not the best, I'm open to suggestions for how to best make this file available for NS.

SebastianBoe (approved):

SebastianBoe:
Nice work :)

modules/trusted-firmware-m/tfm_boards/CMakeLists.txt
```diff
@@ -87,17 +87,25 @@ if (${TFM_PARTITION_CRYPTO})
...
-        ${ZEPHYR_NRF_MODULE_DIR}/lib/hw_unique_key/hw_unique_key_kmu.c
-    )
-  endif()
+   if (${CONFIG_HW_UNIQUE_KEY})
+     target_compile_definitions(platform_s PUBLIC
+       CONFIG_HW_UNIQUE_KEY
+       $<$<BOOL:${CONFIG_HW_UNIQUE_KEY_RANDOM}>:CONFIG_HW_UNIQUE_KEY_RANDOM>)
+
+
+       if((NRF_SOC_VARIANT STREQUAL nrf54l15) OR (target STREQUAL nrf54l15))
```
mswarowsky, line 96: Can't we check here for CRACEN available so we don't have to keep a list platforms

subsys/nrf_security/src/drivers/cracen/cracenpsa/include/cracen_psa.h
```diff
@@ -14,6 +14,7 @@
 #include <cracen/mem_helpers.h>
 #include "cracen_psa_primitives.h"
 #include "cracen_psa_kmu.h"
+#include "cracen_psa_key_ids.h"
```
mswarowsky, line 17: I can't find this file? :(

vili-nordic: forgot to git add it :/

modules/trusted-firmware-m/tfm_boards/CMakeLists.txt
mswarowsky: then we leave it like that

mswarowsky (approved):

PR #14886, opened 2024-04-16
modules: tf-m: Fix 9161 building issue

Vge0rge:
The config.cmake for nRF9120 (used also for nRF9161) was not placed in the correct path for the non secure application. This mainly affected the TF-M regression tests since they use the non secure application building from TF-M.

Ref: NCSDK-26131


SebastianBoe (approved):

mswarowsky (approved):
good catch

PR #14894, opened 2024-04-16
Applications: nrf5340_audio: Oct 2983 statically store and use absolute volume

koffes:


applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -10,14 +10,22 @@
 #include <zephyr/types.h>
 #include <zephyr/bluetooth/conn.h>
 #include <zephyr/bluetooth/audio/vcp.h>
+#include <zephyr/bluetooth/gatt.h>
+#include <../../../../subsys/bluetooth/audio/vcp_internal.h>
```
Thalley, line 14: Suggest to add a comment why you are including the internal header file, and possibly add a reference to https://github.com/zephyrproject-rtos/zephyr/issues/71614

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -70,49 +97,151 @@ static bool vcs_client_peer_exists(struct bt_conn *conn)
...
+
+/**
+ * @brief Get the common volume of all devices
+ *
+ * @param[out] volume
+ * @return int	0 on success,
+ *		-EFAULT if no common volume was found.
+ *		This can happen if no adjustments have been made or no devices are connected.
+ */
+static int common_vol_get(uint8_t *volume)
```
Thalley, line 129: Not sure I follow this - What is a common volume? 

If two servers have e.g. 100 and 110 as their volume states, does this return `100` or an error?

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -70,49 +97,151 @@ static bool vcs_client_peer_exists(struct bt_conn *conn)
...
+ * @param vcs[in]		The Volume Control Server who initiated a volume change.
+ * @param volume_out[in]	The volume to set.
+ * @param include_caller[in]	If true, also send an update to the initiating VCS.
+ */
+static void update_volume_of_all_peers(struct bt_vcp_vol_ctlr const *const vcs, uint8_t volume_out,
+				       bool include_caller)
+{
+	int ret;
+
+	for (int i = 0; i < ARRAY_SIZE(vcs_client_peer); i++) {
```
Thalley, line 155: Consider using `size_t` instead of `int` when using `ARRAY_SIZE`

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -70,49 +97,151 @@ static bool vcs_client_peer_exists(struct bt_conn *conn)
...
+		return;
+	}
+
+	idx = vcs_peer_index_get(vol_ctlr);
+	if (idx < 0) {
+		LOG_ERR("VCS peer not found");
+		return;
+	}
+
+	LOG_DBG("Saving vol %d for dev %d", vol_ctlr->cp_val.volume, idx);
```
Thalley, line 193:
```suggestion
	LOG_DBG("Saving vol %u for dev %u", vol_ctlr->cp_val.volume, idx);
```

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -10,14 +10,22 @@
...
 #include "bt_rendering_and_capture.h"
 
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(bt_vol_ctlr, CONFIG_BT_VOL_LOG_LEVEL);
 
-static struct bt_vcp_vol_ctlr *vcs_client_peer[CONFIG_BT_MAX_CONN];
+struct vcs_peer {
+	struct bt_vcp_vol_ctlr *peer;
+	uint8_t master_vol;
+	bool default_val;
```
Thalley, line 25: Looking at the implementation, it appears that `default_val` indicates "unknown" value, because it's only `true` when you haven't stored a value, right?

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
Thalley, line 129: Consider having the same behavior here as you do for `vcs_peer_index_get`, i.e. negative value => error, positive value => return value (volume in this case)

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -177,13 +321,18 @@ int bt_vol_ctlr_down(void)
...
-					i, ret);
-			}
+
+		if (!vcs_peer_alive(vcs_client_peer[i].peer)) {
+			continue;
+		}
+
+		int vol = vcs_client_peer[i].master_vol - CONFIG_BT_AUDIO_VOL_STEP_SIZE;
+
+		vol = CLAMP(vol, 0, __UINT8_MAX__);
```
Thalley, line 331:
```suggestion
		vol = CLAMP(vol, 0, UINT8_MAX);
```

Minor comment, but you should use the non-underscore versions

applications/nrf5340_audio/unicast_client/main.c
```diff
@@ -326,7 +326,7 @@ static void bt_mgmt_evt_handler(const struct zbus_channel *chan)
 
 		ret = bt_r_and_c_discover(msg->conn);
 		if (ret) {
-			LOG_WRN("Failed to discover rendering services");
+			LOG_WRN("Failed to discover rendering and capture services");
```
Thalley, line 329: Unrelated to the changes in this PR, but consider moving the capture and rendering control out of the module called `unicast_client`. In CAP the unicast client is part of the CAP initiator and does not handle volume control (or microphone control), but that's the behavior of the CAP commander, which does not do stream control (and thus not a unicast client).

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
koffes: With this system that will never happen. 
After discovery, the first server will have 100, and default = true. Then, this will get 100 and default = false.
When the second server is discovered, that will have 110 and default=true. This one will then be updated to 100 and default = false.

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
koffes: You are right. I have renamed.

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -70,49 +98,149 @@ static bool vcs_client_peer_exists(struct bt_conn *conn)
...
+{
+	for (size_t i = 0; i < ARRAY_SIZE(vcs_client_peer); i++) {
+		if (!vcs_peer_alive(vcs)) {
+			/* Skip */
+			continue;
+		}
+
+		if (vcs == vcs_client_peer[i].peer) {
+			return i;
+		}
```
Thalley, line 116: If you clear the `vcs_client_peer[i].peer` on disconnects, you could skip the `vcs_peer_alive` check and simply do `vcs_peer_alive[i].peer == NULL`

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -70,49 +98,149 @@ static bool vcs_client_peer_exists(struct bt_conn *conn)
...
+		common_vol = ret;
+
+		if (common_vol == dev_volume) {
+			/* Volume of this device is the same as the common volume. No update */
+			return;
 		}
+
+		LOG_DBG("Stored vol found. Writing vol %u to remote device %u", common_vol,
+			current_idx);
+		update_volume_of_all_peers(vcs, common_vol, true);
```
Thalley, line 235: How well does this handle situations where multiple servers change volume at the same time? e.g. if you are connected to 2 VCS devices and they both change their volume to different values:

1) Server 1 and Server 2 have volume 100
2) Server 1 locally changes volume to 110 and sends notification, which triggers `update_volume_of_all_peers` to update volume to 110 for Server 2
3) Server 2 locally changes volume to 90 and sends notification before getting the volume request sent in 2), which triggers `update_volume_of_all_peers` to update volume to 90 for Server 1
4) Server 2 now gets volume request for 110, and sends notification for 110, but now you are setting volume to 90 for Server 1, and that triggers yet another call to `update_volume_of_all_peers`

In the above case, won't you effectively end up in an infinitely loop of setting volume between 90 and 110 because they will never really sync?

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -158,14 +296,18 @@ int bt_vol_ctlr_up(void)
...
+	for (size_t i = 0; i < ARRAY_SIZE(vcs_client_peer); i++) {
+		if (!vcs_peer_alive(vcs_client_peer[i].peer)) {
+			continue;
+		}
+
+		int vol = vcs_client_peer[i].master_vol + CONFIG_BT_AUDIO_VOL_STEP_SIZE;
+
+		vol = CLAMP(vol, 0, UINT8_MAX);
+		LOG_DBG("bt_vol_ctlr_up for index %u vol: %u", i, vol);
+		ret = bt_vcp_vol_ctlr_set_vol(vcs_client_peer[i].peer, vol);
```
Thalley, line 308: If `vol == vcs_client_peer[i].master_vol` we don't need to set the volume, right? 

Ditto for the volume down and mute

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -70,49 +98,149 @@ static bool vcs_client_peer_exists(struct bt_conn *conn)
...
+		update_volume_of_all_peers(vcs, dev_volume, false);
 	}
 }
 
-/**
- * @brief	Callback handler for the VCS controller flags.
- *
- * @note	This callback handler will be triggered if VCS flags changed.
- */
-static void vcs_flags_ctlr_cb_handler(struct bt_vcp_vol_ctlr *vcs, int err, uint8_t flags)
```
alexsven, line 115: Why was the documentation removed?

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -121,33 +249,43 @@ static void vcs_flags_ctlr_cb_handler(struct bt_vcp_vol_ctlr *vcs, int err, uint
 	}
 }
 
-/**
- * @brief	Callback handler for the finished VCS discovery.
- *
- * @note	This callback handler will be triggered when the VCS discovery has finished.
- */
```
alexsven, line 128: Why was the documentation removed?

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -121,33 +249,43 @@ static void vcs_flags_ctlr_cb_handler(struct bt_vcp_vol_ctlr *vcs, int err, uint
...
+	int ret;
+
 	if (err) {
-		LOG_WRN("VCS discover finished callback error: %d", err);
-	} else {
-		LOG_INF("VCS discover finished");
+		LOG_ERR("VCS discover finished callback error: %d", err);
+		return;
+	}
+
```
alexsven, line 261:
```suggestion
LOG_INF("VCS discover finished");
```

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
```diff
@@ -121,33 +249,43 @@ static void vcs_flags_ctlr_cb_handler(struct bt_vcp_vol_ctlr *vcs, int err, uint
...
+	/* If a renderer is re-discoverered, reset parameters */
+	for (size_t i = 0; i < ARRAY_SIZE(vcs_client_peer); i++) {
+		if (vcs == vcs_client_peer[i].peer) {
```
alexsven, line 264: Could use `vcs_peer_index_get(vcs)`

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
alexsven: Isn't the mute a separate parameter? So the vol has nothing to do with mute

alexsven (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
Thalley: w.r.t mute and unmute, if a device is already muted, then there's no point in sending the "mute" opcode to the server, and similar to unmute. 

Basically my comment referred to that if we know that the command will have no state change, then we don't need to send the command.

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
koffes: These are static functions and their documentation should reside where this callback is defined in the host.

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
koffes: These are static functions and their documentation should reside where this callback is defined in the host.

applications/nrf5340_audio/src/bluetooth/bt_rendering_and_capture/volume/bt_vol_ctlr.c
koffes: Interesting observation. When tested, I get -EBUSY from `bt_vcp_vol_ctlr_set_vol`

Thalley: > Interesting observation. When tested, I get -EBUSY from `bt_vcp_vol_ctlr_set_vol`

Likely because the previous command has not yet finished for that connection. So depending on timing it may work, it may run in the above loop or it may return an error. To me that sounds like 1/3 cases are handled (although the remaining 2 unhandled cases are rare).

PR #14898, opened 2024-04-17
samples: matter: Align Matter samples to FILE_SUFFIX

ArekBalysNordic:
- Removed all _release.yaml pm_static files because the release configuration has been the same as the debug one.
- Removed all redundant _release overlays from mcuboot child image configuration, because they were the same as for debug build type.
- Changed CONF_FILE to FILE_SUFFIX in sample.yaml files.


ArekBalysNordic:
Closing because there is no support for child images for now.

PR #14903, opened 2024-04-17
samples: wifi: shell: Increase shell stack size

rado17:
Stack size was increased in 77cad230008b55a950a06f8f519401364dae228f. But connection attempts, when done from `conn_mgr` module, add a couple of levels of depth which causes the stack usage to exceed the allocated space. Increasing the stack size by 200 bytes takes care of this.

Fixes SHEL-2599.


kapbh (approved):

krish2718:
Can you please cherry pick the fix from https://github.com/nrfconnect/sdk-nrf/pull/14902 ? We should revisit all threads and usecases that can potentially call `wpa_cli*` and they need to be increased.

VivekUppunda (approved):

krish2718 (approved):

PR #14908, opened 2024-04-17
emds: exclude write time measurement for nrf54

alxelax:
Commit excludes writing time measurement for nrf54l. There are no still RRAM programming digits in documentation.


katgiadla (approved):
It works on-target

ludvigsj (approved):

PR #14921, opened 2024-04-18
driver: wifi: Make WPA supplicant regulatory aware

ajayparida:
[SHEL-2661] If regulatory set is called before WPA supplicant is initialized, then as NM is not registered the calls gets re-routed to driver directly skipping WPA supplicant. In that case, the command should be failed at driver.

[SHEL-2661]: https://nordicsemi.atlassian.net/browse/SHEL-2661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/drivers/wifi/nrf700x/src/fmac_main.c
+++ b/drivers/wifi/nrf700x/src/fmac_main.c
@@ -27,6 +27,7 @@
 
 #ifndef CONFIG_NRF700X_RADIO_TEST
 #ifdef CONFIG_NRF700X_STA_MODE
+#include <zephyr/net/wifi_nm.h>
 #include <wifi_mgmt_scan.h>
 #include <wifi_mgmt.h>
 #include <wpa_supp_if.h>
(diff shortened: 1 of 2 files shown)


sachinthegreen (approved):

drivers/wifi/nrf700x/src/fmac_main.c
```diff
@@ -341,6 +341,27 @@ int nrf_wifi_reg_domain(const struct device *dev, struct wifi_reg_domain *reg_do
 	}
 
 	if (reg_domain->oper == WIFI_MGMT_SET) {
+#ifndef CONFIG_NRF700X_RADIO_TEST
+#ifdef CONFIG_NRF700X_STA_MODE
+		/* Need to check if WPA supplicant is initialized or not.
+		 * Must be cheecked when CONFIG_WPA_SUPP is enabled.
```
krish2718, line 347:
```suggestion
		 * Must be checked when CONFIG_WPA_SUPP is enabled.
```

drivers/wifi/nrf700x/src/fmac_main.c
```diff
@@ -341,6 +341,27 @@ int nrf_wifi_reg_domain(const struct device *dev, struct wifi_reg_domain *reg_do
...
+		 * Not applicable for RADIO_TEST or when CONFIG_WPA_SUPP is not enabled.
+		 */
+		/* It is possbile that during supplicant initialization driver may
+		 * get the command. lock will try to ensure that supplicant
+		 * initialization is complete.
+		 */
+		k_mutex_lock(&vif_ctx_zep->vif_lock, K_FOREVER);
+
+		if (!vif_ctx_zep->supp_drv_if_ctx) {
```
krish2718, line 356: This check won't be enough, as this is set in the `wpa_drv_init` but the NM iface is registered after `add_interface`.

krish2718: You can use `wifi_nm_get_instance_iface` to see if the iface is managed by WPA supplicant or not. But those APIs don't have lock, so, might have to add one.

krish2718 (changes requested):

drivers/wifi/nrf700x/src/fmac_main.c
ajayparida: Added required checks.

PR #14708, opened 2024-04-03
manifest: Update zephyr with SPI tests for nrf54h and nrf54l

nordic-krch:
Update zephyr with support for nrf54h and nrf54l in spi_loopback. Update zephyr with spi_slave test supported on nrf54h and nrf54l.

PR #14929, opened 2024-04-18
manifest: Update zephyr with nrf54l15 overlay for counter

nordic-krch:
Update zephyr with nrf54l15pdk overlay for counter_basic_test.


nordic-krch:
Changes merged to `sdk-zephyr`. They will be fetched with #14708.

PR #14931, opened 2024-04-18
RFC: Bluetooth: Host: introduce priority sets

Tschet1:
In order to better adapt to the needs of the application, add a host API that allows to change the priorities in the controller. These configuration sets are tested internally and have a specific purpose.

Initially, add a configuration which gives more priority to the initiator, in order to improve connection setup latency.

### DIFF
--- a/subsys/bluetooth/host_extensions/hci_types_host_extensions.h
+++ b/subsys/bluetooth/host_extensions/hci_types_host_extensions.h
@@ -39,6 +39,14 @@ struct bt_hci_cp_set_power_control_request_param {
 	uint8_t apr_margin;
 } __packed;
 
+#define BT_HCI_OP_SET_ROLE_PRIORITY	BT_OP(BT_OGF_VS, 0x011B)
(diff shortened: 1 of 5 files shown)


subsys/bluetooth/host_extensions/hci_types_host_extensions.h
```diff
@@ -39,6 +39,14 @@ struct bt_hci_cp_set_power_control_request_param {
 	uint8_t apr_margin;
 } __packed;
 
+#define BT_HCI_OP_SET_ROLE_PRIORITY	BT_OP(BT_OGF_VS, 0x011B)
+
+struct bt_hci_set_role_priority {
```
Tschet1, line 44: This can probably be changed, such that the controller API allows setting multiple priorities at the same time. Right now, if the priority sets grow, then there would be a separate command for each priority in the set.

Tschet1:
We will go with a simpler approach for this initially

PR #14934, opened 2024-04-18
doc: SML - fix nrf54 columns

FrancescoSer:
Edited SML py script to fix nrf54 related columns.


totyz (approved):

doc/_scripts/software_maturity/software_maturity_scanner.py
```diff
@@ -295,7 +295,7 @@ def find_all_socs(all_samples: List[str]) -> List[str]:
 
     # Only use nRF boards, but not nRF51 series
     all_boards = set(filter(BOARD_FILTER, all_boards))
-    all_socs = {re.findall(r"_(nrf[0-9]+)", b)[-1] for b in all_boards}
+    all_socs = {re.findall(r"_(/nrf[0-9]+[hl]?[0-9]*/i)", b)[-1] for b in all_boards}
```
totyz, line 298:
```suggestion
    all_socs = {re.findall(r"_(nrf[0-9]+[hlHL]?[0-9]+)", b)[-1] for b in all_boards}
```

totyz: This regexp will work as expected:
image

totyz (changes requested):

doc/_scripts/software_maturity/software_maturity_scanner.py
totyz: generated yaml indicated that if needed, it must be done in different place.
image
So, I would suggest, to apply the change and we will see how it works

totyz: I found in https://github.com/nrfconnect/sdk-nrf/blob/main/doc/_extensions/software_maturity_table.py 
```
        for soc in all_socs:
            soc = soc.upper().replace("NRF", "nRF")
            entry = nodes.entry()
            entry += nodes.paragraph(text=soc)
            row += entry
        thead.append(row)
```
So, in documentation, the names should look OK

totyz:
The change introduced in this PR is not visible in the documentation built for the PR. This is because the `maturity-samples-report.yaml` file, generated by the `software_maturity_scanner.py` script, is taken from the last sdk-nrf nightly CI run, not from this PR. 

Therefore, I would suggest implementing this PR, as I'm confident it produces a correct and valid `maturity-samples-report.yaml` file.

totyz (approved):

gmarull (approved):

PR #14940, opened 2024-04-18
Allow to disable Partition Manager for a project

de-nordic:
Commits that allow disabling Partition Manager, and multi-image, build for any project that is not built with sysbuild.

For example issuing:
```
 west build -d builds/hello_world -b nrf52840dk_nrf52840 zephyr/samples/hello_world/ -DCONFIG_BOOTLOADER_MCUBOOT=y -DCONFIG_PRE_SELECT_PARTITION_MANAGER=n
```
will build Hello World sample with sdk-nrf, with MCUboot support but with multi-image disabled, and only the Hello worl will be built.
Then you can built the MCUboot without partition manager using:
```
west build -d builds/mcuboot -b nrf52840dk_nrf52840 bootloader/mcuboot/boot/zephyr/ -DCONFIG_PRE_SELECT_PARTITION_MANAGER=n
```

### DIFF
--- a/subsys/partition_manager/Kconfig
+++ b/subsys/partition_manager/Kconfig
@@ -6,8 +6,22 @@
 
 menu "Partition Manager"
 
+config PRE_SELECT_PARTITION_MANAGER
+	bool "Pre-select Partition Manager for project [DO NOT EDIT]"
+	default y
+	help
+	  Use partition manager for specified project.
+	  This should be set from command line before project is built
+	  for the first time, otherwise the Partition Manager and
+	  multi-image support may be already included for a project.
+	  When set to n it directs all partition references to DTS and
+	  disables the multi-image support, so what you will get is
+	  build of the specified project only with partition definitions
+	  taken from the DTS.
+
 config PARTITION_MANAGER_ENABLED
 	bool "Partition manager is enabled (read-only option)"
+	default y if PRE_SELECT_PARTITION_MANAGER
 	help
 	  Option which can be checked to see whether or not the
 	  partition manager is enabled in the current build. Note that this
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,9 @@ endforeach()
 
 include(cmake/extensions.cmake)
 include(cmake/version.cmake)
-include(cmake/multi_image.cmake)
+if (NOT SYSBUILD OR CONFIG_PRE_SELECT_PARTITION_MANAGER)
+  include(cmake/multi_image.cmake)
+endif()
 
 zephyr_include_directories(include)
 
--- a/modules/mcuboot/CMakeLists.txt
+++ b/modules/mcuboot/CMakeLists.txt
@@ -2,9 +2,9 @@ add_subdirectory(${ZEPHYR_MCUBOOT_MODULE_DIR}/boot/bootutil/zephyr
                  ${CMAKE_CURRENT_BINARY_DIR}/boot/bootutil/zephyr
 )
 
-if(SYSBUILD)
+if(SYSBUILD OR NOT CONFIG_PRE_SELECT_PARTITION_MANAGER)
   # Code below is partition manager w/ child image support, thus return when
-  # sysbuild is used.
+  # sysbuild is used or Partition Manager has been disabled.
   return()
 endif()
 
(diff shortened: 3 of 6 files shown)


subsys/partition_manager/Kconfig
```diff
@@ -6,8 +6,22 @@
 
 menu "Partition Manager"
 
+config PARTITION_MANAGER
```
nvlsianpu, line 9: misleading name: maybe `STATR_PROJECT_WITH_PARTITION_MANAGER`

subsys/partition_manager/Kconfig
```diff
@@ -6,8 +6,22 @@
...
+	bool "Partition manager"
```
nvlsianpu, line 10: "Partition Manager pre-selected [DO NOT EDIT]"

subsys/partition_manager/Kconfig
nvlsianpu: or `PRE_SELECT_PARTITION_MANAGER`

nvlsianpu (changes requested):
Can you clarify the PR description: Is this patch for building multiple images using sysbuild (like the app + mcuboot)?

subsys/partition_manager/Kconfig
```diff
@@ -6,8 +6,22 @@
...
 
+config PRE_SELECT_PARTITION_MANAGER
+	bool "Pre-select Partition Manager for project [DO NOT EDIT]"
+	default y
+	help
+	  Use partition manager for specified project.
+	  This should be set from command line before project is built
+	  for the first time, otherwise the Partition Manager and
+	  multi-image support may be already included for a project.
+	  When set to n it directs all partition references to DTS and
```
michalek-no, line 17:
```suggestion
	  When set to 'n' it directs all partition references to DTS and
```

nvlsianpu (approved):

greg-fer (changes requested):
@nvlsianpu , what about documentation update for this? I think doc updates are needed under both DFU and Partition Manager doc sections (@kittydepa and @FrancescoSer ).

nordicjm (commented):
I don't see the purpose in this, you should just use `-DCONFIG_PARTITION_MANAGER_ENABLED=n` and that should disable partition manager, no additional symbol is needed

greg-fer (commented):
@bencefr , @kylebonnici , FYI, this might impact the VSC extension IDE.

nordicjm (changes requested):
and this breaks sysbuild support

tejlmand (changes requested):
Looks very wrong to me.

Can you explain why `-DSB_CONFIG_PARTITION_MANAGER=n` is not enough to disable partition manager when using sysbuild ?

nordicjm:
> > Looks very wrong to me.
> > Can you explain why `-DSB_CONFIG_PARTITION_MANAGER=n` is not enough to disable partition manager when using sysbuild ?
> 
> This disables partition manager when sysbuild is not enabled. Goal is to be able to build anything, as single project, even if CONFIG_BOOTLOADER_MCUBOOT=y; currently if you want to build something for MCUboot the Partition Manager and Multi-imaga kicks in, because cmakes have been modified, for sdk-nrf, to always do so, without override.

Use sysbuild where this feature is natively supported, child/parent image is deprecated, when this upmerge has finished all support for it in applications/samples/tests will be removed and it will remain only to prevent breaking support for customer applications, and will be fully removed in a couple of releases time

nvlsianpu:
@de-nordic I think building stuff (MCUboot + app + ...) using sysbuild & -DSB_CONFIG_PARTITION_MANAGER=n is what we want as target state.

de-nordic:
> @de-nordic I think building stuff (MCUboot + app + ...) using sysbuild & -DSB_CONFIG_PARTITION_MANAGER=n is what we want as target state.

Ok, so this PR makes no sense.

What about user cases where users only want to build application? Customers were asking why they have to build MCUboot while rebuilding the application.

nordicjm:
> > @de-nordic I think building stuff (MCUboot + app + ...) using sysbuild & -DSB_CONFIG_PARTITION_MANAGER=n is what we want as target state.
> 
> Ok, so this PR makes no sense.
> 
> What about user cases where users only want to build application? Customers were asking why they have to build MCUboot while rebuilding the application.

Not supported, if you mean building application which is booted from mcuboot, as the mcuboot parts of the static PM file would be dropped (this is how it's always been in PM). If you just mean building the application without mcuboot and without mcuboot support, that's fine, just do not enable mcuboot in sysbuild

de-nordic:
Seems to be not needed in the end.

PR #14944, opened 2024-04-19
nrf_security: sicrypto: Harden the workqueue

SebastianBoe:
Harden the workqueue against certain types of crashes and attacks.

### DIFF
--- a/subsys/nrf_security/src/drivers/cracen/sicrypto/src/waitqueue.h
+++ b/subsys/nrf_security/src/drivers/cracen/sicrypto/src/waitqueue.h
@@ -8,7 +8,25 @@
 #ifndef WAITQUEUE_HEADER_FILE
 #define WAITQUEUE_HEADER_FILE
 
+#include <nrf.h>
+#include <zephyr/sys/__assert.h>
+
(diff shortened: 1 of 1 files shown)


frkv (approved):
LGTM

subsys/nrf_security/src/drivers/cracen/sicrypto/src/waitqueue.h
```diff
@@ -32,6 +50,8 @@ static inline void si_wq_wake_next(struct sitask *t)
 	struct siwq *w = t->wq;
 
 	while ((w != NULL) && (t->statuscode != SX_ERR_HW_PROCESSING)) {
+		validate_function_pointer(w->run);
```
vili-nordic, line 53: This does not do anything - asserts are not included in release builds and calll on line 56 will anyways trap because MPU will trap on XN bit

vili-nordic (changes requested):

subsys/nrf_security/src/drivers/cracen/sicrypto/src/waitqueue.h
SebastianBoe: My mistake, good catch.

Should have tested this.

PR #14971, opened 2024-04-19
board: thingy91x_nrf9151: Add support for expansion board

maxd-nordic:
This patch adds support for using the expansion board. When using the expansion board, the corresponding power-enable pin must be asserted.
The most convenient way to do this is the HOG feature.


balaji-nordic (approved):

bjda (approved):

cvinayak:
will be good to get a review from @anangl  ?

PR #14981, opened 2024-04-22
snippets: nrf70-debug: Update shell stack size

krish2718:
Shell stack size is now increased, but forgot to update the debug snippet.

Fixes SHEL-2704


sachinthegreen (approved):

samples/wifi/shell/overlay-debug.conf
```diff
@@ -1,5 +1,5 @@
 CONFIG_SHELL=y
-CONFIG_SHELL_STACK_SIZE=4096
+CONFIG_SHELL_STACK_SIZE=5200
```
rado17, line 2: I think increasing shell stack size to 4400 should be enough (ref: https://github.com/nrfconnect/sdk-nrf/pull/14903/files).

krish2718: No, this isn't related to `conn_mgr`, no 2.5 the stack depth is much more, in latest we had done some optimisations for stack in WPA supplicant.

rado17 (approved):

mariusz-nordicsemi (approved):

PR #14982, opened 2024-04-22
NCSDK-26629: Enable SUIT unit tests

tomchy:
Ref: NCSDK-26629


adsz-nordic (approved):

ahasztag (approved):

rlubos:
@nordicjm PTAL at the testcase files.

nordicjm (changes requested):
https://github.com/nrfconnect/sdk-nrf/blob/933a880aa58c943ad84eb6a9b6c98789228d90a0/tests/subsys/suit/unit/app_specific/cmake/test_template.cmake needs updating to use proper variables like `${ZEPHYR_ZCBOR_MODULE_DIR}` i.e. that variable is already defined by the build system, remove it, then use the proper module names for the suit location instead of basing it off ZEPHYR_BASE and same with ncs

nordicjm (approved):

PR #14988, opened 2024-04-22
applications: Refactored Matter bridge storage implementation

kkasperczyk-no:
Refactored Matter bridge storage implementation to use different data structure with a fewer keys what allowed to use ~2.5 times less memory. Additonally, provided a mechanism that detects currently used structure version and migrates data from the older structure to ensure backward-compatibility.

### DIFF
--- a/applications/matter_bridge/src/app_task.cpp
+++ b/applications/matter_bridge/src/app_task.cpp
@@ -106,36 +106,26 @@ CHIP_ERROR AppTask::RestoreBridgedDevices()
 
 	/* Load all devices based on the read count number. */
 	for (size_t i = 0; i < indexesCount; i++) {
-		uint16_t endpointId;
-		char label[Nrf::MatterBridgedDevice::kNodeLabelSize] = { 0 };
-		size_t labelSize;
-		uint16_t deviceType;
+		Nrf::BridgeStorageManager::BridgedDevice device;
+#ifdef CONFIG_BRIDGED_DEVICE_BT
+		bt_addr_le_t btAddr;
+		device.mUserData = reinterpret_cast<uint8_t *>(&btAddr);
+		device.mUserDataSize = sizeof(btAddr);
+#endif
 
-		if (!Nrf::BridgeStorageManager::Instance().LoadBridgedDeviceEndpointId(endpointId, indexes[i])) {
+		if (!Nrf::BridgeStorageManager::Instance().LoadBridgedDevice(device, indexes[i])) {
 			return CHIP_ERROR_NOT_FOUND;
 		}
 
-		/* Ignore an error, as node label is optional, so it may not be found. */
-		Nrf::BridgeStorageManager::Instance().LoadBridgedDeviceNodeLabel(label, sizeof(label), labelSize,
-										 indexes[i]);
-
-		if (!Nrf::BridgeStorageManager::Instance().LoadBridgedDeviceType(deviceType, indexes[i])) {
-			return CHIP_ERROR_NOT_FOUND;
-		}
-
-		LOG_INF("Loaded bridged device on endpoint id %d from the storage", endpointId);
+		LOG_INF("Loaded bridged device on endpoint id %d from the storage", device.mEndpointId);
 
 #ifdef CONFIG_BRIDGED_DEVICE_BT
-		bt_addr_le_t addr;
-
-		if (!Nrf::BridgeStorageManager::Instance().LoadBtAddress(addr, indexes[i])) {
-			return CHIP_ERROR_NOT_FOUND;
-		}
-
-		BleBridgedDeviceFactory::CreateDevice(deviceType, addr, label, indexes[i], endpointId);
+		BleBridgedDeviceFactory::CreateDevice(device.mDeviceType, btAddr, device.mNodeLabel, indexes[i],
+						      device.mEndpointId);
 #else
-		SimulatedBridgedDeviceFactory::CreateDevice(deviceType, label, chip::Optional<uint8_t>(indexes[i]),
-							    chip::Optional<uint16_t>(endpointId));
+		SimulatedBridgedDeviceFactory::CreateDevice(device.mDeviceType, device.mNodeLabel,
+							    chip::Optional<uint8_t>(indexes[i]),
+							    chip::Optional<uint16_t>(device.mEndpointId));
 #endif
 	}
 	return CHIP_NO_ERROR;
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -224,6 +224,10 @@ Matter Bridge
    The :kconfig:option:`CONFIG_BRIDGE_BT_SCAN_TIMEOUT_MS` kconfig option to set the scan timeout.
 
 * Updated the implementation of the persistent storage to leverage ``NonSecure``-prefixed methods from the common Persistent Storage module.
+* Changed data structure of information stored in the persistent storage to use less settings keys.
+  The new structure uses approximately 40% of the memory used by the old structure, and provides a new field to store user-specific data.
+
+  Backward compatibility is kept by using an internal dedicated method that automatically detects the older data format and performs data migration to the new representation.
 
 IPC radio firmware
 ------------------
--- a/samples/matter/common/src/bridge/bridge_storage_manager.cpp
+++ b/samples/matter/common/src/bridge/bridge_storage_manager.cpp
@@ -5,6 +5,11 @@
  */
 
 #include "bridge_storage_manager.h"
+#include "bridge_manager.h"
+
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_DECLARE(app, CONFIG_CHIP_APP_LOG_LEVEL);
 
 namespace
 {
@@ -28,7 +33,90 @@ Nrf::PersistentStorageNode CreateIndexNode(uint8_t bridgedDeviceIndex, Nrf::Pers
 
 } /* namespace */
 
-namespace Nrf {
+namespace Nrf
+{
+
+bool BridgeStorageManager::Init()
+{
+	bool result = Nrf::GetPersistentStorage().NonSecureInit();
+
+	if (!result) {
+		return result;
+	}
+
+	/* Perform data migration from previous data structure versions if needed. */
+	return MigrateData();
+}
+
+bool BridgeStorageManager::MigrateData()
+{
+	uint8_t version;
+
+	/* Check if version key is present in settings to provide backward compatibility between post-2.6.0 releases and
+	 * the previous one. If the version key is present it means that the new keys structure is used (only version
+	 * equal to 1 is for now valid). Otherwise, the deprecated structure is used, so the migration has to be done.
+	 */
+	if (!LoadDataToObject(&mVersion, version)) {
+		uint8_t count;
+		uint8_t indexes[BridgeManager::kMaxBridgedDevices] = { 0 };
+		size_t indexesCount = 0;
+
+		if (LoadBridgedDevicesCount(count) &&
+		    LoadBridgedDevicesIndexes(indexes, BridgeManager::kMaxBridgedDevices, indexesCount)) {
+			/* Load all devices based on the read count number. */
+			for (size_t i = 0; i < indexesCount; i++) {
+				BridgedDevice device;
+				if (!LoadBridgedDeviceEndpointId(device.mEndpointId, i)) {
+					return false;
+				}
+
+				/* Ignore an error, as node label is optional, so it may not be found. */
+				if (!LoadBridgedDeviceNodeLabel(device.mNodeLabel, sizeof(device.mNodeLabel),
+								device.mNodeLabelLength, i)) {
+					device.mNodeLabelLength = 0;
+				}
+
+				if (!LoadBridgedDeviceType(device.mDeviceType, i)) {
+					return false;
+				}
+
+#ifdef CONFIG_BRIDGED_DEVICE_BT
+				bt_addr_le_t btAddr;
+
+				if (!LoadBtAddress(btAddr, i)) {
+					return false;
+				}
+
+				/* Insert Bluetooth LE address as a part of implementation specific user data. */
+				device.mUserDataSize = sizeof(btAddr);
+				device.mUserData = reinterpret_cast<uint8_t *>(&btAddr);
+#endif
+
+				/* Store all information using a new scheme. */
+				if (!StoreBridgedDevice(device, i)) {
+					return false;
+				}
+
+				/* Remove all information described using an old scheme. */
+				RemoveBridgedDeviceEndpointId(i);
+				RemoveBridgedDeviceNodeLabel(i);
+				RemoveBridgedDeviceType(i);
+
+#ifdef CONFIG_BRIDGED_DEVICE_BT
+				RemoveBtAddress(i);
+#endif
+			}
+		}
+
+		version = 1;
+		Nrf::GetPersistentStorage().NonSecureStore(&mVersion, &version, sizeof(version));
+
+	} else if (version != 1) {
+		/* Currently only no-version or version equal to 1 is supported. */
+		return false;
+	}
+	return true;
+}
 
 bool BridgeStorageManager::StoreBridgedDevicesCount(uint8_t count)
 {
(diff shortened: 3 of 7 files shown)


doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -224,6 +224,9 @@ Matter Bridge
    The :kconfig:option:`CONFIG_BRIDGE_BT_SCAN_TIMEOUT_MS` kconfig option to set the scan timeout.
 
 * Updated the implementation of the persistent storage to leverage ``NonSecure``-prefixed methods from the common Persistent Storage module.
+* Changed data structure of information stored in the persistent storage to use less settings keys.
+  It allowed to use approximately 2.5 times less memory and provided new field to store user specific data.
+  Backward compatibility is kept by a internal dedicated method that automatically detects older data format and performs data migration to the new representation.
```
wiba-nordic, line 229:
```suggestion
  The new structure uses approximately 2.5 times less memory, and provides new field to store user-specific data.
  
  Backward compatibility is kept by using an internal dedicated method that automatically detects the older data format and performs data migration to the new representation.
```

wiba-nordic (changes requested):

samples/matter/common/src/bridge/bridge_storage_manager.h
```diff
@@ -23,40 +23,32 @@ namespace Nrf
...
+ *			/0/ /<BridgedDevice>/
+ *			/1/ /<BridgedDevice>/
+ *			.
+ *			.
+ *			/n/ /<BridgedDevice>/
+ *		/ver/ <uint8_t>
  */
 class BridgeStorageManager {
 public:
+	static constexpr auto kMaxUserDataSize = 128;
```
Damian-Nordic, line 35:
```suggestion
	inline static constexpr auto kMaxUserDataSize = 128u;
```
Add `inline` if you're not defining that in .cpp file.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -28,7 +33,101 @@ Nrf::PersistentStorageNode CreateIndexNode(uint8_t bridgedDeviceIndex, Nrf::Pers
...
+	bool result = Nrf::GetPersistentStorage().NonSecureInit();
+
+	if (!result) {
+		return result;
+	}
+
+	/* Perform data migration from previous data structure versions if needed. */
+	result = MigrateData();
+
+	return result;
```
Damian-Nordic, line 50:
```suggestion
	return MigrateData();
```

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -28,7 +33,101 @@ Nrf::PersistentStorageNode CreateIndexNode(uint8_t bridgedDeviceIndex, Nrf::Pers
...
+				if (!LoadBtAddress(btAddr, i)) {
+					return false;
+				}
+
+				/* Insert Bluetooth LE address as a part of implementation specific user data. */
+				device.mUserDataSize = sizeof(btAddr);
+				device.mUserData = reinterpret_cast<uint8_t *>(&btAddr);
+#endif
+
+				/* Remove all information described using an old scheme. */
```
Damian-Nordic, line 97: 1. Isn't it better to first store the new data and than remove the old entries just in case the process is interrupted for some reason?
2. Also, in case you fail to load some old entries it may be worth continuing removing them instead of returning `false`. Overall, if load methods no longer support the old format, it may make the most sense to migrate as much as possible instead of failing the initialization on error.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -123,31 +200,85 @@ bool BridgeStorageManager::RemoveBridgedDeviceType(uint8_t bridgedDeviceIndex)
...
+	uint8_t buffer[sizeof(BridgedDevice) + kMaxUserDataSize];
+	uint16_t counter = 0;
+
+	if (!Nrf::PersistentStorage::Instance().Load(&id, buffer, sizeof(BridgedDevice) + kMaxUserDataSize, readSize)) {
 		return false;
 	}
 
-	if (!StoreBridgedDeviceEndpointId(device->GetEndpointId(), index)) {
-		return false;
+	/* Deserialize data and copy it from buffer into structure's fields. */
```
Damian-Nordic, line 214: Validate `readSize` before reading the buffer?

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -123,31 +200,85 @@ bool BridgeStorageManager::RemoveBridgedDeviceType(uint8_t bridgedDeviceIndex)
...
+
+	/* Check if user prepared a buffer for reading user data. It can be nullptr if not needed. */
+	if (!device.mUserData) {
+		device.mUserDataSize = 0;
+		return true;
 	}
 
-	if (!StoreBridgedDeviceNodeLabel(device->GetNodeLabel(), strlen(device->GetNodeLabel()), index)) {
+	/* Validate if user buffer size is big enough to fit the stored user data size. */
+	if (device.mUserDataSize < readSize - counter - sizeof(device.mUserDataSize)) {
```
Damian-Nordic, line 231:
```suggestion
	if (counter + sizeof(device.mUserDataSize) + device.mUserDataSize < readSize) {
```
to avoid integer underflow and always true condition when dealing with unsigned types.

So this assumes a user must know an exact size of the user data. Probably ok for now but for some user data this may be too restrictive.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
kkasperczyk-no: Do you mean validating that readSize is >= than sizeof(BridgedDevice)? I believe that upper limit is already secured, as we pass sizeof buffer, as a max limit to the `Load` method.

Damian-Nordic: That `readSize >= sizeof(all mandatory fields)` and then `readSize >= sizeof(all mandatory fields) + label length` before reading the label.

applications/matter_bridge/src/app_task.cpp
```diff
@@ -106,36 +106,26 @@ CHIP_ERROR AppTask::RestoreBridgedDevices()
...
-		}
-
-		/* Ignore an error, as node label is optional, so it may not be found. */
-		Nrf::BridgeStorageManager::Instance().LoadBridgedDeviceNodeLabel(label, sizeof(label), labelSize,
-										 indexes[i]);
+		Nrf::BridgeStorageManager::BridgedDevice device;
+#ifdef CONFIG_BRIDGED_DEVICE_BT
+		bt_addr_le_t btAddr;
+		device.mUserData = reinterpret_cast<uint8_t *>(&btAddr);
+		device.mUserDataSize = sizeof(btAddr);
```
ArekBalysNordic, line 113: Shouldn't we pack or align that data?

samples/matter/common/src/bridge/bridge_storage_manager.h
```diff
@@ -23,40 +23,32 @@ namespace Nrf
...
+
+	struct BridgedDevice {
```
ArekBalysNordic, line 37: I think if you want to serialize and deserialize this structure, you need to take care of the alignment and package.

samples/matter/common/src/bridge/bridge_storage_manager.h
```diff
@@ -23,40 +23,32 @@ namespace Nrf
...
+		uint16_t mEndpointId;
+		uint16_t mDeviceType;
+		size_t mNodeLabelLength;
+		char mNodeLabel[MatterBridgedDevice::kNodeLabelSize] = { 0 };
+		size_t mUserDataSize = 0;
+		uint8_t *mUserData = nullptr;
```
ArekBalysNordic, line 43: Why only these fields are initialized, but not other ones?

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -28,7 +33,101 @@ Nrf::PersistentStorageNode CreateIndexNode(uint8_t bridgedDeviceIndex, Nrf::Pers
...
+					return false;
+				}
+			}
+		}
+
+		version = 1;
+		Nrf::PersistentStorage::Instance().Store(&mVersion, &version, sizeof(version));
+
+	} else if (version != 1) {
+		/* Currently only no-version or version equal to 1 is supported. */
```
ArekBalysNordic, line 126: Shouldn't we parametrize it somehow to be ready for future changes?

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -28,7 +33,101 @@ Nrf::PersistentStorageNode CreateIndexNode(uint8_t bridgedDeviceIndex, Nrf::Pers
...
+	result = MigrateData();
+
+	return result;
+}
+
+bool BridgeStorageManager::MigrateData()
+{
+	uint8_t version;
+
+	/* Check if version key is present in settings to provide backward compatibility between post-2.6.0 releases and
```
ArekBalysNordic, line 57: Is the 2.6.1 a one of the post-2.6.0 releases? If so, then the migration is not necessary.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
kkasperczyk-no: Ok I added it and also validation that `readSize >= sizeof(all mandatory fields) +label length + user data size` before reading user data.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
kkasperczyk-no: > So this assumes a user must know an exact size of the user data. 

This wasn't quite my intention. The user should know the max acceptable user data size and then the field is updated with data size read from the storage. I understand it could not be obvious, so I updated the method description.

samples/matter/common/src/bridge/bridge_storage_manager.h
kkasperczyk-no: Because the other fields are mandatory, so they should be always filled with valid data.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -28,7 +33,90 @@ Nrf::PersistentStorageNode CreateIndexNode(uint8_t bridgedDeviceIndex, Nrf::Pers
...
+		uint8_t indexes[BridgeManager::kMaxBridgedDevices] = { 0 };
+		size_t indexesCount = 0;
+
+		if (LoadBridgedDevicesCount(count) &&
+		    LoadBridgedDevicesIndexes(indexes, BridgeManager::kMaxBridgedDevices, indexesCount)) {
+			/* Load all devices based on the read count number. */
+			for (size_t i = 0; i < indexesCount; i++) {
+				BridgedDevice device;
+				if (!LoadBridgedDeviceEndpointId(device.mEndpointId, i)) {
+					return false;
```
Damian-Nordic, line 70: Here you could also `goto` the `Remove...` code block instead of returning.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
kkasperczyk-no: No, it's hard to predict how many bug fix releases there will be, so I meant major releases.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -123,31 +189,109 @@ bool BridgeStorageManager::RemoveBridgedDeviceType(uint8_t bridgedDeviceIndex)
...
 
-bool BridgeStorageManager::StoreBridgedDevice(const MatterBridgedDevice *device, uint8_t index)
+bool BridgeStorageManager::LoadBridgedDevice(BridgedDevice &device, uint8_t index)
 {
-	if (!device) {
+	Nrf::PersistentStorageNode id = CreateIndexNode(index, &mBridgedDevice);
+	size_t readSize = 0;
+	uint8_t buffer[sizeof(BridgedDevice) + kMaxUserDataSize];
+	uint16_t counter = 0;
+	const uint8_t mandatoryItemsSize = sizeof(device.mEndpointId) + sizeof(device.mDeviceType) + sizeof(device.mNodeLabelLength) + sizeof(device.mUserDataSize);
```
Damian-Nordic, line 198: `mUserDataSize` is not a mandatory field, it seems.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -123,31 +189,109 @@ bool BridgeStorageManager::RemoveBridgedDeviceType(uint8_t bridgedDeviceIndex)
...
+
+	/* Validate that read size is big enough to include mandatory data. */
+	if (readSize < mandatoryItemsSize) {
```
Damian-Nordic, line 201: This should be validated after `readSize` is determined :)

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -123,31 +189,109 @@ bool BridgeStorageManager::RemoveBridgedDeviceType(uint8_t bridgedDeviceIndex)
...
+	memcpy(&device.mUserDataSize, buffer + counter, sizeof(device.mUserDataSize));
+	counter += sizeof(device.mUserDataSize);
+
+	/* Validate that user data size value read from the storage is not bigger than the one expected by the user. */
+	if (inUserDataSize < device.mUserDataSize) {
+		return false;
+	}
+
+	/* Validate that read size is big enough to include all expected data. */
+	if (readSize < mandatoryItemsSize + device.mNodeLabelLength + device.mUserDataSize) {
```
Damian-Nordic, line 246:
```suggestion
	if (readSize < counter + device.mUserDataSize) {
```

samples/matter/common/src/bridge/bridge_storage_manager.cpp
```diff
@@ -123,31 +189,109 @@ bool BridgeStorageManager::RemoveBridgedDeviceType(uint8_t bridgedDeviceIndex)
...
+
+	/* Check if user prepared a buffer for reading user data. It can be nullptr if not needed. */
+	if (!device.mUserData) {
+		device.mUserDataSize = 0;
+		return true;
+	}
+
+	uint8_t inUserDataSize = device.mUserDataSize;
+	/* Validate if user buffer size is big enough to fit the stored user data size. */
+	if (device.mUserDataSize + counter + sizeof(device.mUserDataSize) < readSize ) {
```
Damian-Nordic, line 233:
```suggestion
	if (readSize < counter + sizeof(device.mUserDataSize)) {
```
Given that you check the user buffer further (in line 241) it's enough to check that you can read the user data length.

Damian-Nordic (approved):
Last issues with index validation but part from that looks good.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
kkasperczyk-no: I thought about it, but I don't know if it's not an overkill. TBH I think we will remove it after 1, max releases, as we will not maintain this migration forever.

samples/matter/common/src/bridge/bridge_storage_manager.cpp
kkasperczyk-no: Well, even if there is no user data we have to save it to settings with size equal to 0, because we don't have other way to decide whether data are present or not after loading, right?

kkasperczyk-no: Oh wait, what I wrote doesn't match the implementation :D I have to re-think this.

kkasperczyk-no: Ok, you are right, it's not needed.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -224,6 +224,10 @@ Matter Bridge
    The :kconfig:option:`CONFIG_BRIDGE_BT_SCAN_TIMEOUT_MS` kconfig option to set the scan timeout.
 
 * Updated the implementation of the persistent storage to leverage ``NonSecure``-prefixed methods from the common Persistent Storage module.
+* Changed data structure of information stored in the persistent storage to use less settings keys.
+  The new structure uses approximately 2.5 times less memory, and provides a new field to store user-specific data.
```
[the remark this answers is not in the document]
kkasperczyk-no: And how about "uses approximately 40% of the memory used by the old structure"?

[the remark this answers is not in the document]
wiba-nordic: That should also work.

applications/matter_bridge/src/app_task.cpp
kkasperczyk-no: Discussed F2F with @markaj-nordic

samples/matter/common/src/bridge/bridge_storage_manager.h
kkasperczyk-no: Discussed F2F

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -224,6 +224,10 @@ Matter Bridge
    The :kconfig:option:`CONFIG_BRIDGE_BT_SCAN_TIMEOUT_MS` kconfig option to set the scan timeout.
 
 * Updated the implementation of the persistent storage to leverage ``NonSecure``-prefixed methods from the common Persistent Storage module.
+* Changed data structure of information stored in the persistent storage to use less settings keys.
+  The new structure uses approximately approximately 40% of the memory used by the old structure, and provides a new field to store user-specific data.
```
wiba-nordic, line 228:
```suggestion
  The new structure uses approximately 40% of the memory used by the old structure, and provides a new field to store user-specific data.
```

wiba-nordic (approved):

ArekBalysNordic (approved):

PR #14991, opened 2024-04-22
samples: matter: Make the max length storage asset key configurable.

markaj-nordic:
Introduced the CONFIG_NCS_SAMPLE_MATTER_SECURE_STORAGE_MAX_KEY_LEN
option to control the maximum length of the key under which an asset can be written to the persistent storage.

This PR should be merged after the https://github.com/nrfconnect/sdk-nrf/pull/14988 is in.


doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -190,6 +190,8 @@ Both can be used simultaneously by controlling the following Kconfig options:
 * :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_SECURE_STORAGE_BACKEND` - Activates the module based on the ARM PSA Protected Storage API implementation from the :ref:`trusted_storage_readme` |NCS| library.
   This backend implements ``Secure`` methods of the Persistent Storage API and returns ``PSErrorCode::NotSupported`` for ``NonSecure`` methods.
 
+Additionally, the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_SECURE_STORAGE_MAX_KEY_LEN` option allows you to control the maximum length of a string-type key under which an asset can be stored.
```
wiba-nordic, line 193: Wouldn't this make sense as another entry in the list starting on Line 201?

markaj-nordic: I thought about it, but this list only concerns the secure storage backend only whereas the new Kconfig is common for both backends.

wiba-nordic: Ah, right you are. In that case, I'd add this on its own between current Lines 197 and 199, edited as follows:
```suggestion
Both backends allow you to control the maximum length of a string-type key under which an asset can be stored.
You can do this using the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_SECURE_STORAGE_MAX_KEY_LEN` Kconfig option.
```
Now it kind of leads into the the Kconfigs for the secure storage backend.

wiba-nordic (changes requested):

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
markaj-nordic: Ok, thanks :)

kkasperczyk-no (approved):

samples/matter/common/src/persistent_storage/Kconfig
```diff
@@ -16,6 +16,10 @@ config NCS_SAMPLE_MATTER_SECURE_STORAGE_BACKEND
 	select TRUSTED_STORAGE
 	select PSA_PROTECTED_STORAGE
 
+config NCS_SAMPLE_MATTER_SECURE_STORAGE_MAX_KEY_LEN
```
kkasperczyk-no, line 19: Could we change the name, so it would not suggest it applies only to secure storage?

markaj-nordic: Ehhh, sorry, sure.

LuDuda (approved):

wiba-nordic (approved):

PR #14995, opened 2024-04-22
Bluetooth: Mesh: Fix handling of uninitialized values in Sensor

ludvigsj:
This fixes a few bugs where cadence-related functions would attempt to work with uninitalized sensor values causing null pointer dereferencing.

This makes it so that:
 * Cadence updates before cadence is initialized are ignored
 * The first delta check before any `prev` value is available will always return `true`.

Also adds a null check to `bt_mesh_sensor_value_get_status` by using the API function for float conversion (which already does a NULL check) instead of the bare type callback function.

### DIFF
--- a/subsys/bluetooth/mesh/sensor.c
+++ b/subsys/bluetooth/mesh/sensor.c
@@ -127,6 +127,10 @@ bool bt_mesh_sensor_delta_threshold(const struct bt_mesh_sensor *sensor,
 bool bt_mesh_sensor_delta_threshold(const struct bt_mesh_sensor *sensor,
 				    const struct bt_mesh_sensor_value *curr)
 {
+	if (!sensor->state.prev.format) {
+		/* prev is uninitialized, return true as this is the first pub. */
+		return true;
+	}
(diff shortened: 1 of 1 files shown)


m-alperen-sener (approved):

subsys/bluetooth/mesh/sensor.c
```diff
@@ -785,6 +789,11 @@ bool bt_mesh_sensor_value_in_column(const struct bt_mesh_sensor_value *value,
 void sensor_cadence_update(struct bt_mesh_sensor *sensor,
 			   const sensor_value_type *value)
 {
+	if (!sensor->state.configured) {
+		/* Ignore any update before cadence is configured. */
```
alxelax, line 793: Shouldn't `LOG_DBG` be added with appropriate message instead?
Here and above?

ludvigsj: No, because this is expected to happen. This is called for every publication. For publications happening before any cadence is configured (which is a normal thing to happen) we should ignore the cadence update call. I can of course move the ifs to all the call sites instead, but I found it easier to just add the check here and keep the call.

PR #14999, opened 2024-04-22
samples: matter: Fix for device state indications on LEDs

ArekBalysNordic:
Bluetooth LE connection to the Matter Controller was not been handled properly.

Moved the UpdatingDeviceState invocation outside the Matter Event Handler switch.

### DIFF
--- a/samples/matter/common/src/board/board.cpp
+++ b/samples/matter/common/src/board/board.cpp
@@ -298,21 +298,16 @@ void Board::StartBLEAdvertisement()
 void Board::DefaultMatterEventHandler(const ChipDeviceEvent *event, intptr_t /* unused */)
 {
 	static bool isNetworkProvisioned = false;
+	static bool isBleConnected = false;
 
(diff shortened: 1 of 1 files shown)


markaj-nordic (approved):
LGTM. I presume the problem is that we need to handle the state state regardless of the received event type?

kkasperczyk-no (approved):

ArekBalysNordic:
> LGTM. I presume the problem is that we need to handle the state state regardless of the received event type?

Yes, actually there were two problems, we need to set the check isNetworkProvisioned flag regardless of whether we use WiFi or thread -> requirements from WiFi-Thread Switched example, and if we switch state in the kCHIPoBLEAdvertisingChange case, then we cannot switch it again outside the "switch case". So I decided to move it and set the state regardless of the received event type to ensure the proper state will be set.

Damian-Nordic (approved):

PR #15025, opened 2024-04-23
lib: lte_link_control: Add RAI notification support

trantanen:
Adding support for new RAI features introduced in mfw 2.0.x:

- Add support for new `%RAI` AT notifications.
- Using `AT%RAI=2` when enabling RAI to order those notifications. Fallback to `AT%RAI=1` done for older modems.

At the same time, deprecated AT%XRAI related functions are removed, that is, `lte_lc_rai_param_set()`, `lte_lc_rai_req()` and `CONFIG_LTE_RAI_REQ_VALUE`

New AT notification looks as follows:
`%RAI: "00038107","10099",0,0`

`lte_lc` will parse the notification and send an event about it. Modem shell will print it as follows:
`RAI configuration update: Cell Id=229639, MCC=100, MNC=99, AS RAI=0, CP RAI=0`

TODO:

- [x] Unit tests
- [x] Release note
- [x] Test with mfw 1.3.x that fallback works


lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -1561,6 +1562,101 @@ int parse_mdmev(const char *at_response, enum lte_lc_modem_evt *modem_evt)
...
+	 */
+	err = string_to_int(&tmp_str[3], 10, &rai->mnc);
+	if (err) {
+		goto clean_exit;
+	}
+
+	/* Null-terminated MCC, read and store it. */
+	tmp_str[3] = '\0';
+
+	err = string_to_int(tmp_str, 10, &rai->mcc);
```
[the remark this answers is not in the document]
trantanen: Done

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -607,6 +607,15 @@ Modem libraries
     * Convenience function to get :c:struct:`location_data_details` from the :c:struct:`location_event_data`.
     * Location data details for event :c:enum:`LOCATION_EVT_RESULT_UNKNOWN`.
 
+* :ref:`lte_lc_readme` library:
+
+  * Added
+
+    * Support for new ``%RAI`` AT notification.
+    * Using the ``AT%RAI=2`` AT command when enabling RAI to order the ``%RAI`` notification. Fallback done to the ``AT%RAI=1`` command.
```
peknis, line 615:
```suggestion
    * Using the ``AT%RAI=2`` AT command when enabling RAI to order the ``%RAI`` notification.
      Fallback done to the ``AT%RAI=1`` command.
```

lib/lte_link_control/lte_lc_helpers.h
```diff
@@ -341,6 +349,19 @@ int parse_coneval(const char *at_response, struct lte_lc_conn_eval_params *param
  */
 int parse_mdmev(const char *at_response, enum lte_lc_modem_evt *modem_evt);
 
+/* @brief Parses a RAI response and extracts RAI configuration.
```
peknis, line 352:
```suggestion
/* @brief Parse a RAI response and extract RAI configuration.
```

lib/lte_link_control/lte_lc_helpers.h
```diff
@@ -341,6 +349,19 @@ int parse_coneval(const char *at_response, struct lte_lc_conn_eval_params *param
...
+ *
+ * @param at_response AT response.
+ * @param RAI RAI configuration.
```
peknis, line 355:
```suggestion
 * @param rai RAI configuration.
```

peknis (changes requested):

include/modem/lte_lc.h
```diff
@@ -1201,6 +1228,9 @@ struct lte_lc_evt {
 
 		/** Payload for event @ref LTE_LC_EVT_NEIGHBOR_CELL_MEAS. */
 		struct lte_lc_cells_info cells_info;
+
+		/** Payload for event @ref LTE_LC_EVT_RAI_UPDATE. */
+		struct lte_lc_rai rai;
```
tokangas, line 1233: The name `lte_lc_rai` is a bit ambiguous, I think. The struct could be called `lte_lc_rai_cfg` and the member `rai_cfg`. This would also be aligned with other payload types.

lib/lte_link_control/lte_lc.c
```diff
@@ -506,6 +507,26 @@ static void at_handler_mdmev(const char *response)
...
+
+	LOG_DBG("%%RAI notification");
+
+	err = parse_rai(response, &evt.rai);
+	if (err) {
+		LOG_ERR("Can't parse RAI notification, error: %d", err);
+		return;
+	}
+
+	evt.type = LTE_LC_EVT_RAI_UPDATE;
```
tokangas, line 525: I assume that modem only sends the `%RAI` notification when something has changed? So, no need to have a check here?

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -1690,3 +1760,24 @@ int parse_periodic_search_pattern(const char *const pattern_str,
 
 	return 0;
 }
+
+int lte_lc_rai_set(void)
+{
+	int err;
+
+	LOG_DBG("Enabling RAI with notifications");
```
tokangas, line 1768: This log is misleading when `CONFIG_LTE_RAI_REQ` is disabled. Maybe change it to something like "Configuring RAI". If `%RAI=2` fails, there will anyway be a separate log for that.

lib/lte_link_control/lte_lc_helpers.h
```diff
@@ -341,6 +349,19 @@ int parse_coneval(const char *at_response, struct lte_lc_conn_eval_params *param
  */
 int parse_mdmev(const char *at_response, enum lte_lc_modem_evt *modem_evt);
 
+/* @brief Parse a RAI response and extract RAI configuration.
```
tokangas, line 352: There are several functions which talk about responses, when they are really for parsing notifications. IMO this can be merged as it is, but we could later do a cleanup for all such instances. After all, the response to `%RAI` is only `OK` or `ERROR`, only the unsolicited notification contains the configuration.

lib/lte_link_control/lte_lc_helpers.h
```diff
@@ -406,3 +427,13 @@ char *periodic_search_pattern_get(char *const buf, size_t buf_size,
  */
 int parse_periodic_search_pattern(const char *const pattern_str,
 				  struct lte_lc_periodic_search_pattern *pattern);
+
+/* @brief Set RAI based on @kconfig{CONFIG_LTE_RAI_REQ}.
+ *
+ * If enabling of RAI is requested, AT%XRAI=2 is used to order unsolicited RAI notifications.
+ * If setting that fails, AT%XRAI=1 is used.
```
tokangas, line 434:
```suggestion
 * If enabling of RAI is requested, AT%RAI=2 is used to order unsolicited RAI notifications.
 * If setting that fails, AT%RAI=1 is used.
```

tokangas (changes requested):

lib/lte_link_control/lte_lc_helpers.h
trantanen: Yep, I noticed this same thing and didn't want to do a big commenting/documentation changes. There are many such small things I'd like to change in lte_lc comments.

lib/lte_link_control/lte_lc_helpers.c
trantanen: I've made it an if-else-statement so we clearly see in the boot how it's configured:
```
	if (IS_ENABLED(CONFIG_LTE_RAI_REQ)) {
		LOG_DBG("Enabling RAI with notifications");
	} else {
		LOG_DBG("Disabling RAI");
	}
```

lib/lte_link_control/lte_lc.c
trantanen: This notification is sent when RRC connection is established.

peknis (approved):

lib/lte_link_control/lte_lc.c
tokangas: Ok, so should the event be sent only when something has changed? LTE LC has this kind of checks for example for `LTE_LC_EVT_NW_REG_STATUS` and `LTE_LC_EVT_CELL_UPDATE`.

trantanen: And doesn't have those checks for `LTE_LC_EVT_EDRX_UPDATE` and `LTE_LC_EVT_RRC_UPDATE`

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -1561,6 +1559,80 @@ int parse_mdmev(const char *at_response, enum lte_lc_modem_evt *modem_evt)
 	return -ENODATA;
 }
 
+int parse_rai(const char *at_response, struct lte_lc_rai_cfg *rai_cfg)
+{
+	struct at_param_list resp_list;
+	int err;
+	int tmp_int;
```
MirkoCovizzi, line 1566: Is there a better name for this variable?

lib/lte_link_control/lte_lc.c
tokangas: At least `+CSCON` is never sent by the modem unless the value has changed. Most likely it's the same for `+CEDRXP` as well. Let's discuss this offline.

include/modem/lte_lc.h
```diff
@@ -1576,42 +1606,6 @@ int lte_lc_edrx_req(bool enable);
...
 int lte_lc_edrx_get(struct lte_lc_edrx_cfg *edrx_cfg);
 
-/**
- * Set the RAI value to be used.
- *
- * RAI can be subsequently enabled using lte_lc_rai_req().
- *
- * For reference see 3GPP 24.301 Ch. 9.9.4.25.
- *
- * @deprecated Use @kconfig{CONFIG_LTE_RAI_REQ} and socket option ``SO_RAI`` instead.
```
gregersrygg, line 1586: What to use instead was only documented in the code that has been deleted, and will be lost for customers that upgrade from older versions. Please add info about what to use instead in the migration guide for all of the deleted functionality.

gregersrygg (approved):
atv2 and net/udp sample changes looks good to me 👍

lib/lte_link_control/lte_lc.c
tokangas: It seems that modem already does some filtering, so no filtering is needed in LTE LC.

tokangas (approved):

lib/lte_link_control/lte_lc.c
```diff
@@ -510,6 +507,26 @@ static void at_handler_mdmev(const char *response)
...
 }
 
+static void at_handler_rai(const char *response)
+{
+	int err;
+	struct lte_lc_evt evt = {0};
+
+	__ASSERT_NO_MSG(response != NULL);
+
+	LOG_DBG("%%RAI notification");
```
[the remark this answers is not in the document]
trantanen: I guess we should postpone this significantly. I'll likely close this PR and create a new one for removing deprecated AT%XRAI stuff.

lib/lte_link_control/lte_lc_helpers.c
trantanen: I will combine this and enabled as they are temporary variables. This naming is used elsewhere in the file for temporary variable and I guess it's a matter of opinion whether it's ok or really bad. I don't know if `value` etc would be any better so please suggest.

include/modem/lte_lc.h
trantanen: I've added this statement into the release notes as well. This PR will get closed and I'll create another one to remove deprecated stuff from lte_lc.

PR #15030, opened 2024-04-23
bt: kconfig: Support BT_CTLR_ECDH_LIB_OBERON with TFM

SebastianBoe:
Support BT_CTLR_ECDH_LIB_OBERON with TFM and with an "OBERON_BACKEND".

I don't see why this shouldn't be supported.


hermabe (approved):
Seems the `depends on` was only added because the option combination didnt compile. So as long as it compiles, I am fine with this.

SebastianBoe:
compiles for me at least :)

rugeGerritsen (approved):

seko-nordic (approved):

PR #14463, opened 2024-03-08
nRF54L15: Add initial TF-M support

Vge0rge:
Add  initial support for TF-M in nRF54L15. 

This has the DTS files here in this repository until we have more mature support, then we will add upstream TF-M support and we will relocate the files. 

Adds support for the tfm_hello_world sample and the following crypto samples:

- aes_cbc
- aes_ccm
- aes_ctr
- aes_gcm
- chachapoly
- ecdh
- ecdsa
- eddsa
- hkdf
- hmac
- rng
- rsa
- sha256

test_crypto: PR-614
test_tfm: PR-155

PR #15031, opened 2024-04-23
Fix ns dt

SebastianBoe:
misc. 54L+TFM board support fixes


mswarowsky (approved):

mswarowsky:
Should have Jira to group common stuff of the secure and non-secure board inf future to not have to do these updates

SebastianBoe:
I agree. See https://github.com/nrfconnect/sdk-nrf/pull/14463#discussion_r1528056012

vili-nordic (approved):

PR #15035, opened 2024-04-23
applications: nrf5340_audio: Rail the presentation delay.

gWacey:
OCT--2976
	
Ensure the preferred presentation delay is within the range of the	min and max delays, rail if not.
Search for the presentation delay function updated to search over the end points for the direction.

### DIFF
--- a/applications/nrf5340_audio/src/audio/Kconfig
+++ b/applications/nrf5340_audio/src/audio/Kconfig
@@ -34,14 +34,16 @@ config AUDIO_MIN_PRES_DLY_US
 	int "The minimum presentation delay"
 	default 3000
 	help
-	  The minimum presentation delay in micro seconds determined by
-	  the audio system processing and the minimum buffering.
+	  The minimum allowable presentation delay in microseconds.
+	  This needs to allow time for decoding and internal routing.
+	  For 48kHz sampling rate and 96kbps bitrate this is about 4000 us.
 
 config AUDIO_MAX_PRES_DLY_US
 	int "The maximum presentation delay"
 	default 60000
 	help
-	  The maximum presentation delay in micro seconds.
+	  The maximum allowable presentation delay in microseconds.
+	  Increasing this will also increase the FIFO buffers to allow buffering.
 
 choice AUDIO_SYSTEM_SAMPLE_RATE
 	prompt "System audio sample rate"
@@ -107,21 +109,6 @@ config AUDIO_BIT_DEPTH_OCTETS
 	default 4 if AUDIO_BIT_DEPTH_32
 	help
 	  Bit depth of one sample in storage given in octets.
-
-config AUDIO_MIN_PRES_DLY_US
-	int "The minimum presentation delay"
-	default 4000
-	help
-	  The minimum allowable presentation delay in microseconds.
-	  This needs to allow time for decoding and internal routing.
-	  For 48kHz sampling rate and 96kbps bitrate this is about 4000 us.
-
-config AUDIO_MAX_PRES_DLY_US
-	int "The maximum presentation delay"
-	default 60000
-	help
-	  The maximum allowable presentation delay in microseconds.
-	  Increasing this will also increase the FIFO buffers to allow buffering.
 
 choice AUDIO_SOURCE_GATEWAY
 	prompt "Audio source for gateway"
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
@@ -51,6 +51,33 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
 	return false;
 }
 
+bool le_audio_ep_qos_configured(struct bt_bap_ep const *const ep)
+{
+	int ret;
+	struct bt_bap_ep_info ep_info;
+
+	if (ep == NULL) {
+		LOG_DBG("EP is NULL");
+		/* If an endpoint is NULL it is not in any of the states */
+		return false;
+	}
+
+	ret = bt_bap_ep_get_info(ep, &ep_info);
+	if (ret) {
+		LOG_WRN("Unable to get info for ep");
+		return false;
+	}
+
+	if (ep_info.state == BT_BAP_EP_STATE_QOS_CONFIGURED ||
+	    ep_info.state == BT_BAP_EP_STATE_ENABLING ||
+	    ep_info.state == BT_BAP_EP_STATE_STREAMING ||
+	    ep_info.state == BT_BAP_EP_STATE_DISABLING) {
+		return true;
+	}
+
+	return false;
+}
+
 int le_audio_freq_hz_get(const struct bt_audio_codec_cfg *codec, int *freq_hz)
 {
 	int ret;
--- a/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
+++ b/applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
@@ -191,56 +191,82 @@ K_WORK_DEFINE(cap_start_work, cap_start_worker);
  */
 static int device_pres_delay_find(uint8_t index, uint32_t *pres_dly_us)
 {
-	uint32_t pres_dly_min = unicast_servers[index].sink_ep->qos_pref.pd_min;
-	uint32_t pres_dly_max = unicast_servers[index].sink_ep->qos_pref.pd_max;
-	uint32_t pref_dly_min = unicast_servers[index].sink_ep->qos_pref.pref_pd_min;
-	uint32_t pref_dly_max = unicast_servers[index].sink_ep->qos_pref.pref_pd_max;
+	uint32_t pd_min = unicast_servers[index].sink_ep->qos_pref.pd_min;
+	uint32_t pd_max = unicast_servers[index].sink_ep->qos_pref.pd_max;
+	uint32_t pref_pd_min = unicast_servers[index].sink_ep->qos_pref.pref_pd_min;
+	uint32_t pref_pd_max = unicast_servers[index].sink_ep->qos_pref.pref_pd_max;
 
 	LOG_DBG("Index: %d, Pref min: %d, pref max: %d, pres_min: %d, pres_max: %d", index,
-		pref_dly_min, pref_dly_max, pres_dly_min, pres_dly_max);
+		pref_pd_min, pref_pd_max, pd_min, pd_max);
+
+	*pres_dly_us = 0;
 
 	for (int i = 0; i < ARRAY_SIZE(unicast_servers); i++) {
-		if (unicast_servers[i].sink_ep != NULL) {
-			pres_dly_min =
-				MAX(pres_dly_min, unicast_servers[i].sink_ep->qos_pref.pd_min);
-			pres_dly_max =
-				MIN(pres_dly_max, unicast_servers[i].sink_ep->qos_pref.pd_max);
-			pref_dly_min =
-				MAX(pref_dly_min, unicast_servers[i].sink_ep->qos_pref.pref_pd_min);
-			pref_dly_max =
-				MIN(pref_dly_max, unicast_servers[i].sink_ep->qos_pref.pref_pd_max);
+		if (le_audio_ep_qos_configured(unicast_servers[i].sink_ep)) {
+			LOG_DBG("i: %d, Pref min: %d, pref max: %d, pres_min: %d, pres_max: %d", i,
+				unicast_servers[i].sink_ep->qos_pref.pref_pd_min,
+				unicast_servers[i].sink_ep->qos_pref.pref_pd_max,
+				unicast_servers[i].sink_ep->qos_pref.pd_min,
+				unicast_servers[i].sink_ep->qos_pref.pd_max);
+
+			pd_min = MAX(pd_min, unicast_servers[i].sink_ep->qos_pref.pd_min);
+			pref_pd_min =
+				MAX(pref_pd_min, unicast_servers[i].sink_ep->qos_pref.pref_pd_min);
+
+			if (unicast_servers[i].sink_ep->qos_pref.pd_max) {
+				pd_max = MIN(pd_max, unicast_servers[i].sink_ep->qos_pref.pd_max);
+			}
+
+			if (unicast_servers[i].sink_ep->qos_pref.pref_pd_max) {
+				pref_pd_max = MIN(pref_pd_max,
+						  unicast_servers[i].sink_ep->qos_pref.pref_pd_max);
+			}
 		}
 	}
 
 	if (IS_ENABLED(CONFIG_BT_AUDIO_PRES_DELAY_SRCH_MIN)) {
-		*pres_dly_us = pres_dly_min;
+		*pres_dly_us = pd_min;
 
 		return 0;
 	}
 
 	if (IS_ENABLED(CONFIG_BT_AUDIO_PRES_DELAY_SRCH_MAX)) {
-		*pres_dly_us = pres_dly_max;
+		*pres_dly_us = pd_max;
 
 		return 0;
 	}
 
 	if (IS_ENABLED(CONFIG_BT_AUDIO_PRES_DELAY_SRCH_PREF_MIN)) {
-		/* Preferred min is 0, so we set min supported */
-		if (pref_dly_min == 0) {
-			*pres_dly_us = pres_dly_min;
+		if (pref_pd_min == 0) {
+			*pres_dly_us = pd_min;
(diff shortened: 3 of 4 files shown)


applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -270,21 +270,21 @@ static int headset_pres_delay_find(uint8_t index, uint32_t *pres_dly_us)
...
 	if (IS_ENABLED(CONFIG_BT_AUDIO_PRES_DELAY_SRCH_PREF_MIN)) {
 		/* Preferred min is 0, so we set min supported */
-		if (pref_dly_min == 0) {
-			*pres_dly_us = pres_dly_min;
-		} else {
+		if (IN_RANGE(pref_dly_min, pres_dly_min, pres_dly_max)) {
 			*pres_dly_us = pref_dly_min;
+		} else {
+			*pres_dly_us = pres_dly_min;
 		}
```
alexsven, line 277: This code will set pres_dly_min even if pref_dly_min is larger than pres_dly_max.  It is a strange scenario, but I guess it could happen. We should rail the pref_min/pref_max at the start of the function to be within range of pres_min/pres_max and then continue with the rest of the code.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -1020,6 +1020,14 @@ static void stream_configured_cb(struct bt_bap_stream *stream,
 		LOG_INF("%s sink stream configured", headsets[channel_index].ch_name);
 		le_audio_print_codec(headsets[channel_index].sink_stream.codec_cfg,
 				     stream->ep->dir);
+		LOG_DBG("Configured Stream info: %s, %p, dir %d", headsets[channel_index].ch_name,
+			(void *)stream, stream->ep->dir);
+
+		ret = headset_pres_delay_find(headsets[channel_index].sink_ep, &new_pres_dly_us);
```
alexsven, line 1026: There is a build error as well, you need to set new_pres_dly_us on line:1010 to 0 when initializing it.

alexsven (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -231,28 +231,28 @@ static int discover_list_index_get(struct bt_conn const *const conn)
...
+static int headset_pres_delay_find(struct bt_bap_ep *ep, uint32_t *pres_dly_us)
 {
-	uint32_t pres_dly_min = headsets[index].sink_ep->qos_pref.pd_min;
-	uint32_t pres_dly_max = headsets[index].sink_ep->qos_pref.pd_max;
-	uint32_t pref_dly_min = headsets[index].sink_ep->qos_pref.pref_pd_min;
-	uint32_t pref_dly_max = headsets[index].sink_ep->qos_pref.pref_pd_max;
+	uint32_t pres_dly_min = ep->qos_pref.pd_min;
+	uint32_t pres_dly_max = ep->qos_pref.pd_max;
+	uint32_t pref_dly_min = ep->qos_pref.pref_pd_min;
+	uint32_t pref_dly_max = ep->qos_pref.pref_pd_max;
```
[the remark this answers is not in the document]
andvib: I don't see how the old code did iteration either - doesn't this code just get the reference before deciding the delay?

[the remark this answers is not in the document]
gWacey: Oops, apologies, it looks as if I have pushed a partial test build!
Have got the right branch now and it is back to the original with the simple changes for the clamping of the delay.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -231,28 +231,28 @@ static int discover_list_index_get(struct bt_conn const *const conn)
...
-			pres_dly_min = MAX(pres_dly_min, headsets[i].sink_ep->qos_pref.pd_min);
-			pres_dly_max = MIN(pres_dly_max, headsets[i].sink_ep->qos_pref.pd_max);
-			pref_dly_min = MAX(pref_dly_min, headsets[i].sink_ep->qos_pref.pref_pd_min);
-			pref_dly_max = MIN(pref_dly_max, headsets[i].sink_ep->qos_pref.pref_pd_max);
+		if (ep != NULL) {
+			pres_dly_min = MAX(pres_dly_min, ep->qos_pref.pd_min);
+			pres_dly_max = MIN(pres_dly_max, ep->qos_pref.pd_max);
+			pref_dly_min = MAX(pref_dly_min, ep->qos_pref.pref_pd_min);
+			pref_dly_max = MIN(pref_dly_max, ep->qos_pref.pref_pd_max);
 		}
```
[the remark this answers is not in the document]
gWacey: See above comment

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
gWacey: Good spot!

I have updated the code to now test if the pref min or max is outside the pres min and max. It will also error if the pref_dly_max is less than the pres_dly_min or the pref_dly_min is greater than the pres_dly_max. If either of thess occur then there has been an error in the serach or defined delays.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
gWacey: See above comments, moved the search call back for both source and sink.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -269,24 +281,31 @@ static int headset_pres_delay_find(uint8_t index, uint32_t *pres_dly_us)
...
 		return 0;
 	}
 
 	if (IS_ENABLED(CONFIG_BT_AUDIO_PRES_DELAY_SRCH_PREF_MAX)) {
-		/* Preferred max is 0, so we set max supported */
-		if (pref_dly_max == 0) {
+		if (pref_dly_max == 0 || pref_dly_max > pres_dly_max) {
 			*pres_dly_us = pres_dly_max;
-		} else {
+		} else if (pref_dly_max < pres_dly_min) {
```
alexsven, line 301: I think this should be:
```suggestion
		} else if (pref_dly_max >= pres_dly_min) {
```
As it stands now the pres_dly_us would be less than pres_dly_min

alexsven (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
gWacey: Thanks this is a pain to test as I have to change a number of Kconfigs to exercise each issue. Will workout how to test this!

applications/nrf5340_audio/src/audio/Kconfig
alexsven, line 1: The changes in this files should not be pushed up. I'm guessing you removed these for debugging purposes? These are used a couple of key places and are quite important

alexsven (changes requested):

applications/nrf5340_audio/src/audio/Kconfig
gWacey: Err no, these are already defined at line 33 and 40, so thought this might cause issues if defined in two places.
However these have better comments so will update.

alexsven: Ah I see, weird that they have been doubled up for so long.

alexsven (approved):

rick1082 (approved):

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
```diff
@@ -36,6 +36,32 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
...
+		return false;
+	}
+
+	ret = bt_bap_ep_get_info(ep, &ep_info);
+	if (ret) {
+		LOG_WRN("Unable to get info for ep");
+		return false;
+	}
+
+	if (ep_info.state == BT_BAP_EP_STATE_IDLE ||
```
koffes, line 56: This should look for positives and not negatives. I.e. it should check for valid states and return true. If another new state is added, it will return true for this possibly invalid state.

applications/nrf5340_audio/src/bluetooth/bt_stream/le_audio.c
```diff
@@ -36,6 +36,32 @@ bool le_audio_ep_state_check(struct bt_bap_ep *ep, enum bt_bap_ep_state state)
 	return false;
 }
 
+bool le_audio_ep_qos_configured(struct bt_bap_ep *ep)
```
koffes, line 39: ep is not changed in this function, hence:
```suggestion
bool le_audio_ep_qos_configured(struct bt_bap_ep const * const ep)
```

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -144,54 +144,78 @@ static void le_audio_event_publish(enum le_audio_evt_type event, struct bt_conn
...
-	uint32_t pref_dly_min = headsets[index].sink_ep->qos_pref.pref_pd_min;
-	uint32_t pref_dly_max = headsets[index].sink_ep->qos_pref.pref_pd_max;
+	uint32_t pd_min = headsets[index].sink_ep->qos_pref.pd_min;
+	uint32_t pd_max = headsets[index].sink_ep->qos_pref.pd_max;
+	uint32_t pref_pd_min = headsets[index].sink_ep->qos_pref.pref_pd_min;
+	uint32_t pref_pd_max = headsets[index].sink_ep->qos_pref.pref_pd_max;
 
 	LOG_DBG("Index: %d, Pref min: %d, pref max: %d, pres_min: %d, pres_max: %d", index,
-		pref_dly_min, pref_dly_max, pres_dly_min, pres_dly_max);
+		headsets[index].sink_ep->qos_pref.pd_min,
```
koffes, line 153: The print text and values do not match. 
headsets[index].sink_ep->qos_pref.pd_min is printed twice.

koffes (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -144,54 +144,78 @@ static void le_audio_event_publish(enum le_audio_evt_type event, struct bt_conn
...
-		if (headsets[i].sink_ep != NULL) {
-			pres_dly_min = MAX(pres_dly_min, headsets[i].sink_ep->qos_pref.pd_min);
-			pres_dly_max = MIN(pres_dly_max, headsets[i].sink_ep->qos_pref.pd_max);
-			pref_dly_min = MAX(pref_dly_min, headsets[i].sink_ep->qos_pref.pref_pd_min);
-			pref_dly_max = MIN(pref_dly_max, headsets[i].sink_ep->qos_pref.pref_pd_max);
+		if (le_audio_ep_qos_configured(headsets[i].sink_ep)) {
+			LOG_DBG("i: %d, Pref min: %d, pref max: %d, pres_min: %d, pres_max: %d", i,
+				headsets[i].sink_ep->qos_pref.pd_min,
+				headsets[i].sink_ep->qos_pref.pref_pd_max,
+				headsets[i].sink_ep->qos_pref.pd_min,
```
koffes, line 164: pd min is printed twice.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -144,54 +144,78 @@ static void le_audio_event_publish(enum le_audio_evt_type event, struct bt_conn
...
+				headsets[i].sink_ep->qos_pref.pd_max);
+
+			pd_min = MAX(pd_min, headsets[i].sink_ep->qos_pref.pd_min);
+			pref_pd_min = MAX(pref_pd_min, headsets[i].sink_ep->qos_pref.pref_pd_min);
+
+			if (headsets[i].sink_ep->qos_pref.pd_max) {
```
koffes, line 170: Why do we need these if checks here? And why checks on only min and not max above?

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -144,54 +144,78 @@ static void le_audio_event_publish(enum le_audio_evt_type event, struct bt_conn
...
+		*pres_dly_us = pd_max;
 
 		return 0;
 	}
 
 	if (IS_ENABLED(CONFIG_BT_AUDIO_PRES_DELAY_SRCH_PREF_MIN)) {
-		/* Preferred min is 0, so we set min supported */
-		if (pref_dly_min == 0) {
-			*pres_dly_us = pres_dly_min;
+		if (pref_pd_min == 0 || pref_pd_min < pd_min) {
```
koffes, line 194: If the preferred is less than pd_min, would not that be a misconfiguration?

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -144,54 +144,78 @@ static void le_audio_event_publish(enum le_audio_evt_type event, struct bt_conn
  */
 static int headset_pres_delay_find(uint8_t index, uint32_t *pres_dly_us)
```
koffes, line 145: Discuss testing.

koffes (changes requested):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
gWacey: The pd_max can be 0 (BT_AUDIO_PD_PREF_NONE) and if we use MIN then it will always give 0 which is wrong if any other EP has a none zero maximum allowable pd or preferred pd. We are do not have this issue when testing with MAX as any unsigned value is greater than zero.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
gWacey: True and the same for pref_pd_max > pd_max.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
gWacey: This is painful as we use Kconfigs to set all these value and our Kconfigs have great testing for validity of the pd and pref pd values. I have to comment out or hard code values that test the failure conditions. This code will probably all change when additions to the host are added.

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
```diff
@@ -144,54 +144,78 @@ static void le_audio_event_publish(enum le_audio_evt_type event, struct bt_conn
...
 
 	if (IS_ENABLED(CONFIG_BT_AUDIO_PRES_DELAY_SRCH_PREF_MIN)) {
-		/* Preferred min is 0, so we set min supported */
-		if (pref_dly_min == 0) {
-			*pres_dly_us = pres_dly_min;
+		if (pref_pd_min == 0) {
+			*pres_dly_us = pd_min;
+		} else if (pref_pd_min <= pd_max) {
+			*pres_dly_us = pref_pd_min;
 		} else {
```
koffes, line 198:
```suggestion
		if (pref_pd_min == 0) {
			*pres_dly_us = pd_min;
		} else if (pref_pd_min = pd_max) {
			*pres_dly_us = pref_pd_min;
		} else if (pref_pd_min < pd_max) {
			*pres_dly_us = pref_pd_min;
			LOG_WRN("....");
		} else {
```

koffes: And the same below.

koffes (approved):

koffes (approved):

applications/nrf5340_audio/src/bluetooth/bt_stream/unicast/unicast_client.c
gWacey: Not sure I follow the above, why the warning?

PR #15036, opened 2024-04-23
applications: nrf5340_audio: Upmerge 25-03-2024

gWacey:
OCT-2964
	
Update buildprog.py to follow the new Zephyr board file directory structure.
Update the broadcast sink syncable_cb() API.

Can only merge after Zephyr board file directory restructure.


applications/nrf5340_audio/tools/buildprog/buildprog.py
```diff
@@ -39,7 +39,7 @@
 else:
     AUDIO_KIT_SERIAL_NUMBERS_JSON = Path(
         os.getenv("AUDIO_KIT_SERIAL_NUMBERS_JSON"))
-TARGET_BOARD_NRF5340_AUDIO_DK_APP_NAME = "nrf5340_audio_dk_nrf5340_cpuapp"
+TARGET_BOARD_NRF5340_AUDIO_DK_APP_NAME = "nrf5340_audio_dk/nrf5340/cpuapp"
```
rick1082, line 42: Seems this is a big change, would this mean we also need to update the doc for updating the DK name when using west build?
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/applications/nrf5340_audio/doc/building.html#building-the-application

rick1082 (approved):

PR #15038, opened 2024-04-24
manifest: Update zephyr with nrf54l15 counter marking

nordic-krch:
Update zephyr with marking nrf54l15pdk with counter tag so that counter tests can run on this board.


nordic-piks (approved):

nordic-krch:
Got in with upmerge

PR #15045, opened 2024-04-24
samples: wifi: softap: Add supplicant init check

D-Triveni:
Implement a wait check to ensure SoftAP enables
only after WPA supplicant initialization and
interface addition.

Fixes SHEL-2709.

### DIFF
--- a/samples/wifi/softap/src/main.c
+++ b/samples/wifi/softap/src/main.c
@@ -17,7 +17,10 @@ LOG_MODULE_REGISTER(softap, CONFIG_LOG_DEFAULT_LEVEL);
 #include <zephyr/net/dhcpv4_server.h>
 
 #define WIFI_SAP_MGMT_EVENTS (NET_EVENT_WIFI_AP_ENABLE_RESULT)
-
+void wifi_softap_init(struct k_work *work);
+K_WORK_DEFINE(sap_init_work, wifi_softap_init);
+int wpa_supp_events_register(void);
+int wait_for_wpa_s_ready(void);
 static struct net_mgmt_event_callback wifi_sap_mgmt_cb;
 
 static K_MUTEX_DEFINE(wifi_ap_sta_list_lock);
@@ -325,9 +328,28 @@ out:
 		ret = func(__VA_ARGS__); \
 		if (ret) { \
 			LOG_ERR("Failed to configure %s", #func); \
-			return -1; \
+			return; \
 		} \
 	} while (0)
+
+
+void wifi_softap_init(struct k_work *work)
+{
+	int ret;
+
+	ret = wait_for_wpa_s_ready();
+	if (ret < 0) {
+		LOG_ERR("Failed to wait for WPA supplicant to be ready");
+	}
+
+	CHECK_RET(wifi_set_reg_domain);
+
+	CHECK_RET(configure_dhcp_server);
+
+	CHECK_RET(wifi_softap_enable);
+
+	cmd_wifi_status();
+}
 
 int main(void)
 {
@@ -339,13 +361,11 @@ int main(void)
 
 	net_mgmt_add_event_callback(&wifi_sap_mgmt_cb);
 
-	CHECK_RET(wifi_set_reg_domain);
 
-	CHECK_RET(configure_dhcp_server);
-
-	CHECK_RET(wifi_softap_enable);
-
-	cmd_wifi_status();
+	ret = wpa_supp_events_register();
+	if (ret < 0) {
+		LOG_ERR("Failed to register WPA supplicant events");
+	}
 
 	return 0;
 }
--- a/samples/wifi/softap/src/wpas_events.c
+++ b/samples/wifi/softap/src/wpas_events.c
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/net/net_event.h>
+#include <ctrl_iface_zephyr.h>
+#include <supp_events.h>
+#include <supp_main.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/kernel.h>
+LOG_MODULE_REGISTER(wpas_event, CONFIG_LOG_DEFAULT_LEVEL);
+
+#define WPAS_READY_TIMEOUT_MS 500
+
+#define WPA_SUPP_EVENTS (NET_EVENT_WPA_SUPP_READY)
+
+static struct net_mgmt_event_callback net_wpa_supp_cb;
+extern struct k_work sap_init_work;
+K_SEM_DEFINE(wpa_supp_ready_sem, 0, 1);
+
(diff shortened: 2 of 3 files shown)


samples/wifi/softap/src/wpas_events.c
```diff
@@ -0,0 +1,76 @@
...
+LOG_MODULE_REGISTER(wpas_event, CONFIG_LOG_DEFAULT_LEVEL);
+
+/* TODO: Handle other events */
+#define WPA_SUPP_EVENTS (NET_EVENT_WPA_SUPP_READY)
+
+static struct net_mgmt_event_callback net_wpa_supp_cb;
+
+K_SEM_DEFINE(wpa_supp_ready_sem, 0, 1);
+
+static void handle_wpa_supp_ready(struct net_mgmt_event_callback *cb)
```
krish2718, line 22: Rather copy-pasting this file in each and every sample, should we just make a library or a common helper to avoid duplication?

samples/wifi/softap/src/wpas_events.c
```diff
@@ -0,0 +1,76 @@
...
+
+#include <zephyr/net/net_event.h>
+#include <ctrl_iface_zephyr.h>
+#include <supp_events.h>
+#include <supp_main.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/kernel.h>
+LOG_MODULE_REGISTER(wpas_event, CONFIG_LOG_DEFAULT_LEVEL);
+
+/* TODO: Handle other events */
```
krish2718, line 15: We cannot just keep copying TODOs :), as this is the second copy, now is the time to implement them.

krish2718 (changes requested):
Let's discuss about possibility of a wrapper layer that is common to all samples/applications to avoid duplication.

samples/wifi/softap/src/main.c
```diff
@@ -241,6 +242,11 @@ static int wifi_softap_enable(void)
 		goto out;
 	}
 
+	ret = wait_for_wpa_s_ready();
```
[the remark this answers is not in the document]
VivekUppunda: Can the main thread achieve anything without the supplicant coming up as this is the last step in AP main thread enablement? If not, then having a second thread, I am not sure achieves much. The AP operation is dependent on the supplicant and better to fail the process if supplicant has not come up. Adjusting supplicant init time to largest known load value might be better I would believe. 
Thanks 
Vivek

samples/wifi/softap/src/wpas_events.c
VivekUppunda: I would agree to remove this comment. If there is a task to add further events, handle the same in that PR

samples/wifi/softap/src/wpas_events.c
```diff
@@ -0,0 +1,76 @@
...
+
+static void handle_wpa_supp_ready(struct net_mgmt_event_callback *cb)
+{
+	k_sem_give(&wpa_supp_ready_sem);
+}
+
+static void wpa_supp_event_handler(struct net_mgmt_event_callback *cb,
+	uint32_t mgmt_event, struct net_if *iface)
+{
+	/* TODO: Handle other events */
```
VivekUppunda, line 30: same comment as above

samples/wifi/softap/src/wpas_events.c
```diff
@@ -0,0 +1,76 @@
...
+		break;
+	default:
+		LOG_DBG("Unhandled event (%d)", mgmt_event);
+		break;
+	}
+}
+
+int wait_for_wpa_s_ready(void)
+{
+	struct wpa_supplicant *wpa_s = z_wpas_get_handle_by_ifname("nordic_wlan0");
```
VivekUppunda, line 43: can `nordic_wlan0` be used as a #define and later used everywhere else?

samples/wifi/softap/src/main.c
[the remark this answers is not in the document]
krish2718: Yes, remember these samples are not just to get things working, they should *demonstrate* the best usage of Wi-Fi. So, in a system where other connectivity technologies are present, Wi-Fi being operational is optional.

samples/wifi/softap/src/wpas_events.c
krish2718: Missed this one, good catch. This is old code copy-pasted again, now we take this from the DTS.

samples/wifi/softap/src/wpas_events.c
D-Triveni: Will create a task for this and handle in different PR.

samples/wifi/softap/src/wpas_events.c
```diff
@@ -0,0 +1,76 @@
...
+{
+	k_sem_give(&wpa_supp_ready_sem);
+}
+
+static void wpa_supp_event_handler(struct net_mgmt_event_callback *cb,
+	uint32_t mgmt_event, struct net_if *iface)
+{
+	/* TODO: Handle other events */
+	switch (mgmt_event) {
+	case NET_EVENT_WPA_SUPP_READY:
```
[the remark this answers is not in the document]
D-Triveni: As `NET_EVENT_WPA_SUPP_READY` is generated after `NET_EVENT_WPA_SUPP_CMD_IFACE_ADDED`, the supp_ready event is prioritized in the code.

samples/wifi/softap/src/wpas_events.c
D-Triveni: Planning to take up this as another task. Will handle this as a separate task.

samples/wifi/softap/src/main.c
```diff
@@ -325,19 +328,19 @@ static int configure_dhcp_server(void)
 		ret = func(__VA_ARGS__); \
 		if (ret) { \
 			LOG_ERR("Failed to configure %s", #func); \
-			return -1; \
+			return; \
```
krish2718, line 331: This depends on the caller, and I would recommend making the caller return the error always rather than void.

samples/wifi/softap/src/wpas_events.c
krish2718: Well, that would mean that another PR will just revert all these changes? I would suggest fixing this sample in another PR with a simple sleep + TODO comment and then use this PR to properly implement WPA supplicant events for all samples. WDYT?

samples/wifi/softap/src/wpas_events.c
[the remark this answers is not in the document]
krish2718: Agreed, ready means iface is added, but what about the rest i.e., iface removed or not ready event? At the very least

* for a single VIF, implement READY and NOT_READY
* for multi-VIF, implement IFACE ADDED/REMOVED and do a iface comparison

sachinthegreen (commented):
With https://github.com/nrfconnect/sdk-nrf/pull/16046 now raised is this PR needed any longer ?

D-Triveni:
> With #16046 now raised is this PR needed any longer ?

No, it's not needed. Will close

D-Triveni:
Changes are handled in https://github.com/nrfconnect/sdk-nrf/pull/16046.
Hence this is not needed.

PR #15054, opened 2024-04-24
PSA Crypto client support for nRF54H20 devices

frkv:
This pull-request includes changes from @osaether for a core-less build of PSA crypto.

It continues the effort for PSA crypto client support which is explained in one of the commits:

```
Note: PSA Crypto client support means that MBEDTLS_PSA_CRYPTO_CLIENT is 
set while MBEDTLS_PSA_CRYPTO_C is not set. This is used in the non-secure
image when TF-M is enabled. The effect of setting up the build to use
PSA Crypto client support is that context types including operations will
include a simple integer "handle" that corresponds to the ongoing 
multi-part crypto operation instead of expecting the full description
of the operation and any enabled PSA crypto drivers
```

The changes int his pull-request adds better client support and resolves a few regressions with regards to psa_util which is used by TLS/DTLS and X.509 APIs to convert error-codes, key types etc. between PSA and the legacy crypto API equivalence. 

**Note that currently TLS/DTLS and X.509 APIs are reliant on legacy crypto types**


frkv:
> @frkv : FYI, if the bots are to be believed, this increases RAM usage for non-54H platforms by 10kB.
> 
> [#15054 (comment)](https://github.com/nrfconnect/sdk-nrf/pull/15054#issuecomment-2074850946)

Yep. This is likely due to using more PSA crypto APIs (MBEDTLS_USE_PSA_CRYPTO) as well as the fact that there isn't a likely way to disable legacy features without getting problems with check_config.h

This is improved in Mbed TLS 3.6.0, where they add meta-constructs to state that you have a feature without pointing directly to the legacy configurations

Example: CCM support is defined here (MBEDTLS_SSL_HAVE_CCM):
https://github.com/Mbed-TLS/mbedtls/blob/771fd7d1dcd19a157812dde9766bb7781e60d075/include/mbedtls/config_adjust_legacy_crypto.h#L452

Used here
https://github.com/Mbed-TLS/mbedtls/blob/771fd7d1dcd19a157812dde9766bb7781e60d075/include/mbedtls/check_config.h#L958

Unfortunately the 3.5.2 version (which we are using) doesn't have this, and expect legacy features/configurations
https://github.com/Mbed-TLS/mbedtls/blob/daca7a3979c22da155ec9dce49ab1abf3b65d3a9/include/mbedtls/check_config.h#L1031

mswarowsky (approved):

Vge0rge (approved):

anhmolt (approved):

PR #15062, opened 2024-04-24
doc: file_suffix: update for PM support

greg-fer:
Updated documentation about file suffixes for added support for Partition Manager's static partition files.

### DIFF
--- a/scripts/partition_manager/partition_manager.rst
+++ b/scripts/partition_manager/partition_manager.rst
@@ -772,26 +772,25 @@ This file is similar to the regular :file:`pm.yml` configuration files, except t
 
 You can set ``PM_STATIC_YML_FILE`` to contain exactly the static configuration you want to use.
 
-.. important::
-    |file_suffix_partition_manager_exception|
-
 If you do not set ``PM_STATIC_YML_FILE``, the build system will use the following order to look for files in your application source directory to use as a static configuration layout:
 
-* If a :term:`build type` is used, the following order applies:
+* If a :ref:`file suffix <app_build_file_suffixes>` is used, the following order applies:
 
-  1. If the file :file:`pm_static_<board>_<revision>_<buildtype>.yml` exists, it will be used.
-  #. Otherwise, if the file :file:`pm_static_<board>_<buildtype>.yml` exists, it will be used.
-  #. Otherwise, if the file :file:`pm_static_<buildtype>.yml` exists, it will be used.
+  1. If the file :file:`pm_static_<board>_<revision>_<suffix>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<board>_<revision>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<board>_<suffix>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<board>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<suffix>.yml` exists, it will be used.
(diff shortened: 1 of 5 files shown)


scripts/partition_manager/partition_manager.rst
```diff
@@ -772,19 +772,20 @@ This file is similar to the regular :file:`pm.yml` configuration files, except t
...
-* If a :term:`build type` is used, the following order applies:
+* If a :ref:`file suffix <app_build_file_suffixes>` is used, the following order applies:
 
-  1. If the file :file:`pm_static_<board>_<revision>_<buildtype>.yml` exists, it will be used.
-  #. Otherwise, if the file :file:`pm_static_<board>_<buildtype>.yml` exists, it will be used.
-  #. Otherwise, if the file :file:`pm_static_<buildtype>.yml` exists, it will be used.
+  1. If the file :file:`pm_static_<board>_<revision>_<suffix>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<board>_<suffix>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<suffix>.yml` exists, it will be used.
   #. Otherwise, if the file :file:`pm_static.yml` exists, it will be used.
```
nordicjm, line 782: this has fallback, will need to manually check it works as expected but it should be this order:

- pm_static_<board>_<revision>_<suffix>.yml
- pm_static_<board>_<suffix>.yml
- pm_static_<suffix>.yml
- pm_static_<board>_<revision>.yml
- pm_static_<board>.yml
- pm_static.yml


Though now I suspect it might actually be:

- pm_static_<board>_<revision>_<suffix>.yml
- pm_static_<board>_<revision>.yml
- pm_static_<board>_<suffix>.yml
- pm_static_<board>.yml
- pm_static_<suffix>.yml
- pm_static.yml

nordicjm: Have updated the list, also note that when hwmv2 is merged this will need updating to say board target

shanthanordic (approved):

scripts/partition_manager/partition_manager.rst
```diff
@@ -772,19 +772,22 @@ This file is similar to the regular :file:`pm.yml` configuration files, except t
...
+  #. Otherwise, if the file :file:`pm_static_<board>_<revision>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<board>_<suffix>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<board>.yml` exists, it will be used.
+  #. Otherwise, if the file :file:`pm_static_<suffix>.yml` exists, it will be used.
   #. Otherwise, if the file :file:`pm_static.yml` exists, it will be used.
 
-* If a build type is not used, then the same order as above applies, except that *<buildtype>* is not part of the file name:
+  .. note::
+      If your application uses :ref:`multiple images <ug_multi_image>`, you need to use :ref:`build types <modifying_build_types>` instead of file suffixes.
+      Check the application and sample documentation pages.
```
nordicjm, line 788: This is not true, it's how the configuration is applied to other images that is different in child/parent vs sysbuild, so the PM file only uses the suffix if it is set as `-DFILE_SUFFIX=x` which is main application only in child/parent, it will not if you do `-Dmcuboot_FILE_SUFFIX=x` however in sysbuild it will apply if doing `-DFILE_SUFFIX=x` which applies to **all** images or if you just do it for the main image only e.g. `-D<app>_FILE_SUFFIX=x`

nordicjm (approved):

PR #15066, opened 2024-04-24
samples: multicore: hello_world: Align sample configuration

adamkondraciuk:
Align sample for testing purposes.

### DIFF
--- a/samples/multicore/hello_world/power_management_app.overlay
+++ b/samples/multicore/hello_world/power_management_app.overlay
@@ -4,21 +4,22 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 
- /delete-node/ &cpuppr_ram3x_region;
- /delete-node/ &cpuapp_cpuppr_ipc;
+/ {
+	power-states {
+		idle: idle {
+			compatible = "zephyr,power-state";
+			power-state-name = "suspend-to-idle";
+			min-residency-us = <100000>;
+		};
 
- / {
-         reserved-memory {
-                 cpuppr_ram3x_region: memory@2fc00000 {
-                         compatible = "nordic,owned-memory";
-                         reg = < 0x2fc00000 0x10000 >;
-                         status = "okay";
-                         perm-read;
-                         perm-write;
-                         perm-execute;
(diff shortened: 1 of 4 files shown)


coderbyheart:
Hell no!

hubertmis:
Why are there two commits, each of them changing sdk-zephyr SHA to different revisions?

Do I understand correctly the overlays are a temporary solution, but later the power management nodes will be moved to soc's dts files?

samples/multicore/hello_world/power_management_app.overlay
```diff
@@ -1,24 +1,25 @@
 /*
  * Copyright (c) 2024 Nordic Semiconductor ASA
  *
- * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ * SPDX-License-Identifier: Apache-2.0
```
zycz, line 4: Why d we change license here? Isn't it still part of sdk-nrf and we should use Nordic license anyway

samples/multicore/hello_world/power_management_app.overlay
```diff
@@ -1,24 +1,25 @@
...
-                         perm-read;
-                         perm-write;
-                         perm-execute;
-                         #address-cells = < 0x1 >;
-                         #size-cells = < 0x1 >;
-                         ranges = < 0x0 0x2fc00000 0x10000 >;
-                 };
-         };
- };
+		s2ram: s2ram {
```
zycz, line 15: The name might be a bit unfortunate. If someone doesn't know it means suspend to ram he might think it is some strange sram node. Maybe sus2ram to differentiate it from sram?

zycz (approved):

samples/multicore/hello_world/power_management_app.overlay
hubertmis: S2RAM is a common abbreviation used all over computers (ok, mostly in Linux context :) ). I've never seen sus2ram.

gmarull (changes requested):
power states should be defined at soc dts files, please, check upstream examples. This sample will also be removed sometime soon (moved upstream), its purpose is not to demo PM.

PR #15070, opened 2024-04-24
samples: matter: added lock credentials manager module

markaj-nordic:
* implemented generic door lock credentials manager with  secure persistent storage support
* refine door lock sample to use the module and manage PIN credentials

TODO:

- [x] add release notes
- [x] update lock sample README
- [x] run all functional tests for door lock

### DIFF
--- a/samples/matter/lock/Kconfig
+++ b/samples/matter/lock/Kconfig
@@ -6,17 +6,44 @@
 
 mainmenu "Matter Lock sample application"
 
-config LOCK_NUM_USERS
-	int "Maximum number of users supported by lock"
+config LOCK_MAX_NUM_USERS
+	int "Maximum number of users supported by the lock"
 	default 10
 
-config LOCK_NUM_CREDENTIALS
-	int "Maximum number of credentials supported by lock"
-	default 20
+config LOCK_MAX_NUM_CREDENTIALS_PER_USER
+	int "Maximum total number of credentials per one user supported by the lock"
+	default 4
 
-config LOCK_NUM_CREDENTIALS_PER_USER
-	int "Maximum number of credentials per user supported by lock"
-	default 3
+config LOCK_MAX_NUM_CREDENTIALS_PER_TYPE
+	int "Maximum number of credentials of given type supported by the lock"
+	default 15
+
+config LOCK_MAX_CREDENTIAL_LENGTH
+	int "Maximum length of the single credential supported by the lock"
+	default 10
+
+config LOCK_ENABLE_DEBUG
+	bool "Enable debug features to print users and credentials"
+	depends on SHELL
+	help
+	  Debug features allows to checking and printing current users, their state, and
+	  associated credentials. Use these with shell commands combinations to read saved credentials
+	  or users, and manage them (add, clear, read).
+
+if LOCK_ENABLE_DEBUG
+
+# This should be deterined based on the maximum possible command length that depends on the LOCK_MAX_CREDENTIAL_LENGTH
+config SHELL_CMD_BUFF_SIZE
+	default 300
+
+endif
+
+config LOCK_LEAVE_FABRIC_CLEAR_CREDENTIAL
+	bool "Clear all credentials assigned to given fabric index"
+	default y
+	help
+	  This feature allows to remove all credentials which were created or modified
+	  within the fabric which is being removed from the door lock.
 
 config STATE_LEDS
 	bool "Use LEDs to indicate the device state"
@@ -78,6 +105,15 @@ config NRF_WIFI_LOW_POWER
 	default y
 
 endif # CHIP_WIFI
+
+config NCS_SAMPLE_MATTER_PERSISTENT_STORAGE
+	default y
+
+config NCS_SAMPLE_MATTER_SECURE_STORAGE_BACKEND
+	default y if !CHIP_WIFI
+
+config NCS_SAMPLE_MATTER_SETTINGS_STORAGE_BACKEND
+	default n if !CHIP_WIFI
 
 source "${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/config/nrfconnect/chip-module/Kconfig.features"
 source "${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/config/nrfconnect/chip-module/Kconfig.defaults"
--- a/samples/matter/lock/README.rst
+++ b/samples/matter/lock/README.rst
@@ -73,6 +73,27 @@ For details, see the `Commissioning the device`_ section.
 
 .. matter_door_lock_sample_remote_testing_end
 
+Door lock credentials
+=====================
+
+By default, the application supports only PIN code credentials, but it is possible to implement support for other door lock credential types by using the ``CredentialsManager`` module.
+The credentials can be used to control remote access to the bolt lock.
+The PIN code assigned by the Matter controller is stored persistently, which means that it can survive a device reboot.
+Depending on the IPv6 network technology in use, the following storage backends are supported by default to store the PIN code credential:
+
+* Matter over Thread - secure storage backend (:kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_SECURE_STORAGE_BACKEND` Kconfig option enabled by default).
+* Matter over Wi-Fi - non-secure storage backend (:kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_SETTINGS_STORAGE_BACKEND` Kconfig option enabled by default).
+
+You can learn more about the |NCS| Matter persistent storage module and its configuration in the :ref:`ug_matter_persistent_storage` section of the :ref:`ug_matter_device_advanced_kconfigs` documentation.
+
+The application supports multiple door lock users and PIN code credentials.
+The following Kconfig options control the limits of the users and credentials that can be added to the door lock:
+
+* :kconfig:option:`CONFIG_LOCK_MAX_NUM_USERS` - Maximum number of users supported by the door lock.
+* :kconfig:option:`CONFIG_LOCK_MAX_NUM_CREDENTIALS_PER_USER` - Maximum number of credentials that can be assigned to one user.
+* :kconfig:option:`CONFIG_LOCK_MAX_NUM_CREDENTIALS_PER_TYPE` - Maximum number of credentials in total.
+* :kconfig:option:`CONFIG_LOCK_MAX_CREDENTIAL_LENGTH` - Maximum length of a single credential in bytes.
+
 .. _matter_lock_sample_wifi_thread_switching:
 
 Thread and Wi-Fi switching
@@ -459,6 +480,61 @@ Upgrading the device firmware
 =============================
 
 To upgrade the device firmware, complete the steps listed for the selected method in the :doc:`matter:nrfconnect_examples_software_update` tutorial of the Matter documentation.
+
+.. _matter_lock_sample_remote_access_with_pin:
+
+Testing remote access with PIN code credential
+==============================================
+
+.. note::
+   You can test the PIN code credential support with any Matter compatible controller.
+   The following steps use the CHIP Tool controller as an example.
+   For more information about setting user credentials, see the Saving users and credentials on door lock devices section of the :doc:`matter:chip_tool_guide` page in the :doc:`Matter documentation set <matter:index>`.
+
+After building the sample and programming it to your development kit, complete the following steps to test remote access with PIN code credential:
+
+1. |connect_kit|
+#. |connect_terminal_ANSI|
+#. Wait until the device boots.
+#. Commission an accessory with node ID equal to 10 to the Matter network by following the steps described in the `Commissioning the device`_ section.
+#. Make the door lock require a PIN code for remote operations:
+
+   .. code-block:: console
+
+      ./chip-tool doorlock write require-pinfor-remote-operation 1 10 1 --timedInteractionTimeoutMs 5000
+
+#. Add the example ``Home`` door lock user:
+
+   .. code-block:: console
+
+      ./chip-tool doorlock set-user 0 2 Home 123 1 0 0 10 1 --timedInteractionTimeoutMs 5000
+
+   This command creates a ``Home`` user with a unique ID of ``123`` and an index of ``2``.
+   The new user's status is set to ``1``, and both its type and credential rule to ``0``.
+   The user is assigned to the door lock cluster residing on endpoint ``1`` of the node with ID ``10``.
+
+#. Add the example ``12345678`` PIN code credential to the ``Home`` user:
+
+   .. code-block:: console
+
+      ./chip-tool doorlock set-credential 0 '{"credentialType": 1, "credentialIndex": 1}' 12345678 2 null null 10 1 --timedInteractionTimeoutMs 5000
+
+#. Unlock the door lock with the given PIN code:
+
+   .. code-block:: console
+
+      ./chip-tool doorlock unlock-door 10 1 --PINCode 12345678 --timedInteractionTimeoutMs 5000
+
+#. Reboot the device.
+#. Wait until the device it rebooted and attached back to the Matter network.
+#. Unlock the door lock with the PIN code provided before the reboot:
+
+   .. code-block:: console
+
+      ./chip-tool doorlock unlock-door 10 1 --PINCode 12345678 --timedInteractionTimeoutMs 5000
+
+.. note::
+   Accessing the door lock remotely without a valid PIN code credential will fail.
 
 .. _matter_lock_sample_switching_thread_wifi:
 
--- a/samples/matter/lock/src/bolt_lock_manager.cpp
+++ b/samples/matter/lock/src/bolt_lock_manager.cpp
@@ -20,19 +20,15 @@ void BoltLockManager::Init(StateChangeCallback callback)
 	k_timer_init(&mActuatorTimer, &BoltLockManager::ActuatorTimerEventHandler, nullptr);
 	k_timer_user_data_set(&mActuatorTimer, this);
 
+	PinManager::Instance().Init();
+
 	/* Set the default state */
 	Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED2).Set(IsLocked());
 }
 
-bool BoltLockManager::GetUser(uint16_t userIndex, EmberAfPluginDoorLockUserInfo &user) const
+bool BoltLockManager::GetUser(uint16_t userIndex, EmberAfPluginDoorLockUserInfo &user)
 {
-	/* userIndex is guaranteed by the caller to be between 1 and CONFIG_LOCK_NUM_USERS */
-	user = mUsers[userIndex - 1];
-
-	ChipLogProgress(Zcl, "Getting lock user %u: %s", static_cast<unsigned>(userIndex),
-			user.userStatus == UserStatusEnum::kAvailable ? "available" : "occupied");
-
-	return true;
+	return PinManager::Instance().GetUserInfo(userIndex, user);
 }
 
 bool BoltLockManager::SetUser(uint16_t userIndex, FabricIndex creator, FabricIndex modifier, const CharSpan &userName,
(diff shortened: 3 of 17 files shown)


samples/matter/lock/README.rst
```diff
@@ -460,6 +480,66 @@ Upgrading the device firmware
...
+      ./chip-tool doorlock set-credential 0 '{"credentialType": 1, "credentialIndex": 1}' 12345678 1 null null 1 1 --timedInteractionTimeoutMs 5000
+
+#. Unlock the door lock with the given PIN code:
+
+   .. code-block:: console
+
+      ./chip-tool doorlock unlock-door 1 1 --PINCode 12345678 --timedInteractionTimeoutMs 5000
+
+#. Reboot the device.
+#. Wait until the device it rebooted and attached back the the Matter network.
```
wiba-nordic, line 533:
```suggestion
#. Wait until the device it rebooted and attached back to the Matter network.
```

samples/matter/lock/README.rst
```diff
@@ -73,6 +73,26 @@ For details, see the `Commissioning the device`_ section.
...
+
+You can learn more about the |NCS| Matter persistent storage module and its configuration in the :ref:`ug_matter_persistent_storage` section of the :ref:`ug_matter_device_advanced_kconfigs` documentation.
+
+The application supports multiple door lock users and PIN code credentials.
+The following Kconfig options control the limits of the users and credentials that can be added to the door lock:
+
+* :kconfig:option:`CONFIG_LOCK_MAX_NUM_USERS` - Maximum number of users supported by the door lock.
+* :kconfig:option:`CONFIG_LOCK_NUM_CREDENTIALS_PER_USER` - Maximum number of credentials that can be assigned per one user.
+* :kconfig:option:`CONFIG_LOCK_NUM_CREDENTIALS_TYPE` - Maximum number of credentials in total.
+* :kconfig:option:`CONFIG_LOCK_MAX_CREDENTIAL_LENGTH` - Maximum length of the single credential.
```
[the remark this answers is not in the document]
markaj-nordic: It's in bytes, I will add this info.

samples/matter/lock/README.rst
```diff
@@ -460,6 +480,66 @@ Upgrading the device firmware
 
 To upgrade the device firmware, complete the steps listed for the selected method in the :doc:`matter:nrfconnect_examples_software_update` tutorial of the Matter documentation.
 
+.. _matter_lock_sample_remote_access_with_pin:
+
+Testing remote access with PIN code credential
+==============================================
+
+.. note::
+   You can test the PIN code credential support with any Matter compatible controller, however in the following steps the `CHIP Tool` controller is used as an example.
```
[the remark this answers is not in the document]
markaj-nordic: `steps us` -> `steps use` I suppose :)

samples/matter/lock/README.rst
```diff
@@ -460,6 +481,56 @@ Upgrading the device firmware
...
+#. |connect_terminal_ANSI|
+#. Wait until the device boots.
+#. Commission an accessory to the Matter network by following the steps described in the `Commissioning the device`_ section.
+#. Make the door lock require a PIN code for remote operations:
+
+   .. code-block:: console
+
+      ./chip-tool doorlock write require-pinfor-remote-operation 1 1 1 --timedInteractionTimeoutMs 5000
+
+#. Add the example ``Home`` door lock user with the following CHIP Tool command:
```
wiba-nordic, line 505:
```suggestion
#. Add the example ``Home`` door lock user:
```
Just noticed this is the only one that specifies CHIP Tool in the steps, so might as well take it out.

wiba-nordic (changes requested):

samples/matter/lock/README.rst
```diff
@@ -460,6 +480,66 @@ Upgrading the device firmware
...
+#. |connect_terminal_ANSI|
+#. Program the application to the kit using the following command:
+
+   .. code-block:: console
+
+      west flash --erase
+
+#. Wait until the device boots.
+#. Commission an accessory to the Matter network with steps described in the `Commissioning the device`_ section.
+#. Enforce the door lock to require PIN code for remote operations:
```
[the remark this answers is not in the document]
wiba-nordic: Yes, yes it should. And this is why you should always double-check what us tech writers ask you to put in there :D

samples/matter/lock/README.rst
```diff
@@ -460,6 +481,56 @@ Upgrading the device firmware
...
+
+   .. code-block:: console
+
+      ./chip-tool doorlock write require-pinfor-remote-operation 1 1 1 --timedInteractionTimeoutMs 5000
+
+#. Add the example ``Home`` door lock user with the following CHIP Tool command:
+
+   .. code-block:: console
+
+      ./chip-tool doorlock set-user 0 1 Home 0 1 0 0 1 1 --timedInteractionTimeoutMs 5000
```
kkasperczyk-no, line 509: I know that user may read the command description, but arguments like '0 1 0 1 0 0 1 1' look a little bit confusing :)

samples/matter/lock/src/bolt_lock_manager.cpp
```diff
@@ -20,118 +20,52 @@ void BoltLockManager::Init(StateChangeCallback callback)
...
-
-	ChipLogDetail(Zcl, "Invalid lock PIN code provided");
-	err = OperationErrorEnum::kInvalidCredential;
+	return PinManager::Instance().ValidatePIN(pinCode, err);
+}
 
-	return false;
+void BoltLockManager::SetRequirePIN(bool require)
+{
+	return PinManager::Instance().SetRequirePIN(require);
```
kkasperczyk-no, line 64:
```suggestion
	PinManager::Instance().SetRequirePIN(require);
```
It returns void anyway.

samples/matter/lock/src/credentials/credentials_data_types.h
```diff
@@ -0,0 +1,424 @@
...
+	Secret mSecrets;
+
+	/**
+	 * @brief Get required size for single credential entry.
+	 *
+	 * This method can be used to prepare a proper serialization buffer.
+	 *
+	 * @return size of single credential entry.
+	 */
+	constexpr static size_t RequiredBufferSize() { return sizeof(Info) + sizeof(Secret); }
```
kkasperczyk-no, line 68: Does it have to be a method? Why not just constexpr field?

samples/matter/lock/src/credentials/credentials_data_types.h
```diff
@@ -0,0 +1,424 @@
...
+	};
+	static_assert(sizeof(Info) == 6);
+
+	struct Secret {
+		size_t mDataLength;
+		uint8_t mData[kMaxCredentialLength];
+	};
+
+	Info mInfo;
+	Secret mSecrets;
```
kkasperczyk-no, line 59: Super minor - as it's not a list or array it should be singular?
```suggestion
	Secret mSecret;
```

samples/matter/lock/src/credentials/credentials_data_types.h
```diff
@@ -0,0 +1,424 @@
...
+	Name mName;
+
+	/**
+	 * @brief Get required size for single credential entry.
+	 *
+	 * This method can be used to prepare a proper serialization buffer.
+	 *
+	 * @return size of single credential entry.
+	 */
+	constexpr static size_t RequiredBufferSize() { return sizeof(Info) + sizeof(Credentials) + sizeof(Name); }
```
kkasperczyk-no, line 158: Ditto

samples/matter/lock/Kconfig
```diff
@@ -6,17 +6,42 @@
...
+	bool "Enable debug features to print users and credentials"
+	help
+	  Debug features allows to checking and printing current users, their state, and
+	  associated credentials. Use these with shell commands combinations to read saved credentials
+	  or users, and manage them (add, clear, read).
+
+if LOCK_ENABLE_DEBUG
+
+config SHELL_CMD_BUFF_SIZE
+	default 396
```
kkasperczyk-no, line 35: Just out of my curiosity, why 396? Maybe let's add some comment to not wonder about it after some time.

samples/matter/lock/README.rst
```diff
@@ -73,6 +73,27 @@ For details, see the `Commissioning the device`_ section.
 
 .. matter_door_lock_sample_remote_testing_end
 
+Door lock credentials
+=====================
+
+The application supports PIN code credentials.
```
kkasperczyk-no, line 79: Are we able to support other types of credentials or only PIN codes? I mean, I know that application uses pin codes, but the credentials module is able to support other types as well, right? If so, maybe we should write it somewhere that we present only pin codes in our app, but it could be also changed to use something different.

kkasperczyk-no (approved):
I left some comments to consider, but generally it looks good to me!

samples/matter/lock/Kconfig
ArekBalysNordic: Probably I set this value a long time ago, and it describes the maximum size of the very specific credential type. :)

samples/matter/lock/src/bolt_lock_manager.cpp
```diff
@@ -20,118 +20,52 @@ void BoltLockManager::Init(StateChangeCallback callback)
...
 	k_timer_user_data_set(&mActuatorTimer, this);
 
+	PinManager::Instance().Init();
+
 	/* Set the default state */
 	Nrf::GetBoard().GetLED(Nrf::DeviceLeds::LED2).Set(IsLocked());
 }
 
-bool BoltLockManager::GetUser(uint16_t userIndex, EmberAfPluginDoorLockUserInfo &user) const
+bool BoltLockManager::GetUser(uint16_t userIndex, EmberAfPluginDoorLockUserInfo &user)
```
ArekBalysNordic, line 29: Regarding all these changes here, can we remove BoltLockManager methods and use Credentials Manager's ones directly? I can see that most of the methods just call almost the same API from Credential Manager.

ArekBalysNordic (approved):
Two comments to consider, overall LGTM!

markaj-nordic:
Additional functional tests for the PIN code and its persistence passed. Thanks @doublemis1.

samples/matter/lock/README.rst
markaj-nordic: I did it on purpose to indicate that it's the chip-tool controller being used as an example. But that's fair to remove this part to be consistent. We mention above that the chip-tool will be used anyway.

samples/matter/lock/Kconfig
```diff
@@ -6,17 +6,42 @@
...
-	int "Maximum number of credentials supported by lock"
-	default 20
-
 config LOCK_NUM_CREDENTIALS_PER_USER
-	int "Maximum number of credentials per user supported by lock"
-	default 3
+	int "Maximum total number of credentials per one user supported by lock"
+	default 4
+
+config LOCK_NUM_CREDENTIALS_TYPE
```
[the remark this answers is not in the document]
Damian-Nordic: Also, `...CREDENTIALS_PER_TYPE` would be more meaningful.
And some prompts have "supported by lock" and some are not - another thing that can be unified.

samples/matter/lock/Kconfig
```diff
@@ -6,17 +6,42 @@
...
+	int "Maximum number of credentials of given type"
+	default 15
+
+config LOCK_MAX_CREDENTIAL_LENGTH
+	int "Maximum length of the single credential"
+	default 10
+
+config LOCK_ENABLE_DEBUG
+	bool "Enable debug features to print users and credentials"
```
Damian-Nordic, line 26: Should it then depend on `SHELL`?

samples/matter/lock/Kconfig
```diff
@@ -79,7 +104,17 @@ config NRF_WIFI_LOW_POWER
...
+config NCS_SAMPLE_MATTER_SECURE_STORAGE_BACKEND
+	default y if !CHIP_WIFI
+
+config NCS_SAMPLE_MATTER_SETTINGS_STORAGE_BACKEND
+	default n if !CHIP_WIFI
+
 source "${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/config/nrfconnect/chip-module/Kconfig.features"
 source "${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/config/nrfconnect/chip-module/Kconfig.defaults"
 source "${ZEPHYR_NRF_MODULE_DIR}/samples/matter/common/src/Kconfig"
+source "${ZEPHYR_NRF_MODULE_DIR}/samples/matter/common/src/persistent_storage/Kconfig"
```
Damian-Nordic, line 119: Why isn't this sourced from `common/src/Kconfig`?

samples/matter/lock/src/credentials/credentials_data_types.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+size_t Credential::Serialize(void *buff, size_t buffSize)
+{
+	if (!buff || buffSize < RequiredBufferSize() || mSecrets.mDataLength > RequiredBufferSize()) {
+		return 0;
+	}
+	size_t offset = 0;
+
+	pack(buff, mInfo.mRaw, sizeof(mInfo.mRaw), offset);
```
Damian-Nordic, line 62: fyi, Matter has pretty cool `BufferWriter` and `BufferReader` classes that take care of all the buffer size checks.

samples/matter/lock/src/credentials/credentials_data_types.cpp
```diff
@@ -0,0 +1,179 @@
...
+void unpack(void *data, size_t dataSize, const void *buff, size_t &offset)
+{
+	memcpy(data, reinterpret_cast<const uint8_t *>(buff) + offset, dataSize);
+	offset += dataSize;
+}
+
+CHIP_ERROR Credential::FillFromPlugin(EmberAfPluginDoorLockCredentialInfo &credentialInfo)
+{
+	if (credentialInfo.credentialData.size() < kMaxCredentialLength) {
+		mInfo.mFields.mStatus = static_cast<uint8_t>(credentialInfo.status);
```
Damian-Nordic, line 27: Is there a reason we don't want to use Matter enums and cast things to `uint8_t`?

samples/matter/lock/src/credentials/credentials_data_types.h
```diff
@@ -0,0 +1,424 @@
...
+};
+
+template <uint16_t N> inline size_t IndexList<N>::Serialize(void *buff, size_t buffSize)
+{
+	if (!buff || buffSize < RequiredBufferSize()) {
+		return 0;
+	}
+
+	size_t offset = 0;
+	pack(buff, &mList.mLength, sizeof(mList.mLength), offset);
```
Damian-Nordic, line 265: Would it make sense to pack only `sizeof(mList.mIndexes[0]) * mList.mLength` instead of always serialize the full array?
Also, the length doesn't have to be serialized as this is stored separately by the storage subsystem, right?

Damian-Nordic (commented):
partial review

samples/matter/lock/src/bolt_lock_manager.cpp
markaj-nordic: Yes, but I guess it's a good practice to forward the function call by return because the return types of the caller and callee are in line. If the signature of the SetRequirePIN() changes and it suddenly starts returning non-void the compiler will react.

wiba-nordic (approved):
Docs side looks good.

samples/matter/lock/README.rst
markaj-nordic: Ok, I will add some explanation with cooperation with @wiba-nordic .

samples/matter/lock/src/credentials/credentials_data_types.cpp
ArekBalysNordic: Yes, I agree. We did this way because we didn't want to use Matter functions there, but I guess it will be nicer to use BufferedWriter and BufferedReader. :)

samples/matter/lock/src/credentials/credentials_data_types.h
markaj-nordic: I does not change much for the compiler but it's more flexible as we may for instance introduce some logic when calculating the resulting size in the future.

samples/matter/lock/src/credentials/credentials_data_types.h
markaj-nordic: As above.

samples/matter/lock/Kconfig
markaj-nordic: It was more specific to the project Fell and dependent on the LOCK_MAX_CREDENTIAL_LENGTH. I guess we may safely decrease it for the purposes of the lock app having in mind that we have LOCK_MAX_CREDENTIAL_LENGTH equal to 10.

samples/matter/lock/README.rst
markaj-nordic: Sure, good point.

samples/matter/lock/src/bolt_lock_manager.cpp
markaj-nordic: This is how I wanted to do it initially, bu then I though it would be more coherent if the `BoltLockManager` was the only direct user of the `CredentialsManager` and, as a result, only BoltLockManager was called in the zcl_callbacks.cpp. I will re-think it though, thanks.

markaj-nordic:
Resolved remarks from @kkasperczyk-no.

samples/matter/lock/README.rst
```diff
@@ -73,6 +73,27 @@ For details, see the `Commissioning the device`_ section.
 
 .. matter_door_lock_sample_remote_testing_end
 
+Door lock credentials
+=====================
+By default the application supports only PIN code credentials.
+Additionally, it is possible to implement support for other door lock credential types with the use of ``CredentialsManager`` module.
```
wiba-nordic, line 79:
```suggestion
=====================

By default, the application supports only PIN code credentials, but it is possible to implement support for other door lock credential types by using the ``CredentialsManager`` module.
```
Should the rest of the description in this section be made more generic? Alternatively, I can try to reshuffle the first few sentences so that it's clear that the rest of the information is specifically about the default implementation.

wiba-nordic (changes requested):

samples/matter/lock/src/credentials/credentials_data_types.cpp
Damian-Nordic: ok, if you're just merging some preexisting code, feel free to ignore such nits. :) This can be cleaned up later.

samples/matter/lock/src/credentials/credentials_data_types.h
```diff
@@ -0,0 +1,424 @@
...
+private:
+	static constexpr uint8_t RequestedNumOfCredTypesSupported()
+	{
+		uint8_t cnt{ 0 };
+		for (uint8_t i = 0; i < (sizeof(uint8_t) * CHAR_BIT); i++) {
+			if (CRED_BIT_MASK & BIT(i)) {
+				cnt++;
+			}
+		}
+		return cnt;
```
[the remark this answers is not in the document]
markaj-nordic: Seems that's the one I was looking for (but in Zephyr) :)

samples/matter/lock/src/credentials/credentials_data_types.cpp
markaj-nordic: Thanks for the hint, for now let's finalize this PR. I will file a ticket to use Matter utils.

samples/matter/lock/src/credentials/credentials_data_types.cpp
markaj-nordic: I guess we wanted to make the structure more suitable for the serialization, but in fact all underlying enums are inheriting from `uint8_t` anyway and also pack/unpack accept `void*`, so I believe this could be refactored. I hope it's fine to do it as a follow up.

markaj-nordic: Task created to refactor credentials_data_types.

samples/matter/lock/src/bolt_lock_manager.cpp
markaj-nordic: @ArekBalysNordic, @Damian-Nordic do you think my approach is ok?

Damian-Nordic: Both options are fine with me. I agree hiding credentials management may be clearer for existing users but on the other hand less code is always a good thing. Can't make up my mind really :)

samples/matter/lock/src/credentials/credentials_data_types.h
Damian-Nordic: The second part means that you don't have to store the number of items explicitly. When you load this data, you will get the BLOB size, which you can divide by the item size and get the number of items that way. Not sure if that's OK from backwards compatibility but I was just pointing out that storing the length seems redundant.

wiba-nordic (changes requested):

samples/matter/lock/src/bolt_lock_manager.cpp
markaj-nordic: I'll leave as it is then :)

markaj-nordic:
@wiba-nordic, @Damian-Nordic please revisit, hopefully the last time.

samples/matter/lock/README.rst
```diff
@@ -73,6 +73,28 @@ For details, see the `Commissioning the device`_ section.
 
 .. matter_door_lock_sample_remote_testing_end
 
+Door lock credentials
+=====================
+
+By default, the application supports only PIN code credentials, but it is possible to implement support for other door lock credential types by using the ``CredentialsManager`` module.
+Additionally, it is possible to implement support for other door lock credential types with the use of ``CredentialsManager`` module.
```
wiba-nordic, line 80:
```suggestion
```

wiba-nordic (changes requested):

samples/matter/lock/src/credentials/credentials_data_types.cpp
```diff
@@ -0,0 +1,179 @@
...
+	return offset;
+}
+
+CHIP_ERROR Credential::Deserialize(const void *buff, size_t buffSize)
+{
+	if (!buff) {
+		return CHIP_ERROR_INVALID_ARGUMENT;
+	}
+
+	if (buffSize > RequiredBufferSize()) {
```
Damian-Nordic, line 75: nit: you should probably also check if buffSize is not less than `sizeof(mInfo.mRaw) + sizeof(mSecret.mDataLength)`, but this will be solved automatically after switching to `BufferReader`.

samples/matter/lock/src/credentials/credentials_data_types.h
```diff
@@ -0,0 +1,417 @@
...
+	static constexpr size_t kElementSize{ sizeof(mList.mIndexes[0]) };
+
+	/**
+	 * @brief Get required size for single IndexList entry.
+	 *
+	 * This method can be used to prepare a proper serialization buffer.
+	 *
+	 * @return size of single credential entry.
+	 */
+	constexpr static size_t RequiredBufferSize() { return sizeof(List); }
```
Damian-Nordic, line 224:
```suggestion
	size_t RequiredBufferSize() { return mList.mLength * sizeof(mList.mIndexes[0]); }
```
but we can also leave the current code for the worst case.

Damian-Nordic (approved):

samples/matter/lock/src/credentials/credentials_data_types.h
markaj-nordic: The thing is we use this constexpr in the application code to allocate buffers for serialization and deserialization. The only way this change would make sense would be if we allocated these buffers on the heap dynamically, I suppose. But that would require more refactoring and frankly I'm not really sure if it's better to to use stack or heap in this case.

samples/matter/lock/src/credentials/credentials_data_types.cpp
markaj-nordic: Yes, that's a valid point but on the other hand the API documentation clearly states that "buffSize` must be set to at least the RequiredBufferSize size." when using Serialize() methods :)

samples/matter/lock/src/credentials/credentials_data_types.h
Damian-Nordic: ok, thanks

samples/matter/lock/src/credentials/credentials_data_types.cpp
markaj-nordic: On the other hand this does not fit into the `buffSize > RequiredBufferSize()` check :) I guess this should be revisited during the refactoring task that is already scheduled.

wiba-nordic (approved):

PR #15083, opened 2024-04-25
nrf_security: Personalization string for Cracen IKG

vili-nordic:
Adds personalization strings for IKG derived keys based on domain ID.


subsys/nrf_security/src/drivers/cracen/cracenpsa/src/common.c
```diff
@@ -27,6 +27,16 @@
 #define NOT_ENABLED_CURVE    (0)
 #define NOT_ENABLED_HASH_ALG (0)
 
+/* TODO: These defines will come from an external header file. */
```
mswarowsky, line 30: Maybe have a Jira else these will live here forever 👯

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/common.c
```diff
@@ -570,7 +580,48 @@ static int cracen_prepare_ik_key(const uint8_t *user_data)
...
+	cfg.device_secret = device_secret;
+	cfg.device_secret_sz = ARRAY_SIZE(device_secret);
+
+	switch (user_data[1]) {
+#define SET_STR(x)                                                                                 \
+	{                                                                                          \
+		const char lstr_##x[((sizeof(x) + 3) / 4) * 4] = #x;                               \
+		cfg.key_bundle = (uint32_t *)lstr_##x;                                             \
+		cfg.key_bundle_sz = sizeof(lstr_##x) / sizeof(uint32_t);                           \
+	}
```
mswarowsky, line 595: Maybe add a short comment on what this does, especially as has all of these nice magic numbers in there.

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/common.c
```diff
@@ -570,7 +583,54 @@ static int cracen_prepare_ik_key(const uint8_t *user_data)
 		nrf_cracen_seedram_lock_enable_set(NRF_CRACEN, true);
 	}
 #endif
-	return sx_pk_ik_derive_keys(NULL);
+	struct sx_pk_config_ik cfg;
```
Vge0rge, line 586: This is only applicable to the nRF54H20 right? Does it make not make sense to guard this based on the SOC?

vili-nordic: done

Vge0rge: I will approve, if you want  also add the ifdef in the definitions of the domains. Just to make it more clear what their scope is.

Vge0rge (approved):

PR #15094, opened 2024-04-25
samples: suit: Add smp_transfer sample

tomchy:
Add basic sample that demonstrates SUIT-based boot and update procedures.

Ref: NCSDK-26629


adsz-nordic (approved):

samples/suit/smp_transfer/Kconfig
```diff
@@ -0,0 +1,31 @@
...
+	default APP_VERSION
+
+config SUIT_SECDOM_UPDATE_LOG
+	bool "Enable Secure Domain logs in update candidate"
+	default n
+
+config SUIT_SECDOM_EXTMEM_SUPPORT
+	bool "Enable external flash support in SDFW"
+
+config SUIT_BUILD_FLASH_COMPANION
```
ahasztag, line 21: Perhaps we should add 'select Experimental' here until we support it in NCS?

robertstypa (approved):

ahasztag (changes requested):

samples/suit/smp_transfer/boards/nrf54h20dk_nrf54h20_cpuapp.overlay
```diff
@@ -0,0 +1,94 @@
...
+/delete-node/ &cpuapp_rw_partitions;
+
+&mram1x {
+	erase-block-size = < 0x10 >;
+	/* Hardcoded inside the soc_flash_nrf_mram.c (MRAM_WORD_SIZE) */
+	write-block-size = < 0x10 >;
+
+	cpuapp_rx_partitions: cpuapp-rx-partitions {
+		compatible = "nordic,owned-partitions", "fixed-partitions";
+		status = "disabled";
```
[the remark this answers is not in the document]
tomchy: Nice catch!

I copied the default partitioning and it is disabled there.

robertstypa (approved):

e-rk (approved):

samples/suit/smp_transfer/Kconfig
```diff
@@ -0,0 +1,14 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config APP_VERSION
+	int "Application version"
+	default 1
```
nordicjm, line 9: This conflicts with an actual application version, which is set by having a `VERSION` file, you can't have a VERSION file because you want to change this per build but can you instead use `SUIT_ENVELOPE_SEQUENCE_NUM` and check against that rather than conflict with something that is already defined?

samples/suit/smp_transfer/sample.yaml
```diff
@@ -0,0 +1,89 @@
...
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+sample:
+  name: SUIT update through SMP sample
+  description: Sample application that updates firmware using SUIT and SMP protocol
+tests:
+  subsys.suit.smp_transfer.v1:
```
nordicjm, line 11: Anything that's common for all tests here should be in a common section

samples/suit/smp_transfer/sysbuild/hci_ipc.conf
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+CONFIG_BT_MAX_CONN=2
```
nordicjm, line 7: are 2 connections really needed?

tomchy: No, I will remove this line.

jnsgsr:
It looks like MPI for radio core is missing in outcome files when building the configuration using twister "subsys.suit.smp_transfer.bt.v1" there is only "suit_mpi_application_merged"

jnsgsr (changes requested):
It looks like MPI for radio core is missing in outcome files when building the configuration using twister "subsys.suit.smp_transfer.bt.v1" there is only "suit_mpi_application_merged" and from the addresses values there we can tell it does not contain data for MPI for radio core.

jnsgsr:
> It looks like MPI for radio core is missing in outcome files when building the configuration using twister "subsys.suit.smp_transfer.bt.v1" there is only "suit_mpi_application_merged" and from the addresses values there we can tell it does not contain data for MPI for radio core.

Nevermind our script to collect outcome files have to be adapted, no issue here

samples/suit/smp_transfer/prj.conf
```diff
@@ -0,0 +1,33 @@
...
+
+# Enable DK LED library
+CONFIG_DK_LIBRARY=y
+
+# Enable mcumgr
+CONFIG_NET_BUF=y
+CONFIG_ZCBOR=y
+CONFIG_ZCBOR_CANONICAL=y
+CONFIG_MCUMGR=y
+CONFIG_MCUMGR_SMP_LEGACY_RC_BEHAVIOUR=y
```
nordicjm, line 15: why is CONFIG_MCUMGR_SMP_LEGACY_RC_BEHAVIOUR enabled?

samples/suit/smp_transfer/sysbuild/hci_ipc.conf
nordicjm: doesn't look to be fixed?

nordicjm:
license check needs exclusion added

tomchy:
> license check needs exclusion added

Reworked to use Zephyr sources

ahasztag (approved):

nordicjm (approved):

PR #15096, opened 2024-04-25
samples: Bluetooth: mesh: Add additional occupancy sensor types

akredalen:
* Adds three new sensor types to the sensor client and sensor server samples. The new sensor types are: Motion Sensing, Time Since Motion Sensed and People Counting.
* Updated docs to reflect the new sensor types.
* Some refactoring in the server model handler to deal with similar sensor types in a more generic way.

- [x] PTS tests passed
- [x] Add to PTS workspace ICS
- [x] Create task for updating QDID


HaavardRei (changes requested):

samples/bluetooth/mesh/sensor_client/README.rst
```diff
@@ -44,11 +44,18 @@ The following Bluetooth Mesh sensor types, and their settings, are used in this
 
   * :c:var:`bt_mesh_sensor_motion_threshold` - Used as a setting for the :c:var:`bt_mesh_sensor_presence_detected` sensor type to set the time (0-10 seconds) before the presence is detected.
 
-* :c:var:`bt_mesh_sensor_present_amb_light_level` - Periodically requested by the client and published by the server according to its publishing period.
+* :c:var:`bt_mesh_sensor_motion_sensed` - Published when a button is pressed on the server.
+* :c:var:`bt_mesh_sensor_time_since_motion_sensed` - Periodically requested by the client and published by the server according to its publishing period.
+
+  * :c:var:`bt_mesh_sensor_motion_threshold` - Used as a setting for the :c:var:`bt_mesh_sensor_motion_sensed` sensor type to set the time (0-10 seconds) before the motion is sensed.
```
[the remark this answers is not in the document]
akredalen: Agreed. This was copy-pasted from the presence sensor documentation part which also has misplaced the `bt_mesh_sensor_motion_threshold`.

samples/bluetooth/mesh/sensor_client/README.rst
```diff
@@ -44,11 +44,18 @@ The following Bluetooth Mesh sensor types, and their settings, are used in this
 
   * :c:var:`bt_mesh_sensor_motion_threshold` - Used as a setting for the :c:var:`bt_mesh_sensor_presence_detected` sensor type to set the time (0-10 seconds) before the presence is detected.
 
-* :c:var:`bt_mesh_sensor_present_amb_light_level` - Periodically requested by the client and published by the server according to its publishing period.
+* :c:var:`bt_mesh_sensor_motion_sensed` - Published when a button is pressed on the server.
```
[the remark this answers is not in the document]
akredalen: Hmm. We mention that buttons are being used before the `User interface` part in the documentation for multiple samples. I think which buttons are used for what might be too detailed for the `Overview`.

samples/bluetooth/mesh/sensor_server/README.rst
```diff
@@ -137,6 +144,15 @@ Button 2:
...
    The motion threshold has five steps from 0 % (representing 0 seconds) to 100 % (representing 10 seconds) separated by 25 %-steps.
 
+Button 3:
+   Simulates motion sensed.
+   For how long the button has to be pressed before the motion is sensed depends on the motion threshold.
+   The motion threshold has five steps from 0 % (representing 0 seconds) to 100 % (representing 10 seconds) separated by 25 %-steps.
+
+Button 4:
+   Simulates different people count sensor values.
+   These dummy values represent raw values coming from a people count sensor.
```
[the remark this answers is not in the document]
akredalen: Right. Actually, it only has one button, so Thingy never supported the button press functionality for _presence detection_ either. I suggest to add a note.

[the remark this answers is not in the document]
ludvigsj: The Thingy53 has two buttons, the second one is tiny next to the normal one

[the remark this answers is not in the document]
akredalen: Aha! I will update note then

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -77,11 +77,15 @@ static struct bt_mesh_sensor_value pres_mot_thres = {
 };
 static struct bt_mesh_sensor_value amb_light_level_ref = ILLUMINANCE_INIT_MILLIS(0);
 static float amb_light_level_gain = 1.0;
-/* Using a dummy ambient light value because we do not have a real ambient light sensor. */
+/* Using dummy sensor values to simulate having a real sensor. */
 static float dummy_ambient_light_value;
+static float dummy_people_count_value;
```
ludvigsj, line 82: It's a bit weird to use `float` for this since this is a count (it's 16 bit unsigned on the air). Suggest to use an integer type instead

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -77,11 +77,15 @@ static struct bt_mesh_sensor_value pres_mot_thres = {
...
 
 static bool pres_detect;
-static uint32_t prev_detect;
+static uint32_t prev_pres_detect;
+
+static bool mot_sensed;
```
ludvigsj, line 87: This is supposed to be a percentage. I guess we can keep the sample simple by cycling between 0 and 100% motion, but currently, it's cycling between 0% and 1% motion, masquerading as a boolean, which I find confusing.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -473,6 +507,45 @@ static struct bt_mesh_sensor time_since_presence_detected = {
...
+		printk("Error encoding motion sensed (%d)", err);
+	}
+	return err;
+};
+
+static struct bt_mesh_sensor motion_sensor = {
+	.type = &bt_mesh_sensor_motion_sensed,
+	.get = motion_sensed_get,
+	.settings = {
+		.list = (const struct bt_mesh_sensor_setting *)&presence_motion_threshold_setting,
```
ludvigsj, line 526: I'm note sure if you are supposed to refer to the same setting for two different sensors like this. Even if it's legal by the spec, I feel it's confusing, since if you set the threshold on the motion sensor, you wouldn't expect the presence sensor setting to change, and vice versa.

ludvigsj: Are we even required to have a setting here at all? Is this something required by the NLC Profile or something you added?

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -44,6 +54,37 @@ static void sensor_cli_data_cb(struct bt_mesh_sensor_cli *cli,
...
 	} else if (sensor->id == bt_mesh_sensor_present_amb_light_level.id) {
 		printk("Ambient light level: %s\n", bt_mesh_sensor_ch_str(value));
+	} else if (sensor->id == bt_mesh_sensor_motion_sensed.id) {
+		int64_t motion_sensed = 0;
+
+		(void)bt_mesh_sensor_value_to_micro(value, &motion_sensed);
+		if (!!motion_sensed) {
+			is_occupied = true;
+			printk("Motion detected. Area is occupied.\n");
+			k_work_cancel_delayable(&motion_timeout_work);
```
ludvigsj, line 64: This looks a bit over-complicated. Why can't you handle this just like we already handle `presence_detected`, without this extra message 1 minute later?

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -215,7 +271,17 @@ static void button_handler_cb(uint32_t pressed, uint32_t changed)
 				      &bt_mesh_sensor_motion_threshold,
 				      &presence_motion_threshold[motion_threshold_idx++]);
 		if (err) {
-			printk("Error setting motion threshold setting (%d)\n", err);
+			printk("Error setting motion threshold setting for Presence Detected "
+			       "sensor (%d)\n", err);
+		}
+		motion_threshold_idx = motion_threshold_idx % ARRAY_SIZE(presence_motion_threshold);
+
+		err = setting_set_int(&bt_mesh_sensor_motion_sensed,
```
ludvigsj, line 279: Since you have the same backing value for both servers on the server side, this is a no-op (since the same state is written by the other setting set above). See my server side comment on this.

ludvigsj (changes requested):

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -709,12 +806,99 @@ static const double dummy_amb_light_values[] = {
...
+		}
+	} else {
+		if (!mot_sensed) {
+			printk("Publishing motion is aborted\n");
+			k_work_cancel_delayable(&motion_sensed_work);
+		} else {
+			int err;
+			struct bt_mesh_sensor_value val;
+
+			(void)bt_mesh_sensor_value_from_micro(
```
alxelax, line 880: Returned status should be handled

alxelax (changes requested):

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -429,23 +452,23 @@ static struct bt_mesh_sensor presence_sensor = {
...
+
+	else if (prev_detect) {
+		int64_t micro_since_detect = (k_uptime_get() - prev_detect) * USEC_PER_MSEC;
 
 		err = bt_mesh_sensor_value_from_micro(format, micro_since_detect, rsp);
+
 		if (err == -ERANGE) {
-			printk("Warning: Time since presence detected (%u) is out of range and was"
+			printk("Warning: Time since detection (%u) is out of range and was"
 			       " clamped to (%s)\n", (uint32_t)(k_uptime_get() - prev_detect),
```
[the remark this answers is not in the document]
ludvigsj: Nope, I already commented on this when this message was added in the first place: `micro_since_detect` will overflow uint32_t. `k_uptime_get() - prev_detect` is _milli_seconds, not micro, and will fit in uint32_t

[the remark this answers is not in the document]
alxelax: This will be different time to which participated to comparison. I guess it should be `(uint32_t)(micro_since_detect / USEC_PER_MSEC)`

samples/bluetooth/mesh/sensor_server/src/model_handler.c
akredalen: I see your point. And no, the NLC profile does not require it. However, I felt like it was odd to have a "motion threshold" setting and not use this for the motion sensor.

ludvigsj: My point is that the setting is logically located under the Presence sensor. It's the `Presence Sensor/Motion Threshold` setting. If you want a setting here, you should define a separate setting, to avoid `Presence Sensor/Motion Threshold` changing unexpectedly when someone sets `Motion Sensor/Motion Threshold`. But in my opinion, since we already show how settings work for one of the sensors, that is enough, and we don't need anything here.

samples/bluetooth/mesh/sensor_client/src/model_handler.c
akredalen: The motion sensor has the added feature according to the NLC profile that when motion is sensed, the area is to be considered occupied for a certain time which should be handled by the client. This is why the handling is a bit different than for presence detection. Is there still a part of the logic that you think is unnecessary? Or could you be more specific?

> When motion is detected, the space monitored by the sensor is considered occupied for a certain time. This time is not defined by the sensor, but instead by a model receiving the sensor messages.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -664,6 +737,9 @@ static struct bt_mesh_sensor *const ambient_light_sensor[] = {
 static struct bt_mesh_sensor *const occupancy_sensor[] = {
```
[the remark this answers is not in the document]
akredalen: Hmm. Perhaps, but I believe it will require changes in API. Do this in a separate PR?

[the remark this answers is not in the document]
alxelax: API change? Which one?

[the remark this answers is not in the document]
akredalen: With the change suggested here, `BT_MESH_SENSOR_SRV_INIT` will have an incompatible pointer type and give warning, and by making for example `struct bt_mesh_sensor motion_sensor` a const (like suggested in the above comment) the `bt_mesh_sensor_srv_pub` will discard 'const' qualifier and give a warning as well.

samples/bluetooth/mesh/sensor_server/README.rst
```diff
@@ -72,6 +75,12 @@ The following Bluetooth Mesh sensor types, and their settings, are used in this
 
 Moreover, the on-chip ``TEMP_NRF5`` temperature sensor is used for the nRF52 series, and the ``BME680`` temperature sensor for Thingy:53.
 
+.. note::
+  When running this sample on Thingy:53, some functionality will not be available as the device only
+  has two buttons. The two buttons on Thingy:53 will be used for the ambient light sensor and
+  presence detected sensor functionality as described for **Button 1** and **Button 2** in this
+  documentation. **Button 2** can be accessed by removing the top part of the casing.
```
wiba-nordic, line 82:
```suggestion
  When running this sample on Thingy:53, some functionality will not be available as the device only has two buttons.
  The two buttons on Thingy:53 will be used for the ambient light sensor and presence detected sensor functionality as described for **Button 1** and **Button 2** in this documentation.
  **Button 2** can be accessed by removing the top part of the casing.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -350,11 +350,18 @@ Bluetooth Mesh samples
 
 * :ref:`bluetooth_mesh_sensor_client` sample:
 
-  * Added support for the :ref:`nRF54L15 PDK <ug_nrf54l15_gs>` board.
+  * Added:
+
+    * Support for the :ref:`nRF54L15 PDK <ug_nrf54l15_gs>` board.
+    * Motion sensing and people count occupancy sensor types.
```
wiba-nordic, line 356: PR description mentions three sensor types, but only two are mentioned here.

wiba-nordic (changes requested):

samples/bluetooth/mesh/sensor_client/src/model_handler.c
```diff
@@ -11,18 +11,35 @@
...
+
 	if (sensor->id == bt_mesh_sensor_present_dev_op_temp.id) {
 		printk("Chip temperature: %s\n", bt_mesh_sensor_ch_str(value));
 	} else if (sensor->id == bt_mesh_sensor_presence_detected.id) {
 		int64_t presence_detected = 0;
 
-		(void)bt_mesh_sensor_value_to_micro(value, &presence_detected);
+		status = bt_mesh_sensor_value_to_micro(value, &presence_detected);
+		if (status != BT_MESH_SENSOR_VALUE_NUMBER) {
+			printk("Warning: unexpected behaviour during conversion of presence "
```
ludvigsj, line 39: If the code ends up here, we have no guarantees at all about the contents of `presence_detected`, and the code should stop processing and not continue like it currently does. Same for the Motion Sensed-sensor.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -737,46 +922,40 @@ static void button_handler_cb(uint32_t pressed, uint32_t changed)
...
+	}
+	if (BUTTON_PRESSED(pressed, changed, BIT(3))) {
+		int err;
+		static int people_count_idx;
+		struct bt_mesh_sensor_value val;
 
-			err = bt_mesh_sensor_srv_pub(&occupancy_sensor_srv, NULL,
-						&presence_sensor, &val);
+		dummy_people_count_value = dummy_people_count_values[people_count_idx++];
+		people_count_idx = people_count_idx % ARRAY_SIZE(dummy_people_count_values);
```
ludvigsj, line 945: Suggestion: I prefer `people_count_idx %= ARRAY_SIZE(...);`, I think it's more readable. But up to you.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -737,46 +922,40 @@ static void button_handler_cb(uint32_t pressed, uint32_t changed)
...
 
-			if (err) {
-				printk("Error publishing end of presence (%d)\n", err);
-			} else {
-				printk("Publishing end of presence\n");
-			}
+		err = bt_mesh_sensor_value_from_micro(people_count_sensor.type->channels[0].format,
+						      BASE_UNITS_TO_MICRO(dummy_people_count_value),
+						      &val);
+		if (err) {
```
ludvigsj, line 950: 1. I think this should check `if (err && err != -ERANGE)` here. `-ERANGE` is returned when the value is clamped, and I see no reason why clamping should be a problem in this case.
2. If an error other than `-ERANGE` happens, `val` will still be uninitialized, and shouldn't be published

ludvigsj: These go for every case of converting `bt_mesh_sensor_value_from_...` earlier in the file as well. They should probably all accept -ERANGE, and should return if any other error occurs.

samples/bluetooth/mesh/sensor_client/src/model_handler.c
ludvigsj: I'd also prefer to use `bt_mesh_sensor_value_status_is_numeric` instead of comparing to `BT_MESH_SENSOR_VALUE_NUMBER` directly, in all places where this is done. It will not make a difference for these specific sensor types (which don't support less than/greater than), but the best practice to check "did I get a number out of this" is to check for `...is_numeric`.

ludvigsj (changes requested):

wiba-nordic (approved):

samples/bluetooth/mesh/sensor_client/prj.conf
```diff
@@ -47,6 +47,7 @@ CONFIG_BT_MESH_RELAY=y
 CONFIG_BT_MESH_FRIEND=y
 CONFIG_BT_MESH_TX_SEG_MAX=24
 CONFIG_BT_MESH_RX_SEG_MAX=16
+CONFIG_BT_MESH_RX_SEG_MSG_COUNT=2
```
alxelax, line 50: Why is it required?

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -410,9 +442,9 @@ static int presence_detected_get(struct bt_mesh_sensor_srv *srv,
 				 struct bt_mesh_sensor_value *rsp)
 {
 	int err = bt_mesh_sensor_value_from_micro(sensor->type->channels[0].format,
-						  pres_detect * 1000000LL, rsp);
+						  BASE_UNITS_TO_MICRO(pres_detect), rsp);
```
alxelax, line 445: As I wrote before, boolean `True` is just not zero. Behavior of this implementation is not predictable.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -317,6 +329,26 @@ static struct bt_mesh_sensor rel_chip_temp_runtime = {
...
+				 struct bt_mesh_sensor_value *rsp)
+{
+	int err =
+		bt_mesh_sensor_value_from_micro(sensor->type->channels[0].format,
+						BASE_UNITS_TO_MICRO(dummy_people_count_value), rsp);
+
+	if (err && err != -ERANGE) {
+		printk("Error encoding people count (%d)", err);
+	}
+	return err;
```
ludvigsj, line 344: This should be inside the `if`, and the function should return 0 at the end. This is because when it ignores `-ERANGE`, it shouldn't return it as an error either. Same in other getters.

ludvigsj (changes requested):
One small change needed, then it's good

alxelax (changes requested):

samples/bluetooth/mesh/sensor_client/prj.conf
akredalen: With the new sensor types, two segmented messages are being sent at the same time when client requests update or `get`. I kept getting the _No multi-segment message contexts available_ transport error.

samples/bluetooth/mesh/sensor_server/src/model_handler.c
akredalen: Sorry Alex, your previous comment got away from me. How should I fix this? typecast? I guess the (bool) * 1000000LL was done by @ludvigsj initially.

samples/bluetooth/mesh/sensor_client/prj.conf
alxelax: Or do one or both transactions happen over multicast address?

samples/bluetooth/mesh/sensor_server/src/model_handler.c
ludvigsj: `BASE_UNITS_TO_MICRO(!!pres_detect)` is enough here.

alxelax: Nop it should be something like:
```
BASE_UNITS_TO_MICRO(pres_detect ? 1 : 0)
```
I faced that some compilers with max optimization skipped adapting `True` to `1` and just used any digit for it if it is not zero.

ludvigsj: They optimize out `!!`? I've never seen that I think, I thought `!!` was commonly used to force 1 or 0? But sure, your way is 100% unambiguous.

samples/bluetooth/mesh/sensor_client/prj.conf
ludvigsj: I agree, this sounds strange and might be a bug. This started happening once you added the new sensors?

samples/bluetooth/mesh/sensor_server/src/model_handler.c
alxelax: Nop. `!!` manages perfectly with conversion of any numeric type to boolean. This is the opposite situation. Boolean to numeric.

samples/bluetooth/mesh/sensor_client/prj.conf
akredalen: Ok. I am not sure if this issue is related to the changes in this PR then. Could this be something to investigate in a separate task? The sample works like expected despite this error. To answer your question, it happens frequently and is easy to reproduce but it does not happen every time no.

ludvigsj: But does it happen without your addition of extra sensors? If yes, I'd say this is an unrelated bug and should be reported and fixed separately. If it only started happening after you changes, it should probably be fixed here

alxelax: I doubt misbehavior was added here, since @akredalen didn't change anything in transport.

akredalen: I tested the sample without any of my changes now, and I do see the occasional errors/warnings:
```
Time since presence detected(0): <value is not known>                                      
Chip temp: 20.25, total samples: 40                                                        
Chip temp: 20.25, total samples: 41                                                        
Ambient light level: 0                                                                     
Presence detected: 0                                                                       
Time since presence detected(0): <value is not known>                                      
[00:02:54.257,843] <wrn> bt_mesh_delayable_msg: Unable allocate 12 chunks, allocated 4     
Chip temp: 20.25, total samples: 42                                                        
[00:02:54.258,544] <wrn> bt_mesh_delayable_msg: Unable allocate 12 chunks, allocated 4     
[00:02:54.449,096] <err> bt_mesh_transport: No multi-segment message contexts available    
[00:02:54.458,282] <err> bt_mesh_transport: No multi-segment message contexts available    
Ambient light level: 0                                                                     
Presence detected: 0  
```

alxelax: I see messages. This is not related to lack of sar context or parallel sending of 2 sar transaction. This is lack of buffers for delayable messages. It requires adding more chunks `BT_MESH_ACCESS_DELAYABLE_MSG_CHUNK_COUNT`. Nevertheless, we can do better investigation in separate task

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -410,14 +443,15 @@ static int presence_detected_get(struct bt_mesh_sensor_srv *srv,
...
 {
 	int err = bt_mesh_sensor_value_from_micro(sensor->type->channels[0].format,
-						  pres_detect * 1000000LL, rsp);
+						  BASE_UNITS_TO_MICRO(pres_detect ? 1 : 0), rsp);
 
-	if (err) {
+	if (err && err != -ERANGE) {
 		printk("Error encoding presence detected (%d)\n", err);
+		return err;
 	} else {
```
ludvigsj, line 451: `else` is no longer needed with early return

akredalen: seems compliance skipped this one! fixed now.

samples/bluetooth/mesh/sensor_client/prj.conf
akredalen: Hmm. Would it make sense to increase the chunk count based on what was added in this PR or is it still a more general issue? I mean, it was already an issue before but it has certainly gotten worse with my latest changes. I would not mind adding the fix here if it makes more sense. If not I will create an issue on Jira. Let me know.

akredalen:
DNM until PTS issues have been resolved.

akredalen:
PTS tests now pass with the different occupancy sensor types on different elements.

wiba-nordic (approved):

alxelax (approved):

samples/bluetooth/mesh/sensor_server/src/model_handler.c
```diff
@@ -105,11 +111,29 @@ static const struct bt_mesh_comp2_record comp_rec[2] = {
 	.elem_offset_cnt = 1,
 	.elem_offset = cmp2_elem_offset2,
 	.data_len = 0
+	},
+	{
+	.id = BT_MESH_NLC_PROFILE_ID_OCCUPANCY_SENSOR,
+	.version.x = 1,
+	.version.y = 0,
+	.version.z = 0,
+	.elem_offset_cnt = 2,
```
ludvigsj, line 120: This looks weird to me. `cmp2_elem_offset_2` has length 1, why is count set to 2 here (and 3 below)?
`elem_offset_cnt` is according to spec the "Total number of element offsets in the
Element_Offset_List field"

ludvigsj (changes requested):

samples/bluetooth/mesh/sensor_server/src/model_handler.c
ludvigsj: And also: why is `.id = ...OCCUPANCY_SENSOR` listed 3 times, while people count and motion is not mentioned?

akredalen: First comment: I must have misunderstood the intention of the variables. Will fix. Second comment: they are all occupancy sensor types. Should this be updated somewhere? What is your suggestion?

ludvigsj: Oh sorry, second comment can be ignored, they are of course all implementing the same profile and should have the same profile ID.

First comment: The way this should be is that `elem_offset_cnt` should always be 1, and you have to create `cmp2_elem_offset_3`, `cmp2_elem_offset_4`, etc. (Preferably rename them to indicate what the element does, in my opinion, like `cmp2_elem_offset_ambient_light`, `cmp2_elem_offset_presence`, etc.)

PR #15100, opened 2024-04-26
tfm: 54l: Use a more reasonable amount of RAM by default for TFM

SebastianBoe:
Reduce the default amount of RAM for TFM from 200kB to 65kB.

This makes it possible to fit TF-M with other applications that use a lot of RAM, like Matter.


mswarowsky (approved):

frkv (approved):
LGTM

PR #15114, opened 2024-04-26
nrf54L15: Partition configuration for TF-M and MCUBOOT

Vge0rge:
The old configuration of the TF-M partition doesn't work well with the nRF54L15. This changes the order of the TF-M storage partitions to be placed before the non-secure application in an effort to reduce the required secure/non-secure sections.
Check the commit messages for details.

Ref: NCSDK-24863


SebastianBoe (approved):

modules/mcuboot/boot/zephyr/Kconfig
```diff
@@ -15,14 +15,8 @@ partition-size=0x1e000
 source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_config"
 
 partition=MCUBOOT_PAD
-if SOC_SERIES_NRF54LX
-	partition-size=0x800
+partition-size=0x200
```
de-nordic, line 18: I do not think this change is ok. Look at this comment: https://github.com/nrfconnect/sdk-nrf/pull/14246#discussion_r1505590639
there is minimal size for alignment, that depends on IRQs reserved. This platform reserves 271 which in turn gives more than 2k for VTOR table, and when aligned to next power of two gives 4k (0x800).

Are you sure about this change?

nvlsianpu (changes requested):
As Dominik pointed - mcuboot header size must be multiplier of 0x800 for nRF54L15

modules/mcuboot/boot/zephyr/Kconfig
Vge0rge: You are both right. I removed the commit to alter the mcuboot padding size and I fixed my alignment requirements by modifying the TF-M partitions themselves. 

FYI: @nvlsianpu

nvlsianpu (approved):

de-nordic (approved):

sigvartmh (approved):

Vge0rge:
Closing this since we realized that we will have to change this design with a follow up PR soon.

subsys/partition_manager/pm.yml.tfm
```diff
@@ -23,6 +23,43 @@ tfm_nonsecure:
...
+
+tfm_otp_nv_counters:
+  placement:
+#ifdef CONFIG_BOOTLOADER_MCUBOOT
+    before: mcuboot_primary
+#else
+    before: tfm_nonsecure
+#endif
+  inside: tfm_storage
+  size: CONFIG_PM_PARTITION_SIZE_TFM_OTP_NV_COUNTERS
```
sigvartmh, line 59: TODO: Move counters to OTP maybe. Needs to coordinate with NSIBs OTP counters.

PR #15120, opened 2024-04-29
lib: at_parser: add new AT parser

MirkoCovizzi:
Introduces the `at_parser` library for parsing AT command responses, notifications, or events.
It exports an API that lets the user iterate through an AT command string and retrieve any of its elements based on their indices, efficiently and without heap allocations. Under the hood, this library leverages the power of regular expressions converted to deterministic finite automata by the free and open-source lexer generator `re2c`.

Updated components:

- [x] lib/lte_link_control
- [x] lib/modem_info
- [x] lib/location
- [x] subsys/net/lib/zzhc
- [x] subsys/net/lib/lwm2m_client_utils
- [x] samples/cellular/modem_shell
- [x] samples/cellular/fmfu_smp_svr
- [x] applications/serial_lte_modem

Documentation-related changes:

- [x] doc/nrf/libraries/modem/at_parser.rst
- [x] releases/release-notes-changelog.rst
- [x] Update changelogs for libraries
- [x] doc/nrf/releases_and_maturity/migration/migration_guide_2.8.rst
- [x] Replace mentions/references of `at_cmd_parser` library
- [x] Remove unused mentions/references of `at_cmd_parser` library
- [x] Update commit messages
- [x] Document internal headers
- [x] Document static functions where necessary
- [x] Document API

Misc:

- [x] Add deprecation notice for `at_cmd_parser` Kconfig option


include/modem/at_parser.h
```diff
@@ -0,0 +1,50 @@
...
+
+struct at_parser {
+	const char *cursor;
+	size_t count;
+	bool is_next_empty;
+	bool is_init;
+};
+
+int at_parser_init(struct at_parser *parser, const char *at);
+int at_parser_num_get(struct at_parser *parser, size_t index, void *value, enum at_num_type type);
```
lemrey, line 40: here instead of using a `enum` to determine what to copy, I would check if we can use C11 generics or alternatively, let the parameter be a [transparent union](https://gcc.gnu.org/onlinedocs/gcc-8.3.0/gcc/Common-Type-Attributes.html) of all the supported types.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,384 @@
...
+	do {
+		ret = at_parser_tok(parser, &token);
+	} while (ret == 0);
+
+	if (ret != -EAGAIN) {
+		return ret;
+	}
+
+	/* Skip CRLF. */
+	parser->cursor += 2;
```
lemrey, line 332: magic, maybe use `strlen("\r\n")`

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,384 @@
...
+	/* Reset count. */
+	parser->count = 0;
+
+	return 0;
+}
+
+size_t at_valid_count_get(const char *at)
```
lemrey, line 339: What is this function counting? I would add that to the name and prefix the `at_parser_` part, e.g. `at_parser_tok_count_get()`

include/modem/at_parser.h
lemrey, line 1: At least API in this file should be named `at_parser_`

lemrey (commented):
Did a first pass

lemrey:
@MirkoCovizzi please keep separate commits so it's easier to review :)

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,502 @@
...
+{
+	int diff;
+	static const char * const toclip[] = {
+		"\r\nOK\r\n",
+		"\r\nERROR\r\n",
+		"\r\n+CME ERROR",
+		"\r\n+CMS ERROR"
+	};
+
+	for (size_t i = 0; i < 4; i++) {
```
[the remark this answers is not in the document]
MirkoCovizzi: I'm aware but it's not really needed here and it would introduce the inclusion of Zephyr just for that.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,502 @@
...
+	parser->cursor = parser->at;
+	parser->count = 0;
+	memset(&parser->prev_token, 0, sizeof(struct at_token));
+	parser->is_next_empty = false;
+}
+
+int at_parser_init(struct at_parser *parser, const char *at)
+{
+	if (!parser || !at) {
+		return -EINVAL;
```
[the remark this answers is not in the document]
MirkoCovizzi: Perhaps you mean the inverse?

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,502 @@
...
+	}
+
+	if (!parser->initialized) {
+		return -EPERM;
+	}
+
+	if (!parser->cursor) {
+		return -EINVAL;
+	}
```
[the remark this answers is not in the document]
MirkoCovizzi: The pointer to `parser` comes from the outside, the value might have been changed to null.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,502 @@
...
+	if (lookahead_crlf_and_more(remainder) && !is_response(remainder) &&
+	    !parser->is_next_empty) {
+		/* Skip CRLF. */
+		parser->cursor += 2;
+		/* Reset counter. */
+		parser->count = 0;
+		parser->at = remainder;
+
+		return -EAGAIN;
+	}
```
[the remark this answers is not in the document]
MirkoCovizzi: This recognizes if there is a subsequent line after the current one. This section has been refactored and has been moved to the beginning of the function.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,502 @@
...
+
+	return 0;
+}
+
+int at_parser_seek(struct at_parser *parser, size_t index, struct at_token *token)
+{
+	int ret;
+	struct at_token tok = {0};
```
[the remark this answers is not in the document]
MirkoCovizzi: Yes we can.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,384 @@
...
+
+	/* Nothing left to tokenize. */
+	if (cursor_len == 0 || is_crlf(parser->cursor)) {
+		return -EIO;
+	}
+
+	/* Don't tokenize response. */
+	if (is_resp(parser->cursor)) {
+		return -EIO;
+	}
```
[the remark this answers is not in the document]
MirkoCovizzi: I responded above, but in essence responses are not used anywhere.

lib/at_parser/at_parser.c
MirkoCovizzi: I thought about that but this is more efficient and easy enough to understand already.

lib/at_parser/at_parser.c
MirkoCovizzi: I didn't name it with `at_parser_` prefix because it doesn't accept the `at_parser` as a parameter. This counts the number of valid parameters in the first line of the AT string. This counting function is heavily used in multiple libraries so I had to add it for some degree of compatibility with existing code.

include/modem/at_parser.h
MirkoCovizzi: `at_valid_count_get` and `at_cmd_type_get` are more considered to be utility functions than AT Parser APIs or functions that act on the `at_parser` structure.

include/modem/at_parser.h
MirkoCovizzi: Didn't think about that, neat idea.

tests/lib/at_parser/boards/native_posix.conf
```diff
@@ -0,0 +1,7 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+CONFIG_NEWLIB_LIBC=n
```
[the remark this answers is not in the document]
MirkoCovizzi: Not sure, have to check.

lib/at_parser/at_parser.c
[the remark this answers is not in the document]
lemrey: but by response you mean `OK` or `ERROR` etc. right? Because when I think about response I think about the whole AT string returned by the modem in response to a command (solicited).

lib/at_parser/at_parser.c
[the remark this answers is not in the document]
lemrey: I usually use `EFAULT` for input that causes bad memory accesses (e.g. `NULL`) and `EINVAL` for input that is not `NULL` but otherwise incorrect (we do the same in `libmodem` too).

include/modem/at_parser.h
MirkoCovizzi: OK

include/modem/at_parser.h
```diff
@@ -0,0 +1,57 @@
...
+};
+
+struct at_parser {
+	const char *cursor;
+	size_t count;
+	bool is_next_empty;
+	bool is_init;
+};
+
+#define at_parser_num_get(parser, index, value) \
```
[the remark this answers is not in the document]
trantanen: Would it be then client who would need to cast `void*` to `int16_t` etc.?

Copying the signature here for convenience:
`int at_parser_num_get_impl(struct at_parser *parser, size_t index, void *value, enum at_num_type type)`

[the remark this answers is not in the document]
MirkoCovizzi: I don't like exporting that kind of function because it would remove any kind of type-safety. Enums in C are not type-safe. With the current approach, yes, there are still a number of variant functions to get an integer, however the addition of a generic function is a strict improvement over the previous API.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,411 @@
...
+	if (ret) {
+		return ret;
+	}
+
+	do {
+		ret = at_parser_tok(&parser, &token);
+		if (ret != -EBADMSG && ret != -EAGAIN && ret != -EIO) {
+			count++;
+		}
+	} while (ret != -EAGAIN && ret != -EIO);
```
lemrey, line 387: There should be a simpler way to do this than check for 4 different errors.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	/* Don't tokenize response. */
+	if (is_resp(parser->cursor)) {
+		return -EIO;
+	}
+
+	/* Check if there is a new line to parse. */
+	if (lookahead_crlf_and_more(parser->cursor)) {
+		return -EAGAIN;
```
lemrey, line 141: Why do we stop tokenizing on new lines? I think the user should be able to call this function in a loop without looking for special errors that mean "there is another line to parse". The parser should just start tokenizing the next line. If there is a use-case where the use need a function to parse one line at a time, I suggest to introduce a `_tok_line()` function.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,411 @@
...
+
+	/* Update the length to reflect the copied string length. */
+	*len = token.len;
+
+	return 0;
+}
+
+int at_parser_next_line_move(struct at_parser *parser)
+{
+	int ret;
```
lemrey, line 339: nit: I prefer `err` to `ret`, especially when the return value is only an error and not an integer counting something.

Then the loop below would be a bit clearer:

```
do {
	ret = at_parser_tok(parser, &token);
} while (ret == 0);

```
becomes
```
do {
	err = at_parser_tok(parser, &token);
} while (!err);
```

lib/at_parser/Kconfig
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config AT_PARSER
+	bool "AT parser library"
+	depends on NEWLIB_LIBC || EXTERNAL_LIBC || PICOLIBC
```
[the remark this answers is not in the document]
lemrey: Regardless of what function introduces the dependency, if need to express that, you shouldn't depend on a list of C library options. Instead you should `depends on REQUIRES_FULL_LIBC`

lemrey (commented):
Please keep the changes to other libraries in a separate commits.

lib/at_parser/Kconfig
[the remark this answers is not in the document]
MirkoCovizzi: Removed it. The library doesn't use anything out of the ordinary.

lib/at_parser/at_parser.c
MirkoCovizzi: The parsing works line by line and is index based. Parsing without any bounds for lines would compromise this balance. A notification has 5 subparameters and outputs 2 lines of it. We ask for subparameter with index 8, what is it? What if the first line has 4 subparameters because it is not including the last optional? I appreciated the original design choice of returning `-EAGAIN` because it makes sense and informs the user, at runtime, that there is more to parse.

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -299,31 +297,22 @@ int parse_edrx(const char *at_response, struct lte_lc_edrx_cfg *cfg, char *edrx_
...
-		return err;
-	}
-
-	/* Parse response and populate AT parameter list */
-	err = at_parser_params_from_str(at_response,
-					NULL,
-					&resp_list);
-	if (err) {
-		LOG_ERR("Could not parse eDRX response, error: %d", err);
+		LOG_ERR("Could not init AT parser for eDRX response, error: %d", err);
```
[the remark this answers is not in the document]
MirkoCovizzi: Good point

junqingzou (approved):
ZZHC changes approved

applications/serial_lte_modem/src/nativetls/slm_at_cmng.c
MarkusLassila, line 1: at_params_string_ptr_get in this file will require a replacement. The certificates coming can be up to 4096 bytes, hopefully additional buffer is not needed for this.

applications/serial_lte_modem/src/nrfcloud/slm_at_nrfcloud.c
```diff
@@ -703,8 +705,7 @@ static int handle_at_nrf_cloud_pos(enum at_cmd_type cmd_type,
 
 			/* Parse the MAC address. */
 			len = sizeof(mac_addr_str);
-			err = at_params_string_get(
-				param_list, param_idx, mac_addr_str, &len);
+			err = at_parser_string_get(parser, param_idx, mac_addr_str, &len);
```
MarkusLassila, line 708: This seems to be broken. Please run the TestSLMnRFCloud tests for SLM.

Compile with:

python3 /home/<username>/ncs/test-fw-nrfconnect-nrf_iot/utils/oulu/ci/slm/build_images.py /home/<username>/ncs/nrf/applications/serial_lte_modem/ nrf9160dk/nrf9160/ns -i nrf_cloud

MarkusLassila (changes requested):
Looking rather good.

Pushed some other corrections to this branch.

applications/serial_lte_modem/src/nativetls/slm_at_cmng.c
MirkoCovizzi: I have a function for that, will replace. Thanks.

MirkoCovizzi: Added a function.

include/modem/at_parser.h
```diff
@@ -0,0 +1,61 @@
...
+		uint16_t * : at_parser_uint16_get, \
+		int32_t * : at_parser_int32_get,   \
+		uint32_t * : at_parser_uint32_get, \
+		int64_t * : at_parser_int64_get,   \
+		uint64_t * : at_parser_uint64_get  \
+	)(parser, index, value)
+
+int at_parser_init(struct at_parser *parser, const char *at);
+int at_parser_next(struct at_parser *parser);
+int at_parser_cmd_get(struct at_parser *parser, char **at);
```
tokangas, line 44: The parsed string is given into `at_parser_init()` as `const`, but this function returns a non-const pointer to it. If the parsed string is `const`, all returned pointers to it should also be `const`.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,522 @@
...
+		*type = AT_PARSER_CMD_TYPE_TEST;
+		break;
+	case AT_TOKEN_TYPE_CMD_READ:
+		*type = AT_PARSER_CMD_TYPE_READ;
+		break;
+	case AT_TOKEN_TYPE_CMD_SET:
+		*type = AT_PARSER_CMD_TYPE_SET;
+		break;
+	default:
+		return -EOPNOTSUPP;
```
tokangas, line 321: `type` could be set to `AT_PARSER_CMD_TYPE_UNKNOWN` in this case, I guess.

include/modem/at_parser.h
```diff
@@ -0,0 +1,61 @@
...
+	AT_PARSER_CMD_TYPE_SET,
+	AT_PARSER_CMD_TYPE_READ,
+	AT_PARSER_CMD_TYPE_TEST
+};
+
+struct at_parser {
+	const char *at;
+	const char *cursor;
+	size_t count;
+	bool is_init;
```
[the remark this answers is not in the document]
MirkoCovizzi: Good point. Changed it to bitfield.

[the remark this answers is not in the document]
MirkoCovizzi: Changed to use a sentinel value.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,529 @@
...
+	}
+
+	errno = 0;
+	char *next = NULL;
+
+	/* Check unsigned long long first, using its own parsing function. */
+	if (type == AT_NUM_TYPE_UINT64) {
+		uint64_t val = strtoull(token.start, &next, 10);
+
+		if (errno == ERANGE && (val == 0 || val == UINT64_MAX)) {
```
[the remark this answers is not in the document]
MirkoCovizzi: 1. Will double check
2. It cannot fail to parse because the string is guaranteed to begin with an integer due to the tokenization step

include/modem/at_parser.h
```diff
@@ -0,0 +1,307 @@
...
+int at_parser_uint64_get(struct at_parser *parser, size_t index, uint64_t *value);
+
+/**
+ * @brief Get a value as a string.
+ *
+ * The data type must be a string (quoted or non-quoted), an AT command prefix, or an array,
+ * otherwise an error is returned.
+ * The string value is copied to the buffer.
+ * @p len must be bigger than the string length, or an error is returned.
+ * The copied string is not null-terminated.
```
lemrey, line 263: Why is the string not null-terminated? I think it would be quite convenient to have the string be null-terminated, so that it's ready to be printed or otherwise manipulated with `string.h` functions.

include/modem/at_parser.h
```diff
@@ -0,0 +1,307 @@
...
+ * @param[in] parser A pointer to the AT parser.
+ *
+ * @retval 0 If the operation was successful.
+ *           Otherwise, a (negative) error code is returned.
+ * @retval -EINVAL     One or more of the supplied parameters are invalid.
+ * @retval -EPERM      @p parser has not been initialized.
+ * @retval -EOPNOTSUPP The cursor cannot move to the next line. This error occurs if there is no
+ *                     next line to move to or the configured AT command string is malformed.
+ */
+int at_parser_next(struct at_parser *parser);
```
lemrey, line 111: Perhaps `_line_next()` would be a bit more descriptive. It's not clear that this function moves the parser to the next line without reading the documentation.

include/modem/at_parser.h
lemrey, line 1: What happened to the `_tok()` function that retrieved the next token? I think that would have been nice to have.

lib/at_parser/at_parser.c
lemrey: ok

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,551 @@
...
+		*(uint64_t *)value = val;
+
+		return 0;
+	}
+
+	int64_t val = strtoll(token.start, &next, 10);
+
+	if (errno == ERANGE) {
+		return -ERANGE;
+	}
```
lemrey, line 424: Why are we doing this instead of handling this in the `switch` case below?

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,551 @@
...
+	parser->count = 0;
+	/* Set pointer of current AT command string to the current cursor, which points to the
+	 * beginning of a new AT command line.
+	 */
+	parser->at = parser->cursor;
+
+	return 0;
+}
+
+int at_parser_cmd_get(struct at_parser *parser, const char **at)
```
lemrey, line 302: What's the use case for this function? Was there a case where `at_parser` was initialized somewhere, and the parsing happened somewhere else?

doc/nrf/libraries/modem/at_parser.rst
```diff
@@ -0,0 +1,103 @@
...
+==============
+
+To begin using the AT parser, you need to first initialize it. This is done by invoking the :c:func:`at_parser_init` function, where you provide a pointer to the AT parser and the AT command string that needs to be parsed.
+
+The following code snippet shows how to initialize an AT parser for an AT command response:
+
+.. code-block:: c
+
+   int err;
+   struct at_parser parser = { 0 };
```
lemrey, line 26: One comment here, `at_parser_init()` zeros and then initializes the parser instance. Why is then the parser always zero-initialized upon declaration? That way it is actually zeroed twice, which is not necessary.

doc/nrf/libraries/modem/at_parser.rst
```diff
@@ -0,0 +1,103 @@
...
+      return err;
+   }
+
+   /* `at_parser_string_get` does not null-terminate the copied string so null-terminate it. */
+   buffer[len] = '\0';
+
+   /* "Prefix: `+CGCONTRDP`" */
+   printk("Prefix: `%s`\n", buffer);
+
+   /* Retrieve the first subparameter, at index 1 of the AT command string.
```
lemrey, line 67: Why is the first parameter (parameter at index 1) a "subparameter"?
What is a "subparameter", in general?

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,548 @@
...
+{
+	return  (str[0] == CR || str[0] == LF) && str[1] != NULL_TERMINATOR;
+}
+
+/* Lookahead if the string matches either CR or LF, or starts with either CR or LF and has more
+ * data.
+ */
+static bool lookahead_cr_lf_or_more(const char *str)
+{
+	return lookahead_cr_lf(str) || lookahead_cr_lf_and_more(str);
```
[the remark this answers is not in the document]
MirkoCovizzi: `lookahead_cr_lf` matches exactly `"\r\n\0"` (end of the AT command string), while `lookahead_cr_lf_or_more` can match either `"\r\n\0"` or `"\r\nabcdefgh..."`.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,548 @@
...
+			return true;
+		}
+	}
+
+	return false;
+}
+
+static bool is_index_ahead(struct at_parser *parser, size_t index)
+{
+	return index + 1 > parser->count;
```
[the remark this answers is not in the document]
MirkoCovizzi: To me it is clearer this way because of what index and count mean. For example, the first element has index 0, but count 1. The inequality with the `+ 1`  normalizes this comparison and the `>` shows that the element is ahead.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,551 @@
...
+/* Carriage Return. */
+#define CR '\r'
+/* Line Feed. */
+#define LF '\n'
+/* Null Terminator. */
+#define NULL_TERMINATOR '\0'
+/* Minus Sign. */
+#define MINUS_SIGN '-'
+/* Init Sentinel. */
+#define INIT_SENTINEL 0xdeadbeef
```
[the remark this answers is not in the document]
MirkoCovizzi: I can change it.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,548 @@
...
+		"OK\r\n",
+		"ERROR\r\n",
+		"+CME ERROR",
+		"+CMS ERROR"
+	};
+
+	trim_crlf(&str);
+
+	for (size_t i = 0; i < ARRAY_SIZE(resp); i++) {
+		if (strncmp(str, resp[i], strlen(resp[i])) == 0) {
```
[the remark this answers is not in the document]
MirkoCovizzi: I might consider it, `strlen` is consumes a lot of space and I've been removing it from many places.

lib/at_parser/at_parser.c
MirkoCovizzi: Updated.

doc/nrf/libraries/modem/at_parser.rst
MirkoCovizzi: Updated.

doc/nrf/libraries/modem/at_parser.rst
MirkoCovizzi: `Subparameter` is the terminology used in the `nRF91x1 AT Command Reference Guide` when referring to the values of an AT command string.

lib/at_parser/at_parser.c
MirkoCovizzi: I removed it. At some point it was replaced with a different function where it was needed but I forgot to remove this one.

include/modem/at_parser.h
MirkoCovizzi: There is really no usecase for it. Everything can be done with the type-specific get functions.

include/modem/at_parser.h
MirkoCovizzi: Updated.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,532 @@
...
+		/* if the token is a subparameter but does not have a trailing comma then it is the
+		 * last subparameter of the line.
+		 * If there is more to parse, then the token must be followed by CRLF, CR, or LF,
+		 * otherwise the string is malformed.
+		 */
+		bool lookahead = lookahead_cr_lf_or_more(remainder) ||
+				 lookahead_crlf_or_more(remainder);
+
+		if (token->variant != AT_TOKEN_VARIANT_COMMA) {
+			if (remainder[0] != NULL_TERMINATOR && !lookahead) {
```
anhmolt, line 218: Tried to condense this logic a bit since I found parts of it hard to read. Got this:
```
bool next_is_NULL_CR_or_LF = (remainder[0] == NULL_TERMINATOR) ||
			     lookahead_cr_lf_or_more(remainder) ||
			     lookahead_crlf_or_more(remainder);

if (token->variant == AT_TOKEN_VARIANT_COMMA && next_is_NULL_CR_or_LF) {
	parser->is_next_empty = true;
} else if (token->variant == AT_TOKEN_VARIANT_NO_COMMA && !next_is_NULL_CR_or_LF) {
	return -EBADMSG;
}
```
I left out the comments in the snippet.

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,551 @@
...
+	errno = 0;
+	char *next = NULL;
+
+	/* Check unsigned 64-bit integer first, using its own parsing function. */
+	if (type == AT_NUM_TYPE_UINT64) {
+		if (token.start[0] == MINUS_SIGN) {
+			return -ERANGE;
+		}
+
+		uint64_t val = strtoull(token.start, &next, 10);
```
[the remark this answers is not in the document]
MirkoCovizzi: I don't think it's necessary, it's pretty much the default value most of the time.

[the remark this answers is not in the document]
trantanen: Personally, I tend to like the magic number in these cases for readability. You don't need to look for confirmation on the value anywhere. Especially when it's not a constant that you don't intend to change under any circumstances.

divipillai:
@gmarull : Could you please take a look on the doc issue: The issue is because APIs with the same names are used in the different header file. I think the same name cannot be used. Is there any way to resolve this issue by keeping the name as is? 

image

MirkoCovizzi:
> @gmarull : Could you please take a look on the doc issue: The issue is because APIs with the same names are used in the different header file. I think the same name cannot be used. Is there any way to resolve this issue by keeping the name as is?
> 
> image

Added note that while the names are the same, the signature is indeed different so I was surprised when I saw the error.

gmarull:
Doxygen shares a global namespace, so no, there’s no solution to this issue

include/modem/at_parser.h
MirkoCovizzi: Re-added null termination.

trantanen (approved):
Good work! Nice new AT parsing API.

lib/at_parser/at_match.re
```diff
@@ -0,0 +1,193 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/* Generated by re2c 3.0
```
bjda, line 7: As someone unfamiliar with re2c, it's a bit confusing at first glance that the .re file has a "Generated by re2c" comment. Unless this form is standard practice, maybe we should say "at_match.c generated by re2c 3.0" to make it immediately clear that this file is the human-written source.

bjda (approved):

lib/at_parser/at_match.h
```diff
@@ -0,0 +1,77 @@
...
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/**
+ * @file at_match.h
+ *
+ * @defgroup at_match Internal AT parser matching functions.
+ * @ingroup at_parser
```
anhmolt, line 11: Is the thought to include the functions in this header to documentation or to leave them out?

lib/at_parser/at_parser.c
```diff
@@ -0,0 +1,530 @@
...
+
+	switch (type) {
+	case AT_NUM_TYPE_INT16:
+		if (val < INT16_MIN || val > INT16_MAX) {
+			return -ERANGE;
+		}
+		*(int16_t *)value = (int16_t)val;
+		break;
+	case AT_NUM_TYPE_UINT16:
+		if (val < 0 || val > UINT16_MAX) {
```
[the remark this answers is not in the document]
MirkoCovizzi: I would refrain from using a macro for a simple logical expression.

lib/at_parser/at_match.h
MirkoCovizzi: Ah, good point.

doc/nrf/releases_and_maturity/migration/migration_guide_2.8.rst
```diff
@@ -0,0 +1,243 @@
...
+
+AT command parser
+-----------------
+
+.. toggle::
+
+  * The :ref:`at_cmd_parser_readme` library has been deprecated in favor of the :ref:`at_parser_readme` library and will be removed in a future version.
+
+    You can follow this guide to migrate your application to use the :ref:`at_parser_readme` library.
+    This will reduce the footprint of the application and will alleviate memory pressure on the heap.
```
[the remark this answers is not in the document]
MirkoCovizzi: `Reduce` is good.

include/modem/at_cmd_parser.h
```diff
@@ -127,7 +127,7 @@ enum at_cmd_type {
  *
  * @return Command type.
  */
-enum at_cmd_type at_parser_cmd_type_get(const char *at_cmd);
+enum at_cmd_type at_parser_at_cmd_type_get(const char *at_cmd);
```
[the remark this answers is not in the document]
MirkoCovizzi: No, because this commit has to be bisectable and correct by itself. If the change was in a separate commit the documentation build would fail in the AT parser commit.

lib/at_parser/at_match.h
MirkoCovizzi: I think it's fine to leave it as is, `at_utils.h` in the `at_cmd_parser` also has a similar structure and it's supposed to be an internal header only.

doc/nrf/releases_and_maturity/migration/migration_guide_2.8.rst
```diff
@@ -0,0 +1,243 @@
...
+       * Add:
+
+         .. code-block:: C
+
+          size_t count;
+
+          /* `&at_parser` is a pointer to the AT parser.
+           * `&count` is a pointer to the returned parameter count.
+           */
+          at_parser_cmd_count_get(&at_parser, &count);
```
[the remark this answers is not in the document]
MirkoCovizzi: You have a point, it should be added.

tests/lib/at_parser/src/main.c
```diff
@@ -0,0 +1,1986 @@
+/*
```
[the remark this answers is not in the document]
MirkoCovizzi: Extended the `-EOPNOTSUPP` tests with this case.

include/modem/at_parser.h
```diff
@@ -0,0 +1,349 @@
...
+ */
+int at_parser_uint64_get(struct at_parser *parser, size_t index, uint64_t *value);
+
+/**
+ * @brief Get a string value.
+ *
+ * The data type must be a string (quoted or non-quoted), an AT command prefix, or an array,
+ * otherwise an error is returned.
+ * The string value is copied to the buffer.
+ * @p len must be larger than the string length plus one, or an error is returned.
```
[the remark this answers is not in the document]
MirkoCovizzi: Updated the docstring.

lib/at_parser/at_match.h
anhmolt: Yeah. Should be fine. I learned that `:inner:` need to be added after the `.. doxygengroup:: name_of_group` in the `.rst` to show content from subgroups that uses `@ ingroup`.

lib/at_parser/at_match.re
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * re2c at_match.re -o at_match.c -W --no-debug-info --tags
+ */
```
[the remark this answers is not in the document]
MirkoCovizzi: Yes, however there is also a formatting step with `clang` but it requires adding two lines to `.clangformat`.

doc/nrf/libraries/modem/at_parser.rst
```diff
@@ -0,0 +1,97 @@
...
+   }
+
+   /* "Third subparameter: `internet`" */
+   printk("Third subparameter: `%s`\n", buffer);
+
+API documentation
+*****************
+
+| Header file: :file:`include/modem/at_parser.h`
+| Source file: :file:`lib/at_parser/src/at_parser.c`
```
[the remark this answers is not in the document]
MirkoCovizzi: This is not necessary. Other libraries that have multiple source files, like the LTE Link Controller, only include the exported API, not the internal sources/APIs.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -897,6 +897,13 @@ DFU libraries
 Modem libraries
 ---------------
 
+* Added:
+
+   * :ref:`at_parser_readme` library.
+     The :ref:`at_parser_readme` is a library that parses AT command responses, notifications, or events.
```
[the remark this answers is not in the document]
MirkoCovizzi: I believe this is a matter of stylistic choice/taste. I will leave it as it is for the tech writers to chime in with their opinion if they see fit.

doc/nrf/libraries/modem/at_parser.rst
```diff
@@ -0,0 +1,97 @@
...
+
+.. contents::
+   :local:
+   :depth: 2
+
+Overview
+========
+
+The AT parser is a library that parses AT command responses, notifications, or events.
+It exports an API that lets the user iterate through an AT command string and retrieve any of its elements based on their indices, efficiently and without allocations.
```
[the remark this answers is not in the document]
MirkoCovizzi: Same as above. Stylistic choice. Will wait for tech writers.

doc/nrf/releases_and_maturity/migration/migration_guide_2.8.rst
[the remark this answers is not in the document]
MirkoCovizzi: Reworded to: `This will reduce the footprint of the application and will decrease memory requirements on the heap`.

doc/nrf/releases_and_maturity/migration/migration_guide_2.8.rst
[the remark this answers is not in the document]
MirkoCovizzi: Added a note at the end.

tests/lib/at_parser/src/main.c
```diff
@@ -0,0 +1,1986 @@
...
+	zassert_equal(ret, -EINVAL);
+
+	ret = at_parser_uint16_get(&parser, 1, NULL);
+	zassert_equal(ret, -EINVAL);
+
+	ret = at_parser_uint16_get(NULL, 1, &val);
+	zassert_equal(ret, -EINVAL);
+}
+
+ZTEST(at_parser, test_at_parser_uint16_get_erange)
```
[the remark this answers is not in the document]
MirkoCovizzi: Added tests.

tests/lib/at_parser/src/main.c
```diff
@@ -0,0 +1,1986 @@
...
+	zassert_equal(ret, -EINVAL);
+
+	ret = at_parser_uint32_get(&parser, 1, NULL);
+	zassert_equal(ret, -EINVAL);
+
+	ret = at_parser_uint32_get(NULL, 1, &val);
+	zassert_equal(ret, -EINVAL);
+}
+
+ZTEST(at_parser, test_at_parser_uint32_get_erange)
```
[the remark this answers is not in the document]
MirkoCovizzi: Added tests.

tests/lib/at_parser/testcase.yaml
```diff
@@ -0,0 +1,6 @@
+tests:
+  at_parser.at_parser:
+    platform_allow: native_sim
```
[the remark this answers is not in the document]
MirkoCovizzi: Added.

samples/cellular/modem_shell/src/at/at_cmd_mode_custom.c
```diff
@@ -50,44 +50,38 @@ static int at_cmd_ping_handle(struct at_param_list *at_params, enum at_cmd_type
...
-		if (err < 0) {
-			return err;
-		};
-	}
-	if (at_params_valid_count_get(at_params) > 5) {
-		err = at_params_unsigned_int_get(at_params, 5, &ping_args.interval);
-		if (err < 0) {
-			return err;
-		};
+	err = at_parser_num_get(parser, 2, &ping_args.len);
```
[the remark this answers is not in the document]
MirkoCovizzi: No, I forgot to restore the subparam count logic while refactoring. I restored it.

doc/nrf/releases_and_maturity/migration/migration_guide_2.8.rst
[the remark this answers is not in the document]
MirkoCovizzi: I moved the note to the first substitution (the parser initialization) so that it has higher visibility for the user.

lemrey (approved):
Good job!

SeppoTakalo (approved):

applications/serial_lte_modem/src/slm_at_commands.c
```diff
@@ -79,7 +78,7 @@ enum sleep_modes {
 };
 
 /** @brief AT command handler type. */
-typedef int (*slm_at_handler_t) (enum at_cmd_type);
+typedef int (*slm_at_handler_t) (enum at_parser_cmd_type);
```
MarkusLassila, line 81: This is a leftover from AT_CMD_CUSTOM change, it is not used. Feel free to remove.

applications/serial_lte_modem/src/nrfcloud/slm_at_nrfcloud.c
```diff
@@ -703,8 +705,7 @@ static int handle_at_nrf_cloud_pos(enum at_cmd_type cmd_type,
 
 			/* Parse the MAC address. */
 			len = sizeof(mac_addr_str);
-			err = at_params_string_get(
-				param_list, param_idx, mac_addr_str, &len);
+			err = util_string_get(parser, param_idx, mac_addr_str, &len);
 			if (!err && (len != sizeof(mac_addr_str)
```
MarkusLassila, line 709: Increase the space for mac_addr_str to fit the terminating zero and change the check:
```suggestion
			if (!err && (len != strlen(mac_addr_str)
```

MarkusLassila: With old AT-command parser we allowed skipping the null terminating character if it did not fit :facepalm:. I went over the places where `util_string_get` is used, but did not locate any other places where there definitively would not be space for 0.

applications/serial_lte_modem/src/slm_at_commands.c
MirkoCovizzi: Removed.

applications/serial_lte_modem/src/nrfcloud/slm_at_nrfcloud.c
MarkusLassila: @MirkoCovizzi: Still requires the +1 for buffer size, mentioned above: "Increase the space for mac_addr_str to fit the terminating zero"

MirkoCovizzi: Updated.

MarkusLassila (approved):
Excellent!

doc/nrf/libraries/modem/at_parser.rst
```diff
@@ -0,0 +1,98 @@
...
+#########
+
+.. contents::
+   :local:
+   :depth: 2
+
+Overview
+========
+
+The AT parser is a library that parses AT command responses, notifications, or events.
```
eivindj-nordic, line 13: nit: 
```suggestion
The AT parser is a library that parses AT command responses, notifications, and events.
```

eivindj-nordic (approved):

b-gent (approved):
Adding a commit with doc review, please squash it before merging this PR. There are also some conflicts to resolve.

anhmolt (approved):

PR #15122, opened 2024-04-29
doc: wifi: add derivative recertification process

richabp:
Updated Certification page with the information
about the derivative recertification process.

SHEL-2729


ioannisg (approved):
It looks OK to me, thanks. @bama-nordic can you please do a quick check and see if this is accurate?

kittydepa (approved):

PR #15125, opened 2024-04-29
samples: Bluetooth: Mesh: Fix documentation for Light Dimmer sample

akredalen:
The Light Dimmer sample documentation falsely stated that Thingy:53 only has one button and that therefore some functionality would not work on Thingy:53

### DIFF
--- a/samples/bluetooth/mesh/light_dimmer/README.rst
+++ b/samples/bluetooth/mesh/light_dimmer/README.rst
@@ -54,8 +54,8 @@ This mobile application is also used to configure key bindings, and publication 
 After provisioning and configuring the mesh models supported by the sample in the `nRF Mesh mobile app`_, **Button 1** on the Mesh Light Dimmer device can be used to control the configured network nodes' LEDs, while **Button 2** can be used to store and restore scenes on the network nodes.
 
 .. note::
-  When running this sample on Thingy:53 or the :ref:`zephyr:nrf52840dongle_nrf52840`, the scene selection functionality will not be available as the device only has one button.
-  The single button of Thingy:53 and the dongle will be used for dimming and the on/off functionality as described for **Button 1** in this documentation.
+  When running this sample on the :ref:`zephyr:nrf52840dongle_nrf52840`, the scene selection functionality will not be available as the device only has one button.
+  The single button of the dongle will be used for dimming and the on/off functionality as described for **Button 1** in this documentation.
 
 Provisioning
 ============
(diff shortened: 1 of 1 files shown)


ludvigsj:
Have you checked this on an actual Thingy53? I agree, it should work, but it's nice to just check quickly before changing the docs

Andrewpini:
I'm a bit confused. Does Thingy53 have more than one button?
IMG20240430094526

ludvigsj:
> I'm a bit confused. Does Thingy53 have more than one button? IMG20240430094526

Yes 
image

Andrewpini:
> > I'm a bit confused. Does Thingy53 have more than one button? IMG20240430094526
> 
> Yes image

Well I'll be damned!

akredalen:
The sample is tested and works as expected with Thingy:53.

HaavardRei (approved):
LGTM

Andrewpini (approved):

samples/bluetooth/mesh/light_dimmer/README.rst
```diff
@@ -117,7 +117,10 @@ Button 2:
    On long press and release, **Button 2** will publish a Scene Store message using the configured publication parameters of its model instance, and store the current LED state of all the targets under the scene with the most recently recalled scene number.
 
 .. note::
-   :ref:`zephyr:thingy53_nrf5340` and the :ref:`zephyr:nrf52840dongle_nrf52840` only support dimming and the on/off functionality, not the scene selection functionality.
+   The :ref:`zephyr:nrf52840dongle_nrf52840` only support dimming and the on/off functionality, not the scene selection functionality.
+
+.. note::
+  When testing the sample on Thingy:53, remember that **Button 2** can be accessed by removing the top part of the casing.
```
wiba-nordic, line 123:
```suggestion
   On Thingy:53, remember that **Button 2** can be accessed by removing the top part of the casing.
```

samples/bluetooth/mesh/light_dimmer/README.rst
```diff
@@ -117,7 +117,10 @@ Button 2:
    On long press and release, **Button 2** will publish a Scene Store message using the configured publication parameters of its model instance, and store the current LED state of all the targets under the scene with the most recently recalled scene number.
 
 .. note::
-   :ref:`zephyr:thingy53_nrf5340` and the :ref:`zephyr:nrf52840dongle_nrf52840` only support dimming and the on/off functionality, not the scene selection functionality.
+   The :ref:`zephyr:nrf52840dongle_nrf52840` only support dimming and the on/off functionality, not the scene selection functionality.
```
wiba-nordic, line 120:
```
   .. note::
      On the :ref:`zephyr:nrf52840dongle_nrf52840`, the scene selection functionality will not be available as the device only has one button.
```
Indented to be inside the button description, and used the same wording as the earlier note.

Could also move this note to be at the start of the User interface section so we don't have two notes back-to-back. In that case, I'd also add the second line from the earlier note for completeness' sake.

wiba-nordic (changes requested):

samples/bluetooth/mesh/light_dimmer/README.rst
akredalen: Going by your first suggestion, should both notes be indented then?
Or by your second suggestion, this note will be more or less identical to the one in the Overview then. Is that a good thing?

wiba-nordic: Ah, yes, Thingy:53 note (and looking at it, the note under LEDs) should be indented.

I just checked the built documentation for the PR, and it seems that if the nRF54840 Dongle note is moved to the start of User interface, it might be visible on the screen at the same time as the one in Overview. You are right, that is needless duplication. Considering that, I would suggest keeping the note where it is with just one line, and changing the note for Thingy:53 to a tip instead of note to differentiate it a bit. In this case we can remove "remember that" from the line as well so it's just:
```
      On Thingy:53, **Button 2** can be accessed by removing the top part of the casing.
```

wiba-nordic (approved):

PR #15140, opened 2024-04-30
fprotect rram basic support

michalek-no:
- [x] Requires https://github.com/nrfconnect/sdk-zephyr/pull/1693


michalek-no:
@sigvartmh please take a look if we can do something about nonhomogeneous PM/noPM partition names

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,54 @@
...
+		.lock = true,
+		.size_kb = size>>10,
+	};
+
+	nrf_rramc_region_config_set(NRF_RRAMC, 4, &config);
+}
+
+int fprotect_area(uint32_t start, size_t length)
+{
+	if (length <= 0x7c00) {
```
oyvindronningstad, line 40: This may be WIP, but please add comments explaining these if statements for posterity. The magic values should probably have defines also.

nvlsianpu: 31 kB region size limit. This need to be defined somehow.

lib/fprotect/Kconfig
```diff
@@ -55,7 +56,6 @@ menuconfig FPROTECT
 if FPROTECT
 config FPROTECT_APP
 	bool "Lock image partition using FPROTECT"
-	depends on PARTITION_MANAGER_ENABLED
```
nvlsianpu, line 58: is this intentional?
Yes.

subsys/fw_info/Kconfig
```diff
@@ -75,6 +75,8 @@ config FW_INFO_HARDWARE_ID
 	default 51 if SOC_SERIES_NRF51X
 	default 52 if SOC_SERIES_NRF52X
 	default 53 if SOC_SERIES_NRF53X
+	default 154 if SOC_SERIES_NRF54LX
```
nvlsianpu, line 78: how you get these numbers?

lib/fprotect/sys_init_fprotect.c
```diff
@@ -6,25 +6,32 @@
...
 #include <fprotect.h>
+#include <zephyr/storage/flash_map.h>
 
 #define PRIORITY_LEVEL 0 /* Locking should be performed as soon as possible */
 
-BUILD_ASSERT(PM_ADDRESS % CONFIG_FPROTECT_BLOCK_SIZE == 0,
+#ifdef CONFIG_PARTITION_MANAGER_ENABLED
+#define CODE_PARTITION app
+#else
+#define CODE_PARTITION slot0_partition
```
nvlsianpu, line 17: Isn't DT_CHOSEN(zephyr_code_partition) more generic assignment?

lib/fprotect/fprotect_rram.c
michalek-no: Yes, tbh this should be pulled from MDK if possible, especially bacause it's probably going to change.

subsys/fw_info/Kconfig
michalek-no: Going with established scheme meets the wall here. We have two 54's. Any better ideas are more than welcome.

lib/fprotect/sys_init_fprotect.c
nvlsianpu: Unless we have PM and DTS for partitioning - not.

subsys/fw_info/Kconfig
nvlsianpu: You can go with 54 for NRF54LX
For 54H this won't be supported

lib/fprotect/Kconfig
```diff
@@ -41,6 +41,7 @@ config FPROTECT_BLOCK_SIZE
 	default NRF_MPU_FLASH_REGION_SIZE if HAS_HW_NRF_MPU
 	default NRF_BPROT_FLASH_REGION_SIZE if HAS_HW_NRF_BPROT
 	default NRF_ACL_FLASH_REGION_SIZE if HAS_HW_NRF_ACL
+	default 0x400 if SOC_SERIES_NRF54LX
```
hubertmis, line 44: How about something like `NRF_RRAM_REGION_SIZE_GRANULARITY` instead of a magic number?

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,67 @@
...
+
+#include <stdint.h>
+#include <string.h>
+#include <errno.h>
+#include <soc.h>
+#include <nrf_erratas.h>
+#include <hal/nrf_rramc.h>
+
+#define kB_Pos (10)
+#define REGION_SIZE ((RRAMC_REGION_CONFIG_SIZE_Msk >> RRAMC_REGION_CONFIG_SIZE_Pos) << kB_Pos)
```
hubertmis, line 15: It's not obvious to this code reader that RRAMC REGION SIZE unit is a KB.

```suggestion
#define RRAMC_REGION_CONFIG_SIZE_UNIT 1024 /* bytes */
#define REGION_SIZE ((RRAMC_REGION_CONFIG_SIZE_Msk >> RRAMC_REGION_CONFIG_SIZE_Pos) * RRAMC_REGION_CONFIG_SIZE_UNIT)
```

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,67 @@
...
+	nrf_rramc_region_config_t config={
+		.address = 0,
+		.permissions = NRF_RRAMC_REGION_PERM_READ_MASK | NRF_RRAMC_REGION_PERM_EXECUTE_MASK,
+		.writeonce = true,
+		.lock = true,
+		.size_kb = size>>kB_Pos,
+	};
+
+	nrf_rramc_region_config_set(NRF_RRAMC, 3, &config);
+}
```
hubertmis, line 28: This should be done by HW using a UICR configuration to make an Immutable Bootloader code really immutable. Not run-time.

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,67 @@
...
+static void protect_boot_code2(uint32_t start, uint32_t size){
+	nrf_rramc_region_config_t config={
+		.address = start,
+		.permissions = NRF_RRAMC_REGION_PERM_READ_MASK | NRF_RRAMC_REGION_PERM_EXECUTE_MASK,
+		.writeonce = true,
+		.lock = true,
+		.size_kb = size>>kB_Pos,
```
hubertmis, line 35: What if `size` is not a multiplication of 1024? I guess division with rounding up should be used instead.

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,67 @@
...
+		}
+		if(start != 0){
+			/*
+			 * Region[3] has fixed 0x00000000 start address.
+			 * Using both regions implies this constraint.
+			 */
+			return -EINVAL;
+		}
+		protect_boot_code(REGION_SIZE);
+		protect_boot_code2(REGION_SIZE, length - REGION_SIZE);
```
hubertmis, line 64: This part seems invalid. The run-time code should modify only the region 4. Region 3 should be configured by hardware based on the Immutable Bootloader configuration in UICR.
```suggestion
		return -EINVAL;
```

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,67 @@
...
+static void protect_boot_code2(uint32_t start, uint32_t size){
+	nrf_rramc_region_config_t config={
+		.address = start,
+		.permissions = NRF_RRAMC_REGION_PERM_READ_MASK | NRF_RRAMC_REGION_PERM_EXECUTE_MASK,
+		.writeonce = true,
+		.lock = true,
+		.size_kb = size>>kB_Pos,
+	};
+
+	nrf_rramc_region_config_set(NRF_RRAMC, 4, &config);
```
hubertmis, line 38: Please, avoid magic numbers

```suggestion
	nrf_rramc_region_config_set(NRF_RRAMC, RRAMC_REGION_FOR_MUTABLE_BOOTLOADER, &config);
```

michalek-no:
rebase

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,41 @@
...
+	if( start % CONFIG_NRF_RRAM_REGION_ADDRESS_RESOLUTION ||
+	    size % CONFIG_NRF_RRAM_REGION_SIZE_UNIT ) {
+		/*
+		 * We avoid setting up locks from address lower than requested
+		 * which occurs due to difference in resolution.
+		 * Also we don't want to lock more than requested.
+		 */
+		return -EINVAL;
+	}
+	if (size <= REGION_SIZE_MAX) {
```
oyvindronningstad, line 28: IMO if the size is too large this should return an error.

hubertmis: Yeah, according to header file I understand it should be `-EINVAL`.

hubertmis, line 28: You should also check if RRAMC_REGION_FOR_FPROTECT is already in use. If it is, return -NOSPC

lib/fprotect/sys_init_fprotect.c
michalek-no: > Isn't DT_CHOSEN(zephyr_code_partition) more generic assignment?

needs some additional stuff - on it's way https://github.com/nrfconnect/sdk-zephyr/pull/1705

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,42 @@
...
+		/*
+		 * We avoid setting up locks from address lower than requested
+		 * which occurs due to difference in resolution.
+		 * Also we don't want to lock more than requested.
+		 */
+		return -EINVAL;
+	}
+	if (size > REGION_SIZE_MAX) {
+		return -EINVAL;
+	}
```
hubertmis, line 30: It seems this function is still missing checking if RRAMC_REGION_FOR_FPPROTECT is already in use. The function silently fails if the region is locked.

michalek-no: yes, I will resolve conversation when it is addressed.

michalek-no:
rebase and manifest removal

michalek-no:
rebase

michalek-no:
rebase

michalek-no:
rebase

michalek-no:
rebase

lib/fprotect/fprotect_rram.c
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+#include <stdint.h>
+#include <string.h>
+#include <errno.h>
+#include <soc.h>
+#include <nrf_erratas.h>
+#include <hal/nrf_rramc.h>
+
+#define RRAMC_REGION_UNTOUCHED 0x0000000f
```
hubertmis, line 14: Do I understand correctly this is a value of region 4 after a mass erase is performed? If that's the case how about renaming it to RRAMC_REGION_FOR_FPROTECT_DEFAULT_VALUE , _UNPROGRAMMED, _UNTOUCHED, or something like this

hubertmis (approved):

michalek-no:
rebase

nvlsianpu (approved):

west.yml
```diff
@@ -63,7 +63,7 @@ manifest:
     # https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/zephyr/guides/modules.html
     - name: zephyr
       repo-path: sdk-zephyr
-      revision: 2c2f60d1ebd959900405e448cee45c03ecdc6646
+      revision: pull/1825/head
```
nvlsianpu, line 66: Remove west.yaml path

nvlsianpu (approved):

michalek-no:
rebase

PR #15154, opened 2024-05-02
TF-M: Enable Protected Storage encryption and roll back protection

mswarowsky:


include/tfm/tfm_builtin_key_ids.h
```diff
@@ -7,7 +7,9 @@
 #ifndef __TFM_BUILTIN_KEY_IDS_H__
 #define __TFM_BUILTIN_KEY_IDS_H__
 
-#ifdef CONFIG_PSA_CRYPTO_DRIVER_CRACEN
```
SebastianBoe, line 10: What about 54L with

CONFIG_PSA_CRYPTO_DRIVER_CRACEN=n 
CONFIG_PSA_CRYPTO_DRIVER_OBERON=y 

or is that not supported?

SebastianBoe (approved):

vili-nordic (approved):

include/tfm/tfm_builtin_key_ids.h
mswarowsky: no supported

PR #15162, opened 2024-05-03
applications: asset_tracker_v2: Less WiFi APs

trantanen:
Compilation with nrf7002ek configuration had less than 200 bytes of SRAM free, which didn't allow much room for own tweaks. It's evident this would overflow in the near future so decreasing the number of requested WiFi APs from 30 to 20. It could be even 10 like with Thingy91X configuration.


jtguggedal (approved):

tokangas (approved):

maxd-nordic (approved):
I agree with removing the broken configurations. Thingy:91 X is only supported tentatively.

PR #15163, opened 2024-05-03
samples: wifi: softap: Fix synchronization issue

D-Triveni:
Add a sleep to address the issue where the sample doesn't wait for the WPA supplicant to be fully initialized before proceeding with Wi-Fi operations.


krish2718 (approved):

sachinthegreen (approved):

krish2718:
This is needed for any sample that uses regulatory domain setting e.g., Monitor mode, so, can you please do this for all in a single commit, easier to revert once we have a proper fix.

D-Triveni:
> This is needed for any sample that uses regulatory domain setting e.g., Monitor mode, so, can you please do this for all in a single commit, easier to revert once we have a proper fix.

Addressed Chaitanya.

VivekUppunda (approved):

krish2718 (approved):

lave30 (approved):

PR #15169, opened 2024-05-03
drivers: wifi: Create dedicated mem pool for Wi-Fi driver

rado17:
Create dedicated memory pools for Wi-Fi management and
data operations (defaults: 20KB for management and 60KB for data).

Reduce the heap memory pool size since the Wi-Fi driver will
be allocating memory from a dedicated pool.


drivers/wifi/nrf700x/src/shim.c
```diff
@@ -27,19 +28,73 @@
 #include "qspi_if.h"
 
 LOG_MODULE_REGISTER(wifi_nrf, CONFIG_WIFI_NRF700X_LOG_LEVEL);
+K_HEAP_DEFINE(wifi_drv_mgmt_mem_pool, CONFIG_NRF_WIFI_CONFIG_HEAP_MEM_POOL_SIZE);
```
krish2718, line 31:
```suggestion
K_HEAP_DEFINE(wifi_drv_ctrl_mem_pool, CONFIG_NRF_WIFI_CONFIG_HEAP_MEM_POOL_SIZE);
```

drivers/wifi/nrf700x/src/shim.c
```diff
@@ -27,19 +28,73 @@
...
+K_HEAP_DEFINE(wifi_drv_pkt_mem_pool, CONFIG_NRF_WIFI_PKT_HEAP_MEM_POOL_SIZE);
```
krish2718, line 32:
```suggestion
K_HEAP_DEFINE(wifi_drv_data_mem_pool, CONFIG_NRF_WIFI_PKT_HEAP_MEM_POOL_SIZE);
```

drivers/wifi/nrf700x/src/shim.c
```diff
@@ -27,19 +28,73 @@
...
 }
 
 static void *zep_shim_mem_zalloc(size_t size)
 {
+	void *ret;
+    size_t bounds;
+
+	size = (size + 4) & 0xfffffffc;
+
+	if (size_mul_overflow(size, sizeof(char), &bounds)) {
```
krish2718, line 55: is this a new check mandated by new API?

drivers/wifi/nrf700x/src/shim.c
krish2718, line 31: Should we use `K_HEAP_DEFINE_NOCACHE` instead? In case some system supports caching, I don't think our driver will work?

krish2718: Might also need to evaluate any alignment requirements, i.e., using `k_heap_aligned_alloc`?

samples/wifi/shell/overlay-zperf.conf
```diff
@@ -7,7 +7,7 @@ CONFIG_NET_PKT_RX_COUNT=30
 CONFIG_NET_PKT_TX_COUNT=30
 CONFIG_NET_BUF_RX_COUNT=30
 CONFIG_NET_BUF_TX_COUNT=60
-CONFIG_HEAP_MEM_POOL_SIZE=230000
+CONFIG_HEAP_MEM_POOL_SIZE=140000
```
krish2718, line 10: Shouldn't this now be zero? IIRC, this requirement is purely from Wi-Fi driver and now that we are using dedicated pools, this shouldn't be used. And instead we need to use the newly introduced defines here.

krish2718 (changes requested):

drivers/wifi/nrf700x/src/shim.c
rado17: This is taken from the checks that were being used in `k_calloc` call.

samples/wifi/shell/overlay-zperf.conf
rado17: Have reduced the `CONFIG_HEAP_MEM_POOL_SIZE` after introduction of dedicated heap pools for driver. This is to accommodate other subsystems which use `k_malloc`.

drivers/wifi/nrf700x/src/shim.c
rado17: Done

drivers/wifi/nrf700x/src/shim.c
rado17: Done

drivers/wifi/nrf700x/src/shim.c
rado17: Done

krish2718:
Migrated to upstream https://github.com/zephyrproject-rtos/zephyr/pull/78588

PR #15188, opened 2024-05-06
bluetooth: samples: add iso_rx disconnect cb.

henrla:
Add callback that is triggered when an RX ISO channel disconnects. The usecase is to start scanning for a new connection. Scanning for a new connection is also done in the ACL disconnection callback in this sample, but the ISO channel might disconnect after the ACL, which will prevent the central from starting to scan for a new ISO connection.

### DIFF
--- a/samples/bluetooth/iso_time_sync/src/cis_central.c
+++ b/samples/bluetooth/iso_time_sync/src/cis_central.c
@@ -100,7 +100,16 @@ static void scan_start(void)
 
 	if (free_iso_chan_find()) {
 		err = bt_le_scan_start(BT_LE_SCAN_ACTIVE_CONTINUOUS, NULL);
-		if (err) {
+		if (err == -EALREADY) {
+			/** If the central is RXing, both the ISO and the ACL
+			 * disconnection callbacks try to enable the scanner.
+			 * If the ISO channel disconnects before the ACL
+			 * connection, the application will attempt to enable
+			 * the scanner again.
+			 */
+			printk("Scanning did not start because it has already started (err %d)\n",
+				   err);
+		} else if (err) {
 			printk("Scanning failed to start (err %d)\n", err);
 			return;
 		}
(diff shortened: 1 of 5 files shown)


samples/bluetooth/iso_time_sync/src/cis_central.c
```diff
@@ -183,7 +183,13 @@ void cis_central_start(bool do_tx, uint8_t retransmission_number, uint16_t max_t
...
 		iso_tx_init(retransmission_number, &scan_start);
 	} else {
-		iso_rx_init(retransmission_number);
+		/** scan_start is registered as a callback that is triggered when an
+		 * ISO channel disconnects. This is needed because the ISO channel might
+		 * disconnect after the ACL has completed its disconnection callback. If
+		 * this is the case, there are no RX ISO channels available during the
+		 * ACL disconnection callback, which will prevent scanning from starting.
+		 */
+		iso_rx_init(retransmission_number, &scan_start);
```
Tschet1, line 192: I guess this has the side-effect that we sometimes would print https://github.com/nrfconnect/sdk-nrf/blob/c50248c3788c74bbf4fc4c0c349bde9fdd2c2a80/samples/bluetooth/iso_time_sync/src/cis_central.c#L104C12-L104C45 , if ISO is disconnected before the ACL, right?

Tschet1 (approved):

samples/bluetooth/iso_time_sync/src/cis_central.c
henrla: I think you are right. I can add a check if the error returns that the scanner is already enabled.

rugeGerritsen (approved):

samples/bluetooth/iso_time_sync/src/cis_central.c
```diff
@@ -100,7 +100,16 @@ static void scan_start(void)
 
 	if (free_iso_chan_find()) {
 		err = bt_le_scan_start(BT_LE_SCAN_ACTIVE_CONTINUOUS, NULL);
-		if (err) {
+		if (err == -EALREADY) {
+			/** If the central is RXing, both the ISO and the ACL
+			 * disconnection callbacks tries to enable the scanner.
```
Tschet1, line 105:
```suggestion
			 * disconnection callbacks try to enable the scanner.
```

Tschet1 (approved):

PR #15192, opened 2024-05-07
applications: nrf5340_audio: Optionally include nRF5340 ADK features

andvib:
Optionally include HW codec, power measurement and SD card features as these are tied to the nRF5340 audio DK and not the audio application. Move initializations of high frequency clock, button handler and channel assignments from audio DK module and to application.

### DIFF
--- a/applications/nrf5340_audio/boards/nrf5340_audio_dk_nrf5340_cpuapp.conf
+++ b/applications/nrf5340_audio/boards/nrf5340_audio_dk_nrf5340_cpuapp.conf
@@ -1,0 +1,9 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+CONFIG_NRF5340_AUDIO_CS47L63_DRIVER=y
+CONFIG_NRF5340_AUDIO_POWER_MEASUREMENT=y
+CONFIG_NRF5340_AUDIO_SD_CARD_MODULE=y
--- a/applications/nrf5340_audio/src/drivers/CMakeLists.txt
+++ b/applications/nrf5340_audio/src/drivers/CMakeLists.txt
@@ -4,6 +4,6 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
-target_sources(app PRIVATE
-	       ${CMAKE_CURRENT_SOURCE_DIR}/cs47l63_comm.c
+target_sources_ifdef(CONFIG_NRF5340_AUDIO_CS47L63_DRIVER
+		     app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/cs47l63_comm.c
 )
--- a/applications/nrf5340_audio/src/drivers/Kconfig
+++ b/applications/nrf5340_audio/src/drivers/Kconfig
@@ -6,8 +6,13 @@
 
 menu "Drivers"
 
-#----------------------------------------------------------------------------#
-menu "Thread priorities"
+menuconfig NRF5340_AUDIO_CS47L63_DRIVER
+	bool "CS47L63 HW codec driver"
+	select HW_CODEC_CIRRUS_LOGIC
+	help
+	  Include the driver for the Cirrus Logic CS47L63 hardware codec chip
+
+if NRF5340_AUDIO_CS47L63_DRIVER
 
 config CS47L63_THREAD_PRIO
 	int "Priority for CS47L63 thread"
@@ -15,23 +20,13 @@ config CS47L63_THREAD_PRIO
 	help
 	  This is a preemptible thread
 
-endmenu # Thread priorities
-
-#----------------------------------------------------------------------------#
-menu "Stack sizes"
-
 config CS47L63_STACK_SIZE
 	int "Stack size for CS47L63"
 	default 700
-
-endmenu # Stack sizes
-
-#----------------------------------------------------------------------------#
-menu "Log levels"
 
 module = CS47L63
 module-str = cs47l63
 source "subsys/logging/Kconfig.template.log_config"
 
-endmenu # Log levels
+endif # NRF5340_AUDIO_CS47L63_DRIVER
 endmenu # Drivers
--- a/applications/nrf5340_audio/src/modules/CMakeLists.txt
+++ b/applications/nrf5340_audio/src/modules/CMakeLists.txt
@@ -4,20 +4,17 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
-set(SRCS
-  ${CMAKE_CURRENT_SOURCE_DIR}/audio_i2s.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/audio_usb.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/button_handler.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/hw_codec.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/led.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/power_meas.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/sd_card.c
-)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/audio_i2s.c)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/audio_usb.c)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/button_handler.c)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/led.c)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/audio_sync_timer_rtc.c)
 
-list(APPEND SRCS ${CMAKE_CURRENT_SOURCE_DIR}/audio_sync_timer_rtc.c)
-
-if (CONFIG_SD_CARD_PLAYBACK)
-  list(APPEND SRCS ${CMAKE_CURRENT_SOURCE_DIR}/sd_card_playback.c)
-endif()
-
-target_sources(app PRIVATE ${SRCS})
+target_sources_ifdef(CONFIG_NRF5340_AUDIO_CS47L63_DRIVER app PRIVATE
+                     ${CMAKE_CURRENT_SOURCE_DIR}/hw_codec.c)
+target_sources_ifdef(CONFIG_NRF5340_AUDIO_POWER_MEASUREMENT app PRIVATE
+                     ${CMAKE_CURRENT_SOURCE_DIR}/power_meas.c)
+target_sources_ifdef(CONFIG_NRF5340_AUDIO_SD_CARD_MODULE app PRIVATE
+                     ${CMAKE_CURRENT_SOURCE_DIR}/sd_card.c)
+target_sources_ifdef(CONFIG_SD_CARD_PLAYBACK app PRIVATE
+                     ${CMAKE_CURRENT_SOURCE_DIR}/sd_card_playback.c)
--- a/applications/nrf5340_audio/src/modules/Kconfig
+++ b/applications/nrf5340_audio/src/modules/Kconfig
@@ -18,7 +18,12 @@ config AUDIO_SYNC_TIMER_USES_RTC
 	select NRFX_RTC0
 
 #----------------------------------------------------------------------------#
-menu "Power measurement"
+menuconfig NRF5340_AUDIO_POWER_MEASUREMENT
+	bool "Power measurement"
+	help
+	  Include the power measurement driver for the nRF5340 Audio DK
+
+if NRF5340_AUDIO_POWER_MEASUREMENT
 
 config POWER_MEAS_INTERVAL_MS
 	int "Power measurement interval in milliseconds"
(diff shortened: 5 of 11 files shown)


applications/nrf5340_audio/broadcast_sink/main.c
```diff
@@ -451,9 +452,14 @@ int main(void)
 
 	LOG_DBG("nRF5340 APP core started");
 
+	channel_assignment_init();
```
rick1082, line 455: Could we bring this below `nrf5340_audio_dk_init()`?

applications/nrf5340_audio/src/nrf5340_audio_common.c
```diff
@@ -26,5 +27,12 @@ int nrf5340_audio_common_init(void)
 		return ret;
 	}
 
+	/* Use this to turn on 128 MHz clock for cpu_app */
```
rick1082, line 30: Nitpick: Maybe we could move this part to the beginning of the code for speed up a tiny little bit.

rick1082 (approved):

applications/nrf5340_audio/broadcast_sink/main.c
andvib: Unfortunately not, `nrf5340_audio_dk_init()` calls `channel_assignment_set()`, so I assume the channel assignment module should be initialized before that call

applications/nrf5340_audio/src/drivers/CMakeLists.txt
```diff
@@ -4,6 +4,6 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
-target_sources(app PRIVATE
-	       ${CMAKE_CURRENT_SOURCE_DIR}/cs47l63_comm.c
+target_sources_ifdef(CONFIG_HW_CODEC_CIRRUS_LOGIC
```
koffes, line 7: Suggest we also rename:
```suggestion
target_sources_ifdef(CONFIG_HW_CODEC_CS47L63
```

applications/nrf5340_audio/src/drivers/Kconfig
```diff
@@ -6,32 +6,27 @@
 
 menu "Drivers"
 
-#----------------------------------------------------------------------------#
-menu "Thread priorities"
+menuconfig NRF5340_AUDIO_CS47L63_DRIVER
+	bool "CS47L63 codec driver"
```
koffes, line 10:
```suggestion
	bool "CS47L63 HW codec driver"
```

applications/nrf5340_audio/src/nrf5340_audio_common.c
andvib: That's a good idea! I put it in this module as it is common for all the applications, but maybe for this reason it's better to initialize the clock at the beginning of each `main.c` file? I'm not sure if the `nrf5340_audio_common_init()` call can come before the other init's 🤔

applications/nrf5340_audio/src/utils/nrf5340_audio_dk.c
```diff
@@ -157,11 +134,5 @@ int nrf5340_audio_dk_init(void)
 		}
 	}
 
-	ret = audio_system_init();
```
koffes, line 160: I may see this later, but it seems natural that this call remains here.

applications/nrf5340_audio/src/drivers/Kconfig
```diff
@@ -6,32 +6,27 @@
 
 menu "Drivers"
 
-#----------------------------------------------------------------------------#
-menu "Thread priorities"
+menuconfig NRF5340_AUDIO_CS47L63_DRIVER
```
koffes, line 9: Can we somehow combine CONFIG_NRF5340_AUDIO_CS47L63_DRIVER and CONFIG_HW_CODEC_CIRRUS_LOGIC/CONFIG_HW_CODEC_CS47L63 ?

applications/nrf5340_audio/broadcast_sink/main.c
```diff
@@ -451,9 +452,14 @@ int main(void)
...
+
 	ret = nrf5340_audio_dk_init();
 	ERR_CHK(ret);
 
+	ret = audio_system_init();
```
koffes, line 460: These two calls are now done in all four mains. Discuss if we can put this somewhere more generic.

alexsven: It should be moved into nrf5340_audio_common_init then if is there for all four main files

applications/nrf5340_audio/broadcast_sink/main.c
alexsven: What about moving the `channel_assignment_init` into `nrf5340_audio_common_init` and then move that above the `nrf5340_audio_dk_init`? I don't see anything right off the bat that should prevent that at least.

applications/nrf5340_audio/src/drivers/CMakeLists.txt
alexsven: I concur

applications/nrf5340_audio/src/modules/CMakeLists.txt
```diff
@@ -4,20 +4,16 @@
...
-  ${CMAKE_CURRENT_SOURCE_DIR}/hw_codec.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/led.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/power_meas.c
-  ${CMAKE_CURRENT_SOURCE_DIR}/sd_card.c
-)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/audio_i2s.c)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/audio_usb.c)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/button_handler.c)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/led.c)
+target_sources(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/sd_card.c)
```
alexsven, line 11: Shouldn't the SD card be dependent on the nRF5340_audio_DK somehow?

applications/nrf5340_audio/broadcast_sink/main.c
andvib: I moved `audio_system_init()` out of `nrf5340_audio_dk_init()` as it is needed regardless if the application is running on the audio DK or not, so I think it makes sense to have it in each `main.c`. Either way it won't matter functionally in this PR, so if you think it's better to keep `nrf5340_audio_dk` as-is I can move it back and we can discuss the topic when the application has full support for other DKs

applications/nrf5340_audio/src/modules/Kconfig
```diff
@@ -18,7 +18,12 @@ config AUDIO_SYNC_TIMER_USES_RTC
 	select NRFX_RTC0
 
 #----------------------------------------------------------------------------#
-menu "Power measurement"
+menuconfig NRF5340_AUDIO_POWER_MEASURMENT
+	bool "Power measurement"
+	help
+	  Include the power measurment driver for the nRF5340 Audio DK
```
alexsven, line 24:
```suggestion
	  Include the power measurement driver for the nRF5340 Audio DK
```

applications/nrf5340_audio/src/drivers/CMakeLists.txt
andvib: You mean to rename `NRF5340_AUDIO_CS47L63_DRIVER` to `CONFIG_HW_CODEC_CS47L63`?

Anyway this line should definitely be checking `NRF5340_AUDIO_CS47L63_DRIVER` and not `HW_CODEC_CIRRUS_LOGIC`, will update

applications/nrf5340_audio/src/nrf5340_audio_common.c
alexsven: I think it can, as mentioned in a comment earlier.

alexsven (changes requested):

applications/nrf5340_audio/src/modules/CMakeLists.txt
andvib: Good point - it should be possible to not compile `sd_card.c`, but I'm not sure if audio DK is the correct way to filter it though. If a kit has a different SD card slot the same code could be used, right?

applications/nrf5340_audio/broadcast_sink/main.c
andvib: This worked out pretty well! The only blocker I hit was calling `nrf5340_audio_common_init` before `nrf5340_audio_dk_init`, as the `bt_mgmt_init()` call in audio common requires the button handler to be initialized. So I kept the order, but moved `channel_assignment_init()` and `audio_system_init()` into audio common.

applications/nrf5340_audio/broadcast_sink/main.c
andvib: Ignore previous comment, I've moved `audio_system_init()` into `nrf5340_audio_common_init()`

applications/nrf5340_audio/src/nrf5340_audio_common.c
andvib: Moved to top of function, but not to top of `main()` because of blockers explained in other comment

applications/nrf5340_audio/src/modules/CMakeLists.txt
andvib: Added a new kconfig `NRF5340_AUDIO_SD_CARD_MODULE` that is set automatically for the audio DK via the board overlays

applications/nrf5340_audio/src/modules/Kconfig
```diff
@@ -18,7 +18,12 @@ config AUDIO_SYNC_TIMER_USES_RTC
 	select NRFX_RTC0
 
 #----------------------------------------------------------------------------#
-menu "Power measurement"
+menuconfig NRF5340_AUDIO_POWER_MEASURMENT
```
alexsven, line 21:
```suggestion
menuconfig NRF5340_AUDIO_POWER_MEASUREMENT
```

applications/nrf5340_audio/src/modules/Kconfig
```diff
@@ -36,7 +41,7 @@ config POWER_MEAS_START_ON_BOOT
 	  the voltage, current consumption, and power usage for the
 	  following rails: VBAT, VDD1_CODEC, VDD2_CODEC, and VDD2_NRF
 
-endmenu # Power measurement
+endif # NRF5340_AUDIO_POWER_MEASURMENT
```
alexsven, line 44:
```suggestion
endif # NRF5340_AUDIO_POWER_MEASUREMENT
```

applications/nrf5340_audio/src/modules/CMakeLists.txt
```diff
@@ -4,20 +4,17 @@
...
-list(APPEND SRCS ${CMAKE_CURRENT_SOURCE_DIR}/audio_sync_timer_rtc.c)
-
-if (CONFIG_SD_CARD_PLAYBACK)
-  list(APPEND SRCS ${CMAKE_CURRENT_SOURCE_DIR}/sd_card_playback.c)
-endif()
-
-target_sources(app PRIVATE ${SRCS})
+target_sources_ifdef(CONFIG_NRF5340_AUDIO_CS47L63_DRIVER app PRIVATE
+                     ${CMAKE_CURRENT_SOURCE_DIR}/hw_codec.c)
+target_sources_ifdef(CONFIG_NRF5340_AUDIO_POWER_MEASURMENT app PRIVATE
```
alexsven, line 15:
```suggestion
target_sources_ifdef(CONFIG_NRF5340_AUDIO_POWER_MEASUREMENT app PRIVATE
```

applications/nrf5340_audio/boards/nrf5340_audio_dk_nrf5340_cpuapp.conf
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+CONFIG_NRF5340_AUDIO_CS47L63_DRIVER=y
+CONFIG_NRF5340_AUDIO_POWER_MEASURMENT=y
```
alexsven, line 8:
```suggestion
CONFIG_NRF5340_AUDIO_POWER_MEASUREMENT=y
```

applications/nrf5340_audio/src/nrf5340_audio_common.c
```diff
@@ -4,8 +4,12 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 
+#include "audio/audio_system.h"
```
alexsven, line 7:
```suggestion
#include "audio_system.h"
```

applications/nrf5340_audio/src/nrf5340_audio_common.c
```diff
@@ -4,8 +4,12 @@
...
 #include "bt_mgmt.h"
+#include "channel_assignment.h"
 #include "fw_info_app.h"
+#include "nrfx_clock.h"
```
alexsven, line 11:
```suggestion
#include <nrfx_clock.h>
```
and then move up to differentiate "our" header files vs system header files

alexsven (changes requested):

alexsven (approved):

koffes (approved):

FARLY7:
Hi @andvib / @alexsven / @koffes,

Does this mean the nrf5340_audio application is buildable for the nrf5340dk? I tried to build with:

`west build -b nrf5340dk/nrf5340/cpuapp -p -- -DCONFIG_AUDIO_DEV=2`

but get errors (attached) regarding missing SW4 pin.

I would like to use the Broadcast Source configuration and supply the audio over USB, which should be possible on the nrf5340dk.

[nrf5340_audio_build_error.txt](https://github.com/nrfconnect/sdk-nrf/files/15402743/nrf5340_audio_build_error.txt)

PR #15203, opened 2024-05-07
lib: lte_link_control: Avoid unnecessary eDRX enable

trantanen:
lte_lc used to enable eDRX with AT+CEDRXS and send AT%XPTW immediately after. This causes an unnecessary TAU when PTW is set same value that network reported.
Now AT%XPTW is not sent if PTW value given by the network is the one we are supposed to request.

Jira: NRF91-2041

TODO:

- [x] Update lte_lc tests
- [x] Add new tests to cover the new logic in lte_lc_api tests
- [x] Test scenarios where eDRX parameters are set in offline mode and then eDRX enabled and that parameters are applied

### DIFF
--- a/lib/lte_link_control/lte_lc.c
+++ b/lib/lte_link_control/lte_lc.c
@@ -43,6 +43,9 @@ LOG_MODULE_REGISTER(lte_lc, CONFIG_LTE_LINK_CONTROL_LOG_LEVEL);
 		LTE_LC_SYSTEM_MODE_LTEM_NBIOT_GPS		: \
 	LTE_LC_SYSTEM_MODE_DEFAULT)
 
+/* Length for eDRX and PTW values */
+#define LTE_LC_EDRX_VALUE_LEN 5
+
 /* Internal enums */
 
 enum feaconf_oper {
@@ -57,18 +60,26 @@ enum feaconf_feat {
 
 /* Static variables */
 
-/* Request eDRX to be disabled */
-static const char edrx_disable[] = "AT+CEDRXS=3";
-/* Default eDRX setting */
-static char edrx_param_ltem[5] = CONFIG_LTE_EDRX_REQ_VALUE_LTE_M;
-static char edrx_param_nbiot[5] = CONFIG_LTE_EDRX_REQ_VALUE_NBIOT;
-/* Default PTW setting */
-static char ptw_param_ltem[5] = CONFIG_LTE_PTW_VALUE_LTE_M;
-static char ptw_param_nbiot[5] = CONFIG_LTE_PTW_VALUE_NBIOT;
-/* Default PSM RAT setting */
-static char psm_param_rat[9] = CONFIG_LTE_PSM_REQ_RAT;
-/* Default PSM RPTAU setting */
-static char psm_param_rptau[9] = CONFIG_LTE_PSM_REQ_RPTAU;
+/* Previously received LTE mode as indicated by the modem */
+static enum lte_lc_lte_mode prev_lte_mode = LTE_LC_LTE_MODE_NONE;
+/* Requested eDRX state (enabled/disabled) */
+static bool requested_edrx_enable;
+/* Requested eDRX setting */
+static char requested_edrx_value_ltem[LTE_LC_EDRX_VALUE_LEN] = CONFIG_LTE_EDRX_REQ_VALUE_LTE_M;
+static char requested_edrx_value_nbiot[LTE_LC_EDRX_VALUE_LEN] = CONFIG_LTE_EDRX_REQ_VALUE_NBIOT;
+/* Requested PTW setting */
+static char requested_ptw_value_ltem[LTE_LC_EDRX_VALUE_LEN] = CONFIG_LTE_PTW_VALUE_LTE_M;
+static char requested_ptw_value_nbiot[LTE_LC_EDRX_VALUE_LEN] = CONFIG_LTE_PTW_VALUE_NBIOT;
+/* Currently used eDRX setting as indicated by the modem */
+static char edrx_value_ltem[LTE_LC_EDRX_VALUE_LEN];
+static char edrx_value_nbiot[LTE_LC_EDRX_VALUE_LEN];
+/* Currently used PTW setting as indicated by the modem */
+static char ptw_value_ltem[LTE_LC_EDRX_VALUE_LEN];
+static char ptw_value_nbiot[LTE_LC_EDRX_VALUE_LEN];
+/* Requested PSM RAT setting */
+static char requested_psm_param_rat[9] = CONFIG_LTE_PSM_REQ_RAT;
+/* Requested PSM RPTAU setting */
+static char requested_psm_param_rptau[9] = CONFIG_LTE_PSM_REQ_RPTAU;
 /* Request PSM to be disabled and timers set to default values */
 static const char psm_disable[] = "AT+CPSMS=";
 /* Enable CSCON (RRC mode) notifications */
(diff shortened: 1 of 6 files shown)


lib/lte_link_control/lte_lc.c
```diff
@@ -341,17 +393,26 @@ static void at_handler_cedrxp(const char *response)
...
-	err = parse_edrx(response, &evt.edrx_cfg);
+	err = parse_edrx(response, &evt.edrx_cfg, current_edrx_value, current_ptw_value);
 	if (err) {
 		LOG_ERR("Can't parse eDRX, error: %d", err);
 		return;
 	}
 
+	/* PTW must be requested after eDRX is enabled */
+	lte_lc_edrx_params_store(evt.edrx_cfg.mode, current_edrx_value, current_ptw_value);
+	/* Sen PTW setting if eDRX is enabled, i.e., we have network mode */
```
tokangas, line 411:
```suggestion
	/* Send PTW setting if eDRX is enabled, i.e., we have network mode */
```

lib/lte_link_control/lte_lc.c
```diff
@@ -1122,11 +1181,12 @@ int lte_lc_edrx_get(struct lte_lc_edrx_cfg *edrx_cfg)
 		return -EFAULT;
 	}
 
-	err = parse_edrx(response, edrx_cfg);
+	err = parse_edrx(response, edrx_cfg, current_edrx_value, current_ptw_value);
 	if (err) {
 		LOG_ERR("Failed to parse eDRX parameters, error: %d", err);
 		return -EBADMSG;
 	}
+	lte_lc_edrx_params_store(prev_lte_mode, current_edrx_value, current_ptw_value);
```
tokangas, line 1189: This should be unnecessary, but I guess it doesn't hurt either.

lib/lte_link_control/lte_lc.c
```diff
@@ -175,6 +185,49 @@ static void lte_lc_psm_get_work_fn(struct k_work *work_item)
...
 }
 
+static void lte_lc_edrx_params_store(
+	enum lte_lc_lte_mode mode,
+	char *current_edrx_value,
+	char *current_ptw_value)
+{
+	switch (mode) {
+	case LTE_LC_LTE_MODE_LTEM:
+		memcpy(current_edrx_value_ltem, current_edrx_value, 5);
```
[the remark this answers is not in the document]
trantanen: Changed to strcpy

lib/lte_link_control/lte_lc.c
```diff
@@ -57,18 +60,26 @@ enum feaconf_feat {
...
-static char psm_param_rat[9] = CONFIG_LTE_PSM_REQ_RAT;
-/* Default PSM RPTAU setting */
-static char psm_param_rptau[9] = CONFIG_LTE_PSM_REQ_RPTAU;
+/* Requested eDRX setting */
+static char requested_edrx_value_ltem[LTE_LC_EDRX_VALUE_LEN] = CONFIG_LTE_EDRX_REQ_VALUE_LTE_M;
+static char requested_edrx_value_nbiot[LTE_LC_EDRX_VALUE_LEN] = CONFIG_LTE_EDRX_REQ_VALUE_NBIOT;
+/* Requested PTW setting */
+static char requested_ptw_value_ltem[LTE_LC_EDRX_VALUE_LEN] = CONFIG_LTE_PTW_VALUE_LTE_M;
+static char requested_ptw_value_nbiot[LTE_LC_EDRX_VALUE_LEN] = CONFIG_LTE_PTW_VALUE_NBIOT;
+/* Currently set eDRX setting as indicated by the modem */
```
tokangas, line 73: I would make it more explicit that this is the setting currently used. One could understand this so, that this is the value set using `+CEDRXS`, which is not necessarily the same as the value granted by the network. It might be enough to change the word `set` to `used`.

lib/lte_link_control/lte_lc.c
```diff
@@ -57,18 +60,26 @@ enum feaconf_feat {
...
+static char edrx_value_ltem[LTE_LC_EDRX_VALUE_LEN];
+static char edrx_value_nbiot[LTE_LC_EDRX_VALUE_LEN];
+/* Currently set PTW setting as indicated by the modem */
```
tokangas, line 76: Same here.

lib/lte_link_control/lte_lc.c
```diff
@@ -175,6 +188,51 @@ static void lte_lc_psm_get_work_fn(struct k_work *work_item)
...
+		strcpy(edrx_value_ltem, edrx_value);
+		strcpy(ptw_value_ltem, ptw_value);
+		break;
+	case LTE_LC_LTE_MODE_NBIOT:
+		strcpy(edrx_value_nbiot, edrx_value);
+		strcpy(ptw_value_nbiot, ptw_value);
+		break;
+	default:
+		memset(edrx_value_ltem, 0, sizeof(edrx_value_ltem));
+		memset(ptw_value_ltem, 0, sizeof(ptw_value_ltem));
```
tokangas, line 207: Hmm, this should probably also clear `edrx_value_nbiot` and `ptw_value_nbiot`?

Now that I think about it, these values can only be valid for one RAT/LTE mode at a time. So, when mode is LTE-M, should the NB-IoT values be cleared? And vice versa. When switching between modes, the current implementation keeps used values for the old mode. But, we don't know if the same values will be used if modem switches back to that mode. I'm not sure how modem works in this case, though. For example, does it send a new +CEDRXP notification for LTE-M when switching from LTE-M -> NB-IoT -> LTE-M, if the eDRX parameters in LTE-M remain the same?

lib/lte_link_control/lte_lc.c
```diff
@@ -1070,19 +1137,29 @@ int lte_lc_edrx_req(bool enable)
 			LOG_ERR("Failed to disable eDRX, reported error: %d", err);
 			return -EFAULT;
 		}
+		memset(edrx_value_ltem, 0, sizeof(edrx_value_ltem));
+		memset(ptw_value_ltem, 0, sizeof(ptw_value_ltem));
```
tokangas, line 1141: NB?

lib/lte_link_control/lte_lc.c
trantanen: > Hmm, this should probably also clear edrx_value_nbiot and ptw_value_nbiot?

Done

lib/lte_link_control/lte_lc.c
```diff
@@ -341,17 +405,28 @@ static void at_handler_cedrxp(const char *response)
...
 	if (err) {
 		LOG_ERR("Can't parse eDRX, error: %d", err);
 		return;
 	}
 
+	/* PTW must be requested after eDRX is enabled */
+	lte_lc_edrx_values_store(evt.edrx_cfg.mode, edrx_value, ptw_value);
+	/* Send PTW setting if eDRX is enabled, i.e., we have network mode */
+	if (evt.edrx_cfg.mode != LTE_LC_LTE_MODE_NONE) {
+		edrx_enabled = true;
```
tokangas, line 425: I thought that `edrx_enabled` would be set based on whether eDRX was requested by the application (`lte_lc_psm_req()`), not what the network has granted. Modem could be powered off in a cell which does not support eDRX, so `edrx_enabled` would be `false`. If modem is then powered on and camps on a cell which does support eDRX, there would be no unsolicited `+CEDRXP` notification even though eDRX would be used.

lib/lte_link_control/lte_lc.c
```diff
@@ -1122,15 +1198,38 @@ int lte_lc_edrx_get(struct lte_lc_edrx_cfg *edrx_cfg)
...
+NRF_MODEM_LIB_ON_CFUN(lte_lc_edrx_cfun_hook, lte_lc_edrx_on_modem_cfun, NULL);
+
+static void lte_lc_edrx_on_modem_cfun(int mode, void *ctx)
+#endif
+{
+	ARG_UNUSED(ctx);
+
+	/* If eDRX is enabled and modem is going to power off, subscription of unsolicited eDRX
+	 * notifications must be re-newed because modem forgets that information
+	 * although it stores eDRX value and PTW for both syste modes.
```
tokangas, line 1224:
```suggestion
	/* If eDRX is enabled and modem is powered off, subscription of unsolicited eDRX
	 * notifications must be re-newed because modem forgets that information
	 * although it stores eDRX value and PTW for both system modes.
```

lib/lte_link_control/lte_lc.c
trantanen: True. Should be fixed now.

tokangas (approved):

jhirsi (approved):

PR #15214, opened 2024-05-07
samples: matter: Introduce Watchdog for Matter

ArekBalysNordic:
You can create several watchdog instances (each uses its own channel) and feed them in different places in the code.

For example in the Matter sample one watchdog instance can be dedicated to the Main Thread, and another to the Matter Thread.

TODO:

- [x] Release note
- [x] A proper handling of the Test Event for Mainloop
- [x] Page in Matter docs


samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,128 @@
...
+	 * @param minWindowTime Minimum window Watchdog timeout for Watchdog object in milliseconds.
+	 * @param maxWindowTime Maximum window watchdog timeout for Watchdog object in milliseconds.
+	 * @param feedingInterval if the callback argument is provided then after elapsing time in milliseconds of this
+	 * value the callback will be set.
+	 * @param callback FeedingCallback to be called in the specific place to reset the Watchdog timer
+	 */
+	Watchdog(uint32_t minWindowTime, uint32_t maxWindowTime, uint32_t feedingInterval, FeedingCallback callback);
+
+	/**
+	 * @brief Disable and remove an Watchdog object
```
markaj-nordic, line 68: Currently the destructor does nothing.

markaj-nordic, line 68:
```suggestion
	 * @brief Disable and remove a Watchdog object
```

samples/matter/template/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
```diff
@@ -49,3 +49,7 @@
 &uart30 {
 	status = "okay";
 };
+
+&wdt30 {
```
markaj-nordic, line 53: So this is already enabled for other boards, right?

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,128 @@
...
+ * while the new Watchdog object is creating the global Watchdog instance creates a new channel.
+ * Each watchdog object contains own window ranges so all objects may work separately and be fed in a
+ * application-specific way.
+ *
+ * You can feed a Watchdog object in a two ways:
+ * 	- Manually, by calling the Feed() method in a specific context.
+ *	- Automatically, by providing callback function and feed interval value to the constructor as an argument.
+ * 	  If the callback is provided it will be called in each feed interval time periodically.
+ *
+ * @warning All Watchdog object must be created before calling Enable static method.
```
markaj-nordic, line 30:
```suggestion
 * @warning All Watchdog object must be created before calling Enable() static method.
```

markaj-nordic (commented):
Initial review

samples/matter/common/src/event_triggers/default_event_triggers.cpp
```diff
@@ -53,6 +54,42 @@ CHIP_ERROR RebootCallback(TestEventTrigger::TriggerValue delayMs)
 	return CHIP_NO_ERROR;
 }
 
+#ifdef CONFIG_NCS_SAMPLE_MATTER_WATCHDOG_EVENT_TRIGGERS
+CHIP_ERROR BlockMainThreadCallback(TestEventTrigger::TriggerValue blockingTimeMs)
+{
+	Nrf::PostTask([blockingTimeMs] {
+		int64_t startTime = k_uptime_get();
```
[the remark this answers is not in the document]
ArekBalysNordic: That's right but I would like to split them to show customers how they can deal with those handlers.

samples/matter/common/src/event_triggers/default_event_triggers.h
```diff
@@ -19,12 +19,20 @@ enum Ids : TestEventTrigger::EventTriggerId {
 	/* System */
 	FactoryReset = 0xF000'0000'0000'0000,
 	Reboot = 0xF000'0001'0000'0000,
+
+	/* Watchdog */
+	BlockMatterThread = 0xF000'0002'0000'0000,
+	BlockMainThread = 0xF000'0003'0000'0000,
 };
+// 925.5 kB
```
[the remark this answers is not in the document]
ArekBalysNordic: Because of that it is still in draft version :D

samples/matter/common/src/watchdog/watchdog.cpp
```diff
@@ -0,0 +1,82 @@
...
+		k_timer_init(&mTimer, &TimerTimeoutCallback, nullptr);
+		k_timer_user_data_set(&mTimer, this);
+		k_timer_start(&mTimer, K_MSEC(mConfig.feedingTime), K_MSEC(mConfig.feedingTime));
+	}
+
+	LOG_DBG("Registered a watchdog channel %d\n\tMin window: %u\n\tMax window: %u\n\tFeed interval %u", mChannel,
+		mConfig.minWindow, mConfig.maxWindow, mConfig.feedingTime);
+}
+
+Watchdog::~Watchdog() {}
```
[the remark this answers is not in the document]
ArekBalysNordic: Yes that's right, I haven't written the destructor yet :)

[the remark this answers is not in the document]
ArekBalysNordic: Apart form removing the channel, we need to destroy the watchdog internally. I'm going to do it :)

samples/matter/template/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
ArekBalysNordic: Yes, only for nRF54L15 it is required to do it manually for now

ArekBalysNordic: This is done in the same way in Zephyr's sample

samples/matter/common/src/watchdog/watchdog.h
ArekBalysNordic: Currently, because I'm still working on that :)

samples/matter/common/src/watchdog/watchdog.cpp
```diff
@@ -0,0 +1,82 @@
...
+using namespace Nrf;
+
+namespace
+{
+const struct device *const wdt = DEVICE_DT_GET(DT_ALIAS(watchdog0));
+}
+
+bool Watchdog::sEnabled = false;
+
+Watchdog::Watchdog(uint32_t minWindowTime, uint32_t maxWindowTime, uint32_t feedingInterval = 0,
```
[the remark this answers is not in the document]
ArekBalysNordic: Yes, the Watchdog module has only several channels to assign, so we can create Watchdog objects as far as a channel is available. I will add an assert there.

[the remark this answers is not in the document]
ArekBalysNordic: Ok, I checked that it is handled by the `wdt_install_timeout` function. If there are no more channels available the function returns `-ENOMEM` and it will be blocked by the assert below. I will add a simple log there for developers for this error code.

ArekBalysNordic:
Some of nrfx lib watchdog features are incompatible with Zephyr's API, I need to investigate whether it is a bug or real hardware limitation.

ArekBalysNordic:
This PR is ready for review, but it won't work properly until the bug in nrfxlib/zephyr is repaired. The bug related to removing watchdog channels on the `wdt_disable` function call. Zephyr's API assumes that all channels should be removed whereas nrfxlib implementation doesn't do it. It seems to be repaired within this PR: https://github.com/zephyrproject-rtos/zephyr/pull/70220.

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
```diff
@@ -53,17 +53,33 @@ The following table lists the available triggers and their activation codes:
...
+    - `0xF000000000000000` - `0xF0000000FFFFFFFF`
+    - The range of `0x00000000` - `0xFFFFFFFF` represents the delay in ms to wait until the factory reset occurs.
       The maximum time delay is UINT32_MAX ms.
       The value is provided in HEX format.
   * - Reboot
     - None
     - Reboot the device.
-    - `0xF000000000000000` - `0xF00000000000FFFF`
-    - The range of `0x0000` - `0xFFFF` represents the delay in ms to wait until the reboot occurs.
+    - `0xF000000100000000` - `0xF0000001FFFFFFFF`
```
Damian-Nordic, line 63: I think we're not allowed to use these IDs for non-standard test even triggers. Per spec:
> Values of EventTrigger in the range 0xFFFF_FFFF_0000_0000 through 0xFFFF_FFFF_FFFF_FFFF are
reserved for testing use by manufacturers and will not appear in CSA certification test literature.

samples/matter/common/src/event_triggers/default_event_triggers.cpp
```diff
@@ -20,19 +21,68 @@ using namespace chip;
...
+	Platform::UniquePtr<DelayedCtx> delayedContext(Platform::New<DelayedCtx>());
+	if (!delayedContext) {
+		return CHIP_ERROR_NO_MEMORY;
+	}
+
+	delayedContext->action = DelayedAction::FactoryReset;
+	delayedContext->value = delayMs;
+
+	k_timer_start(&sDelayTimer, K_MSEC(delayMs), K_NO_WAIT);
+	k_timer_user_data_set(&sDelayTimer, reinterpret_cast<void *>(delayedContext.get()));
```
Damian-Nordic, line 83: If two delayed actions are executed close to each other, this will result in a leak. If we really need heap allocated context for the timer, we must make sure it's expired before spawning a new timer.

samples/matter/common/src/watchdog/Kconfig
```diff
@@ -0,0 +1,44 @@
...
+    default 10000
+    help
+        The default maximum window time defines a time within which the feeding signal must be received
+        from all created Watchdog sources to reset the watchdog object's timer.
+        If the signals come after elapsing this time, the watchdog timer will not be reset and the
+        reboot occur.
+
+config NCS_SAMPLE_MATTER_WATCHDOG_DEFAULT_FEED_TIME
+    int "Default interval of feeding time in milliseconds"
+    default 1000
```
Damian-Nordic, line 32: This seems quite low from power power consumption perspective. Shouldn't we rather deduce the interval from the timeout?  If timeout is, say, a minute, then feeding the watchdog every 30 seconds should be enough?

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,169 @@
...
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#pragma once
+
+#include <cstdint>
+
+#include "util/finite_map.h"
```
Damian-Nordic, line 11: Unused

samples/matter/common/src/watchdog/watchdog.cpp
```diff
@@ -0,0 +1,215 @@
...
+	/* The callback has no sense, because watchdog interrupt takes only ~64 us and CPU
+	 * cannot do anything useful */
+	wdtConfig.callback = nullptr;
+
+	WatchdogListItem *item = chip::Platform::New<WatchdogListItem>();
+	assert(item);
+
+	mChannel = wdt_install_timeout(wdt, &wdtConfig);
+	if (-ENOMEM == mChannel) {
+		LOG_ERR("Cannot create a new Watchdog source because there is no more available Watchdog channels");
```
Damian-Nordic, line 56: Should we exit then? We shouldn't rely on `asserts` for handling error cases that may occur even in a correctly written software. Asserts can be disabled. Also, it's better to use Zephyr `__ASSERTs` than C ones to have a better control on their behavior (but only for code internal assumption violations, not "normal" things like heap allocation failures).

doc/nrf/protocols/matter/end_product/test_event_triggers.rst
ArekBalysNordic: Oh, right, I didn't notice that. So we need to change also the existing ones to meet the requirements.

samples/matter/common/src/watchdog/Kconfig
ArekBalysNordic: Yes, that's right. I checked that, and a 1-second interval causes ~1uA higher power consumption. If we change it to 30 seconds and timeout to 1 minute it probably will have an insignificant impact on power consumption.

samples/matter/common/src/event_triggers/default_event_triggers.cpp
ArekBalysNordic: Ok, I will think about protection or eliminating heap allocation.

samples/matter/common/src/watchdog/Kconfig
ArekBalysNordic: And deducing this value from the timeout seems good.

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Feeding the Matter Watchdog
+***************************
+
+Each Matter Watchdog source must be fed separately in a specific context.
+There are two options for feeding it:
+
+* Feeding automatically
+
+  You can utilize the automatic feeding feature of the Matter Watchdog to periodically feed the Matter Watchdog source within a defined time.
```
wiba-nordic, line 48:
```suggestion
Feeding automatically
  You can utilize the automatic feeding feature of the Matter Watchdog to periodically feed the Matter Watchdog source within a defined time.
```
Since there are so many steps between the top-level bullets, I think using the term definition formatting will look nicer.

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+
+  3. Enable the Global Watchdog module:
+
+     .. code-block:: c++
+
+        Nrf::Watchdog::Enable();
+
+* Feeding manually
+
+  Manual feeding involves calling the ``Feed()`` method of the specific source
```
wiba-nordic, line 88:
```suggestion
Feeding manually
  Manual feeding involves calling the ``Feed()`` method of the specific source.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+
+     .. code-block:: c++
+
+        myWatchdog.Feed();
+
+
+Enabling and disabling the Watchdog peripheral
+**********************************************
+
+The Global Watchdog used in Matter Samples is a single peripheral that operates independently of the CPU cores and includes multiple channels.
```
wiba-nordic, line 116:
```suggestion
The Global Watchdog used in the |NCS| Matter samples is a single peripheral that operates independently of the CPU cores and includes multiple channels.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+To enable Matter Watchdog sources again, you need to call the ``Restore()`` method on each of them.
+
+For example:
+
+    .. code-block:: c++
+
+        myWatchdog1.Restore();
+        myWatchdog2.Restore();
+
+End then enable the Global Watchdog module:
```
wiba-nordic, line 145:
```suggestion
After enabling the sources, enable the Global Watchdog module:
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+
+.. code-block:: c++
+
+    Nrf::Watchdog::Enable();
+
+Limited lifetime of the Watchdog source
+***************************************
+
+Watchdog source can be created locally and removed when the objects is deleting.
```
wiba-nordic, line 154:
```suggestion
A watchdog source can be created locally and removed when the objects is deleting.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+
+For example you can use it in the specific function:
```
wiba-nordic, line 156:
```suggestion
For example, you can use create a source in a specific function:
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+        {
+            while(condition)
+            {
+                /* Do some time-critical operations and break loop */
+                watchdogSource.Feed();
+            }
+        }
+    }
+
+In the function mentioned above, when the Watchdog source object is created, a new Watchdog is also created and it must be fed periodically.
```
wiba-nordic, line 173:
```suggestion
In the example function above, when the watchdog source object is created, a new watchdog is also created and it must be fed periodically.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+After performing some time-critical operations and exiting the loop in the ``DoTask`` function, the Watchdog source object is deleted, and its destructor frees the assigned channel.
```
wiba-nordic, line 174:
```suggestion
After performing some time-critical operations and exiting the loop in the ``DoTask`` function, the watchdog source object is deleted, and its destructor frees the assigned channel.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+        myWatchdog2.Restore();
+
+End then enable the Global Watchdog module:
+
+.. code-block:: c++
+
+    Nrf::Watchdog::Enable();
+
+Limited lifetime of the Watchdog source
+***************************************
```
wiba-nordic, line 152:
```suggestion
Limited lifetime watchdog source
********************************
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+            }
+        }
+    }
+
+In the function mentioned above, when the Watchdog source object is created, a new Watchdog is also created and it must be fed periodically.
+After performing some time-critical operations and exiting the loop in the ``DoTask`` function, the Watchdog source object is deleted, and its destructor frees the assigned channel.
+
+.. note::
+
+    The maximum number of simultaneously active Watchdog sources is 8. Therefore, creating a ninth Watchdog source will trigger an assert.
```
wiba-nordic, line 178:
```suggestion
    The maximum number of simultaneously active watchdog sources is 8.
    Creating a ninth watchdog source will trigger an assert.
```
Should this note also be in the creating section?

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Default Matter Watchdog implementation
+**************************************
+
+In the Matter common module, there is a default implementation of two Watchdog sources that are automatically created for the release build version of a Matter sample.
```
wiba-nordic, line 183:
```suggestion
In the Matter common module, there is a default implementation of two watchdog sources that are automatically created for the release build version of a Matter sample.
```

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -110,6 +110,8 @@ Matter
 
   * Support for merging the generated factory data HEX file with the firmware HEX file by using the devicetree configuration, when Partition Manager is not enabled in the project.
   * Support for the unified Persistent Storage API, including the implementation of the PSA Persistent Storage.
+  * Watchdog timer implementation for creating multiple Watchdog sources and monitoring the time of executing specific parts of the code.
+    To read more see the :ref:`Matter Watchdog <ug_matter_device_watchdog>` user guide.
```
wiba-nordic, line 114:
```suggestion
  * Watchdog timer implementation for creating multiple :ref:`ug_matter_device_watchdog` sources and monitoring the time of executing specific parts of the code.
```

wiba-nordic (changes requested):

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
+.. _ug_matter_device_watchdog:
+
+Matter Watchdog
+###############
+
+A watchdog is a type of hardware timer responsible for rebooting the device if it does not receive a specific signal within a designated time window.
+"Feeding" refers to the process of sending a signal to the watchdog module to reset the timer before it expires.
+In Matter samples, you can create multiple watchdog sources and assign them to specific functions.
```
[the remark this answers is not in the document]
ArekBalysNordic: I created an abstraction for Zephyr's watchdog and called "Watchdog source" as a name dedicated to the abstraction, so it applies to the NCS Matter samples and Applications(Bridge & Weather station). In Zephyr, it is called differently but actually, I used the Zephyr's API inside.

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,188 @@
...
+The Matter Watchdog is based on Zephyr's watchdog API and allows for the creation of multiple watchdog sources.
+Each Matter Watchdog source utilizes a unique Global Watchdog channel and must be fed individually.
+You can set a time window for the Watchdog timer and enable the automatic feeding feature.
+
+The feeding feature depends on establishing an additional timer that periodically executes the created feeding callback function.
+You can create your own feeding callback function and use it to call the ``Feed()`` method in a specific context.
+This approach eliminates the need to directly call the ``Feed()`` method in your code.
+
+A time window specifies the period within which the feeding signal must be sent to each Watchdog channel to reset the timer and prevent the device from rebooting.
+If the feeding signal is sent after the time window has elapsed, the device will reboot.
```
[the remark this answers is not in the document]
ArekBalysNordic: You're right, it sounds better.

doc/nrf/protocols/matter/end_product/watchdog.rst
ArekBalysNordic: Yes, it should be. Will add it there.

doc/nrf/protocols/matter/end_product/watchdog.rst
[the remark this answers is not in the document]
ArekBalysNordic: Ok, I will add something like that. :)

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,193 @@
...
+.. code-block:: c++
+
+    Nrf::Watchdog::Enable();
+
+Limited lifetime watchdog source
+********************************
+
+A watchdog source can be created locally and removed when the objects is deleting.
+
+For example, you can use create a source in a specific function:
```
wiba-nordic, line 160:
```suggestion
For example, you can create a source in a specific function:
```

wiba-nordic (changes requested):

samples/matter/common/src/event_triggers/default_event_triggers.cpp
ArekBalysNordic: I've added a protection there.

samples/matter/common/src/watchdog/watchdog.cpp
ArekBalysNordic: I've extended those assertions, but I'm still considering whether it is a good approach at all to make it all in a constructor. It works fine, and it is super simple for the user, but I'm wondering whether the errors will be noticed by developers.

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,167 @@
...
+		 *
+		 * This method must be called in order to re-enable a Watchdog source if the Enable() static method has
+		 * been called.
+		 *
+		 * @warning This method must be called before calling Enable() if the Disable function has been called.
+		 *
+		 * @return true if a Watchdog source has been restored and started working again.
+		 * @return false if the Global Watchdog is in the state that disallows restoring the Watchdog source.
+		 */
+		bool Restore();
```
Damian-Nordic, line 144: Should this be private? This is not expected to be called by a user, right? Also, I wonder if this could be called `Install()` and reused in the constructor to reduce the code duplication.

samples/matter/common/src/watchdog/watchdog.cpp
```diff
@@ -0,0 +1,224 @@
...
+		return false;
+	}
+
+	/* Stop all feeding timers*/
+	sys_snode_t *node;
+	sys_snode_t *tmpNodeSafe;
+	WatchdogListItem *item;
+	SYS_SLIST_FOR_EACH_NODE_SAFE (&sWatchdogList, node, tmpNodeSafe) {
+		item = reinterpret_cast<WatchdogListItem *>(node);
+		if (k_timer_status_get(&item->mWatchdog->mTimer)) {
```
Damian-Nordic, line 216: Why cancelling the timer only if it expired at least once?

samples/matter/common/src/event_triggers/default_event_triggers.cpp
```diff
@@ -20,19 +21,75 @@ using namespace chip;
...
+				LOG_INF("Blocking Main thread finished");
+			});
+			break;
+		default:
+			break;
+		}
+	});
+
+	/* Assign that the delayed action has been finished */
+	sDelayedActionInProgress = false;
```
Damian-Nordic, line 71: You still use the timer user data after setting this (as you call `k_timer_user_data_get` from the application thread), which means the user data may be overwritten before the previous is consumed.

samples/matter/common/src/watchdog/watchdog.h
ArekBalysNordic: It can be called Install for sure :) I thought about the approach that the user wants to stop only the several watchdogs. For example, somebody creates 4 Watchog sources and wants to disable 3 of them without removing the 4th one. So it is not typical behaviour and they need to call Disable(), then Restore()/Install() on the desired sources, and then Enable(). On the other hand, after rethinking, we can do the same by calling the destructor on the required source, and thanks to that making the Install() method private.

samples/matter/common/src/watchdog/watchdog.cpp
ArekBalysNordic: That's wrong, I need to stop the timer only if after the node removal there is no one on the list anymore.

samples/matter/common/src/watchdog/watchdog.cpp
```diff
@@ -0,0 +1,224 @@
...
+	/* The minimum window time must be 0 */
+	wdtConfig.window.min = 0;
+	/* The maximum window time must be equal for all channels */
+	wdtConfig.window.max = CONFIG_NCS_SAMPLE_MATTER_WATCHDOG_TIMEOUT;
+	/* The callback has no sense, because watchdog interrupt takes only ~64 us and CPU
+	 * cannot do anything useful */
+	wdtConfig.callback = nullptr;
+
+	WatchdogListItem *item = chip::Platform::New<WatchdogListItem>();
+	AssertAndReturn(item, "No memory to create a new WatchdogListItem");
```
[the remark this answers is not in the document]
ArekBalysNordic: Ok, I don't see a possibility to do it safely within the constructor, I will change it to the `Insert()` function.

samples/matter/common/src/watchdog/watchdog.h
Damian-Nordic: Ok, the rename is optional of course :). I just thought of that in the context of using it in the constructor, so installing the source for the first time.

wiba-nordic (approved):

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -38,10 +38,37 @@ Each Matter watchdog source constructor includes two optional arguments:
 * ``feedingInterval`` - Specifies the duration in milliseconds for automatically calling the attached feeding callback.
 * ``callback`` - A ``FeedingCallback`` implementation that is called periodically at the defined ``feedingInterval``.
 
+After creating a watchdog source you need to install it in the global watchdog module to assign a watchdog channel.
```
wiba-nordic, line 41:
```suggestion
After creating a watchdog source, you need to install it in the global watchdog module to assign a watchdog channel.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -38,10 +38,37 @@ Each Matter watchdog source constructor includes two optional arguments:
...
+To do it call the ``InstallSource(WatchdogSource &source)`` function providing the created source and check if the result of the function is ``true``.
```
wiba-nordic, line 42:
```suggestion
To do install it, call the ``InstallSource(WatchdogSource &source)`` function and provide the created source and check that the result of the function is ``true``.

File: doc/nrf/protocols/matter/end_product/watchdog.rst [hunk 34]
```diff
@@ -38,10 +38,37 @@ Each Matter watchdog source constructor includes two optional arguments:
...
+After that enable the global watchdog by calling ``Enable()`` function.
```
wiba-nordic, line 43:
```suggestion
After this, enable the global watchdog by calling the ``Enable()`` function.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -38,10 +38,37 @@ Each Matter watchdog source constructor includes two optional arguments:
...
+
+This is an example of creating two watchdog sources, one without and second one with the automatic feeding feature enabled:
```
wiba-nordic, line 45:
```suggestion
This following is an example of creating two watchdog sources, one without and second one with the automatic feeding feature enabled:

File: doc/nrf/protocols/matter/end_product/watchdog.rst [hunk 36]
```diff
@@ -51,6 +78,8 @@ There are two options for feeding it:
 
 Feeding automatically
   You can utilize the automatic feeding feature of the Matter watchdog to periodically feed the Matter watchdog source within a defined time.
+  This approach may cause a higher power consumption, especially when the feedback interval is set to the low value.
```
wiba-nordic, line 81:
```suggestion
  This approach can cause a higher power consumption, especially when the feedback interval is set to a low value.

File: doc/nrf/protocols/matter/end_product/watchdog.rst [hunk 37]
```diff
@@ -51,6 +78,8 @@ There are two options for feeding it:
...
+  If you want to keep low power consumption consider using the manually feeding approach.
```
wiba-nordic, line 82:
```suggestion
  If you want to keep low power consumption, consider using the manual feeding approach instead.

File: doc/nrf/protocols/matter/end_product/watchdog.rst [hunk 38]
```diff
@@ -69,27 +98,38 @@ Feeding automatically
...
+  3. Create a Matter watchdog source and provide the ``uint32_t feedingInterval``, and ``FeedingCallback callback`` optional arguments to the Matter watchdog source constructor.
 
      For example, here is a Matter watchdog source object configured to call the previously defined callback every 200 ms:
 
      .. code-block:: c++
 
         Nrf::Watchdog::WatchdogSource myWatchdog(200, FeedFromApp);
 
-  3. Enable the Global watchdog module:
+  4. Try to install the Matter watchdog source:
```
wiba-nordic, line 115: This step should have something after the sample function about what the user should do if the attempt fails.

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -69,27 +98,38 @@ Feeding automatically
...
+
+  5. Enable the Global watchdog module:
 
      .. code-block:: c++
 
         Nrf::Watchdog::Enable();
 
 Feeding manually
   Manual feeding involves calling the ``Feed()`` method of the specific source.
+  This approach provides the best power consumption, because feeding takes place only if the CPU is not in the sleep mode.
```
wiba-nordic, line 132:
```suggestion
  This approach provides the best power consumption, because feeding only takes place if the CPU is not in sleep mode.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -127,30 +176,17 @@ To enable the Global watchdog module, use the ``Nrf::Watchdog::Enable()`` functi
...
-
-        myWatchdog1.Restore();
-        myWatchdog2.Restore();
-
-After enabling the sources, enable the Global watchdog module:
-
-.. code-block:: c++
-
-    Nrf::Watchdog::Enable();
+To disable a specific Matter watchdog source delete the created object.
```
wiba-nordic, line 189:
```suggestion
To disable a specific Matter watchdog source, delete the created object.
```

wiba-nordic (changes requested):

samples/matter/common/src/watchdog/watchdog.cpp
```diff
@@ -0,0 +1,186 @@
...
+{
+	/* Stop the global timer */
+	Disable();
+
+	sys_snode_t *node, *tmpNodeSafe;
+	SYS_SLIST_FOR_EACH_NODE_SAFE (&sWatchdogList, node, tmpNodeSafe) {
+		WatchdogSource *source = static_cast<WatchdogSource *>(node);
+		if (source == this) {
+			/* Remove this object from the list and stop its timer if it was enabled */
+			sys_slist_find_and_remove(&sWatchdogList, node);
```
Damian-Nordic, line 174: I guess both `sys_slist_find_and_remove()` and `k_timer_stop()` can be called unconditionally without checking if `this` is in the list.

Damian-Nordic (approved):
some final nits found, but LGTM.

samples/matter/common/src/event_triggers/default_event_triggers.cpp
ArekBalysNordic: Is better now? :)

Damian-Nordic: yes, thanks :)

doc/nrf/protocols/matter/end_product/watchdog.rst
ArekBalysNordic: I will add a note for that at the bottom of this point to cover both appoaches.

samples/matter/common/src/event_triggers/default_event_triggers.cpp
```diff
@@ -20,19 +21,75 @@ using namespace chip;
...
+		case DelayedAction::FactoryReset:
+			Server::GetInstance().ScheduleFactoryReset();
+			break;
+		case DelayedAction::BlockMainThread:
+			Nrf::PostTask([value] {
+				int64_t startTime = k_uptime_get();
+
+				LOG_INF("Blocking Main thread for %d ms", value);
+
+				while (startTime + value > k_uptime_get()) {
```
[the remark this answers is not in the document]
ArekBalysNordic: Hmm, something doesn't work with `k_msleep` properly, but it is a good idea to use it. I need to investigate what's going on.

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,251 @@
...
+  4. Call the ``Feed()`` method at the specific place in the code where you want to prevent code blocking:
+
+     .. code-block:: c++
+
+        myWatchdog.Feed();
+
+
+.. note::
+
+    If the ``InstallSource(WatchdogSource &source)`` function returns ``false`` it may mean that there is no available watchdog channel to assign.
```
wiba-nordic, line 174:
```suggestion
    If the ``InstallSource(WatchdogSource &source)`` function returns ``false``, it can mean that there is no available watchdog channel to assign.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,251 @@
...
+    Ensure that you have at least one of the eight possible channels is available.
+    It may returns ``false`` also when the global watchdog module is not declared in the Devicetree specification file.
```
wiba-nordic, line 176:
```suggestion
    The function can also return ``false`` when the global watchdog module is not declared in the Devicetree specification file.
    Ensure that the module is properly declared.
```

doc/nrf/protocols/matter/end_product/watchdog.rst
```diff
@@ -0,0 +1,251 @@
...
+
+     .. code-block:: c++
+
+        myWatchdog.Feed();
+
+
+.. note::
+
+    If the ``InstallSource(WatchdogSource &source)`` function returns ``false`` it may mean that there is no available watchdog channel to assign.
+    Ensure that you have at least one of the eight possible channels is available.
```
wiba-nordic, line 175:
```suggestion
    Ensure that you have at least one of the eight possible channels available.
```

wiba-nordic (changes requested):

samples/matter/common/src/event_triggers/default_event_triggers.cpp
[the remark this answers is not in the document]
ArekBalysNordic: Ok, I see a problem, I used the `WDT_OPT_PAUSE_IN_SLEEP` flag to pause the watchdog timer when the device is in sleep mode, but the watchdog is also paused when we use k_sleep. :) So, it seems that we need to block it brutally using a while loop. Moreover, I think it simulates the real blocking issue better than k_sleep.

wiba-nordic (approved):

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+ *
+ * Watchdog sources cannot be disabled separately after enabling the Global Watchdog, but you can disable all of
+ * them at once using the Disable() static method. After that to re-enable Watchdog sources you need to call
+ * Restore() method on each of them and call Enable() method to start the Global Timer once again.
+ *
+ * Similarly to creating the objects, you must call Restore() method before calling the Enable() method.
+ */
+namespace Watchdog
+{
+	constexpr static size_t kMaximumAvailableSources = 8;
```
markaj-nordic, line 35: nit: Is this something we should make configurable with Kconfig? Is this limitation coming from the watchdog peripheral? Maybe we can add a comment to explain it.

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+
+#include <zephyr/drivers/watchdog.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/slist.h>
+
+namespace Nrf
+{
+
+/**
+ * @brief Watchdog peripheral that can be use to detect code blocking and infinite loops
```
markaj-nordic, line 19: nit:
```suggestion
 * @brief Watchdog implementation that can be use to detect code blocking and infinite loops
```
or
```suggestion
 * @brief Watchdog driver wrapper that can be use to detect code blocking and infinite loops
```

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+ *
+ * It contains a possibility to create a different Watchdog sources that must be fed within the
```
markaj-nordic, line 21:
```suggestion
 * It contains a possibility to create a different Watchdog sources that must be fed within the defined timeout.
```
?

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+	constexpr static size_t kMaximumAvailableSources = 8;
+
+	/**
+	 * @brief WatchdogSource class dedicated to detect code blocking and infinite loops.
+	 *
+	 * While the new Watchdog source is creating a new channel for the global Watchdog is created.
+	 * Each watchdog source contains works separately and must be fed in a application-specific way.
+	 *
+	 * If at least one of the created watchdog sources will not be fed in the defined time window then the reboot
+	 * occurs.
```
markaj-nordic, line 44:
```suggestion
	 * If at least one of the created watchdog sources is not fed in the defined time window then the reboot
	 * will occur.
```

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+ */
+namespace Watchdog
+{
+	constexpr static size_t kMaximumAvailableSources = 8;
+
+	/**
+	 * @brief WatchdogSource class dedicated to detect code blocking and infinite loops.
+	 *
+	 * While the new Watchdog source is creating a new channel for the global Watchdog is created.
+	 * Each watchdog source contains works separately and must be fed in a application-specific way.
```
markaj-nordic, line 41:
```suggestion
	 * Each watchdog source handles works separately and must be fed in a application-specific way.
```
?

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+ * Similarly to creating the objects, you must call Restore() method before calling the Enable() method.
+ */
+namespace Watchdog
+{
+	constexpr static size_t kMaximumAvailableSources = 8;
+
+	/**
+	 * @brief WatchdogSource class dedicated to detect code blocking and infinite loops.
+	 *
+	 * While the new Watchdog source is creating a new channel for the global Watchdog is created.
```
markaj-nordic, line 40: I suggest to rephrase this sentence, as it's quite unclear :)

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+	 *    argument. If the callback is provided it will be called in each feed interval time periodically.
+	 */
+	class WatchdogSource : public sys_snode_t {
+	public:
+		/**
+		 * @brief Callback for feeding purposes
+		 *
+		 * This callback will be run when feedingInterval timeout occurs.
+		 * It must call the Feed method within the specific context.
+		 * If this callback cannot be call within the CONFIG_NCS_SAMPLE_MATTER_WATCHDOG_TIMEOUT timeout the
```
markaj-nordic, line 63:
```suggestion
		 * If this callback cannot be called within the CONFIG_NCS_SAMPLE_MATTER_WATCHDOG_TIMEOUT timeout the
```

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+		 * reboot occurs.
+		 *
+		 * For example to feed an Watchdog source dedicated for Main thread
```
markaj-nordic, line 66:
```suggestion
		 * For example to feed a Watchdog source dedicated for Main thread
```

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+
+		/**
+		 * @brief Create new Watchdog feeding source
+		 *
+		 * Each Watchdog source may contain a callback which is called when the feedingInterval time is elapsed.
+		 * If at least one of the watchdog source is not fed within the
+		 * CONFIG_NCS_SAMPLE_MATTER_WATCHDOG_TIMEOUT then the reboot occurs.
+		 *
+		 * @param FeedingInterval if the callback argument is provided then after elapsing time in milliseconds
+		 * of this value the callback will be set.
```
markaj-nordic, line 81:
```suggestion
		 * @param feedingInterval if the callback argument is provided this value is used as an interval [ms]
		 * between subsequent callback occurrences.
```

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+		 */
+		WatchdogSource(uint32_t feedingInterval = 0, FeedingCallback callback = nullptr)
+			: mFeedingTime(feedingInterval), mCallback(callback)
+		{
+		}
+
+		/**
+		 * @brief Remove the Watchdog source
+		 *
+		 * The destructor perform a several steps to remove a WatchdogSource:
```
markaj-nordic, line 92:
```suggestion
		 * The destructor performs a several steps to remove a WatchdogSource:
```

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+		 * 1. Disables the Watchdog peripheral.
+		 * 2. Removes this Watchdog source.
```
markaj-nordic, line 94:
```suggestion
		 * 2. Removes the Watchdog source.
```

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+		/* Pause Watchdog when the CPU is in sleep state or is halted by the debugger */
+		constexpr static uint8_t kWatchdogOptions = WDT_OPT_PAUSE_HALTED_BY_DBG | WDT_OPT_PAUSE_IN_SLEEP;
+
+		/**
+		 * @brief Restore the Watchdog source
+		 *
+		 * This method must be called in order to re-enable a Watchdog source if the Enable() static method has
+		 * been called.
+		 *
+		 * @warning This method must be called before calling Enable() if the Disable function has been called.
```
markaj-nordic, line 127: I'd expect that if this method is private, then it is called in by the class internally as a part of implementation (not API).

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+		uint32_t mFeedingTime;
+		k_timer mTimer;
+		FeedingCallback mCallback = nullptr;
+	};
+
+	/**
+	 * @brief Enable the global Watchdog.
+	 *
+	 * @warning This method must be called after creating all Watchdog sources.
+	 * 			Calling it before initializing next Watchdog source will cause an assert.
```
markaj-nordic, line 151:
```suggestion
	 * 			Calling it before initializing the next Watchdog source will cause an assert.
```

samples/matter/common/src/watchdog/watchdog.h
```diff
@@ -0,0 +1,187 @@
...
+	/**
+	 * @brief Install a new watchdog source
+	 *
+	 * A new source can be installed any time and it exists as long as the origin object exists.
+	 * So to uninstall the source, call its destructor.
+	 *
+	 * @param source Watchdog Source object
+	 * @return true If source has been installed
+	 * @return false If source cannot be installed due to installation error, or there is no available watchdog
+	 * channels.
```
markaj-nordic, line 182:
```suggestion
	 * channel.
```

markaj-nordic (approved):
The implementation LGTM, I only added some suggestions to refine the API docs, as it might be important for the user.

samples/matter/common/src/watchdog/watchdog.h
ArekBalysNordic: This is a limitation that coming from the watchdog so I will add a note here.

samples/matter/common/src/watchdog/watchdog.h
ArekBalysNordic: I see that it can be public, since the user would like to disable only several sources, but to do this they need to disable all of them and then call Install only on the specific objects.

PR #15225, opened 2024-05-08
samples: cellular: modem_shell: Add Thingy:91 X Wi-Fi scanning support

tokangas:
Added devicetree overlay for building the sample for Thingy:91 X with Wi-Fi scanning support. Renamed the nRF700x Kconfig overlay, because the same overlay can be used with nRF7002 EK and Thingy:91 X.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -503,7 +503,11 @@ Cellular samples
 
 * :ref:`modem_shell_application` sample:
 
-  * Added support for sending location data details into nRF Cloud with ``--cloud_details`` command-line option in the ``location`` command.
+  * Added:
+
+    * Support for sending location data details into nRF Cloud with ``--cloud_details`` command-line option in the ``location`` command.
+    * Support for Thingy:91 X Wi-Fi scanning.
+
   * Removed ESP8266 Wi-Fi DTC and Kconfig overlay files.
 
 * :ref:`nrf_cloud_rest_cell_pos_sample` sample:
--- a/samples/cellular/modem_shell/README.rst
+++ b/samples/cellular/modem_shell/README.rst
@@ -1003,8 +1003,8 @@ To program the certificates and connect to nRF Cloud, complete the following ste
 .. parsed-literal::
    :class: highlight
 
-   west build -p -b *board_target* -d build
-   west flash -d build
+   west build -p -b *board_target*
+   west flash
 
 |board_target|
 
@@ -1066,15 +1066,27 @@ To program the certificates and connect to nRF Cloud, complete the following ste
 nRF91 Series DK with nRF7002 EK Wi-Fi support
 =============================================
 
-To build the MoSh sample for an nRF91 Series DK with nRF7002 EK Wi-Fi support, use the ``-DSHIELD=nrf7002ek`` and  ``-DEXTRA_CONF_FILE=overlay-nrf7002ek-wifi-scan-only.conf`` options.
+To build the MoSh sample for an nRF91 Series DK with nRF7002 EK Wi-Fi support, use the ``-DSHIELD=nrf7002ek``, ``-DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf``, ``-DSB_CONFIG_WIFI_NRF700X=y`` and ``-DSB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y`` options.
 For example:
 
 .. parsed-literal::
    :class: highlight
 
-   west build -p -b *board_target* -- -DSHIELD=nrf7002ek -DEXTRA_CONF_FILE=overlay-nrf7002ek-wifi-scan-only.conf
+   west build -p -b *board_target* -- -DSHIELD=nrf7002ek -DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf -DSB_CONFIG_WIFI_NRF700X=y -DSB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y
 
 |board_target|
+
+See :ref:`cmake_options` for more instructions on how to add these options.
+
+Thingy:91 X Wi-Fi support
+=========================
+
+To build the MoSh sample with Thingy:91 X Wi-Fi support, use the ``-DDTC_OVERLAY_FILE=thingy91x_wifi.overlay``, ``-DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf``, ``-DSB_CONFIG_WIFI_NRF700X=y``, and ``-DSB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y`` options.
+For example:
+
+.. code-block:: console
+
+   west build -p -b thingy91x/nrf9151/ns -- -DDTC_OVERLAY_FILE=thingy91x_wifi.overlay -DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf -DSB_CONFIG_WIFI_NRF700X=y -DSB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y
 
 See :ref:`cmake_options` for more instructions on how to add these options.
 
@@ -1148,13 +1160,13 @@ After programming the development kit, test it in the Linux environment by perfo
 Application FOTA support
 ========================
 
-To build the MoSh sample with application FOTA support, use the ``-DEXTRA_CONF_FILE=overlay-app_fota.conf`` option.
+To build the MoSh sample with application FOTA support, use the ``-DEXTRA_CONF_FILE=overlay-app_fota.conf`` and ``-DSB_CONFIG_BOOTLOADER_MCUBOOT=y`` options.
 For example:
 
 .. parsed-literal::
    :class: highlight
 
-   west build -p -b *board_target* -d build -- -DEXTRA_CONF_FILE=overlay-app_fota.conf
+   west build -p -b *board_target* -- -DEXTRA_CONF_FILE=overlay-app_fota.conf -DSB_CONFIG_BOOTLOADER_MCUBOOT=y
 
 |board_target|
 
@@ -1166,7 +1178,7 @@ The following is an example for the nRF9161 DK:
 
 .. code-block:: console
 
-   west build -p -b nrf9161dk/nrf9161/ns -d build -- -DEXTRA_CONF_FILE=overlay-modem_fota_full.conf -DDTC_OVERLAY_FILE=nrf9161dk_ext_flash.overlay
+   west build -p -b nrf9161dk/nrf9161/ns -- -DEXTRA_CONF_FILE=overlay-modem_fota_full.conf -DDTC_OVERLAY_FILE=nrf9161dk_ext_flash.overlay
 
 LwM2M carrier library support
 =============================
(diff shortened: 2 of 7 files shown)


samples/cellular/modem_shell/sample.yaml
```diff
@@ -290,6 +290,13 @@ tests:
       - thingy91/nrf9160/ns
       - thingy91x/nrf9151/ns
     tags: ci_build
+  sample.cellular.modem_shell.thingy91x_wifi:
+    build_only: true
+    platform_allow:
+      - thingy91x/nrf9151/ns
```
trantanen, line 296: Should we add thingy91x compilation for some other basic configurations as well? I mean mainly the default configuration and maybe some other config that might be tight from memory usage perspective.

samples/cellular/modem_shell/thingy91x_wifi.overlay
```diff
@@ -0,0 +1,6 @@
+/* Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <thingy91x_wifi.dtsi>
```
trantanen, line 6: Hmmm, could we actually have this always included (by being under boards folder) even if WiFi would be disabled by default?
I guess it shouldn't cause any power consumption penalty etc. if WiFi is not turned on by CONFIG_WIFI etc. Kconfig symbols?

samples/cellular/modem_shell/sample.yaml
tokangas: Yes, I'll add `thingy91x` to `sample.cellular.modem_shell.thingy91` and `sample.cellular.modem_shell.thingy91_modem_uart_trace` configurations.

As discussed offline, we could have a look if the `integration_platforms` lists are needed at all. Usually, both `integration_platforms` and `platforms_allow` lists contain the same platforms. I'm not sure `integration_platforms` is needed in this case, as the configuration should be built by default for all allowed platforms.

samples/cellular/modem_shell/thingy91x_wifi.overlay
tokangas: I tried building the sample with and without the Wi-Fi include. There's a small (less than 200 bytes) difference in flash consumption. The increased flash consumption is not a problem as such, but I'm worried if the DTSI may affect current consumption even without the driver being enabled.

tokangas: We decided to keep it like it's now, just in case.

peknis (dismissed):

trantanen (approved):

moksanen (approved):
lgtm

jhirsi (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -426,6 +426,7 @@ Cellular samples
 
   * Added support for sending location data details into nRF Cloud with ``--cloud_details`` command-line option in the ``location`` command.
   * Removed ESP8266 Wi-Fi DTC and Kconfig overlay files.
+  * Added support for Thingy:91 X Wi-Fi scanning.
```
peknis, line 429:
```suggestion
  * Added:

    * Support for sending location data details into nRF Cloud with ``--cloud_details`` command-line option in the ``location`` command.
    * Support for Thingy:91 X Wi-Fi scanning.

  * Removed ESP8266 Wi-Fi DTC and Kconfig overlay files.
```

peknis (changes requested):

peknis (approved):

tokangas:
@nordicjm This is PR is ready to be merged, but it's not clear to me what kind of PRs can be merged and what not.

nordicjm (dismissed):
This does not work with sysbuild, so cannot be accepted, unfortunately sysbuild does not support snippets and wifi configuration of which firmware is used is done at a sysbuild level, not an application level

trantanen (approved):

jhirsi (approved):

tokangas:
Pending PR #13153.

tokangas:
Rebased on sysbuild changes.

trantanen (approved):

samples/cellular/modem_shell/README.rst
```diff
@@ -1066,18 +1066,30 @@ To program the certificates and connect to nRF Cloud, complete the following ste
...
 
+Thingy:91 X Wi-Fi support
+=========================
+
+To build the MoSh sample with Thingy:91 X Wi-Fi support, use the ``-DDTC_OVERLAY_FILE=thingy91x_wifi.overlay`` and  ``-DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf`` options.
+For example:
+
+.. code-block:: console
+
+   west build -p -b thingy91x/nrf9151/ns -- -DDTC_OVERLAY_FILE=thingy91x_wifi.overlay -DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf
```
nordicjm, line 1089: this isn't even correct, look at the sample.yaml file

nordicjm (changes requested):
do not dismiss my nack, especially when it isn't even fixed

tokangas:
> do not dismiss my nack, especially when it isn't even fixed

@nordicjm I tested the build command in the example with your PR before it was merged. It worked (and still works) without sysbuild and at that point sysbuild was not enabled by default. Now it seems that the default has been changed to sysbuild and the build command does not work anymore.

Communication about the whole sysbuild change has been bad. It has not been communicated what developers actually need to do. Requesting changes with comment "This does not work with sysbuild" was not very helpful either.

Because of this change, all build command examples in all samples with nRF700x are currently incorrect and need to be updated. This is probably not the only change needed. Is there going to be a treewide update of these or are individual developers just supposed to figure this out themselves?

samples/cellular/modem_shell/sample.yaml
```diff
@@ -157,7 +157,6 @@ tests:
     build_only: true
     extra_args: OVERLAY_CONFIG=overlay-modem_fota_full.conf
                 DTC_OVERLAY_FILE="nrf9160dk_ext_flash.overlay"
-                SB_CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=y
```
tokangas, line 160: @nordicjm This option was added to three `modem_shell` configurations in #13153. I don't think these are needed and they do seem to cause a Kconfig warning. Please check.
```
warning: PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY (defined at
/home/tok1/work/ncs/nrf/sysbuild/Kconfig.sysbuild:46) was assigned the value 'y' but got the value
'n'. Check these unsatisfied dependencies: BOOTLOADER_MCUBOOT (=n). See
http://docs.zephyrproject.org/latest/kconfig.html#CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY and/or
look up PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY in the menuconfig/guiconfig interface. The Application
Development Primer, Setting Configuration Values, and Kconfig - Tips and Best Practices sections of
the manual might be helpful too.
```

tokangas:
@nordicjm `modem_shell` documentation has now been updated.

BTW, I tried to re-request a review from you in the first place, but there was no "Re-request review" button visible in the PR. 
I guess this might have been because you had not submitted a single comment in the code when you finished the review.

samples/cellular/modem_shell/README.rst
```diff
@@ -1066,18 +1066,30 @@ To program the certificates and connect to nRF Cloud, complete the following ste
...
+   west build -p -b *build_target* -- -DSHIELD=nrf7002ek -DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf -DSB_CONFIG_WIFI_NRF700X=y -DSB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y
 
 |build_target|
 
 See :ref:`cmake_options` for more instructions on how to add these options.
 
+Thingy:91 X Wi-Fi support
+=========================
+
+To build the MoSh sample with Thingy:91 X Wi-Fi support, use the ``-DDTC_OVERLAY_FILE=thingy91x_wifi.overlay``, ``-DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf``, ``-DSB_CONFIG_WIFI_NRF700X=y`` and ``-DSB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y`` options.
```
peknis, line 1084:
```suggestion
To build the MoSh sample with Thingy:91 X Wi-Fi support, use the ``-DDTC_OVERLAY_FILE=thingy91x_wifi.overlay``, ``-DEXTRA_CONF_FILE=overlay-nrf700x-wifi-scan-only.conf``, ``-DSB_CONFIG_WIFI_NRF700X=y``, and ``-DSB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y`` options.
```

peknis (approved):
Approving with a minor nit.

trantanen (approved):

samples/cellular/modem_shell/README.rst
```diff
@@ -1148,13 +1160,13 @@ After programming the development kit, test it in the Linux environment by perfo
 Application FOTA support
 ========================
 
-To build the MoSh sample with application FOTA support, use the ``-DEXTRA_CONF_FILE=overlay-app_fota.conf`` option.
+To build the MoSh sample with application FOTA support, use the ``-DEXTRA_CONF_FILE=overlay-app_fota.conf`` and ``-DSB_CONFIG_BOOTLOADER_MCUBOOT=y`` options.
```
tejlmand, line 1163: although the docs seems correct, then I wonder, why haven't this sample been adjust to have the following files instead:
- `prj_fota.conf` (merged content of `prj.conf` + `overlay-app_fota.conf`)
- `sysbuild_fota.conf`

then a `sysbuild_fota.conf` containing `SB_CONFIG_BOOTLOADER_MCUBOOT=y` can be made, and users could simply build the sample with fota support by doing:
```
$ west build -p -b *board_target* -- -DFILE_SUFFIX=fota
```

and similar goes for the other updated and complex examples in this PR.

tejlmand (approved):
lgtm.

An observation, although not directly related to the exact content of this PR.

samples/cellular/modem_shell/README.rst
trantanen: > although the docs seems correct, then I wonder, why haven't this sample been adjust to have the following files instead:

Because we haven't known about `FILE_SUFFIX`. Seems it's not too old thing in Zephyr?
There are probably dozen(s) samples which could utilize this more so some info sharing would be nice on best practices in setting sample/app overlays (both Kconfig and DTC).

> * `prj_fota.conf` (merged content of `prj.conf` + `overlay-app_fota.conf`)

Copying a larger `prj.conf` with 50+ Kconfig options into all overlay config files doesn't sound too good. I guess there are no other possibilities to utilize `FILE_SUFFIX` than copying `prj.conf` into the desired `prj_<FILE_SUFFIX>.conf`?

tokangas: Yes, otherwise that looks nice, but considering the size of `modem_shell` `prj.conf`, we probably don't want to duplicate it. We would lose the benefits of an overlay file, i.e. only having to include those options which are configured differently.

tokangas: We decided that it's not feasible to use this mechanism for `prj.conf` and using it only for sysbuild options would probably not make the configuration clearer.

nordicjm (approved):

PR #15240, opened 2024-05-09
suit: use SUIT boot by default for nrf54h20

robertstypa:
Use SUIT to boot Lilium samples

Ref: NCSDK-27372

Signed-off-by: Robert Stypa <robert.stypa@nordicsemi.no>


lemrey:
Please remember to update the commit title to something like `suit: ` before merging :)

tomchy (approved):

cmake/sysbuild/suit.cmake
```diff
@@ -165,6 +165,11 @@ function(suit_create_package)
 
   suit_set_absolute_or_relative_path(${SB_CONFIG_SUIT_ENVELOPE_EDITABLE_TEMPLATES_LOCATION} ${PROJECT_BINARY_DIR} INPUT_TEMPLATES_DIRECTORY)
   set(ENVELOPE_SHALL_BE_SIGNED ${SB_CONFIG_SUIT_ENVELOPE_SIGN})
+  if(NOT DEFINED west_realpath)
+    # Twister build - ignore and store in the build folder!
+    set(SB_CONFIG_SUIT_ENVELOPE_EDITABLE_TEMPLATES_LOCATION "./")
+  endif()
+  sysbuild_get(ENVELOPE_SHALL_BE_SIGNED IMAGE ${DEFAULT_IMAGE} VAR CONFIG_SUIT_ENVELOPE_SIGN KCONFIG)
```
e-rk, line 172: Not for this PR, but how about copying the templates into `build/<sample-name>/...` directory? The flow would be as follows:
1. If template exists at `<sample-dir>`, then copy it to `build/<sample-name>/...` and use it to generate the envelope
2. If template does not exist at `<sample-dir>`, then copy it from default location into `<sample-dir>` and then do action from 1.

e-rk (approved):

cmake/sysbuild/suit.cmake
robertstypa: I'll create a JIRA ticket to discuss that feature request

robertstypa: https://nordicsemi.atlassian.net/browse/NCSDK-27662

sysbuild/Kconfig.suit
```diff
@@ -6,9 +6,9 @@
 
 menu "SUIT"
 
-# 'default y if SOC_SERIES_NRF54HX' to be added when SUIT is to be enabled globally.
 menuconfig SUIT_ENVELOPE
 	bool "Create SUIT envelope"
+	default y if SOC_SERIES_NRF54HX
```
ahasztag, line 11: @e-rk will this influence flash_companion? Shouldn't the flash_companion app set `SUIT_ENVELOPE` to `n`?

e-rk: No, this is the sysbuild Kconfig that must be enabled always with or without the flash companion.

ahasztag (approved):

PR #15243, opened 2024-05-09
FMDN upstream part 2: Minimal Account Key storage

kapi-no:
Original commits adapted to address the following items:
- alignment with HWMv2
- alignment with testing target strategy
- alignment with NCS coding style requirements (do not init static variables to zero)
- alignment of test case names in the `ZTEST` macro
  - `self_test_basic` -> `test_self_basic`
  - `self_test_rollover` -> `test_self_rollover`
- migrated storage tests to sysbuild


zycz (approved):

tests/subsys/bluetooth/fast_pair/storage/account_key_storage/src/test_corrupted_data.c
```diff
@@ -212,42 +213,69 @@ ZTEST(suite_fast_pair_storage_corrupted, test_inconsistent_key_id)
...
 	key_id = next_account_key_id(key_id);
-	store_key(key_id);
+	store_key(key_id, false);
 
 	initialization_error_validate();
 }
 
 ZTEST(suite_fast_pair_storage_corrupted, test_drop_keys)
 {
+	if (ACCOUNT_KEY_CNT - (IS_ENABLED(CONFIG_TEST_BT_FAST_PAIR_STORAGE_OWNER_ACCOUNT_KEY) ?
```
MarekPieta, line 248: Shouldn't we always skip this test if owner AK is enabled? (or try to perform rollover with the first non-owner AK then?

tests/subsys/bluetooth/fast_pair/storage/account_key_storage/src/test_corrupted_data.c
```diff
@@ -28,7 +28,7 @@ static uint8_t next_account_key_id(uint8_t account_key_id)
 	return account_key_id + 1;
```
MarekPieta, line 28: The function should not be used for backends with Owner Account Key and multiple Account Keys support (and it still would be called for these backends)

MarekPieta: We should fix it or add a build assert `(ACCOUNT_KEY_CNT == 1) || !IS_ENABLED(CONFIG_TEST_BT_FAST_PAIR_STORAGE_OWNER_ACCOUNT_KEY)` (it should be ok too since all of our backends pass this check)

tests/subsys/bluetooth/fast_pair/storage/account_key_storage/src/test_corrupted_data.c
```diff
@@ -79,24 +83,30 @@ static void self_test(void)
...
+			store_key(key_id, true);
+		} else {
+			store_key(key_id, false);
+		}
 		key_cnt++;
 		key_id = next_account_key_id(key_id);
 	}
 
-	store_key(key_id);
+	store_key(key_id, false);
```
MarekPieta, line 107: This would override Owner AK if Account Key count >= 2 in the configuration (not expected)

MarekPieta (changes requested):

subsys/bluetooth/services/fast_pair/fp_storage/Kconfig.fp_storage
```diff
@@ -29,21 +29,44 @@ config BT_FAST_PAIR_STORAGE_USER_RESET_ACTION
 	  bt_fast_pair_factory_reset_user_action_prepare symbol if you want an earlier notification
 	  that the reset operation is due to begin.
 
-config BT_FAST_PAIR_STORAGE_AK
+config BT_FAST_PAIR_STORAGE_OWNER_ACCOUNT_KEY
 	bool
-	default y
 	help
-	  Add Fast Pair storage of Account Keys source files.
+	  Enable support for the Owner Account Key in the Storage module.
```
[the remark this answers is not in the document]
MarekPieta: Discussed offline with @kapi-no. We don't want to make prompts, because they would be available for the end user too. Resolved the comment.

tests/subsys/bluetooth/fast_pair/storage/account_key_storage/src/test_corrupted_data.c
zycz: I agree with assert approach for now we can make follow up ticket to fix it later.

tests/subsys/bluetooth/fast_pair/storage/account_key_storage/src/test_corrupted_data.c
```diff
@@ -212,42 +218,69 @@ ZTEST(suite_fast_pair_storage_corrupted, test_inconsistent_key_id)
...
+	} else {
+		store_key(next_account_key_id(ACCOUNT_KEY_MIN_ID), false);
+	}
 
 	initialization_error_validate();
 }
 
 ZTEST(suite_fast_pair_storage_corrupted, test_drop_key)
 {
+	if ((ACCOUNT_KEY_CNT < 2) &&
```
MarekPieta, line 232: Since you added the build assertion, maybe we could simplify this condition to `if (ACCOUNT_KEY_CNT < 2)` ? (also for `test_drop_keys`)

tests/subsys/bluetooth/fast_pair/storage/account_key_storage/src/test_corrupted_data.c
```diff
@@ -79,24 +88,30 @@ static void self_test(void)
 	cu_account_keys_validate_loaded(ACCOUNT_KEY_MIN_ID, key_cnt);
 }
 
-static void self_test_rollover(void)
+ZTEST(suite_settings_mock_self_test, test_self_rollover)
 {
+	if ((ACCOUNT_KEY_CNT < 2) &&
+	    IS_ENABLED(CONFIG_BT_FAST_PAIR_STORAGE_OWNER_ACCOUNT_KEY)) {
```
MarekPieta, line 94: `if(IS_ENABLED(CONFIG_BT_FAST_PAIR_STORAGE_OWNER_ACCOUNT_KEY)` ?
The test still does not properly cover case where owner AK and multiple AKs are supported in configuration (because of `next_account_key_id` implementation, we would overwrite owner AK here)

MarekPieta (changes requested):

MarekPieta (approved):

kapi-no:
Added a commit with the sysbuild migration for the Fast Pair Storage test suite + slightly renamed the Account Key storage test for minimal backend to follow the naming convention of other tests in the Account KEy storage test suite

PR #15246, opened 2024-05-09
modules: wfa-qt: use net_if_ipv4_set_netmask_by_addr

D-Triveni:
Replace usage of the deprecated `net_if_ipv4_set_netmask` with the `net_if_ipv4_set_netmask_by_addr`.

### DIFF
--- a/modules/wfa-qt/src/utils.c
+++ b/modules/wfa-qt/src/utils.c
@@ -697,7 +697,7 @@ int send_broadcast_arp(char *target_ip, int *send_count, int rate)
 
 void set_netmask(char *ifname)
 {
-	const struct device *dev = device_get_binding(ifname);
+	const struct device *dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_wifi));
 	struct net_if *iface = net_if_lookup_by_dev(dev);
 	struct in_addr addr;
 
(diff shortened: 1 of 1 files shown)


modules/wfa-qt/src/utils.c
```diff
@@ -697,18 +697,28 @@ int send_broadcast_arp(char *target_ip, int *send_count, int rate)
 
 void set_netmask(char *ifname)
 {
-	const struct device *dev = device_get_binding(ifname);
+	const struct device *dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_wifi));
```
krish2718, line 700: Please do this in a separate commit.

modules/wfa-qt/src/utils.c
```diff
@@ -697,18 +697,28 @@ int send_broadcast_arp(char *target_ip, int *send_count, int rate)
...
 	struct net_if *iface = net_if_lookup_by_dev(dev);
-	struct in_addr addr;
+	struct in_addr addr = { 0 };
```
krish2718, line 702: we don't need the unnecessary initialisation as we overwrite with the parse.

krish2718 (changes requested):

krish2718 (approved):

VivekUppunda (approved):

PR #15257, opened 2024-05-10
manifest: Update zephyr with nrf54h20dk overlays for driver tests

nordic-krch:
Update zephyr with overlays for nrf54h20dk for gpio, counter and uart tests.


nordic-segl (approved):

tests/drivers/nrfx_integration_test/src/main.c
```diff
@@ -15,6 +15,22 @@
...
+#define NRFX_GPIOTE_CHANNELS_USED NRFX_GPIOTE20_CHANNELS_USED
+#elif defined(NRFX_GPIOTE30_CHANNELS_USED)
+#define NRFX_GPIOTE_CHANNELS_USED NRFX_GPIOTE30_CHANNELS_USED
+#elif defined(NRFX_GPIOTE130_CHANNELS_USED)
+#define NRFX_GPIOTE_CHANNELS_USED NRFX_GPIOTE130_CHANNELS_USED
+#elif defined(NRFX_GPIOTE131_CHANNELS_USED)
+#define NRFX_GPIOTE_CHANNELS_USED NRFX_GPIOTE131_CHANNELS_USED
+#else
+#define NRFX_GPIOTE_CHANNELS_USED 0
+#endif
```
anangl, line 32: Instead of adding this stuff, I think we should just remove the following line (`NRFX_GPIOTE_CHANNELS_USED` no longer needs to be defined): https://github.com/nrfconnect/sdk-nrf/blob/35ebad3654c76569b68a6134508d88e5461739f1/tests/drivers/nrfx_integration_test/src/main.c#L71

nordic-krch: done

PR #15269, opened 2024-05-13
scripts: add integration quarantine part 13

nordic-piks:
To limit integration scope.
For:
nrf/tests/subsys/zigbee


nordic-segl (approved):

scripts/quarantine_integration.yaml
```diff
@@ -1183,3 +1183,111 @@
...
+    - sample.zigbee.light_switch.with_shell
+  platforms:
+    - nrf21540dk/nrf52840
+  comment: "Configurations excluded to limit resources usage in integration builds"
+
+- scenarios:
+    - sample.zigbee.light_switch.with_shell
+    - sample.zigbee.light_switch
+    - sample.zigbee.light_switch.current_measurement
+    - sample.zigbee.light_switch.with_shell
```
edmont, line 1280: This line is repeated. And what I understood from @mariusz-nordicsemi is that we want to skip it for all platforms, so what about moving it above after line 1217?

nordic-piks: Thanks, removed repeated line and moved at 1217.

edmont (approved):

nordicjm (changes requested):
Temp block to keep these running until sysbuild ncs is merged

nordic-piks:
Rebased after upmerge, to be merged after CI is green.

PR #15272, opened 2024-05-13
samples: cellular: gnss: Update GNSS factory almanacs

tokangas:
Updated GNSS factory almanacs.


lemrey (approved):

trantanen (approved):

samples/cellular/gnss/src/factory_almanac_v3.h
```diff
@@ -7,55 +7,55 @@
 #ifndef FACTORY_ALMANAC_V3_H_
 #define FACTORY_ALMANAC_V3_H_
 
-/* Factory almanac generated on 2024-04-02 09:22:49.
+/* Factory almanac generated on 2024-05-13 14:35:14.
```
trantanen, line 10: Is there a reason why we need to have 6 day gap between v2 and v3 almanacs? Usually I've seen it being an hour or maximum of a day.

tokangas: There were problems with almanac generation and it took longer to get v3 fixed.

PR #15274, opened 2024-05-13
samples: matter: Switch VCOM to the default UART20 on nRF54L15 PDK

ArekBalysNordic:
Disable previously used UART30 and switch to the default UART20. It means that from now the second port of the PDK's VCOM will be in use.

### DIFF
--- a/samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp_0_2_1.overlay
+++ b/samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp_0_2_1.overlay
@@ -5,11 +5,6 @@
  */
 
 / {
-	chosen {
-		zephyr,console = &uart30;
-		zephyr,shell-uart = &uart30;
-	};
-
 	aliases {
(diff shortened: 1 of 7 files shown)


kkasperczyk-no (approved):

doublemis1:
Why only Light bulb and window covering have defined overlay for 0.2.1 and 0.3.0 but other samples have just generic one?

kkasperczyk-no:
> Why only Light bulb and window covering have defined overlay for 0.2.1 and 0.3.0 but other samples have just generic one?

Because only they use PWM.

ArekBalysNordic:
> Why only Light bulb and window covering have defined overlay for 0.2.1 and 0.3.0 but other samples have just generic one?

Because only Light Bulb and Window covering use PWM, and this is the only change in DTS files. So if you build other samples using @0.3.0 board revision there is no additional board DTS file needed, because in that case, 0.2.1 overlay works fine and it is used automatically.

doublemis1 (approved):
LGTM!

LuDuda (approved):

PR #15277, opened 2024-05-13
SUIT extmem Kconfig adjustments

e-rk:
See commit descriptions.

### DIFF
--- a/subsys/suit/stream/Kconfig
+++ b/subsys/suit/stream/Kconfig
@@ -37,7 +37,6 @@ config SUIT_STREAM_SINK_RAM
 
 config SUIT_STREAM_SINK_EXTMEM
 	bool "Enable external memory sink"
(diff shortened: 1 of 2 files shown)


subsys/suit/stream/Kconfig
```diff
@@ -37,7 +37,6 @@ config SUIT_STREAM_SINK_RAM
 
 config SUIT_STREAM_SINK_EXTMEM
 	bool "Enable external memory sink"
-	depends on SSF_EXTMEM_SERVICE_ENABLED
```
tomchy, line 40: Does the code is prepared to work without the service or does this dependency is expressed in a different way?

e-rk: The config that should be in this statement is only defined in sdfw and is not accessible in sdk-nrf.

tomchy (approved):

adsz-nordic (approved):

PR #15282, opened 2024-05-13
drivers: wifi: Fix initialization of extended capabilities length

krish2718:
In case RPU is uninitialized e.g., FW loading failure, then WPA supplicant gets uninitialized extended capabilities from the driver.

Though this doesn't affect directly, but can cause issue without boundary checks. Initialized before the error checks.

Fixes SHEL-2763.

### DIFF
--- a/drivers/wifi/nrf700x/src/wpa_supp_if.c
+++ b/drivers/wifi/nrf700x/src/wpa_supp_if.c
@@ -1779,6 +1779,8 @@ int nrf_wifi_supp_get_capa(void *if_priv, struct wpa_driver_capa *capa)
 		return -1;
 	}
 
+	memset(capa, 0, sizeof(*capa));
+
 	vif_ctx_zep = if_priv;
 	rpu_ctx_zep = vif_ctx_zep->rpu_ctx_zep;
 	if (!rpu_ctx_zep) {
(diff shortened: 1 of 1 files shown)


rado17 (approved):

drivers/wifi/nrf700x/src/wpa_supp_if.c
```diff
@@ -1779,6 +1779,8 @@ int nrf_wifi_supp_get_capa(void *if_priv, struct wpa_driver_capa *capa)
 		return -1;
 	}
 
+	memset(capa, 0, sizeof(*capa));
```
ajayparida, line 1782: I didn't get. Why should we sinitialize it. What if the caller has set some filed set then we aill over write it.

krish2718: Well, the owner of the `wpa_driver_capa` is Driver, so, it won't be filled by WPA supplicant.

ajayparida: That I got about owner of capability.
I think memory is allocated by caller. it should be reponsible for it.'s initialization, 
we just need to use it.

krish2718: > I think memory is allocated by caller. it should be reponsible for it.'s initialization,

Caller ownes memory alloc/free, but contents are owned by this function, so, either it initialises or fills with some values etc is all part of the content population. And remember the initialisation is nothing new, it was there earlier as well, it was just moved.

ajayparida (approved):

PR #15285, opened 2024-05-14
boards: nordic: nrf7002dk: align nRF7001 shared SRAM

jciupis:
Align nRF7001 shared SRAM to fix crashes that occur immediately when running samples/wifi/shell with OpenThread overlay applied.


jciupis:
@krish2718, @Vge0rge please verify if this PR fixes the issue for you

Vge0rge (approved):
I can confirm that this fixes the issue with the wifi/shell sample and the target "nrf7002dk/nrf5340/cpuapp/nrf7001" with the openthread overlay.

krish2718 (approved):

PR #15288, opened 2024-05-14
Samples: Bluetooth: Mesh: Add FWID print on DFU build

Andrewpini:
Adds printing of the Firmware ID to build system for the Mesh DFU samples. The FWID is extracted from the .config file and printed in the bytestring format required by the mesh shell DFU interface. The intention is that this will make it easier for users to configure and run mesh DFU.


samples/bluetooth/mesh/dfu/target/sysbuild.cmake
```diff
@@ -0,0 +1,12 @@
...
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+add_custom_target(ble_mesh_print_dfu_fwid ALL
+  ${PYTHON_EXECUTABLE}
+  ${ZEPHYR_NRF_MODULE_DIR}/samples/bluetooth/mesh/dfu/common/script/ble_mesh_dfu_fwid.py
+  --bin-path ${PROJECT_BINARY_DIR}/${DEFAULT_IMAGE}/zephyr
+  DEPENDS parse_mesh_metadata
```
ludvigsj, line 11: Just a thought: Since this `DEPENDS`, (I think) compilation will fail if someone decides to disable `DFU_ZIP_BLUETOOTH_MESH_METADATA`, right? Could you add an `if` around this depending on that option? (Or: does this need to depend on that target at all? It doesn't use the metadata as far as I can tell, and should work as long as `.config` has been built?)

ludvigsj (changes requested):

samples/bluetooth/mesh/dfu/target/sysbuild.cmake
Andrewpini: Yes, I should add the if clause. To be honest the main reason for the depend is to ensure that the scrupt output is adjacent to the other DFU related build output. Since I'm no Cmake wiz im not sure how to ensure that without this dependency. Also it makes little sence to print the FWID if you are not printing the metadata, so there is a natural binding there IMo

ludvigsj (approved):

akredalen (approved):

PR #15290, opened 2024-05-14
kernel: Fix git commit banner when using worktrees

gregersrygg:
The gitdir (usually .git) does not exist in the nrf folder when using git worktrees.
This commit looks up the absolute path to the gitdir before adding the gitdir dependencies.

More info about [git worktrees in the official docs](https://git-scm.com/docs/git-worktree).

### DIFF
--- a/cmake/version.cmake
+++ b/cmake/version.cmake
@@ -21,6 +21,29 @@ add_custom_command(
 add_custom_target(ncs_version_h DEPENDS ${PROJECT_BINARY_DIR}/include/generated/ncs_version.h)
 add_dependencies(version_h ncs_version_h)
 
+find_package(Git QUIET)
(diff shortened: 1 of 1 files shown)


tejlmand (changes requested):
we need to depend on the index file to properly re-generate the header file when commit changes, so in order to properly support worktrees another approach then proposed in this PR is needed.

tejlmand (approved):
Approved, although we should probably have a common handler for this.

Seems the support for submodules actually ensured that worktrees was also working upstream.
https://github.com/zephyrproject-rtos/zephyr/blob/59212c198cf60f5c1380bb4a21eb9ff059ed4f85/CMakeLists.txt#L505-L533

simensrostad (approved):

PR #15291, opened 2024-05-14
nrf_security: Increase range of PSA key slots

Vge0rge:
Increase the range of the available PSA slots because some use cases require more than 64. The maximum number is an educated guess, I don't expect that any application will need more than 65K keys in our devices.

### DIFF
--- a/subsys/nrf_security/Kconfig.psa
+++ b/subsys/nrf_security/Kconfig.psa
@@ -67,7 +67,7 @@ config MBEDTLS_USE_PSA_CRYPTO
 config MBEDTLS_PSA_KEY_SLOT_COUNT
 	int "Number of PSA key slots available"
 	default 32
-	range 1 64
+	range 1 65535
 	help
 	  Describes the number of available PSA key slots. A key slot is used in PSA
 	  for each transparent key in use. So this number defines the sum of volatile and


Damian-Nordic (approved):
Thank you!

mswarowsky (approved):
let there be more keys

LuDuda (approved):

subsys/nrf_security/Kconfig.psa
```diff
@@ -67,7 +67,7 @@ config MBEDTLS_USE_PSA_CRYPTO
 config MBEDTLS_PSA_KEY_SLOT_COUNT
 	int "Number of PSA key slots available"
 	default 32
-	range 1 64
+	range 1 65535
```
frkv, line 70:
```suggestion
```

I suggest we drop the range, altogether. This is an arbitrary number and it will be limited by the the amount of RAM available on the system. Likely 65535 isn't even possible, at least not in a real-world application

Vge0rge: The annoying part with this is that 0 is not a valid number. So the range gives us something at least, meaning that it doesn't allow invalid configurations to break in some internal part of the code or even at runtime.  So I would say if we don't see that 65535 keys is a limitation we can keep the range.

PR #15299, opened 2024-05-14
manifest: update

SebastianBoe:
update


SebastianBoe:
redundant with another manifest PR

PR #15124, opened 2024-04-29
applications: nrf5340_audio: Use CAP procedures for unicast_server

alexsven:
- Use CAP procedure for starting and stopping streams
- Play/pause enabled for both sink and source in bidirectional mode
- OCT-2542

PR #15303, opened 2024-05-14
applications: nrf5340_audio: Audio lost when other headset disconnects

gWacey:
OCT-3008

For the stereo disconnection stress test, the headset which does not disconnect looses audio when the disconnecting headset disconnects, and never resumes audio. This was due to not testing if any of the other sinks were still streaming in the stream_released_cb().


alexsven (approved):

alexsven:
This is also fixed in https://github.com/nrfconnect/sdk-nrf/pull/15124, so I think we may close this one since it will cause a conflict with the other PR.

PR #15304, opened 2024-05-14
nrf_security: Support single-threaded Cracen

Vge0rge:
Support Cracen when multithreading is not enabled.

Created two macros for mutex handling which are empty when CONFIG_MULTITHREADING is disabled.

Add busy wait implementation for the Cracen interrupts since they use the Zephyr events APIs. The Zephyr events APIs require either multithreading or the Zephyr user mode which is not supported at the moment.

Ref: NCSDK-27385


subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
```diff
@@ -99,3 +102,42 @@ uint32_t cracen_wait_for_pke_interrupt(void)
...
+	while(!nrf_cracen_event_check(NRF_CRACEN, NRF_CRACEN_EVENT_CRYPTOMASTER)) {
+		/* Clear interrupts in CryptoMaster. */
+		sx_wrreg(REG_INT_STATCLR, ~0);
+
+		/* Clear Cracen wrapper interrupts.*/
+		nrf_cracen_event_clear(NRF_CRACEN, NRF_CRACEN_EVENT_CRYPTOMASTER);
+	}
+
+	return 1;
+}
```
SebastianBoe, line 128: This does not appear to be waiting for interrups?

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
```diff
@@ -39,6 +41,7 @@ void CRACEN_IRQHandler(void)
  */
 #endif
 
+
```
SebastianBoe, line 44: accidental whitespace I assume :)

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
```diff
@@ -99,3 +102,42 @@ uint32_t cracen_wait_for_pke_interrupt(void)
...
+	while(!nrf_cracen_event_check(NRF_CRACEN, NRF_CRACEN_EVENT_CRYPTOMASTER)) {
+		/* Clear interrupts in CryptoMaster. */
+		sx_wrreg(REG_INT_STATCLR, ~0);
+
+		/* Clear Cracen wrapper interrupts.*/
+		nrf_cracen_event_clear(NRF_CRACEN, NRF_CRACEN_EVENT_CRYPTOMASTER);
+	}
+
+	return 1;
```
vili-nordic, line 127: I think this should be return 0 ?

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
```diff
@@ -99,3 +102,42 @@ uint32_t cracen_wait_for_pke_interrupt(void)
...
+{
+	while(!nrf_cracen_event_check(NRF_CRACEN, NRF_CRACEN_EVENT_PKE_IKG)) {
+		/* Clear interrupts in PKE. */
+		sx_clear_interrupt(sx_get_current_req());
+
+		/* Clear Cracen wrapper interrupts.*/
+		nrf_cracen_event_clear(NRF_CRACEN, NRF_CRACEN_EVENT_PKE_IKG);
+	}
+
+	return 1;
```
vili-nordic, line 140: I think this should be return 0 ?

vili-nordic (commented):
generally looks fine to me, some questions about return values in the wait-for-interrupt functions

vili-nordic (approved):

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
```diff
@@ -99,3 +102,40 @@ uint32_t cracen_wait_for_pke_interrupt(void)
...
+	/* Clear interrupts in PKE. */
+	sx_clear_interrupt(sx_get_current_req());
+
+	/* Clear Cracen wrapper interrupts.*/
+	nrf_cracen_event_clear(NRF_CRACEN, NRF_CRACEN_EVENT_PKE_IKG);
+
+	return 1;
+}
+
+#endif /* CONFIG_EVENTS || __NRF_TFM__ */
```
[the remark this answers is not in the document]
SebastianBoe: The threadless TF-M solution also sleeps-until-irq instead of busy-waits, which is more power-efficient...

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
Vge0rge: You are right, I updated this now.

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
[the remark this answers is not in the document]
vili-nordic: If you can reuse tf-m solution, and sleep that would probably be nicer.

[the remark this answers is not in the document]
SebastianBoe: Discussed offline.

We'll define some kind of API that can have either a Zephyr_threaded implementation, or a CMSIS_wfe_threadless implementation.

subsys/nrf_security/src/CMakeLists.txt
```diff
@@ -266,3 +266,5 @@ if (NOT BUILD_INSIDE_TFM)
   # Add zephyr-specific code
   add_subdirectory(zephyr)
 endif()
+
+add_subdirectory(utils)
```
SebastianBoe, line 270: whitespace

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/cracen.c
```diff
@@ -23,7 +24,13 @@
...
 static int users;
 
-K_MUTEX_DEFINE(cracen_mutex);
+
+#if defined(CONFIG_MULTITHREADING) && !defined(__NRF_TFM__)
+K_MUTEX_DEFINE(k_cracen_mutex);
+nrf_security_mutex_t cracen_mutex = { .zephyr_mutex = &k_cracen_mutex };
+#else
+nrf_security_mutex_t cracen_mutex;
+#endif
```
SebastianBoe, line 33:
```suggestion
NRF_SECURITY_MUTEX_DEFINE(cracen_mutex);
```

If we define a macro like this and use it here, then users of the API won't need to know API details and it will therefore be more maintainable.

It could be defined something like this:

```
#if defined(CONFIG_MULTITHREADING) && !defined(__NRF_TFM__)
#define NRF_SECURITY_MUTEX_DEFINE(arg) K_MUTEX_DEFINE(k_##arg); nrf_security_mutex_t arg = { .zephyr_mutex = &k_##arg };
#else
nrf_security_mutex_t arg;
#endif
```

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
SebastianBoe: not fixed

subsys/nrf_security/src/utils/nrf_security_events.c
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
```
SebastianBoe, line 6: A comment explaining why this exists would be helpful for a future reader I think.

Something like

This event interface is inspured by Zephyr's event interface. There exists two implementations here. One is backed by Zephyr EVENTS, and other assumes there are no threads and uses cortex-m33 __WFE instructions.

It exists to support three scenarios. TF-M, which doesnt use multithreading. Zephyr with multithreading enabled, and Zephyr with multithreading disabled (bootloaders typically).

subsys/nrf_security/src/utils/nrf_security_events.c
```diff
@@ -0,0 +1,75 @@
...
+uint32_t nrf_security_event_set(nrf_security_event_t event, uint32_t events)
+{
+	uint32_t prev = event.val;
+
+	event.val |= events;
+
+	return prev;
+}
+
+#endif
```
SebastianBoe, line 75: It looks like we have forgotten to delete the old implementation of this.

subsys/nrf_security/src/utils/nrf_security_mutexes.c
```diff
@@ -0,0 +1,40 @@
...
+	(void) mutex;
+	return 0;
+}
+
+int nrf_security_mutex_unlock(nrf_security_mutex_t mutex)
+{
+	(void) mutex;
+	return 0;
+}
+#endif
```
SebastianBoe, line 40: I dont see the need for a struct that contains a single item.

I think a pointer to a typedef would work fine and be simpler.

But perhaps there is some problem with this that I cant see.

SebastianBoe (changes requested):
Much better.

But some feedback requested.

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
Vge0rge: Should be fixed now

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
Vge0rge: I think that now this is not relevant since I rewrited it, right?

subsys/nrf_security/src/CMakeLists.txt
Vge0rge: This is now removed I think

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/cracen.c
Vge0rge: I added this for both the events and mutexes

subsys/nrf_security/src/utils/nrf_security_events.c
Vge0rge: I used your comment thanks, I was planning to add something but glad that you wrote it for me :)

subsys/nrf_security/src/utils/nrf_security_events.c
Vge0rge: I did in a new commit now

subsys/nrf_security/src/utils/nrf_security_mutexes.c
Vge0rge: I converted everything to typedefs now

subsys/nrf_security/src/CMakeLists.txt
```diff
@@ -265,4 +265,4 @@ endif()
 if (NOT BUILD_INSIDE_TFM)
   # Add zephyr-specific code
   add_subdirectory(zephyr)
-endif()
+endif()
```
SebastianBoe, line 268: needs a newline

subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/platform/baremetal/interrupts.c
SebastianBoe: still not fixed

subsys/nrf_security/src/utils/nrf_security_events.h
```diff
@@ -0,0 +1,69 @@
...
+#if !defined(__NRF_TFM__)
+#include <zephyr/kernel.h>
+#endif
+
+/**
+ * @file nrf_security_events.h
+ *
+ * This event interface is inspired by Zephyr's event interface.
+ * There exists two implementations here.
+ * is backed by Zephyr events APIs (k_event_*), and other assumes
```
SebastianBoe, line 16:
```suggestion
 * One is backed by Zephyr events APIs (k_event_*), and the other assumes
```

subsys/nrf_security/src/utils/nrf_security_events.c
```diff
@@ -0,0 +1,74 @@
...
+
+	return *event & events_mask;
+}
+
+void nrf_security_event_clear(nrf_security_event_t event, uint32_t events_mask)
+{
+	*event &= ~events_mask;
+}
+
+uint32_t nrf_security_event_set(nrf_security_event_t event, uint32_t events_mask)
```
vili-nordic, line 65: This implementation does not match the doxygen comment in the header file

Vge0rge: I updated the comment, hopefully makes more sense now

vili-nordic: it still does not match?

k_event_set docs seem to indicate prev & events_mask should be the return value, and the updated documentation says prev | event_mask, and the implementation is a third option?

subsys/nrf_security/src/CMakeLists.txt
Vge0rge: This I cannot do it seems, compliance complains.

subsys/nrf_security/src/utils/nrf_security_events.c
Vge0rge: I think that you are right actually, I updated to to prev & events_mask. Hopefully CI will be happy with this.

vili-nordic: Looks like everything is matching now :smile_cat:

PR #15310, opened 2024-05-15
net: lwm2m_client_utils: Handle Block-wise FOTA on background

SeppoTakalo:
When we receive Block-Wise FOTA packets. We should immediately respond to incomming data. If we instead start erasing modem FOTA area, the server times out and starts retrying the first blockwise packet. This looks like FOTA starts multiple times in a row as the LwM2M engine does not detect duplicates on first block-wise packet. It is handled as a re-started Block-Wise.

Therefore handling of FOTA data is moved to system worker queue, so the LwM2M engine have time to respond to first packet. Semaphore might block the write request on the second packet, but if that is retried, LwM2M engine drops duplicates.

### DIFF
--- a/subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
+++ b/subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
@@ -58,6 +58,7 @@ static struct lwm2m_engine_res_inst pull_protocol_buff[FOTA_PULL_SUPPORTED_COUNT
 
 static lwm2m_firmware_event_cb_t firmware_fota_event_cb;
 static uint8_t firmware_buf[CONFIG_LWM2M_COAP_BLOCK_SIZE];
+static K_SEM_DEFINE(fw_buf_sem, 1, 1);
 
 /* Supported Protocols */
 static uint8_t pull_protocol_support[FOTA_PULL_SUPPORTED_COUNT] = { FOTA_PULL_COAP, FOTA_PULL_HTTP,
@@ -79,21 +80,10 @@ static K_THREAD_STACK_MEMBER(thread_stack,
 #define UNUSED_OBJ_ID 0xffff
 #define PENDING_DELAY K_MSEC(10)
 static uint16_t ongoing_obj_id;
-static uint8_t percent_downloaded;
-static uint32_t bytes_downloaded;
 static int application_obj_id;
 static int modem_obj_id;
 static int smp_obj_id;
 static int target_image_type[FOTA_INSTANCE_COUNT];
-
-
-static void dfu_target_cb(enum dfu_target_evt_id evt);
-static void start_pending_fota_download(struct k_work *work);
-static bool firmware_update_check_linked_instances(int instance_id, bool *reconnect);
-static int target_image_type_get(uint16_t instance);
-static uint8_t get_state(uint16_t id);
-static void set_result(uint16_t id, uint8_t result);
-static K_WORK_DELAYABLE_DEFINE(pending_download_work, start_pending_fota_download);
 static struct update_data {
 	struct k_work_delayable work;
 	enum {APP, MODEM_DELTA, MODEM_FULL, SMP} type;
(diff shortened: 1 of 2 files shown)


subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
```diff
@@ -674,97 +690,123 @@ static int firmware_block_received_cb(uint16_t obj_inst_id, uint16_t res_id, uin
...
+			.image_type = image_type,
+			.data = data,
+			.data_len = data_len,
+			.last_block = last_block,
+		};
+	} else {
+		if (push_data.return_code) {
+			/* If previous write was not succesfull, do not continue */
+			k_sem_give(&fw_buf_sem);
+			return push_data.return_code;
```
MarkusLassila, line 724: Does this mean that we still continue to receive blocks? Should we?

subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
```diff
@@ -674,97 +690,123 @@ static int firmware_block_received_cb(uint16_t obj_inst_id, uint16_t res_id, uin
...
-	ret = dfu_target_offset_get(&offset);
+	ret = dfu_target_offset_get(&dfu_offset);
 	if (ret < 0) {
 		LOG_ERR("Failed to obtain current offset, err: %d", ret);
 		goto cleanup;
 	}
 
+	push_data.bytes_downloaded += push_data.data_len;
+
+	if (dfu_offset > push_data.bytes_downloaded) {
```
MarkusLassila, line 779: So this means that we are retrying some data that was downloaded previously? Reading this I would expect that the download operation was started from the beginning as I do not understand how we would hit this otherwise.

It would be nice if the comment included why we ended up in the situation that we are. It kind of does, but could be more explicit.

MarkusLassila (commented):
Looks good! Some questions as I am not that familiar with LwM2M.

subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
SeppoTakalo: No, when we return the error code from block-write callback, the LwM2M engine return error to the CoAP Block-Wise transfer, usually something like "Internal error 500" so it stops the server from sending more blocks.

subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
SeppoTakalo: CoAP Block-Wise operation starts always from the beginning.
We have no way of telling server the offset where to skip to.

According to [rfc7959#section-2.9](https://datatracker.ietf.org/doc/html/rfc7959#section-2.9) we can only tell server to:
* "2.31 Continue", and it will send a next block.
* any other error code and the server stops sending.

juhaylinen (approved):

subsys/net/lib/lwm2m_client_utils/lwm2m/lwm2m_firmware.c
```diff
@@ -591,6 +599,16 @@ static int firmware_update_cb(uint16_t obj_inst_id, uint8_t *args, uint16_t args
 static void *firmware_get_buf(uint16_t obj_inst_id, uint16_t res_id, uint16_t res_inst_id,
 			      size_t *data_len)
 {
+	while (k_sem_take(&fw_buf_sem, K_SECONDS(1)) != 0) {
+		if (!k_work_is_pending(&push_data.work)) {
```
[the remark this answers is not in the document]
SeppoTakalo: No, because it does not call this module again, if it fails.

`package_write_cb()` is inside Zephyr's firmware object
https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/net/lib/lwm2m/lwm2m_obj_firmware.c#L253-L271
As you can see, there is 3 exit points before it call the `firmware_block_received_cb()` in our module.

The whole sequence can be checked from the `lwm2m_message_handling.c`
https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/net/lib/lwm2m/lwm2m_message_handling.c#L1049-L1242
1. it calls pre-write callback, which is this `firmware_get_buf()` which reserves the semaphore
2. it calls the post-write callback which is the `firmware_block_received_cb()`
3. that callback may call our `firmware_block_received()`, but only if it did not fail earlier.

So the whole purpose of semaphore is to block the second write of valid blocks. Or just delay, not completely block.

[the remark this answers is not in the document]
MarkusLassila: Thanks for the detailed explanation!

MarkusLassila (approved):

SeppoTakalo:
Rebased on top of main and refactored to use the `offset` parameter that came with the latest upmerge.

SeppoTakalo:
This is broken after latest block-wise fixes.. And apparently not so easily fixed, I need to change the approach.

LwM2M engine does not synchronously call pre-write-cb(); post_write_cb().
It might parse packet and call pre_write once, then post_write multiple times. So I cannot block it.

SeppoTakalo:
Closing as https://github.com/nrfconnect/sdk-nrf/pull/15825 fixed the original error. This approach cannot be used, as it turned out pre-write and post-write callbacks are not used synchronously.

PR #15311, opened 2024-05-15
lib: at_monitor: use dedicated workqueue

lemrey:
This patch adds a dedicated workqueue to dispatch AT notifications.

Earlier, AT notifications were dispatched from the system workqueue. This was RAM-efficient but led to problems because of the system workqueue "popularity" that made it used for many tasks, including blocking operations. Blocking the system workqueue would halt the dispatching of AT notifications, causing `at_monitor` to assert after running out of heap.

Dispatching from a dedicated workqueue mitigates this problem, which fundamentally lies in the application performing blocking operations in the system workqueue, at the cost of some RAM.

This is K-configurable, so the old behavior can be retained, if desired.

### DIFF
--- a/lib/at_monitor/Kconfig
+++ b/lib/at_monitor/Kconfig
@@ -14,8 +14,24 @@ config AT_MONITOR_HEAP_SIZE
 	range 64 4096
 	default 256
 
-config SYSTEM_WORKQUEUE_STACK_SIZE
-	default 1152 if (LTE_LINK_CONTROL && LOG)
+config AT_MONITOR_USE_DEDICATED_WORKQUEUE
+	bool "Use dedicated workqueue to dispatch notifications"
+	default y
+
+if AT_MONITOR_USE_DEDICATED_WORKQUEUE
+
+config AT_MONITOR_WORKQ_STACK_SIZE
+	int "Workqueue stack size"
+	default 768
+
+config AT_MONITOR_WORKQ_PRIO_OVERRIDE
+	bool "Override default workqueue priority"
+
+config AT_MONITOR_WORKQ_PRIO
+	int "Workqueue priority" if AT_MONITOR_WORKQ_PRIO_OVERRIDE
+	default -1
+
+endif # AT_MONITOR_USE_DEDICATED_WORKQUEUE
 
 module=AT_MONITOR
 module-dep=LOG
--- a/lib/at_monitor/at_monitor.c
+++ b/lib/at_monitor/at_monitor.c
@@ -20,11 +20,20 @@ struct at_notif_fifo {
 	char data[]; /* Null-terminated AT notification string */
 };
 
+/* Workqueue task to dispatch notifications */
 static void at_monitor_task(struct k_work *work);
+static K_WORK_DEFINE(at_monitor_work, at_monitor_task);
+
+/* Workqueue to dispatch notifications from */
+#if defined(CONFIG_AT_MONITOR_USE_DEDICATED_WORKQUEUE)
+static K_THREAD_STACK_DEFINE(at_monitor_work_q_stack, CONFIG_AT_MONITOR_WORKQ_STACK_SIZE);
+static struct k_work_q at_monitor_work_q;
+#else
+extern struct k_work_q k_sys_work_q;
+#endif
 
 static K_FIFO_DEFINE(at_monitor_fifo);
 static K_HEAP_DEFINE(at_monitor_heap, CONFIG_AT_MONITOR_HEAP_SIZE);
-static K_WORK_DEFINE(at_monitor_work, at_monitor_task);
 
 static bool is_paused(const struct at_monitor_entry *mon)
 {
@@ -83,7 +92,13 @@ void at_monitor_dispatch(const char *notif)
 	strcpy(at_notif->data, notif);
 
 	k_fifo_put(&at_monitor_fifo, at_notif);
+
+	/* Dispatch to queue */
+#if defined(CONFIG_AT_MONITOR_USE_DEDICATED_WORKQUEUE)
+	k_work_submit_to_queue(&at_monitor_work_q, &at_monitor_work);
+#else
 	k_work_submit(&at_monitor_work);
+#endif
 }
 
 static void at_monitor_task(struct k_work *work)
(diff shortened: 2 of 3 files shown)


lemrey:
I am unsure if this is the truly correct approach. The assert in `at_monitor` is showing us there is an issue in the application, somewhere. Perhaps we should just sift through the usage of the system workqueue in Cellular domain and make sure that no blocking operation is performed there.

lib/at_monitor/at_monitor.c
```diff
@@ -83,7 +94,9 @@ void at_monitor_dispatch(const char *notif)
 	strcpy(at_notif->data, notif);
 
 	k_fifo_put(&at_monitor_fifo, at_notif);
-	k_work_submit(&at_monitor_work);
+
+	/* Dispatch to queue */
+	k_work_submit_to_queue(&WORKQUEUE, &at_monitor_work);
```
trantanen, line 99: I think this is simpler to read (with corresponding changes to declarations):
```suggestion
	/* Dispatch to queue */
#if CONFIG_AT_MONITOR_USE_DEDICATED_WORKQUEUE
	k_work_submit_to_queue(&at_monitor_work_q, &at_monitor_work);
#else
	k_work_submit(&at_monitor_work);
#endif
```

lib/at_monitor/at_monitor.c
```diff
@@ -20,11 +20,22 @@ struct at_notif_fifo {
 	char data[]; /* Null-terminated AT notification string */
 };
 
+/* Workqueue task to dispatch notifications */
 static void at_monitor_task(struct k_work *work);
+static K_WORK_DEFINE(at_monitor_work, at_monitor_task);
+
+/* Workqueue to dispatch notifications from */
+#if CONFIG_AT_MONITOR_USE_DEDICATED_WORKQUEUE
```
trantanen, line 28: I'm seeing people mostly using `#if defined`. Is there a specific reason for using plain `#if`?

lib/at_monitor/Kconfig
```diff
@@ -14,8 +14,24 @@ config AT_MONITOR_HEAP_SIZE
 	range 64 4096
 	default 256
 
-config SYSTEM_WORKQUEUE_STACK_SIZE
-	default 1152 if (LTE_LINK_CONTROL && LOG)
+config AT_MONITOR_USE_DEDICATED_WORKQUEUE
```
trantanen, line 17: I'm hesitant to give application developers this option because of ~1,4kB of RAM out of which some can potentially be saved in system work queue stack size.

lib/at_monitor/at_monitor.c
```diff
@@ -107,6 +120,16 @@ static int at_monitor_sys_init(void)
 {
 	int err;
 
+#if CONFIG_AT_MONITOR_USE_DEDICATED_WORKQUEUE
+	const struct k_work_queue_config cfg = {.name = "at_monitor_work_q"};
+
+	k_work_queue_start(&at_monitor_work_q,
+			   at_monitor_work_q_stack,
+			   K_THREAD_STACK_SIZEOF(at_monitor_work_q_stack),
+			   K_LOWEST_APPLICATION_THREAD_PRIO,
```
trantanen, line 129: `CONFIG_AT_MONITOR_WORKQ_PRIO` not taken into account

lib/at_monitor/Kconfig
```diff
@@ -14,8 +14,24 @@ config AT_MONITOR_HEAP_SIZE
...
+	bool "Use dedicated workqueue to dispatch notifications"
+	default y
+
+if AT_MONITOR_USE_DEDICATED_WORKQUEUE
+
+config AT_MONITOR_WORKQ_STACK_SIZE
+	int "Workqueue stack size"
+	default 768
```
trantanen, line 25: Modem shell requires 1016 bytes in the boot. I'd say that minimum default that we should put here is 1024 but should it be a bit more. I haven't been able to get mosh usage above that 1016 bytes yet.

lib/at_monitor/at_monitor.c
trantanen, line 129: Should we keep the current priority that we have from the system work queue? I'm trying to say that this priority should probably be higher than most libraries.

lib/at_monitor/at_monitor.c
lemrey: ok, updated

lib/at_monitor/at_monitor.c
lemrey: Thanks, updated

lib/at_monitor/Kconfig
lemrey: Should I update this value in `mosh` 's `prj.conf` then? I'd like to keep this low by default.

lib/at_monitor/at_monitor.c
lemrey: Agree, I changed it to `-1` by default

lib/at_monitor/Kconfig
lemrey: This is enabled by default, but it's good to keep a it behind a flag for now. Some users might not need this (if they don't use `lte_lc` for example), and would rather save RAM instead.

lib/at_monitor/at_monitor.c
```diff
@@ -20,11 +20,20 @@ struct at_notif_fifo {
...
+/* Workqueue task to dispatch notifications */
 static void at_monitor_task(struct k_work *work);
+static K_WORK_DEFINE(at_monitor_work, at_monitor_task);
+
+/* Workqueue to dispatch notifications from */
+#if defined(CONFIG_AT_MONITOR_USE_DEDICATED_WORKQUEUE)
+static K_THREAD_STACK_DEFINE(at_monitor_work_q_stack, CONFIG_AT_MONITOR_WORKQ_STACK_SIZE);
+static struct k_work_q at_monitor_work_q;
+#else
+extern struct k_work_q k_sys_work_q;
```
trantanen, line 32: `k_sys_work_q` is not used anymore so else-branch could be deleted.

lib/at_monitor/at_monitor.c
```diff
@@ -107,6 +122,19 @@ static int at_monitor_sys_init(void)
 {
 	int err;
 
+#if defined(CONFIG_AT_MONITOR_USE_DEDICATED_WORKQUEUE)
+	/* Default to a cooperative priority to avoid being preempted */
+#define THREAD_PRIORITY                                                                            \
+	COND_CODE_1(CONFIG_AT_MONITOR_WORKQ_PRIO_OVERRIDE, (CONFIG_AT_MONITOR_WORKQ_PRIO), (-1))
+
```
tokangas, line 129:
```suggestion
```

lib/at_monitor/at_monitor.c
```diff
@@ -107,6 +122,19 @@ static int at_monitor_sys_init(void)
...
+	const struct k_work_queue_config cfg = {.name = "at_monitor_work_q"};
+
+	k_work_queue_start(&at_monitor_work_q,
+			   at_monitor_work_q_stack,
+			   K_THREAD_STACK_SIZEOF(at_monitor_work_q_stack),
+			   THREAD_PRIORITY, &cfg);
```
tokangas, line 135:
```suggestion
			   CONFIG_AT_MONITOR_WORKQ_PRIO, &cfg);
```

lemrey:
This may not solve the issue where the AT monitors execute blocking code in their handlers, thus blocking the dispatching of AT notifications. Let's wait and see if that issue arises again.

PR #15314, opened 2024-05-15
applications: nrf_desktop: Make set/get report more generic in USB state

MarekPieta:
Change makes set_report and get_report functions more generic to allow code reuse.

Jira: NCSDK-27014

### DIFF
--- a/applications/nrf_desktop/src/modules/usb_state.c
+++ b/applications/nrf_desktop/src/modules/usb_state.c
@@ -82,44 +82,21 @@ static struct usb_hid_device *dev_to_hid(const struct device *dev)
 	return usb_hid;
 }
 
-static int get_report(const struct device *dev,
-		      struct usb_setup_packet *setup,
-		      int32_t *len_to_set, uint8_t **data)
+static int get_report(const struct device *dev, uint8_t report_type, uint8_t report_id,
+		      uint8_t *buf, size_t size)
 {
-	uint8_t request_value[2];
+	int err = -ENOTSUP;
 
-	sys_put_le16(setup->wValue, request_value);
-
-	switch (request_value[1]) {
+	switch (report_type) {
 	case REPORT_TYPE_FEATURE:
 		if (IS_ENABLED(CONFIG_DESKTOP_CONFIG_CHANNEL_ENABLE) &&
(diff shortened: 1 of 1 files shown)


applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -630,11 +581,84 @@ static void verify_report_bm(void)
...
+
+static int set_report_legacy(const struct device *dev, struct usb_setup_packet *setup,
+			     int32_t *len, uint8_t **data)
+{
+	uint8_t request_value[2];
+
+	sys_put_le16(setup->wValue, request_value);
+
+	uint8_t report_type = request_value[1];
+	uint8_t report_id = request_value[0];
```
kapi-no, line 625: common part, consider defining inline function

```
report_setup_value_parse(uint16_t value, uint8_t *report_type, uint8_t *report_id);
```

kapi-no (approved):
Looking good, a few minor improvements

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -630,11 +581,84 @@ static void verify_report_bm(void)
...
+		*len_to_set = setup->wLength;
+	}
+
+	return err;
+}
+
+static int set_report_legacy(const struct device *dev, struct usb_setup_packet *setup,
+			     int32_t *len, uint8_t **data)
+{
+	uint8_t request_value[2];
```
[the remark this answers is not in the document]
MarekPieta: Will do

applications/nrf_desktop/src/modules/usb_state.c
MarekPieta: Since it's used only in 2 places and the inline function would return 2 values through pointers, I would prefer to desist from doing it (prefer to keep parsing explicitly here).

mkapala-nordic (approved):

applications/nrf_desktop/src/modules/usb_state.c
MarekPieta: Done

kapi-no (approved):

mkapala-nordic (approved):

PR #15319, opened 2024-05-15
mpsl: cx: add software coexistence implementation

Damian-Nordic:
Add an implementation of the MPSL Coexistence interface, which allows an application to control the allowed short range radio operations using software API.

Add nRF RPC layer on top of the MPSL Software Coexistence, which allows to control the granted operations remotely.


canisLupus1313 (approved):

bpienk (approved):

jciupis:
@ankuns you're not listed as a codeowner yet but I think your approval is essential for merging this PR.
@Damian-Nordic please fix compliance check, it's failing on a missing entry in CODEOWNERS file.

jciupis:
@Damian-Nordic please rebase to allow the CI to run a meaningful build

subsys/mpsl/cx/Kconfig
```diff
@@ -80,10 +81,34 @@ config MPSL_CX_BT_1WIRE
...
 module=MPSL_CX
 module-str=MPSL_CX
 source "${ZEPHYR_BASE}/subsys/logging/Kconfig.template.log_config"
 
 endif	# MPSL_CX
+
+config MPSL_CX_SOFTWARE_RPC
+	bool
+
+config MPSL_CX_SOFTWARE_RPC_CLIENT
```
jciupis, line 101: Since `MPSL_CX_SOFTWARE` is one of the options in `MPSL_CX_CHOICE` along with hardware PTA implementations, I'd suggest creating a Kconfig interface similar to the existing approach. For multicore architectures, that means setting the same options for the application core and the core that controls the RADIO peripheral, e.g.
```
west build -p -b nrf5340dk/nrf5340/cpuapp -- -DCONFIG_MPSL_CX=y -DCONFIG_MPSL_CX_NRF700X=y
-Dhci_ipc_CONFIG_MPSL_CX=y -Dhci_ipc_CONFIG_MPSL_CX_NRF700X=y
```
To align with this approach I'd suggest to make `MPSL_CX_SOFTWARE_RPC_...` symbols hidden. Their values would need to be calculated based on the target SoC similarly to `MPSL_CX_PIN_FORWARDER`. Then, the application would only set `CONFIG_MPSL_CX=y` and `CONFIG_MPSL_CX_SOFTWARE=y` and not care about the target platform.

Note that the file you've introduced: `applications/protocols_serialization/client/snippets/coex/coex.conf` needs to set the RPC option explicitly, which breaks the independency from target platform. The snippet by design cannot be applied to single core architectures, which kind of defeats its purpose.

Damian-Nordic: To give you more context, we're building a prototype in which application runs on a different SoC than the OpenThread stack, and they communicate with each other using UART (and nRF RPC). That's why I cannot rely on automatic detection of the client and server role because a user may select any SoC for either role.

Well, I could probably infer that if `CONFIG_MPSL_CX_SOFTWARE && !MPSL` then it means `MPSL_CX_SOFTWARE_RPC_CLIENT` but not sure if this isn't more cryptic. Let me know if you have a better idea having the context :)

Damian-Nordic:
> @Damian-Nordic please rebase to allow the CI to run a meaningful build

Note that this is a PR to a branch where we build our prototype. We will soon start merging it to `main` where the CI will be properly validated.

include/mpsl/mpsl_cx_software.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <stdint.h>
+
+#ifdef CONFIG_MPSL_CX_SOFTWARE_RPC_CLIENT
+/*
+ * MPSL Software Coexistence RPC client provides serialization of the MPSL Software Coexistence
+ * APIs, but it does not pull the actual MPSL Radio Coexistence library into the build system.
+ * Hence the 'mpsl_cx_op_map_t' must be redefined below.
+ */
+typedef uint8_t mpsl_cx_op_map_t;
```
ankuns, line 33: I think you should provide `mpsl_cx_op_t` type from
https://github.com/nrfconnect/sdk-nrfxlib/blob/main/mpsl/include/protocol/mpsl_cx_protocol_api.h#L43

along with `mpsl_cx_op_map_t`. Otherwise we won't have values to build `ops` parameter of `mpsl_cx_software_set_granted_ops`

subsys/mpsl/cx/software/mpsl_cx_software_rpc.c
```diff
@@ -0,0 +1,75 @@
...
+
+	if (K_TIMEOUT_EQ(timeout, K_FOREVER)) {
+		timeout_ms = -1;
+	} else {
+		timeout_ms = k_ticks_to_ms_near64(timeout.ticks);
+	}
+
+	NRF_RPC_CBOR_ALLOC(&mpsl_cx_rpc_group, ctx, 11);
+	zcbor_uint_encode(ctx.zs, &ops, sizeof(ops));
+	zcbor_uint64_put(ctx.zs, timeout_ms);
```
ankuns, line 42: shouldn't be `zcbor_int64_put` ? `timeout_ms` is of `int64_t` type

Damian-Nordic: Good catch, thanks!

subsys/mpsl/cx/Kconfig
jciupis: Thanks for the additional context. I still think that if `MPSL_CX_SOFTWARE` is one of the options in `MPSL_CX_CHOICE`, the interface should be the same for consistency. `CONFIG_MPSL_CX_SOFTWARE && !CONFIG_MPSL` doesn't seem cryptic at all to me, quite the opposite actually ;) it's a good proposal.

SylwesterKonczyk (approved):

subsys/mpsl/cx/Kconfig
Damian-Nordic: Implemented the described pattern (I hope).

ankuns (approved):

PR #15326, opened 2024-05-15
scripts: ci: Detect API-breaking changes in the PRs

doki-nordic:
This script will check the PR's changes in the header files and DTS bindings that may affects compatibility of the public API. This will reduce risk of accidentally breaking the API. Workflow that runs on each PR will add a comment with analysis summary.

### DIFF
--- a/.github/workflows/api-check.yml
+++ b/.github/workflows/api-check.yml
@@ -1,0 +1,129 @@
+name: API Check
+
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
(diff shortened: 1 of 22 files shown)


wbober:
I didn't go deep into the implementation but I would suggest to:
1) add a readme with some details how to use the script, what args it takes, etc.
2) brief description on how it works
3) I'd move the table that details current detection rules into the repository.

wbober (approved):

shanthanordic (approved):

PR #15328, opened 2024-05-16
lib: lte_link_control: New workqueue for event dispatching

trantanen:
LTE link control library has been dispatching its events often in the AT monitors/handlers. These have been called in the system workqueue by the at_monitor library.
Even if at_monitor would implement dedicated workqueue for calling AT monitors, it would still be a good idea to call lte_lc event handlers in a separate workqueue.

A related PR exists for `at_monitor` library: #15311
We will decide later whether this and/or PR #15311 will be merged.

Jira: NCSDK-26227

### DIFF
--- a/lib/lte_link_control/lte_lc.c
+++ b/lib/lte_link_control/lte_lc.c
@@ -1802,6 +1802,11 @@ int lte_lc_factory_reset(enum lte_lc_factory_reset_type type)
 	return nrf_modem_at_printf("AT%%XFACTORYRESET=%d", type) ? -EFAULT : 0;
 }
 
+#define LTE_LC_EVENT_STACK_SIZE 1024 /* TODO: Make size Kconfig option */
+K_THREAD_STACK_DEFINE(lte_lc_event_stack, LTE_LC_EVENT_STACK_SIZE);
+
+struct k_work_q lte_lc_event_work_q;
+
 static int lte_lc_sys_init(void)
 {
 	struct k_work_queue_config cfg = {
@@ -1814,6 +1819,17 @@ static int lte_lc_sys_init(void)
 		K_THREAD_STACK_SIZEOF(lte_lc_psm_get_stack),
 		K_LOWEST_APPLICATION_THREAD_PRIO,
 		&cfg);
+
+	struct k_work_queue_config cfg_event = {
+		.name = "lte_lc_event_work_q",
+	};
+
+	k_work_queue_start(
+		&lte_lc_event_work_q,
+		lte_lc_event_stack,
+		K_THREAD_STACK_SIZEOF(lte_lc_event_stack),
+		K_LOWEST_APPLICATION_THREAD_PRIO,
+		&cfg_event);
 
 	return 0;
 }
--- a/lib/lte_link_control/lte_lc_helpers.c
+++ b/lib/lte_link_control/lte_lc_helpers.c
@@ -30,6 +30,23 @@ struct event_handler {
 
 static sys_slist_t handler_list;
 
+struct lte_lc_event_fifo_item {
+	void *fifo_reserved;
+	struct lte_lc_evt evt;
+	struct lte_lc_ncell *neighbor_cells;
+	struct lte_lc_cell *gci_cells;
+};
+
+static void lte_lc_event_work_fn(struct k_work *work);
+
+#define LTE_LC_EVENT_HEAP_SIZE 512 /* TODO: Make size Kconfig option */
+
+static K_FIFO_DEFINE(lte_lc_event_fifo);
+static K_HEAP_DEFINE(lte_lc_event_heap, LTE_LC_EVENT_HEAP_SIZE);
+static K_WORK_DEFINE(lte_lc_event_work, lte_lc_event_work_fn);
+
+extern struct k_work_q lte_lc_event_work_q;
+
 /**
  * @brief Find the handler from the event handler list.
  *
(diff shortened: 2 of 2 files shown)


trantanen:
This PR is made also for testing purposes and we may decide not to go for this solution. The code is not final in terms of all style issues but it has relevant elements from memory and workqueue handling perspective:

- new workqueue
- stack for the new workqueue
- FIFO for events to be sent
- HEAP to allocate events from the heap

lib/lte_link_control/lte_lc.c
```diff
@@ -1802,6 +1802,11 @@ int lte_lc_factory_reset(enum lte_lc_factory_reset_type type)
 	return nrf_modem_at_printf("AT%%XFACTORYRESET=%d", type) ? -EFAULT : 0;
 }
 
+#define LTE_LC_EVENT_STACK_SIZE 1024 /* TODO: Make size Kconfig option */
+K_THREAD_STACK_DEFINE(lte_lc_event_stack, LTE_LC_EVENT_STACK_SIZE);
+
+struct k_work_q lte_lc_event_work_q;
```
trantanen, line 1808: I'm not sure if we need a new workqueue but using the existing one where AT commands are sent may be troublesome too.

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -30,6 +30,21 @@ struct event_handler {
...
+	void *fifo_reserved;
+	struct lte_lc_evt evt;
+};
+
+static void lte_lc_event_work_fn(struct k_work *work);
+
+#define LTE_LC_EVENT_HEAP_SIZE 512 /* TODO: Make size Kconfig option */
+
+static K_FIFO_DEFINE(lte_lc_event_fifo);
+static K_HEAP_DEFINE(lte_lc_event_heap, LTE_LC_EVENT_HEAP_SIZE);
```
tokangas, line 43: Instead of using separate heap for storing the events, how about using a ring buffer? The Zephyr ring buffer seems to also have a data item mode. Or maybe a message queue, which I believe internally uses the ring buffer implementation.

tokangas: I think FIFO is targeted more for arbitrary size data. The events are fixed size, so there are probably simpler solutions available, which don't need to dynamically allocate memory.

lib/lte_link_control/lte_lc.c
trantanen: I'm afraid we need to use a dedicated workqueue to avoid getting stuck when AT%XMONITOR or AT%XPTW is sent.

lib/lte_link_control/lte_lc_helpers.c
trantanen: Yep, this was just copied from AT monitor quickly and we can change the structure and where it's allocated.

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -111,26 +126,54 @@ int event_handler_list_remove_handler(lte_lc_evt_handler_t handler)
...
+	if (!lte_lc_event_item) {
+		LOG_WRN("No heap space for outgoing lte_lc event type=%d", evt->type);
+		__ASSERT(
+			lte_lc_event_item,
+			"No heap space for lte_lc notification type=%d",
+			evt->type);
+		return;
+	}
+
+	memcpy(&lte_lc_event_item->evt, evt, sizeof(struct lte_lc_evt));
```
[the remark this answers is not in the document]
trantanen: Yep, need to see how this can be resolved. Need to have some additional dynamic memory allocation for those to avoid always reserving the memory for NCELLMEAS results.

lib/lte_link_control/lte_lc_helpers.c
trantanen: > I think FIFO is targeted more for arbitrary size data. The events are fixed size, so there are probably simpler solutions available, which don't need to dynamically allocate memory.

When you don't dynamically allocate the memory, you need to specify the maximum size.
I think the best would be if we could guarantee certain small size and then use extra memory if available with k_malloc. Same discussion should apply to `at_monitor`.

lib/lte_link_control/lte_lc_helpers.c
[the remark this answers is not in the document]
trantanen: I've made now allocation and free functions for the event structures. At this point they are using k_malloc and k_free for ncells and gci_cells but that can be changed once we decide where we want to allocate them.

lib/lte_link_control/lte_lc_helpers.c
tokangas: > When you don't dynamically allocate the memory, you need to specify the maximum size.
> I think the best would be if we could guarantee certain small size and then use extra memory if available with k_malloc. Same discussion should apply to at_monitor.

It's not different from the original implementation, where the `LTE_LC_EVENT_HEAP_SIZE` is the limiting factor. In practice that defines the maximum number of events which can be stored. Even though the allocation is dynamic, the pool from which the allocation is done is static.

lib/lte_link_control/lte_lc_helpers.c
```diff
@@ -108,29 +125,110 @@ int event_handler_list_remove_handler(lte_lc_evt_handler_t handler)
...
+{
+	struct lte_lc_event_fifo_item *lte_lc_event_item;
+
+	lte_lc_event_item = k_heap_alloc(
+				&lte_lc_event_heap,
+				sizeof(struct lte_lc_event_fifo_item),
+				K_NO_WAIT);
+	if (!lte_lc_event_item) {
+		LOG_WRN("No LTE LC heap space for event type=%d", evt->type);
+		lte_lc_event_item = k_malloc(sizeof(struct lte_lc_event_fifo_item));
```
trantanen, line 138: I've made this allocation to provoke discussion also for `at_monitor` PR #15311.
If we do something like this. We can reduce the dedicated heap (lte_lc_event_heap) size a bit and use kernel heap (or even system heap) when available.

tokangas: Wouldn't you have to call `k_free()` for memory allocated with `k_malloc()` and `k_heap_free()` for memory allocated with `k_heap_alloc()`?

That said, I think this gets overly complicated. Are we trying to solve a real problem here? Do we have any numbers how many pending events there could be?

trantanen: > Wouldn't you have to call `k_free()` for memory allocated with `k_malloc()` and `k_heap_free()` for memory allocated with `k_heap_alloc()`?

Yep. Needs to be handled.

> That said, I think this gets overly complicated. Are we trying to solve a real problem here? Do we have any numbers how many pending events there could be?

Well, we either need to reserve extra memory for cases that may happen (and even then we are not sure if it's enough), or we will have fatal issues in form of an assert or dropping events.
I think we at least should have a discussion on something like this if we decide to dispatch lte_lc events from a dedicated workqueue.

trantanen: This was discussed in a meeting and we should not use multiple heaps for allocating memory.

trantanen:
This doesn't solve the issues seen when a blocking operation is performed in LTE LC event handlers or in system workqueue although it would make failure occur a little later. The root cause would still exist.
See Jira NCSDK-26227.

PR #15331, opened 2024-05-16
samples: cellular: modem_shell: Fix P-GPS prediction injection

tokangas:
After recent changes to the nRF Cloud library, P-GPS predictions were no longer being injected. This is now fixed by setting initial A-GNSS data need, so that time and ephemerides are injected also before GNSS has requested for assistance.


tokangas:
I created PR #15336 to address the problem in the nRF Cloud P-GPS library. If that gets merged, then this PR is not needed and can be closed, so marked this as DNM for now.

PR #15337, opened 2024-05-16
Mitigate risk of pin security misconfiguration

SebastianBoe:
Mitigate the risk of pin security misconfiguration by giving more reasonable default values and by logging which pins are configured as secure.


vili-nordic (approved):

mswarowsky (approved):
Good catch

PR #15343, opened 2024-05-16
applications: nrf5340_audio: SYS_INIT channel assignment module

andvib:
Move initialization of the channel assignment module to APPLICATION level of SYS_INIT. This ensures that the module is initialized correctly before any code references it - regardless if app is running on audio DK or not.

OCT-3029

### DIFF
--- a/applications/nrf5340_audio/src/utils/channel_assignment.c
+++ b/applications/nrf5340_audio/src/utils/channel_assignment.c
@@ -35,7 +35,7 @@ void channel_assignment_set(enum audio_channel channel)
 }
 #endif /* CONFIG_AUDIO_HEADSET_CHANNEL_RUNTIME */
 
-void channel_assignment_init(void)
+static int channel_assignment_init(void)
 {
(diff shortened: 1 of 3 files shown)


applications/nrf5340_audio/src/utils/channel_assignment.c
```diff
@@ -35,7 +35,7 @@ void channel_assignment_set(enum audio_channel channel)
 }
 #endif /* CONFIG_AUDIO_HEADSET_CHANNEL_RUNTIME */
 
-void channel_assignment_init(void)
```
rick1082, line 38: Not sure: maybe we could still make this API explicitly?

rick1082 (approved):

gWacey (approved):

applications/nrf5340_audio/src/utils/channel_assignment.c
andvib: I'm not sure what you mean by explicit in this context, can you elaborate?

PR #15344, opened 2024-05-16
suit: add option to erase cache on init

e-rk:
The SUIT cache partitions 1..n may contain left-over data after the last update. To ensure that the cache does not contain left-over data, the partition is erased if it were not empty already.
This can be enabled or disabled with SUIT_CACHEX_ERASE_ON_INIT Kconfig.

### DIFF
--- a/subsys/suit/cache/Kconfig
+++ b/subsys/suit/cache/Kconfig
@@ -52,6 +52,12 @@ config SUIT_CACHE0_ERASE_ON_ENVELOPE_STORED
 	depends on SUIT_CACHE_RW
 	default y
 
+config SUIT_CACHEX_ERASE_ON_INIT
+	bool "Always erase the DFU cache partition 1 .. n on initialization"
+	help
+	  If this option is set the SUIT caches with ID 1 .. n will be erased
+	  during cache initialization, if the cache partition is not already erased.
+
 config SUIT_CACHEX_ERASE_BLOCK_SIZE
 	int "Erase block size"
 	default 4096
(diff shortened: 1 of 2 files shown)


tomchy (approved):

subsys/suit/cache/src/suit_dfu_cache_rw.c
```diff
@@ -606,7 +617,7 @@ static suit_plat_err_t cache_0_update(uint8_t *address, size_t size)
 		}
 #endif /* CONFIG_SUIT_CACHE0_ERASE_ON_ENVELOPE_STORED */
 
-		return partition_initialize(&dfu_partitions_ext[0]);
+		return partition_initialize(&dfu_partitions_ext[0], false);
```
ahasztag, line 620: I think instead of `false` it should be  `IS_ENABLED(CONFIG_SUIT_CACHE0_ERASE_ON_ENVELOPE_STORED)` and delete the part above

subsys/suit/cache/Kconfig
```diff
@@ -52,6 +52,13 @@ config SUIT_CACHE0_ERASE_ON_ENVELOPE_STORED
 	depends on SUIT_CACHE_RW
 	default y
 
+config SUIT_CACHEX_ERASE_ON_INIT
+	bool "Always erase the DFU cache partition 1 .. n on initialization"
+	default y
```
ahasztag, line 57: This is changing the behavior, I think it should be `default n`

ahasztag (changes requested):

subsys/suit/cache/Kconfig
e-rk: Removed the default

subsys/suit/cache/src/suit_dfu_cache_rw.c
e-rk: Done

ahasztag (approved):

adsz-nordic (approved):

PR #15347, opened 2024-05-16
Samples: Bluetooth: Mesh: Increase resources for sensor server

Andrewpini:
Increases the resources for delayable messages and segmented messages to avoid warning and error output under "normal" usage of the sensor server sample.


samples/bluetooth/mesh/sensor_server/prj.conf
```diff
@@ -65,6 +65,12 @@ CONFIG_BT_MESH_SENSOR_SRV=y
 CONFIG_BT_MESH_SENSOR_SRV_SENSORS_MAX=5
 CONFIG_BT_MESH_SENSOR_USE_LEGACY_SENSOR_VALUE=n
 
+# Increase default delayable message config to avoid insuffisient resources
+# during normal sample operation
+CONFIG_BT_MESH_ACCESS_DELAYABLE_MSG_CHUNK_COUNT=60
+CONFIG_BT_MESH_TX_SEG_MSG_COUNT=2
```
alxelax, line 71: Why does it require 2 Tx sar contexts? There are only 2 devices in sample. Two sar transactions to the same address are not allowed.

omkar3141: You could be doing Config Server operations (that result in long responses) and simultaneously fetching Sensor_Series_Get. If these two things happen together, sample will start printing errors and models will fail to send second SAR message, even when DSTs are different.

omkar3141 (approved):

samples/bluetooth/mesh/sensor_server/prj.conf
alxelax: Sensor model transactions before or in parallel sensor model configuration? Hm... sounds doubtful.

Andrewpini: Hmm... Well before increasing this, the following error was triggered from time to time: 
https://github.com/zephyrproject-rtos/zephyr/blob/d651be5d70322e24b8940ae0c3a68a7d552cce38/subsys/bluetooth/mesh/transport.c#L508

If you are correct in your statement, I have no Idea why increasing this makes a difference.

alxelax: However, that might be the case when client received transport ack and started the new sar transaction while server still sends transport acks. Could you set sar configuration to single transport ack (without repeats)?

Andrewpini: The `BT_MESH_SAR_RX_ACK_RETRANS_COUNT` was set to 0 in the sensor server sample, so I have no idea what is causing this. @alxelax maybe we should make a investigation task/bug report to investigate this further? If you want I can remove the mentioned option from this PR for now, but then ofc it will produce error output from time to time.

alxelax: Ok, leave your improvements `as is`. I guess it is worth to investigate the reason of the issue. I'm Ok if it is done it as a separate task.

samples/bluetooth/mesh/sensor_server/prj.conf
```diff
@@ -65,6 +65,12 @@ CONFIG_BT_MESH_SENSOR_SRV=y
...
+CONFIG_BT_MESH_ACCESS_DELAYABLE_MSG_COUNT=6
```
ludvigsj, line 72: Does this really solve anything? As far as I could tell, it was the number of chunks that ran out, not the number of messages?

Andrewpini: Fixing the first issue caused new error/wrn message output.

ludvigsj: Ahh right because it ran out of chunks and then never got to the point where it ran out of messages. Makes sense.

ludvigsj (approved):

PR #15348, opened 2024-05-16
Samples: Bluetooth: Mesh: add nRF54L15 support to light_ctrl

ludvigsj:
This adds support for the nRF54L15 PDK to the light_ctrl sample. It changes the interrupt line used, as the one that was there was not available on the nRF54L15.

Note: Currently, there is an issue with PWM LEDs on the nRF54L15 PDK where the LED dims inverted. I have not handled this here, since the same incorrect behavior shows up in other samples and I expect it to be fixed in Zephyr.

### DIFF
--- a/samples/bluetooth/mesh/light_ctrl/src/main.c
+++ b/samples/bluetooth/mesh/light_ctrl/src/main.c
@@ -21,7 +21,14 @@
 #include <mpsl/mpsl_lib.h>
 #endif
 
-#define EMDS_DEV_IRQ 26
+#if defined(CONFIG_SOC_SERIES_NRF52X)
+	#define EMDS_DEV_IRQ SWI1_EGU1_IRQn
+#elif defined(CONFIG_SOC_SERIES_NRF53X)
+	#define EMDS_DEV_IRQ EGU1_IRQn
+#elif defined(CONFIG_SOC_SERIES_NRF54LX)
+	#define EMDS_DEV_IRQ SWI01_IRQn
+#endif
+
 #define EMDS_DEV_PRIO 0
 #define EMDS_ISR_ARG 0
 #define EMDS_IRQ_FLAGS 0
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -454,7 +454,10 @@ Bluetooth Mesh samples
 
 * :ref:`bluetooth_mesh_light_lc` sample:
 
-  * Added a section about the :ref:`occupancy mode <bluetooth_mesh_light_lc_occupancy_mode>`.
+  * Added
+
+    * A section about the :ref:`occupancy mode <bluetooth_mesh_light_lc_occupancy_mode>`.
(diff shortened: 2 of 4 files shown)


Andrewpini (approved):

samples/bluetooth/mesh/light_ctrl/src/main.c
```diff
@@ -21,7 +21,7 @@
 #include <mpsl/mpsl_lib.h>
 #endif
 
-#define EMDS_DEV_IRQ 26
+#define EMDS_DEV_IRQ 29
```
omkar3141, line 24: I'm curious to know if this change works for all platforms and why did we choose 26 then?

ludvigsj: It works on nRF52840, nRF5340 and nRF54L15. I don't know where 26 came from. 29 was found by iterating through interrupt lines in the running sample and listing the free and implemented ones. (As far as I understand, the 26 came from a similar process)

ludvigsj: It was set to 26 here https://github.com/nrfconnect/sdk-nrf/commit/89b50afd9eefb4966e94f75c04ccd1098d421779

@m-alperen-sener : As far as I understand, it was a similar case of "find a different, available interrupt number" that I had to do here?

alxelax: How about nrf52832/nrf52833? They are in the list of supported platforms too

ludvigsj: In the sample? I don't think we claim the actual sample code to be a stable API?

m-alperen-sener: AFAIK we use the interrupt that is free to use; there is no background to that unfortunately.

ludvigsj: Now I have also tested the interrupt on nRF52832 and nRF52833, it works. Will remove the DNM now.

sigurdnev: Instead of hardcoding the interrupt number, you could use the defines from the MDK.
For nRF52840/nRF52832/nRF52833 29 is the PDM_IRQn.
For the nRF5340, 29 is EGU2_IRQn.
For nRF54L15 29 seems not to be a valid irq actually.
Maybe e.g. Software interrupt 1 would be better choice.

ludvigsj: @sigurdnev I was not aware of these defines from the MDK. I find `SWI01_IRQn = 29` in nrf54l15_application.h, so it seems to be SW interrupt 1 already? How do I use these MDK headers in the application? Do I need to conditionally include the correct one based on platform or is there a single header I can include that will magically include the correct defines for the current platform?

ludvigsj: I'll update this to not use hardcoded IRQn then. DNM again for now.

alxelax (approved):

sigurdnev (approved):

mia-ko (approved):

PR #15350, opened 2024-05-16
samples: net: Add native_sim to sample.yaml

jorgenmk:
Native_sim was missing from several sample.yaml files.

### DIFF
--- a/samples/net/aws_iot/sample.yaml
+++ b/samples/net/aws_iot/sample.yaml
@@ -11,6 +11,7 @@ tests:
       - nrf9151dk/nrf9151/ns
       - thingy91/nrf9160/ns
       - nrf7002dk/nrf5340/cpuapp/ns
+      - native_sim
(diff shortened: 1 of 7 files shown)


jtguggedal (approved):

simensrostad (approved):

lemrey (approved):
Unsure whether a changelog entry is needed for this.
Perhaps @divipillai knows whether we usually have entries for this?

divipillai:
> Unsure whether a changelog entry is needed for this. Perhaps @divipillai knows whether we usually have entries for this?

Since we have an entry for networking sample in general, we won't require a changelog entry here: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/releases_and_maturity/releases/release-notes-changelog.html#networking-samples
_Removed QEMU x86 emulation support and added support for the native simulator board._

PR #15352, opened 2024-05-16
applications: nrf_desktop: Cleanup USB state module

MarekPieta:
Change cleans up USB state application module implementation and aligns Kconfig configuration dependencies to allow integrating USB next stack. The module can only be built with USB next stack - the USB next API is not yet used.

Jira: NCSDK-27014

### DIFF
--- a/applications/nrf_desktop/src/modules/usb_state.c
+++ b/applications/nrf_desktop/src/modules/usb_state.c
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 #include <stdio.h>
+#include <stdint.h>
 
 #include <zephyr/types.h>
 #include <zephyr/sys/byteorder.h>
@@ -12,7 +13,6 @@
 #include <zephyr/usb/usb_device.h>
 #include <zephyr/usb/usb_ch9.h>
 #include <zephyr/usb/class/usb_hid.h>
-
 
 #define MODULE usb_state
 #include <caf/events/module_state_event.h>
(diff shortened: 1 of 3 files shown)


MarekPieta:
`CONFIG_USB_HID_DEVICE_COUNT' is still used in HID forward (that leads to build failures with USB next stack on dongles). I will add fix for it in a separate commit.

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -44,8 +40,10 @@ LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_USB_STATE_LOG_LEVEL);
 #define USB_SUBSCRIBER_PIPELINE_SIZE 0x01
 #define USB_SUBSCRIBER_REPORT_MAX    0x01
 
-#if DT_PROP(DT_NODELABEL(usbd), num_in_endpoints) < (CONFIG_USB_HID_DEVICE_COUNT + 1)
-  #error Too few USB IN Endpoints enabled. Modify dts.overlay file.
+/* Not used for USB next stack. */
+#ifdef CONFIG_DESKTOP_USB_NEXT
+	#define CONFIG_USB_HID_DEVICE_NAME ""
+	#define CONFIG_DESKTOP_USB_HID_PROTOCOL_CODE -1
 #endif
```
mkapala-nordic, line 47: is it necessary to overwrite those kconfigs?

mkapala-nordic (approved):

applications/nrf_desktop/src/modules/usb_state.c
MarekPieta: Yes, otherwise compilation fails because of an undefined symbol

kapi-no (approved):

PR #15354, opened 2024-05-16
lib: pdn: do not error in `pdn_dynamic_params_get()` when MTU is missing

lemrey:
The MTU parameter is optional in `+CGCONTRDP`.

Let `pdn_dynamic_params_get()` handle this scenario without raising an error.
Additionally, make all parameters optional, that is, passing `NULL` prevents the information to be copied in the parameters.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -762,6 +762,11 @@ Modem libraries
     * Convenience function to get :c:struct:`location_data_details` from the :c:struct:`location_event_data`.
     * Location data details for event :c:enum:`LOCATION_EVT_RESULT_UNKNOWN`.
 
+* :ref:`pdn_readme` library:
+
+  * Updated the ``dns4_pri``, ``dns4_sec``, and ``ipv4_mtu`` parameters of the :c:func:`pdn_dynamic_params_get` function to be optional.
+    If the MTU is not reported by the SIM card, the ``ipv4_mtu`` parameter is set to zero.
+
 * :ref:`lte_lc_readme` library:
 
   * Removed ``AT%XRAI`` related deprecated functions ``lte_lc_rai_param_set()`` and ``lte_lc_rai_req()``, and Kconfig option :kconfig:option:`CONFIG_LTE_RAI_REQ_VALUE`.
--- a/lib/pdn/pdn.c
+++ b/lib/pdn/pdn.c
@@ -606,8 +606,9 @@ int pdn_id_get(uint8_t cid)
 int pdn_dynamic_params_get(uint8_t cid, struct in_addr *dns4_pri,
 			   struct in_addr *dns4_sec, unsigned int *ipv4_mtu)
 {
-	int ret;
+	int matched;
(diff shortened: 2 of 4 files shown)


lib/pdn/pdn.c
```diff
@@ -619,15 +620,31 @@ int pdn_dynamic_params_get(uint8_t cid, struct in_addr *dns4_pri,
 	fmt = "+CGCONTRDP: %*u,,\"%*[^\"]\",\"\",\"\",\"%15[0-9.]\",\"%15[0-9.]\",,,,,%u";
 
 	/* If IPv4 is enabled, it will be the first response line. */
-	ret = nrf_modem_at_scanf(at_cmd, fmt, &dns4_pri_str, &dns4_sec_str, ipv4_mtu);
-	if (ret != 3) {
+	matched = nrf_modem_at_scanf(at_cmd, fmt, &dns4_pri_str, &dns4_sec_str, &mtu);
+	/* Need to match at least the two IP addresses, or there is an error */
+	if (matched < 2) {
```
tomi-font, line 625: If `matched == 2`, is it guaranteed that it is the two IP addresses that matched, and not the the first IP address + the MTU?

tomi-font (changes requested):
Can you also update SLM's use of this function [here](https://github.com/nrfconnect/sdk-nrf/blob/main/applications/serial_lte_modem/src/slm_ppp.c#L216) to not set the MTU (or to `sizeof([...])`, doesn't matter) if a MTU of 0 was returned?

lib/pdn/pdn.c
lemrey: yes :)

lemrey:
> Can you also update SLM's use of this function [here](https://github.com/nrfconnect/sdk-nrf/blob/main/applications/serial_lte_modem/src/slm_ppp.c#L216) to not set the MTU (or to `sizeof([...])`, doesn't matter) if a MTU of 0 was returned?

Updated

tomi-font (approved):

MarkusLassila (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -752,6 +752,11 @@ Modem libraries
     * Convenience function to get :c:struct:`location_data_details` from the :c:struct:`location_event_data`.
     * Location data details for event :c:enum:`LOCATION_EVT_RESULT_UNKNOWN`.
 
+* :ref:`pdn_readme`:
+
+  * The ``dns4_pri``, ``dns4_sec``, and ``ipv4_mtu`` parameters of the :c:func:`pdn_dynamic_params_get` function are now optional.
```
divipillai, line 757:
```suggestion
* :ref:`pdn_readme` library:

  * Updated the ``dns4_pri``, ``dns4_sec``, and ``ipv4_mtu`` parameters of the :c:func:`pdn_dynamic_params_get` function to be optional.
```

lemrey: Thank you, updated

divipillai (approved):

SeppoTakalo (approved):

PR #15366, opened 2024-05-17

anangl:
Pull in an upstream fix in the EXMIF dts node definition for nRF54H20.


anangl:
No longer needed.

PR #15381, opened 2024-05-20
test: bluetooth: iso: Update the project for testing with SDC

rick1082:
Update the project for using SDC for testing.
OCT-3011


tests/bluetooth/iso/modules/iso_broadcast_src.c
```diff
@@ -117,7 +117,7 @@ static void broadcaster_t(void *arg1, void *arg2, void *arg3)
 	while (1) {
 		int ret;
 
-		k_sleep(K_USEC(big_create_param.interval));
+		k_sleep(K_USEC(big_create_param.interval - 200));
```
alexsven, line 120: Seems random to reduce by 200. Add a comment?

tests/bluetooth/iso/child_image/hci_ipc.conf
alexsven, line 1: You might get push-back on this, since it probably needs to be using sysbuild and not child-image

tests/bluetooth/iso/modules/iso_broadcast_src.c
rick1082: Will add it, this part is very similar to here:
https://github.com/zephyrproject-rtos/zephyr/blob/b3ffe0aede6db181e118eb95db84ad7537bed0b2/samples/bluetooth/iso_broadcast_benchmark/src/broadcaster.c#L583

alexsven (approved):

koffes (approved):

PR #15384, opened 2024-05-20
Collab serial ble ot

canisLupus1313:


applications/protocols_serialization/client/snippets/debug/debug.conf
```diff
@@ -20,7 +20,8 @@ CONFIG_LOG_MODE_DEFERRED=y
 CONFIG_LOG_MAX_LEVEL=4
 
 CONFIG_LOG_BACKEND_RTT=n
-CONFIG_SHELL_LOG_BACKEND=y
+#CONFIG_SHELL_LOG_BACKEND=y
```
Damian-Nordic, line 23: Why this change? I guess if both logger and shell write to UART, we may expect some interwoven lines and that's why you typically use shell logging backend when shell is enabled.

subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -66,47 +66,62 @@ static void work_handler(struct k_work *work)
...
+		rx_len = ring_buf_put_claim(
+			&uart_tr->rx_ringbuf, &rx_buffer,
+			uart_tr->rx_ringbuf.size);
+		if (rx_len > 0) {
+			rx_len = uart_fifo_read(uart, rx_buffer, rx_len);
+			int err = ring_buf_put_finish(
+				&uart_tr->rx_ringbuf, rx_len);
+			if (err != 0) {
+				__ASSERT_NO_MSG(err == 0);
+			}
```
Damian-Nordic, line 107:
```suggestion
			__ASSERT_NO_MSG(err == 0);
```

subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -66,47 +66,62 @@ static void work_handler(struct k_work *work)
...
-				rx_len, ret);
-			break;
+			/* No space in the ring buffer - consume byte. */
+			LOG_WRN("RX ring buffer full.");
+
+			rx_len = uart_fifo_read(uart, &dummy, 1);
 		}
 
-		k_work_submit_to_queue(&uart_tr->rx_workq, &uart_tr->cb_work);
+		if (new_data) {
```
Damian-Nordic, line 122: nit: You might call this after the loop so that `k_work` is always scheduled only once per ISR.

Damian-Nordic (approved):
Looks great.

applications/protocols_serialization/client/snippets/debug/debug.conf
canisLupus1313: I can change it to a comment. But as it turns out when we are processing the shell command the logs aren't printed out since shell is busy. moving it to UART backend breaks formatting a bit but do not delay logging.

subsys/nrf_rpc/nrf_rpc_uart.c
canisLupus1313: fixed.

kszewenko (approved):

Damian-Nordic (approved):

PR #15391, opened 2024-05-20
samples: suit: add external flash configuration

e-rk:
The smp_transfer can now be build in external flash configuration. The external flash on nRF54H20dk is enabled and the SUIT DFU uses the external flash as a storage for the candidate images.


samples/suit/smp_transfer/boards/nrf54h20dk_nrf54h20_cpuapp.overlay
```diff
@@ -4,11 +4,14 @@
...
  */
 
+#include "../sysbuild/nrf54h20dk_nrf54h20_memory_map.dtsi"
+
 / {
 	chosen {
 		zephyr,entropy = &psa_rng;
 		zephyr,uart_mcumgr = &uart136;
 		zephyr,bt-hci-ipc = &ipc0;
+		extmem-device = &mx25uw63;
```
tomchy, line 14: I'm just wondering if the `extmem-device` should be enabled by default or only with the `FILE_PREFIX=extflash`?

samples/suit/smp_transfer/sample.yaml
```diff
@@ -59,3 +59,15 @@ tests:
...
+  subsys.suit.smp_transfer.full.extflash.v1:
+    extra_args: FILE_SUFFIX=extflash
+    extra_configs:
+      - CONFIG_SUIT_ENVELOPE_SEQUENCE_NUM=1
+    tags: suit
+
+  subsys.suit.smp_transfer.full.extflash.v2:
+    extra_args: FILE_SUFFIX=extflash
+    extra_configs:
+      - CONFIG_SUIT_ENVELOPE_SEQUENCE_NUM=2
```
tomchy, line 72: Don't you need to overwrite something to use the templates from `templates/extflash`?

e-rk: The template path is set in `sysbuild_extflash.conf`. The `FILE_SUFFIX` selects it.

samples/suit/smp_transfer/boards/nrf54h20dk_nrf54h20_cpuapp.overlay
e-rk: I would have to make a second copy of this file with the suffix. I could copy the common parts to some other file and `#include` in in both `nrf54h20dk_nrf54h20_cpuapp.overlay` and `nrf54h20dk_nrf54h20_cpuapp_extflash.overlay`.

adsz-nordic (approved):

samples/suit/smp_transfer/boards/nrf54h20dk_nrf54h20_cpuapp.overlay
tomchy: I think it is better to keep it in the current form - there is no nRF54H DK without extmem :slightly_smiling_face:

tomchy (approved):

samples/suit/smp_transfer/boards/nrf54h20dk_nrf54h20_cpuapp.overlay
e-rk: Whoops, I didn't notice it and split it after all.

samples/suit/smp_transfer/templates/extflash/rad_envelope.yaml.jinja2
```diff
@@ -0,0 +1,106 @@
+{%- set mpi_radio_vendor_name = application['config']['CONFIG_SUIT_MPI_RADio_LOCAL_1_VENDOR_NAME']|default('nordicsemi.com') %}
```
ahasztag, line 1: Aren't these two manifests (radio and root) just copies of suit-generator templates? Why keep them as separate files?

samples/suit/smp_transfer/boards/nrf54h20dk_nrf54h20_cpuapp_common.dtsi
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+ #include "../sysbuild/nrf54h20dk_nrf54h20_memory_map.dtsi"
+
+/ {
+	chosen {
+		zephyr,entropy = &psa_rng;
+		zephyr,uart_mcumgr = &uart136;
```
nordicjm, line 12: this is wrong, it's `uart-mcumgr` with a dash

nordicjm (changes requested):

samples/suit/smp_transfer/boards/nrf54h20dk_nrf54h20_cpuapp_common.dtsi
e-rk: Fixed

samples/suit/smp_transfer/templates/extflash/rad_envelope.yaml.jinja2
e-rk: Those are not exact copies. The manifest has fetch sequence.

nordicjm (approved):

PR #15393, opened 2024-05-20
samples: matter: add support for the TimeSync cluster in lock app

markaj-nordic:
* updated .zap file
* regenerated data model
* enabled ReadClient to support TimeSyncClient feature

- [x] Test with CHIP Tool


samples/matter/lock/prj.conf
```diff
@@ -42,3 +42,6 @@ CONFIG_CHIP_APP_LOG_LEVEL=3
 # Enable Factory Data feature
 CONFIG_CHIP_FACTORY_DATA=y
 CONFIG_CHIP_FACTORY_DATA_BUILD=y
+
+# Enable read client for needed for TimeSyncClient feature
+CONFIG_CHIP_ENABLE_READ_CLIENT=y
```
ArekBalysNordic, line 47: Is it for debugging purposes? I guess we shouldn't enable by default optional features.

ArekBalysNordic (commented):
If the cluster is "heavy" we can consider moving responsibility to enable it to the user and provide proper documentation.

samples/matter/lock/prj.conf
kkasperczyk-no: The TimeSync cluster is configured as a client, so I guess it cannot work without a read client functionality.

kkasperczyk-no: Well, after a second thought, do we really want to be a client and server? Wouldn't server role be enough for us?

markaj-nordic: To support all required Time Sync features I believe we need to be both server and client. The latter is needed for Time Sync Client feature (the device proactively queries trusted nodes about the current time after joining the fabric).

markaj-nordic:
This feature will be a part of https://github.com/nrfconnect/sdk-nrf/pull/15485. Closing this one.

PR #15402, opened 2024-05-21
Revert "samples: applications: Use ipc_radio as default radio firmware."

nordicjm:
This reverts commit e794abeaf0df459c9f2e662b49dafff13cb4026f. This commit does not add configuration properly and breaks builds


KAGA164 (changes requested):
Adding NACK for now to further investigate reason of reverting it. Preferred way would be to fix it if configuration is invalid somewhere

PR #15416, opened 2024-05-21
manifest: update zephyr - allow sysbuild for nrf54

nordic-piks:
Allow testing for PPR and FLPR core at nrf54


nordic-segl (approved):

nordic-segl:
Solution proposed by Maciej Perkowski was merged instead.

PR #15422, opened 2024-05-21
suit: recovery application build separately

ahasztag:
This PR adds the SUIT BLE recovery application. 

The application is build separately and than flashed after the main application is flashed.

test-sdk-nrf: sdk-nrf-pr-15422


cmake/sysbuild/suit.cmake
```diff
@@ -196,22 +196,30 @@ function(suit_create_package)
...
     if(NOT DEFINED ENVELOPE_JINJA_FILE)
       message(SEND_ERROR "DFU: Creation of SUIT artifacts failed.")
       return()
     endif()
     suit_check_template_digest(${INPUT_TEMPLATES_DIRECTORY} "${INPUT_ENVELOPE_JINJA_FILE}")
-    set(BINARY_FILE "${BINARY_FILE}.bin")
-
-    list(APPEND CORE_ARGS
-      --core ${target},${SUIT_ROOT_DIRECTORY}${target}.bin,${BINARY_DIR}/zephyr/edt.pickle,${BINARY_DIR}/zephyr/.config
-    )
```
e-rk, line 209: I think this change can be reverted if you use `CONFIG_SUIT_LOCAL_ENVELOPE_GENERATE`.

samples/suit/recovery/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+#
+
+cmake_minimum_required(VERSION 3.20.0)
+
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+if (NOT SYSBUILD)
+	message(WARNING
+		" This is a multi-image application that should be built using sysbuild.\n"
+		" Add --sysbuild argument to west build command to prepare all the images.")
+endif()
```
e-rk, line 14: It this warning needed? I think only the multicore/hello_world has something like this.

samples/suit/recovery/Kconfig
```diff
@@ -0,0 +1,7 @@
+# #
+# # Copyright (c) 2024 Nordic Semiconductor ASA
+# #
+# # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+# #
```
e-rk, line 5: Remove double `# #`

samples/suit/recovery/Kconfig.sysbuild
```diff
@@ -0,0 +1,7 @@
+# #
+# # Copyright (c) 2024 Nordic Semiconductor ASA
+# #
+# # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+# #
```
e-rk, line 5: Remove double `# #`

samples/suit/recovery/prj.conf
```diff
@@ -0,0 +1,101 @@
...
+CONFIG_MCUMGR=y
+CONFIG_MCUMGR_SMP_LEGACY_RC_BEHAVIOUR=y
+
+# Enable SUIT services
+CONFIG_SUIT=y
+CONFIG_MGMT_SUITFU=y
+CONFIG_FLASH=y
+
+##################
+#Bluetooth config
```
e-rk, line 26:
```suggestion
# Bluetooth config
```

samples/suit/recovery/prj.conf
```diff
@@ -0,0 +1,101 @@
+# #
+# # Copyright (c) 2024 Nordic Semiconductor ASA
+# #
+# # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+# #
```
e-rk, line 5: Remove double `# #`

samples/suit/recovery/prj.conf
```diff
@@ -0,0 +1,101 @@
...
+
+CONFIG_SUIT_MPI_GENERATE=n
+CONFIG_SUIT_ENVELOPE_OUTPUT_MPI_MERGE=n
+
+CONFIG_SUIT_ENVELOPE_TEMPLATE=""
```
e-rk, line 10: You can use `CONFIG_SUIT_LOCAL_ENVELOPE_GENERATE` instead.

samples/suit/recovery/src/main.c
```diff
@@ -0,0 +1,25 @@
...
+
+int main(void)
+{
+	#ifdef CONFIG_MCUMGR_TRANSPORT_BT
+	start_smp_bluetooth_adverts();
+	#endif
+
+	while (1) {
+		k_msleep(1000);
+	}
```
e-rk, line 22: This can be removed.

samples/suit/recovery/src/main.c
```diff
@@ -0,0 +1,25 @@
...
+#include <zephyr/logging/log.h>
+#include "common.h"
+
+LOG_MODULE_REGISTER(app);
+
+int main(void)
+{
+	#ifdef CONFIG_MCUMGR_TRANSPORT_BT
+	start_smp_bluetooth_adverts();
+	#endif
```
e-rk, line 18:
```suggestion
	if (IS_ENABLED(CONFIG_MCUMGR_TRANSPORT_BT)) {
	    start_smp_bluetooth_adverts();
	}
```

samples/suit/smp_transfer/prj.conf
```diff
@@ -28,3 +28,6 @@ CONFIG_FLASH=y
 
 # Extended SUIT commands over USER group
 CONFIG_MGMT_SUITFU_GRP_SUIT=y
+
+CONFIG_SUIT_MPI_APP_RECOVERY=y # TODO: do this automatically?
+CONFIG_SUIT_MPI_RAD_RECOVERY=y # TODO: do this automatically?
```
e-rk, line 33: You can put this into `sysbuild/smp_transfer_recovery.conf` for the time being. You will then be able to enable this configuration using `FILE_SUFFIX=recovery`

samples/suit/smp_transfer/rad_envelope.yaml.jinja2.digest
```diff
@@ -0,0 +1 @@
+9d46b70df831d38071543b56193edaa269dfadbab4da8f384459a0a141a834a4
```
e-rk, line 1: If you need to put a dedicated template I would suggest putting them into `templates/recovery` without the digests.

e-rk (changes requested):

cmake/sysbuild/suit.cmake
ahasztag: Introduced changes here, as we've discussed

samples/suit/recovery/CMakeLists.txt
ahasztag: Removed

samples/suit/recovery/Kconfig
ahasztag: Done

samples/suit/recovery/Kconfig.sysbuild
ahasztag: Done

samples/suit/recovery/prj.conf
ahasztag: Done

samples/suit/recovery/prj.conf
ahasztag: Changed

samples/suit/recovery/prj.conf
ahasztag: Fixed

samples/suit/recovery/src/main.c
ahasztag: Why is this solution better?

samples/suit/smp_transfer/rad_envelope.yaml.jinja2.digest
ahasztag: This was of course committed here by accident, removed.

samples/suit/smp_transfer/prj.conf
ahasztag: I am not sure this would work, as we already need `FILE_SUFFIX=bt` to build with BLE, which would have to be used at the same time.

I actually consider generating MPI-s even if no recovery application is present harmless.

cmake/sysbuild/suit.cmake
```diff
@@ -176,13 +176,19 @@ function(suit_create_package)
...
 
     set(BINARY_FILE "${BINARY_FILE}.bin")
 
     list(APPEND CORE_ARGS
       --core ${image},${SUIT_ROOT_DIRECTORY}${image}.bin,${BINARY_DIR}/zephyr/edt.pickle,${BINARY_DIR}/zephyr/.config
     )
+
+    list(APPEND CORE_ARGS
+    --core ${target},${SUIT_ROOT_DIRECTORY}${target}.bin,${BINARY_DIR}/zephyr/edt.pickle,${BINARY_DIR}/zephyr/.config
+    )
```
e-rk, line 189: I have one concern here. in `flash_companion` I didn't set any value for `CONFIG_SUIT_ENVELOPE_TARGET`, which means that it has `application` by default. In that case both `smp_transfer` and `flash_companion` would have the same target name, which is a recipe for problems.

I am wondering how `CONFIG_SUIT_ENVELOPE_TARGET` should be approached. On one hand, I think that using `sysbuild` image name by default is convenient and would be easier for the user.

On the second hand, we want the default manifest templates to be generic regardless of the application name. 

Perhaps in the `flash_companion` the `CONFIG_SUIT_ENVELOPE_TARGET` should be set to `"" (empty)` and in this loop there should be a check if there are multiple images with the same `CONFIG_SUIT_ENVELOPE_TARGET` and a warning should be displayed in such case.

e-rk (changes requested):

samples/suit/recovery/Kconfig
```diff
@@ -0,0 +1,7 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+source "Kconfig.zephyr"
```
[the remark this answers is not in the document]
ahasztag: It only seems so, without this I get hundreds of Kconfig errors. I've also taken a look at other samples and each one had this `source "Kconfig.zephyr"` line in its Kconfig file.

samples/suit/recovery/src/main.c
ahasztag: Ok, changed also in smp_transfer

samples/suit/smp_transfer/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 
 project(suit_smp_transfer)
 
+
```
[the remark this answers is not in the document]
ahasztag: Done

samples/suit/smp_transfer/prj.conf
ahasztag: Added `default y` for these Kconfig options and removed them from here.

adsz-nordic (approved):

samples/suit/smp_transfer/root_with_binary_nordic_top.yaml.jinja2
```diff
@@ -0,0 +1,198 @@
+{%- set component_index = 0 %}
```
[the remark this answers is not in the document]
ahasztag: Committed by accident, removed

sysbuild/Kconfig.suit
```diff
@@ -24,9 +24,13 @@ config SUIT_ENVELOPE_SEQUENCE_NUM
...
 	default 1
 
+config SUIT_RECOVERY
+	bool "The application is a SUIT recovery firmware"
+
 config SUIT_ENVELOPE_ROOT_TEMPLATE
 	string "Path to the default root envelope template"
-	default "${ZEPHYR_SUIT_GENERATOR_MODULE_DIR}/ncs/root_with_binary_nordic_top.yaml.jinja2"
+	default "${ZEPHYR_SUIT_GENERATOR_MODULE_DIR}/ncs/root_with_binary_nordic_top.yaml.jinja2" if !SUIT_RECOVERY
+	default "${ZEPHYR_SUIT_GENERATOR_MODULE_DIR}/ncs/app_recovery_envelope.yaml.jinja2" if SUIT_RECOVERY
```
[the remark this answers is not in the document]
ahasztag: As discussed I'm trying another way, removed.

cmake/sysbuild/suit.cmake
```diff
@@ -176,13 +176,19 @@ function(suit_create_package)
   foreach(image ${IMAGES})
     sysbuild_get(BINARY_DIR IMAGE ${image} VAR APPLICATION_BINARY_DIR CACHE)
     sysbuild_get(BINARY_FILE IMAGE ${image} VAR CONFIG_KERNEL_BIN_NAME KCONFIG)
+    sysbuild_get(target IMAGE ${image} VAR CONFIG_SUIT_ENVELOPE_TARGET KCONFIG)
```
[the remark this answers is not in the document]
ahasztag: 'target' is for example "radio" and "application" (instead of "image": where we have for example "hci_ipc"). Manifests use "radio" so if we want this tp be accessible to all manifests we need to put this line here.

samples/suit/recovery/sysbuild.conf
```diff
@@ -0,0 +1,13 @@
+# #
+# # Copyright (c) 2024 Nordic Semiconductor ASA
```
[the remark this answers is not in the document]
ahasztag: Fixed

samples/suit/smp_transfer/sysbuild/hci_ipc.overlay
```diff
@@ -7,3 +7,23 @@
...
+		perm-execute;
+		perm-secure;
+		#address-cells = < 0x1 >;
+		#size-cells = < 0x1 >;
+		cpurad_slot0_partition: partition@66000 {
+			reg = < 0x66000 DT_SIZE_K(256) >;
+		};
+
+		recovery_partition: partition@a6000 {
+			reg = < 0xa6000 DT_SIZE_K(100) >;
```
tomchy, line 26: The recovery partition size does not match?

tomchy: Discussed offline: this is the `cpurad` recovery partition and the new name makes it obvious.

tomchy (approved):

cmake/sysbuild/suit.cmake
```diff
@@ -176,12 +176,20 @@ function(suit_create_package)
...
```
e-rk, line 179: It is necessary to add `unset` here, otherwise the parameters to suit-generator are wrong.

e-rk (approved):

e-rk:
> If it is an Application, it needs to be moved to Application folder.

The situation is currently not clear. It would better fit the core_fw category and the current recommendation is to put such applications in the sample directory. See [NCSDK-10551](https://nordicsemi.atlassian.net/browse/NCSDK-10551)

ahasztag:
> Some initial comments:
> 
> * If it is an Application, it needs to be moved to Application folder.
> * It need to follow the Application/Sample templates that can be found https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/dev_model_and_contributions/documentation/templates.html.
> * Since it is a new doc, it needs to be added in the Changelog.

The documentation will be extended and Changelog will be added in a separate PR. 

When it comes to the "application" part - this will be finally placed in a completely different category - named something like "core firmware" together with some other applications. It is just temporarily kept in a sample - this was consulted with @wbober.

richabp:
> > Some initial comments:
> > 
> > * If it is an Application, it needs to be moved to Application folder.
> > * It need to follow the Application/Sample templates that can be found https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/dev_model_and_contributions/documentation/templates.html.
> > * Since it is a new doc, it needs to be added in the Changelog.
> 
> The documentation will be extended and Changelog will be added in a separate PR.
> 
> When it comes to the "application" part - this will be finally placed in a completely different category - named something like "core firmware" together with some other applications. It is just temporarily kept in a sample - this was consulted with @wbober.

Ok. Approving, assuming all the doc changes will be taken care of in another PR.

richabp (approved):
Approving. All the required doc changes will be done in another PR.

samples/suit/recovery/README.rst
```diff
@@ -0,0 +1,50 @@
+:orphan:
```
greg-fer, line 1: This is orphaned, but still appearing in the docs because of the inclusion at the higher level: 
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/suit.html

kittydepa: Yes, this is/will be addressed in a different PR that is currently open, but thanks for catching!

PR #15427, opened 2024-05-21
samples: wifi: wfa_qt_app: Fix SVD failures

krish2718:
Hostap now uses more stack, so, calibrate and increase stack sizes.

Fixes SHEL-2754.


SachinDKulkarni (approved):

sachinthegreen (approved):

ioannisg (approved):

ioannisg:
Side comment @krish2718  and @sachinthegreen : it would be nice if we could document (e.g. as part of the architecture documentation, SDK docs, etc.) which core threads are expected to invoke WPA supplicant calls, and hence are sensitive to the stack requirements of the hostap. This is very useful for application developers, so they can do proper selection of stack sizes.

D-Triveni (approved):

bama-nordic (approved):

PR #15440, opened 2024-05-22
suit: Add SDFW Recovery update

adsz-nordic:
Add new functionality of sdfw_recovery_sink and use it for performing SDFW Recovery update.

Ref: NCSDK-26627


tomchy (approved):

subsys/suit/platform/sink_selector/src/suit_sdfw_sink_selector.c
```diff
@@ -162,6 +165,13 @@ int suit_sink_select(suit_component_t dst_handle, struct stream_sink *sink)
 			}
 
 			LOG_ERR("SDFW sink not enabled");
+		} else if (number == 2) {
```
ahasztag, line 168: I think the magic numbers 1 and 2 should be defined as macros somewhere (`suit_metadata.h` seems to be a good place)

subsys/suit/stream/stream_sinks/src/suit_sdfw_recovery_sink.c
```diff
@@ -0,0 +1,366 @@
...
+			LOG_ERR("Failed to calculate digest: %d", err);
+		}
+
+		return SUIT_PLAT_ERR_CRASH;
+	}
+
+	LOG_DBG("Candidate consistent");
+
+	/* Then compare candidate's digest with current SDFW Recovery digest */
+	err = verify_digest(candidate_binary_start,
```
ahasztag, line 160: Can't we use `memcmp` between `current_sdfw_recovery_digest` and `candidate_digest_in_manifest`  - as it is currently done in SDFW sink ? Comparing it to the whole digest seems a waste of time here.

Same for `check_recovery_activated`.

subsys/suit/stream/stream_sinks/src/suit_sdfw_recovery_sink.c
```diff
@@ -0,0 +1,366 @@
...
+
+		LOG_PANIC();
+
+		sys_reboot(SYS_REBOOT_COLD);
+	} else {
+		LOG_DBG("Reboot disabled - perform manually");
+	}
+}
+
+static suit_plat_err_t check_urot_none(const uint8_t *buf, size_t size)
```
ahasztag, line 193: Is `urot` the correct name in case of recovery?

subsys/suit/stream/stream_sinks/src/suit_sdfw_recovery_sink.c
e-rk: I think it's better to keep this implementation of checking the digest and optimize later if the performance ever turns out to be a problem.

robertstypa (approved):

e-rk (approved):

subsys/suit/stream/stream_sinks/src/suit_sdfw_recovery_sink.c
adsz-nordic: Given our deadlines, I would prefer to leave current implementation as it is and optimize it in future a e-rk suggests.

subsys/suit/stream/stream_sinks/src/suit_sdfw_recovery_sink.c
adsz-nordic: Yes, this is a name of the register, which is used for both sdfw and sdfw recovery updates.

PR #15448, opened 2024-05-22
emds: add preliminary timings for rram

alxelax:
Commit adds preliminary writing timings for flash driver based on rram.

### DIFF
--- a/subsys/emds/Kconfig
+++ b/subsys/emds/Kconfig
@@ -37,11 +37,15 @@ config EMDS_THREAD_PRIORITY
 
 config EMDS_FLASH_TIME_WRITE_ONE_WORD_US
 	int "Time to write one word into flash"
+	default 210 if SOC_FLASH_NRF_RRAM
 	default 41
 	help
 	  Max time to write one word (4 bytes) in flash (in microseconds). This
 	  value is dependent on the chip used, and should be checked against the
 	  chip datasheet.
+	  For RRAM-based flash drivers, data is written in blocks of 32 bytes over
+	  buffered writing. This value is used to calculate the time to write
+	  one 32-byte block.
 
 config EMDS_FLASH_TIME_ENTRY_OVERHEAD_US
 	int "Time to schedule write of one entry"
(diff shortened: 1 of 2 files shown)


subsys/emds/Kconfig
```diff
@@ -37,11 +37,15 @@ config EMDS_THREAD_PRIORITY
...
 	int "Time to write one word into flash"
+	default 210 if SOC_FLASH_NRF_RRAM
 	default 41
 	help
 	  Max time to write one word (4 bytes) in flash (in microseconds). This
 	  value is dependent on the chip used, and should be checked against the
 	  chip datasheet.
+	  For rram based flash driver, data is written by blocks of 32 bytes over
+	  buffered writing. This value is used to calculate the time to write
+	  one 32 bytes block.
```
mia-ko, line 48:
```suggestion
	  For RRAM-based flash drivers, data is written in blocks of 32 bytes over
	  buffered writing. This value is used to calculate the time to write
	  one 32-byte block.
```

subsys/emds/Kconfig
```diff
@@ -63,6 +68,8 @@ config EMDS_FLASH_TIME_BASE_OVERHEAD_US
 	   worst case scenario, before starting storing entries. This value
 	   is dependent on the chip used, and should be checked against the chip
 	   datasheet.
+	   For rram based flash driver, the time is defined by the erasing block size.
+	   The default erase block size is 4096 bytes.
```
mia-ko, line 72:
```suggestion
	   For RRAM-based flash drivers, the time is defined by the erasing block size.
	   The default erasing block size is 4096 bytes.
```

mia-ko (changes requested):
Do we need a changelog entry for this?

Andrewpini (approved):

alxelax:
> Do we need a changelog entry for this?

No we don't. We didn't release anything related to nrf54l. Nothing to change.

mia-ko (approved):

PR #15450, opened 2024-05-22
net: lib: nrf_cloud: properly update last job string

jayteemo:
Update the last job string when job validation is done. 
IRIS-8846

### DIFF
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_fota.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_fota.c
@@ -11,6 +11,7 @@
 #include "nrf_cloud_download.h"
 
 #include <zephyr/kernel.h>
+#include <zephyr/spinlock.h>
 #include <errno.h>
 #include <stdio.h>
 #include <fcntl.h>
@@ -90,6 +91,7 @@ static struct nrf_cloud_fota_job current_fota = {
 	.info = { .type = NRF_CLOUD_FOTA_TYPE__INVALID }
 };
 
+static struct k_spinlock last_job_slock;
 static uint8_t last_job[NRF_CLOUD_FOTA_JOB_ID_SIZE];
 static struct nrf_cloud_settings_fota_job saved_job = {
 	.type = NRF_CLOUD_FOTA_TYPE__INVALID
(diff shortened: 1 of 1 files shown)


plskeggs (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota.c
```diff
@@ -104,6 +106,31 @@ static bool reboot_on_init;
...
+
+	LOG_DBG("Updated last job: %s", job_id);
+}
+
+static bool is_last_job(char const *const job_id)
+{
+	bool match;
+
+	K_SPINLOCK(&last_job_slock) {
+		match = (strncmp(last_job, job_id, sizeof(last_job)) == 0);
```
glarsennordic, line 128: This will return false if a job_id which was longer than `NRF_CLOUD_FOTA_JOB_ID_SIZE` was stored, even if the same job_id was called

I don't have a problem with that, just want to make sure its the intended behavior

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota.c
```diff
@@ -104,6 +106,31 @@ static bool reboot_on_init;
...
 static bool fota_report_ack_pending;
 
+static void update_last_job(char const *const job_id)
+{
+	if ((job_id == NULL) || (job_id[0] == '\0')) {
+		return;
+	}
+
+	K_SPINLOCK(&last_job_slock) {
+		strncpy(last_job, job_id, sizeof(last_job) - 1);
```
glarsennordic, line 116: Maybe we should print a warning if the job ID is too large?

Or do you think that is too unlikely to happen to justify the memory cost?

glarsennordic (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_fota.c
jayteemo: job id should have already been checked for length.

tony-le-24 (commented):
LGTM! The issue seems to be gone in my local testing.

tony-le-24 (approved):

PR #15455, opened 2024-05-23
samples: matter: Refactored diagnostic crash logs module

kkasperczyk-no:
Refactored and enhanced diagnostic crash logs module:
* Extracted crash logs specific methods from generic diagnostic logs provider to the diagnostic logs crash module.
* Removed usage of persistent storage as a back-up for the retention RAM memory
* Added saving intent implementation object as a part of mIntentMap what allows to perform several independent log collection sessions at the same time.


samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
```diff
@@ -22,6 +22,11 @@ namespace Nrf::Matter
 
 class DiagnosticLogProvider : public chip::app::Clusters::DiagnosticLogs::DiagnosticLogsProviderDelegate {
 public:
+	struct IntentData {
+		chip::app::Clusters::DiagnosticLogs::IntentEnum mIntent;
+		void *mIntentImpl = nullptr;
```
ArekBalysNordic, line 27: I guess it is better to use here an abstract class and then define 3 intent implementations based on that rather than using `void *`. `Void*` can suggest that it can be used for different purposes, whereas you need to use a specific API inside DiagnosticLogProvider methods.

samples/matter/common/src/diagnostic/diagnostic_logs_crash.h
```diff
@@ -81,6 +86,11 @@ class CrashData {
 	 */
 	size_t CalculateSize();
 
+	CHIP_ERROR GetCrashLogs(chip::MutableByteSpan &outBuffer, bool &outIsEndOfLog);
+	CHIP_ERROR FinishCrashLogs();
+	size_t GetCrashLogsSize();
+	CHIP_ERROR MoveCrashLogsToNVS();
+
```
ArekBalysNordic, line 93: They are public methods so maybe it is worth adding descriptions to them.

samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
```diff
@@ -77,30 +79,11 @@ class DiagnosticLogProvider : public chip::app::Clusters::DiagnosticLogs::Diagno
...
-	Nrf::CrashData *mCrashData = nullptr;
-#endif
-
-#ifndef CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_SAVE_CRASH_TO_SETTINGS
 	DiagnosticLogProvider() = default;
-#endif
+
+	CHIP_ERROR SetIntentImplementation(chip::app::Clusters::DiagnosticLogs::IntentEnum intent,
+					   IntentData &intentData);
+	void DeleteIntentImplementation(chip::app::Clusters::DiagnosticLogs::IntentEnum intent, IntentData &intentData);
```
ArekBalysNordic, line 86: You can make them private, or add a description if you want to keep them public.

samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
kkasperczyk-no: Done, I've implemented abstract DiagnosticLogsIntentImpl class. Please re-review.

samples/matter/common/src/diagnostic/diagnostic_logs_crash.h
kkasperczyk-no: Done

samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
kkasperczyk-no: Done

ArekBalysNordic (approved):
LGTM

samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
```diff
@@ -20,8 +14,38 @@
...
+
+	/**
+	 * @brief Finish the log capturing. This method is used to perform potential clean up after session.
+	 */
+	virtual CHIP_ERROR FinishLogs() = 0;
+};
+
 class DiagnosticLogProvider : public chip::app::Clusters::DiagnosticLogs::DiagnosticLogsProviderDelegate {
 public:
+	struct IntentData {
```
markaj-nordic, line 44: Theoretically the `IntentData` does not match FiniteMap requirements: `* T2 must have move semantics and bool()/==operators implemented`. Also, if we implement the proper destructor here, we can do without `DeleteIntentImplementation` because the `FiniteMap::Erase()` will do the job then. You can use the following as a reference:
https://github.com/nrfconnect/sdk-nrf/blob/main/samples/matter/common/src/persistent_storage/backends/persistent_storage_secure.h#L36

samples/matter/common/src/diagnostic/Kconfig
```diff
@@ -6,7 +6,7 @@
 
 menuconfig NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS
 	bool "Enable diagnostics logs support"
-	select CHIP_ENABLE_BDX_LOG_TRANSFER
+	imply CHIP_ENABLE_BDX_LOG_TRANSFER
```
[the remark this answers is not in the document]
kkasperczyk-no: Yes, we can do without BDX, so I thought that select is too strict. If the device does not support BDX and the size is too big we will just return the max available size using the Matter command. The rest of data will be cut.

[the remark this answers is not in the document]
markaj-nordic: Ok, fair enough.

samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
```diff
@@ -20,8 +14,38 @@
 namespace Nrf::Matter
 {
 
+class DiagnosticLogsIntentImpl {
```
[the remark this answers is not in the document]
kkasperczyk-no: Yeah I think that `Impl` name is a little bit misleading. I didn't mean to refer to the polymorphism architecture we use in Matter. I just wanted to create a class that will define API to be implemented by specific intents implementation. I will rename it.

samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
kkasperczyk-no: Done

wiba-nordic (approved):

markaj-nordic (approved):
Looks great now.

samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
```diff
@@ -20,6 +14,74 @@
...
+			Clear();
+			mIntent = other.mIntent;
+			mIntentImpl = other.mIntentImpl;
+			/* Nullptr the copied data to prevent from deleting it by the source. */
+			other.mIntentImpl = nullptr;
+		}
+		return *this;
+	}
+
+	operator bool() const { return true; }
```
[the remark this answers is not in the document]
kkasperczyk-no: For now mIntentImpl can be nullptr. This is needed,because we don't have network logs implementation and we want to use it for testing purposes.

samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
```diff
@@ -20,6 +14,74 @@
...
+	}
+
+	IntentData &operator=(IntentData &&other)
+	{
+		if (this != &other) {
+			Clear();
+			mIntent = other.mIntent;
+			mIntentImpl = other.mIntentImpl;
+			/* Nullptr the copied data to prevent from deleting it by the source. */
+			other.mIntentImpl = nullptr;
```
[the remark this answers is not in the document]
kkasperczyk-no: I will add it in the next PR. I will open it just after merging this one.

PR #15459, opened 2024-05-23
sysbuild: Fix issue with re-generating provision data

nordicjm:
Fixes an issue with sysbuild whereby settings were changed and the provision data was not re-generated, which would then cause an unbootable image

### DIFF
--- a/cmake/sysbuild/provision_hex.cmake
+++ b/cmake/sysbuild/provision_hex.cmake
@@ -47,7 +47,7 @@ function(provision application prefix_name)
         --public-key-files "${PUBLIC_KEY_FILES}"
       )
 
-      set(PROVISION_DEPENDS signature_public_key_file_target ${application}_extra_byproducts ${application}/zephyr/.config)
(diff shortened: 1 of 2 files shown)


cmake/sysbuild/provision_hex.cmake
```diff
@@ -125,6 +125,7 @@ function(provision application prefix_name)
 
   add_custom_target(
     ${prefix_name}provision_target
+    ALL
```
tejlmand, line 128: why is `ALL` needed here ?

The merged_hex target already depends on `merged.hex` file creation which again depends on the `<name>_provision.hex` file creation.

https://github.com/nrfconnect/sdk-nrf/blob/41c2ad2b39a8408059432f0a54495763f7db9723/cmake/sysbuild/partition_manager.cmake#L272-L282
https://github.com/nrfconnect/sdk-nrf/blob/41c2ad2b39a8408059432f0a54495763f7db9723/cmake/sysbuild/partition_manager.cmake#L253

nordicjm: Fixed

tejlmand (approved):

nordicjm:
@sigvartmh @nvlsianpu please review

PR #15461, opened 2024-05-23
samples: wifi: Enable SHA1

Vge0rge:
Since it is required by WIFI.

Ref: SHEL-2566

### DIFF
--- a/samples/wifi/shell/overlay-openthread.conf
+++ b/samples/wifi/shell/overlay-openthread.conf
@@ -31,3 +31,6 @@ CONFIG_LOG_MAX_LEVEL=1
 
 # Consumes more memory
 CONFIG_WIFI_CREDENTIALS=n
+
+# Required by WIFI
+CONFIG_MBEDTLS_SHA1_C=y


samples/wifi/shell/prj.conf
```diff
@@ -93,3 +93,5 @@ CONFIG_SETTINGS_NVS=y
 # environment.
 CONFIG_NET_MGMT_EVENT_QUEUE_TIMEOUT=5000
 CONFIG_NET_SOCKETS_POLL_MAX=10
+
+CONFIG_MBEDTLS_SHA1_C=y
```
krish2718, line 97: This needs to done at `modules/hostap/Kconfig` so that it applies to all users of Wi-Fi. And also why is this only a problem with OT and not others? By default without OT Wi-Fi works fine.

krish2718: Or as its only for OT, we should move this to `overaly-openthread.conf`

Vge0rge: Sure, I moved it to the overlay of openthread.

krish2718 (approved):

krish2718:
@muraliThokala  please try OT Co-ex with this fix.

sachinthegreen (approved):

PR #15418, opened 2024-05-21
Enable ITS encryption by default on 54L + other TF-M updates

mswarowsky:

PR #15466, opened 2024-05-23
manifest: update TF-M

nordic-mare:
Update TF-M


Vge0rge (approved):

mswarowsky:
CI is green so we include it in #15418, to reduce the amounts of manifest PR's

PR #15491, opened 2024-05-24
sysbuild: MCUboot fixes

nordicjm:


carlescufi:
@ArekBalysNordic @sigvartmh @nvlsianpu @de-nordic please review ASAP

sysbuild/Kconfig.mcuboot
```diff
@@ -143,6 +143,13 @@ config MCUBOOT_UPDATEABLE_IMAGES
 	help
 	  The number of images that MCUboot will be built with.
 
+config MCUBOOT_APP_SYNC_UPDATEABLE_IMAGES
+	bool "Sync updateable image value to main application"
+	default y
+	help
+	  If enabled then will synchronise the value of ``MCUBOOT_UPDATEABLE_IMAGES`` to MCUboot
```
de-nordic, line 150: You mean the `UPDATEABLE_IMAGE_NUMBER` or the MCUmgr Kconfig `MCUMGR_GRP_IMG_UPDATABLE_IMAGE_NUMBER`?
(The `MCUBOOT_UPDATEABLE_IMAGES` is something new? I can not find it...)

nordicjm: `MCUBOOT_UPDATEABLE_IMAGES` is the sysbuild setting, if this is set then it will set it to both mcuboot and application, if not then it will only set it to mcuboot... though not sure why the help text mentions mcuboot

de-nordic (approved):

cmake/sysbuild/b0_mcuboot_signing.cmake
```diff
@@ -69,7 +69,7 @@ function(ncs_secure_boot_mcuboot_sign application bin_files signed_targets prefi
         # Hence, if a programmer is given this hex file, it will flash it
         # to the secondary slot, and upon reboot mcuboot will swap in the
         # contents of the hex file.
-        ${imgtool_sign} ${PROJECT_BINARY_DIR}/signed_by_b0_${application}.bin ${output}.bin
+        ${imgtool_sign} ${imgtool_extra} ${PROJECT_BINARY_DIR}/signed_by_b0_${application}.bin ${output}.bin
```
tejlmand, line 72: the use of `imgtool_extra` is actually problematic because it will inherit `imgtool_extra` from parent scope value which can result in surprises.

Use of imgtool_extra should be an argument to the function here:
https://github.com/nrfconnect/sdk-nrf/blob/f30dcc20c0860d962676b0f54f9b7c27816c2ddd/cmake/sysbuild/b0_mcuboot_signing.cmake#L13

tejlmand (changes requested):

cmake/sysbuild/b0_mcuboot_signing.cmake
nordicjm: it won't inherit anything because this is from a sysbuild context, there is no imgtool_extra variable in sysbuild

sysbuild/Kconfig.mcuboot
nordicjm: Have re-read the text and main application is on the line after, so is correct

ArekBalysNordic:
> @ArekBalysNordic @sigvartmh @nvlsianpu @de-nordic please review ASAP

@doublemis1 Could you please check whether our DFU tests work with this PR?

PR #15496, opened 2024-05-24
Add NB-IoT support for HERE positioning

JyriLehtinen:
HERE Api v2 has added separate objects for LTE, LTE-M and NB-IoT.
[https://www.here.com/docs/bundle/network-positioning-api-developer-guide-v2/page/topics/example-nbiot.html]

Based on my tests with NB-IoT cell measurements from Finland and Germany (Elisa and Vodafone networks), the "lte" object does not work while "lteCatM" supports both LTE-M and NB-IoT cells.

Thus, the current implementation does not strictly follow the HERE Api v2 documentation, and while this PR doesn't either, it supports NB-IoT.

Better solution would be to format the payload based the current network mode, but  lte_lc_cells_info -struct does not include that information.

### DIFF
--- a/lib/location/cloud_service/cloud_service_here_rest.c
+++ b/lib/location/cloud_service/cloud_service_here_rest.c
@@ -52,7 +52,7 @@ BUILD_ASSERT(sizeof(CONFIG_LOCATION_SERVICE_HERE_API_KEY) > 1,
 
 #define HTTP_REQUEST_BODY						\
 	"{"								\
-		"\"lte\":["						\
+		"\"lteCatM\":["						\
 			"{"						\
 				"\"mcc\":%d,"				\
 				"\"mnc\":%d,"				\
@@ -70,7 +70,7 @@ BUILD_ASSERT(sizeof(CONFIG_LOCATION_SERVICE_HERE_API_KEY) > 1,
 
 #define HTTP_REQUEST_BODY_NO_NEIGHBORS					\
 	"{"								\
-		"\"lte\":["						\
+		"\"lteCatM\":["						\
 			"{"						\
 				"\"mcc\":%d,"				\
 				"\"mnc\":%d,"				\


lib/location/cloud_service/cloud_service_here_rest.c
```diff
@@ -52,7 +52,7 @@ BUILD_ASSERT(sizeof(CONFIG_LOCATION_SERVICE_HERE_API_KEY) > 1,
 
 #define HTTP_REQUEST_BODY						\
 	"{"								\
-		"\"lte\":["						\
+		"\"lteCatM\":["						\
```
trantanen, line 55: I'm not sure we can do this. As you've said, it's not done according to specs.

> Better solution would be to format the payload based the current network mode, but lte_lc_cells_info -struct does not include that information.

You could add it there if you wanted to. LTE LC already has it internally in the `prev_lte_mode` variable.

We had thought of this earlier but skipped as the advantages hadn't been confirmed with the cloud services.

JyriLehtinen: Right, that should be a rather small change.
But I assume it still can't be merged as it's not following specs and the benefits are quite minimal?

trantanen: I'm not sure what you are referring to now with "it still can't be merged".
Adding network mode into `lte_lc_cells_info` structure is fine and its use in this file.

tokangas: Sending cells according to the current system in either `lteCatM` or `nbIot` object would be fine and according to the specs, right? However, it seems that some of the objects are named differently for LTE(-M) and NB-IoT, for example `rsrp` and `nrsrp`, so this needs to be taken into account in the implementation.

JyriLehtinen: Yes, that should be correct. Though I wonder when they changed the API, and what is the purpose of having the 'n' in `nrsrp`. On a side note, the example in the API documentation uses same cell data for LTE-M and NB-IoT, which I assume are not real measurements for NB-IoT.

tokangas: Yes, the examples seem to contain identical data, the object names are just different. I don't know why they have decided to use different naming, it seems unnecessary and only complicates the code.

JyriLehtinen: I noticed LWM2M object /4/0/2 specifies NRSRP for NB-IoT signal strength, and the referred 3GPP TS 36.133 uses NRSRP for NB-IoT, so it seems it's best to stick with that.

PR #15506, opened 2024-05-27
suit: Introduce invoke recovery failed state

tomchy:
It is crucial to detect the situation in which the recovery path fails, so the access to the SUIT storage may be unblocked to recover the device.

Ref: NCSDK-NONE


tests/subsys/suit/execution_mode/src/main.c
```diff
@@ -25,7 +25,7 @@ ZTEST(suit_execution_mode_tests, test_suit_execution_mode_set_NOK)
 	zassert_equal(err, SUIT_PLAT_ERR_INVAL,
 		      "suit_execution_mode_set should have failed - invalid input");
 
-	err = suit_execution_mode_set(EXECUTION_MODE_FAIL_MPI_UNSUPPORTED + 1);
+	err = suit_execution_mode_set(EXECUTION_MODE_FAIL_INVOKE_RECOVERY + 1);
```
e-rk, line 28: nit: Maybe you could add `EXECUTION_MODE_INVALID` as the last item and modify the condition in `suit_execution_mode_set` to ensure that the next person adding new items doesn't have to remember to do such change.

adsz-nordic (approved):

e-rk (approved):

tests/subsys/suit/execution_mode/src/main.c
tomchy: If I correctly remember, there were some discussions about it and it was decided to not include such non-existing (invalid) value.

PR #15521, opened 2024-05-27
lib: location: Send GNSS coordinates to nRF Cloud

trantanen:
Added support for sending GNSS coordinates to nRF Cloud. This is only supported with nRF Cloud and is configured with a Kconfig option.

Jira: NCSDK-24322

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -815,6 +815,7 @@ Modem libraries
 
     * Convenience function to get :c:struct:`location_data_details` from the :c:struct:`location_event_data`.
     * Location data details for event :c:enum:`LOCATION_EVT_RESULT_UNKNOWN`.
+    * Sending GNSS coordinates to nRF Cloud when the :kconfig:option:`CONFIG_LOCATION_SERVICE_NRF_CLOUD_GNSS_POS_SEND` Kconfig option is set.
 
 * :ref:`pdn_readme` library:
 
--- a/lib/location/Kconfig
+++ b/lib/location/Kconfig
@@ -86,6 +86,13 @@ config LOCATION_METHOD_GNSS_PRIORITIZE_QZSS_ASSISTANCE
 	  needed at the same time. Enabling this option allows A-GNSS data request to be sent also
 	  when only QZSS assistance data (usually ephemerides) is needed.
 
+config LOCATION_SERVICE_NRF_CLOUD_GNSS_POS_SEND
+	bool "Send GNSS coordinates to nRF Cloud"
+	depends on !LOCATION_SERVICE_EXTERNAL
+	depends on NRF_CLOUD_REST || NRF_CLOUD_MQTT || NRF_CLOUD_COAP
+	help
+	  Send GNSS coordinates to nRF Cloud.
+
 endif # LOCATION_METHOD_GNSS
 
 if LOCATION_METHOD_WIFI
--- a/lib/location/method_gnss.c
+++ b/lib/location/method_gnss.c
@@ -27,6 +27,11 @@
 #if defined(CONFIG_NRF_CLOUD_COAP)
 #include <net/nrf_cloud_coap.h>
 #endif
+#if defined(CONFIG_LOCATION_SERVICE_NRF_CLOUD_GNSS_POS_SEND)
+#include <net/nrf_cloud_codec.h>
+#include <zephyr/sys/timeutil.h>
+#include <time.h>
+#endif
 
 LOG_MODULE_DECLARE(location, CONFIG_LOCATION_LOG_LEVEL);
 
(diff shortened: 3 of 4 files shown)


lib/location/method_gnss.c
```diff
@@ -866,6 +869,106 @@ static void method_gnss_print_pvt(const struct nrf_modem_gnss_pvt_data_frame *pv
...
+}
+
+void method_gnss_cloud_send(struct location_data *location_result)
+{
+	int err = 0;
+	char *json_str_out = NULL;
+	cJSON *gnss_data_obj = NULL;
+	struct nrf_cloud_gnss_data gnss_data = {
+		.type = NRF_CLOUD_GNSS_TYPE_PVT,
+		.ts_ms = NRF_CLOUD_NO_TIMESTAMP, /* TODO: What should be here? */
```
trantanen, line 927: Not sure we want to add dependency to Date Time library to use `date_time_now()`.
location_result.datetime would be available which has been taken from PVT data.
Or then NRF_CLOUD_NO_TIMESTAMP which means cloud will use a time when it receives position data.

trantanen: ATv2 seems to use `k_uptime_get()`.

tokangas: The timestamp could also be calculated from the time in the PVT. That would be the most accurate time available.

trantanen: Yep, time from PVT is possible. I'm wondering why ATv2 hasn't used it but I guess we've just made things "easier" without that conversion.
The easiest would be to use NRF_CLOUD_NO_TIMESTAMP and cloud uses the time from when the data arrives there.

trantanen: Using time from location event which is time from PVT data. `timeutil_timegm64()` used for conversion.

tokangas: I think it would be best to use GNSS time or date_time library time, those should be accurate and because why not. Cloud time is probably also accurate, but it may take some time before the data is received by the cloud.

lib/location/Kconfig
```diff
@@ -86,6 +86,11 @@ config LOCATION_METHOD_GNSS_PRIORITIZE_QZSS_ASSISTANCE
 	  needed at the same time. Enabling this option allows A-GNSS data request to be sent also
 	  when only QZSS assistance data (usually ephemerides) is needed.
 
+config LOCATION_SERVICE_NRF_CLOUD_GNSS_POS_CLOUD_SEND
```
tokangas, line 89: Would the name `LOCATION_SERVICE_NRF_CLOUD_GNSS_POS_SEND` be enough?

lib/location/method_gnss.c
```diff
@@ -866,6 +871,113 @@ static void method_gnss_print_pvt(const struct nrf_modem_gnss_pvt_data_frame *pv
...
+{
+	int err = 0;
+	char *json_str = NULL;
+	cJSON *gnss_data_obj = NULL;
+	struct nrf_cloud_gnss_data gnss_data = {
+		.type = NRF_CLOUD_GNSS_TYPE_PVT,
+		.pvt = {
+			.lon	  = location_result->longitude,
+			.lat	  = location_result->latitude,
+			.accuracy = location_result->accuracy,
```
maxd-nordic, line 932: We could add more info here:
```suggestion
			.accuracy = location_result->accuracy,
			.speed =	pvt_data->speed,
			.has_speed =	location_result->flags & NRF_MODEM_GNSS_PVT_FLAG_VELOCITY_VALID,
			.heading =	location_result->heading,
			.has_heading =	location_result->flags & NRF_MODEM_GNSS_PVT_FLAG_VELOCITY_VALID,
```
Also, altitude might be available if we're not using low-accuracy mode,

lib/location/method_gnss.c
```diff
@@ -866,6 +871,113 @@ static void method_gnss_print_pvt(const struct nrf_modem_gnss_pvt_data_frame *pv
...
+
+	gnss_data_obj = cJSON_CreateObject();
+	if (gnss_data_obj == NULL) {
+		LOG_ERR("Failed to encode GNSS position data, out of memory");
+		err = -ENOMEM;
+		goto cleanup;
+	}
+
+	/* Encode the GNSS location data */
+	err = nrf_cloud_gnss_msg_json_encode(&gnss_data, gnss_data_obj);
```
maxd-nordic, line 954: would be nice to use `nrf_cloud_coap_location_send` if we're using the CoAP backend otherwise.

lib/location/method_gnss.c
```diff
@@ -866,6 +871,113 @@ static void method_gnss_print_pvt(const struct nrf_modem_gnss_pvt_data_frame *pv
...
+		.qos = MQTT_QOS_1_AT_LEAST_ONCE,
+		.topic_type = NRF_CLOUD_TOPIC_MESSAGE,
+	};
+
+	ret = nrf_cloud_send(&mqtt_msg);
+	if (ret) {
+		LOG_ERR("MQTT: location data sending failed: %d", ret);
+	}
+#elif defined(CONFIG_NRF_CLOUD_COAP)
+	ret = nrf_cloud_coap_json_message_send(body, false, true);
```
maxd-nordic, line 893: would be good if we could use cbor with the coap backend using `nrf_cloud_coap_location_send`.

maxd-nordic:
Thanks @trantanen for addressing our feature-request!

lib/location/method_gnss.c
tokangas: I implemented similar support for ATv2 some time ago. I didn't remember that there's a flag for velocity validity in the GNSS API. 😃 But, `NRF_MODEM_GNSS_PVT_FLAG_VELOCITY_VALID` does not tell whether heading is valid or not. For heading, I suggest something like this instead:

https://github.com/nrfconnect/sdk-nrf/blob/e5e367d13d164c3dcfb67e0bcac16111cdd5e4c0/applications/asset_tracker_v2/src/cloud/cloud_codec/nrf_cloud/nrf_cloud_codec.c#L167

Altitude is always valid and could be sent to the cloud. Low accuracy mode doesn't mean that altitude would not be valid. Valid altitude is needed to be able to calculate the position using only three satellites. If the (injected) altitude is incorrect, also the calculated position will have error. If we want, there could be some filtering for the altitude based on `altitude_accuracy`, but we don't do that in ATv2 currently.

trantanen: I'll need to pass PVT data into this function instead of location_result to get this done properly.

trantanen: This is now handled by using `nrf_modem_gnss_pvt_data_frame` and `NRF_CLOUD_GNSS_TYPE_MODEM_PVT`.

Heading issue has to be handled in `nrf_cloud_modem_pvt_data_encode()` if needed via separate PR.

lib/location/method_gnss.c
trantanen: Seems we cannot encode the data into JSON string with CoAP then. I need to see how to do this at code level.

lib/location/Kconfig
trantanen: Changed

lib/location/method_gnss.c
trantanen: Because of CoAP CBOR, I need to go back to using `NRF_CLOUD_GNSS_TYPE_PVT`.

lib/location/method_gnss.c
trantanen: Changed to `nrf_cloud_coap_location_send()` now.

lib/location/method_gnss.c
```diff
@@ -866,6 +871,130 @@ static void method_gnss_print_pvt(const struct nrf_modem_gnss_pvt_data_frame *pv
...
+cleanup:
+	if (gnss_data_obj) {
+		cJSON_Delete(gnss_data_obj);
+	}
+	if (json_str) {
+		cJSON_free(json_str);
+	}
+
+#elif defined(CONFIG_NRF_CLOUD_COAP)
+	/* CoAP is handled differently because we are sending CBOR instead of JSOn data */
```
tokangas, line 988: Also, would it be nice to have a debug log here as well, although there's no body to trace? Now there's no log output with CoAP.
```suggestion
	/* CoAP is handled differently because we are sending CBOR instead of JSON data */
```

tokangas (approved):
Looks good to me, only a minor comment.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -815,6 +815,7 @@ Modem libraries
 
     * Convenience function to get :c:struct:`location_data_details` from the :c:struct:`location_event_data`.
     * Location data details for event :c:enum:`LOCATION_EVT_RESULT_UNKNOWN`.
+    * Sending GNSS coordinates to nRF Cloud when :kconfig:option:`CONFIG_LOCATION_SERVICE_NRF_CLOUD_GNSS_POS_SEND` Kconfig option is set.
```
peknis, line 818:
```suggestion
    * Sending GNSS coordinates to nRF Cloud when the :kconfig:option:`CONFIG_LOCATION_SERVICE_NRF_CLOUD_GNSS_POS_SEND` Kconfig option is set.
```

peknis (approved):

maxd-nordic (approved):

PR #15523, opened 2024-05-27
Prevent local domains from booting if UICR is missing

tomchy:


ahasztag (approved):

e-rk (approved):

subsys/suit/mci/src/suit_mci_nrf54h20.c
```diff
@@ -402,17 +403,43 @@ mci_err_t suit_mci_memory_access_rights_validate(const suit_manifest_class_id_t
 	case SUIT_MANIFEST_APP_RECOVERY:
 	case SUIT_MANIFEST_APP_LOCAL_1:
 	case SUIT_MANIFEST_APP_LOCAL_2:
-	case SUIT_MANIFEST_APP_LOCAL_3:
-		/* App manifest - TODO - implement checks based on UICR
-		 */
+	case SUIT_MANIFEST_APP_LOCAL_3: {
```
SylwesterKonczyk, line 406: We should make sure that the SDFW does NOT overwrite any candidate manifest during the installation process (It would open possibility to inject malicious instructions to already validated candidate manifest). I am not certain, if such checking shall be performed here (during the validation of declared components) or rather in runtime, in implementation of directives like fetch, copy or write.

tomchy: I agree that this code does not prevent from such behavior.
To do so here, we would need to pass the update candidate info to this module, since there are no special markers for the DFU partition(s) in the UICR.
The other option is to add a runtime reservation inside the arbiter, so the Update path may block them before manifests gets processed.
Generally speaking, in this PR the main concern is to stop booting domains if there are no UICRs (thus - no R+S regions defined).
Let's keep the discussion about the address validation in scope of NCSDK-22883 or NCSDK-27375.

PR #15528, opened 2024-05-27
applications: nrf5340_audio: Expand broadcast_source_enable

alexsven:
- Allow user to input their own create_param
- Create helper function to produce create_param
- Alter the API of broadcast_source_enable to take in a struct.
- Struct will determine the number of subgroups, BISes and their info.
- OCT-3016

### DIFF
--- a/applications/nrf5340_audio/broadcast_source/main.c
+++ b/applications/nrf5340_audio/broadcast_source/main.c
@@ -9,6 +9,7 @@
 #include <zephyr/bluetooth/audio/audio.h>
 #include <zephyr/kernel.h>
 #include <zephyr/zbus/zbus.h>
+#include <zephyr/sys/byteorder.h>
 
 #include "nrf5340_audio_common.h"
 #include "nrf5340_audio_dk.h"
@@ -38,6 +39,8 @@ static struct k_thread le_audio_msg_sub_thread_data;
 
 static k_tid_t button_msg_sub_thread_id;
 static k_tid_t le_audio_msg_sub_thread_id;
+
+struct bt_le_ext_adv *ext_adv;
 
 K_THREAD_STACK_DEFINE(button_msg_sub_thread_stack, CONFIG_BUTTON_MSG_SUB_STACK_SIZE);
 K_THREAD_STACK_DEFINE(le_audio_msg_sub_thread_stack, CONFIG_LE_AUDIO_MSG_SUB_STACK_SIZE);
@@ -78,12 +81,12 @@ static void button_msg_sub_thread(void)
 		switch (msg.button_pin) {
 		case BUTTON_PLAY_PAUSE:
 			if (strm_state == STATE_STREAMING) {
-				ret = broadcast_source_stop();
+				ret = broadcast_source_stop(0);
 				if (ret) {
 					LOG_WRN("Failed to stop broadcaster: %d", ret);
 				}
 			} else if (strm_state == STATE_PAUSED) {
-				ret = broadcast_source_start(NULL);
+				ret = broadcast_source_start(0, ext_adv);
 				if (ret) {
 					LOG_WRN("Failed to start broadcaster: %d", ret);
 				}
@@ -236,7 +239,9 @@ static void bt_mgmt_evt_handler(const struct zbus_channel *chan)
 	case BT_MGMT_EXT_ADV_WITH_PA_READY:
 		LOG_INF("Ext adv ready");
 
-		ret = broadcast_source_start(msg->ext_adv);
+		ext_adv = msg->ext_adv;
+
+		ret = broadcast_source_start(msg->index, ext_adv);
 		if (ret) {
 			LOG_ERR("Failed to start broadcaster: %d", ret);
 		}
@@ -298,7 +303,7 @@ void streamctrl_send(void const *const data, size_t size, uint8_t num_ch)
 	struct le_audio_encoded_audio enc_audio = {.data = data, .size = size, .num_ch = num_ch};
 
 	if (strm_state == STATE_STREAMING) {
-		ret = broadcast_source_send(enc_audio);
+		ret = broadcast_source_send(0, enc_audio);
 
 		if (ret != 0 && ret != prev_ret) {
 			if (ret == -ECANCELED) {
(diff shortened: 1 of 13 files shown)


applications/nrf5340_audio/src/bluetooth/bt_management/advertising/bt_mgmt_adv.c
```diff
@@ -144,7 +145,7 @@ static const struct bt_le_ext_adv_cb adv_cb = {
 #endif /* CONFIG_BT_PRIVACY */
 };
 
-static int direct_adv_create(bt_addr_le_t addr)
+static int direct_adv_create(uint8_t ext_adv_index, bt_addr_le_t addr)
```
rick1082, line 148: Would we need to bring this feature to dir-adv creation?

applications/nrf5340_audio/src/bluetooth/bt_management/advertising/bt_mgmt_adv.c
```diff
@@ -247,41 +257,42 @@ static void advertising_process(struct k_work *work)
 	bt_addr_le_t addr;
 
 	if (!k_msgq_get(&bonds_queue, &addr, K_NO_WAIT) && !dir_adv_timed_out) {
-		ret = direct_adv_create(addr);
+		ret = direct_adv_create(ext_adv_index, addr);
```
rick1082, line 260: Probably we don't need to change the connection related ADV-EXT?

applications/nrf5340_audio/src/bluetooth/bt_management/advertising/bt_mgmt_adv.c
alexsven: If we want to support multiple CIGs we will have to include here as well. Also the API for adv is the same for BIS and CIS, so since we are changing it for BIS we need to apply it to CIS. Otherwise it would have to be hard coded to ext_adv[0] everywhere.

applications/nrf5340_audio/src/bluetooth/bt_management/advertising/bt_mgmt_adv.c
alexsven: Same answer as above

applications/nrf5340_audio/src/bluetooth/bt_management/advertising/bt_mgmt_adv.c
```diff
@@ -154,19 +155,19 @@ static int direct_adv_create(bt_addr_le_t addr)
 	ext_adv_param.options |= BT_LE_ADV_OPT_DIR_ADDR_RPA;
 
 	/* Clear ADV data set before update to direct advertising */
-	ret = bt_le_ext_adv_set_data(ext_adv, NULL, 0, NULL, 0);
+	ret = bt_le_ext_adv_set_data(ext_adv[ext_adv_index], NULL, 0, NULL, 0);
 	if (ret) {
 		LOG_ERR("Failed to clear advertising data. Err: %d", ret);
```
rick1082, line 160: Not sure if we show the ext_adv_index when print error log could do help or not. We can discuss.

rick1082 (approved):
I don't see issues, but seems the complexity for broadcast_source is growing a bit.

applications/nrf5340_audio/src/bluetooth/bt_management/advertising/bt_mgmt_adv.c
alexsven: Not until we have multiple advertising sets for CIS at least. It doesn't hurt to include it either, will add it

applications/nrf5340_audio/broadcast_source/main.c
```diff
@@ -312,11 +317,67 @@ void streamctrl_send(void const *const data, size_t size, uint8_t num_ch)
...
+#error Feature is incomplete and should only be used as a guideline for now
+static struct bt_bap_lc3_preset lc3_preset_48 = BT_BAP_LC3_BROADCAST_PRESET_48_4_1(
+	BT_AUDIO_LOCATION_FRONT_LEFT | BT_AUDIO_LOCATION_FRONT_RIGHT, BT_AUDIO_CONTEXT_TYPE_MEDIA);
+
+static void broadcast_create(struct broadcast_source_big *broadcast_param)
+{
+	static enum bt_audio_location location[2] = {BT_AUDIO_LOCATION_FRONT_LEFT,
+						     BT_AUDIO_LOCATION_FRONT_RIGHT};
+	static struct subgroup_config subgroups[2];
+
```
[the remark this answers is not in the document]
alexsven: It is not related to the number of locations though, subgroups are above that and used for things like language. This is only here as an example of how a user can use the new API.

applications/nrf5340_audio/broadcast_source/main.c
```diff
@@ -312,11 +317,67 @@ void streamctrl_send(void const *const data, size_t size, uint8_t num_ch)
...
+	subgroups[0].group_lc3_preset = lc3_preset_48;
+	subgroups[0].num_bises = 2;
+	subgroups[0].context = BT_AUDIO_CONTEXT_TYPE_MEDIA;
```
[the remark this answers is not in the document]
alexsven: Same as above. I won't change these magic numbers as they are only there as examples on how to create a custom broadcaster. Using defines would make it harder to understand for customers, I believe.

applications/nrf5340_audio/broadcast_source/main.c
```diff
@@ -312,11 +317,67 @@ void streamctrl_send(void const *const data, size_t size, uint8_t num_ch)
...
+	broadcast_param->encryption = false;
+
+	bt_audio_codec_cfg_meta_set_bcast_audio_immediate_rend_flag(
+		&subgroups[0].group_lc3_preset.codec_cfg);
+	bt_audio_codec_cfg_meta_set_bcast_audio_immediate_rend_flag(
+		&subgroups[1].group_lc3_preset.codec_cfg);
+
+	uint8_t spanish_src[3] = "spa";
+	uint8_t english_src[3] = "eng";
+
```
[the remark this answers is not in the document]
alexsven: Same as above, won't use defines for this code since it is only temporary here. Partly because of the new main file that is coming and partly to show what is possible to do with this API. There is no define for `eng` in Zephyr, and I guess part of the reason is that ISO 639-3 contains roughly 520 entries just for the letter 'A'...

gWacey (approved):

PR #15529, opened 2024-05-27
Bluetooth: Mesh: Add note on blocking API calls to CLI headers

akredalen:
Added a paragraph to model_types.rst in the Bluetooth Mesh library documentation to explain restrictions on using blocking API calls.

### DIFF
--- a/doc/nrf/libraries/bluetooth_services/mesh/model_types.rst
+++ b/doc/nrf/libraries/bluetooth_services/mesh/model_types.rst
@@ -57,6 +57,15 @@ The delay should not be taken into account when calculating the remaining transi
 Server models are taking care of publishing of status messages, when receiving a state changing message, as well as sending a response back to a client, when an acknowledged message is received.
 If a state change is non-instantaneous, for example when :c:func:`bt_mesh_model_transition_time` returns a nonzero value, the application is responsible for publishing a new value of the state at the end of the transition.
 
+.. _bt_mesh_models_common_blocking_api_rule:
+
+Common rule for the model's blocking API
+****************************************
+
+Some client APIs can be called as blocking if the response buffer is set to non-NULL.
+Blocking calls must only be made from threads where they do not disrupt Zephyr’s system services.
+Thus, do not make blocking calls in interrupt handlers, such as button handlers, or kernel services, such as system work queue handlers.
+
 .. _bt_mesh_models_common_types:
 
 Common types for all models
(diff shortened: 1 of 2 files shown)


include/bluetooth/mesh/gen_battery_cli.h
```diff
@@ -83,6 +83,9 @@ struct bt_mesh_battery_cli {
  * function will return, and the response will be passed to the
  * @ref bt_mesh_battery_cli::status_handler callback.
  *
+ * @note Blocking calls must be made from the main cooperative thread only. Do not use in
+ * button handlers or workqueues.
```
ludvigsj, line 87: They probably _can_ be used in a workqueue, if you set it up to run in a thread with correct priority, I think. Suggest to change to
```suggestion
 * button handlers or the system workqueue.
```

ludvigsj (changes requested):
I'm not sure if I like adding this to every single API function. Would it be possible to find a more central location for this note?

akredalen:
> I'm not sure if I like adding this to every single API function. Would it be possible to find a more central location for this note?

I am not sure either. At least now it is very clear whenever using the API. I did consider finding a more centralized location for this note, but I worry it might end up being too hidden and thereby useless. Any suggestions?

include/bluetooth/mesh/gen_loc_cli.h
```diff
@@ -100,6 +100,9 @@ struct bt_mesh_loc_cli {
  * function will return, and the response will be passed to the
  * @ref bt_mesh_loc_cli_handlers::global_status callback.
  *
+ * @note Blocking calls must be made from the main cooperative thread only. Do not use in
```
alxelax, line 103: > 
Blocking calls must be made from the threads, blocking of which doesn't disrupt functionality of Zephyr system services. Do not use blocking calls from interrupt handlers (for example button handlers) and kernel services (for example the system work queue handlers).

alxelax (changes requested):

akredalen:
Any opinion about where to add this note, @alxelax? See Ludvig's comment.

alxelax:
> Any opinion about where to add this note, @alxelax? See Ludvig's comment.

In general, I agree with Ludvig, but I do not have any good suggestion for this. I leave it up to your decision.

alxelax:
Probably this page: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/libraries/bluetooth_services/mesh/model_types.html
as `Common rule of the model's blocking API` 
but again, if you think this is bad idea just ignore this comment

akredalen:
> Probably this page: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/libraries/bluetooth_services/mesh/model_types.html as `Common rule of the model's blocking API` but again, if you think this is bad idea just ignore this comment

Great! Thanks. I agree this is a logical place to add the note. It might not be the most obvious place for a user to check before using the API but I still think it is good enough for now. We could relocate later if we see the need to do so.

alxelax (approved):

ludvigsj (approved):

Andrewpini (approved):

mia-ko (approved):
LGTM
The placement looks ok as well, let's see if we need any changes if any feedback comes from the customers.
And I always ask this... do we need a changelog entry? :)

PR #15535, opened 2024-05-28
lib: at_host: use CRLF termination as default

eivindj-nordic:
Use CRLF termination as default as that is the standard AT termination characters and also used by the Serial Terminal in nRF Connect for Desktop.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -767,6 +767,10 @@ DFU libraries
 Modem libraries
 ---------------
 
+* :ref:`lib_at_host`:
+
+   * Updated to set the default termination mode to the :kconfig:option:`CONFIG_CR_LF_TERMINATION` Kconfig option instead of the :kconfig:option:`CONFIG_CR_TERMINATION` Kconfig option.
+
 * :ref:`nrf_modem_lib_readme`:
 
   * Fixed an issue with the CFUN hooks when the Modem library is initialized during ``SYS_INIT`` at kernel level and makes calls to the :ref:`nrf_modem_at` interface before the application level initialization is done.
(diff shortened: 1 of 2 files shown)


MirkoCovizzi (approved):

lemrey:
CC @charlieshao5189

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -767,6 +767,10 @@ DFU libraries
 Modem libraries
 ---------------
 
+* :ref:`lib_at_host`:
+
+  * Changed the default termination mode from the :kconfig:option:`CONFIG_CR_TERMINATION` Kconfig option to the :kconfig:option:`CONFIG_CR_LF_TERMINATION` kconfig option.
```
divipillai, line 772:
```suggestion
  * Updated to set the default termination mode in the :kconfig:option:`CONFIG_CR_LF_TERMINATION` Kconfig option instead of the ``CONFIG_CR_TERMINATION`` kconfig option.
```

eivindj-nordic: `CONFIG_CR_TERMINATION` is still a valid option, so I'll keep the format for it.

balaji-nordic (approved):

lemrey (approved):
Thanks

divipillai (approved):

jtguggedal:
I think this PR will need to be reverted, as CR is the default termination on non-Windows OS and this change will cause issues for a lot of users

jtguggedal:
> In addition \r\n is the standard for marking an end of an AT command.

eivindj-nordic:
> But if `<CR><LF>` is required to work witho our tools by default, I suppose we have to make this change.

You're right `<CR>` is standard in https://www.etsi.org/deliver/etsi_ts/127000_127099/127007/18.06.00_60/ts_127007v180600p.pdf. 
If <CR> works well with Linux too, maybe we can make the fix in the tools instead (change to use `<CR>` as default)? What do you think @chunfantasy?

PR #15467, opened 2024-05-23
sysbuild: Add missing board configurations

nordicjm:

PR #15542, opened 2024-05-28
applications: serial_lte_modem: Add GNSS data to carrier library

stig-bjorlykke:
When carrier library is enabled the GNSS data will be provided using the lwm2m_carrier_location_set() and lwm2m_carrier_velocity_set() API.

### DIFF
--- a/applications/serial_lte_modem/src/gnss/slm_at_gnss.c
+++ b/applications/serial_lte_modem/src/gnss/slm_at_gnss.c
@@ -28,6 +28,11 @@ static void pgps_event_handler(struct nrf_cloud_pgps_event *event);
 
 #endif /* CONFIG_SLM_NRF_CLOUD */
 
+#if defined(CONFIG_SLM_CARRIER)
+#include <time.h>
+#include <lwm2m_carrier.h>
+#endif
+
 #include "slm_util.h"
 #include "slm_at_host.h"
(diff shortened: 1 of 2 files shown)


MarkusLassila:
@stig-bjorlykke: This will fail until this PR goes in: https://github.com/nrfconnect/sdk-nrf/pull/15467

applications/serial_lte_modem/src/gnss/slm_at_gnss.c
```diff
@@ -533,14 +531,36 @@ static void fix_rep_wk(struct k_work *work)
...
+#endif
+
+#if defined(CONFIG_SLM_NRF_CLOUD)
+
+	if (slm_nrf_cloud_send_location && slm_nrf_cloud_ready) {
+		/* Report to nRF Cloud by best-effort */
+		send_location(&pvt);
+	}
 
+#if defined(CONFIG_NRF_CLOUD_PGPS) && defined(CONFIG_SLM_PGPS_INJECT_FIX_DATA)
```
tomi-font, line 543: As you removed it from the previous `#ifdef`, might as well here:
```suggestion
#if defined(CONFIG_SLM_PGPS_INJECT_FIX_DATA)
```

tomi-font (approved):

wiba-nordic (approved):

PR #15549, opened 2024-05-28
boards: nrf53: config EXTDOMAIN in soc to reduce code duplication

SebastianBoe:
Configure the "cpunet" to be secure from the soc system initialization to avoid code duplication across all the boards.

cpunet is now configured to be secure earlier in the boot process, from POST_KERNEL to PRE_KERNEL, but AFAIK this is not a problem.


SebastianBoe:
no test results yet because the BLE test plan is red on master.

Waiting for BLE to fix their test plan.

PR #15560, opened 2024-05-29
bords: shields: pca63566: Fix overlay for nrf54h20

nordic-segl:
Instance 130 can't be i2c and spi at the same time.
Fix build error resulting from not enabled DTS nodes.

### DIFF
--- a/boards/shields/pca63566/boards/nrf54h20dk_nrf54h20_cpuapp.overlay
+++ b/boards/shields/pca63566/boards/nrf54h20dk_nrf54h20_cpuapp.overlay
@@ -6,10 +6,6 @@
 	};
 };
 
-&cpuapp_dma_region {
-	status="okay";
-};
-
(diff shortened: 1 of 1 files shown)


nordic-bami (approved):

kl-cruz (approved):

maje-emb (commented):
Can we also add configurations for the PPR core?

nordic-segl:
> Can we also add configurations for the PPR core?

Can this be done in other PR?
Because:
- I need this ASAP to have results of TWI/SPI driver tests on nrf54h20 before NCS2.7.0 release.
- I don't know how to prepare overlay for PPR.
- I heard that people struggle with executing driver tests on PPR. PPR doesn't have it's own UICR, so every resource used by PPR must be requested by its owner. When using SysBuild, SHIELD is applied to all cores.

PR #15579, opened 2024-05-29
suit: build recovery image from main application

ahasztag:
This PR allows to build the recovery image as a child image of a main application by using the SB_CONFIG_SUIT_BUILD_RECOVERY option.

### DIFF
--- a/cmake/sysbuild/suit.cmake
+++ b/cmake/sysbuild/suit.cmake
@@ -134,6 +134,10 @@ function(suit_register_post_build_commands)
     list(APPEND dependencies "${BINARY_DIR}/zephyr/${BINARY_FILE}.bin")
   endforeach()
 
+  if (SB_CONFIG_SUIT_BUILD_RECOVERY)
+    list(APPEND dependencies recovery)
+  endif()
+
   add_custom_target(
     create_suit_artifacts
     ALL
@@ -318,6 +322,16 @@ function(suit_setup_merge)
       list(APPEND ARTIFACTS_TO_MERGE ${IMAGE_BINARY_DIR}/zephyr/uicr.hex)
     endif()
 
+    if(SB_CONFIG_SUIT_BUILD_RECOVERY)
+      get_property(
+        recovery_artifacts
+        GLOBAL PROPERTY
+        SUIT_RECOVERY_ARTIFACTS_TO_MERGE_${IMAGE_TARGET_NAME}
+      )
+
+      list(APPEND ARTIFACTS_TO_MERGE ${recovery_artifacts})
+    endif()
+
     set_property(
       GLOBAL APPEND PROPERTY SUIT_POST_BUILD_COMMANDS
       COMMAND ${PYTHON_EXECUTABLE} ${ZEPHYR_BASE}/scripts/build/mergehex.py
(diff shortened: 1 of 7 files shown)


sysbuild/Kconfig.suit
```diff
@@ -69,6 +69,12 @@ config SUIT_ENVELOPE_NORDIC_TOP_DIRECTORY
 
 endif # SUIT_ENVELOPE
 
+config SUIT_BUILD_RECOVERY
+	bool "Build SUIT recovery firmware"
```
tomchy, line 73: Looking at the integration (`hci_ipc` files are assumed to be present), I guess it `depends on SB_CONFIG_NETCORE_HCI_IPC`.

tomchy (approved):

sysbuild/Kconfig.suit
ahasztag: Added

cmake/sysbuild/suit.cmake
```diff
@@ -317,6 +331,50 @@ function(suit_setup_merge)
...
+  set_property(
+    GLOBAL APPEND PROPERTY SUIT_RECOVERY_ARTIFACTS_TO_MERGE_application ${BINARY_DIR}/recovery/zephyr/zephyr.hex)
+
+  set_property(
+    GLOBAL APPEND PROPERTY SUIT_RECOVERY_ARTIFACTS_TO_MERGE_radio
+    ${BINARY_DIR}/recovery/zephyr/suit_installed_envelopes_radio_merged.hex)
+  set_property(
+    GLOBAL APPEND PROPERTY SUIT_RECOVERY_ARTIFACTS_TO_MERGE_radio ${BINARY_DIR}/hci_ipc/zephyr/zephyr.hex)
+
+endfunction()
```
e-rk, line 372: Shouldn't this go to `sysbuild/suit.cmake`?

cmake/sysbuild/suit.cmake
```diff
@@ -317,6 +331,50 @@ function(suit_setup_merge)
...
+  set(sysbuild_root_path "${ZEPHYR_NRF_MODULE_DIR}/../zephyr/share/sysbuild")
+
+  ExternalProject_Add(
+    recovery
+    SOURCE_DIR ${sysbuild_root_path}
+    PREFIX ${CMAKE_BINARY_DIR}/recovery
+    INSTALL_COMMAND ""
+    CMAKE_CACHE_ARGS
+      "-DAPP_DIR:PATH=${ZEPHYR_NRF_MODULE_DIR}/samples/suit/recovery"
+      "-DBOARD:STRING=nrf54h20dk/nrf54h20/cpuapp"
```
e-rk, line 351: `nrf54h20dk` cannot be assumed. See https://github.com/nrfconnect/sdk-nrf/blob/main/sysbuild/suit.cmake#L16 to see how to make the target string.

e-rk (changes requested):

cmake/sysbuild/suit.cmake
ahasztag: The way it is done currently it would work, however I've originally placed it here because I needed overlay files from specific images, which were only accessible during the POST_CMAKE functions. 

I wouldn't say we have a guarantee that we won't need accessing such per-image variable. For example, I think we might need to access the MPI configuration which is defined by the application core image.

That is why, I'd rather not move this.

cmake/sysbuild/suit.cmake
ahasztag: Ok fixed it. I didn't go for exactly the same solution, however the DK is no longer assumed.

cmake/sysbuild/suit.cmake
ahasztag: By the way I see I should add passing MPI configuration (vendor and application name)

e-rk (approved):

subsys/suit/provisioning/soc/Kconfig.nrf54h20
```diff
@@ -39,7 +39,7 @@ rsource "Kconfig.template.manifest_config"
 endmenu
 
 manifest=APP_RECOVERY
-manifest-class=nRF54H20_sample_app_recovery
+manifest-class=nRF54H20_app_recovery
```
tomchy, line 42: Please remove the `default y` for recovery MPIs. Currently the ROOT manifest is not used as recovery manifest in case of missing recovery manifest.

tomchy (changes requested):

subsys/suit/provisioning/soc/Kconfig.nrf54h20
ahasztag: Fixed here: https://github.com/nrfconnect/sdk-nrf/pull/15639 - additionally, added https://github.com/ahasztag/sdk-nrf/blob/c13c4ba5c14751933190970008178624dc1422e9/sysbuild/CMakeLists.txt#L405-L408 to achieve MPI generation when building with recovery

tomchy (approved):

nordicjm:
`This PR allows to build the recovery image as a child image of`

Never refer to sysbuild images as `child image`, they are just `images`

ahasztag:
> `This PR allows to build the recovery image as a child image of`
> 
> Never refer to sysbuild images as `child image`, they are just `images`

But it is not a sysbuild image. It is in fact a child image. Achieving what we wanted turned out to be impossible with a sysbuild image.

nordicjm:
> > `This PR allows to build the recovery image as a child image of`
> > Never refer to sysbuild images as `child image`, they are just `images`
> 
> But it is not a sysbuild image. It is in fact a child image. Achieving what we wanted turned out to be impossible with a sysbuild image.

What? SB_CONFIG_* is sysbuild, it has no relation to child image

PR #15580, opened 2024-05-29
doc: update VSC extension links in links.txt

greg-fer:
Updated entries pointing to the VSC extension docs in links.txt.


peknis:
A couple of nRF5-related Infocenter links still remaining. But do we have the v17.0.2 targets in Tech docs available? And the one generic link to Infocenter.

peknis (approved):

greg-fer:
> A couple of nRF5-related Infocenter links still remaining. But do we have the v17.0.2 targets in Tech docs available? And the one generic link to Infocenter.

On second thought, I removed fixes to Infocenter links that I put in the original commit. All of these should be handled separately. Let's keep this PR about the extension links.

PR #15588, opened 2024-05-30
tests: drivers: migrate flash_rpc to sysbuild

sigvartmh:
Migrate the flash rpc driver test to use sysbuild.

### DIFF
--- a/tests/drivers/flash/flash_rpc/Kconfig
+++ b/tests/drivers/flash/flash_rpc/Kconfig
@@ -15,4 +15,5 @@ config APP_INCLUDE_REMOTE_IMAGE
 
 config APP_REMOTE_BOARD
 	string "The name of the CORE to be used by remote image"
+	default "nrf5340dk/nrf5340/cpunet"
 	depends on APP_INCLUDE_REMOTE_IMAGE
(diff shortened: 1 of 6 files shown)


sigvartmh:
Tested with ` west build -b nrf5340dk/nrf5340/cpuapp --sysbuild -p -- -DSB_CONF_FILE=sysbuild/nrf5340dk_nrf5340_cpunet.conf`

tests/drivers/flash/flash_rpc/Kconfig
```diff
@@ -15,4 +15,5 @@ config APP_INCLUDE_REMOTE_IMAGE
 
 config APP_REMOTE_BOARD
```
sigvartmh, line 16: I'm a bit uncertain if this is still needed

nvlsianpu: not needed anymore

nvlsianpu (approved):

tests/drivers/flash/flash_rpc/Kconfig.sysbuild
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+source "${ZEPHYR_BASE}/share/sysbuild/Kconfig"
+
+config REMOTE_BOARD
+	string "The board used for remote target"
```
nordicjm, line 10: Should be using a qualifier not a board name, and it can be defaulted in here e.g.
```
default "cpunet" if SOC_SERIES_NORDIC_NRF53X
```

PR #15600, opened 2024-05-30
sysbuild: use specific certificate for Thingy:91

maxd-nordic:
This patch makes sure that for the Thingy:91,
the correct bootloader certificate is used.


balaji-nordic (approved):

jorgenmk (approved):
Tested working

PR #15603, opened 2024-05-31
net: lib: nrf_cloud_coap: Enable ground fix conf

plskeggs:
Now that cloud support is released, enable CoAP support for the ground fix configuration flags do_reply, hi_conf, and fallback.

Also set default value for .config to NULL when initializing struct nrf_cloud_rest_location_request in the location library. This structure is shared for REST and CoAP use.

Improve recent SO_KEEPOPEN support to better handle mfw_2.0.0.

Jira: IRIS-9014

### DIFF
--- a/include/net/nrf_cloud_coap.h
+++ b/include/net/nrf_cloud_coap.h
@@ -94,6 +94,15 @@ int nrf_cloud_coap_pause(void);
  */
 int nrf_cloud_coap_resume(void);
 
+/**@brief Check if you can pause and resume safely.
+ *
+ * Check if the device can avoid performing a full handshake after a temporary network outage ends.
+ *
+ * @retval false if SO_KEEPOPEN is not supported.
+ * @retval true if it is supported.
+ */
+bool nrf_cloud_coap_keepopen_is_supported(void);
+
 /**
  * @brief Disconnect the nRF Cloud CoAP connection.
  *
--- a/lib/location/cloud_service/cloud_service_nrf.c
+++ b/lib/location/cloud_service/cloud_service_nrf.c
@@ -115,7 +115,8 @@ int cloud_service_nrf_pos_get(
 #endif
 	const struct nrf_cloud_rest_location_request loc_req = {
 		.cell_info = params->cell_data,
-		.wifi_info = params->wifi_data
+		.wifi_info = params->wifi_data,
+		.config = NULL
 	};
 	struct nrf_cloud_location_result result;
 
--- a/subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap
+++ b/subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap
@@ -27,15 +27,6 @@ config NRF_CLOUD_COAP_SERVER_PORT
 	int "CoAP server port"
 	default 5684
 
-config NRF_CLOUD_COAP_GF_CONF
-	bool "Enable configuration options for ground-fix requests"
-	help
-	  The initial release of CoAP for nRF Cloud did not provide a way to
-	  configure the response; it defaulted to always returning the location
-	  and method used to acquire the location. The cloud will soon support
-	  configuration values do_reply, fallback, and hi_conf. When the cloud
-	  support is ready, NRF_CLOUD_COAP_GF_CONF can be set true.
-
 config NRF_CLOUD_COAP_KEEPOPEN
 	bool "Enable SO_KEEPOPEN to prevent unnecessary socket closures"
 	default y
--- a/subsys/net/lib/nrf_cloud/coap/include/nrfc_dtls.h
+++ b/subsys/net/lib/nrf_cloud/coap/include/nrfc_dtls.h
@@ -54,4 +54,13 @@ int nrfc_dtls_session_save(int sock);
  */
 int nrfc_dtls_session_load(int sock);
 
+/** @brief Check if you can use SO_KEEPOPEN.
+ *
+ * Modem firmware < 2.0.1 does not support this.
+ *
+ * @retval true - SO_KEEPOPEN can be used.
+ * @retval false - it cannot be used.
+ */
+bool nrfc_keepopen_is_supported(void);
+
 #endif /* NRFC_DTLS_H_ */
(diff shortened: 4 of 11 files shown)


lib/location/cloud_service/cloud_service_nrf.c
```diff
@@ -115,7 +115,8 @@ int cloud_service_nrf_pos_get(
 #endif
 	const struct nrf_cloud_rest_location_request loc_req = {
 		.cell_info = params->cell_data,
-		.wifi_info = params->wifi_data
+		.wifi_info = params->wifi_data,
+		.config = NULL
```
tokangas, line 119: This is not strictly necessary, because when you initialize a struct like this, rest of the fields are automatically initialized to zero.

subsys/net/lib/nrf_cloud/Kconfig.nrf_cloud_coap
```diff
@@ -29,12 +29,15 @@ config NRF_CLOUD_COAP_SERVER_PORT
...
 	bool "Enable configuration options for ground-fix requests"
+	default y
 	help
-	  The initial release of CoAP for nRF Cloud did not provide a way to
-	  configure the response; it defaulted to always returning the location
-	  and method used to acquire the location. The cloud will soon support
-	  configuration values do_reply, fallback, and hi_conf. When the cloud
-	  support is ready, NRF_CLOUD_COAP_GF_CONF can be set true.
+	  Allow configuration of the ground fix request with nrf_cloud_coap_location_get().
+	  The do_reply flag controls whether RF Cloud will send the
```
tokangas, line 35:
```suggestion
	  The do_reply flag controls whether nRF Cloud will send the
```

tokangas (commented):
The description for `CONFIG_NRF_CLOUD_COAP_GF_CONF` is quite complex. While trying to understand the behavior, I started wondering if the option is still needed. If the user doesn't want to use the configuration, the `config` can be set to `NULL`.

jayteemo:
> While trying to understand the behavior, I started wondering if the option is still needed. If the user doesn't want to use the configuration, the `config` can be set to `NULL`.

i agree.  
if the default value is left at `n`, then you could say it is preventing the config from being unintentionally used now that it is 
functional.
but since it is defaults to `y` now, i don't understand the value the config option provides.

jayteemo (approved):

plskeggs:
Added DNM so I can test the new commit.

subsys/net/lib/nrf_cloud/coap/src/nrfc_dtls.c
```diff
@@ -167,19 +168,30 @@ int nrfc_dtls_setup(int sock)
...
 	}
 
+	keepopen_supported = false;
 	if (IS_ENABLED(CONFIG_NRF_CLOUD_COAP_KEEPOPEN)) {
 		err = setsockopt(sock, SOL_SOCKET, SO_KEEPOPEN, &(int){1}, sizeof(int));
 		if (err) {
-			LOG_ERR("Failed to set SO_KEEPOPEN: %d", -errno);
-			err = -errno;
+			if (errno != EINVAL) {
+				LOG_ERR("Failed to set SO_KEEPOPEN: %d", -errno);
```
jayteemo, line 183: when reading `keepopen_supported`, i thought "supported by the modem", as in, it should work.
but this handling seems to be more like `keepopen_enabled`.

for this error case, does it mean that keepopen _is_ supported, but failed for whatever reason?

plskeggs: I am not sure what all possible errno values are for this socket option and what they all mean. I can probably just simplify this to assume any error indicates we can't use SO_KEEPOPEN though.

jayteemo (approved):

subsys/net/lib/nrf_cloud/coap/src/nrf_cloud_coap_transport.c
```diff
@@ -835,6 +840,10 @@ int nrf_cloud_coap_transport_pause(struct nrf_cloud_coap_client *const client)
 	int err = 0;
 	int tmp;
 
+	if (client->sock == -1) {
+		return -EBADF; /* Device is disconnected. */
```
jayteemo, line 844: add this to to the possible retval's in the header comments

subsys/net/lib/nrf_cloud/coap/src/nrfc_dtls.c
trantanen: > I can probably just simplify this to assume any error indicates we can't use SO_KEEPOPEN though.

Agreed.

trantanen:
> The description for `CONFIG_NRF_CLOUD_COAP_GF_CONF` is quite complex. While trying to understand the behavior, I started wondering if the option is still needed. If the user doesn't want to use the configuration, the `config` can be set to `NULL`.

@tokangas is out this week but removing this Kconfig option looks good now.

trantanen (approved):

include/net/nrf_cloud_coap.h
```diff
@@ -94,6 +94,15 @@ int nrf_cloud_coap_pause(void);
  */
 int nrf_cloud_coap_resume(void);
 
+/**@brief Check if we can pause and resume safely.
+ *
+ * Check if device can avoid performing a full handshake after temporary network outage ends.
```
peknis, line 99:
```suggestion
/**@brief Check if you can pause and resume safely.
 *
 * Check if the device can avoid performing a full handshake after a temporary network outage ends.
```

subsys/net/lib/nrf_cloud/coap/include/nrfc_dtls.h
```diff
@@ -54,4 +54,13 @@ int nrfc_dtls_session_save(int sock);
  */
 int nrfc_dtls_session_load(int sock);
 
+/** @brief Check if we can use SO_KEEPOPEN.
```
peknis, line 57:
```suggestion
/** @brief Check if you can use SO_KEEPOPEN.
```

peknis (approved):
Approved with minor nits.

PR #15613, opened 2024-05-31
manifest: bring CAN support to nRF54H20

gmarull:
Pull Zephyr changes that bring CAN support to nRF54H20.


gmarull:
sha updated to newer revision in another PR, closing.

PR #15614, opened 2024-05-31
suit: enable SUIT envelope creation for app and rad cores only

robertstypa:
Fix: enable generation of SUIT envelopes only in case of CPUAPP or CPURAD cores

Ref: NONE

Signed-off-by: Robert Stypa <robert.stypa@nordicsemi.no>


adsz-nordic (approved):
This fixed the problem in my use case

PR #15620, opened 2024-05-31
samples: bluetooth: add missing sysbuild config

henrla:
Add missing sysbuild configuration for multicore
targets running the iso_time_sync sample.


hermabe (approved):

yuxCai (approved):

yuxCai:
are these sysbuild configs also missing for the cis and bis sample? It'd be nice if you could fix that sample build too

henrla:
> are these sysbuild configs also missing for the cis and bis sample? It'd be nice if you could fix that sample build too

No, it did not seem like it was needed.

PR #15622, opened 2024-05-31
suit: remove duplicate artifacts

e-rk:
Certain SUIT artifacts were duplicated. For example, when building the smp_transfer sample, it would produce two identical binaries smp_transfer.bin and application.bin. With this commit only smp_transfer.bin would be produced.


ahasztag (changes requested):
I think I need a call to see if this won't break the recovery application build.

For the recovery application, `CONFIG_SUIT_LOCAL_ENVELOPE_GENERATE` is set to n, yet the `application` data is still required in the manifest. Won't this PR break it?

cmake/sysbuild/suit.cmake
```diff
@@ -184,27 +184,25 @@ function(suit_create_package)
...
     set(BINARY_FILE "${BINARY_FILE}.bin")
 
     list(APPEND CORE_ARGS
-      --core ${image},${SUIT_ROOT_DIRECTORY}${image}.bin,${BINARY_DIR}/zephyr/edt.pickle,${BINARY_DIR}/zephyr/.config
+      --core ${target},${SUIT_ROOT_DIRECTORY}${image}.bin,${BINARY_DIR}/zephyr/edt.pickle,${BINARY_DIR}/zephyr/.config
     )
 
-    if(DEFINED target AND NOT target STREQUAL "")
-      list(APPEND CORE_ARGS
-      --core ${target},${SUIT_ROOT_DIRECTORY}${target}.bin,${BINARY_DIR}/zephyr/edt.pickle,${BINARY_DIR}/zephyr/.config
```
ahasztag, line 195: This seems dangerous, couldn't we leave this part and change it to:

`--core ${target},${SUIT_ROOT_DIRECTORY}${image}.bin,${BINARY_DIR}/zephyr/edt.pickle,${BINARY_DIR}/zephyr/.config`

to fix the problem that `${target}.bin` won't be available?

ahasztag (changes requested):

ahasztag (approved):

tomchy (approved):

PR #15624, opened 2024-05-31
lib: pdn: destroy context on CFUN 0

eivindj-nordic:
Destroy context on CFUN 0 as that is done in the modem.

PR requires changes to the carrier library, which will not go in for NCS 2.7.0.

TODOs: 

- [x]  Check if `pdn_ctx_destroy()` can be removed from some samples.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -671,16 +671,18 @@ Modem libraries
 
 * :ref:`nrf_modem_lib_readme`:
 
+  * Updated the RTT trace backend to allocate the RTT channel at boot, instead of when the modem is activated.
   * Rename the nRF91 socket offload layer from ``nrf91_sockets`` to ``nrf9x_sockets`` to reflect that the offload layer is not exclusive to the nRF91 Series SiPs.
+  * Removed support for deprecated RAI socket options ``SO_RAI_LAST``, ``SO_RAI_NO_DATA``, ``SO_RAI_ONE_RESP``, ``SO_RAI_ONGOING``, and ``SO_RAI_WAIT_MORE``.
 
 * :ref:`modem_info_readme` library:
 
   * Fixed a potential issue with scanf in the :c:func:`modem_info_get_current_band` function, which could lead to memory corruption.
 
-* :ref:`nrf_modem_lib_readme` library:
+* :ref:`pdn_readme` library:
 
-  * Updated the RTT trace backend to allocate the RTT channel at boot, instead of when the modem is activated.
-  * Removed support for deprecated RAI socket options ``SO_RAI_LAST``, ``SO_RAI_NO_DATA``, ``SO_RAI_ONE_RESP``, ``SO_RAI_ONGOING``, and ``SO_RAI_WAIT_MORE``.
+  * Added the event ``PDN_EVENT_CTX_DESTROYED`` to indicate when a PDP context is destroyed.
+    This happens when the modem is switched to minimum functionality mode (``CFUN=0``).
 
 Multiprotocol Service Layer libraries
 -------------------------------------
--- a/include/modem/pdn.h
+++ b/include/modem/pdn.h
@@ -56,14 +56,26 @@ struct pdn_pdp_opt {
 
 /** @brief PDN library event */
 enum pdn_event {
-	PDN_EVENT_CNEC_ESM,		/**< +CNEC ESM error code */
-	PDN_EVENT_ACTIVATED,		/**< PDN connection activated */
-	PDN_EVENT_DEACTIVATED,		/**< PDN connection deactivated */
-	PDN_EVENT_IPV6_UP,		/**< PDN has IPv6 connectivity */
-	PDN_EVENT_IPV6_DOWN,		/**< PDN has lost IPv6 connectivity */
-	PDN_EVENT_NETWORK_DETACH,	/**< Network detached */
-	PDN_EVENT_APN_RATE_CONTROL_ON,	/**< APN rate control is ON for given PDN */
-	PDN_EVENT_APN_RATE_CONTROL_OFF, /**< APN rate control is OFF for given PDN */
+	/** +CNEC ESM error code */
+	PDN_EVENT_CNEC_ESM,
+	/** PDN connection activated */
+	PDN_EVENT_ACTIVATED,
+	/** PDN connection deactivated */
+	PDN_EVENT_DEACTIVATED,
+	/** PDN has IPv6 connectivity */
+	PDN_EVENT_IPV6_UP,
+	/** PDN has lost IPv6 connectivity */
+	PDN_EVENT_IPV6_DOWN,
+	/** Network detached */
+	PDN_EVENT_NETWORK_DETACH,
+	/** APN rate control is ON for given PDN */
+	PDN_EVENT_APN_RATE_CONTROL_ON,
+	/** APN rate control is OFF for given PDN */
+	PDN_EVENT_APN_RATE_CONTROL_OFF,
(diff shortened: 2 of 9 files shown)


include/modem/pdn.h
```diff
@@ -64,6 +64,7 @@ enum pdn_event {
 	PDN_EVENT_NETWORK_DETACH,	/**< Network detached */
 	PDN_EVENT_APN_RATE_CONTROL_ON,	/**< APN rate control is ON for given PDN */
 	PDN_EVENT_APN_RATE_CONTROL_OFF, /**< APN rate control is OFF for given PDN */
+	PDN_EVENT_CTX_DESTROYED,	/**< Context is destroyed for given PDN */
```
lemrey, line 67: We could add a comment here so that it's clear that this happened because of `CFUN=0`, for example,
`Modem switched to minimum functionality mode, PDP context was destroyed`.

eivindj-nordic: updated

lib/pdn/pdn.c
```diff
@@ -692,6 +698,17 @@ static void on_cfun(int mode, void *ctx)
...
+			}
+
+			sys_slist_find_and_remove(&pdn_contexts, &pdn->node);
+			k_free(pdn);
+		}
+		k_mutex_unlock(&list_mutex);
+
 #if defined(CONFIG_PDN_DEFAULTS_OVERRIDE)
 		pdn_defaults_override();
 #endif
```
lemrey, line 714: I think these should be overridden only when changing to `MODEM_CFUN_NORMAL`, `MODEM_CFUN_ACTIVATE_LTE`.

eivindj-nordic: PDP configuration is lost on modem side on cfun=0. Therefore I think it makes sense to do it on CFUN=0.

lemrey: Alright time for another coffee ☕

lemrey (approved):

anhmolt (approved):

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -588,14 +588,15 @@ Modem libraries
...
 
 * :ref:`modem_info_readme` library:
 
   * Fixed a potential issue with scanf in the :c:func:`modem_info_get_current_band` function, which could lead to memory corruption.
 
-* :ref:`nrf_modem_lib_readme` library:
+* :ref:`pdn_readme` library:
 
-  * Removed support for deprecated RAI socket options ``SO_RAI_LAST``, ``SO_RAI_NO_DATA``, ``SO_RAI_ONE_RESP``, ``SO_RAI_ONGOING``, and ``SO_RAI_WAIT_MORE``.
+  * Added the event ``PDN_EVENT_CTX_DESTROYED`` to indicate when a PDNs PDP context is destroyed. This happens when the modem is switched to minimum functionality mode (``CFUN=0``).
```
divipillai, line 599:
```suggestion
  * Added the event ``PDN_EVENT_CTX_DESTROYED`` to indicate when a PDN connection for a PDP context is destroyed. 
    This happens when the modem is switched to minimum functionality mode (``CFUN=0``).
```

divipillai (approved):

lib/pdn/pdn.c
```diff
@@ -692,6 +698,17 @@ static void on_cfun(int mode, void *ctx)
 	}
 
 	if (mode == MODEM_CFUN_POWER_OFF) {
+		k_mutex_lock(&list_mutex, K_FOREVER);
+		SYS_SLIST_FOR_EACH_CONTAINER_SAFE(&pdn_contexts, pdn, tmp, node) {
+			if (pdn->callback) {
+				pdn->callback(pdn->context_id, PDN_EVENT_CTX_DESTROYED, 0);
+			}
+
+			sys_slist_find_and_remove(&pdn_contexts, &pdn->node);
```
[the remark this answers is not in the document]
eivindj-nordic: Thanks for your input, I have updated the `pdn_on_modem_cfun` to only call `k_free()` only if the context is removed using `pdn_ctx_free()`. Also did some more cleanup on the usage of `sys_slist_find_and_remove()`.

lucasvdp (approved):
Maybe add the new events to MOSH and Asset tracker.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -659,16 +659,18 @@ Modem libraries
...
 * :ref:`modem_info_readme` library:
 
   * Fixed a potential issue with scanf in the :c:func:`modem_info_get_current_band` function, which could lead to memory corruption.
 
-* :ref:`nrf_modem_lib_readme` library:
+* :ref:`pdn_readme` library:
 
-  * Updated the RTT trace backend to allocate the RTT channel at boot, instead of when the modem is activated.
-  * Removed support for deprecated RAI socket options ``SO_RAI_LAST``, ``SO_RAI_NO_DATA``, ``SO_RAI_ONE_RESP``, ``SO_RAI_ONGOING``, and ``SO_RAI_WAIT_MORE``.
+  * Added the event ``PDN_EVENT_CTX_DESTROYED`` to indicate when a PDN connection for a PDP context is destroyed.
```
tokangas, line 672: If I understand this correctly, the purpose of the new event is to indicate, that the PDP context (i.e. the configuration) has been destroyed/removed. Naturally, all PDN connections are always disconnected when modem is switched to flight mode (`+CFUN=4`) or minimal functionality mode (`+CFUN=0`). The text should only talk about the context, not the connection, I think.

lib/pdn/pdn.c
```diff
@@ -692,6 +709,21 @@ static void on_cfun(int mode, void *ctx)
 	}
 
 	if (mode == MODEM_CFUN_POWER_OFF) {
+		k_mutex_lock(&list_mutex, K_FOREVER);
+		SYS_SLIST_FOR_EACH_CONTAINER_SAFE(&pdn_contexts, pdn, tmp, node) {
+			if (pdn->context_id == 0) {
+				/* Default context is not destroyed */
```
tokangas, line 715: When modem is switched to minimum functionality mode (`+CFUN=0`), also the configuration for the default context is lost. So, I'm not sure this is the right way to handle default context in the library.

Example:

```
mosh:~$ at at+cgdcont=0,"IP","foobar"
OK

mosh:~$ at at+cgdcont?
+CGDCONT: 0,"IP","foobar","",0,0
OK

mosh:~$ at at+cfun=0
OK

mosh:~$ at at+cgdcont?
OK
```

tokangas (changes requested):
Changing review to "request changes" to prevent PR from being merged prematurely.

lib/pdn/pdn.c
lucasvdp: I think it is not intended for the user to change the default APN themselves, we have `CONFIG_PDN_DEFAULTS_OVERRIDE` for that. When using this config the configuration is automatically re-applied on CFUN=0. So it would make sense to still have the callback.

When would you suggest the user to set the callback again if this exception is not made? We cannot do it (without extra changes) in the `PDN_EVENT_CTX_DESTROYED` event because we are in the loop deleting contexts. We cannot use our own CFUN=0 hook as there is no way to tell in which order they will be handled.

tokangas: I'm not that familiar with the PDN library and I did not remember that the default context is handled differently. I think you're right and the current behavior is correct.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
eivindj-nordic: Thanks, I'll update the release note.

tokangas (approved):

bjda (approved):

PR #15412, opened 2024-05-21
FMDN upstream part 6: locator tag sample documentation and links relevant for the FMDN extension

kapi-no:
Based on the code additions done in part 5 and PRs with the former parts:

https://github.com/nrfconnect/sdk-nrf/pull/15370

PR #15643, opened 2024-06-03
FMDN upstream part 7: DULT documentation

kapi-no:
Follow-up upstreaming activity for the documentation content after the following PR (sample Readme):

https://github.com/nrfconnect/sdk-nrf/pull/15412

### DIFF
--- a/doc/nrf/external_comp/dult.rst
+++ b/doc/nrf/external_comp/dult.rst
@@ -1,0 +1,289 @@
+.. _ug_dult:
+
+Detecting Unwanted Location Trackers (DULT) integration
+#######################################################
+
+.. contents::
+   :local:
+   :depth: 2
+
+The Detecting Unwanted Location Trackers (DULT) is a specification that lists a set of best practices and protocols for manufacturers of products with built-in location tracking capabilities.
+Following the specification improves the privacy and safety of individuals by preventing the location tracking products from tracking users without their knowledge or consent.
+For detailed information about supported functionalities, see the official `DULT`_ documentation.
+
+.. note::
+   The DULT support in the |NCS| is :ref:`experimental <software_maturity>`.
+   The optional features are not fully supported.
+
+Integration prerequisites
+*************************
+
+Before you start the |NCS| integration with DULT, make sure that the following prerequisites are fulfilled:
+
+* :ref:`Install the nRF Connect SDK <installation>`.
+* Set up a supported :term:`Development Kit (DK)`.
+  See :ref:`device_guides` for more information on setting up the DK you are using.
+
+Solution architecture
+*********************
+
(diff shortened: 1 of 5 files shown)


mkapala-nordic (approved):

kapi-no:
> There is a problem with how the API header renders in the documentation: https://developer.nordicsemi.com/nRF_Connect_SDK_dev/doc/PR-15643/nrf/libraries/others/dult.html#api-documentation. Some enums and define lacks the description. Also some of the struct members are not shown. @divipillai, do you know why this happens? I looked into `dult.h` file and everything looks good in this file,

Could it be because we don't use the @brief syntax?

kapi-no:
> > There is a problem with how the API header renders in the documentation: https://developer.nordicsemi.com/nRF_Connect_SDK_dev/doc/PR-15643/nrf/libraries/others/dult.html#api-documentation. Some enums and define lacks the description. Also some of the struct members are not shown. @divipillai, do you know why this happens? I looked into `dult.h` file and everything looks good in this file,
> 
> Could it be because we don't use the @brief syntax?

Okay, I think this is because we use `/* xxxx` instead of `/** xxx` when documenting defines and struct fields

divipillai:
> I pushed an update as an attempt to fix the incorrect oxygen documentation

This should fix the issue. Checked the local build, and works as expected.

kapi-no:
> > I pushed an update as an attempt to fix the incorrect oxygen documentation
> 
> This should fix the issue. Checked the local build, and works as expected.

Yes, API docs seem to be rendered correctly now :)

greg-fer (commented):
Doc structure looks OK. Please consider adding a changelog entry.
Leaving the detailed doc review to @divipillai .

alstrzebonski (approved):

kapi-no:
> Doc structure looks OK. Please consider adding a changelog entry. Leaving the detailed doc review to @divipillai .

We agreed with @divipillai and @annwoj to add a changelog at the end of the upstreaming process. As you can see this is part 7 and we haven't documented any changelog entries since part 1. Would that be okay with you?

We have even bigger documentation PR (part 8) after this one as we want to squeeze in the FMDN support before the NCS feature freeze.

kapi-no:
Depends on the following PR:

https://github.com/nrfconnect/sdk-nrf/pull/15662

divipillai (approved):

kapi-no:
> ~Depends on the following PR:~
> 
> #15662

Removing the dependency as fixing the FMDN (Bluetooth) header causes the Bluetooth integration test to fail. This test suite is broken currently so I am just bringing the DULT header fixes as part of this PR and removing the dependency on the mentioned PR.

PR #15674, opened 2024-06-03
shields: nrf54h20dk: Rejig the memory layout

krish2718:
To avoid removing too many nodes causing compatibility issues, for now, remove CPU Radio related stuff to make room for Wi-Fi, this gives use 768K of Flash from MRAM10 and 436K of RAM from RAM0.

Of course any sample that uses CPU Radio (network) will not work with Wi-Fi shield.

Fixes SHEL-2833.

@hubertmis  after internal discussion with @tomchy I have rejigged the layout by just removing CPU Radio support (affects Wi-Fi BLE combo). @ArekBalysNordic  hopefully this is a better version than earlier and Matter can override, please confirm.


hubertmis:
LGTM, as long as it works for @ArekBalysNordic to redefine Radio partitions in Matter overlays.

ArekBalysNordic:
> LGTM, as long as it works for @ArekBalysNordic to redefine Radio partitions in Matter overlays.

Yes, I'm going to do it in Matter samples.

ArekBalysNordic (changes requested):
Oh, I was too fast with approving :/ I checked and I cannot use this shield if you remove the radio partition. It is because I need to have the same overlay for Thread and WiFi so in both cases I need to use Radio partition. In the case of Thread, I don't have any shield, so I need to remove and define the radio partition, but if I use your shield dts file, I cannot remove the radio partition, because it doesn't exist anymore. So I would suggest redefining the radio core partition size to for example 1kB instead of removing it at all. It would be a good compromise. Moreover what about your `wifi_coex` sample, where you use radio core? Do you plan to support it on nRF54h20?

krish2718:
> So I would suggest redefining the radio core partition size to for example 1kB instead of removing it at all. It would be a good compromise.

I did that first, but then I decided no to support BLE + Wi-Fi, anyways, I can do that.

> Moreover what about your `wifi_coex` sample, where you use radio core? Do you plan to support it on nRF54h20?

As I said in the commit log, due to the memory layout limitations, Wi-Fi + Radio won't work with this change, so, Wi-Fi + BLE combo builds aren't supported in 54H.

sachinthegreen (approved):

boards/shields/nrf700x_nrf54h20dk/nrf700x_nrf54h20dk.overlay
```diff
@@ -127,8 +28,67 @@
 		zephyr,wifi = &nordic_wlan0;
 		zephyr,bt-hci-ipc = "";
```
ArekBalysNordic, line 29: Could you please remove it also? It removes my choice of bt-hci-ipc in the sample application, and I cannot overwrite it :)

hubertmis (approved):

PR #15687, opened 2024-06-04
applications: nrf5340_audio: Sysbuild support for DFU.

andvib:
Use sysbuild kconfigs to include modules needed for DFU support. By setting the sysbuild kconfig SB_CONFIG_AUDIO_DFU_EXTERNAL_FLASH the build system will include all images and configurations needed to support DFu using external flash.

OCT-3007


applications/nrf5340_audio/Kconfig.sysbuild
```diff
@@ -4,7 +4,37 @@
...
+	  Use external flash to store the images downloaded during DFU. This requires the external
+	  flash shield with the MX25R6435F on the SPI bus.
+
+config AUDIO_DFU_INTERNAL_FLASH
+	bool "Audio Application DFU using internal flash"
+	default n
+	help
+	  Use the internal flash to store the images downloaded during DFU. This mode only support
+	  single image mode, i.e. image for application and network must be upgraded separately.
+
```
alexsven, line 23: Let's discuss this

applications/nrf5340_audio/src/bluetooth/bt_management/bt_mgmt.c
gWacey, line 1: Can you add the comment:

/* CONFIG_AUDIO_DFU_ENABLE */

To line 444.

applications/nrf5340_audio/dfu/conf/overlay-mcuboot-ext-dfu.conf
```diff
@@ -4,38 +4,30 @@
...
-CONFIG_PCD_APP=y
-CONFIG_PM_PARTITION_SIZE_MCUBOOT=0x10000
 CONFIG_BOOT_MAX_IMG_SECTORS=2048
 
-# Enable serial debug for MCUBOOT
-CONFIG_CBPRINTF_NANO=y
-CONFIG_CONSOLE=y
-CONFIG_LOG=y
-CONFIG_UART_NRFX=y
+# Use minimal C library instead of the Picolib
```
gWacey, line 9: Nit-pick: Full stop at end of comment

gWacey (changes requested):
May need to talk me through some of this!

applications/nrf5340_audio/Kconfig.sysbuild
```diff
@@ -4,7 +4,37 @@
...
 #
 
-source "${ZEPHYR_BASE}/share/sysbuild/Kconfig"
-
 config NRF_DEFAULT_IPC_RADIO
 	default y
+
+config AUDIO_DFU_EXTERNAL_FLASH
+	bool "Audio Application DFU using external flash"
+	default n
```
koffes, line 12: This should be default y? And, there must be a big fat warning if setting this to n.

applications/nrf5340_audio/Kconfig.sysbuild
```diff
@@ -4,7 +4,37 @@
...
+	help
+	  Use external flash to store the images downloaded during DFU. This requires the external
+	  flash shield with the MX25R6435F on the SPI bus.
+
+config AUDIO_DFU_INTERNAL_FLASH
+	bool "Audio Application DFU using internal flash"
+	default n
+	help
+	  Use the internal flash to store the images downloaded during DFU. This mode only support
```
koffes, line 21: We need to add more warnings in caps that this may break NET<->APP comms.

applications/nrf5340_audio/Kconfig.sysbuild
```diff
@@ -4,7 +4,37 @@
...
+config AUDIO_DFU_EXTERNAL_FLASH
+	bool "Audio Application DFU using external flash"
+	default n
+	help
+	  Use external flash to store the images downloaded during DFU. This requires the external
+	  flash shield with the MX25R6435F on the SPI bus.
+
+config AUDIO_DFU_INTERNAL_FLASH
+	bool "Audio Application DFU using internal flash"
+	default n
```
koffes, line 19: Should there be some `depends on` here? Both of these cannot be set to "y" at the same time.

applications/nrf5340_audio/Kconfig.sysbuild
```diff
@@ -4,7 +4,37 @@
...
+	default BOOTLOADER_MCUBOOT if AUDIO_DFU_EXTERNAL_FLASH || AUDIO_DFU_INTERNAL_FLASH
+endchoice
+
+config SECURE_BOOT
+	default y if AUDIO_DFU_EXTERNAL_FLASH || AUDIO_DFU_INTERNAL_FLASH
+
+config SECURE_BOOT_NETCORE
+	default y if AUDIO_DFU_EXTERNAL_FLASH || AUDIO_DFU_INTERNAL_FLASH
+
+config NETCORE_APP_UPDATE
```
koffes, line 34: PLease add a note here as well about APP<->NET possibly being broken if this is not set.

applications/nrf5340_audio/dfu/conf/overlay-dfu_external_flash.overlay
```diff
@@ -36,6 +30,12 @@
 		t-exit-dpd = <5000>;
 		dpd-wakeup-sequence = <30000 20 45000>;
 	};
+	sdhc0: sdhc@1 {
```
koffes, line 33: Nice catch. This was changed upstream.

koffes: But why do we need to have this here? THis has nothing to do with ext flash?

applications/nrf5340_audio/dfu/conf/overlay-dfu_external_flash.overlay
```diff
@@ -9,18 +9,12 @@
...
-	};
-	cs47l63: cs47l63@1 {
-		status = "okay";
-	};
-	mx25r64: mx25r6435f@2 {
+	mx25r64: mx25r6435f@0 {
 
 		compatible = "jedec,spi-nor";
-		reg = <2>;
+		reg = <0>;
```
koffes, line 17: What happened here?

applications/nrf5340_audio/dfu/conf/overlay-dfu_external_flash.overlay
```diff
@@ -9,18 +9,12 @@
...
+	cs-gpios = <&gpio1 10 GPIO_ACTIVE_LOW>, <&gpio0 11 GPIO_ACTIVE_LOW>, <&gpio0 17 GPIO_ACTIVE_LOW>;
 	status = "okay";
-	sdhc0: sdhc@0 {
-		status = "okay";
-	};
-	cs47l63: cs47l63@1 {
-		status = "okay";
-	};
-	mx25r64: mx25r6435f@2 {
+	mx25r64: mx25r6435f@0 {
```
koffes, line 14: Discuss. Will this impact backwards compatibility?

applications/nrf5340_audio/dfu/conf/overlay-dfu_external_flash.overlay
```diff
@@ -9,18 +9,12 @@
 };
 
 &spi4 {
-	cs-gpios = <&gpio0 11 GPIO_ACTIVE_LOW>, <&gpio0 17 GPIO_ACTIVE_LOW>, < &gpio1 10 GPIO_ACTIVE_LOW >;
+	cs-gpios = <&gpio1 10 GPIO_ACTIVE_LOW>, <&gpio0 11 GPIO_ACTIVE_LOW>, <&gpio0 17 GPIO_ACTIVE_LOW>;
```
koffes, line 12: Discuss.

applications/nrf5340_audio/dfu/conf/overlay-mcuboot-ext-dfu.conf
```diff
@@ -4,38 +4,30 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
```
koffes, line 4: We need to go through this file, step-by-step.

applications/nrf5340_audio/sysbuild.cmake
```diff
@@ -0,0 +1,19 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+if(SB_CONFIG_AUDIO_DFU_EXTERNAL_FLASH)
+	add_overlay_dts(nrf5340_audio
```
koffes, line 7: Please enlighten me on why this needs to be done like this.

applications/nrf5340_audio/sysbuild/mcuboot/prj.conf
```diff
@@ -7,9 +7,11 @@
...
 CONFIG_MBEDTLS_CFG_FILE="mcuboot-mbedtls-cfg.h"
-CONFIG_PCD_APP=y
 
 # Enable serial debug for MCUBOOT
 CONFIG_CBPRINTF_NANO=y
 CONFIG_LOG=y
 CONFIG_UART_NRFX=y
+
+CONFIG_PCD_APP=y
+CONFIG_BOOT_UPGRADE_ONLY=y
```
koffes, line 17: Why is this here?

koffes (changes requested):

applications/nrf5340_audio/Kconfig.sysbuild
andvib: This doesn't select between internal or external, it simply enables external. So if this is set to yes it will include MCUboot and B0N, and require an external flash shield, which is not the intention.

applications/nrf5340_audio/dfu/conf/overlay-dfu_external_flash.overlay
andvib: This change is needed as the external flash previously was listed as unit #2 on the bus, while this is now occupied by the HW codec. The change upstream was to reserve the #0 slot for an Arduino shield connection, which the external flash shield is. This change only relates to the order of the cs-pins in the DTS file.

applications/nrf5340_audio/dfu/conf/overlay-dfu_external_flash.overlay
andvib: This value has to be the same slot as the node name says (in this case `mx25r6435f@0`), otherwise a warning will be displayed. Not sure why the `reg` property is needed in addition to the node tag

applications/nrf5340_audio/sysbuild/mcuboot/prj.conf
andvib: It shouldn't :) I forgot to move it before push, will move to `overlay-mcuboot-ext-dfu.conf` so they are only applied when  external flash DFU is enabled

applications/nrf5340_audio/dfu/conf/overlay-dfu_external_flash.overlay
andvib: I tested not having this here, and that will cause the chip-select pin defined in the board file to be used - which is wrong for the flash shield making the application fail. Adding only the flash shield CS will overwrite the existing pins, causing the SD reader and HW codec to not work. So we need to set all three CS pins in this file

applications/nrf5340_audio/Kconfig.sysbuild
```diff
@@ -37,4 +37,7 @@ config NETCORE_APP_UPDATE
 config PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY
 	default y if AUDIO_DFU_EXTERNAL_FLASH
 
+config PM_OVERRIDE_EXTERNAL_DRIVER_CHECK
```
koffes, line 33: Would appreciate an explanation.

koffes (changes requested):

alexsven (approved):

koffes (approved):

PR #15716, opened 2024-06-05
applications: Matter Bridge: avoid runtime crash caused by bad alloc

markaj-nordic:
When adding multiple devices to the bridge it may happen that we run out of heap. In such case the allocation of mDataVersion may fail and we end up with nullptr that is later passed to the Span causing its constructor to abort().
Avoid abort() by adding a sanity check and more error handing to the BridgeManager.

BTW. It doesn't seem to be a particularly good design that we allocate objects in MatterBridgedDevice c'tors.
As a follow up I suggest to either use static memory for mDataVersion, similarly to the mDeviceTypeList,
or consider adding Init() methods that would pose errors in case of lack of memory.


kkasperczyk-no (approved):
Good job! I agree with the comment on the follow-up. Would you mind creating a JIRA ticket just to not forget about it?

ArekBalysNordic (approved):

LuDuda (approved):

PR #15720, opened 2024-06-05
logging: add nRF RPC logging backend

Damian-Nordic:
Add logging backend that sends logs as nRF RPC events to the log forwarder. The log forwarder receives nRF RPC log events and puts the log messages into the Zephyr logging subsystem.

The logging backend filters out logs coming from nRF RPC to avoid the avalanche of logs generated while sending already generated logs.


subsys/logging/log_rpc.c
```diff
@@ -0,0 +1,225 @@
...
+
+#ifdef CONFIG_NRF_RPC_IPC_SERVICE
+NRF_RPC_IPC_TRANSPORT(log_rpc_tr, DEVICE_DT_GET(DT_NODELABEL(ipc0)), "log_rpc_ept");
+#elif defined(CONFIG_NRF_RPC_UART_TRANSPORT)
+#define log_rpc_tr NRF_RPC_UART_TRANSPORT(DT_NODELABEL(uart1))
+#endif
+NRF_RPC_GROUP_DEFINE(log_rpc_group, "log", &log_rpc_tr, NULL, NULL, NULL);
+
+enum log_rpc_evt_forwarder {
+	LOG_RPC_EVT_MSG = 0,
```
alxelax, line 32: Enum initializing to zero is a bit useless since it starts from zero according to C language.

subsys/logging/log_rpc.c
```diff
@@ -0,0 +1,225 @@
...
+
+enum log_rpc_cmd_backend {
+	LOG_RPC_CMD_GET_CRASH_LOG = 0,
+};
+
+#ifdef CONFIG_LOG_FORWARDER_RPC
+
+LOG_MODULE_REGISTER(rpc);
+
+void log_rpc_msg_handler(const struct nrf_rpc_group *group, struct nrf_rpc_cbor_ctx *ctx,
```
alxelax, line 43:
```suggestion
static void log_rpc_msg_handler(const struct nrf_rpc_group *group, struct nrf_rpc_cbor_ctx *ctx,
```

subsys/logging/log_rpc.c
```diff
@@ -0,0 +1,225 @@
...
+
+/*
+ * Drop messages coming from nRF RPC to avoid the avalanche of logs:
+ * 1. log added
+ * 2. log sent over nRF RPC
+ * 3. more logs generated by nRF RPC
+ * 4. more logs sent over nRF RPC
+ * ...
+ */
+const char *const filtered_out_sources[] = {
```
alxelax, line 94:
```suggestion
static const char *const filtered_out_sources[] = {
```

subsys/logging/log_rpc.c
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	nrf_rpc_cbor_evt_no_err(&log_rpc_group, LOG_RPC_EVT_MSG, &ctx);
+
+	return length;
+}
+
+LOG_OUTPUT_DEFINE(log_output_rpc, send, tx_buffer, sizeof(tx_buffer));
+
+static inline const char *log_msg_source_name_get(struct log_msg *msg)
```
alxelax, line 133: Adding `inline` qualifier to the static functions(in C file) doesn't have sense since compiler will do it or won't according to optimization level and ability to do it.
Up to you to leave it or not.

subsys/logging/log_rpc.c
```diff
@@ -0,0 +1,225 @@
...
+static int format_set(const struct log_backend *const backend, uint32_t log_type)
+{
+	ARG_UNUSED(backend);
+
+	log_format = log_type;
+
+	return 0;
+}
+
+const struct log_backend_api log_backend_rpc_api = {
```
alxelax, line 215:
```suggestion
static const struct log_backend_api log_backend_rpc_api = {
```

alxelax (commented):
LGTM just minor comments. 
How about code ownership?

subsys/logging/log_rpc.c
Damian-Nordic: Right, though for things like message IDs it's useful to have all values assigned explicitly for faster lookups (in the logs you only have the numeric representation available).

subsys/logging/log_rpc.c
Damian-Nordic: Yes, you're right.

subsys/logging/log_rpc.c
```diff
@@ -0,0 +1,225 @@
...
+
+static int send(uint8_t *data, size_t length, void *context)
+{
+	struct nrf_rpc_cbor_ctx ctx;
+	const struct zcbor_string data_string = {
+		.value = data,
+		.len = length,
+	};
+
+	NRF_RPC_CBOR_ALLOC(&log_rpc_group, ctx, 4 + length);
```
alxelax, line 114: Just curious. Why `4 + length`?

alxelax (approved):
LGTM, but I suggest keeping all functionality related to rpc in one place. This logging is located in common place that creates illusion it is general logging but this is not the truth.

subsys/logging/log_rpc.c
Damian-Nordic: 1 for log level (CBOR small uint)
3 for log string (1 for CBOR major, 2 for length)

But I agree, we need some sort of helper for that as current CBOR API is way too low level to be readable.

Damian-Nordic:
> LGTM, but I suggest keeping all functionality related to rpc in one place. This logging is located in common place that creates illusion it is general logging but this is not the truth.

Well, nothing prevents using `subsys/logging` for logging other than over RPC. I tried to follow the existing pattern where RPC functions belong to particular subsystems, but if there's an agreement to put all RPC functions into a single directory, I would welcome it warmly :)
I just thought it would be strange to put just logging RPC in the RPC directory but keep all BLE, OT, CoEx etc. RPC in their respective subsystems.

canisLupus1313 (approved):

CODEOWNERS
```diff
@@ -262,6 +262,7 @@ Kconfig*                                  @tejlmand
 /subsys/event_manager_proxy/              @rakons
 /subsys/fw_info/                          @hakonfam
 /subsys/gazell/                           @leewkb4567
+/subsys/logging/                          @Damian-Nordic
```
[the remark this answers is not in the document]
Damian-Nordic: ok, I'll add you in the next PR :)

PR #15723, opened 2024-06-05
samples: cellular: modem_shell: sock: Fix 'packet' socket connect

trantanen:
PR #13772 added an assert requiring family to be 'inet' or 'inet6' when port is set into address structure.
This must be removed for 'packet' (raw) sockets to work.

Jira: MOSH-588

### DIFF
--- a/samples/cellular/modem_shell/src/sock/sock.c
+++ b/samples/cellular/modem_shell/src/sock/sock.c
@@ -566,12 +566,16 @@ int sock_open_and_connect(
 	}
 
 	/* Set port to address info */
-	if (family == AF_INET) {
-		((struct sockaddr_in *)socket_info->addrinfo->ai_addr)->sin_port = htons(port);
-	} else if (family == AF_INET6) {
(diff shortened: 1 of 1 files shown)


MirkoCovizzi (approved):

samples/cellular/modem_shell/src/sock/sock.c
```diff
@@ -570,8 +570,6 @@ int sock_open_and_connect(
 		((struct sockaddr_in *)socket_info->addrinfo->ai_addr)->sin_port = htons(port);
 	} else if (family == AF_INET6) {
 		((struct sockaddr_in6 *)socket_info->addrinfo->ai_addr)->sin6_port = htons(port);
-	} else {
-		assert(0);
```
trantanen, line 574: By removing this assert we still have a crash if `sock connect -t raw` is given as family will be `inet` and addr_info NULL pointer accessed.

I'll just check that there is address and keep the assert.

jhirsi (approved):

PR #15724, opened 2024-06-05
bl_storage: Change key invalidation to be more strict

oyvindronningstad:
Also fix the bl_validation test to put the key at the end again.

Ref: NCSDK-NONE


nvlsianpu (commented):
Way of improving the mechanism is appropriate (its introduce writing values different form 0xffff to both halfwords, so nothing else can be written later, also checking for the equal value is more strict condition).

nvlsianpu (approved):

PR #15731, opened 2024-06-05
tests: suit: Use sysbuild

jnsgsr:
Use sysbuild in SUIT tests

### DIFF
--- a/tests/subsys/suit/cache_pool_digest/testcase.yaml
+++ b/tests/subsys/suit/cache_pool_digest/testcase.yaml
@@ -1,3 +1,5 @@
+common:
+  sysbuild: true
 tests:
   suit-platform.integration.cache_pool_digest:
     platform_allow: nrf54h20dk/nrf54h20/cpuapp native_posix native_posix/native/64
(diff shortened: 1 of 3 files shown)


nordic-segl (approved):

tests/subsys/suit/cache_pool_digest/testcase.yaml
```diff
@@ -1,3 +1,5 @@
+common:
+  sysbuild: true
```
tomchy, line 2: In theory, you can limit the change to the tests that define `nrf54h20dk/nrf54h20/cpuapp` as platform, but I agree that making it consistent across all tests has some advantages (i.e. common build directory structure).

tomchy (approved):

tests/subsys/suit/cache_pool_digest/testcase.yaml
jnsgsr: There are issues to build tests projects for nrf52840. I think I will update it as you suggest and enable sysbuild only for nRF54H20 for now, and later we can get back to those issues when sysbuild will be enabled globally for all tests.

robertstypa (approved):

PR #15732, opened 2024-06-05
doc: nrf54l15 structure reorg

annwoj:
adjusted the doc structure for nRF54L15

### DIFF
--- a/doc/nrf/device_guides/nrf54l.rst
+++ b/doc/nrf/device_guides/nrf54l.rst
@@ -32,4 +32,5 @@ Zephyr and the |NCS| provide support and contain board definitions for developin
    :caption: Subpages:
 
    working_with_nrf/nrf54l/features
-   working_with_nrf/nrf54l/ug_nrf54l15_gs
+   working_with_nrf/nrf54l/nrf54l15_gs
+   working_with_nrf/nrf54l/testing_dfu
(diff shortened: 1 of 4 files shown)


kittydepa (approved):

doc/nrf/gsg_guides.rst
```diff
@@ -27,6 +26,7 @@ This is not a comprehensive section that includes all of the available getting s
    gsg_guides/nrf9160_gs
    gsg_guides/thingy91_gsg
    gsg_guides/nrf7002_gs
+   gsg_guides/nrf54l15_gs
```
greg-fer, line 29: Wait, this page is still not meeting the requirements listed in lines 11 and 12:

> These guides do not require installing the |NCS|.
They provide a guided experience using Nordic Semiconductor tools and precompiled binaries.

So putting this page here will create problems in the other direction than now: we will have a getting started guide that requires installing NCS and compiling firmware before the NCS is installed for the user and the build system is described. (Yes, the user can search for all of this, but the structure still needs to be followed.) 

I would suggest keeping the getting started guide for nRF54L15 where it is now until the full tool support for nRF54L15 is in place. We can then move the guide when it meets the requirements above.

greg-fer (changes requested):

doc/nrf/gsg_guides.rst
annwoj: Yes, you still need ncs installed first, thought we were trying to prioritize organizing the docs. 
Will reverse this change then, it should be possible to move it for the next release - I'll keep it as a task, thanks!

divipillai (approved):

greg-fer (approved):

PR #15735, opened 2024-06-05
manifest: improve west debug support for nRF54H20

gmarull:
Some extra quirks have been added so that we can e.g. add a breakpoint to main.


anangl (approved):

gmarull:
sha already ahead

PR #15741, opened 2024-06-05
sysbuild: Add overrides for nRF91-based NSIB MCUboot updates

nordicjm:
Adds an override to the number of default images when an nRF91 application is built so that the number of updateable images is not synchronised to the main application and to set the default number of images for MCUboot to 2


jorgenmk (approved):
Tested working for bootloader fota on nrf91dk.

maxd-nordic (approved):

PR #15748, opened 2024-06-05
sysbuild: Do not use ECDSA for nRF53

nordicjm:
This removes nRF53 support for using ECDSA-P256 by default for MCUboot, it seems that this was overlooked when nRF53 support was added to child/parent and it uses RSA which is not hardware acceleration, this change is needed to align support with the original thingy53 bootloader


carlescufi (approved):

nvlsianpu (approved):

anangl:
I cherry-picked the sysbuild commit into #15770 which updates the zephyr revision further.

sigvartmh (approved):
I understand the change and the rational behind it but still this may lead to some customers builds to break. Just a heads up.

nordicjm:
> I understand the change and the rational behind it but still this may lead to some customers builds to break. Just a heads up. The fix is relatively easy where support asks them to enable ECDSA in their build explicitly.

Note: this actually fixes an issue that was introduced, ECDSA-P256 was only ever defaulted on devices with a CC310, the nRF5340 has a CC312, so it has never been the default. Despite the CC312 being able to handle ECDSA-P256 and it being faster and smaller then mbedtls RSA. In a sysbuild commit I set ECDSA-P256 as the default for all devices with hardware acceleration but seemingly that was wrong because then firmware updates from child/parent image builds did not work, this restores the child/parent image firmware signing key type (which arguably, is not the right default for the hardware)

PR #15750, opened 2024-06-05
samples: matter: added support for network diagnostic logs

kkasperczyk-no:
Introduced support for Matter network diagnostic logs. Btw. conducted refactor and extracted separate module with common functionality for managing the retention memory and implementing the ring buffer.


samples/matter/common/cmake/source_common.cmake
```diff
@@ -73,5 +73,11 @@ if(CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS)
 
     if(CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_END_USER_LOGS)
         target_sources(app PRIVATE ${MATTER_COMMONS_SRC_DIR}/diagnostic/diagnostic_logs_end_user.cpp)
+        target_sources(app PRIVATE ${MATTER_COMMONS_SRC_DIR}/diagnostic/diagnostic_logs_retention.cpp)
+    endif()
+
+    if(CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_NETWORK_LOGS)
+        target_sources(app PRIVATE ${MATTER_COMMONS_SRC_DIR}/diagnostic/diagnostic_logs_network.cpp)
+        target_sources(app PRIVATE ${MATTER_COMMONS_SRC_DIR}/diagnostic/diagnostic_logs_retention.cpp)
```
ArekBalysNordic, line 81: is it a good idea to set a source twice? This will be a case when both network and user logs are enabled. I guess it can be optimized somehow here.

samples/matter/common/src/diagnostic/diagnostic_logs_end_user.cpp
```diff
@@ -8,174 +8,33 @@
 
 #include <lib/support/CodeUtils.h>
 #include <system/SystemError.h>
```
ArekBalysNordic, line 10: Redundant now?

samples/matter/common/src/diagnostic/diagnostic_logs_network.cpp
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include "diagnostic_logs_network.h"
+
+#include <lib/support/CodeUtils.h>
+#include <system/SystemError.h>
```
ArekBalysNordic, line 10: Redundant?

samples/matter/common/src/diagnostic/diagnostic_logs_network.h
```diff
@@ -0,0 +1,64 @@
...
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#pragma once
+
+#include "diagnostic_logs_provider.h"
+#include "diagnostic_logs_retention.h"
+#include <lib/core/CHIPError.h>
+
+class DiagnosticLogsNetworkReader : public Nrf::Matter::DiagnosticLogsIntentIface {
```
ArekBalysNordic, line 13: Hmm, I know that we talked about it but now I'm looking at this PR and it seems that both User Logs and Network Logs are exactly the same, but only have different constructor inputs. So why don't we resign with singleton here and allow the creation of multiple objects? It seems more proper in this case.

ArekBalysNordic: I guess the reader also does the same things but on a different RAM space.

ArekBalysNordic: Should the diagnostic logs provider contain those two objects?

wiba-nordic (approved):

samples/matter/common/cmake/source_common.cmake
kkasperczyk-no: Yeah, sure, I just missed it. Fixed.

shanthanordic:
removed milestone as no approvals yet

shanthanordic:
removed milestone as no approvals yet  and has a DNM

ArekBalysNordic (approved):

LuDuda (approved):
LGTM :+1:

Seems that approves are in, and there is no dependency to other teams, so I would re-add milestone.

PR #15753, opened 2024-06-05
cmake: Update format-version of generated dfu_application.zip

MarekPieta:
Change updates `format-version` in `manifest.json` of generated `dfu_application.zip` files. The new `format-version` is necessary
because changes introduced by HW model v2 and sysbuild break backwards compatibility.

PR also updates `generate_zip.py` to allow specifying `format-version`.

Jira: NCSDK-27569

### DIFF
--- a/scripts/bootloader/generate_zip.py
+++ b/scripts/bootloader/generate_zip.py
@@ -36,6 +36,8 @@ def parse_args():
                         file.''')
     parser.add_argument('--name', required=False, help='Optional name to display to the user to help identify the '
                         'purpose of this update')
+    parser.add_argument('--format-version', required=False, type=int, default=1,
+                        help='The format-version field is used to determine manifest.json format')
     return parser.parse_known_args()
 
 
@@ -47,7 +49,7 @@ if __name__ == '__main__':
     args, info = parse_args()
 
     manifest = {
-        'format-version': 0,
+        'format-version': args.format_version,
         'time': int(time.time()),
         'files': list()
     }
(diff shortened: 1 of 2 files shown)


scripts/bootloader/generate_zip.py
```diff
@@ -52,6 +54,8 @@ def get_name(text_wrapper):
         'files': list()
     }
 
+    manifest['format-version'] = args.format_version
```
mkapala-nordic, line 57: Can't it be set in the L52 instead?

MarekPieta: Updated. Thanks

LuDuda:
@MarekPieta what does it mean that HWv2 and sysbuild break backward-compatibility in case of DFU? What stopped working and what are the implications for the customers?

mkapala-nordic (approved):

MarekPieta:
> @MarekPieta what does it mean that HWv2 and sysbuild break backward-compatibility in case of DFU? What stopped working and what are the implications for the customers?

Properties of files that are described by the `manifest.json` file from `dfu_application.zip` were modified, namely board and file names.

scripts/bootloader/generate_zip.py
```diff
@@ -36,6 +36,8 @@ def parse_args():
                         file.''')
     parser.add_argument('--name', required=False, help='Optional name to display to the user to help identify the '
                         'purpose of this update')
+    parser.add_argument('--format-version', required=False, type=int, default=0,
```
carlescufi, line 39: I think we should default to `1`, which is the current one.

MarekPieta: Discussed offline. I will set default to `1` here

carlescufi (approved):

PR #15784, opened 2024-06-06
Samples: Bluetooth: Mesh: fix common mesh adv data

m-alperen-sener:
SMP service UUID must be presented in adv data.

### DIFF
--- a/samples/bluetooth/mesh/common/smp_bt.c
+++ b/samples/bluetooth/mesh/common/smp_bt.c
@@ -34,12 +34,12 @@ static struct bt_le_adv_param adv_params = {
 
 static const struct bt_data ad[] = {
 	BT_DATA_BYTES(BT_DATA_FLAGS, (BT_LE_AD_GENERAL | BT_LE_AD_NO_BREDR)),
+	BT_DATA_BYTES(BT_DATA_UUID128_ALL, 0x84, 0xaa, 0x60, 0x74, 0x52, 0x8a, 0x8b, 0x86, 0xd3,
+		      0x4c, 0xb7, 0x1d, 0x1d, 0xdc, 0x53, 0x8d),
 	BT_DATA(BT_DATA_NAME_COMPLETE, CONFIG_BT_DEVICE_NAME, sizeof(CONFIG_BT_DEVICE_NAME) - 1),
 };
 
 static const struct bt_data sd[] = {
-	BT_DATA_BYTES(BT_DATA_UUID128_ALL, 0x84, 0xaa, 0x60, 0x74, 0x52, 0x8a, 0x8b, 0x86, 0xd3,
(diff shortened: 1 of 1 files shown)


omkar3141:
I don't think this is a good idea. You can't assume that the external client will always try to do initiate scan-response from the peripheral.

m-alperen-sener:
> I don't think this is a good idea. You can't assume that the external client will always try to do initiate scan-response from the peripheral.

Can you elaborate please, I don't understand the point.

samples/bluetooth/mesh/common/smp_bt.c
```diff
@@ -34,12 +34,11 @@ static struct bt_le_adv_param adv_params = {
 
 static const struct bt_data ad[] = {
 	BT_DATA_BYTES(BT_DATA_FLAGS, (BT_LE_AD_GENERAL | BT_LE_AD_NO_BREDR)),
-	BT_DATA(BT_DATA_NAME_COMPLETE, CONFIG_BT_DEVICE_NAME, sizeof(CONFIG_BT_DEVICE_NAME) - 1),
+	BT_DATA_BYTES(BT_DATA_UUID128_ALL, 0x84, 0xaa, 0x60, 0x74, 0x52, 0x8a, 0x8b, 0x86, 0xd3,
```
alxelax, line 37: Will it show device name in the list of adv devices? I guess passive scanning is used for this and device name in form of AD.
Could you check it?
Scan Response is available only for active scanning.

m-alperen-sener: yes that is true; I ll update the PR so in passive scanning the name also appears. Thanks for warning.

m-alperen-sener: I have reverted that part and made sure that UUID is in adv data.

alxelax (approved):

omkar3141:
Is that UUID a mesh node UUID or SMP Service UUID? If it is mesh node UUID, it does not seem right to add this as `BT_DATA_UUID128_ALL` (this is intended for services), it should be added as `BT_DATA_MANUFACTURER_DATA`.

ludvigsj:
> Is that UUID a mesh node UUID or SMP Service UUID? If it is mesh node UUID, it does not seem right to add this as `BT_DATA_UUID128_ALL` (this is intended for services), it should be added as `BT_DATA_MANUFACTURER_DATA`.

It is the service UUID for SMP, see this page https://docs.zephyrproject.org/latest/services/device_mgmt/smp_transport.html

omkar3141:
> It is the service UUID for SMP, see this page https://docs.zephyrproject.org/latest/services/device_mgmt/smp_transport.html

Then the mystery is solved!

omkar3141 (approved):
It will nice to add one liner comment there to make it tiny bit self explanatory.

PR #15794, opened 2024-06-06
snippets: bt-rpc: Align overlay with new Zephyr

tomchy:
Currently there is a dfu_partition defined between app code and settings partition.
This PR shrinks the DFU partition to fit radio settings and leave the code partitions as-is.
Additionally, the radio core settings end at 0xe1e9000, which is the final location of Secure storage area.

Ref: NCSDK-NONE


ppryga-nordic (approved):
LGTM

hubertmis (approved):

PR #15798, opened 2024-06-06
Bluetooth: Add hci_vs_sdc.h which expose wrappers VS HCI commands

rugeGerritsen:
The SDC provides vendor specific Bluetooth functionality through vendor specific HCI commands.
The APIs added in this commit handles:
 - Building the VS HCI commands
 - Sending the VS HCI command to the controller. This works both when the controller is on the same core, or if the controller is separated from the host by another HCI layer.
 - Return parameters

This removes the need for boilerplate code in applications.

Some of the vendor specific APIs take in handles.
Those can be obtained from host types using the APIs defined in `zephyr/bluetooth/hci.h`.

The API documentation is referring to the HCI command documentation to avoid duplication.

### DIFF
--- a/subsys/bluetooth/hci_vs_sdc.c
+++ b/subsys/bluetooth/hci_vs_sdc.c
@@ -1,0 +1,341 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/bluetooth/hci.h>
+#include <bluetooth/hci_vs_sdc.h>
+
+static int hci_vs_cmd_with_rsp_only(uint16_t opcode, void *rsp, size_t rsp_size)
+{
+	int err;
+	struct net_buf *rsp_buf;
+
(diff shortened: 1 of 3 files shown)


subsys/bluetooth/hci_vs_sdc.c
```diff
@@ -0,0 +1,329 @@
...
+#include <zephyr/bluetooth/hci.h>
+#include <bluetooth/hci_vs_sdc.h>
+
+static int hci_vs_cmd_with_rsp_only(uint16_t opcode, void *rsp, size_t rsp_size)
+{
+	int err;
+	struct net_buf *buf;
+	struct net_buf *rsp_buf;
+
+	buf = bt_hci_cmd_create(opcode, 0);
```
alwa-nordic, line 16: `buf` is never used, so this just leaks a buffer

alwa-nordic (changes requested):

alwa-nordic (approved):

theob-pro (approved):

KAGA164 (approved):
LGTM

PR #15802, opened 2024-06-06
samples/nrf5340/extxip_smp_svr: no sysbuild for twister

nvlsianpu:
Added sysbuild: false to the testcase in sample.yaml as the sample is expected to build using childimage.


shanthanordic (approved):

gchwier (approved):

nvlsianpu:
This was not relay needed.

PR #15807, opened 2024-06-06
doc/nrf/../nrf5340_qspi_xip: note on childimage reqirement

nvlsianpu:
Clarified that childimage must be used.

### DIFF
--- a/samples/nrf5340/extxip_smp_svr/README.rst
+++ b/samples/nrf5340/extxip_smp_svr/README.rst
@@ -43,6 +43,12 @@ Building and running
 
 .. |sample path| replace:: :file:`samples/nrf5340/extxip_smp_svr`
 .. include:: /includes/build_and_run.txt
+Because sysbuild is currently not supported, you must add the ``--no-sysbuild`` argument when building the sample:
+
+.. parsed-literal::
+   :class: highlight
+
+   west build -b *board_target* --no-sysbuild
 
 The |NCS| build system generates all essential binaries, including the application for internal and external QSPI flash, the networking stack, and the MCUboot.
 The build process involves signing all the application binaries except for MCUboot which does not require signing in this configuration.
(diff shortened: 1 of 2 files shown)


nvlsianpu:
@kittydepa @annwoj I probably did this wrong - we need to put the info that `extxip` sample must be build using childimage and that nrf5340 QSPI XIP feature is supported only by childimage buildsystem part.

doc/nrf/device_guides/working_with_nrf/nrf53/qspi_xip_guide.rst
```diff
@@ -273,6 +273,11 @@ Similarly, it is possible to relocate certain libraries, for example:
    zephyr_code_relocate(LIBRARY subsys__mgmt__mcumgr__mgmt LOCATION EXTFLASH_TEXT NOCOPY)
    zephyr_code_relocate(LIBRARY subsys__mgmt__mcumgr__mgmt LOCATION RAM_DATA)
 
+Building the project
+********************
+
+Only specific requiremet is using childimage instead of sysbuild as xip qspi sppport is not ported to sysbuild yet.
```
annwoj, line 279:
```suggestion
You must use child image when building your project, as XIP QSPI does not currently support sysbuild.
```
I'm wondering if this information should be in this section, or rather if the title of the section really covers what comes after. What do you think @kittydepa ?

samples/nrf5340/extxip_smp_svr/README.rst
```diff
@@ -43,6 +43,7 @@ Building and running
 
 .. |sample path| replace:: :file:`samples/nrf5340/extxip_smp_svr`
 .. include:: /includes/build_and_run.txt
+This sample need to by build with --no-sysbuild option.
```
annwoj, line 46:
```suggestion
Because sysbuild is currently not supported, you must add the ``--no-sysbuild`` argument when building the sample:

.. parsed-literal::
   :class: highlight

   west build -b *board_target* --no-sysbuild
```
Would it be helpful to include an example command?

kittydepa: I agree

doc/nrf/device_guides/working_with_nrf/nrf53/qspi_xip_guide.rst
kittydepa: Hm, I think it is okay here since it is the first section on "building" (?) I am fine with either though.

annwoj: Let's keep as is then :)

annwoj (approved):

PR #15828, opened 2024-06-07
doc: software_maturity: Update BLE feature support

laxiLang:
Update Bluetooth LE features support in software maturity doc.


rugeGerritsen (changes requested):
The output does not match the expectations. Only a few features are listed, and the maturity level on different platforms seems wrong

.github/test-spec.yml
```diff
@@ -1,6 +1,6 @@
 # This is the Jenkins ci variant of the .github/labler.yaml
 
-"CI-run-twister":
```
jangalda-nsc, line 3: Please do not change this name. Some CI feature relies on this naming. In case of CI issues please open ticket in helpdesk

jangalda-nsc (changes requested):

.github/test-spec.yml
laxiLang: https://ci-help.nordicsemi.no/#ticket/zoom/3679

PR #15840, opened 2024-06-10
applications: serial_lte_modem: fix the format of Carrier app data

kacperradoszewski:
The data sent in the #XCARRIER: 12 unsolicited notification is a string and so should be enclosed in double quotes.
Print the data on the same line.

### DIFF
--- a/applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
+++ b/applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
@@ -124,9 +124,10 @@ static void on_event_app_data(const lwm2m_carrier_event_t *event)
 			return;
 		}
 
-		rsp_send("\r\n#XCARRIEREVT: %u,%hhu,\"%s\",%d\r\n", event->type, app_data->type,
+		rsp_send("\r\n#XCARRIEREVT: %u,%hhu,\"%s\",%d\r\n\"", event->type, app_data->type,
 			 uri_path, ret);
 		data_send(slm_data_buf, ret);
+		rsp_send("\"");
 	} else {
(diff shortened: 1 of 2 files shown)


applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -124,9 +124,10 @@ static void on_event_app_data(const lwm2m_carrier_event_t *event)
 			return;
 		}
 
-		rsp_send("\r\n#XCARRIEREVT: %u,%hhu,\"%s\",%d\r\n", event->type, app_data->type,
```
junqingzou, line 127: Let's print the data in the second line like TCP and UDP clients

wiba-nordic (approved):

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
MarkusLassila: I think that it is being done in here. The data is send in the data_send below.

MarkusLassila (approved):

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
MarkusLassila: Ah, nvm. I see what you mean. There is no line change after the data. Yes, I agree, it would be good for this to work similarly to TCP and UDP

kacperradoszewski: Ok, I will move it back

kacperradoszewski: I believe I have now aligned it with the format for UDP and TCP.

junqingzou (approved):

PR #15844, opened 2024-06-10
applications: machine_learning: Update memory layout

maje-emb:
This change updates the memory layout to prevent FLASH overflow. 
Expands the APP slot partition and removes the DFU partition.


dchat-nordic (approved):

grochu (approved):
Let's use it as a temporary fix before the DFU integration.

PR #15847, opened 2024-06-10
nrf_rpc: Add CRC16_CCITT calculation to hdlc frame.

canisLupus1313:
Adding calculation of CRC to HDLC framer in uart transport.

### DIFF
--- a/subsys/nrf_rpc/nrf_rpc_uart.c
+++ b/subsys/nrf_rpc/nrf_rpc_uart.c
@@ -12,6 +12,7 @@
 #include <sys/_stdint.h>
 #include <zephyr/drivers/uart.h>
 #include <zephyr/sys/util.h>
+#include <zephyr/sys/byteorder.h>
 #include <zephyr/sys/crc.h>
 
 #include <zephyr/logging/log.h>
@@ -23,6 +24,8 @@ enum {
 	hdlc_char_escape = 0x7d,
 	hdlc_char_delimiter = 0x7e,
 };
+
+#define CRC_SIZE sizeof(uint16_t)
 
 static int hdlc_input_byte(struct nrf_rpc_uart *uart_tr, uint8_t byte)
 {
@@ -44,6 +47,12 @@ static int hdlc_input_byte(struct nrf_rpc_uart *uart_tr, uint8_t byte)
 
 	if (uart_tr->hdlc_state == hdlc_state_frame_found) {
 		LOG_HEXDUMP_DBG(uart_tr->frame_buffer, uart_tr->frame_len, "Received frame");
+		if (uart_tr->frame_len > CRC_SIZE) {
+			uart_tr->frame_len -= CRC_SIZE;
+		} else {
+			uart_tr->frame_len = 0;
+			uart_tr->hdlc_state = hdlc_state_unsync;
+		}
 	}
 
 	return 0;
@@ -56,11 +65,27 @@ static void work_handler(struct k_work *work)
 	uint8_t *data;
 	size_t len;
 	int ret;
+	uint16_t crc_received = 0;
+	uint16_t crc_calculated = 0;
+
 	while (!ring_buf_is_empty(&uart_tr->rx_ringbuf)) {
 		len = ring_buf_get_claim(&uart_tr->rx_ringbuf, &data, NRF_RPC_MAX_FRAME_SIZE);
 		for (size_t i = 0; i < len; i++) {
 			if (hdlc_input_byte(uart_tr, data[i]) == 0) {
+
 				if (uart_tr->hdlc_state == hdlc_state_frame_found) {
+					crc_received = sys_get_le16(uart_tr->frame_buffer +
+							      uart_tr->frame_len);
+					crc_calculated = crc16_ccitt(0xffff, uart_tr->frame_buffer,
+								     uart_tr->frame_len);
+					if (crc_calculated != crc_received) {
+						LOG_ERR("Invalid CRC, calculated %x received %x",
+							crc_calculated, crc_received);
+						uart_tr->frame_len = 0;
+						uart_tr->hdlc_state = hdlc_state_unsync;
+						return;
+					}
+
 					uart_tr->receive_callback(transport, uart_tr->frame_buffer,
 								  uart_tr->frame_len,
 								  uart_tr->receive_ctx);
@@ -171,9 +196,22 @@ static int init(const struct nrf_rpc_tr *transport, nrf_rpc_tr_receive_handler_t
 	return 0;
 }
 
+static void send_byte(const struct device *dev, uint8_t byte)
+{
+
+	if (byte == hdlc_char_delimiter || byte == hdlc_char_escape) {
+		uart_poll_out(dev, hdlc_char_escape);
+		byte ^= 0x20;
+	}
+
+	uart_poll_out(dev, byte);
+}
+
 static int send(const struct nrf_rpc_tr *transport, const uint8_t *data, size_t length)
 {
+	uint8_t crc[2];
 	LOG_HEXDUMP_DBG(data, length, "Sending frame");
+	LOG_HEXDUMP_DBG(crc, CRC_SIZE, "frame crc");
 	struct nrf_rpc_uart *uart_tr = transport->ctx;
 
 	k_sem_take(&uart_tr->uart_tx_sem, K_FOREVER);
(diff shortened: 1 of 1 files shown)


subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -56,11 +59,24 @@ static void work_handler(struct k_work *work)
...
 	int ret;
+	uint16_t crc_received = 0;
+	uint16_t crc_calculated = 0;
+
 	while (!ring_buf_is_empty(&uart_tr->rx_ringbuf)) {
 		len = ring_buf_get_claim(&uart_tr->rx_ringbuf, &data, NRF_RPC_MAX_FRAME_SIZE);
 		for (size_t i = 0; i < len; i++) {
 			if (hdlc_input_byte(uart_tr, data[i]) == 0) {
 				if (uart_tr->hdlc_state == hdlc_state_frame_found) {
+					crc_received = (uint16_t)uart_tr->frame_buffer[uart_tr->frame_len];
```
alxelax, line 70:
```suggestion
					crc_received = sys_le16_to_cpu(*(uint16_t *)(uart_tr->frame_buffer + uart_tr->frame_len));
```
Maybe to be endianness independent?

subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -171,26 +187,35 @@ static int init(const struct nrf_rpc_tr *transport, nrf_rpc_tr_receive_handler_t
...
-		uint8_t byte = data[i];
-
-		if (data[i] == hdlc_char_delimiter || data[i] == hdlc_char_escape) {
-			uart_poll_out(uart_tr->uart, hdlc_char_escape);
-			byte ^= 0x20;
-		}
-
-		uart_poll_out(uart_tr->uart, byte);
+		send_byte(uart_tr->uart, data[i]);
 	}
```
alxelax, line 215:
```
crc = sys_cpu_to_le16(crc);
```

alxelax (commented):
Should this functionality be endianness independent?

subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -56,11 +59,24 @@ static void work_handler(struct k_work *work)
...
 		len = ring_buf_get_claim(&uart_tr->rx_ringbuf, &data, NRF_RPC_MAX_FRAME_SIZE);
 		for (size_t i = 0; i < len; i++) {
 			if (hdlc_input_byte(uart_tr, data[i]) == 0) {
 				if (uart_tr->hdlc_state == hdlc_state_frame_found) {
+					crc_received = (uint16_t)uart_tr->frame_buffer[uart_tr->frame_len];
+					crc_received |= (uint16_t)uart_tr->frame_buffer[uart_tr->frame_len + 1] << 8;
+					crc_calculated = crc16_ccitt(0xffff, uart_tr->frame_buffer, uart_tr->frame_len);
+					if (crc_calculated != crc_received) {
+						LOG_DBG("Invalid crc");
+						LOG_HEXDUMP_DBG((uint8_t*)&crc_calculated, CRC_SIZE, "CRC calculated");
```
Damian-Nordic, line 75: `LOG_HEXDUMP` sounds like an overkill for two bytes :). How about a single `LOG_ERR("Invalid CRC, calculated %x received %x`.

canisLupus1313: fixed.

subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -44,6 +46,7 @@ static int hdlc_input_byte(struct nrf_rpc_uart *uart_tr, uint8_t byte)
 
 	if (uart_tr->hdlc_state == hdlc_state_frame_found) {
 		LOG_HEXDUMP_DBG(uart_tr->frame_buffer, uart_tr->frame_len, "Received frame");
+		uart_tr->frame_len -= CRC_SIZE;
```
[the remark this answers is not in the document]
canisLupus1313: fixed.

subsys/nrf_rpc/nrf_rpc_uart.c
canisLupus1313: fixed.

subsys/nrf_rpc/nrf_rpc_uart.c
canisLupus1313: I think I don't understand this comment. `data[i]` is actually a single byte. Have you meant it for 222 line?

alxelax: Some lines are shifted, but you applied fix. 👍🏻

subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -56,11 +65,28 @@ static void work_handler(struct k_work *work)
...
+
 	while (!ring_buf_is_empty(&uart_tr->rx_ringbuf)) {
 		len = ring_buf_get_claim(&uart_tr->rx_ringbuf, &data, NRF_RPC_MAX_FRAME_SIZE);
 		for (size_t i = 0; i < len; i++) {
 			if (hdlc_input_byte(uart_tr, data[i]) == 0) {
+
 				if (uart_tr->hdlc_state == hdlc_state_frame_found) {
+					crc_received = sys_le16_to_cpu(
+						*(uint16_t *)(uart_tr->frame_buffer +
+							      uart_tr->frame_len));
```
Damian-Nordic, line 79: Never cast bytes to `uint16_t` on ARM (unaligned access ;)). 
```suggestion
					crc_received = sys_get_le16(uart_tr->frame_buffer + uart_tr->frame_len));
```

alxelax (approved):

subsys/nrf_rpc/nrf_rpc_uart.c
Damian-Nordic: This sems wrong now. `CRC` is `uint16_t` value that independent of the endiannness, and then it is put in the frame in little endian, so `sys_cpu_to_le16` isn't necessary.

alxelax: Whether the value should be converted to the protocol endianness before serialization?

subsys/nrf_rpc/nrf_rpc_uart.c
alxelax: yep, my fault in suggestion :(

canisLupus1313: fixed.

subsys/nrf_rpc/nrf_rpc_uart.c
Damian-Nordic: This would be true if we sent `uint16_t` cast to `uint8_t*`. In this case, we explicitly send the least significant byte first: `crc & 0xff`, thus no bswap is needed before.

Damian-Nordic (approved):
Approving provided that `sys_cpu_to_le16` is removed :)

subsys/nrf_rpc/nrf_rpc_uart.c
alxelax: Seems, you're right.

subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -56,11 +65,27 @@ static void work_handler(struct k_work *work)
...
+	uint16_t crc_received = 0;
+	uint16_t crc_calculated = 0;
+
 	while (!ring_buf_is_empty(&uart_tr->rx_ringbuf)) {
 		len = ring_buf_get_claim(&uart_tr->rx_ringbuf, &data, NRF_RPC_MAX_FRAME_SIZE);
 		for (size_t i = 0; i < len; i++) {
 			if (hdlc_input_byte(uart_tr, data[i]) == 0) {
+
 				if (uart_tr->hdlc_state == hdlc_state_frame_found) {
+					crc_received = sys_le16_to_cpu(uart_tr->frame_buffer +
```
alxelax, line 77: Let's wait a little bit. Now it isn't correct. The pointer is rotated instead of value.
I apologize for the confusing discussion. :)

subsys/nrf_rpc/nrf_rpc_uart.c
canisLupus1313: @alxelax @Damian-Nordic I have changed the proposed functions to [this](https://docs.zephyrproject.org/latest/doxygen/html/sys_2byteorder_8h.html#a68df8f14dfc0d8715d16ac87dd4c73d2) and [this](https://docs.zephyrproject.org/latest/doxygen/html/sys_2byteorder_8h.html#a91fae25b12126c2b35d240f56866184d). According to documentation those are endianess safe (storing and getting data in Big Endian or network byte order) and are alignment safe as well.

Damian-Nordic: Any reason Big Endian instead of Little Endian is used? I can see OpenThread and PPP driver uses LE for HDLC(-lite) frames so maybe would be more intuitive to align?

Damian-Nordic (approved):
I suggest using little endian instead of big endian for FCS, but otherwise LGTM.

subsys/nrf_rpc/nrf_rpc_uart.c
canisLupus1313: When comes to networking big endian is more intuitive.

Damian-Nordic: Yes, but when it comes to HDLC, this is what RFC includes:
> Frame Check Sequence (FCS) Field

> The Frame Check Sequence field defaults to 16 bits (two octets). The FCS is transmitted least significant octet first, which contains the coefficient of the highest term.

canisLupus1313: fixed.

Damian-Nordic: But feel free to disregard this comment if you think we should diverge from HDLC at this.

Damian-Nordic (approved):

PR #15848, opened 2024-06-10
bl_storage: Change key invalidation to be more strict

oyvindronningstad:
Also fix the bl_validation test to put the key at the end and add debug keys.


nvlsianpu (approved):

subsys/bootloader/bl_storage/bl_storage.c
```diff
@@ -72,15 +73,39 @@ uint32_t num_public_keys_read(void)
...
+#define INVALID_VAL 0x50FA0000
+#define INVALID_WRITE_VAL 0xFFFF0000
+#define VALID_VAL 0x50FAFFFF
+#define MAX_NUM_PUBLIC_KEYS 16
+#define TOKEN_IDX_OFFSET 24
 
 static bool key_is_valid(uint32_t key_idx)
 {
-	return nrfx_nvmc_uicr_word_read(&BL_STORAGE->key_data[key_idx].valid) != INVALID_VAL;
+	if (key_idx > num_public_keys_read()) {
```
pdunaj, line 84: shouldn't it be `>=`?

pdunaj (changes requested):

subsys/bootloader/bl_storage/bl_storage.c
oyvindronningstad: You're right. Fixed.

subsys/bootloader/bl_storage/bl_storage.c
```diff
@@ -72,15 +73,39 @@ uint32_t num_public_keys_read(void)
...
+	}
+
+	/* Invalid value. */
+	k_panic();
+	return false;
 }
 
 int verify_public_keys(void)
 {
+	if (num_public_keys_read() >= MAX_NUM_PUBLIC_KEYS) {
```
pdunaj, line 105: This was correct in the earlier version.
In this case it should be simply `>`. This is a number of keys that is checked here not an index (like in line 84).

pdunaj (changes requested):

nvlsianpu:
Verified that patch is working correctly.
Need to apply Pawel's remark.

nvlsianpu (approved):

nvlsianpu (changes requested):
as above

pdunaj (approved):

nvlsianpu (approved):

subsys/bootloader/bl_storage/bl_storage.c
```diff
@@ -72,15 +72,39 @@ uint32_t num_public_keys_read(void)
...
+	const uint32_t valid_val = VALID_VAL | key_idx << TOKEN_IDX_OFFSET;
+
+	if (token == invalid_val) {
+		return false;
+	} else if (token == valid_val) {
+		return true;
+	}
+
+	/* Invalid value. */
+	NVIC_SystemReset();
```
pdunaj, line 98: I don't understand this change. Can you tell why it was needed?

I expect it should use a system oops or panic, which in release version should reboot. If it does not we have a problem in the entire system.
BTW - how are other fatal errors handled in NSIB?

pdunaj (changes requested):

pdunaj:
> nrf54h20[nrf54h20dk/nrf54h20/cpuapp

Hi @miha-nordic , this change modifies NSIB but log in memory consumption increase shows regression in builds that do not use it (based on SUIT and MCUBoot). Is it some false positive or an issue in check itself?

pdunaj (approved):

pdunaj:
@shanthanordic , this can go in

PR #15855, opened 2024-06-10
applications: nrf5340_audio: Speed up discovery process

alexsven:
- Set ACL interval to 10ms during discovery and configuration
- Set ACL back to 80ms before ISO starts streaming to free up air time
- Fixes bug where set up time is too long
- OCT-3035

### DIFF
--- a/applications/nrf5340_audio/src/bluetooth/bt_management/scanning/Kconfig
+++ b/applications/nrf5340_audio/src/bluetooth/bt_management/scanning/Kconfig
@@ -25,11 +25,18 @@ menu "Connection"
 
 config BLE_ACL_CONN_INTERVAL
 	int "Bluetooth LE ACL Connection Interval (x*1.25ms)"
-	default 32
+	default 8
 	help
-	  When the LE Audio Controller Subsystem for nRF53 is used, this
-	  interval should be a multiple of the ISO interval and maximum 4x
-	  larger than the lowest interval.
+	  This interval should be a multiple of the ISO interval used. The recommendation is to
+	  increase the interval to something like BLE_ACL_CONN_INTERVAL_SLOW after the discovery
+	  process is done, to free up time on air.
+
+config BLE_ACL_CONN_INTERVAL_SLOW
+	int "Bluetooth LE ACL Connection Interval (x*1.25ms)"
+	default 72
+	help
+	  This interval should be a multiple of the ISO interval used. 72*1.25=90 which will
+	  suit both 7.5ms and 10ms.
 
 config BLE_ACL_SLAVE_LATENCY
 	int "Bluetooth LE Slave Latency"
--- a/applications/nrf5340_audio/unicast_client/main.c
+++ b/applications/nrf5340_audio/unicast_client/main.c
@@ -269,6 +269,19 @@ static void le_audio_msg_sub_thread(void)
 			break;
 
 		case LE_AUDIO_EVT_CONFIG_RECEIVED:
+			struct bt_le_conn_param param;
+
+			/* Set the ACL interval up to allow more time for ISO packets */
+			param.interval_min = CONFIG_BLE_ACL_CONN_INTERVAL_SLOW;
+			param.interval_max = CONFIG_BLE_ACL_CONN_INTERVAL_SLOW;
+			param.latency = CONFIG_BLE_ACL_SLAVE_LATENCY;
+			param.timeout = CONFIG_BLE_ACL_SUP_TIMEOUT;
+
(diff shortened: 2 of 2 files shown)


applications/nrf5340_audio/src/bluetooth/bt_management/scanning/Kconfig
```diff
@@ -25,11 +25,16 @@ menu "Connection"
 
 config BLE_ACL_CONN_INTERVAL
 	int "Bluetooth LE ACL Connection Interval (x*1.25ms)"
-	default 32
+	default 64
 	help
-	  When the LE Audio Controller Subsystem for nRF53 is used, this
-	  interval should be a multiple of the ISO interval and maximum 4x
-	  larger than the lowest interval.
+	  This interval should be a multiple of the ISO interval.
```
koffes, line 30: Should be or must be?

koffes (approved):

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/Kconfig
alexsven: Should be, it is still valid to set it to something that is not a multiple of the ISO interval, but it is in no way advisable ;)

applications/nrf5340_audio/unicast_client/main.c
```diff
@@ -354,6 +369,20 @@ static void bt_mgmt_evt_handler(const struct zbus_channel *chan)
...
+		param.interval_min = CONFIG_BLE_ACL_CONN_INTERVAL_FAST;
+		param.interval_max = CONFIG_BLE_ACL_CONN_INTERVAL_FAST;
+		param.latency = CONFIG_BLE_ACL_SLAVE_LATENCY;
+		param.timeout = CONFIG_BLE_ACL_SUP_TIMEOUT;
+
+		ret = bt_conn_le_param_update(msg->conn, &param);
+		if (ret) {
+			LOG_WRN("Failed to update conn parameters: %d", ret);
+		}
+
```
rugeGerritsen, line 385: As a Bluetooth Central we can set the initial ACL interval when connecting to avoid running a connection update after connection setup. Should that be done instead? Or are we here trying to improve the situation where the unicast client is a Bluetooth peripheral?

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/Kconfig
```diff
@@ -25,11 +25,16 @@ menu "Connection"
 
 config BLE_ACL_CONN_INTERVAL
 	int "Bluetooth LE ACL Connection Interval (x*1.25ms)"
-	default 32
+	default 64
```
rugeGerritsen, line 28: Should the default instead be a multiple of both 7.5 ms and 10 ms to be prepared for future use cases?

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/Kconfig
```diff
@@ -25,11 +25,16 @@ menu "Connection"
...
 	help
-	  When the LE Audio Controller Subsystem for nRF53 is used, this
-	  interval should be a multiple of the ISO interval and maximum 4x
-	  larger than the lowest interval.
+	  This interval should be a multiple of the ISO interval.
+
+config BLE_ACL_CONN_INTERVAL_FAST
```
rugeGerritsen, line 32: Do we have to make BLE_ACL_CONN_INTERVAL_FAST a Kconfig entry? Can we instead use SDU interval?

rugeGerritsen (changes requested):
It would be nice if the commit message explained why the selected values were chosen

applications/nrf5340_audio/src/bluetooth/bt_management/scanning/Kconfig
alexsven: Yes, can do that.

applications/nrf5340_audio/unicast_client/main.c
alexsven: As discussed, I don't want non-ISO related ACls to fall in to the default of having 10ms, so default should be larger as it is less intrusive on air and then we increase the interval when we know that a lot of data will go over air.

alexsven: After more discussions with @koffes I ended up setting the default to 10, and then setting it to 90 later. We need to do some more thinking related to supporting both 7.5 and 10, as well as non-ISO related ACL devices.

PR #15865, opened 2024-06-10
samples: suit: smp_transfer: do not generate MPI for hci_ipc image

ahasztag:
Do not generate or merge the MPI generated by the hci_ipc image. It would result in the MPI configuration managed by the application core (main) image to be overwritten by the default configuration from hci_ipc.


robertstypa (approved):

e-rk (approved):

e-rk:
Should this be documented as a known issue or fixed for all applications? I would assume that someone could fall into this in their own application.

tomchy (approved):

PR #15868, opened 2024-06-10
modules: hostap: Fix SHA1 selection

krish2718:
Though this is enabled by default except for OT, we had seen issues with conflicting MbedTLS configuration from multiple modules. But for Wi-Fi SHA1 is a mandatory configuration.

Explicitly enabling required configurations ensures that it is always enabled.

Fixes SHEL-2875.


glarsennordic (approved):
LGTM, thanks!

VivekUppunda (approved):

sachinthegreen (approved):

PR #15875, opened 2024-06-11
samples: matter: Added diagnostic logs required by the certification test case

kkasperczyk-no:
This PR introduces log backend for redirecting logs to the diagnostic logs module. Diagnostic logs generation is required to pass the Matter certification tests.

This PR includes also small refactor of modules for diagnostic end user logs and network logs that were the same and the only difference was used memory partition. They were removed and replaced by creating two instances of DiagnosticLogsRetention objects by the DiagnosticLogsProvider.

Additionally the DiagnosticLogsRetentionReader was created.

Fixes: KRKNWK-19080


samples/matter/common/src/diagnostic/diagnostic_logs_provider.h
```diff
@@ -138,7 +131,20 @@ class DiagnosticLogProvider : public chip::app::Clusters::DiagnosticLogs::Diagno
...
+		: mNetworkLogs(DEVICE_DT_GET(DT_NODELABEL(network_logs_retention))),
+		  mEndUserLogs(DEVICE_DT_GET(DT_NODELABEL(end_user_logs_retention)))
+	{
+	}
+#elif defined(CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_NETWORK_LOGS)
+	DiagnosticLogProvider() : mNetworkLogs(DEVICE_DT_GET(DT_NODELABEL(network_logs_retention))) {}
+#elif defined(CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_END_USER_LOGS)
+	DiagnosticLogProvider() : mEndUserLogs(DEVICE_DT_GET(DT_NODELABEL(end_user_logs_retention))) {}
+#else
+	DiagnosticLogProvider() {}
```
ArekBalysNordic, line 146: super NIT.

```suggestion
	DiagnosticLogProvider() =  default;
```

ArekBalysNordic (commented):
So far it looks good.

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -263,13 +263,17 @@ The diagnostic logs snippet enables the set of configurations needed for full Ma
...
+The parition sizes are configured using example values and may not be sufficient for your use cases.
+To change the partition sizes, for example increase them to be able to store more logs, you need to change the configuration in the devicetree overlay.
 The snippet sets the following kconfig options:
 
   * :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS` to ``y``.
   * :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_CRASH_LOGS` to ``y``.
   * :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_REMOVE_CRASH_AFTER_READ` to ``y``.
   * :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_END_USER_LOGS` to ``y``.
   * :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_NETWORK_LOGS` to ``y``.
+  * :kconfig:option:`CONFIG_LOG_MODE_DEFERRED` to ``y``.
```
ArekBalysNordic, line 275: Should we have more comments on that, or do you think it is redundant?

samples/matter/common/src/diagnostic/diagnostic_logs_retention.cpp
```diff
@@ -61,6 +61,10 @@ CHIP_ERROR DiagnosticLogsRetention::PushLog(const void *data, size_t size)
 	VerifyOrReturnError(mIsInitialized, CHIP_ERROR_INCORRECT_STATE);
 	VerifyOrReturnError(size < mCapacity, CHIP_ERROR_BUFFER_TOO_SMALL);
 
+	/* Make sure that no-one will start write operation from different thread in the same time. */
+	VerifyOrReturnError(!mWriteInProgress, CHIP_ERROR_INCORRECT_STATE);
```
ArekBalysNordic, line 65: Can we use a zephyr lock (semaphore/mutex) here similarly to matter init purposes? I guess we can wait some time for the write/read operation (using mutex) instead of dropping messages due to multi-thread operations.

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -263,13 +263,17 @@ The diagnostic logs snippet enables the set of configurations needed for full Ma
 The configuration set consist of devicetree overlays for each supported target board, and a config file that enables all diagnostic logs features by default.
 The devicetree overlays add new RAM partitions which are configured as retained to keep the log data persistent and survive the device reboot.
 They also reduce the SRAM size according to the size of the retained partition.
+The parition sizes are configured using example values and may not be sufficient for your use cases.
```
wiba-nordic, line 266:
```suggestion
The partition sizes are configured using example values and may not be sufficient for all use cases.
```

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -263,13 +263,17 @@ The diagnostic logs snippet enables the set of configurations needed for full Ma
...
+To change the partition sizes, for example increase them to be able to store more logs, you need to change the configuration in the devicetree overlay.
```
wiba-nordic, line 267:
```suggestion
To change the partition sizes, you need to change the configuration in the devicetree overlay.
You can, for example, increase the partition sizes to be able to store more logs.

```
Note the empty line at the end. The paragraph is long enough now that it makes sense to separate the sentence introducing the Kconfig options from the rest of the text.

samples/matter/common/src/diagnostic/log_backend_diagnostic.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+#include <zephyr/logging/log.h>
+#include <zephyr/logging/log_backend.h>
+
+static uint32_t sLogFormat = LOG_OUTPUT_TEXT;
+/* The buffer size was selected to be able to store single log line (usually in Matter we have logs < 200 B).
+ * In case the log will not fit it will be printed anyway, but usng fragmentation, so setting a small buffer
+ * increases the time of printing.
+ */
+static uint8_t sLogBuffer[256];
```
ArekBalysNordic, line 17: Can we use here an existing kconfig instead of raw value?

wiba-nordic (changes requested):

samples/matter/common/src/diagnostic/log_backend_diagnostic.cpp
```diff
@@ -0,0 +1,90 @@
...
+static int RedirectEndUserLog(uint8_t *data, size_t length, void *context)
+{
+	Nrf::Matter::DiagnosticLogProvider::GetInstance().PushLog(
+		chip::app::Clusters::DiagnosticLogs::IntentEnum::kEndUserSupport, data, length);
+
+	return length;
+}
+
+LOG_OUTPUT_DEFINE(log_output_network, RedirectNetworkLog, sLogBuffer, sizeof(sLogBuffer));
+LOG_OUTPUT_DEFINE(log_output_end_user, RedirectEndUserLog, sLogBuffer, sizeof(sLogBuffer));
```
ArekBalysNordic, line 36: Is it ok that we use the same buffer for both log output backends?

ArekBalysNordic: I guess, it is OK because logs are added sequentially, but I wonder if there is a possibility to corrupt this buffer when it is filled by two sources.

samples/matter/common/src/diagnostic/log_backend_diagnostic.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+	const log_format_func_t logFormatter = log_format_func_t_get(sLogFormat);
+
+	/* The assumption following redirection is done:
+	 * - chip -> diagnostic network logs
+	 * - app -> diagnostic end user logs
+	 * - other -> dropped
+	 */
+#ifdef CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_NETWORK_LOGS
+	if (strncmp(sourceName, "chip", strlen("chip")) == 0) {
```
ArekBalysNordic, line 65: What about using an existing name somewhere, if exists, instead of hardcoding `chip` and `app`? Otherwise, we will need to remember to update that if someone changes it.

samples/matter/common/src/diagnostic/log_backend_diagnostic.cpp
kkasperczyk-no: Do you have in mind any specific Kconfig? Other backends just create own Kconfig to control it, but I thought maybe we don't need to provide a way to change the value.

samples/matter/common/src/diagnostic/log_backend_diagnostic.cpp
kkasperczyk-no: I don't think it's possible, because as you said these calls are used sequentially.

samples/matter/common/src/diagnostic/diagnostic_logs_retention.cpp
kkasperczyk-no: Probably yes, I will try.

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
kkasperczyk-no: I can add additional explanation, similarly to the one that is in the .conf files's comment.

samples/matter/common/src/diagnostic/log_backend_diagnostic.cpp
kkasperczyk-no: I don't know what you mean by existing name?

kkasperczyk-no:
@wiba-nordic I forgot to add some part of documentation before. I've recently added a new section about logs redirecting, please take a look on that as well.

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
kkasperczyk-no: Done.

samples/matter/common/src/diagnostic/log_backend_diagnostic.cpp
ArekBalysNordic: No, maybe there is no appropriate one. :/ So, ok let's leave it as it is.

samples/matter/common/src/diagnostic/log_backend_diagnostic.cpp
ArekBalysNordic: I mean that you compare `sourceName` to "chip". Is the "chip" string defined somewhere?

kkasperczyk-no: No, it's hardcoded in the `src/platform/Zephyr/Logging.cpp`, where we register this log module.

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -254,6 +254,28 @@ The device sends converted logs to the Matter controller as a response.
 After the crash data is successfully read, it will be removed and further read attempts will notify the user that there is no available data to read.
 To keep the crash log in the memory after reading it, set the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_REMOVE_CRASH_AFTER_READ` Kconfig option to ``n``.
 
+Network and end user logs
+-------------------------
+
+The diagnostic network and end user logs are saved in the dedicated retained RAM partitions.
+The logs are not removed after reading and in case of attempt to write new logs to the already full buffer, the oldest logs are be replaced.
+
+The diagnostic network and end user logs are designed to be pushed on the user demand.
```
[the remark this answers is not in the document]
kkasperczyk-no: It's ok, I just wanted to highlight it, because e.g. crash logs are pushed internally by the app and user can't request it.

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -254,6 +254,28 @@ The device sends converted logs to the Matter controller as a response.
...
+Usually it is not convenient to duplicate all logger calls to pass the same information, but using other API.
```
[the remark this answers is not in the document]
kkasperczyk-no: I'm not sure. I mean that one option is to redirect Zephyr logger, so we only call log and have two outputs - one printed on the console and the other saved to memory.
Other option would be to not redirect logs, so we would need to call some additional API just after log to send the same information to memory and this is not convenient (but someone may want to do it this way for some reason).

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -254,6 +254,28 @@ The device sends converted logs to the Matter controller as a response.
...
+Due to that, for the network and the end user logs the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_REDIRECT` Kconfig option is enabled by default.
+
+The redirection functionality allows to save logs passed to the Zephyr logger in the retained RAM as a Matter diagnostic logs.
```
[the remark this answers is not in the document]
kkasperczyk-no: I slightly changed it to:
With the Kconfig option enabled, the redirect functionality takes logs passed to the Zephyr logger and saves them in the retained RAM as Matter diagnostic logs.

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -254,6 +254,31 @@ The device sends converted logs to the Matter controller as a response.
...
+The diagnostic network and end user logs are saved in the dedicated retained RAM partitions.
+The logs are not removed after reading, but when attempting to write new logs to an already full buffer, the oldest logs are replaced.
+
+The diagnostic network and end user logs are designed to be pushed when requested by the user.
+Usually, it is not convenient to duplicate all logger calls to pass the same information, but using other API.
+Because of this, for the network and the end user logs the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_REDIRECT` Kconfig option is enabled by default.
+
+With the Kconfig option enabled, the redirect functionality takes logs passed to the Zephyr logger and saves them in the retained RAM as Matter diagnostic logs.
+Only the following logs are redirected:
+* Logs from the ``chip`` module are redirected into diagnostic network logs.
```
wiba-nordic, line 269:
```suggestion
Only the following logs are redirected:

* Logs from the ``chip`` module are redirected into diagnostic network logs.
```

samples/matter/common/src/diagnostic/diagnostic_logs_retention.cpp
```diff
@@ -62,6 +63,9 @@ CHIP_ERROR DiagnosticLogsRetention::PushLog(const void *data, size_t size)
 	VerifyOrReturnError(mIsInitialized, CHIP_ERROR_INCORRECT_STATE);
 	VerifyOrReturnError(size < mCapacity, CHIP_ERROR_BUFFER_TOO_SMALL);
 
+	/* Make sure that no-one will start write operation from different thread in the same time. */
+	k_mutex_lock(&sWriteMutex, K_FOREVER);
```
ArekBalysNordic, line 67: Should we wait forever, or give some time to execute, and return?

ArekBalysNordic (approved):
One thing to consider, overall looks good.

doc/nrf/protocols/matter/getting_started/advanced_kconfigs.rst
```diff
@@ -254,6 +254,31 @@ The device sends converted logs to the Matter controller as a response.
...
 To keep the crash log in the memory after reading it, set the :kconfig:option:`CONFIG_NCS_SAMPLE_MATTER_DIAGNOSTIC_LOGS_REMOVE_CRASH_AFTER_READ` Kconfig option to ``n``.
 
+Network and end user logs
+-------------------------
+
+The diagnostic network and end user logs are saved in the dedicated retained RAM partitions.
+The logs are not removed after reading, but when attempting to write new logs to an already full buffer, the oldest logs are replaced.
+
+The diagnostic network and end user logs are designed to be pushed when requested by the user.
+Usually, it is not convenient to duplicate all logger calls to pass the same information, but using other API.
```
wiba-nordic, line 264:
```suggestion
This can result in the same information being called by multiple APIs, which is usually not desirable bahavior.
```
Does this wording work?

wiba-nordic (changes requested):

samples/matter/common/src/diagnostic/diagnostic_logs_retention.cpp
kkasperczyk-no: Discussed f2f

wiba-nordic (approved):

PR #15877, opened 2024-06-11
samples: cellular: gnss: Update GNSS factory almanacs

tokangas:
Updated GNSS factory almanacs.


tokangas:
@shanthanordic Is is still possible to get updated GNSS factory almanacs into v2.7.0? I have updated those in the beginning of each month, but I was on vacation and the update was delayed a bit.

trantanen (approved):

lemrey (approved):

jhirsi (approved):

shanthanordic (approved):

PR #15881, opened 2024-06-11
manifest: add soc-hwmv1 module

gmarull:
The soc-hwmv1 module adds support for nRF52/nRF53/nRF91 SoCs in HWMv1 (Hardware Model V1), so that users can continue using the old hardware model while they transition to HWMv2.


tejlmand (approved):
we should cherry-pick https://github.com/zephyrproject-rtos/zephyr/pull/74057 into sdk-zephyr and update the manifest change in this PR accordingly.

PR #15896, opened 2024-06-11
manifest: Update zephyr with uart legacy shim for nrf54h20

nordic-krch:
Allow to use legacy uart shim for nrf54h20 (in particular PPR).


kapi-no:
The NCS CoreMark sample requires this change so that the PPR core can correctly log the benchmark results.

anangl:
Rebased and updated with SHA.

anangl (approved):

nordicjm (approved):

shanthanordic (approved):

PR #15901, opened 2024-06-11
samples: cellular: gnss: Build integration CI images with twister

moksanen:
To optimize integration CI positioning test execution times, use twister to build test images with required configurations. The images configured to be built by twister will be copied to test environment instead of re-building similar images in Jenkins jobs.

test-sdk-nrf: sdk-nrf-pr-16010

### DIFF
--- a/applications/asset_tracker_v2/sample.yaml
+++ b/applications/asset_tracker_v2/sample.yaml
@@ -364,3 +364,18 @@ tests:
                 EXTRA_CONF_FILE="overlay-nrf7002ek-wifi-scan-only.conf;overlay-debug.conf"
                 SB_CONFIG_WIFI_NRF700X=y SB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y
     tags: ci_build sysbuild
+
+  # Configuration which will be used by the positioning CI integration job to verify PRs
+  applications.asset_tracker_v2.integration_config_positioning:
+    sysbuild: true
+    build_only: true
+    extra_configs:
+      - CONFIG_MODEM_ANTENNA_GNSS_EXTERNAL=y
+      - CONFIG_LOCATION_LOG_LEVEL_DBG=y
+      - CONFIG_LOG_BUFFER_SIZE=2048
+    extra_args: asset_tracker_v2_SNIPPET="nrf91-modem-trace-uart"
+    integration_platforms:
+      - nrf9151dk/nrf9151/ns
+    platform_allow:
+      - nrf9151dk/nrf9151/ns
+    tags: ci_build sysbuild
--- a/samples/cellular/gnss/sample.yaml
+++ b/samples/cellular/gnss/sample.yaml
@@ -13,3 +13,99 @@ tests:
       - nrf9160dk/nrf9160/ns
       - nrf9161dk/nrf9161/ns
     tags: ci_build sysbuild
+
+  # Following configurations will be used by the positioning CI integration job to verify PRs
+  sample.cellular.gnss.integration_config_positioning_agnss_nrfcloud_ltem_pvt:
+    sysbuild: true
+    build_only: true
+    extra_configs:
+      - CONFIG_LTE_NETWORK_MODE_LTE_M_GPS=y
+      - CONFIG_LTE_NETWORK_MODE_NBIOT_GPS=n
+      - CONFIG_GNSS_SAMPLE_LTE_ON_DEMAND=y
+      - CONFIG_GNSS_SAMPLE_ASSISTANCE_NRF_CLOUD=y
+      - CONFIG_GNSS_SAMPLE_NMEA_ONLY=n
+      - CONFIG_GNSS_SAMPLE_MODE_PERIODIC=n
+      - CONFIG_GNSS_SAMPLE_ASSISTANCE_MINIMAL=n
+      - CONFIG_LTE_EDRX_REQ=y
+      - CONFIG_FPU=y
+      - CONFIG_GNSS_SAMPLE_REFERENCE_LATITUDE="61.49375330"
+      - CONFIG_GNSS_SAMPLE_REFERENCE_LONGITUDE="23.77588976"
+      - CONFIG_MODEM_ANTENNA_GNSS_EXTERNAL=y
+      - CONFIG_LOG_BUFFER_SIZE=2048
+    extra_args: gnss_SNIPPET="nrf91-modem-trace-uart"
+    integration_platforms:
+      - nrf9151dk/nrf9151/ns
+    platform_allow:
+      - nrf9151dk/nrf9151/ns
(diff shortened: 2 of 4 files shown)


bjda (approved):

samples/cellular/gnss/sample.yaml
```diff
@@ -13,3 +13,105 @@ tests:
...
       - nrf9161dk/nrf9161/ns
     tags: ci_build sysbuild
+  # Following configurations will be used by the positioning CI integration
+  # job to verify PRs
+  sample.cellular.gnss.integration_config_positioning_agnss_nrfcloud_ltem_pvt:
+    sysbuild: true
+    build_only: true
+    extra_configs:
+      - CONFIG_LTE_NETWORK_MODE_LTE_M_GPS=y
+      - CONFIG_LTE_NETWORK_MODE_NBIOT_GPS=n
```
trantanen, line 23: I would remove those network modes that are 'n'. This is a choice symbol so you can only have one of them on anyway.

samples/cellular/gnss/sample.yaml
```diff
@@ -13,3 +13,105 @@ tests:
       - nrf9160dk/nrf9160/ns
       - nrf9161dk/nrf9161/ns
     tags: ci_build sysbuild
+  # Following configurations will be used by the positioning CI integration
+  # job to verify PRs
```
trantanen, line 17: These would fit into one 100 character line.
And an empty line before integration config makes a nice separation from other compilation configs.

```suggestion

    # Following configurations will be used by the positioning CI integration job to verify PRs
```

trantanen (commented):
Just some nit-picking

samples/cellular/gnss/sample.yaml
moksanen: We configure all gnss sample images like this for extra clarity, so that it's obvious what the image does, and what not. I'd rather use the same format for these images, so that it's easy to see that the configurations match.

trantanen: OK, fine. I understand this in many symbols but for choice symbols it may also be confusing because setting them all 'y' doesn't really do that and you don't necessarily know which one is used.
But I'll leave this to you.

moksanen: For testers it's not always obvious which kconfigs are choice and which are not. You need to search it in source to understand how it works after it is passed on to the compiler. For me it is better to state clearly that this image _is LTE-M_ and _is not NB-IoT_.
We pass all these y/n parameters to test code as text vectors, and generate required variants automatically. It's better to keep the list of y's and n's unmodified, so that we know that images build in place A have exactly the same configuration as images built in place B.

jvaliharju (approved):

tokangas (approved):
I don't have any other comments, but as Tommi R. already said, the configuration lists would be a lot cleaner if all default values would be removed. But I understand your point and leave the choice up to you.

samples/cellular/gnss/sample.yaml
trantanen: > For testers it's not always obvious which kconfigs are choice and which are not. You need to search it in source to understand how it works after it is passed on to the compiler.

Same applies for developers. You only know it by looking at the KConfig from the source code or documentation.

> For me it is better to state clearly that this image _is LTE-M_ and _is not NB-IoT_.

I understand this but there is a danger of misconfiguration as well if you set both to 'y'. You don't basically know what will be chosen and you don't get compilation errors or warnings.

> We pass all these y/n parameters to test code as text vectors, and generate required variants automatically. It's better to keep the list of y's and n's unmodified, so that we know that images build in place A have exactly the same configuration as images built in place B.

I don't necessarily see how this would change (but I don't know how these text vectors are implemented in detail) if you would have just one line of Kconfig for a choice symbol but yes, you need to know that it's choice symbol by looking at the documentation. But this is something you should be aware of anyway to avoid setting multiple values for it.

I do understand the value of setting other (non-choice symbols) configurations even if it was a default.

trantanen (approved):

applications/asset_tracker_v2/sample.yaml
```diff
@@ -364,3 +364,18 @@ tests:
                 EXTRA_CONF_FILE="overlay-nrf7002ek-wifi-scan-only.conf;overlay-debug.conf"
                 SB_CONFIG_WIFI_NRF700X=y SB_CONFIG_WIFI_NRF700X_SCAN_ONLY=y
     tags: ci_build sysbuild
+
```
balaji-nordic, line 367: Can't the CI be modified instead? That would be cleaner I believe

balaji-srin (changes requested):

applications/asset_tracker_v2/sample.yaml
trantanen: We have CI compiling these configurations at the moment so everything is working but it would be more efficient from integration testing perspective if we could use the images from twister stage.
We are already doing that with MOSH tests. This PR targets for us to be able to do that also for positioning test set, which is using ATv2, GNSS sample and Location sample.

balaji-nordic: I see. I believe it is better to call this new configuration `locationing-ext-gnss-antenna-debug` . This will make it similar to other configurations. That way we do not unnecessarily expose details about our CI here. You can remove the comment in line 368 as this may be a useful configuration to have for customers as well. 

@jorgenmk  Will adding this new configuration have any effect on our CI plan?

balaji-srin (changes requested):

applications/asset_tracker_v2/sample.yaml
trantanen: What are the potential secret details of the CI this is revealing? I'm asking it because MOSH has it already. And I don't really see any problems sharing information that we are testing with this configuration.

There is a benefit on the naming and comment indicating that this configuration is used for testing purposes. People don't break it as easily and are more careful in modifying it.

I don't see how this would affect on other CI plans than the ones that intend to use it.

balaji-nordic: I agree that there are no secrets being revealed. But it just makes the sample.yaml clumsy. I still prefer that you that change the name to keep it more generic. But I wont insist further. Will approve.

balaji-srin (approved):

applications/asset_tracker_v2/sample.yaml
trantanen: I agree that having those integration_config's doesn't look nice.
However, I find it even worse if we add a somewhat weird config there without any reasoning. It will cause more confusion, discussion and questioning why it's there. So instead of keeping the reasoning part secret, writing it down clearly is better in my opinion.

balaji-nordic: Got it. Go for it 🙂

PR #15911, opened 2024-06-11
bluetooth: Fix potential nullptr dereference in mds

jelledevleeschouwer:
MDS allows registration of a user supplied callback function to control acces to the Memfault characteristic and descriptors data. However, when the presence of this callback registration is tested in data_export_write, data_export_ccc_write, and mds_read, it is falsely assumed that the callback structure itself is always registered. Since this can considered programming mistake, an assert would also resolve this but that would bypass the optional nature of the registration API.

### DIFF
--- a/subsys/bluetooth/services/mds.c
+++ b/subsys/bluetooth/services/mds.c
@@ -192,7 +192,7 @@ static ssize_t mds_read(struct bt_conn *conn, const struct bt_gatt_attr *attr, v
 	ssize_t ret;
 	enum mds_read_char characteristic = (enum mds_read_char)attr->user_data;
 
-	if (mds_instance.cb->access_enable) {
+	if (mds_instance.cb && mds_instance.cb->access_enable) {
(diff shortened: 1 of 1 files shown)


KAGA164:
Thanks a lot for your contribution. Please check compliance check and apply it. Then we should be ready to merge your PR

KAGA164 (approved):

jtguggedal:
@jelledevleeschouwer I created #15969 as a copy of this PR, with the only change being to fix the compliance check. The reason is that we might be able to get this fix into NCS v2.7.0, but we're in a bit of a hurry.

jelledevleeschouwer:
> @jelledevleeschouwer I created #15969 as a copy of this PR, with the only change being to fix the compliance check. The reason is that we might be able to get this fix into NCS v2.7.0, but we're in a bit of a hurry.

@jtguggedal, sure, no problem! Would be great to see it fixed in NCS v2.7.0.

PR #15917, opened 2024-06-11
applications: nrf_desktop: Fix USB state when disconnecting cable

alstrzebonski:
Adds option to call usbd_enable/usbd_disable on USB cable connection/disconnection. It workarounds USB issues on nRF54H20.

Jira: NCSDK-27880
Jira: NCSDK-27878

### DIFF
--- a/applications/nrf_desktop/src/modules/Kconfig.usb_state
+++ b/applications/nrf_desktop/src/modules/Kconfig.usb_state
@@ -68,11 +68,11 @@ config DESKTOP_USB_INIT_THREAD_STACK_SIZE
 
 config DESKTOP_USB_INIT_THREAD_PRIORITY
 	int
-	default 0
+	default -1
 
 config DESKTOP_USB_INIT_THREAD_DELAY_MS
 	int
-	default 1
+	default 30
 	help
 	  Due to an issue in the USB stack initialization on the nRF54HX SoC,
 	  the usbd_init() cannot be called too early. The delay is introduced
@@ -177,6 +177,24 @@ config USBD_HID_IN_BUF_COUNT
 	help
 	  nRF Desktop queues HID reports at the source. There is no need to use
 	  multiple buffers in the IN pool per HID instance.
+
+config DESKTOP_USB_STACK_NEXT_DISABLE_ON_VBUS_REMOVAL
+	bool
+	default y if SOC_SERIES_NRF54HX
+	depends on DESKTOP_USB_INIT_THREAD
+	select REBOOT
+	help
+	  Disable USB stack on VBUS removal. This is a workaround for the USB
+	  driver not working correctly on the nRF54HX SoC. After the USB cable
+	  is removed, the USB driver is powered down and doesn't call appropriate
+	  callbacks to the USB stack. It may lead to the USB stack ending in a
+	  broken state. Calling usbd_disable() on VBUS removal workarounds this
+	  issue. When this option is enabled, the usb_init_thread will be used to
+	  call the usbd_enable() and usbd_disable() functions on VBUS events. This
+	  is done because the usbd_enable() function blocks if the USB cable is
+	  not connected. A separate thread allows the application to trigger a
+	  reboot from the workqueue context to prevent blocking the application
+	  forever.
 
 choice USBD_LOG_LEVEL_CHOICE
 	default USBD_LOG_LEVEL_WRN
--- a/applications/nrf_desktop/src/modules/usb_state.c
+++ b/applications/nrf_desktop/src/modules/usb_state.c
@@ -11,6 +11,8 @@
 #include <zephyr/types.h>
 #include <zephyr/sys/byteorder.h>
 #include <zephyr/sys/util.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/reboot.h>
 
 #include <zephyr/usb/usbd.h>
 #include <zephyr/usb/usbd_msg.h>
@@ -24,6 +26,7 @@
 #include <caf/events/module_state_event.h>
 
 #include <zephyr/logging/log.h>
+#include <zephyr/logging/log_ctrl.h>
 LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_USB_STATE_LOG_LEVEL);
 
 #include "hid_report_desc.h"
@@ -104,11 +107,24 @@ BUILD_ASSERT(!IS_ENABLED(CONFIG_DESKTOP_HID_STATE_ENABLE) ||
 	     (ARRAY_SIZE(usb_hid_device) <= 1));
 
 static struct usbd_context *usbd_ctx;
+static bool usb_enabled;
 
 static struct config_channel_transport cfg_chan_transport;
 
 static struct k_thread usb_init_thread;
 static K_THREAD_STACK_DEFINE(usb_init_thread_stack, CONFIG_DESKTOP_USB_INIT_THREAD_STACK_SIZE);
+
+#define USB_REQ_TIMEOUT_MS	100
+
+/* USB events */
+enum {
+	USB_REQ_ENABLE		= BIT(0),
+	USB_REQ_DISABLE		= BIT(1),
+	USB_RSP_SUCCESS		= BIT(2),
+	USB_RSP_FAIL		= BIT(3),
+};
+
+static K_EVENT_DEFINE(usb_event);
 
 
 static struct usb_hid_device *dev_to_hid(const struct device *dev)
@@ -514,7 +530,8 @@ static void usb_wakeup(void)
 {
 	int err = 0;
 
-	if (IS_ENABLED(CONFIG_DESKTOP_USB_STACK_NEXT) && usbd_ctx) {
+	if (IS_ENABLED(CONFIG_DESKTOP_USB_STACK_NEXT) && usb_enabled) {
+		__ASSERT_NO_MSG(usbd_ctx);
 		err = usbd_wakeup_request(usbd_ctx);
 	} else {
 		__ASSERT_NO_MSG(IS_ENABLED(CONFIG_DESKTOP_USB_STACK_LEGACY));
@@ -962,6 +979,64 @@ static int usb_init_next_hids_init(void)
 	return err;
 }
 
+static int usb_thread_req_send(bool enable)
+{
+	uint32_t events;
+	int err;
+
+	__ASSERT_NO_MSG(!k_event_test(&usb_event, USB_REQ_ENABLE | USB_REQ_DISABLE));
+
+	LOG_DBG("%sabling USB", enable ? "En" : "Dis");
+	(void) k_event_set(&usb_event, enable ? USB_REQ_ENABLE : USB_REQ_DISABLE);
+	events = k_event_wait(&usb_event,
+			      USB_RSP_SUCCESS | USB_RSP_FAIL,
+			      false,
+			      K_MSEC(USB_REQ_TIMEOUT_MS));
+	(void) k_event_clear(&usb_event, events);
+
+	if (events & USB_RSP_SUCCESS) {
+		LOG_DBG("USB %sabled", enable ? "en" : "dis");
+		usb_enabled = enable;
+		err = 0;
+	} else if (events & USB_RSP_FAIL) {
+		LOG_ERR("Failed to %sable USB", enable ? "en" : "dis");
+		module_set_state(MODULE_STATE_ERROR);
+		err = -EIO;
+	} else {
+		LOG_ERR("Fatal error - USB %sable timeout. Rebooting...", enable ? "en" : "dis");
+		LOG_PANIC();
+		sys_reboot(SYS_REBOOT_WARM);
+		err = -EIO;
+	}
+
+	return err;
+}
+
+static int handle_usbd_state_on_status_change(enum usbd_msg_type type)
+{
+	int err = 0;
+
+	switch (type) {
+	case USBD_MSG_VBUS_READY:
+		if (!usb_enabled) {
+			err = usb_thread_req_send(true);
+		}
+		break;
+
+	case USBD_MSG_VBUS_REMOVED:
+		if (usb_enabled) {
+			err = usb_thread_req_send(false);
+		}
+		break;
+
+	default:
+		/* Ignore */
+		break;
+	}
+
+	return err;
+}
+
 static void usb_init_next_status_cb(struct usbd_context *const usbd,
 				    const struct usbd_msg *const msg)
 {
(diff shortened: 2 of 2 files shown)


applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -1185,19 +1236,29 @@ static int usb_init_next(void)
...
+		return -EIO;
+	}
+
 	err = usbd_msg_register_cb(usbd, usb_init_next_status_cb);
 	if (err) {
 		LOG_ERR("usbd_msg_register_cb failed (err: %d)", err);
 		return err;
 	}
 
+	if (IS_ENABLED(CONFIG_DESKTOP_USB_STACK_NEXT_DISABLE_ON_VBUS_REMOVAL)) {
```
kapi-no, line 1250: what if the cable is connected during the flash operation? Would it still work correctly?

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -103,7 +103,7 @@ BUILD_ASSERT(!IS_ENABLED(CONFIG_DESKTOP_HID_STATE_ENABLE) ||
 	     IS_ENABLED(CONFIG_DESKTOP_USB_SELECTIVE_REPORT_SUBSCRIPTION) ||
 	     (ARRAY_SIZE(usb_hid_device) <= 1));
 
-static struct usbd_context *usbd_ctx;
+static atomic_ptr_t usbd_ctx = ATOMIC_PTR_INIT(NULL);
```
kapi-no, line 106: include atomic header?

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/usb_state.c
alstrzebonski: Yes, the callback is called after registration.

applications/nrf_desktop/src/modules/usb_state.c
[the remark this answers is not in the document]
alstrzebonski: I verified that it works

applications/nrf_desktop/src/modules/usb_state.c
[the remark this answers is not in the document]
alstrzebonski: Yes, calling the `usbd_msg_register_cb` should be moved before calling `usbd_init`. Otherwise we may theoretically miss the vbus connected callback.

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -1185,19 +1236,29 @@ static int usb_init_next(void)
 		return -ENXIO;
 	}
 
+	if (!usbd_can_detect_vbus(usbd)) {
+		LOG_ERR("USBD controller cannot detect VBUS state change");
```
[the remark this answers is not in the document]
alstrzebonski: Yes, it works fine.

applications/nrf_desktop/src/modules/usb_state.c
alstrzebonski: Actually it's called after usbd_init so I will register it befot calling usbd_init - that way we have no races.

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -1228,6 +1354,47 @@ static int usb_init(void)
...
+			if (err) {
+				LOG_ERR("usbd_disable failed (err: %d)", err);
+				k_event_post(&usb_event, USB_DISABLE_FAIL_MASK);
+			} else {
+				usbd_ctx = NULL;
+				k_event_post(&usb_event, USB_DISABLE_SUCCESS_MASK);
+			}
+		}
+
+		k_event_clear(&usb_event, events);
```
MarekPieta, line 1394: We need to clear requests at the beginning

MarekPieta (changes requested):

alstrzebonski:
Addressed review comments. Setting DNM as I need to conduct detailed tests before merging.

applications/nrf_desktop/src/modules/Kconfig.usb_state
```diff
@@ -178,6 +178,22 @@ config USBD_HID_IN_BUF_COUNT
 	  nRF Desktop queues HID reports at the source. There is no need to use
 	  multiple buffers in the IN pool per HID instance.
 
+config DESKTOP_USB_STACK_NEXT_DISABLE_ON_VBUS_REMOVAL
+	bool
+	default y if SOC_SERIES_NRF54HX
```
MarekPieta, line 183: `select REBOOT`

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -962,14 +977,91 @@ static int usb_init_next_hids_init(void)
...
+{
+	switch (type) {
+	case USBD_MSG_VBUS_READY:
+		if (!usb_enabled) {
+			uint32_t events;
+
+			__ASSERT_NO_MSG(!is_usb_req_pending());
+
+			LOG_DBG("Enabling USB");
+			k_event_set(&usb_event, USB_REQ_ENABLE);
```
MarekPieta, line 995: Can we make a common function like `usb_thread_req_send(bool usb_enable)` to avoid code duplication here? Then you could even remove `is_usb_req_pending` because it would be called only once (whole function explicitly uses `USB_REQ_...` flags so I would suggest not to use additional wrapper for the assertion too)

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -962,14 +977,91 @@ static int usb_init_next_hids_init(void)
...
+			events = k_event_wait(&usb_event,
+					      USB_RSP_SUCCESS | USB_RSP_FAIL,
+					      false,
+					      K_MSEC(100));
```
MarekPieta, line 999: Can we make the timeout as Kconfig/define, please?

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -1228,6 +1340,44 @@ static int usb_init(void)
...
+			if (err) {
+				LOG_ERR("usbd_disable failed (err: %d)", err);
+				k_event_post(&usb_event, USB_RSP_FAIL);
+			}
+		} else {
+			__ASSERT_NO_MSG(false);
+		}
+
+		if (err) {
+			k_event_post(&usb_event, USB_RSP_FAIL);
```
MarekPieta, line 1374: `k_event_post(&usb_event, err ? USB_RSP_FAIL : USB_RSP_SUCCESS)`?

MarekPieta: You could also validate return value (to ensure no bit was set in rsp before)

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -1228,6 +1340,44 @@ static int usb_init(void)
...
+			err = usbd_enable(usbd_ctx);
+			if (err) {
+				LOG_ERR("usbd_enable failed (err: %d)", err);
+				k_event_post(&usb_event, USB_RSP_FAIL);
+			}
+		} else if (events & USB_REQ_DISABLE) {
+			err = usbd_disable(usbd_ctx);
+			if (err) {
+				LOG_ERR("usbd_disable failed (err: %d)", err);
+				k_event_post(&usb_event, USB_RSP_FAIL);
```
MarekPieta, line 1367: Post done at the end. should not be done also here

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -1228,6 +1340,44 @@ static int usb_init(void)
...
+				      K_FOREVER);
+		k_event_clear(&usb_event, events);
+		__ASSERT_NO_MSG(!(events & USB_REQ_ENABLE) || !(events & USB_REQ_DISABLE));
+
+		__ASSERT_NO_MSG(usbd_ctx);
+		if (events & USB_REQ_ENABLE) {
+			err = usbd_enable(usbd_ctx);
+			if (err) {
+				LOG_ERR("usbd_enable failed (err: %d)", err);
+				k_event_post(&usb_event, USB_RSP_FAIL);
```
MarekPieta, line 1361: Post done at the end.  should not be done also here

applications/nrf_desktop/src/modules/Kconfig.usb_state
MarekPieta: btw. Does reboot already work on nRF54H? IIRC there used to be a related issue before

applications/nrf_desktop/src/modules/usb_state.c
MarekPieta, line 995: `(void)` cast to indicate that you ignore the returned value. Also in other places

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -962,14 +977,91 @@ static int usb_init_next_hids_init(void)
...
+				LOG_DBG("USB enabled");
+				usb_enabled = true;
+			} else if (events & USB_RSP_FAIL) {
+				LOG_ERR("Failed to enable USB");
+				module_set_state(MODULE_STATE_ERROR);
+				return -EIO;
+			} else {
+				LOG_ERR("Fatal error - USB enable timeout. Rebooting...");
+				LOG_PANIC();
+				sys_reboot(SYS_REBOOT_WARM);
```
MarekPieta, line 1011: I would return `-EIO` here for sanity (even though code should not reach it)

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -1021,6 +1113,14 @@ static void usb_init_next_status_cb(struct usbd_context *const usbd,
 		break;
 	}
 
+	if (IS_ENABLED(CONFIG_DESKTOP_USB_STACK_NEXT_DISABLE_ON_VBUS_REMOVAL)) {
+		err = handle_usbd_state_on_status_change(msg->type);
```
MarekPieta, line 1117: Please keep `err` locally in this scope

MarekPieta: You actually may not emit an extra log here, `handle_usbd_state_on_status_change` already does it

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/usb_state.c
alstrzebonski: I'll keep the error log.

applications/nrf_desktop/src/modules/Kconfig.usb_state
kapi-no: Sys reboot is also used in the SUIT DFU using CC and SMP. It has to work

kapi-no:
```
static void reboot_request_handler(struct k_work *work)
{
	LOG_INF("Handle reset request now");

	if (flash_area) {
		terminate_dfu();
	}

	LOG_PANIC();

#if CONFIG_SUIT
	if ((update_candidate.mem != 0) && (update_candidate.size != 0) &&
	    (cur_offset == img_length) && !slot_was_used_by_other_transport) {
		LOG_INF("Configuration channel reboot request triggered SUIT update");
		int ret = suit_trigger_update(&update_candidate, UPDATE_CANDIDATE_CNT);

		LOG_INF("SUIT update triggered with status: %d", ret);
	}
#endif

	sys_reboot(SYS_REBOOT_WARM);

	if (!k_work_delayable_is_pending(&background_erase)) {
		config_channel_dfu_lock_release();
	}
}

```

MarekPieta: Just double checking, since I recall some related issues

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -962,6 +979,68 @@ static int usb_init_next_hids_init(void)
...
+
+	LOG_DBG("%sabling USB", enable ? "En" : "Dis");
+	(void) k_event_set(&usb_event, enable ? USB_REQ_ENABLE : USB_REQ_DISABLE);
+	events = k_event_wait(&usb_event,
+			      USB_RSP_SUCCESS | USB_RSP_FAIL,
+			      false,
+			      K_MSEC(USB_REQ_TIMEOUT_MS));
+	(void) k_event_clear(&usb_event, events);
+
+	if (!events) {
```
MarekPieta, line 996: Why not `if ... else if ... else` with `err` local variable ? You could avoid early returns and assertion at the end of this function

applications/nrf_desktop/src/modules/Kconfig.usb_state
```diff
@@ -178,6 +178,23 @@ config USBD_HID_IN_BUF_COUNT
...
+	  Disable USB stack on VBUS removal. This is a workaround for the USB
+	  driver not working correctly on the nRF54HX SoC. After the USB cable
+	  is removed, the USB driver is powered down and doesn't call appropriate
+	  callbacks to the USB stack. It may lead to the USB stack ending in the
+	  broken state. Calling usbd_disable() on VBUS removal workarounds this
+	  issue. When this option is enabled, the usb_init_thread will be used to
+	  call usbd_enable() and usbd_disable() functions on VBUS events. This is
+	  done because the usbd_enable() function blocks if USB cable is not
+	  connected. A separate thread allows application to trigger reboot from
+	  the workqueue context to prevent blocking application forever.
```
kapi-no, line 196:
```suggestion
	  Disable USB stack on VBUS removal. This is a workaround for the USB
	  driver not working correctly on the nRF54HX SoC. After the USB cable
	  is removed, the USB driver is powered down and doesn't call appropriate
	  callbacks to the USB stack. It may lead to the USB stack ending in a
	  broken state. Calling usbd_disable() on VBUS removal workarounds this
	  issue. When this option is enabled, the usb_init_thread will be used to
	  call the usbd_enable() and usbd_disable() functions on VBUS events. This is
	  done because the usbd_enable() function blocks if the USB cable is not
	  connected. A separate thread allows the application to trigger a reboot from
	  the workqueue context to prevent blocking the application forever.
```

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -962,6 +979,68 @@ static int usb_init_next_hids_init(void)
...
+	__ASSERT_NO_MSG(false);
+	return -EINVAL;
+}
+
+static int handle_usbd_state_on_status_change(enum usbd_msg_type type)
+{
+	switch (type) {
+	case USBD_MSG_VBUS_READY:
+		if (!usb_enabled) {
+			return usb_thread_req_send(true);
```
MarekPieta, line 1024: Can we avoid early returns here by using local `err` variable?

applications/nrf_desktop/src/modules/usb_state.c
```diff
@@ -962,6 +979,68 @@ static int usb_init_next_hids_init(void)
...
+		}
+
```
MarekPieta, line 1026: Empty line not needed before break

MarekPieta (changes requested):

kapi-no (approved):
Looking good

tejlmand (approved):

alstrzebonski:
Fixed build warning in release configuration. Finished testing - everything seems to be working.

alstrzebonski:
@anangl, please merge :)

PR #15918, opened 2024-06-11
doc: sml: add new fast pair features

kapi-no:
Added new Fast Pair features that will be used in the Software Maturity documentation.

### DIFF
--- a/doc/_scripts/software_maturity/software_maturity_features.yaml
+++ b/doc/_scripts/software_maturity/software_maturity_features.yaml
@@ -132,6 +132,9 @@ features:
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
+  fast_pair:
+    Personalized Name extension: BT_FAST_PAIR_PN
+    Find My Device Network extension: BT_FAST_PAIR_FMDN
   trusted_firmware_m:
     Minimal Build: BUILD_WITH_TFM && TFM_PROFILE_TYPE_MINIMAL
     Full build: BUILD_WITH_TFM && TFM_PROFILE_TYPE_NOT_SET && NRF_SECURITY
--- a/subsys/bluetooth/services/fast_pair/Kconfig.fast_pair
+++ b/subsys/bluetooth/services/fast_pair/Kconfig.fast_pair
@@ -22,10 +22,11 @@ config BT_FAST_PAIR_REQ_PAIRING
 	  procedure that is performed without the Bluetooth pairing operation.
 
 config BT_FAST_PAIR_PN
-	bool "Fast Pair Personalized Name extension"
+	bool "Fast Pair Personalized Name extension [EXPERIMENTAL]"
 	default y if !BT_FAST_PAIR_FMDN
 	select BT_FAST_PAIR_GATT_SERVICE_ADDITIONAL_DATA
 	select BT_FAST_PAIR_STORAGE_PN
+	select EXPERIMENTAL
 	help
 	  Enable Fast Pair Personalized Name extension.
 


doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -132,6 +132,12 @@ features:
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
+  fast_pair:
+    Initial Pairing: BT_FAST_PAIR
+    Subsequent pairing: BT_FAST_PAIR && EXPERIMENTAL
+    Battery Notification extension: BT_FAST_PAIR && EXPERIMENTAL
+    Personalized Name extension: BT_FAST_PAIR && BT_FAST_PAIR_PN
```
alstrzebonski, line 139: I wonder if we shouldn't make it experimental - there are unresolved bugs on Google side.

kapi-no: all features will be marked currently as experimental, but if we were to remove the `EXPERIMENTAL` option from the `BT_FAST_PAIR` symbol selection, then I would add it under the `BT_FAST_PAIR_PN` option

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -132,6 +132,12 @@ features:
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
+  fast_pair:
+    Initial Pairing: BT_FAST_PAIR
+    Subsequent pairing: BT_FAST_PAIR && EXPERIMENTAL
+    Battery Notification extension: BT_FAST_PAIR && EXPERIMENTAL
```
MarekPieta, line 138: Do we need to explicitly add `EXPERIMENTAL` here? `CONFIG_BT_FAST_PAIR` selects experimental already.

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -132,6 +132,12 @@ features:
...
+    Personalized Name extension: BT_FAST_PAIR && BT_FAST_PAIR_PN
+    Find My Device Network extension: BT_FAST_PAIR && BT_FAST_PAIR_FMDN
```
MarekPieta, line 140: I wonder if we need to explicitly list `BT_FAST_PAIR` if the second listed option depends on it in Kconfig

doc/_scripts/software_maturity/software_maturity_features.yaml
kapi-no: Yes, but we want to remove the EXPERIMENTAL selection from the BT_FAST_PAIR while still tagging the BN extension as experimental.

doc/_scripts/software_maturity/software_maturity_features.yaml
kapi-no: yes, maybe we don't have to do it

doc/_scripts/software_maturity/software_maturity_features.yaml
alstrzebonski: So why don't we add experimental to BT_FAST_PAIR_PN? Or it will be done in Kconfig.fast_pair?

doc/_scripts/software_maturity/software_maturity_features.yaml
kapi-no: I will improve it

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -132,6 +132,12 @@ features:
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
+  fast_pair:
+    Initial Pairing: BT_FAST_PAIR
+    Subsequent pairing: BT_FAST_PAIR && EXPERIMENTAL
```
MarekPieta, line 137: Can we combine initial and subsequent pairing (it's controlled by the same Kconfig symbol)

kapi-no: We want to declare different maturity level for both and as you indicated we will have a dedicated Kconfig for enabling support for subsequent pairing in the future

alstrzebonski (approved):

subsys/bluetooth/services/fast_pair/Kconfig.fast_pair
```diff
@@ -26,6 +26,7 @@ config BT_FAST_PAIR_PN
 	default y
 	select BT_FAST_PAIR_GATT_SERVICE_ADDITIONAL_DATA
 	select BT_FAST_PAIR_STORAGE_PN
+	select EXPERIMENTAL
```
MarekPieta, line 29: We should update prompt here

kapi-no:
Rebased due to conflicts

doc/_scripts/software_maturity/software_maturity_features.yaml
MarekPieta: I am worried that we may need a separate Kconfig for subequent pairing in that case. Current solution may work by an accident (we are looking for configurations with both BT_FAST_PAIR && EXPERIMENTAL enabled). Similar for battery notification.

kapi-no:
put on hold

kapi-no:
Reduced the scope to only indicate features that have Kconfig option

subsys/bluetooth/services/fast_pair/Kconfig.fast_pair
```diff
@@ -22,10 +22,11 @@ config BT_FAST_PAIR_REQ_PAIRING
 	  procedure that is performed without the Bluetooth pairing operation.
 
 config BT_FAST_PAIR_PN
```
MarekPieta, line 24: In commit message should be `Personalized Name Extension` (it's `Personalized Extension` now)

MarekPieta: Also: `experimnetal` -> `experimental`

MarekPieta (approved):

subsys/bluetooth/services/fast_pair/Kconfig.fast_pair
kapi-no: fixed

MarekPieta (approved):

alstrzebonski (approved):

kapi-no:
@anangl, I think this is ready for merging

annwoj (approved):

PR #15937, opened 2024-06-12
Bluetooth: Mesh: Fix scene recall for LLC server

HaavardRei:
When recalling a scene during when the controller is not in standby, the Light Lightness Server would turn off the light even if the scene had the light turned on.

This commit fixes the issue by removing a check for whether the light is on when recalling a scene. The check was added to prevent recalling the same scene multiple times, but it is not necessary as this is checked in in `bt_mesh_scene_srv_set` (in `scene_srv.c`).

### DIFF
--- a/subsys/bluetooth/mesh/light_ctrl_srv.c
+++ b/subsys/bluetooth/mesh/light_ctrl_srv.c
@@ -1803,7 +1803,7 @@ static void scene_recall(const struct bt_mesh_model *model, const uint8_t data[]
 			ctrl_enable(srv);
 		}
 
-		if (!!scene->light && !atomic_test_bit(&srv->flags, FLAG_ON)) {
+		if (!!scene->light) {
 			turn_on(srv, transition, true);
 		} else if (atomic_test_bit(&srv->flags, FLAG_ON)) {
 			transition_start(srv, LIGHT_CTRL_STATE_STANDBY, 0);


subsys/bluetooth/mesh/light_ctrl_srv.c
ludvigsj, line 1: As far as I can tell, the behavior after this change when the state machine is in the PROLONG state or in the middle of transitioning to it is that the lights will go back to the ON state when receiving a scene where `light == true`. Is this the intention? I would suggest to rather add a `!scene->light &&` to the `else if`, which makes the behavior: if the light is in PROLONG and `scene->light == true`, stay in PROLONG.

ludvigsj: Also, the scene server checking with early return on scene == current_scene was already there when this `if ... else if ...` was augmented with the flag check, but it was apparently needed. I think the problem might be recalling _different_ scenes which all have light == true? @omkar3141 will have to elaborate on this.

HaavardRei: To me, it makes sense that the lights will go back to on. But I agree that it may be a bit ambiguous as the state of the controller is not stored with the scene.

HaavardRei: > Also, the scene server checking with early return on scene == current_scene was already there when this `if ... else if ...` was augmented with the flag check, but it was apparently needed. I think the problem might be recalling _different_ scenes which all have light == true? @omkar3141 will have to elaborate on this.

I've tested the change with multiple scenes with the lights on, which didn't result in any weird behavior (I. e. turning the lights off which was the initial problem).

ludvigsj (dismissed):

omkar3141 (approved):

subsys/bluetooth/mesh/light_ctrl_srv.c
omkar3141: Upon Scene Recall, irrespective of current state machine state, the "Light On" event gets generated (if Light On == 1, as saved in the Scene), that pull the state machine from whatever state it is in ... to ON state.

ludvigsj (approved):
Looks good now that Omkar confirmed the spec interpretation

PR #15845, opened 2024-06-10
doc: Add sysbuild migration guide

nordicjm:
Adds a sysbuild migration guide and updates documentation

PR #15943, opened 2024-06-12
wifi: Fix radio test combo with sysbuild

krish2718:


bama-nordic (approved):

nordicjm (commented):
can you add this to the application itself since I don't think anyone else is going to need this image? See doc on https://github.com/nrfconnect/sdk-nrf/pull/15845/files#diff-574bd304f11ae194ae13adcbeada903eea9549b3364faec29d1592f6d14ec088R41

nordicjm (changes requested):
Move to application

krish2718:
> > can you add this to the application itself since I don't think anyone else is going to need this image? See doc on https://github.com/nrfconnect/sdk-nrf/pull/15845/files#diff-574bd304f11ae194ae13adcbeada903eea9549b3364faec29d1592f6d14ec088R41
> 
> I can see two users, BLE only radio test and Wi-Fi + BLE combo, so, have chosen a global sysbuild. It would be good to get some feedback from BLE team.

Moved to application as requested.

krga2022 (approved):

samples/wifi/radio_test/Kconfig.sysbuild
```diff
@@ -0,0 +1,40 @@
...
+config SUPPORT_NETCORE_PERIPHERAL_RADIO_TEST
+	bool
+	default y if NRF700X_RADIO_TEST_COMBO
+
+choice NETCORE
+	prompt "Netcore image"
+	depends on SUPPORT_NETCORE && !EXTERNAL_CONFIGURED_NETCORE
+
+config NETCORE_PERIPHERAL_RADIO_TEST
+	bool "radio_test"
```
nordicjm, line 23:
```suggestion
	bool "peripheral_radio_test"
```

samples/wifi/radio_test/Kconfig.sysbuild
```diff
@@ -0,0 +1,40 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config NRF700X_RADIO_TEST_COMBO
+	bool "Enable radio test combo build for Wi-Fi and Bluetooth"
+	depends on SOC_NRF5340_CPUAPP
```
nordicjm, line 9:
```suggestion
```

samples/wifi/radio_test/Kconfig.sysbuild
```diff
@@ -0,0 +1,40 @@
...
+
+menu "Network core configuration"
+	depends on SUPPORT_NETCORE
+
+config SUPPORT_NETCORE_PERIPHERAL_RADIO_TEST
+	bool
+	default y if NRF700X_RADIO_TEST_COMBO
```
nordicjm, line 16: default y, the if isn't needed as support netcore is NRF5340 soc series (though includes NRF54H20 too)

samples/wifi/radio_test/Kconfig.sysbuild
```diff
@@ -0,0 +1,40 @@
...
+choice NETCORE
+	prompt "Netcore image"
+	depends on SUPPORT_NETCORE && !EXTERNAL_CONFIGURED_NETCORE
+
+config NETCORE_PERIPHERAL_RADIO_TEST
+	bool "radio_test"
+	depends on SUPPORT_NETCORE_PERIPHERAL_RADIO_TEST
+	help
+	  Include radio_test as the netcore image to use
+endchoice
```
nordicjm, line 27: newline

samples/wifi/radio_test/CMakeLists.txt
```diff
@@ -20,3 +20,36 @@ target_sources(app PRIVATE
 	src/nrf_wifi_radio_test_shell.c
 	src/nrf_wifi_radio_ficr_shell.c
 )
+
+
+if(SB_CONFIG_NETCORE_PERIPHERAL_RADIO_TEST)
```
nordicjm, line 25: None of this will work, it needs to go in the sysbuild.cmake file, plus needs 2 space indentation for cmake

nordicjm (changes requested):

samples/wifi/radio_test/Kconfig.sysbuild
rado17: Could you provide reason for this suggestion. Chaitanya is not available today and I am looking for some context before making this change.

Thanks

nordicjm: This isn't needed, in your app you would just add to sysbuild.conf (or sample.yaml): `SB_CONFIG_NETCORE_PERIPHERAL_RADIO_TEST=y`

samples/wifi/radio_test/CMakeLists.txt
krish2718: > None of this will work, it needs to go in the sysbuild.cmake file, plus needs 2 space indentation for cmake

I did test this at least build wise and had seen hex files for both cores and merged files. My initial attempt was to use sysbuild.cmake but faced some errors and upon re-reading the doc link that you have shared above, for other processors child images only CMakeLists.txt was mentioned and sysbuild.cmake.

samples/wifi/radio_test/Kconfig.sysbuild
krish2718: Sure, but what about no-sysbuilt support? As this is the first sysbuild release I intentionally left both versions of macros.WDYT?

sachinthegreen (approved):

tejlmand (approved):

PR #15946, opened 2024-06-12
applications: serial_lte_modem: adjust SLM_CARRIER_APP_DATA_BUFFER_LEN

kacperradoszewski:
The buffer used to set application data via AT#XCARRIER="app_data_set" is of size SLM_CARRIER_APP_DATA_BUFFER_LEN. This buffer however must also accommodate a null-terminator on top of the actual data when retrieving the string from the issued AT command. Increase the size by one byte as it is crucial to pass a test item of a certain operator, which verifies the upper limit of data that can be set.


kacperradoszewski:
@shanthanordic please consider for the v2.7.0 milestone. This is a crucial fix to pass a certification test case of one of the operators.

SeppoTakalo (approved):

PR #15949, opened 2024-06-12
sysbuild: Only apply external flash to MCUboot and main image

nordicjm:
Resolves some warnings from appearing by only applying the value of using external flash for the secondary slot to the main image and MCUboot


Balaklaka:
It is important that this PR is reviewed and added to 2.7.0. We have tested that this will remove a specific Kconfig warnings from all samples with Thingy:53.

nordicjm:
> It is important that this PR is reviewed and added to 2.7.0. We have tested that this will remove a specific Kconfig warnings from all samples with Thingy:53.

Can you approve changes?

tejlmand (approved):

nordic-segl (approved):

PR #15960, opened 2024-06-12
samples: bluetooth: peripheral_hids_mouse: Don't start advertising when active

maje-emb:
Added a check to ensure Bluetooth advertising is not initiated if it is already running.

### DIFF
--- a/samples/bluetooth/peripheral_hids_mouse/src/main.c
+++ b/samples/bluetooth/peripheral_hids_mouse/src/main.c
@@ -116,6 +116,8 @@ static struct conn_mode {
 	bool in_boot_mode;
 } conn_mode[CONFIG_BT_HIDS_MAX_CLIENT_COUNT];
 
+static volatile bool is_adv_running;
+
 static struct k_work adv_work;
 
 static struct k_work pairing_work;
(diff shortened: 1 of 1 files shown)


grochu (approved):
The updated advertising logic looks good to me.

samples/bluetooth/peripheral_hids_mouse/src/main.c
```diff
@@ -162,14 +166,23 @@ static void advertising_continue(void)
 
 	if (!k_msgq_get(&bonds_queue, &addr, K_NO_WAIT)) {
 		char addr_buf[BT_ADDR_LE_STR_LEN];
+		int err;
+
+		if (atomic_test_bit(&adv_state, ADV_RUNNING)) {
```
KAGA164, line 171: I would say that we do not need atomic here. Every access to adv_state seems to be done from the same no preemptive context

maje-emb: `adv_state` is used in the connected and disconnected callbacks. As far as I know, these callbacks are called from different threads, BT and sysworkq.

KAGA164: Yes and they are cooperative threads. You also call advertising_continue only from sysworkq

doki-nordic (approved):

samples/bluetooth/peripheral_hids_mouse/src/main.c
maje-emb: Right. Changed to bool.

KAGA164 (approved):

PR #15971, opened 2024-06-13
Samples: Bluetooth: Mesh: Fix chat sample warning

HaavardRei:
Removes shell print statement from the model handler initialization which would cause a warning due to the shell not being initialized yet.

### DIFF
--- a/samples/bluetooth/mesh/chat/src/model_handler.c
+++ b/samples/bluetooth/mesh/chat/src/model_handler.c
@@ -425,7 +425,7 @@ const struct bt_mesh_comp *model_handler_init(void)
 	k_work_init_delayable(&attention_blink_work, attention_blink);
 
 	chat_shell = shell_backend_uart_get_ptr();
-	shell_print(chat_shell, ">>> Bluetooth Mesh Chat sample <<<");
+	printk(">>> Bluetooth Mesh Chat sample <<<\n");
 
 	return &comp;
 }


samples/bluetooth/mesh/chat/src/model_handler.c
```diff
@@ -425,7 +425,6 @@ const struct bt_mesh_comp *model_handler_init(void)
 	k_work_init_delayable(&attention_blink_work, attention_blink);
 
 	chat_shell = shell_backend_uart_get_ptr();
-	shell_print(chat_shell, ">>> Bluetooth Mesh Chat sample <<<");
```
ludvigsj, line 428: If the shell isn't initialized yet at this point, it should also not create UART conflicts if you just change this to a `printk` instead, so we still get this welcome message. Have you tried this? If that works, I think it would be better than removing the welcome message.

ludvigsj (changes requested):

samples/bluetooth/mesh/chat/src/model_handler.c
alxelax: Shouldn't application banner be used instead?

ludvigsj: The application banner with custom application name only applies to applications with a VERSION file, which this does not. I'll add an issue to add VERSION files to all our samples that we can do after the 2.7 release, but for now I think changing this to a printk is the better solution for the 2.7 milestone, then we can do it "properly" after release (I really don't want to modify samples more than necessary now that we have completed testing of them)

HaavardRei: Changing to printk works, but I had the same thought as @alxelax . But if we add application banner here shouldn't we add it to all samples?

ludvigsj: All samples which currently have a VERSION file also has the correct banner, this was added in #15948 . Like mentioned, samples which currently do _not_ have a VERSION file will not get this for NCS 2.7.0, to avoid changing samples more than necessary shortly before release.

HaavardRei: Changed now, will you add a note to the issue once it is added to remove the printk from this sample?

alxelax (approved):

ludvigsj (approved):

PR #15972, opened 2024-06-13
samples: matter: Fixed secure backend remove condition

kkasperczyk-no:
The SecureRemove method removes object only if it is NOT present in the map, what seems to be a broken condition.


ArekBalysNordic (approved):

LuDuda (approved):
LGTM :+1:

PR #15975, opened 2024-06-13
samples: event_manager_proxy: verify host console

nordic-piks:
Enable automatic testing by checkout console output.

### DIFF
--- a/.github/test-spec.yml
+++ b/.github/test-spec.yml
@@ -543,3 +543,11 @@
   - "samples/suit/**/*"
   - "subsys/mgmt/suitfu/**/*"
   - "tests/subsys/suit/**/*"
+
+"CI-test-low-level":
+  - "include/event_manager_proxy.h"
+  - "subsys/event_manager_proxy/**/*"
+  - "samples/ipc/**/*"
+  - "samples/event_manager_proxy/**/*"
+  - "tests/subsys/event_manager_proxy/**/*"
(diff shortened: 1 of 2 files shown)


.github/test-spec.yml
```diff
@@ -543,3 +543,10 @@
   - "samples/suit/**/*"
   - "subsys/mgmt/suitfu/**/*"
   - "tests/subsys/suit/**/*"
+
+"CI-test-low-level":
+  - "include/event_manager_proxy.h"
+  - "subsys/event_manager_proxy/**/*"
+  - "samples/ipc/**/*"
+  - "samples/event_manager_proxy/**/*"
```
miha-nordic, line 551: This is in scope of apps pipeline. Do you want to duplicate that or it is tested in other way?

miha-nordic, line 551: This is in scope of apps pipeline. Do you want to duplicate that or it is tested in other way?

miha-nordic (changes requested):

.github/test-spec.yml
nordic-piks: Low level is using twister to check sample - which is different than apps.
Beside that, this will enable checking IPC at nrf54*.
Beside that, low-level will use it to calculate code coverage of IPC driver.

miha-nordic (approved):

nordic-segl (approved):

PR #15980, opened 2024-06-13
samples: matter: Fix radio core image for nRF54H20 + nRF7002EK

ArekBalysNordic:
Fixed the WiFi variant for nRF54H20.

### DIFF
--- a/samples/matter/lock/Kconfig.sysbuild
+++ b/samples/matter/lock/Kconfig.sysbuild
@@ -8,6 +8,10 @@
 config NRF_DEFAULT_IPC_RADIO
 	default y
 
+# Use HCI IPC as radio core image for nRF54H20 + nRF7002EK
+config NRF_DEFAULT_BLUETOOTH
+	default y if SOC_SERIES_NRF54HX && WIFI_NRF700X
+
 # Enable IEEE802.15.4 serialization to network core
 config NETCORE_IPC_RADIO_IEEE802154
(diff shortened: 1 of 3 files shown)


samples/matter/lock/Kconfig.sysbuild
```diff
@@ -8,6 +8,10 @@
 config NRF_DEFAULT_IPC_RADIO
 	default y
 
+# Use HCI IPC as radio core image for nRF54H20 + nRF7002EK
+config NRF_DEFAULT_BLUETOOTH
+	default y if SOC_SERIES_NRF54HX && WIFI_NRF700X
```
doublemis1, line 13: radio IPC does not work for nRF54H20 with WiFi?

ArekBalysNordic: Something doesn't work here with WiFi, for now, we need to use hci_ipc and in the next step, we will switch to radio_ipc. Are you OK with that?

LuDuda (approved):

samples/matter/lock/Kconfig.sysbuild
ArekBalysNordic: Currently, there is an additional dts overlay file as a workaround for radio IPC that works only if IEEE802154 is enabled. For the WiFi variant, we don't need it so for now I switched to hci_ipc. In the future, I'm going to use radio IPC for this purpose as well, once it is ready.

doublemis1 (approved):

PR #9865, opened 2023-01-10
bluetooth: mesh: Fix Light PI regulator-related issues.

omkar3141:
Fix Light PI regulator-related issues.

- make sure regulator is enabled for Light LC model
- recall controller state only when current state differs
- run regulator when ambient lux level is reported

PR #15985, opened 2024-06-13
Bluetooth: Mesh: Don't start LC regulator on powerup

ludvigsj:
After PR #9865, the PI-regulator should only start once the server has started receiving ambient lux level data, and not when the LC starts. This was fixed for initial boot and when manually enabling the LC server, however the Power OnOff handling was not updated to align with this, causing every subsequent reboot to indiscriminately start the regulator at boot without waiting for sensor data. This is fixed with this PR.


HaavardRei (approved):

subsys/bluetooth/mesh/light_ctrl_srv.c
```diff
@@ -1970,7 +1970,6 @@ static int light_ctrl_srv_start(const struct bt_mesh_model *model)
 		break;
 	case BT_MESH_ON_POWER_UP_RESTORE:
 		if (is_enabled(srv)) {
-			reg_start(srv);
```
Balaklaka, line 1973: I have one questions. Are we sure that FLAG_AMBIENT_LUXLEVEL_SET is cleared? So when handle_sensor_status is run it calls the reg_start()? If not reg_start will not be called before a disable/enable is called.

Balaklaka (approved):

omkar3141 (approved):

PR #15992, opened 2024-06-13
scripts: hid_configurator: Add support in Windows for dll folder

zycz:
Adding the path to find a .dll to PATH
environment variable has been removed in Python 3.8. Now os.add_dll_directory is recommended solution for adding dll. This commit adds an optional argument to script with path to dll folder.

Jira: NCSDK-27863

### DIFF
--- a/scripts/hid_configurator/README.rst
+++ b/scripts/hid_configurator/README.rst
@@ -54,10 +54,7 @@ Complete the following steps:
    Use the bundled DLL or build it according to instructions in `HIDAPI library`_.
 
 #. Select the appropriate HIDADPI library version for the used Windows system (either ``x86`` or ``x64``).
-   Copy the :file:`hidapi.dll` file and paste it into either of the following directory:
-
-   * The directory where used Python executable is located.
-   * The :file:`Windows\\System32` directory, for example :file:`C:\\Windows\\System32`.
+   Copy the :file:`hidapi.dll` file and paste it into the directory where the :file:`configurator_cli.py` script is located.
 
 #. Install `pyhidapi Python wrapper`_ and other required libraries with the following command:
 
--- a/scripts/hid_configurator/configurator_cli.py
+++ b/scripts/hid_configurator/configurator_cli.py
@@ -8,6 +8,15 @@ import logging
 import os
 import pprint
 
+if os.name == "nt":
+    try:
+        script_dir = os.path.dirname(os.path.abspath(__file__))
+        os.add_dll_directory(script_dir)
+    except AttributeError:
+        print('Cannot add directory containing configurator_cli.py script to DLL search path. \
+               Please update to Python 3.8 or newer to allow placing hidapi.dll in the directory \
+               containing configurator_cli.py script')
+
 from NrfHidManager import NrfHidManager
 
 from modules.module_config import MODULE_CONFIG
@@ -186,12 +195,7 @@ def perform_led_stream(dev, args):
 
 
 def parse_arguments():
-    try:
-        parser = argparse.ArgumentParser(allow_abbrev=False)
-    except TypeError:
-        # The allow_abbrev argument was added in Python 3.5.
-        # Skip setting the value if used Python version does not support it.
-        parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(allow_abbrev=False)
 
     parser.add_argument(dest='device', default=None, nargs='?',
                         help='Device specified by type, board name or HW ID '
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ ext/lib/bin/libsupl.zip
 ext/BSEC_*
 
 applications/nrf5340_audio/src/fw_info_app.c
+scripts/hid_configurator/hidapi.dll


scripts/hid_configurator/configurator_cli.py
```diff
@@ -186,13 +196,11 @@ def perform_led_stream(dev, args):
 
 
 def parse_arguments():
-    try:
```
zycz, line 189: removed parser = argparse.ArgumentParser() - as compliance did not allow it.
We require usage of python => 3.9  so I think we can do it

scripts/hid_configurator/README.rst
```diff
@@ -56,8 +56,13 @@ Complete the following steps:
 #. Select the appropriate HIDADPI library version for the used Windows system (either ``x86`` or ``x64``).
    Copy the :file:`hidapi.dll` file and paste it into either of the following directory:
 
-   * The directory where used Python executable is located.
-   * The :file:`Windows\\System32` directory, for example :file:`C:\\Windows\\System32`.
+   * The directory hid_configurator directory.
```
MarekPieta, line 59: The directory where `configurator_cli.py` script is located (to be more precise)?

scripts/hid_configurator/README.rst
```diff
@@ -56,8 +56,13 @@ Complete the following steps:
...
+   * Any folder you want, and add the folder path as an argument to the script:
+
+      .. parsed-literal::
+      :class: highlight
+
+      python3 configurator_cli.py --dll_path "C:\\Users\\user\\hidapi-win\\x64" show
```
MarekPieta, line 65: Discussed offline. We will desist from making the dll path configurable from CLI argument (it significantly complicates implementation and causes some compliance issues while not brining much added value).

scripts/hid_configurator/configurator_cli.py
MarekPieta: I will double check this and get back to you

MarekPieta (changes requested):

scripts/hid_configurator/README.rst
```diff
@@ -54,10 +54,9 @@ Complete the following steps:
    Use the bundled DLL or build it according to instructions in `HIDAPI library`_.
 
 #. Select the appropriate HIDADPI library version for the used Windows system (either ``x86`` or ``x64``).
-   Copy the :file:`hidapi.dll` file and paste it into either of the following directory:
+   Copy the :file:`hidapi.dll` file and paste it into the following directory:
```
mkapala-nordic, line 57:
```
Copy the :file:`hidapi.dll` file and paste it into the directory where :file:`configurator_cli.py` script is located.
```
?

mkapala-nordic (approved):

scripts/hid_configurator/README.rst
```diff
@@ -54,10 +54,7 @@ Complete the following steps:
    Use the bundled DLL or build it according to instructions in `HIDAPI library`_.
 
 #. Select the appropriate HIDADPI library version for the used Windows system (either ``x86`` or ``x64``).
-   Copy the :file:`hidapi.dll` file and paste it into either of the following directory:
-
-   * The directory where used Python executable is located.
-   * The :file:`Windows\\System32` directory, for example :file:`C:\\Windows\\System32`.
+   Copy the :file:`hidapi.dll` file and paste it into the directory where :file:`configurator_cli.py` script is located.
```
annwoj, line 57:
```suggestion
   Copy the :file:`hidapi.dll` file and paste it into the directory where the :file:`configurator_cli.py` script is located.
```

annwoj (approved):
Just an article suggestion - approving

scripts/hid_configurator/configurator_cli.py
```diff
@@ -8,6 +8,13 @@
 import os
 import pprint
 
+if os.name == "nt":
+    try:
+        script_dir = os.path.dirname(os.path.abspath(__file__))
+        os.add_dll_directory(script_dir)
+    except AttributeError:
+        print('Cannot add DLL directory. Consider using Python 3.8 or newer')
```
MarekPieta, line 16: We could require Python >= 3.8 (and state here that `you must use Python 3.8 or newer` - consider is not a right word then; btw. maybe we could also inform about minimal Python version in doc?).

Alternatively we can allow user to still use older Python version, remove this warning and let old Python rely on old ways to find dll files (make sure to update doc then).

MarekPieta (changes requested):

scripts/hid_configurator/configurator_cli.py
zycz: I thought that we can leave it as I did. If it works for someone he can ignore my print warning. I don't want to support old versions of python but want to give a notice to someone who is using old version that he might want to update it

scripts/hid_configurator/configurator_cli.py
```diff
@@ -8,6 +8,13 @@
 import os
 import pprint
 
+if os.name == "nt":
```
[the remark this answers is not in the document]
zycz: > Maybe we could try to add dll directory unconditionally here? - bad idea - Lets not mix windows specific instructions with Linux

> Or we could make this check more straightforward by relying on sys.platform ?
- yes we can, thought about it,  but then we need to import extra module so I decided this is a better way.

scripts/hid_configurator/configurator_cli.py
MarekPieta: `Cannot add directory containing configurator_cli.py script to DLL search path. Please update to Python 3.8+ or make sure that hidapi.dll can be located by Python version that you use.`

PR #15993, opened 2024-06-13
lib: sms: Add support for Service Center Address

trantanen:
Added support for reading SMS Service Center Address (number) with AT+CSCA? and include it into the SMS PDU sent with AT+CMGS.
SMS sample made to use it.
Fixing gcf_sms library also related to AT+CSCA which responded with +CSMS due to a typo.

TODO list:

- [x] Add SMS library tests
- [x] Check gcf_sms library tests
- [x] Add gcf_sms configuration into sample.yaml
- [x] Add `AT+CSCA` documentation
- [x] Add mention of `AT+CSCA` or CONFIG_GCF_SMS into SMS sample and library documentation
- [ ] Add release note

Jira: NCSDK-27729

NOTE 27th Jun: This PR is kind of ready but will be finalized only after summer holidays in the end of July.

### DIFF
--- a/doc/nrf/libraries/modem/gcf_sms_lib.rst
+++ b/doc/nrf/libraries/modem/gcf_sms_lib.rst
@@ -13,6 +13,25 @@ The library implements the following AT commands:
 * ``AT+CPMS``
 * ``AT+CSMS``
 * ``AT+CSCA``
+
+  * Set command: ``+CSCA=<sca>[,<tosca>]``
+
+    * ``<sca>``: string
+    * ``<tosca>``: Type-of-Address octet in integer format, default ``145``.
+
+   .. code-block:: console
+
+      AT+CSCA=""+358501234567""
+      OK"
+
+  * Read command: ``+CSCA?``
+
+    .. code-block:: console
+
+       AT+CSCA?
+       +CSCA: ""+358501234567"",145"
+       OK
+
 * ``AT+CSCS``
 * ``AT+CMGD``
 * ``AT+CMSS``
--- a/doc/nrf/libraries/modem/sms.rst
+++ b/doc/nrf/libraries/modem/sms.rst
@@ -23,13 +23,21 @@ In addition, AT commands are also used to send SMS messages.
 SMS notifications are received using AT commands, but those are not visible for the users of this module.
 The module automatically acknowledges the SMS messages received on behalf of each listener.
 
+The SMS module reads the SMS service center number using the ``AT+CSCA?`` AT command.
+Most applications do not need to set a custom SMS service center number because the default value is enough.
+However, if your application requires a custom SMS service center number, you can set it using the ``AT+CSCA`` AT command.
+Because ``AT+CSCA`` AT command is not supported by the modem firmware, you can use the :ref:`lib_gcf_sms_readme` library.
+
 Configuration
 *************
 
-Configure the following Kconfig options when using this library:
+Configure the following mandatory Kconfig options when using this library:
 
-* :kconfig:option:`CONFIG_SMS` - Enables the SMS subscriber library.
+* :kconfig:option:`CONFIG_SMS` - Enables the SMS library.
 * :kconfig:option:`CONFIG_SMS_SUBSCRIBERS_MAX_CNT` - Sets the maximum number of SMS subscribers.
+
+Check and configure the optional :kconfig:option:`CONFIG_GCF_SMS` Kconfig option when using this library.
+It enables the :ref:`lib_gcf_sms_readme` library for the ``AT+CSCA`` AT command support to allow a custom SMS service center number.
 
 Limitations
 ***********
--- a/doc/nrf/libraries/modem/sms.rst
+++ b/doc/nrf/libraries/modem/sms.rst
@@ -23,9 +23,9 @@ In addition, AT commands are also used to send SMS messages.
 SMS notifications are received using AT commands, but those are not visible for the users of this module.
 The module automatically acknowledges the SMS messages received on behalf of each listener.
 
-The SMS module reads the SMS service center number using the ``AT+CSCA?`` AT command.
-Most applications do not need to set a custom SMS service center number because the default value is enough.
-However, if your application requires a custom SMS service center number, you can set it using the ``AT+CSCA`` AT command.
+The SMS module reads the SMS service center address, that is, phone number of the SMS service center, using the ``AT+CSCA?`` AT command.
+Most applications do not need to set a custom SMS service center address because the default value is enough.
+However, if your application requires a custom SMS service center address, you can set it using the ``AT+CSCA`` AT command.
 Because ``AT+CSCA`` AT command is not supported by the modem firmware, you can use the :ref:`lib_gcf_sms_readme` library.
 
 Configuration
@@ -37,7 +37,7 @@ Configure the following mandatory Kconfig options when using this library:
 * :kconfig:option:`CONFIG_SMS_SUBSCRIBERS_MAX_CNT` - Sets the maximum number of SMS subscribers.
 
 Check and configure the optional :kconfig:option:`CONFIG_GCF_SMS` Kconfig option when using this library.
-It enables the :ref:`lib_gcf_sms_readme` library for the ``AT+CSCA`` AT command support to allow a custom SMS service center number.
+It enables the :ref:`lib_gcf_sms_readme` library for the ``AT+CSCA`` AT command support to allow a custom SMS service center address.
 
 Limitations
 ***********
--- a/lib/gcf_sms/gcf_sms.c
+++ b/lib/gcf_sms/gcf_sms.c
@@ -29,8 +29,8 @@ LOG_MODULE_REGISTER(gcf_sms, CONFIG_GCF_SMS_LOG_LEVEL);
 #define SMS_BUFFER_ARRAY_SIZE (3)
 #define SMS_BUFFER_PDU_SIZE_MAX (352)
 
-/* Max size of Service Center Address. */
-#define SCA_SIZE_MAX (16)
+/* Max size of +CSCA response including Service Center Address. */
+#define CSCA_SIZE_MAX (32)
 
 /* Buffer for SMS messages. */
 struct sms_buffer {
@@ -42,7 +42,7 @@ struct sms_buffer {
 static struct sms_buffer sms_buffers[SMS_BUFFER_ARRAY_SIZE];
 
 /* Buffer for SCA. */
-static char sca_buff[SCA_SIZE_MAX + 1];
+static char sca_buff[CSCA_SIZE_MAX + 1];
 
 /* AT filters
  * Including all commands the filter should check for and the respective
@@ -281,16 +281,24 @@ static int at_cmd_callback_csms(char *buf, size_t len, char *at_cmd)
 
 static int at_cmd_callback_csca(char *buf, size_t len, char *at_cmd)
 {
+	char *csca_data;
+
 	/* Set */
 	if (prefix_match("AT+CSCA=", at_cmd)) {
-		strncpy(sca_buff, &(at_cmd[8]), SCA_SIZE_MAX);
-		sca_buff[sizeof(sca_buff) - 1] = '\0';
+		csca_data = &(at_cmd[8]);
+		if (strlen(csca_data) > CSCA_SIZE_MAX) {
+			LOG_ERR("CSCA data too long (%d) to fit in the buffer (%d)",
+				strlen(csca_data), CSCA_SIZE_MAX);
+			return at_cmd_custom_respond(buf, len, "ERROR\r\n");
+		}
+		sca_buff[0] = '\0';
+		strncat(sca_buff, &(at_cmd[8]), CSCA_SIZE_MAX);
 		return at_cmd_custom_respond(buf, len, "OK\r\n");
 	}
 
 	/* Read */
 	if (prefix_match("AT+CSCA?", at_cmd)) {
-		return at_cmd_custom_respond(buf, len, "+CSMS: %s\r\nOK\r\n", sca_buff);
+		return at_cmd_custom_respond(buf, len, "+CSCA: %s\r\nOK\r\n", sca_buff);
 	}
 
 	return at_cmd_custom_respond(buf, len, "ERROR\r\n");
@@ -455,8 +463,6 @@ static int at_cmd_callback_cmgf(char *buf, size_t len, char *at_cmd)
 {
 	/* Set */
 	if (prefix_match("AT+CMGF=0", at_cmd)) {
-		strncpy(sca_buff, &(at_cmd[8]), SCA_SIZE_MAX);
-		sca_buff[sizeof(sca_buff) - 1] = '\0';
 		return at_cmd_custom_respond(buf, len, "OK\r\n");
 	}
 
--- a/lib/sms/sms_submit.c
+++ b/lib/sms/sms_submit.c
@@ -63,7 +63,7 @@ static int sms_submit_encode_number(
 	__ASSERT_NO_MSG(number != NULL);
 
 	if (*number_size == 0) {
-		LOG_ERR("Number not given but zero length");
+		LOG_ERR("Number given but zero length");
 		return -EINVAL;
 	}
 
--- a/lib/sms/sms_submit.c
+++ b/lib/sms/sms_submit.c
@@ -41,6 +41,7 @@ LOG_MODULE_DECLARE(sms, CONFIG_SMS_LOG_LEVEL);
  * @param[out] encoded_number_size_octets Number of octets/bytes in encoded_number.
  *
  * @retval -EINVAL Invalid parameter.
+ * @retval -E2BIG Input number too long.
  * @return Zero on success, otherwise error code.
  */
 static int sms_submit_encode_number(
@@ -54,7 +55,7 @@ static int sms_submit_encode_number(
 	__ASSERT_NO_MSG(number != NULL);
 
 	if (*number_size == 0) {
-		LOG_ERR("SMS number not given but zero length");
+		LOG_ERR("Number not given but zero length");
 		return -EINVAL;
 	}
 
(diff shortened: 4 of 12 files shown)


lib/gcf_sms/gcf_sms.c
```diff
@@ -283,14 +283,14 @@ static int at_cmd_callback_csca(char *buf, size_t len, char *at_cmd)
...
-		strncpy(sca_buff, &(at_cmd[8]), SCA_SIZE_MAX);
+		strncpy(sca_buff, &(at_cmd[8]), CSCA_SIZE_MAX);
 		sca_buff[sizeof(sca_buff) - 1] = '\0';
 		return at_cmd_custom_respond(buf, len, "OK\r\n");
 	}
 
 	/* Read */
 	if (prefix_match("AT+CSCA?", at_cmd)) {
-		return at_cmd_custom_respond(buf, len, "+CSMS: %s\r\nOK\r\n", sca_buff);
+		return at_cmd_custom_respond(buf, len, "+CSCA: %s\r\nOK\r\n", sca_buff);
```
trantanen, line 293: @eivindj-nordic, was this intentionally `+CSMS` and if yes, why? Tests have the same so I'm starting to suspect that this is not a copy-paste bug.

lib/gcf_sms/gcf_sms.c
trantanen, line 1: @eivindj-nordic, is it fine to use this library for final product to handle those specific AT commands?
In this PR, I'm using `AT+CSCA`.

samples/cellular/sms/src/main.c
```diff
@@ -78,6 +80,25 @@ int main(void)
 		return 0;
 	}
 
+	/* Setting SMS service center number if set into the configuration,
+	 * or if it's left empty, an information text is printed.
```
tokangas, line 84: Is this comment correct? The code does not print anything if the service center number is empty.

samples/cellular/sms/src/main.c
```diff
@@ -78,6 +80,25 @@ int main(void)
...
+	 */
+	if (strcmp(CONFIG_SMS_SERVICE_CENTER_NUMBER, "")) {
```
tokangas, line 86: It's easy to read this condition so, that the code below is executed when the configured service center number is empty, when it's actually the other way around. `sizeof() > 1` would be clearer, I think. Not sure if using `strcmp()` even generates code, which is unnecessary.

samples/cellular/sms/src/main.c
trantanen: Copy-paste. Fixed.

samples/cellular/sms/src/main.c
trantanen: Good point. Changed here and for `CONFIG_SMS_SEND_PHONE_NUMBER` below.

lib/gcf_sms/gcf_sms.c
eivindj-nordic: Hi, it was intentional, I'll send you the (internal) reference.

lib/gcf_sms/gcf_sms.c
eivindj-nordic: As long as the application is satisfied with what it gets, it should not be a problem I think, though note that some of the responses are hardcoded. The library is implementing a minimum set of AT commands that is not supported in the modem, which is required to pass some GCF specifications.

samples/cellular/sms/Kconfig
```diff
@@ -9,6 +9,16 @@ menu "SMS Sample Settings"
 config SMS_SEND_PHONE_NUMBER
 	string "Phone number, including country code, where the SMS message is sent"
 	default ""
+	help
+	  Phone number, including country code, where the SMS message is sent.
+		If empty, SMS will not be sent at all but the sample waits for received messages.
```
tokangas, line 14:
```suggestion
	  Phone number, including country code, where the SMS message is sent.
	  If empty, SMS will not be sent at all but the sample waits for received messages.
```

samples/cellular/sms/Kconfig
```diff
@@ -9,6 +9,16 @@ menu "SMS Sample Settings"
...
+
+config SMS_SERVICE_CENTER_NUMBER
+	string "Phone number, including country code, of the SMS service center"
+	default ""
+	help
+		If empty, SMS will not add service center number into the SMS PDU,
+		and modem/network will use default values.
```
tokangas, line 21:
```suggestion
	  If empty, SMS will not add service center number into the SMS PDU,
	  and modem/network will use default values.
```

samples/cellular/sms/src/main.c
```diff
@@ -78,12 +80,29 @@ int main(void)
...
 
+	/* Setting SMS service center number if set into the configuration. */
+	if (sizeof(CONFIG_SMS_SERVICE_CENTER_NUMBER) > 1) {
+		int err = nrf_modem_at_cmd(
+				response, sizeof(response),
+				"AT+CSCA=\"%s\",91",
+				CONFIG_SMS_SERVICE_CENTER_NUMBER);
+		if (err) {
+			printk("Setting service center number %s failed, err=%d\n",
+				CONFIG_SMS_SERVICE_CENTER_NUMBER, err);
```
tokangas, line 91:
```suggestion
			printk("Setting service center number %s failed, err=%d\n",
			       CONFIG_SMS_SERVICE_CENTER_NUMBER, err);
```

samples/cellular/sms/src/main.c
```diff
@@ -78,12 +80,29 @@ int main(void)
...
+		} else {
+			printk(
+				"Setting service center number %s succeeded\n",
+				CONFIG_SMS_SERVICE_CENTER_NUMBER);
```
tokangas, line 95:
```suggestion
			printk("Setting service center number %s succeeded\n",
			       CONFIG_SMS_SERVICE_CENTER_NUMBER);
```

lib/gcf_sms/gcf_sms.c
trantanen: Yep, I understand the limitations. I'm only using AT+CSCA set and read functionalities and returning just the stored string is fine.

samples/cellular/sms/src/main.c
trantanen: The old way would be the accurate if parameters are on separate lines when they don't fit into the same line as function name but I know this will make people feel uncomfortable 😄

lib/gcf_sms/gcf_sms.c
trantanen: Seems this was not intentional and I'm discussing what fixes we do.

lib/sms/sms_submit.c
```diff
@@ -88,6 +104,80 @@ static int sms_submit_encode_number(
...
 }
 
+static void sms_submit_encode_sca(char *sca_field)
+{
+	int ret;
+
+	uint8_t encoded_sca[SMS_MAX_ADDRESS_LEN_CHARS + 1];
+	uint8_t encoded_sca_size;
+	uint8_t encoded_sca_size_octets = 0;
+	/* 36 is enough and has few characters extra. Maximum required should be:
```
tokangas, line 114: The size does not match the implementation anymore, you could consider removing the size of the buffer from the comment.

doc/nrf/libraries/modem/gcf_sms_lib.rst
```diff
@@ -13,6 +13,25 @@ The library implements the following AT commands:
 * ``AT+CPMS``
 * ``AT+CSMS``
 * ``AT+CSCA``
+
+  * Set command: +CSCA=<sca>[,<tosca>]
```
trantanen, line 17: I do not know what would be a better place to describe `AT+CSCA` command.
It has to be described because AT commands from GCF SMS library are not in AT Command Reference Guide.

trantanen:
Initial comments from documentation side would be nice. @peknis

doc/nrf/libraries/modem/gcf_sms_lib.rst
peknis, line 17:
```suggestion
  * Set command: ``+CSCA=<sca>[,<tosca>]``
```

doc/nrf/libraries/modem/gcf_sms_lib.rst
```diff
@@ -13,6 +13,25 @@ The library implements the following AT commands:
...
+
+    * <sca>: string
+    * <tosca>: Type-of-Address octet in integer format. Default is 145.
```
peknis, line 20:
```suggestion
    * ``<sca>``: string
    * ``<tosca>``: Type-of-Address octet in integer format, default ``145``.
```

doc/nrf/libraries/modem/gcf_sms_lib.rst
```diff
@@ -13,6 +13,25 @@ The library implements the following AT commands:
...
+      AT+CSCA=""+358501234567""
+      OK"
+
+  * Read command: +CSCA?
+
+   .. code-block:: console
+
+      AT+CSCA?
+      +CSCA: ""+358501234567"",145"
+      OK
```
peknis, line 33:
```suggestion
  * Read command: ``+CSCA?``

    .. code-block:: console

       AT+CSCA?
       +CSCA: ""+358501234567"",145"
       OK
```

doc/nrf/libraries/modem/sms.rst
```diff
@@ -23,14 +23,23 @@ In addition, AT commands are also used to send SMS messages.
 SMS notifications are received using AT commands, but those are not visible for the users of this module.
 The module automatically acknowledges the SMS messages received on behalf of each listener.
 
+The SMS module reads SMS service center number using ``AT+CSCA?`` AT command.
+Most applications do not need to set custom SMS service center number but default value is enough.
+However, if your application requires custom SMS service center number, you can set it using ``AT+CSCA`` AT command.
+Because ``AT+CSCA`` AT command is not supported by the modem firmware, you can use the :ref:`lib_gcf_sms_readme` library.
```
peknis, line 29:
```suggestion
The SMS module reads the SMS service center number using the ``AT+CSCA?`` AT command.
Most applications do not need to set a custom SMS service center number because the default value is enough.
However, if your application requires a custom SMS service center number, you can set it using the ``AT+CSCA`` AT command.
Because ``AT+CSCA`` AT command is not supported by the modem firmware, you can use the :ref:`lib_gcf_sms_readme` library.
```

doc/nrf/libraries/modem/sms.rst
```diff
@@ -23,14 +23,23 @@ In addition, AT commands are also used to send SMS messages.
...
-Configure the following Kconfig options when using this library:
+Configure the following mandatory Kconfig options when using this library:
 
-* :kconfig:option:`CONFIG_SMS` - Enables the SMS subscriber library.
+* :kconfig:option:`CONFIG_SMS` - Enables the SMS library.
 * :kconfig:option:`CONFIG_SMS_SUBSCRIBERS_MAX_CNT` - Sets the maximum number of SMS subscribers.
 
+Check and configure the following optional Kconfig option when using this library:
+
+* :kconfig:option:`CONFIG_GCF_SMS` - Enables the :ref:`lib_gcf_sms_readme` library for ``AT+CSCA`` AT command support to allow custom SMS service center number.
```
peknis, line 41:
```suggestion
Check and configure the optional :kconfig:option:`CONFIG_GCF_SMS` Kconfig option when using this library.
It enables the :ref:`lib_gcf_sms_readme` library for the ``AT+CSCA`` AT command support to allow a custom SMS service center number.
```

samples/cellular/sms/Kconfig
```diff
@@ -9,6 +9,16 @@ menu "SMS Sample Settings"
 config SMS_SEND_PHONE_NUMBER
 	string "Phone number, including country code, where the SMS message is sent"
 	default ""
+	help
+	  Phone number, including country code, where the SMS message is sent.
+	  If empty, SMS will not be sent at all but the sample waits for received messages.
+
+config SMS_SERVICE_CENTER_NUMBER
+	string "Phone number, including country code, of the SMS service center"
```
peknis, line 17:
```suggestion
	string "Phone number of the SMS service center, including the country code."
```

samples/cellular/sms/Kconfig
```diff
@@ -9,6 +9,16 @@ menu "SMS Sample Settings"
 config SMS_SEND_PHONE_NUMBER
 	string "Phone number, including country code, where the SMS message is sent"
 	default ""
+	help
+	  Phone number, including country code, where the SMS message is sent.
```
peknis, line 13:
```suggestion
	  Phone number to which the SMS message is sent, including the country code.
```

samples/cellular/sms/Kconfig
```diff
@@ -9,6 +9,16 @@ menu "SMS Sample Settings"
 config SMS_SEND_PHONE_NUMBER
 	string "Phone number, including country code, where the SMS message is sent"
```
peknis, line 10:
```suggestion
	string "Phone number to which the SMS message is sent, including the country code."
```

samples/cellular/sms/Kconfig
```diff
@@ -9,6 +9,16 @@ menu "SMS Sample Settings"
...
+	help
+	  Phone number, including country code, where the SMS message is sent.
+	  If empty, SMS will not be sent at all but the sample waits for received messages.
+
+config SMS_SERVICE_CENTER_NUMBER
+	string "Phone number, including country code, of the SMS service center"
+	default ""
+	help
+	  If empty, SMS will not add service center number into the SMS PDU,
+	  and modem/network will use default values.
```
peknis, line 21:
```suggestion
	  If empty, SMS will not add the service center number into the SMS PDU,
	  and the modem/network will use the default values.
```

samples/cellular/sms/README.rst
```diff
@@ -27,6 +27,8 @@ The sample requires an LTE connection.
 When the sample starts, it sends an SMS if a recipient phone number is set in the configuration.
 The sample then receives all the SMS messages and displays the information about the messages including the text that is sent.
 
+Custom SMS service center number can be set through configuration if needed.
```
peknis, line 30:
```suggestion
If you need a custom SMS service center number, you can set it through the configuration.
```

peknis (changes requested):

lib/gcf_sms/gcf_sms.c
trantanen: I will close this comment as we have decided to fix the typo.

peknis (approved):

lib/sms/sms_submit.c
```diff
@@ -88,6 +104,93 @@ static int sms_submit_encode_number(
...
+	 * - double quotes for address: 2
+	 * - comma: 1
+	 * - Type of address: 3
+	 * - \r\nOK\r\n: 6
+	 * - NUL termination: 1
+	 */
+	char csca_resp[48] = { 0 };
+	/* +1 for NUL termination and another +1 for potential '+' sign */
+	char sca[SMS_MAX_ADDRESS_LEN_CHARS + 2] = { 0 };
+	int tosca = 0x91;
```
tokangas, line 139: This value is in decimal in all other places, it could be clearer to use `145` here. Maybe also add a comment where this value comes from.

trantanen: I'm adding a comment and changed it to decimal format. However, it's in decimal format in `AT+CSCA` AT command, but in hexadecimal format in the encoded number in `AT+CMGS` (see `sms_submit_encode`)

lib/sms/sms_submit.c
```diff
@@ -88,6 +104,102 @@ static int sms_submit_encode_number(
...
+ *   - 3GPP TS 27.005 chapter 3.1 (see <pdu>, <tosca> and <toda>)
+ *   - 3GPP TS 27.005 chapter 3.3.1
+ *   - 3GPP TS 27.005 chapter 4.3
+ *
+ * @param[out] sca_field Buffer for SCA field.
+ *                       Must be at least 23 bytes long to include the following information:
+ *                       - Address/number: 20
+ *                       - Address length: 1
+ *                       - Type of address: 1
+ *                       - NUL termination: 1
```
eivindj-nordic, line 122: nit: 
```suggestion
 *                       - NULL termination: 1
```
Here and others

trantanen: Hmmm, NULL and NUL termination is used in a mixed manner. Google tells that NUL is the name of the '\0' character and hence probably more correct here although there are reasons why you would call it NULL termination.

doc/nrf/libraries/modem/sms.rst
```diff
@@ -23,14 +23,22 @@ In addition, AT commands are also used to send SMS messages.
 SMS notifications are received using AT commands, but those are not visible for the users of this module.
 The module automatically acknowledges the SMS messages received on behalf of each listener.
 
+The SMS module reads the SMS service center number using the ``AT+CSCA?`` AT command.
```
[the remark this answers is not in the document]
trantanen: I've been struggling to select the terminology for the document for the phone number (and service center number is the same). 3GPP talks about destination address and source address instead of recipient and sender phone numbers. I've kind of made a choice to talk about "number" instead of "address" in the documentation to be more understandable for the average user of SMS library who doesn't know the protocol or 3GPP details.

Although I realize word "address" is used in the the code and even in some parts of the public API, I've stick to the "number" in the documentation of the SMS library.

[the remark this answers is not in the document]
eivindj-nordic: ok, I found it a bit confusing mixing terminology of service center number and service center address, as the latter is what I've seen elsewhere in our documentation etc.. So I thought it might be better to align, but it is ok by me to keep it as is.

lib/sms/sms_submit.c
```diff
@@ -88,6 +104,102 @@ static int sms_submit_encode_number(
...
+ */
+static void sms_submit_encode_sca(char *sca_field)
```
trantanen, line 124: This function (which is most of this PR) takes flash as follows for configurations not using CSCA, i.e., not include gcf_sms library and using SMS as earlier:
* 760 bytes with `CONFIG_SMS_LOG_LEVEL_DBG=y`,
* 432 bytes with default logging, that is, `CONFIG_SMS_LOG_LEVEL_INF=y`
* 320 bytes with `CONFIG_LOG=n`

I was thinking of making a Kconfig for `AT+CSCA` usage to optimize flash usage but I'm hesitant given it's only 320 bytes.

doc/nrf/libraries/modem/sms.rst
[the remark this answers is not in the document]
trantanen: Äh, seems both terms are used in the public API. I'll change this to address and mention that it means the service center phone number so everybody should be able to understand it.

trantanen:
This is closed as it wasn't needed after all in this form because this solution doesn't store SMS service center address into persistent memory.

Another solution was made to read SCA from the SIM for type approval purposes:
PR #17152

PR #16014, opened 2024-06-14
Fast Pair: locator tag use case: upgrade to "Supported" maturity level

kapi-no:


doc/nrf/releases_and_maturity/software_maturity.rst
miha-nordic, line 1: Everything for nrf54L15 should be marked as supported.
Should nrf5340 support for Fast Pair features follows Locator tag maturity and should be experimental?

miha-nordic (changes requested):

doc/nrf/releases_and_maturity/software_maturity.rst
kapi-no: The locator tag uses only the following FP features: "Initial pairing" and "FMDN extension".

kapi-no: The use case level support for nRF5340 and nRF54L15 platforms is considered experimental as they ideally should support the "Security provisioning" feature and keep the provisioned data as part of the Secure Storage. Apart from that, nRF54L15 SoC is experimental so we shouldn't encourage customers to deploy experimental SoC into their products.

Individual features like "Initial pairing" and "FMDN extension" are not platform-specific so they are declared as supported for all platforms if they are deployed in products for at least one use case and SoC (e.g. Locator Tag, nRF52840 SoC). 

Platform-specific features (e.g. "Secure provisioning") can differ for various SoCs.

I am happy to change the approach, but I just wanted to describe the wider context so that we can make a more informed decision. I would love to hear input from others.

@alstrzebonski, @MarekPieta, @PaalKastnes

alstrzebonski: I agree with @kapi-no. I think there's no reason to claim that e.g. "Initial pairing" feature support is experimental on one platform and supported on another platform. The platform itself might be experimental, but the feature is not. That's why we have two tables.

alstrzebonski (approved):

doc/nrf/releases_and_maturity/software_maturity.rst
miha-nordic: @kapi-no @alstrzebonski I understand but then it shouldn't be divided through platforms at all. If you leave it like that you give false impression that anything is production ready for nrf54L15 when it is still SOC1 not even checked on FP1 targets. So you have production ready support on experimental HW. That doesn't make sense in my opinion. If you remove this division into platforms then it will be fine for me.

kapi-no: @miha-nordic, documenting the software maturity level for every platform is the NCS documentation policy for this type of information. It is used for other protocols, libraries and NCS modules. The Fast Pair software maturity table follows this documentation policy. We don't want to be the outlier here.

miha-nordic: Then it can't be supported for the platforms that it is not production ready.

miha-nordic:
@kapi-no This section: Fast Pair feature needs to be change I don't approve it in this shape.

miha-nordic (changes requested):
This section: Fast Pair feature needs to be change I don't approve it in this shape.

kapi-no:
@miha-nordic, please request specific changes that you think are mandatory. This way, we can progress with our discussion

miha-nordic:
> @miha-nordic, please request specific changes that you think are mandatory. This way, we can progress with our discussion

@kapi-no   As mention before:
- Initial pairing needs to be marked as experimental for nrf5340 and nrf54l15
- Find My Device Network extension needs to be marked as experimental for nrf5340 and nrf54l15

MarekPieta (approved):
I would suggest to apply my comment related to nRF54L to align with @miha-nordic request

kapi-no:
> > @miha-nordic, please request specific changes that you think are mandatory. This way, we can progress with our discussion
> 
> @kapi-no As mention before:
> 
> * Initial pairing needs to be marked as experimental for nrf5340 and nrf54l15
> * Find My Device Network extension needs to be marked as experimental for nrf5340 and nrf54l15

Done

miha-nordic (approved):

MarekPieta (approved):

alstrzebonski (approved):

kapi-no:
@divipillai or @annwoj, could you ACK this PR? This is technically a documentation update.

PR #16020, opened 2024-06-14
Sml update for Wifi 

VivekUppunda:


doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -104,6 +104,14 @@ features:
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
+    Wi-Fi + Thread Co-existence:
```
sachinthegreen, line 107:
```suggestion
    Thread Co-existence:
```

sachinthegreen (changes requested):

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -104,6 +104,14 @@ features:
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
+    Thread Co-existence:
+      rule: WIFI_NRF700X && NRF700X_THREAD_COEX
+      boards_and_shields:
+        - SHIELD_NRF7002EK
+        - SHIELD_NRF7002EK_NRF7001
+        - SHIELD_NRF7002EB
+        - BOARD_NRF7002DK_NRF5340_CPUAPP
```
bama-nordic, line 113: Presently it has been tested only on `BOARD_NRF7002DK_NRF5340_CPUAPP` - please remove others from the supported list. We'll add once we have tested them.

bama-nordic (changes requested):

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -104,6 +104,10 @@ features:
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
+    Thread Co-existence:
```
richabp, line 107: Can you also change line 99 as **Bluetooth LE coexistence** ?
```suggestion
    Thread coexistence:
```

richabp (changes requested):

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -96,14 +96,18 @@ features:
...
         - SHIELD_NRF7002EK
         - SHIELD_NRF7002EK_NRF7001
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
+    Thread coexistence:
+      rule: WIFI_NRF700X && NRF700X_THREAD_COEX
+      boards_and_shields:
+        - BOARD_NRF7002DK_NRF5340_CPUAPP
     SoftAP Mode:
```
sachinthegreen, line 111: Can a note be added to this as a superscript with text saying `Support limited to what is essential for Wi-Fi provisioning` ?

sachinthegreen (changes requested):

doc/_scripts/software_maturity/software_maturity_features.yaml
VivekUppunda: Updated the SoftAP mode tag as discussed

b-gent (approved):

sachinthegreen (approved):

richabp (approved):

doc/_scripts/software_maturity/software_maturity_features.yaml
```diff
@@ -96,15 +96,19 @@ features:
...
       rule: WIFI_NRF700X && NRF700X_SR_COEX
       boards_and_shields:
         - SHIELD_NRF7002EK
         - SHIELD_NRF7002EK_NRF7001
         - SHIELD_NRF7002EB
         - BOARD_NRF7002DK_NRF5340_CPUAPP
         - BOARD_NRF7002DK_NRF5340_CPUAPP_NRF7001
-    SoftAP Mode:
+    Thread coexistence:
+      rule: WIFI_NRF700X && NRF700X_THREAD_COEX
```
krish2718, line 108: rather than a dummy Kconfig we can use `WIFI_NRF700X && OPENTHREAD` this is the combination for OTBR where thread co-ex comes in to picture.

krish2718: Also the SoftAP change should be in a separate commit, it's not matching commit log

drivers/wifi/nrf700x/Kconfig
```diff
@@ -241,6 +241,10 @@ config NRF700X_SR_COEX
 	bool "Enable Wi-Fi and SR coexistence support"
 	def_bool $(dt_nodelabel_enabled,nrf_radio_coex)
 
+config NRF700X_THREAD_COEX
```
krish2718, line 244: This should be before the SML change as it depends on this, but anyways we don't need this change as suggested above.

krish2718 (changes requested):

doc/_scripts/software_maturity/software_maturity_features.yaml
VivekUppunda: I have modified the same as provided in the comment and moved SoftAP and bluetooth co-existence naming to a new commit

drivers/wifi/nrf700x/Kconfig
VivekUppunda: The Config is updated and the commit is moved down accordingly

samples/wifi/thread_coex/sample.yaml
```diff
@@ -10,6 +10,7 @@ tests:
       - nrf7002dk/nrf5340/cpuapp
     extra_args: CONFIG_MPSL_CX=y ipc_radio_CONFIG_MPSL_CX=y
                 CONFIG_COEX_SEP_ANTENNAS=y
+                CONFIG_OPENTHREAD=y
```
krish2718, line 13: This isn't needed, `overlay-ot.conf` is enabled for all which enables `NET_L2_OPENTHREAD` which selects `OPENTHREAD`.

@muraliThokala as the sample itself it thread co-ex, there is no point in having a separate `overlay-ot.conf`, we should merge it with `prj.conf`. WDYT?

VivekUppunda: Removed the CONFIG from sample.yaml

krish2718 (approved):

PR #16033, opened 2024-06-17
manifest: update suit-generator revision

e-rk:
The envelope component digest algorithm is now separate from the authentication block algorithm of the child envelope.


tomchy:
Replaced by: https://github.com/nrfconnect/sdk-nrf/pull/16383

PR #16049, opened 2024-06-17
partition_manager: Add support for nrf54l15_cpuflpr with vpr_launcher

oyvindronningstad:
Add case for cpuflpr_sram in PM_SRAM_(BASE|SIZE)

Add pm.yml.vpr_launcher file

Include a noup patch in Zephyr that overwrites FLASH_BASE_ADDRESS and
FLASH_SIZE to allow Partition Manager to use the whole RRAM instead of
trying to put both the flpr firmware and the vpr_launcher into the flpr
RRAM partition.

### DIFF
--- a/Kconfig.nrf
+++ b/Kconfig.nrf
@@ -109,6 +109,17 @@ config NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE
 config GETOPT
 	default n
 
+# Temporary hack to be able to build samples and tests on the nRF51L15 FLPR core
+if SOC_NRF54L15_ENGA_CPUFLPR && PARTITION_MANAGER_ENABLED
+config FLASH_BASE_ADDRESS
+	hex
+	default 0x0
+config FLASH_SIZE
+	int
+	default 1524
+endif
+
+
 rsource "samples/Kconfig"
 rsource "subsys/Kconfig"
 rsource "modules/Kconfig"
--- a/subsys/partition_manager/CMakeLists.txt
+++ b/subsys/partition_manager/CMakeLists.txt
@@ -130,6 +130,10 @@ if (CONFIG_NRF_MODEM_LIB_TRACE_BACKEND_FLASH)
   ncs_add_partition_manager_config(pm.yml.modem_trace)
 endif()
 
+if (CONFIG_SOC_NRF54L15_ENGA_CPUFLPR)
+  ncs_add_partition_manager_config(pm.yml.vpr_launcher)
+endif()
+
 # We are using partition manager if we are a child image or if we are
 # the root image and the 'partition_manager' target exists.
 if (SYSBUILD)
--- a/subsys/partition_manager/Kconfig
+++ b/subsys/partition_manager/Kconfig
@@ -120,6 +120,16 @@ rsource "Kconfig.template.partition_config"
 rsource "Kconfig.template.partition_region"
 endif
 
+if SOC_NRF54L15_ENGA_CPUFLPR
+config PM_PARTITION_SIZE_VPR_LAUNCHER
+	hex
+	default $(dt_node_reg_addr_hex,/soc/rram-controller@5004b000/rram@165000)
+	help
+	  Memory set aside for the vpr_launcher partition.
+	  Must match the size of the cpuapp_rram partition which is deleted.
+	  This to place the flpr app's code at the address found in the devicetree.
+endif
+
 endmenu # Zephyr subsystem configurations
 menu "NCS subsystem configurations"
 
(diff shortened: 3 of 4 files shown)


subsys/partition_manager/CMakeLists.txt
```diff
@@ -130,6 +130,10 @@ if (CONFIG_NRF_MODEM_LIB_TRACE_BACKEND_FLASH)
   ncs_add_partition_manager_config(pm.yml.modem_trace)
 endif()
 
+if (SB_CONFIG_VPR_LAUNCHER) # Currently doesn't work
```
nordicjm, line 133: this will never be visible, the file here is something that runs per image, it does not run at a sysbuild context, those files are in cmake/sysbuild

oyvindronningstad: I guessed as much, so we need something that's visible here, how can we do that?

masz-nordic: I agree it should not be based on `SB_CONFIG_VPR_LAUNCHER`, we need to support both: builds with vpr_launcher and standalone ones.
Not sure about how PM works, so I'm most likely missing something - is it possible to use two separate yaml files for APP build and FLPR build?

oyvindronningstad: Ideally, you build everything at once. The point of PM is to have full knowledge of the system and partition things accordingly. In that case, you would use cpuapp as the top-level board instead of cpuflpr, so it would also solve the other problem I found. In that case you could always add the PM file for nrf54l15, and the user could optionally remove the flpr partition by setting its size to 0 in Kconfig.

But I understand that the current way is practical when developing for cpuflpr. We could maybe generalize the partition to something like "cpuapp_fw" since you would always need some FW for the appcore. I'm looking into making the necessary DTS overlays for building in this way.

oyvindronningstad: Changed it now

oyvindronningstad:
The current solution works when using the xip variant of cpuflpr. Need to fix some RAM stuff for the default variant to work.

nordic-segl (approved):

nvlsianpu:
@michalek-no Can you review this?

Kconfig.nrf
```diff
@@ -109,6 +109,17 @@ config NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE
 config GETOPT
 	default n
 
+# Hack to be able to run tests on the flpr core
```
hakonfam, line 112: Is there some kconfig we could check here to ensure that it is a test that is being built? Should we do this in `test/Kconfig` instead of here?

subsys/partition_manager/Kconfig
```diff
@@ -205,11 +205,13 @@ config PM_OVERRIDE_EXTERNAL_DRIVER_CHECK
 config PM_SRAM_BASE
 	hex
 	default $(dt_node_reg_addr_hex,/soc/memory@21000000) if SOC_NRF5340_CPUNET
+	default 0x20000000 if SOC_NRF54L15_ENGA_CPUFLPR
```
hakonfam, line 208: I assume we don't have a DT node available for this?

subsys/partition_manager/Kconfig
```diff
@@ -205,11 +205,13 @@ config PM_OVERRIDE_EXTERNAL_DRIVER_CHECK
...
 	default $(dt_node_reg_addr_hex,/soc/memory@20000000)
 
 config PM_SRAM_SIZE
 	hex
 	default $(dt_node_reg_size_hex,/soc/memory@21000000) if SOC_NRF5340_CPUNET
+	default 0x40000 if SOC_NRF54L15_ENGA_CPUFLPR
```
hakonfam, line 214: Don't we have a DT node for this?

subsys/partition_manager/pm.yml.vpr_launcher
```diff
@@ -0,0 +1,7 @@
+vpr_launcher:
```
hakonfam, line 1: Should we have kconfig options for these values so that they can be changed from kconfig instead of directly modifying these files?

hakonfam (approved):
Only nits so no blockers. I assume that this will be replaced for next release anyway.

subsys/partition_manager/Kconfig
oyvindronningstad: It's deleted

subsys/partition_manager/pm.yml.vpr_launcher
oyvindronningstad: Currently they must match DT.

subsys/partition_manager/Kconfig
oyvindronningstad: It's also deleted in nrf54l15_cpuflpr.dtsi.

Kconfig.nrf
oyvindronningstad: I suppose it's also used to test samples, so I don't think we can rely on a kconfig.

subsys/partition_manager/pm.yml.vpr_launcher
oyvindronningstad: Added Kconfig that takes value from DT.

Kconfig.nrf
```diff
@@ -109,6 +109,17 @@ config NORDIC_QSPI_NOR_FLASH_LAYOUT_PAGE_SIZE
 config GETOPT
 	default n
 
+# Temporary hack to be able to build samples and tests on the nRF51L15 FLPR core
```
tejlmand, line 112: Please ensure this is addressed in in a followup task.

subsys/partition_manager/Kconfig
```diff
@@ -205,11 +215,13 @@ config PM_OVERRIDE_EXTERNAL_DRIVER_CHECK
 config PM_SRAM_BASE
 	hex
 	default $(dt_node_reg_addr_hex,/soc/memory@21000000) if SOC_NRF5340_CPUNET
+	default 0x20000000 if SOC_NRF54L15_ENGA_CPUFLPR  # Because the cpuapp_sram DT node is deleted in nrf54l15_cpuflpr.dtsi
```
tejlmand, line 218: this is not ideal, please ensure this is part of the follow up issues created to remove the temporary solution.

tejlmand (approved):
Approved as temporary solution, but please ensure comments are addressed in followup work.

oyvindronningstad:
CI is green in Jenkins

PR #16050, opened 2024-06-17
connectivity_bridge: enable debug functionality

maxd-nordic:
This patch enables debug functionality in the connectivity bridge using a CMSIS-DAP v2.0 compatible debug probe. This can be used with pyocd, nRF91 support has been implemented there.


maxd-nordic:
TODO: this adds 100uA power usage - need to find the source and fix.
Possible culprits:
* GPIO
* USB
* app control flow

PR #16052, opened 2024-06-17
modules: TF-M: Allow RRAMC configuration from the non-secure app

Vge0rge:
Adds a service which allows writing to secure memory from the non-secure application for specific memory locations.

Adds an option to configure the RRAMC controller to low power mode from the non-secure application.

Ref: NCSDK-27159

### DIFF
--- a/modules/trusted-firmware-m/tfm_boards/services/include/tfm_platform_user_memory_ranges.h
+++ b/modules/trusted-firmware-m/tfm_boards/services/include/tfm_platform_user_memory_ranges.h
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
 
-#ifndef TFM_READ_RANGES_H__
-#define TFM_READ_RANGES_H__
+#ifndef TFM_PLATFORM_USER_MEMORY_RANGES_H__
+#define TFM_PLATFORM_USER_MEMORY_RANGES_H__
 
 #include <pm_config.h>
 
@@ -76,4 +76,9 @@ static const struct tfm_read_service_range ranges[] = {
 #endif
 };
 
-#endif /* TFM_READ_RANGES_H__ */
+static const struct tfm_write32_service_address tfm_write32_service_addresses[] = {
+	/* This is a dummy value because this table cannot be empty */
+	{.addr = 0xFFFFFFFF, .mask = 0x0, .allowed_values = NULL, .allowed_values_array_size = 0},
+};
+
+#endif /* TFM_PLATFORM_USER_MEMORY_RANGES_H__ */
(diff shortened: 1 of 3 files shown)


mswarowsky (approved):

vili-nordic (approved):

modules/trusted-firmware-m/tfm_boards/services/include/tfm_platform_user_memory_ranges.h
```diff
@@ -76,4 +76,26 @@ static const struct tfm_read_service_range ranges[] = {
...
+ */
+#define NRF_RRAMC_LOWPOWER_CONFIG_ADDR (NRF_RRAMC_S_BASE + 0x518)
+
+/* Values: 0x0 = Power down mode, 0x1 = Standby mode */
+static const uint32_t rramc_lowpower_config_allowed[] = {0x0, 0x1};
+#endif /* NRF_RRAMC_S */
+
+static const struct tfm_write32_service_address tfm_write32_service_addresses[] = {
+	/* This is a dummy value because this table cannot be empty */
+	{.addr = 0xFFFFFFFF, .mask = 0x0, .allowed_values = NULL, .allowed_values_array_size = 0},
```
tomi-font, line 92: Why not have this only `#ifndef NRF_RRAMC_S`?

tomi-font (approved):

modules/trusted-firmware-m/tfm_boards/services/include/tfm_platform_user_memory_ranges.h
Vge0rge: Do you refer to the dummy value here?

tomi-font: Yes.

Vge0rge: Now doing this it makes sense, but afterwards I expect that this list will be expanded with more entries and having an ifndef for all the other entries doesn't seem very good to me.

tomi-font: Of course it can then be modified appropriately later on. I would just leave out those entries that are not needed by principle (and heh, saves a few bytes). But I'm not familiar with this code anyway. Just thinking out loud.

Vge0rge: So the thing is that we need at least one entry, otherwise the build will fail all together. I added an  if/else to avoid having the dummy value when we have rramc defined and we can decide what do when we have more entries later. This will save some bytes when RRAMC is enable but it will still add the dummy value when the RRAMC is disabled. Saving these bytes requires allowing to enable/disable specific platform services which is not supported at the moment and it is not in scope for this PR.

PR #16072, opened 2024-06-18
suit: add support for more key types and add key path kconfig

e-rk:
Added support for more key types in the sign script. Added option to set the path to the private key file.


tejlmand (approved):

nordicjm (approved):

tomchy:
Replaced by: https://github.com/nrfconnect/sdk-nrf/pull/16383

PR #16076, opened 2024-06-18
bl_crypto: cc310: Do not initialize cc310 from EXT_ABI

sigvartmh:
Fix-up for 723194f47bfdc3572c3e5cb27eeab74df23522b9 to add handling for a case where you call this function when you are running from PSP and have modified MSP so that you would still initialize the cc310.

Ref. NCSDK-23962

### DIFF
--- a/subsys/bootloader/bl_crypto/bl_crypto_cc310_common.c
+++ b/subsys/bootloader/bl_crypto/bl_crypto_cc310_common.c
@@ -33,8 +33,11 @@ int cc310_bl_init(void)
 #if (defined(CONFIG_BL_ROT_VERIFY_EXT_API_ENABLED) || defined(CONFIG_BL_SHA256_EXT_API_ENABLED) \
 	|| defined(CONFIG_BL_SECP256R1_EXT_API_ENABLED)) && defined(PM_B0_END_ADDRESS)
 	uint32_t msp = __get_MSP();
+	uint32_t psp = __get_PSP();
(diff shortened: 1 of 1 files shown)


hakonfam (approved):
So if invoked by EXT_ABI, then the PSP is selected, and if its invoked by B0 directly then MSP is selected? Should we add a comment describing this?

sigvartmh:
No. It checks if you are coming from outside of B0. So it's more if you call ext abi from an external context like app or MCUBoot.

Before you could skip this check by changing to PSP then call the function from PSP context

sigvartmh:
Not enough approvals so removing it from 2.7

oyvindronningstad (changes requested):
CONTROL.SPSEL doesn't alone specify whether PSP is being used. If execution is in handler mode, MSP is always used, regardless of CONTROL.SPSEL.

pdunaj:
I wonder why it is done the way it is. Can't it be called from some place that we know it's B0? Since it is clearly done once it does not look like you have to reinitialize on every call, right?

sigvartmh:
> CONTROL.SPSEL doesn't alone specify whether PSP is being used. If execution is in handler mode, MSP is always used, regardless of CONTROL.SPSEL.

So it should check that MSP is larger regardless then and only check PSP if SPSEL is set.

oyvindronningstad:
You must check if you are in handler mode (i.e. executing an interrupt). I don't quite remember how you can do that.

PR #16087, opened 2024-06-19
sysbuild: update image specific names to use DEFAULT_IMAGE

tejlmand:
The image name is hard coded for dependency configuration, but when using sysbuild the image name of the default image is based on the name of the folder. This means that if sample is copied elsewhere for local modification the hard coded name may no longer match the default image name.

Improve this by using DEFAULT_IMAGE variable instead of hard coded name.

Jira: NCSDK-28028

### DIFF
--- a/applications/machine_learning/sysbuild.cmake
+++ b/applications/machine_learning/sysbuild.cmake
@@ -31,8 +31,8 @@ if(SB_CONFIG_PARTITION_MANAGER)
 endif()
 
 # Add a dependency so that the remote image will be built first.
-sysbuild_add_dependencies(CONFIGURE machine_learning ipc_radio remote)
+sysbuild_add_dependencies(CONFIGURE ${DEFAULT_IMAGE} ipc_radio remote)
 # Add dependency so that the remote image is flashed first.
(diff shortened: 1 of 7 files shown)


shanthanordic (approved):

sigvartmh (approved):

shanthanordic:
add DNM to get some extra test covera

thst-nordic:
> add DNM to get some extra test covera

all the HW tests passed except 1 which was missing twister artifact.   
I think ok to remove `ci-all-test`

nordicjm (approved):

nordicjm:
> Getting CI green is not happening. It is not abug but more an annoyancem so pushing it out of release scope.

Without these changes, if these are copied out using vscode then they will not work, adding @bencefr for comment

bencefr:
> > Getting CI green is not happening. It is not abug but more an annoyancem so pushing it out of release scope.
> 
> Without these changes, if these are copied out using vscode then they will not work, adding @bencefr for comment

Not my requirement, if this doesn't make the release, it doesn't.

nordicjm (approved):

PR #16090, opened 2024-06-19
nrf_wifi: Remove dependency on OSAL layer handle

sachinthegreen:
Removes the requirement for the different layers in the OS agnostic parts of the driver having to maintain a handle to the OS interface layer in order to call the OS interface calls. The OS interface layer now maitains the handle to OS-specific ops internally and invokes the appropriate functions.

Fixes SHEL-2639

### DIFF
--- a/drivers/wifi/nrf700x/src/fmac_main.c
+++ b/drivers/wifi/nrf700x/src/fmac_main.c
@@ -43,6 +43,7 @@
 LOG_MODULE_DECLARE(wifi_nrf, CONFIG_WIFI_NRF700X_LOG_LEVEL);
 
 struct nrf_wifi_drv_priv_zep rpu_drv_priv_zep;
+extern const struct nrf_wifi_osal_ops nrf_wifi_os_zep_ops;
 
 /* 3 bytes for addreess, 3 bytes for length */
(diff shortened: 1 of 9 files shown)


krish2718 (commented):
if we consider the FMAC API as the public interface, then this needs a changelog + migration guide entry.

sachinthegreen:
> if we consider the FMAC API as the public interface, then this needs a changelog + migration guide entry.

This is not changing any of the FMAC API declarations, except `nrf_wifi_phy_rf_params_init`, but I think that function has been incorrectly exposed as an API...it needs to be `static`

krish2718:
> > if we consider the FMAC API as the public interface, then this needs a changelog + migration guide entry.
> 
> This is not changing any of the FMAC API declarations, except `nrf_wifi_phy_rf_params_init`, but I think that function has been incorrectly exposed as an API...it needs to be `static`

I was referring to the exception itself :). And there are also utils API e.g., `nrf_wifi_utils_is_mac_addr_valid`, `nrf_wifi_utils_q_len` etc those are changed in this PR.

krish2718 (approved):

rado17 (approved):

VivekUppunda (approved):

kapbh (approved):

PR #16098, opened 2024-06-20
sysbuild: Fix firmware loader for MCUboot support

nordicjm:
Adds support for having an image designated as a firmware loader image

NCSDK-27560


nvlsianpu:
I removed the milestone as this is to late to have it in release scope.

shanthanordic:
I think even if it is fix, such changes are coming very latem, pushing it out of NCS 2.7.0

nvlsianpu (approved):

PR #16099, opened 2024-06-20
Bluetooth: Define connection event nofitication callbacks

rugeGerritsen:
The connection event callback feature replaces
the radio notification feature. It can be used to
synchronize data sampling with Bluetooth connection events.

The old feature had several weaknesses:
 - It was not possible to know which connection
   the notification belonged to.
 - The trigger also happened upon flash access.
 - The time from trigger until event start was fixed.
   That is, not configurable per use case.

The new feature is built on top of the connection event
trigger feature and is therefore able to known
which connection a given notification belongs to.

This PR is done on top of https://github.com/nrfconnect/sdk-nrf/pull/15798 to simplify the implementation


include/bluetooth/conn_evt_prepare_cb.h
```diff
@@ -0,0 +1,66 @@
...
+ * it is sent on air.
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** Connection event prepare callback
+ *
+ * The callback is called from ISR context.
```
seko-nordic, line 35: would it be better to support thread mode instead so user can do anything in the callback, including BLE operations? As I see the primary goal is to call BLE API in timely manner. That also may be easier for us since thread mode by definition cannot be as accurate in terms of timing.

rugeGerritsen: @MarekPieta , do you have some input on this?

rugeGerritsen:
Other suggestion: Add the possibility to trigger a callback at the start of the events.

Then it might make sense to define the callback as a struct so that it is possible to later add `start()` and `end()` if requested.

```c
struct bt_conn_evt_cb {
	 void (*prepare)(struct bt_conn *conn, void *user_data);
	 void (*start)((struct bt_conn *conn, void *user_data);
	 void (*end)((struct bt_conn *conn, void *user_data);
}, 

int bt_conn_evt_cb_register(const struct bt_conn *conn, struct bt_conn_evt_cb cb,
			       void *user_data,
			       uint32_t prepare_distance_us);
```

include/bluetooth/conn_evt_prepare_cb.h
rugeGerritsen: Suggestion: Make this compile-time configurable through Kconfig where the default is to offload to thread mode

seko-nordic: may be we can start with thread mode only and add config later if needed? I guess applications that need accurate timing will use the conn event trigger directly anyway?

subsys/bluetooth/host_extensions/conn_evt_prepare_cb.c
```diff
@@ -0,0 +1,205 @@
...
+};
+
+static int setup_connection_event_trigger(struct notification_cfg *cfg)
+{
+	int err;
+	uint16_t conn_handle;
+
+	err = bt_hci_get_conn_handle(cfg->conn, &conn_handle);
+	if (err) {
+		printk("Failed obtaining conn_handle (err %d)\n", err);
```
Tschet1, line 113: Should we use the logger here?

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,37 @@
...
+
+The connection event prepare callback will be called for every connection event for both the peripheral and central device.
+If the peripheral application provides data to be sent in this callback, the peripheral device will wake up for that given connection event.
+
+.. figure:: images/ConnEvtCbPeripheralLatency.svg
+   :alt: Connection event prepare callback with peripheral latency
+
+   The peripheral wakes up if data is provided in a connection event prepare callback
+
+For this configuration, the application is required to configure the prepare callback to trigger at least 3 ms before the connection event starts.
```
Tschet1, line 35: Should we add a constant somewhere for these 3ms?
Like that, we could change it more easily, if needed (e.g. to account for different HWs)

rugeGerritsen: Yes, I believe we can add something like `BT_CONN_EVT_CB_PREPARE_DISTANCE_US_RECOMMENDED` or something similar to `evt_cb.h`

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
+.. _ug_bt_conn_evt_cb:
+
+Synchronize data sampling with Bluetooth connection events
+##########################################################
+
+The Bluetooth connection event callback feature is used to inform the application about connection events.
+The connection event prepare callback triggered right before of a Bluetooth connection event.
```
peknis, line 7:
```suggestion
The ``connection event prepare`` callback is triggered right before a Bluetooth connection event.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
+.. _ug_bt_conn_evt_cb:
+
+Synchronize data sampling with Bluetooth connection events
+##########################################################
```
peknis, line 4:
```suggestion
Connection event callback
#########################
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+The Bluetooth connection event callback feature is used to inform the application about connection events.
+The connection event prepare callback triggered right before of a Bluetooth connection event.
+This allows the application to trigger sensor data collection for transmission during the upcoming connection event.
+
+.. figure:: images/ConnEvtCb.svg
+   :alt: Connection event prepare callback
+
+   Connection event prepare callback is called before a connection event
+
+The timing of the prepare callback is configured with the parameter `prepare_time_us`.
```
peknis, line 15:
```suggestion
The timing of the prepare callback is configured with the parameter ``prepare_time_us``.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+The parameter must be set sufficiently large to allow the Bluetooth Stack to enqueue a data packet for the upcoming connection event.
```
peknis, line 16:
```suggestion
You must set the parameter value to large enough to allow the Bluetooth stack to enqueue a data packet for the upcoming connection event.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+
+The :ref:`ble_conn_evt_cb` sample application demonstrates how this feature is used to reduce the latency between data sampling and data transmission.
```
peknis, line 18:
```suggestion
The :ref:`ble_conn_evt_cb` sample demonstrates how you can use this feature to reduce the latency between data sampling and data transmission.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+
+This feature can only be used with the :ref:`SoftDevice Controller <nrfxlib:softdevice_controller>` and only when the Bluetooth Controller and application are located on the same core.
```
peknis, line 20:
```suggestion
You can use this feature only with the :ref:`SoftDevice Controller <nrfxlib:softdevice_controller>` and only when the Bluetooth controller and application are located on the same core.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+
+Connection event callback with peripheral latency
+*************************************************
+
+When a connection is configured with peripheral latency, the peripheral device may sleep for `peripheral_latency` consecutive connection events to save power.
```
peknis, line 25:
```suggestion
When a connection is configured with peripheral latency, the peripheral device may sleep for ``peripheral_latency`` consecutive connection events to save power.
```
How is the peripheral latency configured (using a Kconfig option or something else?)

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+It may wake up more often if, for example if it has data to send.
```
peknis, line 26:
```suggestion
The device may wake up more often if, for example, it has data to send.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+Therefore, the central device assumes the peripheral will wake up every connection event.
```
peknis, line 27:
```suggestion
The central device assumes that the peripheral will wake up at every connection event.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+
+The connection event prepare callback will be called for every connection event for both the peripheral and central device.
```
peknis, line 29:
```suggestion
The ``connection event prepare`` callback will be called for every connection event for both the peripheral and central device.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+If the peripheral application provides data to be sent in this callback, the peripheral device will wake up for that given connection event.
+
+.. figure:: images/ConnEvtCbPeripheralLatency.svg
+   :alt: Connection event prepare callback with peripheral latency
+
+   The peripheral wakes up if data is provided in a connection event prepare callback
+
+For this configuration, the application is required to configure the prepare callback to trigger at least 3 ms before the connection event starts.
```
peknis, line 37:
```suggestion
For this configuration, the application must configure the prepare callback to trigger at least three ms before the connection event starts.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+This allows the Bluetooth stack to start the high frequency clock and radio peripheral in case they are not already running.
```
peknis, line 38:
```suggestion
This allows the Bluetooth stack to start the high frequency clock and radio peripheral if they are not already running.
```

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+If the data is provided too close to the start of the connection event, the data will be sent in future connection events.
```
peknis, line 39:
```suggestion
If the data is provided too close to the start of the connection event, the data will be sent in upcoming events.
```

include/bluetooth/evt_cb.h
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+#ifndef BT_EVT_CB_H__
+#define BT_EVT_CB_H__
+
+/**
+ * @file
+ * @defgroup bt_evt_cb Event callback
+ * @{
+ * @brief APIs to setup event callbacks
```
peknis, line 14:
```suggestion
 * @brief APIs to set up event callbacks
```

include/bluetooth/evt_cb.h
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	void (*prepare)(struct bt_conn *conn,
+			void *user_data);
+};
+
+/** Register a connect event callback for a connection.
+ *
+ * @param[in] conn                The connection context.
+ * @param[in] cb                  The callback to be used.
+ * @param[in] user_data           User data which will be provided in the callback.
```
peknis, line 51:
```suggestion
 * @param[in] user_data           User data that will be provided in the callback.
```

include/bluetooth/evt_cb.h
```diff
@@ -0,0 +1,72 @@
...
+ * @param[in] prepare_distance_us The distance in time from the start of the
+ *                                callback to the start of the connection event.
+ *
+ *  @retval 0          The device callbak has been successfully registered.
```
peknis, line 55:
```suggestion
 *  @retval 0          The device callback has been successfully registered.
```

include/bluetooth/evt_cb.h
```diff
@@ -0,0 +1,72 @@
...
+ *  @retval -EALREADY  A callback has already been registered for this connection.
+ *  @retval -ENOMEM    There is not enough PPI channels available to use this feature.
```
peknis, line 57:
```suggestion
 *  @retval -ENOMEM    There are not enough PPI channels available to use this feature.
```

samples/bluetooth/conn_evt_cb/README.rst
```diff
@@ -0,0 +1,136 @@
+.. _ble_conn_evt_cb:
+
+Bluetooth: Connection event callback
+#####################################
+
+.. contents::
+   :local:
+   :depth: 2
+
+The Connection event callback sample demonstrates how to use the connection event prepare callback feature described in :ref:`ug_bt_conn_evt_cb`.
```
peknis, line 10:
```suggestion
This sample demonstrates how to use the :ref:`ug_bt_conn_evt_cb` feature.
```

samples/bluetooth/conn_evt_cb/README.rst
```diff
@@ -0,0 +1,136 @@
...
+It uses the :ref:`latency_readme` and the :ref:`latency_client_readme` to showcase how this feature can be used to minimize the time between data sampling and data transmission.
```
peknis, line 11:
```suggestion
It uses the :ref:`latency_readme` and the :ref:`latency_client_readme` to showcase how you can use this feature to minimize the time between data sampling and data transmission.
```

samples/bluetooth/conn_evt_cb/README.rst
```diff
@@ -0,0 +1,136 @@
...
+
+The sample supports the following development kits:
+
+.. table-from-sample-yaml::
+
+You can use any two of the development kits mentioned above and mix different development kits.
+
+Additionally, the sample requires a connection to a computer with a serial terminal for each of the development kits.
+
+Note that the feature involves triggering a (D)PPI task directly from the SoftDevice Controller link layer, and therefore it is expected that the application and SoftDevice Controller are running on the same core.
```
peknis, line 24:
```suggestion
.. note::
   The feature involves triggering a (D)PPI task directly from the SoftDevice Controller link layer.
   The application and SoftDevice Controller are supposed to be running on the same core.
```

samples/bluetooth/conn_evt_cb/README.rst
```diff
@@ -0,0 +1,136 @@
...
+Building and running
+********************
+.. |sample path| replace:: :file:`samples/bluetooth/llpm`
+
+.. include:: /includes/build_and_run.txt
+
+Testing
+=======
+
+After programming the sample to both development kits, test it by performing the following steps:
```
peknis, line 35:
```suggestion
After programming the sample to both development kits, perform the following steps to test it:
```

samples/bluetooth/conn_evt_cb/README.rst
```diff
@@ -0,0 +1,136 @@
...
+
+1. Connect to both kits with a terminal emulator (for example, `nRF Connect Serial Terminal`_).
+   See :ref:`test_and_optimize` for the required settings and steps.
+#. Reset both kits.
+#. In one of the terminal emulators, type "c" to start the application on the connected board in the central role.
+#. In the other terminal emulator, type "p" to start the application in the peripheral role.
```
peknis, line 41:
```suggestion
#. In one of the terminal emulators, type ``c`` to start the application on the connected board in the central role.
#. In the other terminal emulator, type ``p`` to start the application in the peripheral role.
```

samples/bluetooth/conn_evt_cb/README.rst
```diff
@@ -0,0 +1,136 @@
...
+#. Observe that latency measurements are printed in the terminals.
+
+Sample output
+=============
+
+The result should look similar to the following output.
+
+- For the central::
```
peknis, line 49:
```suggestion
For the central::
```

samples/bluetooth/conn_evt_cb/README.rst
```diff
@@ -0,0 +1,136 @@
...
+   Latency: 1116 us, round trip latency: 51116 us
+   Latency: 1116 us, round trip latency: 51116 us
+   Latency: 1116 us, round trip latency: 51116 us
+   Latency: 1116 us, round trip latency: 51116 us
+   Latency: 1116 us, round trip latency: 51116 us
+   Latency: 1116 us, round trip latency: 51116 us
+   Latency: 1116 us, round trip latency: 51116 us
+   Latency: 1116 us, round trip latency: 51116 us
+
+- For the peripheral::
```
peknis, line 78:
```suggestion
For the peripheral::
```

peknis (changes requested):

doc/nrf/protocols/bt/ble/conn_evt_cb.rst
rugeGerritsen: The peripheral latency is configured run-time. It may change over the life-time of a connection. The central sets the initial value. The central may at any point in time change the value. The peripheral may request the central to change it

samples/bluetooth/conn_evt_cb/src/central.c
```diff
@@ -0,0 +1,78 @@
...
+
+	struct bt_conn *unused_conn;
+
+	err = bt_conn_le_create(addr, BT_CONN_LE_CREATE_CONN,
+				BT_LE_CONN_PARAM_DEFAULT, &unused_conn);
+	if (err) {
+		printk("Create conn to %s failed (%d)\n", addr_str, err);
+		scan_start();
+	}
+}
```
alwa-nordic, line 65:
```suggestion

	bt_conn_unref(unused_conn);
}
```

subsys/bluetooth/host_extensions/conn_evt_cb.c
```diff
@@ -0,0 +1,207 @@
...
+static struct conn_evt_cb_cfg cfg[CONFIG_BT_MAX_CONN];
+static struct k_work_delayable work_queue_items[CONFIG_BT_MAX_CONN];
+
+static void work_handler(struct k_work *work)
+{
+	uint8_t index = ARRAY_INDEX(work_queue_items, work);
+
+	cfg[index].cb.prepare(cfg[index].conn, cfg[index].user_data);
+
+	/* Schedule the callback in the connection event does not occur because
```
alwa-nordic, line 55:
```suggestion
	/* Schedule the callback in case the connection event does not occur because
```

subsys/bluetooth/host_extensions/conn_evt_cb.c
```diff
@@ -0,0 +1,207 @@
...
+	 * of peripheral latency or scheduling conflicts.
+	 */
+	k_work_reschedule(&work_queue_items[index], K_USEC(cfg[index].prepare_time_us));
+}
+
+static void egu_event_handler(uint8_t event_idx, void *context)
```
alwa-nordic, line 61: "EGU event" is a bad name for this event. The name would be better if it relates to `SDC_HCI_VS_CONN_EVENT_TRIGGER_ROLE_CONN`.

subsys/bluetooth/host_extensions/conn_evt_cb.c
```diff
@@ -0,0 +1,207 @@
...
+{
+	ARG_UNUSED(context);
+
+	__ASSERT_NO_MSG(event_idx < ARRAY_SIZE(work_queue_items));
+	__ASSERT_NO_MSG(cfg[event_idx].conn_interval_us > cfg[event_idx].prepare_time_us);
+
+	const uint32_t timer_distance_us =
+		cfg[event_idx].conn_interval_us - cfg[event_idx].prepare_time_us;
```
alwa-nordic, line 69: Do we need to account for some offset from `SDC_HCI_VS_CONN_EVENT_TRIGGER_ROLE_CONN` to the PHY connection event?

subsys/bluetooth/host_extensions/conn_evt_cb.c
```diff
@@ -0,0 +1,207 @@
...
+static void work_handler(struct k_work *work)
+{
+	uint8_t index = ARRAY_INDEX(work_queue_items, work);
+
+	cfg[index].cb.prepare(cfg[index].conn, cfg[index].user_data);
+
+	/* Schedule the callback in the connection event does not occur because
+	 * of peripheral latency or scheduling conflicts.
+	 */
+	k_work_reschedule(&work_queue_items[index], K_USEC(cfg[index].prepare_time_us));
```
alwa-nordic, line 58:
```suggestion
	k_work_reschedule(&work_queue_items[index], K_USEC(cfg[index].conn_interval_us));
```

We should maybe also record the time of the `SDC_HCI_VS_CONN_EVENT_TRIGGER_ROLE_CONN` event and use that as anchor? I worry about scheduling delays.

subsys/bluetooth/host_extensions/conn_evt_cb.c
```diff
@@ -0,0 +1,207 @@
...
+		return err;
+	}
+
+	const sdc_hci_cmd_vs_set_conn_event_trigger_t params = {
+		.conn_handle = conn_handle,
+		.role = SDC_HCI_VS_CONN_EVENT_TRIGGER_ROLE_CONN,
+		.ppi_ch_id = cfg->ppi_channel,
+		.period_in_events = 1, /* Trigger every connection event. */
+		.conn_evt_counter_start =
+			get_next_conn_event_count_ret_params.next_conn_event_counter + 10,
```
alwa-nordic, line 137: Why the `+ 10`?

subsys/bluetooth/host_extensions/conn_evt_cb.c
```diff
@@ -0,0 +1,207 @@
...
+static nrfx_egu_t egu_inst = NRFX_EGU_INSTANCE(EGU_INST_IDX);
+
+struct conn_evt_cb_cfg {
+	struct bt_conn_evt_cb cb;
+	struct bt_conn *conn;
+	void *user_data;
+	uint32_t prepare_time_us;
+
+	uint8_t ppi_channel;
+	uint16_t conn_interval_us;
```
alwa-nordic, line 43: Can we just read it from `bt_conn_get_info()` instead of storing it?

alwa-nordic (dismissed):

subsys/bluetooth/host_extensions/conn_evt_cb.c
rugeGerritsen: We can. I have no strong preference of how to obtain it. Caching it seemed like the simpler option

subsys/bluetooth/host_extensions/conn_evt_cb.c
```diff
@@ -0,0 +1,207 @@
...
+};
+
+static struct conn_evt_cb_cfg cfg[CONFIG_BT_MAX_CONN];
+static struct k_work_delayable work_queue_items[CONFIG_BT_MAX_CONN];
+
+static void work_handler(struct k_work *work)
```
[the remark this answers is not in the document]
rugeGerritsen: I'm not sure if I understand what the concern is here. Can you elaborate?
The `on_disconnected` callback cancels any outstanding work for the given connection

subsys/bluetooth/host_extensions/conn_evt_cb.c
rugeGerritsen: We could store the time of the trigger, but that will require some additional HW resources. We have some plans on how to expose this information. That will allow us to simplify the implementation.

I would assume that the current implementation is good enough to start out with. It is anyways marked with "experimental".

subsys/bluetooth/host_extensions/conn_evt_cb.c
rugeGerritsen: It is added just to ensure we don't start it too late (which is not allowed). We plan to replace it with a new HCI command which doesn't have this logic

subsys/bluetooth/host_extensions/conn_evt_cb.c
rugeGerritsen: We can assume that the trigger occurs right before the radio activity. We can improve this, but we will not get too much of a benefit from it. Also related to https://github.com/nrfconnect/sdk-nrf/pull/16099/files/a92d9934100a4141874556a6386ff608a9c03a1e#r1672204188

include/bluetooth/evt_cb.h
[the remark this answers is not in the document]
rugeGerritsen: Based upon this. Maybe we should name this feature "Bluetooth radio event callback" with the corresponding `radio_evt_cb.h` and API named `bt_conn_radio_evt_cb_register`?

subsys/bluetooth/host_extensions/conn_evt_cb.c
alwa-nordic: Ok. I was hoping that if we just set it to zero it would trigger on the next event.

I note that depending on connection interval, 10 connection events is a noticeable delay for a e.g. a mouse to start reporting after reconnecting. This could be critical to make a 'deep sleep until accelerometer input' feel responsive to an end user.

subsys/bluetooth/host_extensions/conn_evt_cb.c
alwa-nordic: I was thinking of capturing `k_cycle_get_32` in the EGU ISR. But I agree it may not be necessary.

Obs: The suggestion above in this thread point out a bug, I think. Do you agree the "backup plan" reschedule should be a connection interval in the future?

rugeGerritsen: Yes, it should be a connection interval!

subsys/bluetooth/host_extensions/conn_evt_cb.c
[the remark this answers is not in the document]
alwa-nordic: I'm thinking that these functions can be called from separate threads, which requires attention. The cancel can fail and then we might call the NULL pointer, pass a dangling conn, or reschedule the work, etc..

I remember now since Zephyr 3.7 the disconnected callback is invoked from the system work queue, so these functions will indeed naturally synchronize. But this is not true before Zephyr 3.7, and I'm not sure it's sound design to rely on this anyway.

It may also be possible to find some reasoning about timings of connection radio events and disconnections that shows it's impossible for these functions to run at the same time, but that's very complex, and I actually doubt there is a bulletproof argument. It's much less work to assume the worst.

My opinion is that we should never rely on a simple cancel. The cancel is purely a optimization, and we must assume the work will run, possibly racing the function that calls cancel.

From an API standpoint, what guarantees do we give about when the last moment the callback can run is?

peknis (approved):

include/bluetooth/radio_notification_cb.h
```diff
@@ -0,0 +1,70 @@
...
+ * @param[in] cb                  The callback structure to be used.
+ *                                The memory pointed to needs to be kept alive by the user.
+ * @param[in] prepare_distance_us The distance in time from the start of the
+ *                                prepare callback to the start of the connection event.
+ *
+ *  @retval 0          The callback structure has been successfully registered.
+ *  @retval -EALREADY  A callback has already been registered.
+ *  @retval -ENOMEM    There are not enough PPI channels available to use this feature.
+ */
+int bt_radio_notification_conn_cb_register(struct bt_radio_notification_conn_cb *cb,
```
seko-nordic, line 59:
```suggestion
int bt_radio_notification_conn_cb_register(struct bt_radio_notification_conn_cb const* cb,
```
so app can put it to flash

samples/bluetooth/radio_notification_cb/README.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Additionally, the sample requires a connection to a computer with a serial terminal for each of the development kits.
+
+.. note::
+   The feature involves triggering a (D)PPI task directly from the SoftDevice Controller link layer.
+   The application and SoftDevice Controller are supposed to be running on the same core.
+
+Building and running
+********************
+.. |sample path| replace:: :file:`samples/bluetooth/llpm`
```
seko-nordic, line 30: llpm?

samples/bluetooth/radio_notification_cb/sample.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: Connection event callback
+tests:
+  sample.bluetooth.conn_evt_cb:
+    sysbuild: true
+    build_only: true
+    integration_platforms:
+      - nrf52dk/nrf52832
+      - nrf52840dk/nrf52840
+      - nrf5340dk/nrf5340/cpunet
+    platform_allow: nrf52dk/nrf52832 nrf52840dk/nrf52840 nrf5340dk/nrf5340/cpunet
```
seko-nordic, line 12: does not work on 54?

samples/bluetooth/radio_notification_cb/src/main.c
```diff
@@ -0,0 +1,226 @@
...
+	err = bt_latency_request(&latency_client,
+				 &latency_request_time,
+				 sizeof(latency_request_time));
+	if (err) {
+		printk("Latency request failed (err %d)\n", err);
+	}
+	latency_request_time_in_use = true;
+}
+
+static struct bt_radio_notification_conn_cb callbacks = {
```
seko-nordic, line 175:
```suggestion
static const struct bt_radio_notification_conn_cb callbacks = {
```

include/bluetooth/radio_notification_cb.h
KAGA164: I would change an order to `const struct`

samples/bluetooth/radio_notification_cb/src/main.c
```diff
@@ -0,0 +1,226 @@
...
+#include <zephyr/bluetooth/conn.h>
+#include <bluetooth/services/latency.h>
+#include <bluetooth/services/latency_client.h>
+#include <bluetooth/gatt_dm.h>
+#include <bluetooth/radio_notification_cb.h>
+
+static struct bt_latency latency;
+static struct bt_latency_client latency_client;
+
+void scan_start(void);
```
KAGA164, line 19: static?

samples/bluetooth/radio_notification_cb/src/main.c
```diff
@@ -0,0 +1,226 @@
...
+void adv_start(void);
```
KAGA164, line 20: static?

samples/bluetooth/radio_notification_cb/src/main.c
```diff
@@ -0,0 +1,226 @@
...
+{
+	printk("Service not found\n");
+}
+
+static void discovery_error(struct bt_conn *conn, int err, void *context)
+{
+	printk("Error while discovering GATT database: (err %d)\n", err);
+}
+
+struct bt_gatt_dm_cb discovery_cb = {
```
KAGA164, line 50: static const?

samples/bluetooth/radio_notification_cb/src/main.c
```diff
@@ -0,0 +1,226 @@
...
+		} else {
+			adv_start();
+		}
+		return;
+	}
+
+	printk("Connected: %s\n", addr);
+
+	if (bt_conn_get_info(conn, &info)) {
+		printk("Failed getting conn info\n");
```
KAGA164, line 76: Could you add error code to printk message?

subsys/bluetooth/host_extensions/radio_notification_conn_cb.c
```diff
@@ -0,0 +1,189 @@
...
+		return err;
+	}
+
+	const sdc_hci_cmd_vs_set_conn_event_trigger_t params = {
+		.conn_handle = conn_handle,
+		.role = SDC_HCI_VS_CONN_EVENT_TRIGGER_ROLE_CONN,
+		.ppi_ch_id = ppi_channels[conn_index],
+		.period_in_events = 1, /* Trigger every connection event. */
+		.conn_evt_counter_start =
+			get_next_conn_event_count_ret_params.next_conn_event_counter + 10,
```
KAGA164, line 115: +10 ? What does it mean or why is it needed?

subsys/bluetooth/host_extensions/radio_notification_conn_cb.c
```diff
@@ -0,0 +1,189 @@
...
+	struct bt_conn *conn = conn_pointers[index];
+
+	registered_cb->prepare(conn);
+
+	uint32_t conn_interval_us = conn_interval_us_get(conn);
+
+	/* Schedule the callback in the connection event does not occur because
+	 * of peripheral latency or scheduling conflicts.
+	 */
+	k_work_reschedule(&work_queue_items[index], K_USEC(conn_interval_us));
```
KAGA164, line 65: What does happen after disconnection? It looks like this work is never canceled

KAGA164 (changes requested):

samples/bluetooth/radio_notification_cb/src/main.c
rugeGerritsen: The function is defined in another file, so it should not be declared as static

subsys/bluetooth/host_extensions/radio_notification_conn_cb.c
rugeGerritsen: Fixed

subsys/bluetooth/host_extensions/radio_notification_conn_cb.c
rugeGerritsen: Added a comment in the code. We indent to remove this logic later

samples/bluetooth/radio_notification_cb/src/main.c
```diff
@@ -0,0 +1,226 @@
...
+
+	printk("Starting radio notification callback sample.\n");
+
+	err = bt_enable(NULL);
+	if (err) {
+		printk("Bluetooth init failed (err %d)\n", err);
+		return 0;
+	}
+
+	err = bt_radio_notification_conn_cb_register(&callbacks,
```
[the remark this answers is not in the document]
rugeGerritsen: Added a check

samples/bluetooth/radio_notification_cb/sample.yaml
rugeGerritsen: Not yet. Need to add EGU definitions here: https://github.com/zephyrproject-rtos/zephyr/blob/main/modules/hal_nordic/nrfx/Kconfig#L64. I believe we can do that separately to unblock other work

samples/bluetooth/radio_notification_cb/src/main.c
rugeGerritsen: See https://github.com/nrfconnect/sdk-nrf/pull/16099#discussion_r1675785374

samples/bluetooth/radio_notification_cb/src/central.c
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/bluetooth/bluetooth.h>
+#include <zephyr/bluetooth/conn.h>
+
+#define ADV_NAME_STR_MAX_LEN (sizeof(CONFIG_BT_DEVICE_NAME))
+
+void scan_start(void);
```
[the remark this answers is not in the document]
rugeGerritsen: See https://github.com/nrfconnect/sdk-nrf/pull/16099#discussion_r1675785374

subsys/bluetooth/host_extensions/Kconfig
```diff
@@ -0,0 +1,76 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+config BT_RADIO_NOTIFICATION_CONN_CB
```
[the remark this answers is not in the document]
rugeGerritsen: I guess you suggest this so that the other options are hidden when this is not enabled?

doc/nrf/protocols/bt/ble/radio_notification_conn_cb.rst
```diff
@@ -0,0 +1,39 @@
...
+You can use this feature only with the :ref:`SoftDevice Controller <nrfxlib:softdevice_controller>` and only when the Bluetooth controller and application are located on the same core.
+
+Radio Notification connection callback with peripheral latency
+**************************************************************
+
+When a connection is configured with peripheral latency, the peripheral device may sleep for ``peripheral_latency`` consecutive connection events to save power.
+The device may wake up more often if, for example, it has data to send.
+The central device assumes that the peripheral will wake up at every connection event.
+
+The :c:member:`ble_radio_notification_conn_cb.prepare` callback will be called before every connection event for both the peripheral and central device.
```
Tschet1, line 29: Should we combine this with the previous paragraph? I think they logically belong together.

include/bluetooth/radio_notification_cb.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include <stdint.h>
+#include <zephyr/bluetooth/conn.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** Recommended prepare distance for the connection event callback. */
+#define BT_RADIO_NOTIFICATION_CONN_CB_PREPARE_DISTANCE_US_RECOMMENDED 3000
```
Tschet1, line 31: I think this is not enough if the application configures the maximum connection interval of 4s and using 500ppm clock accuracy.

subsys/bluetooth/host_extensions/Kconfig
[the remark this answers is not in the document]
KAGA164: Yes, You are right

subsys/bluetooth/host_extensions/radio_notification_conn_cb.c
```diff
@@ -0,0 +1,200 @@
...
+	if (bt_conn_get_info(conn, &info)) {
+		return;
+	}
+
+	if (info.state == BT_CONN_STATE_CONNECTED) {
+		registered_cb->prepare(conn);
+
+		uint32_t conn_interval_us = conn_interval_us_get(conn);
+
+		/* Schedule the callback in the connection event does not occur because
```
Tschet1, line 68:
```suggestion
		/* Schedule the callback in case the connection event does not occur because
```

Tschet1 (approved):

include/bluetooth/radio_notification_cb.h
rugeGerritsen: Maybe we should add some documentation to state that this value needs to be chosen according to the maximum connection interval and clock drift? We can calculate the maximum clock drift for some typically use cases

Tschet1: not sure on what is best. We could consider
1. documenting until what connection interval the constant can be used
2. have different constants for different connection intervals (one for "short" and one for "long")
3. document on how the value can be chosen (e.g. provide some kind of formula)
4. have a runtime check based on the actual connection interval

KAGA164 (approved):

samples/bluetooth/radio_notification_cb/sample.yaml
rugeGerritsen: Will be possible after https://github.com/zephyrproject-rtos/zephyr/pull/75866 is in NCS

PR #16102, opened 2024-06-20
lib: nrf_modem: correct help text for NRF_MODEM_LIB_ON_FAULT_LTE_NET_IF

eivindj-nordic:
Correct help text for CONFIG_NRF_MODEM_LIB_ON_FAULT_LTE_NET_IF Kconfig option.

### DIFF
--- a/lib/nrf_modem_lib/Kconfig.modemlib
+++ b/lib/nrf_modem_lib/Kconfig.modemlib
@@ -220,10 +220,11 @@ config NRF_MODEM_LIB_ON_FAULT_APPLICATION_SPECIFIC
 	  Let the application define the fault handler function.
 
 config NRF_MODEM_LIB_ON_FAULT_LTE_NET_IF
-	bool "Application defined"
+	bool "Raise a fatal connectivity error network event"
 	depends on NRF_MODEM_LIB_NET_IF
 	help
-	  Let the application define the fault handler function.
+	  Pass modem faults to the network interface driver.
+	  The network interface driver will throw a NET_EVENT_CONN_IF_FATAL_ERROR event.
 
 endchoice # NRF_MODEM_LIB_ON_FAULT
 


balaji-srin (approved):
Thanks for the fix.

lib/nrf_modem_lib/Kconfig.modemlib
```diff
@@ -220,10 +220,10 @@ config NRF_MODEM_LIB_ON_FAULT_APPLICATION_SPECIFIC
 	  Let the application define the fault handler function.
 
 config NRF_MODEM_LIB_ON_FAULT_LTE_NET_IF
-	bool "Application defined"
+	bool "Pass to network interface driver"
 	depends on NRF_MODEM_LIB_NET_IF
 	help
-	  Let the application define the fault handler function.
+	  Pass modem faults to the network interface driver.
```
lemrey, line 226: We should mention it's a `NET_EVENT_CONN_IF_FATAL_ERROR` error in the help text, so the user knows which event to handle.

lib/nrf_modem_lib/Kconfig.modemlib
```diff
@@ -220,10 +220,10 @@ config NRF_MODEM_LIB_ON_FAULT_APPLICATION_SPECIFIC
 	  Let the application define the fault handler function.
 
 config NRF_MODEM_LIB_ON_FAULT_LTE_NET_IF
-	bool "Application defined"
+	bool "Pass to network interface driver"
```
lemrey, line 223: Well, in the end the network interface driver will emit a `NET_EVENT_CONN_IF_FATAL_ERROR` error, which must be handled by the application. That's not what "pass to network interface driver" suggests, as in, the network interface driver is handling the fault.

What about "Raise a fatal connectivity error network event" ? Bit long, but it reflects the description of the event

image

and does not suggest that the event is handled by the network interface driver (it isn't).

lemrey (approved):

PR #16106, opened 2024-06-20
applications: nrf5340_audio: Update buildprog.py for FEM

rick1082:
Update the naming in buildprog.py for FEM.
OCT-NONE


alexsven (approved):

koffes (approved):
The change itself is good, but I do think removing it all together is a preferable option.

PR #16108, opened 2024-06-20

anangl:
Pull in a few more commits from upstream that are needed for 2.7.0.


anangl:
No longer needed. Zephyr revision will be updated further by #16125.

PR #16119, opened 2024-06-21
manifest: Update zephyr with fix in uart driver

nordic-krch:
Fixes NCSDK-28005.


shanthanordic:
Cleared milestone too late for 2.7.0

PR #16124, opened 2024-06-21
cmake: sysbuild: Fix secure boot public key relative path usage

nordicjm:
Fixes an issue whereby the public key files for secure boot were not looked up using the application's configuration directory as the base

### DIFF
--- a/cmake/sysbuild/debug_keys.cmake
+++ b/cmake/sysbuild/debug_keys.cmake
@@ -63,13 +63,11 @@ if(NOT SB_CONFIG_SECURE_BOOT_SIGNING_CUSTOM AND "${SB_CONFIG_SECURE_BOOT_SIGNING
     )
   set(SIGN_KEY_FILE_DEPENDS debug_sign_key_target)
 else()
-  # Resolve path relative to the main application's configuration directory.
-  sysbuild_get(app_config_dir IMAGE ${DEFAULT_IMAGE} VAR APPLICATION_CONFIG_DIR CACHE)
-
   if(IS_ABSOLUTE ${SB_CONFIG_SECURE_BOOT_SIGNING_KEY_FILE})
     set(SIGNATURE_PRIVATE_KEY_FILE ${SB_CONFIG_SECURE_BOOT_SIGNING_KEY_FILE})
   else()
-    set(SIGNATURE_PRIVATE_KEY_FILE ${app_config_dir}/${SB_CONFIG_SECURE_BOOT_SIGNING_KEY_FILE})
+    # Resolve path relative to the application configuration directory.
+    set(SIGNATURE_PRIVATE_KEY_FILE ${APPLICATION_CONFIG_DIR}/${SB_CONFIG_SECURE_BOOT_SIGNING_KEY_FILE})
   endif()
 
   if(NOT EXISTS ${SIGNATURE_PRIVATE_KEY_FILE})
@@ -133,11 +131,11 @@ if("${SB_CONFIG_SECURE_BOOT_PUBLIC_KEY_FILES}" STREQUAL "debug")
   set(PROVISION_KEY_DEPENDS provision_key_target)
 else()
   foreach(key ${PUBLIC_KEY_FILES_LIST})
-    # Resolve path.
     if(IS_ABSOLUTE ${key})
       list(APPEND PUBLIC_KEY_FILES ${key})
     else()
-      list(APPEND PUBLIC_KEY_FILES ${CMAKE_SOURCE_DIR}/${key})
+      # Resolve path relative to the application configuration directory.
+      list(APPEND PUBLIC_KEY_FILES ${APPLICATION_CONFIG_DIR}/${key})
     endif()
   endforeach()
 endif()


cmake/sysbuild/debug_keys.cmake
```diff
@@ -132,12 +132,13 @@ if("${SB_CONFIG_SECURE_BOOT_PUBLIC_KEY_FILES}" STREQUAL "debug")
...
   set(PROVISION_KEY_DEPENDS provision_key_target)
 else()
+  sysbuild_get(app_config_dir IMAGE ${DEFAULT_IMAGE} VAR APPLICATION_CONFIG_DIR CACHE)
   foreach(key ${PUBLIC_KEY_FILES_LIST})
     # Resolve path.
     if(IS_ABSOLUTE ${key})
       list(APPEND PUBLIC_KEY_FILES ${key})
     else()
-      list(APPEND PUBLIC_KEY_FILES ${CMAKE_SOURCE_DIR}/${key})
+      list(APPEND PUBLIC_KEY_FILES ${app_config_dir}/${key})
```
tejlmand, line 141: I'm not convinced this is correct for the following reasons.

The code is looping `PUBLIC_KEY_FILES_LIST`: https://github.com/nrfconnect/sdk-nrf/blob/a6e9c1bd9b408bb929790c282d88d1654e787a47/cmake/sysbuild/debug_keys.cmake#L136

a list which is defined by the sysbuild variable: `SB_CONFIG_SECURE_BOOT_PUBLIC_KEY_FILES` here:
https://github.com/nrfconnect/sdk-nrf/blob/a6e9c1bd9b408bb929790c282d88d1654e787a47/cmake/sysbuild/debug_keys.cmake#L7

As sysbuild itself is the entry point, and not the image, then having files relative depending on the image's config dir can be challenging, especially as the image's own `APPLICATION_CONFIG_DIR` will vary depending on the presence of `<main_app>/sysbuild/<image>`.

For example it means that first key in list could be looked up from `<main-app>/key1.pem`, 2nd key from `<main-app>/sysbuild/<image>/key2.pem`, and 3rd from `<workspace>/path/to/<image>/key3.pem`.
Also the list would just look as: `key1.pem,key2.pem,key3.pem`.

Current solution actually means keys as above can be written as (note, APPLICATION_CONFIG_DIR default for sysbuild will be the main app dir): `key1.pem,sysbuild/<image>/key2.pem;<abspath>/key3.pem` 
which also hints the image for which the key is intended as part of the path.


Note, the help text for `SB_CONFIG_SECURE_BOOT_PUBLIC_KEY_FILES` actually states abs paths must be used, so regardless of implementation, the help text must be updated with a precise description of relative path behavior.
https://github.com/nrfconnect/sdk-nrf/blob/a6e9c1bd9b408bb929790c282d88d1654e787a47/sysbuild/Kconfig.secureboot#L99

An alternative can be to first lookup the key relative to sysbuild's APPLICATION_CONFIG_DIR (current behavior), and if not found there, then look relative to the application's APPLICATION_CONFIG_DIR (proposed change).

tejlmand (changes requested):

cmake/sysbuild/debug_keys.cmake
nordicjm: > As sysbuild itself is the entry point, and not the image, then having files relative depending on the image's config dir can be challenging, especially as the image's own APPLICATION_CONFIG_DIR will vary depending on the presence of <main_app>/sysbuild/<image>.

This is for the default image **only** so will never use those folders, so the keys will be in `<main_app>` by default or the alternative configuration directory if one is set, e.g. for nrf_desktop it could be `applications/nrf_desktop/configuration/nrf52840dk_nrf52840`, it will not be of another image. The keys should be relative to the application configuration directory, not sysbuild itself, since everything in sysbuild configuration in NCS is relative to the application config dir. Will update help text as part of another PR as there's a few Kconfig entries that need updating

tejlmand: > This is for the default image only so will never use those folders,

right, for some reason I read the code as picking up files per image application config dir when I first looked at it.

> The keys should be relative to the application configuration directory, not sysbuild itself,

The default value for the sysbuild application config dir is the same as the main app's config dir:
https://github.com/nrfconnect/sdk-zephyr/blob/297b00dade9f72873f9dd9147008e4a90516a3ff/share/sysbuild/cmake/modules/sysbuild_kconfig.cmake#L34-L36
```
elseif(NOT DEFINED APPLICATION_CONFIG_DIR)
  get_filename_component(APP_DIR  ${APP_DIR} ABSOLUTE)
  set(APPLICATION_CONFIG_DIR ${APP_DIR})
```

But we need to ensure consistency on how relative paths are treated when configured through sysbuild, cause we really don't want to end in a situation where relative in sysbuild's kconfig means something different for each setting. We have had such issues in the past in Zephyr.

Current situation afaict is actually that the both current and proposed change will end in same location, but the keys are still configured through sysbuild kconfig and used in sysbuild code.

Do we have an actual issue with current implementation, or is it simply an observation, and you wanted to avoid issues in future ?

> e.g. for nrf_desktop it could be applications/nrf_desktop/configuration/nrf52840dk_nrf52840

and actually nRF Desktop sets the APPLICATION_CONFIG_DIR in sysbuild which is then passed on to the main app (and other images).
https://github.com/nrfconnect/sdk-nrf/blob/0862baa53463f8bf51418752315c535349920add/applications/nrf_desktop/sysbuild/CMakeLists.txt#L8

 > Will update help text as part of another PR as there's a few Kconfig entries that need updating

:+1:

nordicjm: > Do we have an actual issue with current implementation, or is it simply an observation, and you wanted to avoid issues in future ?

Without this, this is an issue as the path supplied must be absolute, not relative as the help text implies.

Help text update PR: https://github.com/nrfconnect/sdk-nrf/pull/16152

tejlmand: > not relative as the help text implies.

current help text doesn't imply anything wrt. relative paths, as it explicitly states that path must be absolute.

> > But we need to ensure consistency on how relative paths are treated when configured through sysbuild, cause we really don't want to end in a situation where relative in sysbuild's kconfig means something different for each setting. We have had such issues in the past in Zephyr.

this part was never commented.

Sysbuild follows the value of APPLICATION_CONFIG_DIR when looking up `sysbuild.conf`, APPLICATION_CONFIG_DIR is also the value used by main app.
And if not explicitly given by user or the main app, then the value of `APPLICATION_CONFIG_DIR` is identical to the CMake source dir of the main app.

However, in case a user decides to use different app directories for sysbuild and the main app, by specifying `SB_APPLICATION_CONFIG_DIR` for sysbuild and `APPLICATION_CONFIG_DIR` (or `<app>_APPLICATION_CONFIG_DIR`) for for the main app, then it would be confusing if a SB_CONFIG setting for a relative path inside the `<SB_APPLICATION_CONFIG_DIR>/sysbuild.conf` file is suddenly referring outside of `SB_APPLICATION_CONFIG_DIR` and inside the `<app>_APPLICATION_CONFIG_DIR` directory.
The value is defined for sysbuild and used by sysbuild should be relative to sysbuild's APPLICATION_CONFIG_DIR.

In practice I don't see a reason to have `SB_APPLICATION_CONFIG_DIR` pointing differently than `APPLICATION_CONFIG_DIR` so the use of using the APPLICATION_CONFIG_DIR inside sysbuild will not change the intended behavior with this PR.

But as long as it is indeed possible to specify a dedicated `SB_APPLICATION_CONFIG_DIR` then the most correct thing is to treat relative path set in sysbuild.conf as relative to `SB_APPLICATION_CONFIG_DIR`.

nordicjm: I've given this a try over this morning and from what I can see the `SB_APPLICATION_CONFIG_DIR` implementation is completely broken. When you set it to a directory that doesn't exist, no error is printed, in fact, the project configures normally like you have specified a valid sysbuild.conf file - this is wrong, it should throw an error. And using your method of having a sysbuild.conf file in the application directory and not setting `SB_APPLICATION_CONFIG_DIR` to anything leads it to be nothing, so using `SB_APPLICATION_CONFIG_DIR` for this PR in `cmake/sysbuild/debug_keys.cmake` for the relative path gives:
```
-- Build files have been written to: /tmp/bb/zephyr/samples/hello_world/_AA/s1_image
CMake Error at /tmp/bb/nrf/cmake/sysbuild/debug_keys.cmake:77 (message):
  Config points to non-existing PEM file '/app.pem'
```
It only works when you do specify `SB_APPLICATION_CONFIG_DIR` on the command line and specify it to a valid folder.

tejlmand: > so using `SB_APPLICATION_CONFIG_DIR` for this PR in cmake/sysbuild/debug_keys.cmake for the relative path gives:

that's not what I wrote, I wrote:
> The value is defined for sysbuild and used by sysbuild should be relative to sysbuild's APPLICATION_CONFIG_DIR.

and sysbuild will take care of updating the value of `APPLICATION_CONFIG_DIR` to the value of `SB_APPLICATION_CONFIG_DIR` when that one is set.

Just like an any CMake code for an application uses `APPLICATION_CONFIG_DIR`, not `<image>_APPLICATION_CONFIG_DIR`. But the value of the application's `APPLICATION_CONFIG_DIR` can be controlled using `<image>_APPLICATION_CONFIG_DIR`.

tejlmand: but hindsight I see that the additional description of how SB_APPLICATION_CONFIG_DIR can be used to control the value of `APPLICATION_CONFIG_DIR` can lead to confusion on what I tried to describe.

tejlmand (approved):

anangl (approved):

PR #16129, opened 2024-06-21
samples: suit: Disable MRAM auto poweroff feature

tomchy:
Disable auto poweroff feature in SMP transfer sample that uses bluetooth.

Ref: NCSDK-27870


kapi-no:
There is a discussion to introduce this change at the board level.

CC: @shanthanordic, @ns-tolu 

If we use a fix-per-project approach (suggested in this PR), we need to also align our nRF Desktop app and perhaps other projects that support SUIT SMP DFU over Bluetooth.

samples/suit/recovery/prj.conf
```diff
@@ -114,3 +114,7 @@ CONFIG_BT_CTLR_PHY_2M=n
 
 CONFIG_SSF_ECHO_SERVICE_ENABLED=n
 CONFIG_SSF_RESET_EVT_SERVICE_ENABLED=n
+
+# Enable MRAM latency NRFS service
+CONFIG_NRFS=y
+CONFIG_NRFS_MRAM_SERVICE_ENABLED=y
```
ahasztag, line 120: Does recovery still fit after that? I assume yes, but I'd like to be sure.

PR #16138, opened 2024-06-21
samples: net: add net tag for samples

nordic-piks:
Extend list of paths used to check if net tests/samples should be executed.

### DIFF
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -65,18 +65,54 @@ bluetooth:
 net:
   files:
     - nrf/samples/net/
+    - nrf/tests/subsys/net/
+    - zephyr/subsys/mgmt/
     - zephyr/subsys/net/
+    - zephyr/subsys/random/
+    - zephyr/subsys/settings/
+    - zephyr/subsys/shell/
+    - zephyr/subsys/storage/
     - zephyr/include/zephyr/net/
     - zephyr/drivers/wifi/
     - zephyr/drivers/net/
     - zephyr/drivers/ethernet/
     - zephyr/drivers/ieee802154/
     - zephyr/drivers/ptp_clock/
-#    Examples:
-#    - nrf/subsys/net/
-#    - nrf/include/net/
-#    - nrf/drivers/wifi/
-#    - nrf/drivers/net/
+    - nrf/include/net/
+    - nrf/include/modem/
+    - nrf/drivers/wifi/
+    - nrf/drivers/net/
+    - nrf/drivers/hw_cc3xx/
+    - nrf/lib/at_cmd_parser/
+    - nrf/lib/at_host/
+    - nrf/lib/at_monitor/
+    - nrf/lib/dk_buttons_and_leds/
+    - nrf/lib/fprotect/
+    - nrf/lib/hw_id/
+    - nrf/lib/lte_link_control/
+    - nrf/lib/modem_antenna/
+    - nrf/lib/modem_info/
+    - nrf/lib/modem_key_mgmt/
+    - nrf/lib/nrf_modem_lib/
+    - nrf/lib/pdn/
+    - nrf/modules/cjson/
+    - nrf/modules/hostap/
+    - nrf/modules/trusted-firmware-m/
+    - nrf/subsys/dfu/
+    - nrf/subsys/fw_info/
+    - nrf/subsys/net/
+    - nrf/subsys/nrf_security/
+    - nrf/subsys/partition_manager/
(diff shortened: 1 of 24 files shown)


rlubos (approved):

nordic-piks:
In this case, there is defined tag (e.g. net) and files/folder with source code related to that area.

miha-nordic (approved):

nordic-piks:
DNM until NCS 2.7

scripts/ci/tags.yaml
```diff
@@ -62,11 +62,10 @@ net:
...
     - zephyr/drivers/ptp_clock/
-#    Examples:
-#    - nrf/subsys/net/
-#    - nrf/include/net/
-#    - nrf/drivers/wifi/
-#    - nrf/drivers/net/
+    - nrf/subsys/net/
+    - nrf/include/net/
+    - nrf/drivers/wifi/
+    - nrf/drivers/net/
```
trantanen, line 68: If you want to make this work, you need similar dependencies as in cellular case because many of the net samples you cellular connectivity only or as one of it's configuration.

lemrey: That's right, we should probably have at least the following here:
`nrf/include/modem`
`nrfxlib/nrf_modem/include`

trantanen: Because this tagging means that we wouldn't compile these sample configurations that have been tagged if the folders/files listed in tags.yaml for `net` are not modified, we would need to make this list a lot more complete. It's very close to `cellular` tag listing in PR #16140. You need to have `nrf/lib/lte_link_control` here and it explodes then easily.

tejlmand (approved):

nordic-piks:
Update:
Added also tag in tests.
Extended list of decencies base on actual compilation scope:
compiled all tests and samples, analyzed compile_commnad.json and generated list of decencies - that should limit the risk of missing dependency.

scripts/ci/tags.yaml
nordic-piks: I think I addressed this comment: I generated list of decencies for all net samples and tests and extended the list.

trantanen: These generated lists works nicely although are not updated when changes occur. I guess the change process will be to react when something is getting broken due to changed dependencies.

nordic-piks: Indeed, dependencies needs to be update. The main trigger for that may be that main will be broken after some PR.
Once I will finish this activity I will remind about that, thanks.

trantanen: This is something people will forget so I guess there should be some task once or twice a year to check some of the most important dependency lists (at least cellular and net from my perspective) and update this file.

trantanen: You could probably even automate the dependency list generation per sample if parsing of the output (compile_command.json) is doable. This way all samples could be compiled only when relevant changes had occurred.

nordic-piks: Yes and no :)
Parsing compile_command.json to evaluate dependencies for each sample was an option from the beginning.
That suppose to be an ultimate, precise solution.
However, as I see it now - after adding at lot of tags and checking a lot of compile_command outputs, this will not work automatically. There is still human needed to decide is some dependency is important and should actually trigger this specific test. E.g. all samples depends on zephyr code but we do not want to run them e.g. on some kernel changes. zephyr kernel tests should verify them, only.
So compile_command itself does not handle test isolation. 
We could end up in situation that we use compile_command exact dependencies but tweak them with some exceptions create manually. 
So with tagging I did shortcut of this - created dependencies and manually adjusted them.
Beside that, tagging was simpler to implement and configure.

Manual work - adjustments - are needed in both solutions.

trantanen: > There is still human needed to decide is some dependency is important and should actually trigger this specific test. E.g. all samples depends on zephyr code but we do not want to run them e.g. on some kernel changes. zephyr kernel tests should verify them, only. So compile_command itself does not handle test isolation. We could end up in situation that we use compile_command exact dependencies but tweak them with some exceptions create manually.

Yep, this makes sense although I'm not sure if it's a bad thing if we run all tests for fundamental Zephyr kernel changes.

> So with tagging I did shortcut of this - created dependencies and manually adjusted them. Beside that, tagging was simpler to implement and configure.

While tagging is simpler, it's visible in every sample.yaml file for each configuration. Automatic creation of dependencies and adjusting them manually would be a little less visible in the repository (if I've understood it correctly) and hence a little cleaner to me and probably more flexible to changes.

But let's see how this evolves in the future.

PR #16140, opened 2024-06-21
samples: tests: cellular: add cellular tag for samples and tests

nordic-piks:
Extend list of paths used to check if cellular tests/samples should be executed.


juhaylinen (approved):

scripts/ci/tags.yaml
```diff
@@ -97,7 +97,11 @@ posix:
...
 
-# cbprintf:
-#    files:
-#        - zephyr/lib/os/cbprintf*
-#        - zephyr/lib/posix
+cellular:
+  files:
+    - nrf/lib/at*
+    - nrf/lib/*modem*
+    - nrf/lib/sms
```
trantanen, line 104: There are many more libraries with cellular dependency or that are used only or mainly with our cellular nRF91 chips. So is this list now based on some well-thought criteria?

trantanen (commented):
Not knowing how `cellular` tag is intended to be used, I don't know if these are valid but are `asset_tracker_v2` and `serial_lte_modem` applications left without the tag intentionally? And same question for `samples/net/*`?

scripts/ci/tags.yaml
nordic-piks: Great!
Can you specify them?
ie all relevant folders from NCS with source code related to cellular?

trantanen:
> The idea behind this change is to allow automatic detection of PRs not related to given code area. In this case, there is defined tag (e.g. cellular) and files/folder with source code related to that area. This list of files/folders is then compared to files changed in PR. If particular PR does not change them, tests related to specified tag should not be executed - will be excluded.

Ok, the dependencies are probably a lot harder to create and maintain than it may initially look. E.g., if sample_X uses lib_A, which uses lib_B, we would need to list lib_A and lib_B in the dependencies. This way the dependencies will spread a lot.

@shanthanordic, are you aware of this exercise and how do we coordinate this from the dependencies point of view? I'm not sure it should be just me listing my best guess but this should be well organized.

> Regarding `asset_tracker_v2` - is this related to cellular area (in general)? (I simply do not know that).

It's using cellular connection to transmit data and the whole application wouldn't exist if there was no cellular connection. So, I'd say it's very much related to cellular area.

> `samples/net` are handled here - #16138

Many of the net samples should have similar dependencies as cellular if they are using cellular connections. And vice versa, i.e., cellular samples use `nrf/subsys/net` libraries a lot.

nordic-piks:
> > The idea behind this change is to allow automatic detection of PRs not related to given code area. In this case, there is defined tag (e.g. cellular) and files/folder with source code related to that area. This list of files/folders is then compared to files changed in PR. If particular PR does not change them, tests related to specified tag should not be executed - will be excluded.
> 
> Ok, the dependencies are probably a lot harder to create and maintain than it may initially look. E.g., if sample_X uses lib_A, which uses lib_B, we would need to list lib_A and lib_B in the dependencies. This way the dependencies will spread a lot.
> 
> @shanthanordic, are you aware of this exercise and how do we coordinate this from the dependencies point of view? I'm not sure it should be just me listing my best guess but this should be well organized.
> 
> > Regarding `asset_tracker_v2` - is this related to cellular area (in general)? (I simply do not know that).
> 
> It's using cellular connection to transmit data and the whole application wouldn't exist if there was no cellular connection. So, I'd say it's very much related to cellular area.
> 
> > `samples/net` are handled here - #16138
> 
> Many of the net samples should have similar dependencies as cellular if they are using cellular connections. And vice versa, i.e., cellular samples use `nrf/subsys/net` libraries a lot.

In general, this change (and others) aim to reduce test scope at CI (PRs) - the goal from task force exercise .
There was already finished exercise with static approach - integration quarantine (https://nordicsemi.atlassian.net/wiki/spaces/NCS/pages/530843071/Compilation+scope+table+in+NCS+integration+Twister+scope), however it did not bring expected speedups.
Now I do dynamic test selection base on modified files. There is used test_plan mechanisms from zephyr (more info @PerMac )
I would like to create groups of tests and its related code - scripts/ci/tags.yaml

If `cellular` area is big, that is fine, the list of folder will be longer, not a problem.
Maybe you can suggest different way of grouping tests and samples and linking them with code?

If there needs to be more work done to gather dependencies, I can help to do that (we can look e.g. on compile_commands.json results for all cellular samples).

Please now that I do not want to end in situation that e.g. cellular depends on everything, whole zephyr etc.
That may be true, but we do not want to verify e.g. kernel behavior with cellular tests.
Please think about `cellular` area only, with assumption that e.g. platform is already prepared and tested.

nordic-piks:
Fortunately, there is already list of dependent folders defined:
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L419

trantanen:
> Fortunately, there is already list of dependent folders defined: https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L419

This is for modem shell sample. While it's some sort of a starting point, it's not complete for all cellular samples.

tony-le-24 (approved):

tomi-font (approved):

samples/cellular/nrf_cloud_multi_service/sample.yaml
```diff
@@ -94,4 +94,4 @@ tests:
     extra_args: nrf_cloud_multi_service_SHIELD=nrf7002ek
       EXTRA_CONF_FILE="overlay_nrf7002ek_wifi_no_lte.conf"
       SB_CONF_FILE="sysbuild_nrf700x-wifi-conn.conf"
-    tags: ci_build sysbuild
+    tags: ci_build sysbuild cellular
```
glarsennordic, line 97: This one is a Wi-Fi-only build, so you can skip it for the cellular tag

trantanen: This should probably then have `net` tag.

nordic-piks: Hm, that would be tricky as it is in `samples/cellular` folder. I want to keep this simple and not do cross-functional.
So in this case I prefer to remove `cellular` tag from this configuration.

nordic-piks:
Updated after evaluation of:
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L419C2-L462
https://github.com/nrfconnect/sdk-nrfxlib/blob/main/.github/test-spec.yml#L142-L146
https://github.com/nrfconnect/sdk-zephyr/blob/main/.github/test-spec.yml#L211-L220
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L27C2-L40
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L42-L56
https://github.com/nrfconnect/sdk-nrfxlib/blob/main/.github/test-spec.yml#L2-L7
https://github.com/nrfconnect/sdk-zephyr/blob/main/.github/test-spec.yml#L17-L24
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L72-L101
https://github.com/nrfconnect/sdk-nrfxlib/blob/main/.github/test-spec.yml#L9-L13
https://github.com/nrfconnect/sdk-zephyr/blob/main/.github/test-spec.yml#L26-L30
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L103-L120
https://github.com/nrfconnect/sdk-nrfxlib/blob/main/.github/test-spec.yml#L21-L25
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L464-L495
https://github.com/nrfconnect/sdk-nrfxlib/blob/main/.github/test-spec.yml#L148-L152
https://github.com/nrfconnect/sdk-zephyr/blob/main/.github/test-spec.yml#L222-L228

nordic-piks:
DNM until NCS 2.7

gregersrygg (approved):
As mentioned by @trantanen, asset_tracker_v2 should probably be included.

nordic-piks:
> As mentioned by @trantanen, asset_tracker_v2 should probably be included.

Thanks, just added.

plskeggs (approved):

trantanen:
> IMO this dynamic scoping based on tags can only be helpful if we tag tests/apps that has limited dependencies, hence the assumption: "we haven't touch this region, hence we don't need to verify building this app" will work.

So is `cellular` good for tagging if we need to list at least ~40 folders and dozens of header files? Same question applies for `net` tag.

trantanen:
> > IMO this dynamic scoping based on tags can only be helpful if we tag tests/apps that has limited dependencies, hence the assumption: "we haven't touch this region, hence we don't need to verify building this app" will work.
> 
> So is `cellular` good for tagging if we need to list at least ~40 folders and dozens of header files? Same question applies for `net` tag.

And it's not only about the number of folders we need to list. To make it complete or good enough is not easy. How much failures we are allowed to accept because of the incompleteness of the information related to tagging? Is it fine that every now and then PR slips in and full twister run fails in nightly (or wherever it's run) as all configurations are not working. In testing side we are taking certain risk as dependencies for different CI plans are limited to increase performance but here we are talking about leaking compilation errors. Is that fine? @PerMac

> In general, this change (and others) aim to reduce test scope at CI (PRs) - the goal from task force exercise.

I understand the intent of this exercise but I'm not sure this is the right approach given the number of dependencies there are for `cellular` and `net` tags.
I could ask why there is a tag and why not just do it based on cellular and net folders? I know tagging is more flexible but you say elsewhere that a simple solution is preferred and not set `net` tag into a configuration in `cellular` folder.
But that wouldn't change the fundamental idea of this PR which I'm not convinced of.

> There was already finished exercise with static approach - integration quarantine
> (https://nordicsemi.atlassian.net/wiki/spaces/NCS/pages/530843071/Compilation+scope+table+in+NCS+integration+Twister+scope), however it did not bring expected speedups.

There are few reasons why this didn't work IMO. First of all it's a manual process and not everybody participates into those. You would need to poll those people who haven't done it several times and for a longer period of time.
Secondly, people (including myself) doesn't have any clue how much we should really strip out from integration runs. What's the target and how much risk we can tolerate? I'm sure we could do a lot more with this approach if we wanted as there are so many similar sample configurations and duplicates for different chips that are the same from the sample perspective.
That being said, I think this method is also very cumbersome from maintenance point of view. We add a new chip and it's missed from the quarantine perspective 99,99% of the time. Same for any added configuration or sample. Nobody remembers this kind of configuration that's outside the sample code folder. It would maybe be easier if that was defined in the sample.yaml rather than the outside-of-sample-codebase huge quarantine file. I've heard many people thinking (including myself) that "integration_platforms" would have been config for PR runs but that's not the case.

> Maybe you can suggest different way of grouping tests and samples and linking them with code?

The only easy optimization would be to only compile specific sample if only its codebase is modified. This doesn't help for those modifying libraries but this would reduce some load from CI and make twister stage really quick for sample/app modifications.
My question would be that do we absolutely need this kind of optimization for libraries? And if yes, has everything being done in the infrastructure side to speed things up.

tejlmand (approved):

gregersrygg:
> @gregersrygg Is that what you'd like? IMO this dynamic scoping based on tags can only be helpful if we tag tests/apps that has limited dependencies, hence the assumption: "we haven't touch this region, hence we don't need to verify building this app" will work.

@PerMac I mostly agree with you, but I also think it's fine to experiment with filtering of the tests. My biggest worry is that it's already quite hard to understand what twister tests will be run when already, and this change won't make that any easier. sdk-nrf is almost a monorepo, in where we have many applications, samples and libraries in the same repository. Monorepos usually have some way of limiting all tests to run on every change, and I think it would be wise for us to do the same.

nordic-piks:
> > > IMO this dynamic scoping based on tags can only be helpful if we tag tests/apps that has limited dependencies, hence the assumption: "we haven't touch this region, hence we don't need to verify building this app" will work.
> > 
> > 
> > So is `cellular` good for tagging if we need to list at least ~40 folders and dozens of header files? Same question applies for `net` tag.
> 
> And it's not only about the number of folders we need to list. To make it complete or good enough is not easy. How much failures we are allowed to accept because of the incompleteness of the information related to tagging? Is it fine that every now and then PR slips in and full twister run fails in nightly (or wherever it's run) as all configurations are not working. In testing side we are taking certain risk as dependencies for different CI plans are limited to increase performance but here we are talking about leaking compilation errors. Is that fine? @PerMac

My assumption when adding tags was to base on test-specs. They already include some risk of failure due to incomplete information. I am asking here for the same amount of risk to accept.
We will not do this perfectly (gathering all dependencies) for the first time. That will be incremental work.
In particular, dependencies should be updated after failure is found and not detected at PR verification.

Full scope (without this dynamic test scope selection) will be run when PR is merged so that failure will be detected earlier than nightly.

> 
> > In general, this change (and others) aim to reduce test scope at CI (PRs) - the goal from task force exercise.
> 
> I understand the intent of this exercise but I'm not sure this is the right approach given the number of dependencies there are for `cellular` and `net` tags. I could ask why there is a tag and why not just do it based on cellular and net folders? I know tagging is more flexible but you say elsewhere that a simple solution is preferred and not set `net` tag into a configuration in `cellular` folder. But that wouldn't change the fundamental idea of this PR which I'm not convinced of.
> 

Tagging tests and samples is very similar to selecting based on folders, and as you said, it does not change the fundamental idea.
If there were mechanisms to select tests by folders, the list of dependencies still would need to be created.
Regarding not setting `net` tag into `cellular` folder, this was only my idea (to keep this simple), we can change it and have e.g. `net` tag spread in multiple different folders.
A different approach would be to divide tags more e.g. create cellular_iot, cellular_yyy and add them to tests/samples that are more isolated (with a small number of dependencies). And remove tags from those which has a lot of dependencies.
(but as I said, the number of dependencies is not a problem for this solution, still even 50 folders is much less than the whole sdk-nrf and other modules)
I am asking for some help from your side on how to organize this better.

> > There was already finished exercise with static approach - integration quarantine
> > (https://nordicsemi.atlassian.net/wiki/spaces/NCS/pages/530843071/Compilation+scope+table+in+NCS+integration+Twister+scope), however it did not bring expected speedups.
> 
> There are few reasons why this didn't work IMO. First of all it's a manual process and not everybody participates into those. You would need to poll those people who haven't done it several times and for a longer period of time. Secondly, people (including myself) doesn't have any clue how much we should really strip out from integration runs. What's the target and how much risk we can tolerate? I'm sure we could do a lot more with this approach if we wanted as there are so many similar sample configurations and duplicates for different chips that are the same from the sample perspective. That being said, I think this method is also very cumbersome from maintenance point of view. We add a new chip and it's missed from the quarantine perspective 99,99% of the time. Same for any added configuration or sample. Nobody remembers this kind of configuration that's outside the sample code folder. It would maybe be easier if that was defined in the sample.yaml rather than the outside-of-sample-codebase huge quarantine file. I've heard many people thinking (including myself) that "integration_platforms" would have been config for PR runs but that's not the case.

At the https://nordicsemi.atlassian.net/wiki/spaces/NCS/pages/530843071/Compilation+scope+table+in+NCS+integration+Twister+scope
there is a summary that tries to explain the results of this exercise.


> 
> > Maybe you can suggest different way of grouping tests and samples and linking them with code?
> 
> The only easy optimization would be to only compile specific sample if only its codebase is modified. This doesn't help for those modifying libraries but this would reduce some load from CI and make twister stage really quick for sample/app modifications. My question would be that do we absolutely need this kind of optimization for libraries? And if yes, has everything being done in the infrastructure side to speed things up.

Building only specific tests and samples when its code has changed is already implemented.
Regarding different optimizations, yes, there were done (ccache, faster machines). But none of them solved the most important issue - why do we try to build something that has not changed and its dependencies also did not change - this is simply a waste of time and money. Why developer should wait for tests that are not related to its change?

PerMac:
@gregersrygg @trantanen Sorry for late reply. I was on holidays but also I am in contact with @nordic-piks and these changes where discussed between us and we are on the same page, so no need of repetition:). You are correct with all your worries and ideas. We also thought about them and decided to give a try to a method which is already used in the upstream zephyr. It is obviously not perfect and has some more potential pitfalls. But as Piotr mentioned: we decided to take the risk of regression leaking into the main. The failure will be seen almost immediately, hence the risk that it will affect a release is minimal. 
BTW you can check e.g. this PR https://github.com/nrfconnect/sdk-nrf/pull/16357 and it's scope https://jenkins-ncs.nordicsemi.no/job/latest/job/sdk-nrf/job/PR-16357/1/testReport/ to see dynamic scope at work when only samples/tests/applications are modified. Only 4 configurations were built by twister and it took 5 mins. We won't be able to be faster due to the infrastructure setup time. Such cases were the easiest, since it is clear only a relevant build can be affected.

nordic-piks:
asser_tracker has its own tag - https://github.com/nrfconnect/sdk-nrf/pull/16424
thus removing from there.

nordic-piks:
Rebased and removed tag from tests (so they can executed on simulated platforms). Keep tag for samples.

nordic-piks:
Rebased + added dependencies to all sources which are compiled when all samples from cellular are build - did such analysis to minimize risk of missing dependency.

PR #16141, opened 2024-06-21
applications: nrf_desktop: Turn off MRAM suspension

zycz:
Turn off suspending MRAM for nrf_desktop application. This change is required to make radio core working correctly during suit smp dfu.


zycz:
Probably functionality should be moved to separate nrf_dresktop module.

PR #16148, opened 2024-06-24
scripts: Update minimum required toolchain for the v2.7.0 release

jangalda-nsc:
'2.6.20240605' is pointer to 'v2.7.0-rc2' in toolchain-manager index This is required by CMAKE_VERSION.
https://cmake.org/cmake/help/latest/variable/CMAKE_VERSION.html

### DIFF
--- a/scripts/ncs-toolchain-version-minimum.txt
+++ b/scripts/ncs-toolchain-version-minimum.txt
@@ -1,3 +1,3 @@
 # This file specifies the minimum nRF Connect SDK Toolchain that will be
 # working with this release.
-nrf-connect-sdk-toolchain=2.6.20240304
+nrf-connect-sdk-toolchain=2.6.20240605


thst-nordic (approved):

scripts/ncs-toolchain-version-minimum.txt
```diff
@@ -1,3 +1,3 @@
 # This file specifies the minimum nRF Connect SDK Toolchain that will be
 # working with this release.
-nrf-connect-sdk-toolchain=2.6.20240304
+nrf-connect-sdk-toolchain=2.6.20240605
```
tejlmand, line 3: shouldn't we be having a `2.7.<something>` ?
Or is the toolchain itself still reporting it as a 2.6 release.

tejlmand (approved):

shanthanordic (approved):

scripts/ncs-toolchain-version-minimum.txt
jangalda-nsc: We decided to decouple (or at least loosely couple) sdk-nrf's and bundle's versions. 
Bundle version is built based on:
* content of sdk-nrf's VERSION file  
* date when bundle was built.

For sdk-nrf `v2.7.0` release we are using toolchain bundle built before `v2.7.0-rc-1` tag was placed (so sdk-nrf version was 2.6.99), so its version starts with `2.6`
For `v2.6.0` release, we used toolchain bundle built after `v2.6.0-rc1` was placed (there were changes in toolchain after rc`), so its version started with `v2.6` too

PR #16150, opened 2024-06-24
sysbuild: Remove include functions

nordicjm:
Removes include functions and just includes files directly where this is appropriate

### DIFF
--- a/cmake/mesh_dfu_metadata.cmake
+++ b/cmake/mesh_dfu_metadata.cmake
@@ -41,3 +41,7 @@ function(mesh_dfu_metadata)
     COMMAND_EXPAND_LISTS
   )
 endfunction()
+
+if(SYSBUILD)
+  mesh_dfu_metadata()
+endif()
--- a/cmake/sysbuild/b0_packaging.cmake
+++ b/cmake/sysbuild/b0_packaging.cmake
@@ -1,32 +1,38 @@
 # Copyright (c) 2024 Nordic Semiconductor ASA
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 
-# B0 will boot the app directly, create the DFU zip with both update candidates
-include(${ZEPHYR_NRF_MODULE_DIR}/cmake/fw_zip.cmake)
+function(dfu_app_b0_zip_package)
+  # B0 will boot the app directly, create the DFU zip with both update candidates
+  include(${ZEPHYR_NRF_MODULE_DIR}/cmake/fw_zip.cmake)
 
-sysbuild_get(app_fw_info_firmware_version IMAGE ${DEFAULT_IMAGE} VAR CONFIG_FW_INFO_FIRMWARE_VERSION KCONFIG)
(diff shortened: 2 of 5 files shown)


tejlmand:
why is this needed for NCS v2.7.0 ?

Although i'm much in favor of this cleanup then it doesn't seem to be fixing any issues.

nordicjm:
> why is this needed for NCS v2.7.0 ?
> 
> Although i'm much in favor of this cleanup then it doesn't seem to be fixing any issues.

You left a comment in another PR to clean it up

shanthanordic:
Not a critical for NCS 2.7.0, cleared milestone

tejlmand (commented):
minor nit observed.

tejlmand:
> You left a comment in another PR to clean it up

correct, but it not a priority to have it cleaned up during stabilization period, can wait till after release.

cmake/sysbuild/fast_pair_hex.cmake
```diff
@@ -4,43 +4,47 @@
...
-  "${FP_PROVISIONING_DATA_HEX}"
-  )
-
-set_property(
-  GLOBAL PROPERTY
-  ${FP_PARTITION_NAME}_PM_TARGET
-  ${FP_PARTITION_NAME}_target
-  )
+function(fast_pair_hex)
+  set(FP_PARTITION_NAME bt_fast_pair)
```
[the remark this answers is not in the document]
nordicjm: Fixed

cmake/mesh_dfu_metadata.cmake
```diff
@@ -41,3 +41,7 @@ function(mesh_dfu_metadata)
     COMMAND_EXPAND_LISTS
   )
 endfunction()
+
+if(SYSBUILD)
+  mesh_dfu_metadata()
```
57300, line 46: Does this work? I don't think `SYSBUILD` is set in sysbuild itself, only in images.

cmake/sysbuild/b0_packaging.cmake
```diff
@@ -1,32 +1,38 @@
...
+      DEPENDS
+      ${DEFAULT_IMAGE}_extra_byproducts
+      ${DEFAULT_IMAGE}_signed_kernel_hex_target
+      s1_image_extra_byproducts
+      s1_image_signed_kernel_hex_target
+      )
+  endif()
+endfunction()
+
+dfu_app_b0_zip_package()
```
57300, line 38: nit: How about wrapping this call in `if(SB_CONFIG_DFU_ZIP_APP)`, instead of the function body?

cmake/mesh_dfu_metadata.cmake
nordicjm: Yes, it is set

cmake/sysbuild/b0_packaging.cmake
nordicjm: Fixed

57300 (approved):

carlescufi (approved):

PR #16151, opened 2024-06-24
logging: crash log support for nRF RPC backend

Damian-Nordic:
* Put log forwarder and backend implementations into separate source files.
* Add Kconfig option that enables saving the crash log in the "crash_log" retained RAM partition, and adds an nRF RPC command that allows a remote user to read the crash log after the device reboots.
* Add snippets to enable logging over nRF RPC, including crash log support, on both the client and server.
* Add "logging crash_log" shell command for retrieving the crash log from the server device.
* Enable testing the crash log retention by triggering k_oops() when all buttons are pressed in the server appplication.


applications/protocols_serialization/client/snippets/logging/logging.conf
```diff
@@ -0,0 +1,7 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+CONFIG_LOG_FORWARDER_RPC=y
```
canisLupus1313, line 7: Seems that this config is incomplete or name is confusing. In logging snippet I would expect enabled logging subsystem. And it seems that for logging we have to add both debug and logging snippet. Order of those might matter.

canisLupus1313:
@Damian-Nordic  Can you explain why haven't You used: https://docs.zephyrproject.org/latest/services/debugging/coredump.html

applications/protocols_serialization/client/snippets/logging/logging.conf
Damian-Nordic: I can rename the snippet to "logrpc" or someting. Didn't want to do it because we don't add "rpc" to other snippets that enable RPC groups, but I can change this in this case.

Damian-Nordic:
> @Damian-Nordic Can you explain why haven't You used: https://docs.zephyrproject.org/latest/services/debugging/coredump.html

The coredump module has currently two backends:
- DEBUG_COREDUMP_BACKEND_LOGGING - which can't be used directly in our project because we can't send any RPC events anymore once the core dump is generated.
- DEBUG_COREDUMP_BACKEND_FLASH_PARTITION - which requires a dedicated flash partition - I don't really like storing the core dump into flash as it can easily wear out the flash if a faulty firmware is installed.

We could use `coredump` if it supported the retained memory but currently it doesn't (we could actually contribute a backend for that).

But... the good thing about the current approach is that retains **any** log messages after entering the panic mode, which means that if you reserve enough space in the "crash_log" partition (like 4kB), you can enable `DEBUG_COREDUMP_BACKEND_LOGGING` and the core dump will be stored in the retained RAM as a log.

subsys/nrf_rpc/nrf_rpc_uart.c
```diff
@@ -211,7 +211,6 @@ static int send(const struct nrf_rpc_tr *transport, const uint8_t *data, size_t
 {
 	uint8_t crc[2];
 	LOG_HEXDUMP_DBG(data, length, "Sending frame");
-	LOG_HEXDUMP_DBG(crc, CRC_SIZE, "frame crc");
```
alxelax, line 214: Why was it removed?

applications/protocols_serialization/client/snippets/logging/logging.conf
alxelax: Does logging have sense without debugging?

applications/protocols_serialization/client/src/logging_shell.c
```diff
@@ -0,0 +1,42 @@
...
+#include <logging/log_rpc.h>
+#include <zephyr/kernel.h>
+#include <zephyr/shell/shell.h>
+
+#include <stdlib.h>
+
+static int logging_get_crash_log_cmd(const struct shell *sh, size_t argc, char *argv[])
+{
+	int rc;
+	char buffer[256];
```
alxelax, line 16: Could it be configurable?

applications/protocols_serialization/server/prj.conf
```diff
@@ -13,6 +13,7 @@ CONFIG_NRF_RPC_ZCBOR_BACKUPS=1
 CONFIG_NRF_RPC_THREAD_STACK_SIZE=4096
 CONFIG_NRF_RPC_GROUP_INIT_WAIT_TIME=-1
 CONFIG_NRF_RPC_UART_TRANSPORT=y
+CONFIG_DK_LIBRARY=y
```
alxelax, line 16: Why is it required?

applications/protocols_serialization/server/snippets/debug/debug.conf
```diff
@@ -15,15 +15,16 @@ CONFIG_LOG=y
 
 # Use separate thread for logging
 CONFIG_LOG_MODE_DEFERRED=y
+CONFIG_LOG_BUFFER_SIZE=4096
```
alxelax, line 18: 1. Is it possible to pass 4k logging buffer over the serialization interface?
2. If retained memory area is 1k why logging buffer is 4k?

applications/protocols_serialization/server/src/main.c
```diff
@@ -16,6 +17,13 @@ static void err_handler(const struct nrf_rpc_err_report *report)
 	//k_oops();
 }
 
+static void button_handler(uint32_t button_state, uint32_t has_changed)
+{
+	if (button_state & DK_ALL_BTNS_MSK) {
+		k_oops();
```
alxelax, line 23: Seems this is testing stuff that shouldn't be merged.

applications/protocols_serialization/server/src/main.c
```diff
@@ -30,5 +38,11 @@ int main(void)
 
 	LOG_INF("RPC server ready");
 
+	ret = dk_buttons_init(button_handler);
```
alxelax, line 41: This is the same. We shouldn't pollute source by stub testing.

include/logging/log_rpc.h
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef LOG_RPC_H_
+#define LOG_RPC_H_
+
+#include <zephyr/shell/shell.h>
```
alxelax, line 10: Why is this header required here?

subsys/logging/CMakeLists.txt
```diff
@@ -4,4 +4,8 @@
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 #
 
-zephyr_sources_ifdef(CONFIG_LOG_RPC log_rpc.c)
+if (CONFIG_LOG_FORWARDER_RPC OR CONFIG_LOG_BACKEND_RPC)
```
alxelax, line 7: Does this condition have a sense in general? Is it possible to have logging without Zephyr?

subsys/logging/log_backend_rpc.c
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	length = MIN(length, header.size - offset);
+	NRF_RPC_CBOR_ALLOC(group, rsp_ctx, 5 + length);
+
+	if (!zcbor_bstr_start_encode(rsp_ctx.zs)) {
+		goto discard;
+	}
+
+	if (retention_read(crash_log, sizeof(header) + offset, rsp_ctx.zs[0].payload_mut, length)) {
```
alxelax, line 140: Why is this reading required?

subsys/logging/log_backend_rpc.c
```diff
@@ -0,0 +1,310 @@
...
+	if (!zcbor_bstr_end_encode(rsp_ctx.zs, NULL)) {
+		goto discard;
+	}
+
+	goto send;
+
+discard:
+	NRF_RPC_CBOR_DISCARD(group, rsp_ctx);
+err:
+	/* Send empty string */
```
alxelax, line 155: Why is the empty string required?

subsys/logging/Kconfig
```diff
@@ -37,6 +37,21 @@ config LOG_BACKEND_RPC_BUFFER_SIZE
 	  Defines the maximum log line length that can be sent to the log forwarder
 	  over nRF RPC without fragmentation.
 
+config LOG_BACKEND_RPC_CRASH_LOG
+	bool "nRF RPC crash log"
+	select RETENTION
+	select RETENTION_MUTEX_FORCE_DISABLE
```
alxelax, line 43: Are these mutexes used anywhere?

subsys/logging/log_forwarder_rpc.c
```diff
@@ -0,0 +1,89 @@
...
+	uint8_t level;
+	struct zcbor_string message;
+	bool decoded_ok;
+
+	decoded_ok = zcbor_uint_decode(ctx->zs, &level, sizeof(level));
+	decoded_ok = decoded_ok && zcbor_bstr_decode(ctx->zs, &message);
+	nrf_rpc_cbor_decoding_done(group, ctx);
+
+	if (!decoded_ok) {
+		goto out;
```
alxelax, line 35: Probably immediate return is simpler? :)

subsys/logging/log_rpc_internal.h
```diff
@@ -0,0 +1,42 @@
+/*
```
alxelax, line 1: I guess everything what isn't in public API folder is internal. How much sense is in `internal` suffix?

alxelax (commented):
I have a couple of questions. Please take a look at my comments.

subsys/nrf_rpc/nrf_rpc_uart.c
Damian-Nordic: Because it was incorrect (it printed `crc` before it was calculated). Also, we print the entire frame so we can always calculate the actual CRC if needed and it's worth reducing the amount of logs.

applications/protocols_serialization/client/src/logging_shell.c
Damian-Nordic: Sure, I mean, this is just an application, not a generic module so I wanted to keep it simple, but I can add a Kconfig.

applications/protocols_serialization/server/src/main.c
Damian-Nordic: How will this feature be tested if we don't provide such an trigger?

subsys/logging/log_backend_rpc.c
Damian-Nordic: I don't understand. Could you elaborate what's unclear?

applications/protocols_serialization/client/snippets/logging/logging.conf
Damian-Nordic: Sure, the crash log feature can be tested without logger enabled on the client side so I didn't want needlessly couple these two.

applications/protocols_serialization/server/src/main.c
Damian-Nordic: This meant to be a way to test this feature in our CI, not stub testing.

include/logging/log_rpc.h
Damian-Nordic: Right, copy-paste error.

subsys/logging/CMakeLists.txt
Damian-Nordic: I don't understand the question, I just didn't want to create a Zephyr library if neither log forwarder nor backend are enabled.

subsys/logging/Kconfig
Damian-Nordic: Not anymore after I force-disable them. :) They're used internally by the retention API, which is undesired given that we use the API from the fatal error handler.

subsys/logging/log_backend_rpc.c
Damian-Nordic: Because the API contract which I documented is that an empty string means "end of file". We could return an error code as well but for a caller it doesn't really make a difference why the crash log is not accessible (I thought).

subsys/logging/log_rpc_internal.h
Damian-Nordic: I added also a public `log_rpc.h` header so I wanted to make it faster to find the correct file and easier to distinguish open tabs in VS Code. I also found a few other examples of that in `nrf/subsys/` so I didn't think I would break any convention.

applications/protocols_serialization/server/src/main.c
alxelax: I guess there is misunderstanding of the ncs-sdk application purposes.
Application is logically completed functionality that is ready for being used by Nordic's customers `as is` for their production\deployment and etc `from the box`.

This is not for testing, not even for the feature demonstration (for demonstration purposes there are samples).

It is why I'm confused by seeing a lot of garbage code, inaccurate configuration with huge buffers (without clear explanation of the reason of them) etc.

All functionality that is added within the serialization project should (shall?) be tested separately in form of system\unit\performance\robustness tests implemented in the test environment and integrated in CI.

Application should be cleaned up from any testing stubs and stuffs.
If the initially it was assumed for testing then application is wrong place for it. Probably we do not need application at all and just a couple of samples will be sufficient, for example for demonstration crash logging.

applications/protocols_serialization/server/prj.conf
alxelax: I guess my comment (https://github.com/nrfconnect/sdk-nrf/pull/16151#discussion_r1654891873) is fair enough for this too.

applications/protocols_serialization/server/src/main.c
alxelax: Mark as solved to not duplicate discussion started above.

applications/protocols_serialization/server/prj.conf
alxelax: Mark as solved to not duplicate discussion referenced by link.

subsys/logging/CMakeLists.txt
alxelax: Question was mostly about ability to run server's part without Zephyr considering that Bluetooth Host is part of Zephyr.

subsys/logging/log_backend_rpc.c
alxelax: Nevermind. Understood by self :)

applications/protocols_serialization/server/src/main.c
Damian-Nordic: I'm not sure what you mean by the garbage code and huge buffers etc. I thought it's clear that while the server can satisfy your requirements of an application to some extent, the client is just an exemplary code because we need some way to validate the server's functionality.

I personally don't feel that turning the client into a sample in unrelated directory would make it more logical for our customers but feel free to bring this up at the stand up meeting. Eventually, this is up to PMT what exactly they want to be delivered.

I get your point that an application (let's confine this to our server) should be optimized etc. but I don't think we have to artificially copy its functionality into a bunch of samples just to satisfy the requirements that "applications" shouldn't be used for testing. I don't think this is true even today - Thingy:53 applications do have dedicated integration or nightly tests.

And if we add extra snippets to enable debug features, what's exactly the harm for the customers in deploying the application in production (without enabling the debug features)?

subsys/logging/CMakeLists.txt
Damian-Nordic: Sorry, but I still don't get it :) This library depends on Zephyr logging subsystem but the condition is needed not because I think it's possible to have logging without Zephyr so I need more context.

subsys/logging/log_rpc_internal.h
alxelax: They are distinguished by C including rules.
#include <nrf_rpc.h> is always public since it is passed with system paths.
#include "nrf_rpc.h" is always internal, since developer explicitly said to compiler to look for it in the current folder.

but up to you if it is easier to develop with word suffix, let it be.

Damian-Nordic: I'm not saying there's any issues for the compiler. Rather for me when searching for the right file :). But I will rename it to `log_rpc_group.h` as this is what it actually provides.

applications/protocols_serialization/server/src/main.c
alxelax: Sure, I'll raise up the question on stand up. I'm really not a big fan of maintaining some code that looks like `all inclusive` just for testing purposes.

That the application exists for not testing purposes doesn't mean the application shouldn't be tested by self. I do not see any contradiction in testing application functionality as you pointed for some platforms. Even more, listed above type of tests should include application testing to not do it manually before every release.

Also, there is no contradiction to have debug logging(in form of snippet for example) in the final product if a customer deployed the network in test mode to gather statistics or figure out some instabilities.

subsys/logging/CMakeLists.txt
alxelax: Considering said above means it logging was enabled, then Zephyr is present. Am I wrong?

Damian-Nordic: No, you're correct. How is it related to the condition though?

applications/protocols_serialization/client/snippets/logging/logging.conf
canisLupus1313: @tomaszkob89 From CI POV how will it be easier to organise this?

subsys/logging/CMakeLists.txt
alxelax: This is just clean up. The change isn't worth time that we've already for discussion. If you feel that you are 100% right in your assurance, just leave at `as is`.

Damian-Nordic: It seems you assume that this code intents to do something which in fact it does not. Hence I'm confused :)
If I remove the `if`, yes - it will work, but it will generate a warning when neither log forwarder nor backend are enabled. It is because `zephyr_library()` will just call CMake's `add_library()` under the hood but then no `target_sources()` will be called for the library.

To that end, Zephyr provides `zephyr_library_property(ALLOW_EMPTY TRUE)` functionality, which I can use instead of the condition if you prefer it, but I saw such a condition in other places in NCS as well.

applications/protocols_serialization/client/snippets/logging/logging.conf
tomaszkob89: Can't this be just a part of proj.conf? (As I understand we want to have it enabled even if debugging is disabled, so far it is a single flag so a separate snippet is unnecessary?)

Damian-Nordic: I'll keep the current approach until we talk with PMT on what we want to deliver.

Damian-Nordic:
@canisLupus1313 @alxelax I removed the testing code and addressed other comments.

alxelax (commented):
LGTM just fix compliance issues.

alxelax (approved):

canisLupus1313 (approved):

PR #16157, opened 2024-06-24
samples: wifi: twt: Fix build with default configuration

D-Triveni:
TWT sample fails without extra `TRAFFIC_GEN_REMOTE_IPV4_ADDR` config, so added it as a default configuration in prj.conf.

### DIFF
--- a/samples/wifi/twt/README.rst
+++ b/samples/wifi/twt/README.rst
@@ -95,6 +95,10 @@ The following is an example of the CLI command:
 
    west build -b nrf7002dk/nrf5340/cpuapp
 
+.. important::
+
+   It is mandatory to set the :kconfig:option:`CONFIG_TRAFFIC_GEN_REMOTE_IPV4_ADDR` Kconfig option to the IPv4 address of the host running the traffic generator server. If this is set incorrectly, the sample may fail to work as expected.
+
 Testing
 =======
 
--- a/samples/wifi/twt/prj.conf
+++ b/samples/wifi/twt/prj.conf
@@ -90,5 +90,7 @@ CONFIG_NET_CONFIG_MY_IPV4_GW="192.168.1.1"
 CONFIG_NET_MGMT_EVENT_QUEUE_TIMEOUT=5000
 
 CONFIG_TRAFFIC_GEN=y
+# IPv4 address of the host running the traffic generator server
+CONFIG_TRAFFIC_GEN_REMOTE_IPV4_ADDR="192.168.0.1"
 # Needed because traffic gen uses more sockets, so, keeping an upper limit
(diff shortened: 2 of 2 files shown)


sachinthegreen (approved):

VivekUppunda (approved):

samples/wifi/twt/prj.conf
```diff
@@ -90,5 +90,7 @@ CONFIG_NET_CONFIG_MY_IPV4_GW="192.168.1.1"
 CONFIG_NET_MGMT_EVENT_QUEUE_TIMEOUT=5000
 
 CONFIG_TRAFFIC_GEN=y
+# IPv4 address of the host running the traffic generator server
+CONFIG_TRAFFIC_GEN_REMOTE_IPV4_ADDR="192.168.0.1"
```
krish2718, line 94: This was intentionally left out of default config, we don't have put some default address without user's knowledge, it will start failing and user has no idea why. So, a build assert is added to force the user to choose an IP, see https://github.com/nrfconnect/sdk-nrf/blob/main/samples/wifi/twt/modules/traffic_gen/src/main.c#L319

krish2718 (changes requested):

D-Triveni:
> Then in the sample documentation we can stress a bit better that this Kconfig symbol needs to be looked-at and modified accordingly (if needed) and state the reason for that (e.g. otherwise the sample can start to fail).

Addressed.

samples/wifi/twt/README.rst
```diff
@@ -95,6 +95,10 @@ The following is an example of the CLI command:
 
    west build -b nrf7002dk/nrf5340/cpuapp
 
+.. note::
```
krish2718, line 98:
```suggestion
.. important::
```

krish2718 (approved):

ioannisg (approved):

ioannisg (approved):
lgtm

krish2718 (approved):

kittydepa (approved):

samples/wifi/twt/README.rst
```diff
@@ -95,6 +95,10 @@ The following is an example of the CLI command:
 
    west build -b nrf7002dk/nrf5340/cpuapp
 
+.. important::
+
+   It is mandatory to set the :kconfig:option:`CONFIG_TRAFFIC_GEN_REMOTE_IPV4_ADDR` Kconfig option to the IPv4 address of the host running the traffic generator server. If this is set incorrectly, the sample may fail to work as expected.
```
richabp, line 100: I see that this PR has been merged, thus the changes can be implemented in another PR.
The sentence in line 100 should begin on a new line, following the one sentence per line syntax rule in RST.

PR #16159, opened 2024-06-24
samples: crypto: add crypto tag for samples

nordic-piks:
Extend list of paths used to check if crypto tests/samples should be executed.

### DIFF
--- a/samples/crypto/aes_cbc/sample.yaml
+++ b/samples/crypto/aes_cbc/sample.yaml
@@ -6,7 +6,7 @@ sample:
 tests:
   sample.aes_cbc.cc3xx:
     sysbuild: true
-    tags: introduction psa cc3xx sysbuild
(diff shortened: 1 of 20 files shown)


nordic-piks:
In this case, there is defined tag (e.g. crypto) and files/folder with source code related to that area.

miha-nordic (approved):

nordic-piks:
> LGTM (But I wonder if TF-M tests and samples also should get this treatment)

Thanks, I will add them in separated PR.

nordic-piks:
> Extend list of paths used to check if crypto tests/samples should be executed.

Here it is: https://github.com/nrfconnect/sdk-nrf/pull/16174

nordic-piks:
DNM until NCS 2.7

tejlmand (approved):

PR #16165, opened 2024-06-24
samples: wifi: add wifi tag for samples

nordic-piks:
Extend list of paths used to check if wifi
samples should be executed.

### DIFF
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -97,7 +97,12 @@ posix:
   files:
     - zephyr/lib/posix/
 
-# cbprintf:
-#    files:
-#        - zephyr/lib/os/cbprintf*
-#        - zephyr/lib/posix
+wifi:
+  files:
+    - nrf/drivers/wifi/
+    - nrf/modules/hostap/
+    - modules/lib/hostap/
+    - zephyr/subsys/net/l2/wifi/
+    - zephyr/subsys/net/l2/ethernet/
+    - nrfxlib/nrf_wifi/
+    - nrfxlib/crypto/
--- a/samples/wifi/ble_coex/sample.yaml
+++ b/samples/wifi/ble_coex/sample.yaml
@@ -11,7 +11,7 @@ tests:
     extra_args: CONFIG_MPSL_CX=y ipc_radio_CONFIG_MPSL_CX=y
                 CONFIG_COEX_SEP_ANTENNAS=y
     platform_allow: nrf7002dk/nrf5340/cpuapp
-    tags: ci_build sysbuild
+    tags: ci_build sysbuild wifi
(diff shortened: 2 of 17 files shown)


scripts/ci/tags.yaml
```diff
@@ -97,7 +97,7 @@ posix:
   files:
     - zephyr/lib/posix/
 
-# cbprintf:
-#    files:
-#        - zephyr/lib/os/cbprintf*
-#        - zephyr/lib/posix
+wifi:
+  files:
+    - nrf/drivers/wifi
```
krish2718, line 102: `modules/hostap/` and `../modules/lib/hostap` too.

scripts/ci/tags.yaml
```diff
@@ -97,7 +97,9 @@ posix:
...
 
-# cbprintf:
-#    files:
-#        - zephyr/lib/os/cbprintf*
-#        - zephyr/lib/posix
+wifi:
+  files:
+    - nrf/drivers/wifi/
+    - nrf/modules/hostap/
+    - nrf/modules/lib/hostap/
```
krish2718, line 104: This is outside `nrf`, hence the `..`

scripts/ci/tags.yaml
```diff
@@ -97,7 +97,9 @@ posix:
...
+    - zephyr/drivers/wifi/
```
krish2718, line 105: This isn't needed for NCS

scripts/ci/tags.yaml
nordic-piks: Thanks, fixed.
The root folder is NCS, so it would be `modules/lib/hostap`

scripts/ci/tags.yaml
nordic-piks: Thanks, removed.

scripts/ci/tags.yaml
nordic-piks: And `modules/hostap` is actually at `sdk-nrf`

nordic-piks:
Updated after evaluation of:
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L497-L501
https://github.com/nrfconnect/sdk-zephyr/blob/main/.github/test-spec.yml#L246-L248
https://github.com/nrfconnect/sdk-nrfxlib/blob/main/.github/test-spec.yml#L168-L173

VivekUppunda (approved):

bama-nordic (approved):

bjda (approved):

D-Triveni (approved):

tejlmand (approved):

muraliThokala (approved):

nordic-piks:
Just rebased after NCS 2.7

PR #16167, opened 2024-06-24
samples: zigbee: add zigbee tag for samples

nordic-piks:
Extend list of paths used to check if zigbee
samples should be executed.

### DIFF
--- a/samples/zigbee/ncp/sample.yaml
+++ b/samples/zigbee/ncp/sample.yaml
@@ -12,13 +12,13 @@ tests:
       - nrf52833dk/nrf52833
       - nrf21540dk/nrf52840
       - nrf5340dk/nrf5340/cpuapp
-    tags: sysbuild
+    tags: sysbuild zigbee
   sample.zigbee.ncp.usb:
     sysbuild: true
     build_only: true
     platform_allow: nrf52840dk/nrf52840 nrf52833dk/nrf52833 nrf21540dk/nrf52840
       nrf5340dk/nrf5340/cpuapp
-    tags: ci_build ncp sysbuild
+    tags: ci_build ncp sysbuild zigbee
     extra_args: FILE_SUFFIX=usb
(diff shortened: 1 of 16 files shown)


nordic-piks:
In this case, there is defined tag (e.g. zigbee) and files/folder with source code related to that area.

nordic-piks:
Updated after evaluation of:
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L222-L241
https://github.com/nrfconnect/sdk-nrfxlib/blob/main/.github/test-spec.yml#L47-L58
https://github.com/nrfconnect/sdk-zephyr/blob/main/.github/test-spec.yml#L94-L98

nordic-piks:
DNM until NCS 2.7

tejlmand (approved):

nordic-piks:
Just rebase and added path for tests and samples.

scripts/ci/tags.yaml
```diff
@@ -154,3 +154,35 @@ crypto:
     - zephyr/include/drivers/serial/
     - zephyr/modules/mbedtls/
     - modules/crypto/
+
+zigbee:
+  files:
```
mariusz-nordicsemi, line 159: What about following files included in test-spec.yaml?
- nrf/applications/zigbee_weather_station
- nrf/samples/nrf5340/multiprotocol_rpmsg

mariusz-nordicsemi (changes requested):

samples/zigbee/ncp/sample.yaml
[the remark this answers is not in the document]
nordic-piks: Added, thanks ;)

scripts/ci/tags.yaml
nordic-piks: Added.

mariusz-nordicsemi (approved):

PR #16169, opened 2024-06-24
tests: subsys: suit: add suit tag for samples

nordic-piks:
Extend list of paths used to check if suit
samples should be executed.

### DIFF
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -230,3 +230,22 @@ tfm:
     - zephyr/modules/trusted-firmware-m/
     - modules/crypto/
     - nrfxlib/crypto/
+
+suit:
+  files:
+    - nrf/include/bl*
+    - nrf/include/fprotect.h
+    - nrf/include/dfu/*suit*
+    - nrf/include/sdfw/sdfw_services/
+    - nrf/cmake/
+    - nrf/sysbuild/
+    - nrf/lib/dk_buttons_and_leds/
+    - nrf/subsys/suit/
+    - nrf/subsys/dfu/
+    - nrf/subsys/nrf_rpc/
+    - nrf/subsys/nrf_security/
+    - nrf/subsys/sdfw_services/
+    - nrf/subsys/mgmt/suitfu/
+    - nrf/tests/subsys/suit/
+    - modules/lib/suit-processor/
+    - modules/lib/suit-generator/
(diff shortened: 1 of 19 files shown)


nordic-piks:
In this case, there is defined tag (e.g. suit) and files/folder with source code related to that area.

tomchy (approved):

scripts/ci/tags.yaml
```diff
@@ -97,7 +97,14 @@ posix:
   files:
     - zephyr/lib/posix/
 
-# cbprintf:
-#    files:
-#        - zephyr/lib/os/cbprintf*
-#        - zephyr/lib/posix
+suit:
```
robertstypa, line 100: test-sdk-dfu is triggered by changes done in the following files/folders (test-spec.yaml):
```
"CI-suit-dfu-test":
  - "include/bl*"
  - "include/fprotect.h"
  - "include/dfu/*suit*"
  - "include/sdfw/sdfw_services/*"
  - "cmake/**/*"
  - "sysbuild/**/*"
  - "lib/dk_buttons_and_leds/**/*"
  - "subsys/dfu/*"
  - "subsys/nrf_rpc/**/*"
  - "subsys/nrf_security/**/*"
  - "subsys/sdfw_services/**/*"
  - "subsys/suit/**/*"
  - "samples/suit/**/*"
  - "subsys/mgmt/suitfu/**/*"
  - "tests/subsys/suit/**/*"

"CI-suit-dfu-test-extended":
  - "include/dfu/*suit*"
  - "include/sdfw/sdfw_services/*"
  - "subsys/sdfw_services/**/*"
  - "subsys/suit/**/*"
  - "samples/suit/**/*"
  - "subsys/mgmt/suitfu/**/*"
  - "tests/subsys/suit/**/*"
```
List above has been created during dependency analysis done by the team responsible for SUIT.
In my opinion the same files/folders shall be used to limit Twister tests/builds - if there is a plan to do so.

robertstypa (changes requested):

scripts/ci/tags.yaml
nordic-piks: Updated after evaluation of:
https://github.com/nrfconnect/sdk-nrf/blob/main/.github/test-spec.yml#L521-L545
https://github.com/nrfconnect/suit-generator/blob/ad6502a937d1e6e11824d3fb546f3d2027c5766a/.github/test-spec.yml#L1-L2
https://github.com/nrfconnect/suit-processor/blob/05089f4f7452ff50f09e2e2d776105c630381b1e/.github/test-spec.yml#L1-L2

robertstypa (approved):

tejlmand (approved):

nordic-piks:
Just rebased and added path for tests.

PR #16154, opened 2024-06-24
sysbuild: Remove non-functional environmental key support

nordicjm:
Removes using the environmental or command line parameters for specifying a key as the underlying functionality in child/parent has never worked so does not need reimplementation in sysbuild

PR #16173, opened 2024-06-24
subsy/bootloader/cmake/debug_keys: fix SB_SIGNING_KEY supply

nvlsianpu:
ref:NCSDK-28124

### DIFF
--- a/subsys/bootloader/cmake/debug_keys.cmake
+++ b/subsys/bootloader/cmake/debug_keys.cmake
@@ -22,17 +22,17 @@ if (DEFINED ENV{SB_SIGNING_KEY_FILE} AND NOT SB_SIGNING_KEY_FILE)
   if (NOT EXISTS "$ENV{SB_SIGNING_KEY_FILE}")
     message(FATAL_ERROR "ENV points to non-existing PEM file '$ENV{SB_SIGNING_KEY_FILE}'")
   else()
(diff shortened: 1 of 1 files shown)


sigvartmh (approved):

nvlsianpu:
@tejlmand
@shanthanordic 
If this PR won't go into NCS 2.7.0 then I have to add known issue instead.

shanthanordic:
Pushed it to NCS 2.8.0 as this was already there

hakonfam (approved):

tejlmand (changes requested):
Has this been tested to work ?

For sure it doesn't work with absolute paths.

tejlmand (changes requested):
This has not been working for several NCS releases, so therefore this functionality was removed for sysbuild here: https://github.com/nrfconnect/sdk-nrf/pull/16154

If we consider this to be a bug that we want to fix for parent/child images, then we should support the same for sysbuild.

@nordicjm feel free to give additional comments.

Also, if this PR should be merged then a proper commit message should be written.

nordicjm:
I do not think this should be fixed, it should be listed as a known issue and references to it removed from the help text. Fixing this gives a false sense of security of the feature working for users that might go back to older versions and be aware that this is a silent failure. The feature has never worked, let's just remove it.

nvlsianpu:
@nordicjm @tejlmand I agree with your opinion.
This was never been seen working, child images are close to the end-of-life.

PR #16174, opened 2024-06-24
samples: tfm: add tfm tag for samples

nordic-piks:
Extend list of paths used to check if tfm
tests/samples should be executed.


nordic-piks:
DNM until NCS 2.7

Vge0rge (approved):

tejlmand (approved):

nordic-piks:
Just rebased and added samples and tests folders.

PR #16039, opened 2024-06-17
nrf_security/cracen: KMU keys support

vili-nordic:
Adds missing cases for Cracen KMU driver to PSA crypto driver wrappers.

test_crypto: PR-641

PR #16181, opened 2024-06-25
Reserve memory for the KMU push aread

Vge0rge:
This reserves 64 bytes on the top RAM address for the KMU to push keys into.  Since the KMU sets it's push address when we provision the device we need to do to allow using it across different images/dfus.

### DIFF
--- a/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
+++ b/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
@@ -8,6 +8,7 @@
 #include <cracen/mem_helpers.h>
 #include "cracen_psa.h"
 #include "platform_keys/platform_keys.h"
+#include <nrf_security_mutexes.h>
 
 #include <sicrypto/drbghash.h>
 #include <sicrypto/ecc.h>
@@ -34,6 +35,8 @@ enum asn1_tags {
 };
 
 extern const uint8_t cracen_N3072[384];
+
+extern nrf_security_mutex_t cracen_mutex_symmetric;
 
 static psa_status_t check_brainpool_alg_and_key_bits(psa_algorithm_t alg, size_t key_bits)
 {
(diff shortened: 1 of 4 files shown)


subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
```diff
@@ -1325,21 +1329,26 @@ psa_status_t cracen_export_key(const psa_key_attributes_t *attributes, const uin
...
-		return PSA_SUCCESS;
+		/* The kmu_push_area is guarded by the symmetric mutex since it is the most common use case.
+		 * Here the decision was to avoid defining another mutex to handle the push buffer for the rest
+		 * of the use cases.
+		 */
+		nrf_security_mutex_lock(cracen_mutex_symmetric);
+		status = cracen_kmu_prepare_key(key_buffer);
+		if (status == SX_OK) {
+			memcpy(data, kmu_push_area, key_out_size);
+			*data_length = key_out_size;
```
vili-nordic, line 1346: We should clean key here

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.c
```diff
@@ -211,7 +218,13 @@ int cracen_kmu_prepare_key(const uint8_t *user_data)
 int cracen_kmu_clean_key(const uint8_t *user_data)
 {
 	(void)user_data;
+	/* The kmu_push_area is guarded by the symmetric mutex since it is the most common use case.
+	 * Here the decision was to avoid defining another mutex to handle the push buffer for the rest
+	 * of the use cases.
+	 */
+	nrf_security_mutex_lock(cracen_mutex_symmetric);
```
vili-nordic, line 225: I don't think we need to lock the mutex here (normally called in sx symcrypt free functions?)

vili-nordic:
I pulled the changes into 
https://github.com/nrfconnect/sdk-nrf/pull/16039
since they are related

PR #16183, opened 2024-06-25
manifest: Update zephyr with extended gains support in SAADC shim

kl-cruz:
Update zephyr sha.


shanthanordic:
Discussed this @barsok  Not going for NCS 2.7.0

PR #16185, opened 2024-06-25
manifest: zephyr: Turn on NRFS for NRF54H20 DK

zycz:
Turn on NRFS globally so turning off
suspending MRAM for NRF54H20 DK will work.

Jira: NCSDK-27870

### DIFF
--- a/samples/suit/flash_companion/prj.conf
+++ b/samples/suit/flash_companion/prj.conf
@@ -20,6 +20,9 @@ CONFIG_FLASH=y
 # Disable power management
 CONFIG_PM=n
 
+# Disable NRFS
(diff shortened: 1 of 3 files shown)


tomchy (changes requested):
Please disable `NRFS` inside `flash_companion` image - it does not use radio core and should be optimized for memory (https://github.com/nrfconnect/sdk-nrf/blob/main/samples/suit/flash_companion/prj.conf).

tomchy (approved):

pdunaj (approved):

e-rk (approved):

shanthanordic (approved):

PR #16186, opened 2024-06-25
samples: coremark: use default logging mode

kapi-no:
Use the default NCS sample configuration in the CoreMark sample that enabled the minimal logging mode. Removed the flush logic in the sample as the minimal logging mode works synchronously.

Ref: NCSDK-26794


samples/benchmarks/coremark/README.rst
```diff
@@ -187,21 +187,21 @@ After programming the sample to your development kit, complete the following ste
...
       .. code-block:: console
 
-         *** Booting Zephyr OS build bf606fc00ec1  ***
-         [00:00:00.502,166] <inf> app: Coremark sample for nrf52dk/nrf52832. Call address: 00007fa1
-         [00:00:00.502,197] <inf> app: Press Push button switch 0 to start the test ...
-         [00:00:14.483,764] <inf> app: Push button switch 0 pressed!
-         [00:00:14.511,627] <inf> app: Coremark started!
-         [00:00:14.511,627] <inf> app: CPU FREQ: 64000000 Hz
-         [00:00:14.511,627] <inf> app: (threads: 1, data size: 2000; iterations: 2000)
+         [00:00:27.759,582] <inf> app: Coremark finished! Press Push button switch 0 to restart ...
```
zycz, line 190: Lets start from booting

samples/benchmarks/coremark/src/main.c
```diff
@@ -50,18 +49,10 @@ static const struct gpio_dt_spec status_led;
...
-		k_sleep(K_MSEC(100));
-	} else {
-		while (LOG_PROCESS()) {
-		}
-	}
-}
+/* Enfore synchronous logging as the sample doesn't flush logs. */
+BUILD_ASSERT(IS_ENABLED(CONFIG_LOG_MODE_MINIMAL) || IS_ENABLED(CONFIG_LOG_MODE_IMMEDIATE),
+	     "Logs should be processed synchronously to avoid negative impact on the "
+	     "benchamrk peformance");
```
zycz, line 55: typo peformance -> performance

samples/benchmarks/coremark/src/main.c
```diff
@@ -50,18 +49,10 @@ static const struct gpio_dt_spec status_led;
...
-			k_sleep(K_MSEC(100));
-		}
-		k_sleep(K_MSEC(100));
-	} else {
-		while (LOG_PROCESS()) {
-		}
-	}
-}
+/* Enfore synchronous logging as the sample doesn't flush logs. */
+BUILD_ASSERT(IS_ENABLED(CONFIG_LOG_MODE_MINIMAL) || IS_ENABLED(CONFIG_LOG_MODE_IMMEDIATE),
```
zycz, line 53: I wouldn't add this assert - we can have referred logging and use sample correctly - our current sample is prove of that.

samples/benchmarks/coremark/prj.conf
```diff
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
 
+CONFIG_NCS_SAMPLES_DEFAULTS=y
```
MarekPieta, line 6: This also implies `CONFIG_ASSERT` and HW stack protection. Maybe we could simply explicitly enable minimal logging to avoid enabling unwanted features accidentally?

samples/benchmarks/coremark/src/main.c
MarekPieta: Assert allows to simplify logic, so maybe it's not a bad idea to change it like that (there is no reason to use this sample with deferred logging I think)

MarekPieta (changes requested):

samples/benchmarks/coremark/prj.conf
kapi-no: Yes, it does. Why are those features unwanted?

Shouldn't we strive to be aligned with other samples?

samples/benchmarks/coremark/README.rst
kapi-no: right, this was leftover from the previous logging format

samples/benchmarks/coremark/src/main.c
kapi-no: yes, but I removed the logic for flushing logs so switching to the deferred mode would give you incorrect results. This sample is not the CoreMark module and it does not support the deferred logging after my change

samples/benchmarks/coremark/prj.conf
MarekPieta: While testing performance, generally assertions should not be enabled (assertions may affect performance measurements). HW stack protection may be enabled (should not impact performance).
Still, if a feature will be silently added to `CONFIG_NCS_SAMPLES_DEFAULTS` at some point, it may negatively affect our performance results, that's why I suggested to avoid using NCS samples defaults here

kapi-no: ASSERTs do not affect performance as the CoreMark benchmark does not use them. On the other hand, asserts give us some additional safety measures before we start the benchmark and after we finish executing it.

zycz (approved):

samples/benchmarks/coremark/prj.conf
MarekPieta: Zephyr may use asserts under the hood (e.g. when switching threads). Since we measure performance here, I would suggest to disable assertions because of that.

MarekPieta (approved):

kapi-no:
Rebased on top of the main branch. No file diff has changed.

kapi-no:
Added the changelog entry and update the boot banner in the Readme documentation.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -338,7 +338,9 @@ Wi-Fi samples
 Other samples
 -------------
 
-|no_changes_yet_note|
+* :ref:`coremark_sample` sample:
+
+  * Updated the logging mode to the minimal mode (:kconfig:option:`CONFIG_LOG_MODE_MINIMAL`) to minimize the memory footprint of this sample.
```
zycz, line 343: to minimize the memory footprint of this sample. ->  to minimize the memory footprint of this sample and guarantee that logging wouldn't interfere with running benchmark.

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -338,7 +338,9 @@ Wi-Fi samples
 Other samples
 -------------
```
MarekPieta, line 339: Consider aligning commit message (you no longer use the default NCS sample configuration)

MarekPieta (approved):

zycz (approved):

MarekPieta (approved):

samples/benchmarks/coremark/src/main.c
```diff
@@ -50,18 +49,10 @@ static const struct gpio_dt_spec status_led;
...
-		while (log_data_pending()) {
-			k_sleep(K_MSEC(100));
-		}
-		k_sleep(K_MSEC(100));
-	} else {
-		while (LOG_PROCESS()) {
-		}
-	}
-}
+/* Enfore synchronous logging as the sample doesn't flush logs. */
```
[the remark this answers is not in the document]
kapi-no: you are right, thanks for spotting this :)

annwoj (approved):
lgtm :)

kapi-no:
Ready to merge once the main branch is unblocked.

kapi-no:
Simple rebase as requested by @anangl. No file diff has changed

PR #16188, opened 2024-06-25
ci: docpublish: temporarily ignore old server publish failures

gmarull:
Until server is fixed, let's ignore failures during the upload process. Zoomin is still published, so it should not be critical.


gmarull:
note: CI won't test this, so it can be merged straight away.

gmarull:
Server should be fixed now, so closing...

PR #16199, opened 2024-06-25
cmake: multi_image: Fix handling of HWMv1 boards with revisions

57300:
Function `get_board_without_ns_suffix` cannot recognize the non-secure suffix if the board revision is placed after it, as is the case in the old hardware model. Example: `nrf9160dk_nrf9160_ns@0.14.0`

Before d7a19af4fe5b5d944befb66b6d9d1de3df67e275, board revisions were not passed to `get_board_without_ns_suffix`, because they are handled independently in the `ncs_file` call further down. Undo this change.

### DIFF
--- a/cmake/multi_image.cmake
+++ b/cmake/multi_image.cmake
@@ -234,11 +234,7 @@ function(add_child_image_from_source)
   elseif (NOT ACI_BOARD)
     # No BOARD is given as argument, this triggers automatic conversion of
     # *.ns board from parent image.
-    if(DEFINED BOARD_REVISION)
-      get_board_without_ns_suffix(${BOARD}@${BOARD_REVISION}${BOARD_QUALIFIERS} ACI_BOARD)
-    else()
-      get_board_without_ns_suffix(${BOARD}${BOARD_QUALIFIERS} ACI_BOARD)
-    endif()
+    get_board_without_ns_suffix(${BOARD}${BOARD_QUALIFIERS} ACI_BOARD)
   endif()
 
   if (NOT ACI_DOMAIN AND DOMAIN)


eivindj-nordic (approved):
Fixes my issue with HWMv1 and custom nRF9160 board. Thanks!

nordicjm (approved):

cmake/multi_image.cmake
```diff
@@ -234,10 +234,9 @@ function(add_child_image_from_source)
   elseif (NOT ACI_BOARD)
     # No BOARD is given as argument, this triggers automatic conversion of
     # *.ns board from parent image.
+    get_board_without_ns_suffix(${BOARD}${BOARD_QUALIFIERS} ACI_BOARD)
```
tejlmand, line 237: something does not sound correct.
In old HWM then the `BOARD` should not contain the revision at this point in time.

In ncs 2.6.x the revision was already stripped from the `BOARD` variable at this point in time.
This could be a regression in support for old HWM, so taking an extra look.

Blocking the PR in the meantime.

tejlmand (changes requested):
temp block to allow some more investigation

cmake/multi_image.cmake
tejlmand: No changes needed upstgream.

All that need to be changes is to remove all board revision specific handling, see other suggestion fix.

Board revision is passed independently to the child image here:
https://github.com/nrfconnect/sdk-nrf/blob/e6e4696ce35340b475b402dc84a2852d3a960b3a/cmake/multi_image.cmake#L338-L339

cmake/multi_image.cmake
```diff
@@ -234,10 +234,9 @@ function(add_child_image_from_source)
...
     if(DEFINED BOARD_REVISION)
-      get_board_without_ns_suffix(${BOARD}@${BOARD_REVISION}${BOARD_QUALIFIERS} ACI_BOARD)
-    else()
-      get_board_without_ns_suffix(${BOARD}${BOARD_QUALIFIERS} ACI_BOARD)
+      string(REGEX REPLACE "(^[^/]+)(/?.*$)" "\\1@${BOARD_REVISION}\\2" ACI_BOARD ${ACI_BOARD})
     endif()
```
tejlmand, line 240: Just remove all revision specific code:
```
    if(DEFINED BOARD_REVISION)
      ...
    endif()
```

```suggestion
```

tejlmand (changes requested):

cmake/multi_image.cmake
57300: I think it makes sense. Thanks for investigating.

cmake/multi_image.cmake
57300: Updated.

nordicjm (approved):

tejlmand (approved):

eivindj-nordic (approved):

PR #16205, opened 2024-06-26
applications/samples/test: Use DEFAULT_IMAGE in sysbuild

nordicjm:
Switches to use the DEFAULT_IMAGE variable when adding dependencies in sysbuild configuration to allow for applications to be copied out of tree

Fixes: NCSDK-28028


bencefr (approved):

tejlmand (changes requested):
duplicate of https://github.com/nrfconnect/sdk-nrf/pull/16087

PR #16214, opened 2024-06-26
samples: matter: Add basic SUIT support to Template sample

ArekBalysNordic:
Added base SUIT support to Template to allow the next development.

- Added a config overlay file to enable all SUIT dependencies and BLE SMP transfer.
- Added a DTS overlay file for SUIT purposes, enabling communication between the Application core and secure domain and reorganized partition layout.
- Aligned the dfu_over_smp Matter's implementation to be compatible with SUIT.


samples/matter/template/boards/nrf54h20_suit_overlay.conf
```diff
@@ -0,0 +1,33 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Enable SUIT services
+CONFIG_SUIT=y
```
kkasperczyk-no, line 8: I have few comments on that:
1. Why do we create nrf54h20_suit overlay that seems to enable only BT SMP? Can't SUIT be used with Matter OTA?
2. Why do we create such overlay at all, since we have `CONFIG_CHIP_DFU_OVER_BT_SMP` that is used by the clients to enable BT SMP? Wouldn't be better to modify the existing Kconfig to e.g. enable MCUBOOT or SUIT conditionally, depending on the platform used?

samples/matter/template/boards/nrf54h20_suit_overlay.conf
```diff
@@ -0,0 +1,33 @@
...
+CONFIG_BT_L2CAP_TX_MTU=498
+CONFIG_BT_BUF_ACL_RX_SIZE=502
+CONFIG_BT_BUF_ACL_TX_SIZE=502
+
+# Enable the Bluetooth mcumgr transport (unauthenticated).
+CONFIG_MCUMGR_TRANSPORT_BT=y
+CONFIG_MCUMGR_TRANSPORT_BT_AUTHEN=n
+CONFIG_MCUMGR_TRANSPORT_BT_CONN_PARAM_CONTROL=y
+CONFIG_MCUMGR_TRANSPORT_BT_REASSEMBLY=y
+CONFIG_MCUMGR_TRANSPORT_NETBUF_SIZE=2475
```
kkasperczyk-no, line 31: Why 2475? We have this buffer set to 1024 for other configurations and we use the same MTU size, so either this one does not have to be that big, or we should increase it for other configurations as well.

samples/matter/template/boards/nrf54h20_suit.overlay
```diff
@@ -0,0 +1,73 @@
+/* Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+/* Reorganize partitions for SUIT */
```
kkasperczyk-no, line 6: Correct me, if I'm wrong, but do we really need an overlay? As far as I understand SUIT is required for the DFU purposes, so we don't want to support configuration without it. I think we should just add it to the common nrf54h20 board overlay.

samples/matter/template/sample.yaml
```diff
@@ -76,3 +76,8 @@ tests:
       - nrf54h20dk/nrf54h20/cpuapp
     platform_allow: nrf54h20dk/nrf54h20/cpuapp
     tags: sysbuild
+  samples.matter.template.nrf54h20.suit:
```
kkasperczyk-no, line 79: I would suggest sample.matter.template.smp_dfu with the new board instead of creating new configuration.

samples/matter/template/boards/nrf54h20_suit_overlay.conf
ArekBalysNordic: 1. Because it is only a base for SUIT. We need to check whether SUIT works with dfu_target to run OTA.
2. OK, that's a good point to edit the existing kconfig. I forgot that it belongs to us, not to Zephyr. I will move the configs there.

samples/matter/template/boards/nrf54h20_suit_overlay.conf
ArekBalysNordic: I will investigate it. This value was proposed by the SUIT team. Maybe the lower value indeed is better.

samples/matter/template/boards/nrf54h20_suit.overlay
ArekBalysNordic: For now, we don't have implemented and tested DFU using external flash - it is planned in some weeks. Currently, we cannot perform dfu both for app + net core because we don't have enough space, so for development purposes, I would like to create an overlay that allows us to build the App with 98% occupancy but leave the original DTS for building the App with 68% occupancy. Once the external flash is supported it will be removed or changed.

samples/matter/template/sample.yaml
ArekBalysNordic: ok

ArekBalysNordic:
@kkasperczyk-no @markaj-nordic I've moved all SUIT manifests to the common directory because they will be the same for all our samples. If a user wants to use their own, we will describe this process in future documentation. Moreover, I've also created a new dfu_smp file dedicated to SUIT purposes and used the existing `CONFIG_CHIP_DFU_OVER_BT_SMP` config to enable all SUIT dependencies for nRF54H20.

samples/matter/template/boards/nrf54h20_suit_overlay.conf
ArekBalysNordic: They seem redundant to us. I removed them.

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
```diff
@@ -16,10 +17,12 @@
 
 #include <lib/support/logging/CHIPLogging.h>
 
+#ifndef CONFIG_SUIT
 #include <zephyr/dfu/mcuboot.h>
+#include <zephyr/mgmt/mcumgr/grp/img_mgmt/img_mgmt.h>
 #include <zephyr/mgmt/mcumgr/mgmt/callbacks.h>
 #include <zephyr/mgmt/mcumgr/mgmt/mgmt.h>
-#include <zephyr/mgmt/mcumgr/grp/img_mgmt/img_mgmt.h>
+#endif /* CONFIG_SUIT */
```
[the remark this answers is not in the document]
ArekBalysNordic: Done.

samples/matter/template/boards/nrf54h20dk_nrf54h20_cpuapp.conf
```diff
@@ -22,4 +22,4 @@ CONFIG_CHIP_FACTORY_DATA=n
 CONFIG_CHIP_FACTORY_DATA_WRITE_PROTECT=n
 
 # Set SUIT template paths
-CONFIG_SUIT_ENVELOPE_TEMPLATE="suit_templates/app_envelope.yaml.jinja2"
+CONFIG_SUIT_ENVELOPE_TEMPLATE="../common/suit_templates/app_envelope.yaml.jinja2"
```
kkasperczyk-no, line 25: I have a doubt about it. We rather try not to use relative paths based on the current location (../), but e.g. we use some macros in cmake (like ZEPHYR_NRF_MODULE_DIR). The problem with the current approach is that it will not work with the copying sample feature of the vs code, because someone may want to copy sample to other location than matter/samples/...

sysbuild/Kconfig.matter
```diff
@@ -39,4 +39,7 @@ config MATTER_OTA_IMAGE_FILE_NAME
 	help
 	  Provides the file name of the generated Matter OTA image in the build directory.
 
+config SUIT_ENVELOPE_ROOT_TEMPLATE
+	default "../common/suit_templates/root_with_binary_nordic_top.yaml.jinja2"
```
tejlmand, line 43: using a relative path from within a specific sample can make sense as you know the location of the sample itself.

But making the default relative to a sample's path when you don't know the sample being built looks wrong to me, and something which is bound to fail at some point.

Is is not possible to have the location defaulting to a path within a specific Zephyr module, such as ZEPHYR_NRF_MODULE_DIR ?

ArekBalysNordic: I checked that and indeed using the `ZEPHYR_NRF_MODULE_DIR` variable works, but if I use it within the sample directory, it points to the `<workdir>/ncs/nrf/samples/matter/template/` instead of `<workdir>/ncs/nrf`, but while using it within the `sysbuild/Kconfig.matter` file it works in the proper way and points to the `<workdir>/ncs/nrf` I have no idea why...

samples/matter/common/dts/nrf54h20/nrf54h20_cpuapp_memory_map.dtsi
```diff
@@ -0,0 +1,69 @@
...
+			reg = < 0x174000 DT_SIZE_K(4) >;
+		};
+
+		/* DFU partition to store SUIT manifests and Nordic Firmware update */
+		dfu_partition: partition@175000 {
+			reg = < 0x175000 DT_SIZE_K(464) >;
+		};
+
+	};
+};
```
markaj-nordic, line 69: Missing an empty line here and there.

samples/matter/common/src/dfu/smp/suit_dfu_over_smp.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+using namespace ::chip;
+using namespace ::chip::DeviceLayer;
+
+namespace Nrf
+{
+
+DFUOverSMP DFUOverSMP::sDFUOverSMP;
+
+void DFUOverSMP::Init()
```
markaj-nordic, line 26: I'd consider abstracting the common code to the base class with default implementation of the whole public interface and then only adding different bits in the specific SUIT DFUOverSMP version by overwriting.

samples/matter/common/dts/nrf54h20/nrf54h20_cpuapp_memory_map.dtsi
ArekBalysNordic: fixed.

samples/matter/template/sample.yaml
```diff
@@ -20,15 +20,20 @@ tests:
...
       CONFIG_NCS_SAMPLE_MATTER_OPERATIONAL_KEYS_MIGRATION_TO_ITS=y
-      SB_CONFIG_BOOTLOADER_MCUBOOT=y SB_CONFIG_PARTITION_MANAGER=y
     integration_platforms:
       - nrf52840dk/nrf52840
       - nrf5340dk/nrf5340/cpuapp
       - nrf7002dk/nrf5340/cpuapp
       - nrf54l15pdk/nrf54l15/cpuapp
+      - nrf54h20dk/nrf54h20/cpuapp
     platform_allow: nrf52840dk/nrf52840 nrf5340dk/nrf5340/cpuapp nrf7002dk/nrf5340/cpuapp
+<<<<<<< HEAD
```
markaj-nordic, line 30: Some unresolved conflicts sneaked in.

markaj-nordic (approved):

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
```diff
@@ -6,8 +6,9 @@
 
 #include "dfu_over_smp.h"
 
-#if !defined(CONFIG_MCUMGR_TRANSPORT_BT) || !defined(CONFIG_MCUMGR_GRP_IMG) || !defined(CONFIG_MCUMGR_GRP_OS)
-#error "DFUOverSMP requires MCUMGR module configs enabled"
+#if !defined(CONFIG_MCUMGR_TRANSPORT_BT) ||                                                                            \
+	(!defined(CONFIG_SUIT) && (!defined(CONFIG_MCUMGR_GRP_IMG) || !defined(CONFIG_MCUMGR_GRP_OS)))
+#error "DFUOverSMP requires MCUMGR module configs enabled and SUIT for the nRF54H20 platform"
```
nordicjm, line 11: *MCUmgr, and shouldn't it be DFU over SMP?

samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf
```diff
@@ -22,4 +22,4 @@ CONFIG_CHIP_FACTORY_DATA=n
 CONFIG_CHIP_FACTORY_DATA_WRITE_PROTECT=n
 
 # Set SUIT template paths
-CONFIG_SUIT_ENVELOPE_TEMPLATE="suit_templates/app_envelope.yaml.jinja2"
+CONFIG_SUIT_ENVELOPE_TEMPLATE="${ZEPHYR_NRF_MODULE_DIR}../common/suit_templates/app_envelope.yaml.jinja2"
```
nordicjm, line 25: bit confused by this, if this is reading in a file, why is there a "common" folder in the same location as `nrf`, if this is writing a file then why is it being written to here?

sysbuild/Kconfig.matter
```diff
@@ -39,4 +39,7 @@ config MATTER_OTA_IMAGE_FILE_NAME
 	help
 	  Provides the file name of the generated Matter OTA image in the build directory.
 
+config SUIT_ENVELOPE_ROOT_TEMPLATE
+	default "${ZEPHYR_NRF_MODULE_DIR}/samples/matter/common/suit_templates/root_with_binary_nordic_top.yaml.jinja2"
```
nordicjm, line 43: based on this I'm thinking your other paths are wrong

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
```diff
@@ -16,43 +17,42 @@
...
 {
-enum mgmt_cb_return UploadConfirmHandler(uint32_t event,
-					 enum mgmt_cb_return prev_status,
-					 int32_t *rc, uint16_t *group,
-					 bool *abort_more, void *data,
-					 size_t data_size)
+enum mgmt_cb_return UploadConfirmHandler(uint32_t event, enum mgmt_cb_return prev_status, int32_t *rc, uint16_t *group,
+					 bool *abort_more, void *data, size_t data_size)
 {
+/* Currently img_mgmt hooks are not supported by SUIT */
```
kkasperczyk-no, line 42: Are we able to track DFU progress some other way? If not, could we somehow inform the user that DFU started and he should wait some time for it to end? I'm afraid that customer can be worried that application stuck, as progress will not be shown.

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
```diff
@@ -106,6 +107,7 @@ void DFUOverSMP::Init()
 
 void DFUOverSMP::ConfirmNewImage()
```
kkasperczyk-no, line 108: Just to my knowledge is any form of confirmation required in the SUIT?

sysbuild/Kconfig.matter
kkasperczyk-no: Yeah, I think this should be the same in both places.

ArekBalysNordic: As I wrote here: https://github.com/nrfconnect/sdk-nrf/pull/16214#discussion_r1666459774 there is a different approach to paths. I need to use different paths depending on the config location. @nordicjm maybe you could explain to me why I cannot use the same paths in `sysbuild/Kconfig.matter` and in `samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf` ?

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
ArekBalysNordic: No, SUIT can verify new images using its semantics and digests, but we can use it to confirm it additionally in the application if needed, or if some users want to confirm something in their application.

sysbuild/Kconfig.matter
ArekBalysNordic: I wanted to use the same paths here and there, but I couldn't :/

samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf
ArekBalysNordic: This is a base SUIT manifest that we suggest to customers, but they can have their own manifests and use the `CONFIG_SUIT_ENVELOPE_TEMPLATE` config to set the new path. Do you recommend setting it by default within the sample's Kconfig file?

samples/matter/template/sample.yaml
ArekBalysNordic: yeah thanks, seems that I forgot to save the file.

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
ArekBalysNordic: Yes, I will change it.

kkasperczyk-no, line 11: Could we split this into two separate error messages to be more informative? E.g.
```suggestion
#if !defined(CONFIG_MCUMGR_TRANSPORT_BT)
#error "DFUOverSMP requires MCUMGR module configs enabled"
#endif

#if (!defined(CONFIG_SUIT) && (!defined(CONFIG_MCUMGR_GRP_IMG) || !defined(CONFIG_MCUMGR_GRP_OS)))
#error  "DFUOverSMP requires SUIT configs enabled for the nRF54H20 platform"
#endif
```

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
kkasperczyk-no: Not necessarily, it's ok, as long as we track this.

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
ArekBalysNordic: Sure.

ArekBalysNordic: done.

samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf
ArekBalysNordic: To explain it more, actually, we will have one "base" SUIT manifest template that is valid for all Matter samples, and it is located in the `nrf/samples/matter/common` directory. We need to set the `CONFIG_SUIT_ENVELOPE_TEMPLATE` config anyway to provide a template manifest path so we decided to provide a path to the common directory instead of having the copy of the manifests in each app separately. Users can write their own manifest templates, and then they can easily set the `CONFIG_SUIT_ENVELOPE_TEMPLATE` config to other paths and use them. They can both modify `samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf` file or provide the config as CMAKE/NINJA build argument as well.

kkasperczyk-no (approved):

sysbuild/Kconfig.matter
ArekBalysNordic: @nordicjm Could you help me and explain why I need to use different paths?

ArekBalysNordic:
@nordicjm Do you have any more input here? If not I would like to merge this PR as soon as possible and if there is something wrong, we can address it as a follow-up because without this PR on main, we are blocked from doing the next tasks.

samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf
nordicjm: I would really appreciate it if you didn't rush to get things merged that clearly are wrong. To highlight:
```
message(WARNING "path: ${ZEPHYR_NRF_MODULE_DIR}../common/suit_templates/app_envelope.yaml.jinja2")
```
=
`  path: /tmp/bb/nrf../common/suit_templates/app_envelope.yaml.jinja2`
Notice something odd about the path? Please fix

ArekBalysNordic: 3 our tasks were blocked because we've been waiting to merge this PR for two weeks... I explained why is wrong and I hadn't got any input from you for two days, so to unlock the other tasks, we merged this one, and as I wrote below, I can do a follow-up if this is wrong. I didn't provide any regression, because it works for me, so I still don't understand why your message points to the wrong path, whereas SUIT generation works properly...

PR #16215, opened 2024-06-26
net: lib: nrf_cloud: Fix SDK version string

simensrostad:
Set SDK version string reported to the cloud to the correct value.

### DIFF
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
@@ -23,9 +23,13 @@
 #include <zephyr/sys/util.h>
 #include <zephyr/logging/log.h>
 #include <modem/modem_info.h>
+#include <ncs_version.h>
+#include <ncs_commit.h>
 #include "cJSON_os.h"
 
 LOG_MODULE_REGISTER(nrf_cloud_codec_internal, CONFIG_NRF_CLOUD_LOG_LEVEL);
+
+#define SDK_VERSION NCS_VERSION_STRING "-" NCS_COMMIT_STRING
 
 /** @brief How the control section is handled when either a trimmed shadow
  *  or a delta shadow is received.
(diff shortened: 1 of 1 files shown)


subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -1391,7 +1395,7 @@ static int encode_modem_info_device(struct device_param *device, cJSON *json_obj
 		return -ENOMEM;
 	}
 
-	if (json_add_str_cs(json_obj, "sdkVer", device->app_version)) {
```
jayteemo, line 1394: this "app_version" comes from modem_info lib.
it holds the SDK version... so perhaps it should be renamed in modem_info?

example:
image

do you see different behavior?

plskeggs:
@simensrostad is there a ticket associated with this?

What observations led you to want to make this change?

As Justin shared in his comment, the data in the shadow looks correct already to me.

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
simensrostad: Sorry for the late reply. In the [OOB firmware](https://github.com/hello-nrfcloud/firmware), the sdkVer appears as APP_VERSION. This might be because it's out-of-tree, where the CMAKE variable isn't available, but I'm not certain.

I didn't see why a modem library should be used to retrieve the SDK version, so I found it more appropriate to fix the issue by referencing the actual NCS version and commit strings, which should always be available.

The modem info library parameter refers to an APP_VERSION, which isn't really an SDK version at all. This is another reason not to use it.

Overall, I don't see why the modem info library should provide either the app or SDK version, so I would suggest removing the parameter in a later effort.

simensrostad:
> @simensrostad is there a ticket associated with this?
> 
> What observations led you to want to make this change?
> 
> As Justin shared in his comment, the data in the shadow looks correct already to me.

I added a reply to Justin's comment, this is the original ticket:
https://github.com/hello-nrfcloud/firmware/issues/198

subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
jayteemo: created `IRIS-9301`

jayteemo (approved):

glarsennordic (approved):

PR #16223, opened 2024-06-27
samples: ipc: ipc_service: disable hw_flow_control for cpurad nRF54H20

maje-emb:
hw-flow-control disabled to prevent sample blocking during UART data transmission when the terminal is not connected.


grochu (approved):

doki-nordic (approved):

dchat-nordic (approved):

dchat-nordic:
As a comment - does cpuapp need to have disabled flow control also?

nordic-piks:
Please keep in mind that this is just a workaround.
It would be the best to keep flow-control and fix hanging core on logging when there no receiving side.
Beside that, similar configuration with app and ppr works fine (ppr has flow control enabled).

PR #16230, opened 2024-06-27
suit: Adress validation - missing parts

ahasztag:
This commit adds missing parts of the address validation in SUIT:

- validation of update regions which are passed to SDFW during update
- validation of execute permissions on components which are being booted


adsz-nordic (approved):

subsys/suit/orchestrator/src/suit_orchestrator_sdfw.c
```diff
@@ -86,6 +88,44 @@ static void leave_emergency_recovery(void)
...
+	 * 2) address accessible by application core
+	 * 3) address accessible by radio core
+	 *  
+	 * Other cases can lead to security issues, leading to potential leaks
+	 * of protected data.
+	 */
+	
+	if (suit_memory_global_address_range_is_in_external_memory((uintptr_t) address, size)) {
+		return true;
+	}
```
e-rk, line 104: The `update_region[0]` must be in MRAM. I would suggest  that this check is done only for the `update_regions[i > 0]`

e-rk (changes requested):

ahasztag:
Changing to draft, the way it is now I cannot make the tests compile due to undefined arbiter module on sdk-nrf...

PR #16236, opened 2024-06-27
samples: add board overlay and conf file for nrf5340dk in HWMv1 design

tejlmand:
nRF5340dk is also available in HWMv1 description format for testing purposes.
Create board overlay and conf files targeting the nRF5340dk board in hardware model v1 description format.

Jira: NCSDK-28165


PerMac:
It works

nordicjm:
Why can't the twister overlay be used for this for another repo?

nordicjm:
> > Why can't the twister overlay be used for this for another repo?
> 
> what do you mean exactly?

Where you have the files in another repo and twister overlays them for the test

tejlmand:
> Where you have the files in another repo and twister overlays them for the test

they can, but the main point here is that they are able to run `twister` as-is for the sample.
If moving those files to other repo, then they need to apply `EXTRA_CONF_FILE` + `EXTRA_DTC_OVERLAY_FILE`, or use `APPLICATIN_CONFIG_DIR` (in which case also `prj.conf` must be duplicated (and twister call must be adjusted).

Third option is to duplicate whole sample to other repo, and run from there. (duplicates all code, but twister can run as-is).

All options are possible, question just is which solution to choose when considering maintenance vs. what we want to achieve.

nordicjm:
Not that, there is a twister option to use an alternative root

PerMac:
this --alt-config-root only allows to modify yaml file. We can of course have a version of yaml where overlays are listed explicitly, but for twister the build dir would be the sdk-nrf one so path will require adapting. This option is not optimal to use. IMO, we should consider copying all the tests/samples we want to test to the test repo and align them then. We already had to copy nrf_desktop and asset_tracker from nt 2.6 due to how different they are now in compare to 2.6. With copied versions we at least will have a scenario as at a client side: out-of-tree board and out-of-tree configurations. If we uses sources from nrf main than at one point (as it already happened some configs) we might need extra effort in aligning our setup, since the app from main would no longer work with v1

shanthanordic:
Need approvals from @PerMac @nvlsianpu

PerMac (changes requested):
I don't think this fix is acceptable. It can cause some confusion since this v1 board is not available in NCS, hence their names shouldn't be referred in NCS. As in comments, IMO we need to copy and align samples we want to verify with hwmv1 to the dedicated test repo

nvlsianpu:
I do understand that we want to make sure that HWMv1 is working by testing it.
Are we want to put patch to NCS?

shanthanordic:
Discussed with @PerMac  and @nvlsianpu. Cleared milestone, 
Reason is customer will have his own boards and his own application. 
Also we do not want to pollute the tree with overlay files for V1

tejlmand:
Closing.

PR has served its purpose by providing a solution to NCSDK-28165.
Whether the request to ensure NCS supports boards HWMv1 format by testing those is done in-tree (this PR) or by having custom test sample (outside this PR) has been discussed and agreed, and thus no need for this PR.

PR #16242, opened 2024-06-27
net: lib: nrf_cloud_coap: Allow null net_info

plskeggs:
For an AGNSS request, allow the net_info request parameter to be null.

Jira: NCSDK-28192

### DIFF
--- a/subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
+++ b/subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
@@ -412,10 +412,22 @@ int coap_codec_agnss_encode(struct nrf_cloud_rest_agnss_request const *const req
 	int cnt;
 
 	memset(&input, 0, sizeof(struct agnss_req));
-	input.agnss_req_eci = request->net_info->current_cell.id;
-	input.agnss_req_mcc = request->net_info->current_cell.mcc;
-	input.agnss_req_mnc = request->net_info->current_cell.mnc;
-	input.agnss_req_tac = request->net_info->current_cell.tac;
+
+	if (request->net_info != NULL) {
+		input.agnss_req_eci = request->net_info->current_cell.id;
+		input.agnss_req_mcc = request->net_info->current_cell.mcc;
+		input.agnss_req_mnc = request->net_info->current_cell.mnc;
+		input.agnss_req_tac = request->net_info->current_cell.tac;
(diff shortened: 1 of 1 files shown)


balaji-nordic (approved):

jayteemo (approved):

subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
```diff
@@ -412,10 +412,20 @@ int coap_codec_agnss_encode(struct nrf_cloud_rest_agnss_request const *const req
...
+		input.agnss_req_mcc = request->net_info->current_cell.mcc;
+		input.agnss_req_mnc = request->net_info->current_cell.mnc;
+		input.agnss_req_tac = request->net_info->current_cell.tac;
+		if (request->net_info->current_cell.rsrp != NRF_CLOUD_LOCATION_CELL_OMIT_RSRP) {
+			input.agnss_req_rsrp.agnss_req_rsrp = request->net_info->current_cell.rsrp;
+			input.agnss_req_rsrp_present = true;
+		}
+	} else {
+		LOG_DBG("No net_info provided.");
+	}
```
tokangas, line 427: This leads to `eci`, `mcc`, `mnc` and `tac` fields to be encoded as zeroes. This is problematic, because `0` is a valid cell ID, which could theoretically be in use somewhere.

As far as I understand the CDDL definition, the fields are currently mandatory, so they have to be given some values. Can we change the CDDL definition and make the fields optional? How does the cloud side implementation work?

If we can not make the values optional, value `0xFFFFFFFF` could be used to indicate that the cell ID is not valid.

shanthanordic:
Too late, pushed out NCS 2.7.0

subsys/net/lib/nrf_cloud/coap/src/coap_codec.c
plskeggs: This was seemingly an emergency bug fix at the last second for NCS 2.7.0, so I did the simplest and smallest change.

Since it was decided that this could not be included now in NCS 2.7.0, I will go ahead and modify the CDDL and regenerate the codec.

plskeggs: Actually, changing the CDDL will need to be done in coordination with the backend team, so I will made that a separate PR.

Instead, I'll set `eci` and `tac` to 0xFFFFFFFF if there is no network information available.

glarsennordic (approved):

PR #16246, opened 2024-06-28
net: lib: nrf_cloud: Add RSRP reporting

balaji-nordic:
Add RSRP (Reference Signal Received Power) to the modem information reported to the cloud.

### DIFF
--- a/subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
+++ b/subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
@@ -1307,6 +1307,11 @@ static int encode_modem_info_network(struct network_param *network, cJSON *json_
 		return ret;
 	}
 
+	ret = add_modem_info_data(&network->rsrp, json_obj);
+	if (ret < 0) {
+		return ret;
+	}
+
 	ret = modem_info_name_get(network->cellid_hex.type, data_name);
 	if (ret < 0) {
 		return ret;


subsys/net/lib/nrf_cloud/src/nrf_cloud_codec_internal.c
```diff
@@ -1307,6 +1307,11 @@ static int encode_modem_info_network(struct network_param *network, cJSON *json_
 		return ret;
 	}
 
+	ret = add_modem_info_data(&network->rsrp, json_obj);
```
jayteemo, line 1310: should this be reported in dBm?

if so, add something like this in `add_modem_info_data`

```
       } else {
		uint16_t val = param->value;

		if (param->type == MODEM_INFO_RSRP) {

			if (val == 255) {
				/* If RSRP is not valid, add a null */
				if (cJSON_AddNullToObjectCS(json_obj, data_name) == NULL) {
					return -ENOMEM;
				}
				return 0;
			}

			val = RSRP_IDX_TO_DBM(val);
		}

		if (cJSON_AddNumberToObject(json_obj, data_name, val) == NULL) {
			return -ENOMEM;
		}
	}
```

i believe RSRP is only valid if CESQ notifications have been enabled.

balaji-nordic: > i believe RSRP is only valid if CESQ notifications have been enabled.

Ah I did not know know. Thanks. I think I will need to solve it differently then. Using `modem_info_rsrp_register()` and then report the RSRP received in the callback function.

PR #16274, opened 2024-07-01
suit: Added address validator module and used it in the orchestrator

ahasztag:


subsys/suit/validator/Kconfig
```diff
@@ -0,0 +1,26 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
```
tomchy, line 2: 2024

subsys/suit/validator/include/suit_validator.h
```diff
@@ -0,0 +1,20 @@
...
+#define SUIT_VALIDATOR_H__
+
+#include <stdint.h>
+#include <stddef.h>
+#include <suit_plat_err.h>
+
+suit_plat_err_t suit_validator_validate_update_candidate_location(const uint8_t *address,
+								  size_t size);
+
+suit_plat_err_t suit_validator_validate_dfu_partition_location(const uint8_t *address,
```
tomchy, line 17: Missing docs?

subsys/suit/validator/include/suit_validator.h
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef SUIT_VALIDATOR_H__
+#define SUIT_VALIDATOR_H__
+
+#include <stdint.h>
+#include <stddef.h>
+#include <suit_plat_err.h>
+
```
tomchy, line 13: The `#ifdef __cplusplus` is not strictly necessary, but it i better to be consistent and add it here too :slightly_smiling_face:

tests/subsys/suit/common/validator_test/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2023 Nordic Semiconductor ASA
```
tomchy, line 1: 2024?

tomchy (changes requested):

subsys/suit/validator/include/suit_validator.h
ahasztag: Fixed

subsys/suit/validator/Kconfig
ahasztag: Fixed

tomchy (approved):

robertstypa (approved):

subsys/suit/validator/src/suit_validator_nrf54h20.c
```diff
@@ -0,0 +1,51 @@
...
+				.size = size,
+			},
+	};
+
+	if (arbiter_mem_access_check(&mem_params) == ARBITER_STATUS_OK) {
+		return SUIT_PLAT_SUCCESS;
+	}
+
+	mem_params.access.owner = NRF_OWNER_RADIOCORE;
+
```
SylwesterKonczyk, line 29: I wonder about the use case where candidate is stored in memory belonging to radio core... let's discuss it

subsys/suit/validator/include/suit_validator.h
```diff
@@ -0,0 +1,48 @@
...
+extern "C" {
+#endif
+
+/**
+ * @brief Validate the location of the update candidate.
+ *
+ * @retval SUIT_PLAT_SUCCESS     The location is valid.
+ * @retval SUIT_PLAT_ERR_ACCESS  The location is forbidden.
+ */
+suit_plat_err_t suit_validator_validate_update_candidate_location(const uint8_t *address,
```
SylwesterKonczyk, line 32: Looking at function names: 

suit_validator_validate_update_candidate_location, suit_validator_validate_dfu_partition_location - I am to some extent confused. 

We have two kind of regions to distinguish:
1. an area holding a 'primary' envelope (one containing candidate manifest(s)) - that one must be placed in MRAM
2. area(s) with dfu cache entries - those may be placed both on MRAM and external flash (actually- it can be any kind of non-volatile memory represented as 'external flash')

So what is a mapping between discussed functions and regions listed above?

Perhaps we could consider bettter-describing function names?

ahasztag: My mistake, I have skipped the word "cache" by accident. Changed the name of the second function to `suit_validator_validate_cache_pool_location` now it should be more clear and consistent.

PR #16276, opened 2024-07-01
samples: matter: optimize the factory reset time

markaj-nordic:
Currently the Matter core uses generic implementation of GroupDataProvider that performs multiple operations on persistent storage (including erasing of the related keys) even though the KVS is erased at the end and the device is rebooted. Implemented custom GroupDataProvider that can be configured at runtime to avoid redundant and time-consuming tasks.

It decreases the factory reset time on nRF52 by ~40%.

### DIFF
--- a/samples/matter/common/src/app/group_data_provider.h
+++ b/samples/matter/common/src/app/group_data_provider.h
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#pragma once
+
+#include <credentials/GroupDataProviderImpl.h>
+
+namespace Nrf::Matter
+{
+class GroupDataProviderImpl : public chip::Credentials::GroupDataProviderImpl {
+public:
+	/**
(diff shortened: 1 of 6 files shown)


ArekBalysNordic (approved):
LGTM

samples/matter/common/src/app/group_data_provider.h
```diff
@@ -0,0 +1,62 @@
...
+	 *
+	 * This function informs the GroupDataProviderImpl that the factory
+	 * reset will happen, therefore we can skip the time-consuming default
+	 * RemoveFabric's operations that heavily use persistent storage.
+	 *
+	 * It should be called before scheduling the factory reset with
+	 * chip::Server::ScheduleFactoryReset().
+	 *
+	 */
+	void WillBeFactoryReseted() { mClearStorage = false; }
```
kkasperczyk-no, line 35: How about?
```suggestion
	void WillBeFactoryReset() { mClearStorage = false; }
```

kkasperczyk-no (approved):

Damian-Nordic (approved):

samples/matter/common/src/app/group_data_provider.h
markaj-nordic: I had doubts about it :)

PR #16277, opened 2024-07-01
NCS 2.7.0 rebase

anangl:
Update zephyr revision with a snapshot tag. There is no need for a new tag in sdk-mcuboot as that -rc3 one will preserve the history. Other OSS repositories do not need to be rebased.

Update zephyr and mcuboot revisions after they were rebased.


anangl:
The first commit extracted to a separate PR to verify the snapshot tag: #16291.

carlescufi (approved):

PR #16279, opened 2024-07-01
services: hids: Fix possible out-of-bound access

KAGA164:
Fixes coverity issue 122949

### DIFF
--- a/subsys/bluetooth/services/hids.c
+++ b/subsys/bluetooth/services/hids.c
@@ -708,7 +708,7 @@ static ssize_t hids_ctrl_point_write(struct bt_conn *conn,
 		return BT_GATT_ERR(BT_ATT_ERR_INVALID_OFFSET);
 	}
 
-	if (len > sizeof(uint8_t)) {
+	if (len != sizeof(uint8_t)) {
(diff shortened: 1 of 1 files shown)


alwa-nordic (changes requested):
I think you also have to fix the `if (len > sizeof(uint8_t)) {` so it is `!=` instead.

https://github.com/nrfconnect/sdk-nrf/blob/d4de8ecbc9599abcbdf4393f3350dadc280a7b84/subsys/bluetooth/services/hids.c#L711

dchat-nordic (changes requested):
I agree with @alwa-nordic

KAGA164:
> I think you also have to fix the `if (len > sizeof(uint8_t)) {` so it is `if (len != sizeof(uint8_t)) {` instead.
> 
> https://github.com/nrfconnect/sdk-nrf/blob/d4de8ecbc9599abcbdf4393f3350dadc280a7b84/subsys/bluetooth/services/hids.c#L711

@alwa-nordic @dchat-nordic  Done

alwa-nordic (approved):

dchat-nordic (approved):

maje-emb (approved):

grochu (approved):

PR #16284, opened 2024-07-01
lib: nrf_modem_lib: lte_net_if: Enable modem events

gregersrygg:
Enable modem events when using Zephyr Connection Manager. When the events are enabled, a warning will be printed when the modem detects a reset loop.

### DIFF
--- a/lib/nrf_modem_lib/lte_net_if/lte_net_if.c
+++ b/lib/nrf_modem_lib/lte_net_if/lte_net_if.c
@@ -415,6 +415,11 @@ static int lte_net_if_connect(struct conn_mgr_conn_binding *const if_conn)
 		return ret;
 	}
 
+	ret = lte_lc_modem_events_enable();
+	if (ret) {
+		LOG_WRN("lte_lc_modem_events_enable, error: %d", ret);
+	}
+
 	connection_timeout_schedule();
 
 	return 0;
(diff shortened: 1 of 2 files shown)


lib/nrf_modem_lib/lte_net_if/lte_net_if.c
```diff
@@ -413,6 +413,11 @@ static int lte_net_if_connect(struct conn_mgr_conn_binding *const if_conn)
 		return ret;
 	}
 
+	ret = lte_lc_modem_events_enable();
+	if (ret) {
+		LOG_WRN("lte_lc_modem_events_enable, error: %d", ret);
```
balaji-nordic, line 418: Add a 

`return ret;`

balaji-nordic (changes requested):

lib/nrf_modem_lib/lte_net_if/lte_net_if.c
gregersrygg: Do we want to return en error just because we're not getting some events from the modem? These events are only used for printing warnings. So I considered a failure to enable them isn't critical enough. Can you please explain why you think this is a critical error?

lemrey: I think it's fine to continue rather than return; it's not a critical error (in fact we print a warning).

balaji-nordic: I see. I just expect any LTE_LC_* API to succeed at the connection establishment stage. If it fails and we ignore it, then something else is bound to fail at a later point. But since two of you consider this as non-critical failure, I am willing to leave it as is.

balaji-nordic (approved):

lemrey (approved):

PR #16285, opened 2024-07-01
doc: migration: hwmv2: fix board migration invocation

gmarull:
- Remove $
- Fix board root
- Add missing group (-g) parameter.


carlescufi (changes requested):
Blocking until https://github.com/nrfconnect/sdk-nrf/pull/16286 is merged so that we get it used here

greg-fer:
Now cherry-picked into #16286 . I suggest we close this PR, @gmarull .

PR #16289, opened 2024-07-02
samples: matter: Increased stacks and settings sizes for nRF54L15

kkasperczyk-no:
Increased CHIP and MPSL stack sizes, and settings size for nRF54L15, as it was proven in tests that usage is bigger for this platform.


markaj-nordic (approved):

samples/matter/light_bulb/boards/nrf54l15pdk_nrf54l15_cpuapp.conf
```diff
@@ -22,3 +22,7 @@ CONFIG_CHIP_FACTORY_DATA_WRITE_PROTECT=n
 # reset + 1. Hence, the reboot time increases more and more.
 # To avoid it enable tickles kernel for mcuboot.
 CONFIG_TICKLESS_KERNEL=y
+
+# nRF54L15 requires bigger stack sizes than nRF52/nRF53 families
```
ArekBalysNordic, line 26: Should we explain that it is caused by the crypto backend which uses more RAM than backends for NRF52/NRF53?

ArekBalysNordic (approved):

PR #16296, opened 2024-07-02
applications: nrf5340_audio: Add a deinitialise function

gWacey:
The data FIFO module did not have a deinitialise function, so added one, with a twister test.

### DIFF
--- a/include/data_fifo.h
+++ b/include/data_fifo.h
@@ -43,14 +43,14 @@ struct data_fifo {
 
 #define DATA_FIFO_DEFINE(name, elements_max_in, block_size_max_in)                                 \
 	char __aligned(WB_UP(1))                                                                   \
-		_msgq_buffer_##name[(elements_max_in) * sizeof(struct data_fifo_msgq)] = { 0 };    \
+		_msgq_buffer_##name[(elements_max_in) * sizeof(struct data_fifo_msgq)] = {0};      \
 	char __aligned(WB_UP(1))                                                                   \
-		_slab_buffer_##name[(elements_max_in) * (block_size_max_in)] = { 0 };              \
-	struct data_fifo name = { .msgq_buffer = _msgq_buffer_##name,                              \
-				  .slab_buffer = _slab_buffer_##name,                              \
-				  .block_size_max = block_size_max_in,                             \
-				  .elements_max = elements_max_in,                                 \
-				  .initialized = false }
+		_slab_buffer_##name[(elements_max_in) * (block_size_max_in)] = {0};                \
+	struct data_fifo name = {.msgq_buffer = _msgq_buffer_##name,                               \
+				 .slab_buffer = _slab_buffer_##name,                               \
+				 .block_size_max = block_size_max_in,                              \
+				 .elements_max = elements_max_in,                                  \
+				 .initialized = false}
 
 /**
  * @brief Get pointer to the first vacant block in slab.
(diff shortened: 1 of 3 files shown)


lib/data_fifo/data_fifo.c
```diff
@@ -167,6 +167,22 @@ int data_fifo_empty(struct data_fifo *data_fifo)
 	return 0;
 }
 
+int data_fifo_deinit(struct data_fifo *data_fifo)
+{
+	__ASSERT_NO_MSG(data_fifo != NULL);
```
koffes, line 172: Recommend using ASSERT with message.

include/data_fifo.h
```diff
@@ -150,6 +150,18 @@ int data_fifo_num_used_get(struct data_fifo *data_fifo, uint32_t *alloced_num,
  */
 int data_fifo_empty(struct data_fifo *data_fifo);
 
+/**
+ * @brief Deinitialise the data_fifo.
+ *
+ * @note The fifo is emptied first, so it is the users responsibility to release any data items it
```
koffes, line 156: What happens if the user does not release all the data items?

tests/lib/data_fifo/src/main.c
```diff
@@ -31,6 +31,21 @@ static void internal_test_remaining_elements(struct data_fifo *data_fifo, uint32
 		      num_locked, line);
 }
 
+ZTEST(suite_data_fifo, test_data_fifo_deinit_ok)
```
koffes, line 34: Nice.

Suggest we add a bit more substantial test where we init a fifo, fill it, deinit it, init it again and fill it.

koffes (changes requested):

include/data_fifo.h
gWacey: What will happen to any buffers that are carried in the message?
We get a message and load it with a pointer to some data (on a slab for example). When all the messages are released what happens to this item and the slab it is on? It may be lost to the system!

tests/lib/data_fifo/src/main.c
gWacey: Ok will add something.

lib/data_fifo/data_fifo.c
gWacey: On this, It is good to test like this, but if there is an issue in the release that would have exercised this assert but it has been optimised out, then the codes action would be undetermined and potentially difficult to debug. Would be more code but better in my option to explicitly test for NULL. Actually it shouldn't be more than a test zero and branch (with branch prediction this will be even better). An error is already returned and tested for, so the code could gracefully fail, rather than go off somewhere into memory!

andvib (approved):

include/data_fifo.h
```diff
@@ -150,6 +150,18 @@ int data_fifo_num_used_get(struct data_fifo *data_fifo, uint32_t *alloced_num,
...
+ *       has queued. The internal slab and message buffer are not released.
+ *
+ * @param data_fifo Pointer to the data_fifo structure.
+ *
+ * @retval 0 if success, error otherwise.
+ */
+int data_fifo_deinit(struct data_fifo *data_fifo);
```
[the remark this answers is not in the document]
gWacey: Will correct.

include/data_fifo.h
```diff
@@ -150,6 +150,18 @@ int data_fifo_num_used_get(struct data_fifo *data_fifo, uint32_t *alloced_num,
  */
 int data_fifo_empty(struct data_fifo *data_fifo);
 
+/**
+ * @brief Uninitialise the data_fifo.
```
greg-fer, line 154:
```suggestion
 * @brief Deinitialize data_fifo.
```
We follow American English (`z`).

include/data_fifo.h
```diff
@@ -150,6 +150,18 @@ int data_fifo_num_used_get(struct data_fifo *data_fifo, uint32_t *alloced_num,
...
+ *
+ * @note The fifo is emptied first, so it is the users responsibility to release any data items it
```
greg-fer, line 156:
```suggestion
 * @note data_fifo is emptied first, so it is the user's responsibility to release any data items it
```

greg-fer (changes requested):

greg-fer (changes requested):
My comments not yet applied.
Can you rebase the PR properly, as I'm getting lost what is new and what is not new among the 120 commits and 300+ files edited? Thanks.

greg-fer (approved):

koffes (approved):

PR #16304, opened 2024-07-03
cmake: pass root cache variables to child images

tejlmand:
When a ROOT cache variable is set by the user in the parent image then those cache variables should be passed to child images so that the child images will use the same root folders.

Jira: NCSDK-28257


shanthanordic (approved):

shanthanordic:
NEed +1 from @kylebonnici  and @magnev  before merging

gmarull (approved):

magnev (approved):

PR #16308, opened 2024-07-03
NCS v2.7.99 preparation

kl-cruz:


ns-tolu (approved):

nordic-piks (approved):
Test also locally - ADC and power consumption - OK.

PR #16312, opened 2024-07-03
net: openthread: rpc: interface management functions

Damian-Nordic:
Implement serialization of OpenThread APIs for managing the Thread interface state.
Add shell commands in the client application for testing the serialization.


applications/protocols_serialization/client/src/ot_shell.c
```diff
@@ -19,16 +23,244 @@ static int ot_console_cb(void *context, const char *format, va_list arg)
...
+	ot_cli_lazy_init(sh);
+	error = handler(sh, argc, argv);
+
+	if (error == OT_ERROR_NONE) {
+		shell_print(sh, "Done");
+	} else {
+		shell_print(sh, "Error %d", error);
+	}
+
+	return 0;
```
alxelax, line 59: Shouldn't it return any error code in case when handler returns error?

applications/protocols_serialization/client/src/ot_shell.c
```diff
@@ -19,16 +23,244 @@ static int ot_console_cb(void *context, const char *format, va_list arg)
...
+		shell_print(sh, "%u", (unsigned int)otLinkGetPollPeriod(NULL));
+		return OT_ERROR_NONE;
+	}
+
+	return otLinkSetPollPeriod(NULL, strtoul(argv[1], NULL, 10));
+}
+
+static otError ot_cli_command_state(const struct shell *sh, size_t argc, char *argv[])
+{
+	static const char *const role_str[] = {"disabled", "detached", "child", "router", "leader"};
```
alxelax, line 193: why is it static?

subsys/net/openthread/rpc/server/ot_rpc_ctrl_server.c
```diff
@@ -116,6 +117,35 @@ static void ot_rpc_cmd_ip6_get_unicast_addrs(const struct nrf_rpc_group *group,
...
+static void ot_rpc_cmd_ip6_get_multicast_addrs(const struct nrf_rpc_group *group,
+					       struct nrf_rpc_cbor_ctx *ctx, void *handler_data)
+{
+	struct nrf_rpc_cbor_ctx rsp_ctx;
+	const otNetifMulticastAddress *addr;
+	size_t addr_count;
+
+	nrf_rpc_cbor_decoding_done(group, ctx);
+
+	openthread_api_mutex_lock(openthread_get_default_context());
```
alxelax, line 129: Is it mandatory for any open thread api call?

alxelax (commented):
LGTM just a couple of questions.

applications/protocols_serialization/client/src/ot_shell.c
Damian-Nordic: it makes sense that it be in `rodata` instead of being constructed on stack on each call, no?

applications/protocols_serialization/client/src/ot_shell.c
Damian-Nordic: Does this change any behavior? The shell core does not display any error code on error etc. from my experience so I'm not even sure what the error code does in the shell handler.

subsys/net/openthread/rpc/server/ot_rpc_ctrl_server.c
Damian-Nordic: Yes, per nRF RPC architecture there can be multiple handling threads so we need a synchronization at OT level.

applications/protocols_serialization/client/src/ot_shell.c
alxelax: Names will be located in rodata even without static, the only array with pointers is constructed on stack (5 * 4) = 20 bytes. That doesn't seem a lot.

Damian-Nordic: Yes, I meant the array, of course. The assembly is a bit shorter that way. Is there a specific reason you don't like `static`?

alxelax: I do not like static if it is allocated on ram and static that consumes ram for rarely used function. If it is in flash then it is Ok.

applications/protocols_serialization/client/src/ot_shell.c
alxelax: Yes, it does. https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/shell/shell.h#L162-L175 api description.
[shell_get_return_value](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/shell/shell.h#L1292-L1299)

Damian-Nordic: ok, fixed.

applications/protocols_serialization/client/src/ot_shell.c
Damian-Nordic: Ok, I agree but in this case it's in flash.

alxelax (approved):

canisLupus1313 (approved):

PR #16209, opened 2024-06-26
nrf: rpc: move serializer and callback proxy to nrf rpc

alxelax:
Commit moves serialize and callback proxy functionality from bluetooth rpc to nrf rpc to make it available for the others.

PR #16335, opened 2024-07-05
net: openthread: rpc: serialized commissioning functions

alxelax:
PR adds:
-  serialization of some OpenThread commissioning functions;
-  serialization callback for discovery functionality;
-  shell commands to use serialized function in client application;

Known issues and limitations:
PR is based on changes from these PRs: 

-   https://github.com/nrfconnect/sdk-nrf/pull/16209
-   https://github.com/nrfconnect/sdk-nrf/pull/16312

So there is huge review overhead at the moment. Listed above PRs should be merged first.

Active dataset TLVs payload cannot fit with maximum length into serialization low levels.


subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	nrf_rpc_encode_uint(&ctx, (uintptr_t)context);
+	nrf_rpc_encode_uint(&ctx, callback_slot);
+
+	nrf_rpc_cbor_cmd_no_err(&ot_group, OT_RPC_CMD_THREAD_DISCOVER_CB, &ctx, ot_rpc_decode_void,
+				NULL);
+}
+
+NRF_RPC_CBKPROXY_HANDLER(ot_thread_discover_cb_encoder, ot_thread_discover_cb,
```
Damian-Nordic, line 55: Before I finish the review - if we start using the callback proxy for OpenThread group, we won't be able to easily write POSIX unit tests for the server side (which I started to work on). I think we have two choices:
1. Port the callback proxy assembly to x64 and x86. Kinda meh but possible :)
2. Do not use the callback proxy in OT commands taking advantage of the fact that all OT APIs that take callback parameters take an extra context as well. The server can allocate a pair `<client's callback, client's context>` and pass that pair as a context to the actual OT API. It's more portable and more flexible than the callback proxy approach.

What are your thoughts @canisLupus1313 @alxelax?

subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
```diff
@@ -0,0 +1,169 @@
...
+{
+	struct nrf_rpc_cbor_ctx ctx;
+	struct ot_rpc_active_scan_result pkt_data;
+	size_t cbor_buffer_size = result ? sizeof(pkt_data) + 1 : 1;
+
+	cbor_buffer_size += 10; /* for context and callback slot */
+	NRF_RPC_CBOR_ALLOC(&ot_group, ctx, cbor_buffer_size);
+
+	if (result) {
+		memcpy(pkt_data.ext_addr, result->mExtAddress.m8, OT_EXT_ADDRESS_SIZE);
```
Damian-Nordic, line 28: I wonder if it's OK to serialize structures as BLOBs or we should encode each field as a separate CBOR data item. This is probably a topic for a broader discussion but we should take a unified strategy on this. I guess separate CBOR data items make it easier to validate the data correctness and detect client & server incompatibility when a structure format changes.

Damian-Nordic (commented):
I think this PR raises two questions that we have to answer as a team. We can probably take some time in the stand up to discuss them.

subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
alxelax: > 
> 1. Port the callback proxy assembly to x64 and x86. Kinda meh but possible :)

Kind if disagree with this statement. Unit tests check serialization\deserialization of API. Callback is out of scope of this. All callback proxy API can be mocked, so unit tests will be possible to run on any architecture.

Damian-Nordic: Yes, but we still have to be able to build this file and test that the callback results in a proper serialization.

alxelax: It is possible to exclude it for unit tests. I do not see unresolvable issues. If OpenThread paradigm of context usage is changed, then ot rpc will require reimplementation.
Even more Bluetooth RPC should be tested as well. Potentially with unit tests too. There is no context in Bluetooth API.

Damian-Nordic: I'm also not saying this is unresolvable. I just mentioned the limitations this implementation imposes but like with other things we can discuss live and choose the best option.

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
```diff
@@ -0,0 +1,145 @@
...
+#include <ot_rpc_common.h>
+
+#include <nrf_rpc/nrf_rpc_serialize.h>
+#include <nrf_rpc/nrf_rpc_cbkproxy.h>
+
+static void ot_rpc_thread_discover_cb_rpc_handler(const struct nrf_rpc_group *group,
+						  struct nrf_rpc_cbor_ctx *ctx, void *handler_data)
+{
+	otHandleActiveScanResult callback_slot;
+	otActiveScanResult aResult;
```
Damian-Nordic, line 18: `aName` convention is meant to be used for function arguments only. I suggest renaming to follow NCS convention.

subsys/net/openthread/rpc/common/ot_rpc_common.c
```diff
@@ -0,0 +1,45 @@
...
+
+	if (data == NULL) {
+		handler_data = NULL;
+	} else {
+		dataset->mLength = length;
+		memcpy(dataset->mTlvs, data, dataset->mLength);
+	}
+}
+
+void ot_rpc_report_decoding_error(uint8_t cmd_evt_id, void *data)
```
Damian-Nordic, line 42: This is specifically for command decoding, not response decoding, right? Also, why is `data` argument needed if not used?

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
```diff
@@ -0,0 +1,145 @@
...
+#include <ot_rpc_types.h>
+#include <ot_rpc_common.h>
+
+#include <nrf_rpc/nrf_rpc_serialize.h>
+#include <nrf_rpc/nrf_rpc_cbkproxy.h>
+
+static void ot_rpc_thread_discover_cb_rpc_handler(const struct nrf_rpc_group *group,
+						  struct nrf_rpc_cbor_ctx *ctx, void *handler_data)
+{
+	otHandleActiveScanResult callback_slot;
```
Damian-Nordic, line 17: It's not a slot, it's already decoded callback, right?
```suggestion
	otHandleActiveScanResult callback;
```

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
```diff
@@ -0,0 +1,145 @@
...
+			 otHandleActiveScanResult aCallback, void *aCallbackContext)
+{
+	struct nrf_rpc_cbor_ctx ctx;
+	size_t cbor_buffer_size = 0;
+	otError error;
+
+	ARG_UNUSED(aInstance);
+
+	cbor_buffer_size += sizeof(uint32_t) + 1;  /* aScanChannels */
+	cbor_buffer_size += sizeof(uint32_t) + 1;  /* aPanId */
```
Damian-Nordic, line 68:
```suggestion
	cbor_buffer_size += sizeof(uint16_t) + 1;  /* aPanId */
```

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
```diff
@@ -0,0 +1,145 @@
...
+	size_t cbor_buffer_size;
+	otError error;
+
+	ARG_UNUSED(aInstance);
+
+	if (aDataset == NULL || aDataset->mLength > OT_OPERATIONAL_DATASET_MAX_LENGTH) {
+		return OT_ERROR_INVALID_ARGS;
+	}
+
+	cbor_buffer_size = aDataset->mLength + 1;
```
Damian-Nordic, line 116: This may be insufficient, you need an extra byte to encode the length itself.
```suggestion
	cbor_buffer_size = aDataset->mLength + 2;
```

subsys/net/openthread/rpc/common/ot_rpc_common.c
```diff
@@ -0,0 +1,45 @@
...
+				void *handler_data)
+{
+	otOperationalDatasetTlvs *dataset = *(otOperationalDatasetTlvs **)handler_data;
+	const void *data;
+	size_t length;
+
+	data = nrf_rpc_decode_buffer_ptr_and_size(ctx, &length);
+
+	if (data == NULL) {
+		handler_data = NULL;
```
Damian-Nordic, line 35: This does nothing. You should probably make `dataset` of type `otOperationalDatasetTlvs *` and then do `*dataset = NULL` in here.

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
```diff
@@ -0,0 +1,145 @@
...
+static void ot_rpc_thread_discover_cb_rpc_handler(const struct nrf_rpc_group *group,
+						  struct nrf_rpc_cbor_ctx *ctx, void *handler_data)
+{
+	otHandleActiveScanResult callback_slot;
+	otActiveScanResult aResult;
+	struct ot_rpc_active_scan_result result;
+	void *pkt_data = &result;
+	void *context;
+
+	pkt_data = nrf_rpc_decode_buffer(ctx, pkt_data, sizeof(result));
```
Damian-Nordic, line 23: This doesn't check if the payload length is exactly `sizeof(result)`, only if it's no bigger than that.

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
```diff
@@ -0,0 +1,145 @@
...
+	if (pkt_data) {
+		memcpy(aResult.mExtAddress.m8, result.ext_addr, OT_EXT_ADDRESS_SIZE);
+		memcpy(aResult.mNetworkName.m8, result.nwk_name, OT_NETWORK_NAME_MAX_SIZE + 1);
+		memcpy(aResult.mExtendedPanId.m8, result.ext_pan_id, OT_EXT_PAN_ID_SIZE);
+		memcpy(aResult.mSteeringData.m8, result.steering_data, OT_STEERING_DATA_MAX_LENGTH);
+		aResult.mPanId = result.pan_id;
```
Damian-Nordic, line 29: Hmm. This assumes `pan_id` is aligned while `result` is a packed structure, so it may not be, right? Probably using some Zephyr helpers for copying int byte by byte might be more portable/future-proof.

subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
```diff
@@ -0,0 +1,169 @@
...
+
+#include <zephyr/net/openthread.h>
+
+#include <openthread/link.h>
+
+static void ot_thread_discover_cb(otActiveScanResult *result, void *context, uint32_t callback_slot)
+{
+	struct nrf_rpc_cbor_ctx ctx;
+	struct ot_rpc_active_scan_result pkt_data;
+	size_t cbor_buffer_size = result ? sizeof(pkt_data) + 1 : 1;
```
Damian-Nordic, line 22:
```suggestion
	size_t cbor_buffer_size = result ? sizeof(pkt_data) + 2 : 1;
```
Extra byte for encoding the length.

subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
```diff
@@ -0,0 +1,169 @@
...
+						    struct nrf_rpc_cbor_ctx *ctx,
+						    void *handler_data)
+{
+	otOperationalDatasetTlvs dataset;
+	void *data_ptr = &dataset;
+	otError error;
+
+	ot_rpc_decode_dataset_tlvs(group, ctx, &data_ptr);
+
+	if (!nrf_rpc_decoding_done_and_check(group, ctx)) {
```
Damian-Nordic, line 124: You should also check if `data_ptr` is not nulled out since ot_rpc_decode_dataset_tlvs allows for NULL.

subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
alxelax: BLOB was splitted for set of separate parameters. 1 cbor data per structure member.

subsys/net/openthread/rpc/common/ot_rpc_common.c
alxelax: Yep, this is for command decoding.
I'll remove `data`. Already do not remember my initial reason for it. :)

subsys/net/openthread/rpc/common/ot_rpc_common.c
alxelax: Didn't get your statement.
It does what is written here, writes NULL to notify the function above about absence of data. `dataset` of type `otOperationalDatasetTlvs *` has been made a couple of lines above.
Please clarify what you meant.

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
alxelax: it was redesigned to separate cbor types. not actual anymore

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
alxelax: It was redesigned, not actual anymore.
Nevertheless, this implementation is correct. Compiler follows up data structure attributes and generates alignment safe backend code to get access to packed members.
The nonalignment fault is possible only if the type was cast explicitly and the compiler lost knowledge about packing in a new type.

alxelax: btw. this is classic approach for parsing \ composing of serialized data without taking into account shifts, positioning etc. Compiler will do everything for you by self on safe manner.
Read it in some computer science book, unfortunately forgot which one. :)

subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
alxelax: was redesigned. not actual anymore

subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
alxelax: My initial assumption was that Open Thread will check it is not NULL for me and return an appropriate error code if it is NULL.
Is it wrong?

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
Damian-Nordic: Ok, never found any GCC guarantees that this is handled already, but I checked in assembly that you're right. Thanks!

subsys/net/openthread/rpc/common/ot_rpc_common.c
Damian-Nordic: I meant it's not obvious from the function name so it's worth adding a comment in the declaration :)

subsys/net/openthread/rpc/client/ot_rpc_commissioning_client.c
alxelax: Yep, I checked it too on previous projects where I participated with several compilers and platforms. GCC, IAR, Keil all are managed this correctly for different platforms (checked for arm, avr32, arc).  I guess this is considered by any compiler developers.

subsys/net/openthread/rpc/common/ot_rpc_common.c
Damian-Nordic: `handler_data` is a local variable,  so you're not updating the user provided variable anyhow.

subsys/net/openthread/rpc/server/ot_rpc_commissioning_server.c
Damian-Nordic: The thing was that you always passed a valid `&dataset` pointer even if this was not populated. But now you added the `data_ptr` check so this is OK.

subsys/net/openthread/rpc/common/ot_rpc_common.c
alxelax: You're right, good catch. Will fix it.

Damian-Nordic (approved):

canisLupus1313 (approved):

PR #15210, opened 2024-05-07
Fix TF-M flash region configuration for nRF54L15

Vge0rge:
See commit messages for descriptions.

PR #16343, opened 2024-07-05
MCUBoot + TF-M with PS/ITS support for nRF54LX

sigvartmh:
PR [#15210](https://github.com/nrfconnect/sdk-nrf/pull/15210) needs to be merged before this and I'll rebase it on-top when that's done. Then I'll remove the TF-M assert change.

This brings changes which was in an old PR which allows for TF-M to be built together with NCS MCUBoot. It also adds support for placing TF-M protected storage and internal trusted storage in front of the firmware application according to memory layout design document.

Ref: NCSDK-27354
Ref: NCSDK-24863


modules/trusted-firmware-m/tfm_boards/common/assert.c
```diff
@@ -24,7 +24,7 @@
 	"\nRefer to the documentation section 'TF-M partition alignment requirements'" \
 	"\nfor more information.\n\n"
 
-#error "TF-M non-secure start address is not aligned to SPU HW requirements"
+#warning"TF-M non-secure start address is not aligned to SPU HW requirements"
```
nordicjm, line 27: this should be an error because it deals with hardware registers

modules/trusted-firmware-m/tfm_boards/common/assert.c
```diff
@@ -24,7 +24,7 @@
 	"\nRefer to the documentation section 'TF-M partition alignment requirements'" \
```
vili-nordic, line 24: The MPC handles this on nrf54l, so it may be a bit misleading to refer to spu requirements.

modules/trusted-firmware-m/tfm_boards/common/assert.c
sigvartmh: I tried to inform that this is a temporary work around until the PR which fixes this is merged.

> PR https://github.com/nrfconnect/sdk-nrf/pull/15210 needs to be merged before this and I'll rebase it on-top when that's done. 

This is only a temporary workaround just to have it compiling since the limitations doesn't apply for nRF54L15

modules/trusted-firmware-m/tfm_boards/common/assert.c
sigvartmh: If you have any feed back on this assert please use the following PR to provide it: 

> PR https://github.com/nrfconnect/sdk-nrf/pull/15210 needs to be merged before this and I'll rebase it on-top when that's done. 

This commit is not intended to be included in this PRs final form. It's to be rebased on-top of the changes to this but that PR was stale which is why I posted with a workaround.

sigvartmh: > Currently waiting for the fix found in PR https://github.com/nrfconnect/sdk-nrf/pull/15210 so this is a work
around since the alignment requirements is wrong for NRF54LX.

modules/trusted-firmware-m/tfm_boards/common/assert.c
sigvartmh: Rebased PR on-top now.

nordicjm (approved):

Vge0rge (approved):

vili-nordic (approved):

PR #16348, opened 2024-07-08
sysbuild: Add psuedo imply for MCUboot flash sim mediate

nordicjm:
Adds an overlay file for MCUboot on nRF53 series devices which will set the flash simulator mediate Kconfig value, this acts as an imply as the setting can later be changed


modules/mcuboot/flash_sim.conf
```diff
@@ -0,0 +1 @@
+CONFIG_ADD_MCUBOOT_MEDIATE_SIM_FLASH_DTS=y
```
joerchan, line 1: This does not look right.

This config depends on CONFIG_BOOTLOADER_MCUBOOT, so it cannot be added to the MCUboot image.
This option was used to add a flash_sim DTS overlay file to mcuboot, to enable CONFIG_FLASH_SIMULATOR=y in mcuboot.

joerchan:
There should be something in CI that builds this.
The only thing I could find was `samples/bluetooth/peripheral_uart` with `nrf5340dk_app_sr_net.conf` as the overlay config.
However the nRF5340DK does not have external flash, so it would perhaps be better to have this for thingy53 instead to test it?

PR #16350, opened 2024-07-08
sysbuild: Add support for selecting app core image on net core builds

nordicjm:
Adds support for selecting an application core image when building for the network core of a supported device. Allows selection of none, empty app or remote shell

NCSDK-28221


krish2718 (approved):

sysbuild/Kconfig.appcore
```diff
@@ -0,0 +1,89 @@
...
+	bool
+
+config NRF_DEFAULT_REMOTE_SHELL
+	bool
+
+choice APPCORE
+	prompt "Appcore image"
+	default APPCORE_REMOTE_SHELL if NRF_DEFAULT_REMOTE_SHELL && APPCORE_REMOTE_BOARD_NAME != ""
+	default APPCORE_EMPTY if NRF_DEFAULT_EMPTY && APPCORE_REMOTE_BOARD_NAME != ""
+	depends on SUPPORT_APPCORE && !EXTERNAL_CONFIGURED_APPCORE
```
dchat-nordic, line 38: I don't see the option `EXTERNAL_CONFIGURED_APPCORE` defined anywhere. Should it be added?

nordicjm: No, that is a workaround for upstream samples that add a network core image in some builds, there aren't any that add an application core image

dchat-nordic: So to clarify if a sample uses different appcore image (ie. DTM uses remote_hci) what is the proper way to configure it? Is it:
- Use the EXTERNAL_CONFIGURED_APPCORE
- or add the options and image names through Kconfig.sysbuild in sample directory?

nordicjm: See second commit in PR

57300 (approved):

dchat-nordic (approved):

sysbuild/Kconfig.appcore
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) 2024 Nordic Semiconductor
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+
+config SUPPORT_APPCORE
+	bool
+	default y if (SOC_NRF5340_CPUNET || SOC_NRF54H20_CPURAD)
```
KAGA164, line 7: I am not sure if this should be enabled bu default for nRF54h20. It is mandatory for nRF53 because netcore is enabled by application core so app core needs to run any application always

dchat-nordic: Good catch, but I think the `SUPPORT_APPCORE` is fine. The `APPCORE` choice should have a default for APPCORE_NONE in case of nRF54.

PR #16352, opened 2024-07-08
adv_prov: fast_pair: Fix uninitialized structure

KAGA164:
Fixes coverity issue 204627

### DIFF
--- a/subsys/bluetooth/adv_prov/providers/fast_pair.c
+++ b/subsys/bluetooth/adv_prov/providers/fast_pair.c
@@ -82,7 +82,7 @@ static int get_data(struct bt_data *ad, const struct bt_le_adv_prov_adv_state *s
 		    struct bt_le_adv_prov_feedback *fb)
 {
 	static uint8_t buf[ADV_DATA_BUF_SIZE];
-	struct bt_fast_pair_adv_config adv_config;
+	struct bt_fast_pair_adv_config adv_config = { 0 };
 
 	if (!enabled) {
 		return -ENOENT;


dchat-nordic (approved):

maje-emb (approved):

subsys/bluetooth/adv_prov/providers/fast_pair.c
```diff
@@ -82,7 +82,7 @@ static int get_data(struct bt_data *ad, const struct bt_le_adv_prov_adv_state *s
 		    struct bt_le_adv_prov_feedback *fb)
 {
 	static uint8_t buf[ADV_DATA_BUF_SIZE];
-	struct bt_fast_pair_adv_config adv_config;
+	struct bt_fast_pair_adv_config adv_config = { 0 };
```
MarekPieta, line 85: All of the fields are filled a few lines below before the structure is used. Is the structure actually uninitialized?

MarekPieta: Discussed offline. There is no bug in the current code. Still, zeroing struct fields would not hurt here.

MarekPieta (approved):

subsys/bluetooth/adv_prov/providers/fast_pair.c
KAGA164: Maybe. But it is good practice to clean-up local structure when field filling is spread across whole function body

kapi-no (approved):

PR #16355, opened 2024-07-08
applications: nrf5340_audio: Module to read LC3 files from SD card

andvib:
Reads and parses LC3 files from the SD card, and provides a function to read the LC3 file one frame at a time. Uses the existing SD card module.

OCT-3031


applications/nrf5340_audio/src/modules/lc3_file.c
```diff
@@ -0,0 +1,101 @@
...
+
+#include "lc3_file.h"
+#include "sd_card.h"
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(sd_card_lc3_file, CONFIG_MODULE_SD_CARD_LC3_FILE_LOG_LEVEL);
+
+int lc3_file_frame_get(struct lc3_file_ctx *file, uint8_t *buffer, size_t buffer_size)
+{
+	int err;
```
alexsven, line 15:
```suggestion
	int ret;
```

applications/nrf5340_audio/src/modules/lc3_file.c
```diff
@@ -0,0 +1,101 @@
...
+
```
alexsven, line 16:
```suggestion
if (file == NULL) {
        LOG_ERR("Invalid pointer");
        return -EINVAL;
}
```
Do the same for buffer

alexsven: Same for all functions in this file

applications/nrf5340_audio/src/modules/lc3_file.c
alexsven: Same for entire file. We use `ret` in these neck of the woods ;)

applications/nrf5340_audio/src/modules/lc3_file.c
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	/* Read LC3 header and store in struct */
+	err = sd_card_read((char *)&file->lc3_header, &size, &file->file_object);
+	if (err) {
+		LOG_ERR("Failed to read the LC3 header: %d", err);
+		return err;
+	}
+
+	if (file->lc3_header.file_id != 0xcc1c) {
```
alexsven, line 69: Create a define for this ID. Is this ID unique to the LC3 tool we use, all will it be the same for all LC3 PC tools?

tests/nrf5340_audio/lc3_file/src/main.c
```diff
@@ -0,0 +1,348 @@
...
+}
+
+ZTEST(lc3_file, test_lc3_file_close)
+{
+	int ret;
+	struct lc3_file_ctx file;
+
+	ret = lc3_file_close(&file);
+
+	zassert_equal(0, ret, "lc3_file_close() should return success");
```
alexsven, line 311: Shouldn't this return `-EPERM` since there is no open before this?

alexsven (changes requested):

tests/nrf5340_audio/lc3_file/src/main.c
andvib: The underlying API probably does, but this functionality gets a bit "lost" in the mocking. lc3_file_close does nothing to check whether a file has been opened, but as long as the underlying API does the error will be forwarded, as is tested by `test_lc3_file_close_invalid`.

It could be a discussion if the `lc3_file` module should check this though. Right now I can't think of this being something that must be guarded against as only the calls to `fs` (probably) must be initialized correctly before using the objects, which should be checked in that API.

gWacey (changes requested):

alexsven (approved):

tests/nrf5340_audio/lc3_file/src/main.c
```diff
@@ -0,0 +1,348 @@
...
+
+	FFF_RESET_HISTORY();
+
+	read_lc3_fake_bytes_read = 0;
+}
+
+int sd_card_read_header_fake_invalid(char *buf, size_t *size, struct fs_file_t *f_seg_read_entry)
+{
+	uint8_t lc3_headerarray[] = {
+		0x1c, 0xFF, /* Invalid file id, is normally 0xcc1c */
```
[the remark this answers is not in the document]
andvib: Good idea! This will already be useful for the next module. Added a `mocks` folder under `tests/nrf5340_audio` that contains everything - please let me know if there's something more that could be done to make it more readable and reusable!

tests/nrf5340_audio/lc3_file/src/main.c
```diff
@@ -0,0 +1,348 @@
...
+	memcpy(buf, &lc3_data[read_lc3_fake_bytes_read], read_size);
+
+	read_lc3_fake_bytes_read += read_size;
+	*size = read_size;
+
+	return 0;
+}
+
+ZTEST(lc3_file, test_lc3_file_frame_get_valid)
+{
```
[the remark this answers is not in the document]
andvib: Not sure I get completely what you're thinking of here - this would be to split the tests into lc3-parsing specific tests and into more general file handling tests?

tests/nrf5340_audio/lc3_file/src/main.c
```diff
@@ -0,0 +1,348 @@
...
+
+static int read_lc3_fake_bytes_read;
+
+static void test_setup(void *f)
+{
+	RESET_FAKE(sd_card_init);
+	RESET_FAKE(sd_card_open);
+	RESET_FAKE(sd_card_close);
+	RESET_FAKE(sd_card_read);
+
```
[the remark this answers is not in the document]
andvib: I haven't seen this before - but I do like the idea! But `DO_FOREACH_FAKE` is something that must be implemented in this file, and not available in a FFF header or something? As it would only be used here with the current structure I'm not sure if the abstraction makes it better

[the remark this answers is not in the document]
gWacey: Sorry yes it would be implemented in the fakes.h as follows:

/* List of fakes used by this unit tester */
#define DO_FOREACH_FAKE(FUNC)                                                                      \
	do {                                                                                       \
		FUNC(sd_card_init)                                                                \
		FUNC(sd_card_open)                                                              \
		FUNC(sd_card_close)                                                         \
		FUNC(sd_card_read)                                                         \
	} while (0)

tests/nrf5340_audio/lc3_file/src/main.c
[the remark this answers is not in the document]
gWacey: Ignore this comment!

gWacey (changes requested):
Only other point is the mixing of mock and fake. These are different techniques, there is also stubs, for testing and I think it is all fakes here?

**_Mocking_**: Mock functions are also known as "spies", because they let you spy on the behaviour of a function that is called indirectly by some other code, rather than only testing the output.
**_Stubs_**: Stubs are simulations of actual functions, which can be used to isolate the function under test and to check that calls to the stubbed function are correctly formulated. 
**_Fakes_**: The purpose of a fake is not to affect the behaviour of the system under test, but rather to simplify the implementation of it. A fake is an implementation that behaves “naturally” but is not “real”.

tests/nrf5340_audio/lc3_file/src/main.c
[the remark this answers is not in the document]
gWacey: Rather than mocks could it be fakes?

gWacey (approved):
Looks good, nice work.

PR #16356, opened 2024-07-08
nrf_security: Support legacy and PSA APIs in 54L15

Vge0rge:
This adds support for the configuration
NRF_SECURITY_LEGACY_AND_PSA with the nRF54L15
devices.

Ref: NCSDK-27932

### DIFF
--- a/subsys/nrf_security/Kconfig
+++ b/subsys/nrf_security/Kconfig
@@ -39,7 +39,7 @@ config NRF_SECURITY
 	default y if ENTROPY_PSA_CRYPTO_RNG && SOC_SERIES_NRF54LX
 	select DISABLE_MBEDTLS_BUILTIN if MBEDTLS
 	# Generating random requires a CRACEN PSA Crypto driver on nrf54L
-	select PSA_CRYPTO_DRIVER_CRACEN if PSA_WANT_GENERATE_RANDOM && SOC_SERIES_NRF54LX
+	select PSA_CRYPTO_DRIVER_CRACEN if (PSA_WANT_GENERATE_RANDOM && SOC_SERIES_NRF54LX) && !NRF_SECURITY_LEGACY_AND_PSA
 	# NCS does not use TF-M's BL2 bootloader, but uses it's own fork
 	# of MCUBoot instead (CONFIG_BOOTLOADER_MCUBOOT).
 	#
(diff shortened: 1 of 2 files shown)


subsys/nrf_security/src/drivers/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ else()
 	add_subdirectory(nrf_cc3xx)
   endif()
 
-  if(CONFIG_PSA_CRYPTO_DRIVER_CRACEN)
+  if(CONFIG_PSA_CRYPTO_DRIVER_CRACEN OR (CONFIG_NRF_SECURITY_LEGACY_AND_PSA AND CONFIG_SOC_SERIES_NRF54LX))
```
tomi-font, line 15: Seems a bit counterintuitive as in the new case `CONFIG_PSA_CRYPTO_DRIVER_CRACEN` is not selected, but still added in the CMake logic.

tomi-font (approved):

subsys/nrf_security/Kconfig
```diff
@@ -39,7 +39,7 @@ config NRF_SECURITY
 	default y if ENTROPY_PSA_CRYPTO_RNG && SOC_SERIES_NRF54LX
 	select DISABLE_MBEDTLS_BUILTIN if MBEDTLS
 	# Generating random requires a CRACEN PSA Crypto driver on nrf54L
-	select PSA_CRYPTO_DRIVER_CRACEN if PSA_WANT_GENERATE_RANDOM && SOC_SERIES_NRF54LX
+	select PSA_CRYPTO_DRIVER_CRACEN if (PSA_WANT_GENERATE_RANDOM && SOC_SERIES_NRF54LX) && !NRF_SECURITY_LEGACY_AND_PSA
```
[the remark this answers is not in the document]
Vge0rge: I added it because they are logically grouped, I hoped that this makes it more clear.

subsys/nrf_security/src/drivers/CMakeLists.txt
Vge0rge: Yes, it is indeed counterintuitive but for this particular use case necessary. For the  legacy and PSA config we need to include that to get RNG even though we don't use the rest of the driver. We use Oberon for everything except RNG. This is an experimental feature (legacy and PSA) which we hope to get it removed later when then adoption of PSA is better.

vili-nordic (approved):

PR #16360, opened 2024-07-09
application: serial_lte_modem: Automation of LwM2M requests

junqingzou:
Add `CONFIG_SLM_CARRIER_AUTO_CONTROL` to control automatical handling of below requests from LwM2M in SLM, instead of by host MCU:
 - Link down/up
 - Power off modem
 - Reboot

By default this CONFIG is enabled.

JIRA-ticket: NCSDK-28267
JIRA-ticket: NCSDK-28431

### DIFF
--- a/applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
+++ b/applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
@@ -31,27 +31,27 @@ These may be the following:
 
 * ``#XCARRIEREVT: 1,<status>``
 
-  Request to set the modem to full functional mode.
+  Request to use the ``AT+CFUN=1`` AT command to set the modem to full functional mode.
   ``<status>`` returns two possible values:
 
-  * ``0`` - Request handled successfully.
-  * ``-1`` - Request handling was deferred and must be performed by the application at its earliest convenience.
+  * ``0`` - Request handling is fulfilled automatically.
+  * ``-1`` - Request handling is deferred and must be fulfilled by the application at its earliest convenience.
 
 * ``#XCARRIEREVT: 2,<status>``
 
-  Request to set the modem to flight functional mode.
+  Request to use the ``AT+CFUN=4`` AT command to set the modem to flight functional mode.
   ``<status>`` returns two possible values:
 
-  * ``0`` - Request handled successfully.
-  * ``-1`` - Request handling was deferred and must be performed by the application at its earliest convenience.
+  * ``0`` - Request handling is fulfilled automatically.
+  * ``-1`` - Request handling is deferred and must be fulfilled by the application at its earliest convenience.
 
 * ``#XCARRIEREVT: 3,<status>``
 
-  Request to set the modem to minimum functional mode.
+  Request to use the ``AT+CFUN=0`` AT command to set the modem to minimum functional mode.
   ``<status>`` returns two possible values:
 
-  * ``0`` - Request handled successfully.
-  * ``-1`` - Request handling was deferred and must be performed by the application at its earliest convenience.
+  * ``0`` - Request handling is fulfilled automatically.
+  * ``-1`` - Request handling is deferred and must be fulfilled by the application at its earliest convenience.
 
 * ``#XCARRIEREVT: 4,0``
 
@@ -94,12 +94,12 @@ These may be the following:
 
 * ``#XCARRIEREVT: 10,<status>``
 
-  Request to perform an application reboot.
+  Request to perform an application reboot, for example using the ``AT#XRESET`` AT command.
 
   ``<status>`` returns two possible values:
 
-  * ``0`` - Request handled successfully.
-  * ``-1`` - Request handling was deferred and must be performed by the application at its earliest convenience.
+  * ``0`` - Request handling is fulfilled automatically.
+  * ``-1`` - Request handling is deferred and must be fulfilled by the application at its earliest convenience.
 
 * ``#XCARRIEREVT: 11,0``
 
--- a/applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
+++ b/applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
@@ -25,4 +25,13 @@ config SLM_CARRIER_AUTO_STARTUP
 	  Enable automatic startup of the library on device boot. If this configuration is
 	  disabled, automatic startup is controlled through a dedicated AT command.
 
+config SLM_CARRIER_AUTO_CONTROL
+	bool "Automatic handling of LwM2M requests"
+	depends on SLM_AUTO_CONNECT
+	default y
+	help
+	  Enable automatic power off, reboot, link up or link down when requested by the LwM2M
+	  stack. If this is disabled, the host MCU must perform these operations via AT commands
+	  when requested.
+
 endif # SLM_CARRIER
--- a/applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
+++ b/applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
@@ -148,18 +148,36 @@ static void reconnect_wk(struct k_work *work)
 int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 {
 	int err = 0;
-	static bool fota_started;
 
 	switch (event->type) {
+#if defined(CONFIG_SLM_CARRIER_AUTO_CONTROL)
 	case LWM2M_CARRIER_EVENT_LTE_LINK_UP:
 		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_LINK_UP");
-		if (fota_started) {
-			fota_started = false;
-			k_work_reschedule(&reconnect_work, K_MSEC(100));
-		} else {
-			/* AT+CFUN=1 to be issued. */
+		k_work_reschedule(&reconnect_work, SLM_UART_RESPONSE_DELAY);
+		break;
+	case LWM2M_CARRIER_EVENT_LTE_LINK_DOWN:
+		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_LINK_DOWN");
+		err = slm_util_at_printf("AT+CFUN=4");
+		if (err) {
 			err = -1;
 		}
+		break;
+	case LWM2M_CARRIER_EVENT_LTE_POWER_OFF:
+		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_POWER_OFF");
+		err = slm_util_at_printf("AT+CFUN=0");
+		if (err) {
+			err = -1;
+		}
+		break;
+	case LWM2M_CARRIER_EVENT_REBOOT:
+		LOG_DBG("LWM2M_CARRIER_EVENT_REBOOT");
+		/* Return 0 to reboot right after a URC. */
+		break;
+#else
+	case LWM2M_CARRIER_EVENT_LTE_LINK_UP:
+		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_LINK_UP");
+		/* AT+CFUN=1 to be issued. */
+		err = -1;
 		break;
 	case LWM2M_CARRIER_EVENT_LTE_LINK_DOWN:
 		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_LINK_DOWN");
@@ -168,9 +186,15 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 		break;
 	case LWM2M_CARRIER_EVENT_LTE_POWER_OFF:
 		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_POWER_OFF");
-		/* TODO: defer setting the modem to minimum funtional mode. */
-		err = slm_util_at_printf("AT+CFUN=0");
+		/* AT+CFUN=0 to be issued. */
+		err = -1;
 		break;
+	case LWM2M_CARRIER_EVENT_REBOOT:
+		LOG_DBG("LWM2M_CARRIER_EVENT_REBOOT");
+		/* Return -1 to defer the reboot until the application decides to do so. */
+		err = -1;
+		break;
+#endif /* CONFIG_SLM_CARRIER_AUTO_CONTROL */
 	case LWM2M_CARRIER_EVENT_BOOTSTRAPPED:
 		LOG_DBG("LWM2M_CARRIER_EVENT_BOOTSTRAPPED");
 		break;
@@ -185,15 +209,9 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 		return 0;
 	case LWM2M_CARRIER_EVENT_FOTA_START:
 		LOG_DBG("LWM2M_CARRIER_EVENT_FOTA_START");
-		fota_started = true;
 		break;
 	case LWM2M_CARRIER_EVENT_FOTA_SUCCESS:
 		LOG_DBG("LWM2M_CARRIER_EVENT_FOTA_SUCCESS");
-		break;
-	case LWM2M_CARRIER_EVENT_REBOOT:
-		LOG_DBG("LWM2M_CARRIER_EVENT_REBOOT");
-		/* Return -1 to defer the reboot until the application decides to do so. */
-		err = -1;
 		break;
 	case LWM2M_CARRIER_EVENT_MODEM_DOMAIN:
 		LOG_DBG("LWM2M_CARRIER_EVENT_MODEM_DOMAIN");
@@ -219,6 +237,13 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 	}
 
 	rsp_send("\r\n#XCARRIEREVT: %d,%d\r\n", event->type, err);
+
+#if defined(CONFIG_SLM_CARRIER_AUTO_CONTROL)
+	/* Allow time for the URC be flushed before reboot */
+	if (event->type == LWM2M_CARRIER_EVENT_REBOOT && err == 0) {
+		k_sleep(SLM_UART_RESPONSE_DELAY);
+	}
+#endif
 
 	return err;
 }


applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -220,6 +226,11 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 
 	rsp_send("\r\n#XCARRIEREVT: %d,%d\r\n", event->type, err);
 
+	/* Allow time for the URC be flushed before reboot */
+	if (event->type == LWM2M_CARRIER_EVENT_REBOOT && err == 0) {
+		k_sleep(K_MSEC(100));
```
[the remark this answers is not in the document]
junqingzou: SBM asked for a delay of 5s and I offered 100ms as I'm not sure about context of callback in LwM2M stack. SLM_UART_RESPONSE_DELAY is 50ms.

lucasvdp (commented):
This solution only works for a succesfull fota. I think we should add a setting, either compile time or in the form of an AT command. This setting could allow SLM to automatically do everything or let everything up to the user.
It will be hard to capture all the scenarios after a FOTA_START event because some events require us to do something after the FOTA_SUCCES or ERROR FOTA_FAIL event.
Also for other Carrier we get these events that require an action during bootstrap. It would make sense to also allow these to be automated.

junqingzou:
> This solution only works for a succesfull fota. I think we should add a setting, either compile time or in the form of an AT command. This setting could allow SLM to automatically do everything or let everything up to the user. It will be hard to capture all the scenarios after a FOTA_START event because some events require us to do something after the FOTA_SUCCES or ERROR FOTA_FAIL event. Also for other Carrier we get these events that require an action during bootstrap. It would make sense to also allow these to be automated.

CONFIG_SLM_CARRIER_AUTO_REBOOT is added to control the behavior of auto reboot after APP FOTA download. Other scenario can refer to this config but not in the scope of this PR, I think.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -192,8 +192,20 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
...
+			fota_started = false;
+			err = 0;
+		} else {
+			/* Return -1 to defer the reboot until the application decides to do so. */
+			err = -1;
+		}
+#else
 		/* Return -1 to defer the reboot until the application decides to do so. */
 		err = -1;
+
```
tomi-font, line 207:
```suggestion
```

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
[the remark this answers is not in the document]
tomi-font: `SLM_UART_RESPONSE_DELAY` could be increased to 100ms if you think 100ms is needed. I was thinking it'd be nice to use it here because it's for the same purpose than other places using it.

tomi-font (approved):

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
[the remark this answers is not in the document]
junqingzou: @tomi-font SBM just accepted current behavior without further requesting time (5min) for host side to prepare for a reboot. So I'll silently change the delay from 100ms to 50ms in next submit.

junqingzou:
> `CONFIG_SLM_CARRIER_AUTO_REBOOT` is a good improvement, but I think we should also get rid of the `fota_started` variable. It is still not unset when a fota does not end with a reboot (app fota finished) or link up (modem fota finished). We can simply allow all reboots, link_up, link_down and power_off. Maybe rename `CONFIG_SLM_CARRIER_AUTO_REBOOT` to `CONFIG_SLM_CARRIER_AUTO_HANDLE_EVENTS`? I understand wanting to limit the scope of this PR to reboot only, but I don't like having a half working inbetween state and my suggestion is a quicker fix than keeping proper track of the `fota_started`.

Let's get this in first for SBM IOT. For further improvement or design change, I'll leave it to Kacper.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -153,18 +153,33 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 	switch (event->type) {
 	case LWM2M_CARRIER_EVENT_LTE_LINK_UP:
 		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_LINK_UP");
+#if defined(SLM_CARRIER_FOTA_AUTOMATION)
```
tomi-font, line 156:
```suggestion
#if defined(CONFIG_SLM_CARRIER_FOTA_AUTOMATION)
```
Same for the other instances of this.

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -220,6 +246,13 @@ int lwm2m_carrier_event_handler(const lwm2m_carrier_event_t *event)
 
 	rsp_send("\r\n#XCARRIEREVT: %d,%d\r\n", event->type, err);
 
+#if defined(SLM_CARRIER_FOTA_AUTOMATION)
+	/* Allow time for the URC be flushed before reboot */
+	if (event->type == LWM2M_CARRIER_EVENT_REBOOT && err == 0) {
+		k_sleep(SLM_UART_RESPONSE_DELAY);
```
tomi-font, line 252: Why not rather put this [here](https://github.com/nrfconnect/sdk-nrf/pull/16360/files#diff-1c276f4a29001e10696ebbf7251866a78b4593747545d5a151ca4c4e44040dd4R214)?

applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
```diff
@@ -25,4 +25,14 @@ config SLM_CARRIER_AUTO_STARTUP
...
 	  disabled, automatic startup is controlled through a dedicated AT command.
 
+config SLM_CARRIER_FOTA_AUTOMATION
+	bool "Enable automatic handling after FOTA download"
+	depends on SLM_AUTO_CONNECT
+	default y
+	help
+	  Enable automatic reboot when requested by LwM2M stack. If this configuration is
+	  disabled, reboot is controlled by host MCU.
+	  Enable automatic link down/up when requested by LwM2M stack. If this configuration is
```
tomi-font, line 35: `during a FOTA`?

tomi-font (changes requested):

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
junqingzou: No, after URC is sent

applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
junqingzou: Automation for FOTA context only

tomi-font: Yeah, that's why I think it would be good to make this more clear in the Kconfig option help message.

kacperradoszewski:
This works for the simple success/failure FOTA cases but does not behave well with some other scenarios, that don't always apply to SBM but do affect other carriers, which can lead to the `fota_started` variable to be out of sync (RAT switch, some errors that happen before we push anything to DFU, and so on). So similarly to @lucasvdp, I also suggest removing that one and handle the requests unconditionally, perhaps under a common `CONFIG_SLM_CARRIER_AUTO_REQUEST_HANDLING`, `CONFIG_SLM_CARRIER_AUTO_HANDLE_EVENTS` or something like that.

LINK_UP and LINK_DOWN events are only really sent during FOTA and bootstrap process (to write PSKs into the modem, this does not apply to SBM), so I don't think it hurts to automate these altogether, since we already wanted to automate the bootstrap too. It also won't hurt to automate REBOOT.

I don't mind if this is controlled with a KConfig or an AT command, I think it will suffice if this is a static setting.

junqingzou:
> This works for the simple success/failure FOTA cases but does not behave well with some other scenarios, that don't always apply to SBM but do affect other carriers, which can lead to the `fota_started` variable to be out of sync (RAT switch, some errors that happen before we push anything to DFU, and so on). So similarly to @lucasvdp, I also suggest removing that one and handle the requests unconditionally, perhaps under a common `CONFIG_SLM_CARRIER_AUTO_REQUEST_HANDLING`, `CONFIG_SLM_CARRIER_AUTO_HANDLE_EVENTS` or something like that.
> 
> LINK_UP and LINK_DOWN events are only really sent during FOTA and bootstrap process (to write PSKs into the modem, this does not apply to SBM), so I don't think it hurts to automate these altogether, since we already wanted to automate the bootstrap too. It also won't hurt to automate REBOOT.
> 
> I don't mind if this is controlled with a KConfig or an AT command, I think it will suffice if this is a static setting.

What you said is automate LINK_UP/LINK_DOWN/REBOOT with an static CONFIG?
SLM has unconditional POWER OFF/SHUTDOWN MODEM/INIT MODEM as of now

kacperradoszewski:
> What you said is to automate LINK_UP/LINK_DOWN/REBOOT with an static CONFIG beyond FOTA scenario?

Yes, this will simplify the code and ensure we don’t enter some weird out-of-sync states.

> SLM has unconditional POWER OFF/SHUTDOWN MODEM/INIT MODEM as of now

Modem shutdown and init are unconditional because there are no AT commands to handle those I think? (I might be mistaken, perhaps SLM has some mechanism)
Don’t remember the reason for different handling of power off vs link down/up, but we can potentially put that one under that config too.

junqingzou:
> > What you said is to automate LINK_UP/LINK_DOWN/REBOOT with an static CONFIG beyond FOTA scenario?
> 
> Yes, this will simplify the code and ensure we don’t enter some weird out-of-sync states.
> 
> > SLM has unconditional POWER OFF/SHUTDOWN MODEM/INIT MODEM as of now
> 
> Modem shutdown and init are unconditional because there are no AT commands to handle those I think? (I might be mistaken, perhaps SLM has some mechanism) Don’t remember the reason for different handling of power off vs link down/up, but we can potentially put that one under that config too.

OK, I'll make the automation for REBOOT/LINK UP/LINK DOWN unconditional.
Modem shutdown and init were made unconditional when rebootless MFW DFU was added in LwM2M by Stig as I remember.

applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
```diff
@@ -25,4 +25,12 @@ config SLM_CARRIER_AUTO_STARTUP
 	  Enable automatic startup of the library on device boot. If this configuration is
 	  disabled, automatic startup is controlled through a dedicated AT command.
 
+config SLM_CARRIER_AUTO_CONTROL
+	bool "Enable automatic handling of LwM2M request"
```
tomi-font, line 29:
```suggestion
	bool "Automatic handling of LwM2M requests"
```

applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
```diff
@@ -25,4 +25,12 @@ config SLM_CARRIER_AUTO_STARTUP
...
+	depends on SLM_AUTO_CONNECT
+	default y
+	help
+	  Enable automatic power off, reboot, link up or link down when requested by LwM2M stack.
```
tomi-font, line 33:
```suggestion
	  Enable automatic power off, reboot, link up or link down when requested by the LwM2M stack.
```

applications/serial_lte_modem/src/lwm2m_carrier/Kconfig
```diff
@@ -25,4 +25,12 @@ config SLM_CARRIER_AUTO_STARTUP
...
+	  If this configuration is disabled, control is by the host MCU via AT command.
```
tomi-font, line 34:
```suggestion
	  If this is disabled, the host MCU must perform these operations via AT commands when requested.
```

applications/serial_lte_modem/src/lwm2m_carrier/slm_at_carrier.c
```diff
@@ -148,18 +148,30 @@ static void reconnect_wk(struct k_work *work)
...
-			k_work_reschedule(&reconnect_work, K_MSEC(100));
-		} else {
-			/* AT+CFUN=1 to be issued. */
-			err = -1;
-		}
+		k_work_reschedule(&reconnect_work, SLM_UART_RESPONSE_DELAY);
+		break;
+	case LWM2M_CARRIER_EVENT_LTE_LINK_DOWN:
+		LOG_DBG("LWM2M_CARRIER_EVENT_LTE_LINK_DOWN");
+		slm_util_at_printf("AT+CFUN=4");
```
tomi-font, line 160:
```suggestion
		err = slm_util_at_printf("AT+CFUN=4");
```
or maybe even (and same for the other occurences)
```suggestion
		err = slm_util_at_printf("AT+CFUN=4");
		if (err) {
			err = -1;
		}
```

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -94,12 +94,12 @@ These may be the following:
 
 * ``#XCARRIEREVT: 10,<status>``
 
-  Request to perform an application reboot.
+  Request to perform an application reboot, e.g. AT#XRESET.
```
tomi-font, line 97:
```suggestion
  Request to perform an application reboot, i.e. ``AT#XRESET``.
```

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -31,27 +31,27 @@ These may be the following:
...
 
-  * ``0`` - Request handled successfully.
-  * ``-1`` - Request handling was deferred and must be performed by the application at its earliest convenience.
+  * ``0`` - request handling is fulfilled automatically.
+  * ``-1`` - request handling is deferred and must be fulfilled by the application at its earliest convenience.
 
 * ``#XCARRIEREVT: 3,<status>``
 
-  Request to set the modem to minimum functional mode.
+  Request to set the modem to minimum functional mode, i.e. AT+CFUN=0.
```
tomi-font, line 50:
```suggestion
  Request to set the modem to minimum functional mode, i.e. ``AT+CFUN=0``.
```

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -31,27 +31,27 @@ These may be the following:
...
 
-  * ``0`` - Request handled successfully.
-  * ``-1`` - Request handling was deferred and must be performed by the application at its earliest convenience.
+  * ``0`` - request handling is fulfilled automatically.
+  * ``-1`` - request handling is deferred and must be fulfilled by the application at its earliest convenience.
 
 * ``#XCARRIEREVT: 2,<status>``
 
-  Request to set the modem to flight functional mode.
+  Request to set the modem to flight functional mode, i.e. AT+CFUN=4.
```
tomi-font, line 42:
```suggestion
  Request to set the modem to flight functional mode, i.e. ``AT+CFUN=4``.
```

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -31,27 +31,27 @@ These may be the following:
 
 * ``#XCARRIEREVT: 1,<status>``
 
-  Request to set the modem to full functional mode.
+  Request to set the modem to full functional mode, i.e. AT+CFUN=1.
```
tomi-font, line 34:
```suggestion
  Request to set the modem to full functional mode, i.e. ``AT+CFUN=1``.
```

tomi-font (changes requested):

kacperradoszewski:
I imagine the

kacperradoszewski:
Sorry about that, I've mis-clicked... I imagine the SLM tests will need to be adjusted now that SLM_CARRIER_AUTO_CONTROL is enabled by default @tomi-font @MarkusLassila I can help if needed

junqingzou:
> Sorry about that, I've mis-clicked... I imagine the SLM tests will need to be adjusted now that SLM_CARRIER_AUTO_CONTROL is enabled by default @tomi-font @MarkusLassila I can help if needed

If not wanted I could enable it in NIDD overlay instead. But POWEROFF will still behave differently

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -94,12 +94,12 @@ These may be the following:
 
 * ``#XCARRIEREVT: 10,<status>``
 
-  Request to perform an application reboot.
+  Request to perform an application reboot, e.g. ``AT#XRESET``.
```
tomi-font, line 97:
```suggestion
  Request to perform an application reboot, i.e. ``AT#XRESET``.
```

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -31,27 +31,27 @@ These may be the following:
 
 * ``#XCARRIEREVT: 1,<status>``
 
-  Request to set the modem to full functional mode.
+  Request to set the modem to full functional mode, i.e. ``AT+CFUN=1``.
   ``<status>`` returns two possible values:
 
-  * ``0`` - Request handled successfully.
-  * ``-1`` - Request handling was deferred and must be performed by the application at its earliest convenience.
+  * ``0`` - request handling is fulfilled automatically.
```
tomi-font, line 37: Uppercase R is missing on every line that you modified.

tomi-font (changes requested):
Getting there!

kacperradoszewski:
> > Sorry about that, I've mis-clicked... I imagine the SLM tests will need to be adjusted now that SLM_CARRIER_AUTO_CONTROL is enabled by default @tomi-font @MarkusLassila I can help if needed
> 
> If not wanted I could enable it in NIDD overlay instead. But POWEROFF will still behave differently

I think with the different POWEROFF behaviour, the tests will probably pass but might lead to a reboot without a modem power off first (potentially problematic over several reboots).

kacperradoszewski (approved):

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
junqingzou: `e.g.` means there are other ways than `AT#XRESET`

junqingzou:
This PR have helped in passing SB FOTA over LwM2M/NIDD (1)APP split-DFU (2)MFW DFU and (3) MFW DFU Abort test cases.

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
tomi-font: Like what?

junqingzou: .AT#XCARRIER="reboot"
.nRESET system reset (pin reset)
.ENABLE toggle
.nrfjprog -r (soft reset)
.powercycle

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
tomi-font: This is not resolved.

junqingzou: Ah, just locally, next commit after CI is updated

MarkusLassila (approved):
Looks good.

Created PR for moving the LwM2M carrier tests to nightly. This can be merged when that get's in.

Created a separate PR for adapting the changes to CI and ran this against it. To be merged after this has gone in.

Edit: @junqingzou, LwM2M carrier tests moved to nightly. This should now pass the CI.

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -31,27 +31,27 @@ These may be the following:
 
 * ``#XCARRIEREVT: 1,<status>``
 
-  Request to set the modem to full functional mode.
+  Request to set the modem to full functional mode, i.e. ``AT+CFUN=1``.
```
wiba-nordic, line 34: We don't use "i.e." and "e.g." in the documentation, so this will need to be opened up.
```suggestion
  Request to set the modem to full functional mode, equivalent to the ``AT+CFUN=1` AT command`.
```

Should be applied to all the lines where these are used.

wiba-nordic (commented):
Does this need a changelog entry?

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
junqingzou: Not equivalent but must issue the AT command as mentioned.
Change to another wording?

wiba-nordic: Ah, so it doesn't automatically do it. Then, would this be correct:
```suggestion
  Request to set the modem to full functional mode, such as using the ``AT+CFUN=1`` AT command.
```
Alternatively, if the AT command is basically the only option, we can change the order:
```suggestion
  Request to use the ``AT+CFUN=1`` AT command to set the modem to full functional mode.
```

applications/serial_lte_modem/doc/CARRIER_AT_commands.rst
```diff
@@ -94,12 +94,12 @@ These may be the following:
...
 
-  Request to perform an application reboot.
+  Request to perform an application reboot, for example using the ``AT#XRESET`` AT command.
 
   ``<status>`` returns two possible values:
 
-  * ``0`` - Request handled successfully.
-  * ``-1`` - Request handling was deferred and must be performed by the application at its earliest convenience.
+  * ``0`` - request handling is fulfilled automatically.
+  * ``-1`` - request handling is deferred and must be fulfilled by the application at its earliest convenience.
```
tomi-font, line 102: Still missing uppercase letters here.

tomi-font (changes requested):

tomi-font (approved):

wiba-nordic (approved):

PR #16365, opened 2024-07-09
samples: matter: Used DFUsync module in the DFU over SMP

kkasperczyk-no:
Updated Matter repository to pull the change introducing DFUSync module and use in the DFU over SMP implementation in Matter.

### DIFF
--- a/samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
+++ b/samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
@@ -17,6 +17,7 @@
 #include "dfu/ota/ota_util.h"
 
 #include <platform/CHIPDeviceLayer.h>
+#include <platform/nrfconnect/DFUSync.h>
 
 #include <lib/support/logging/CHIPLogging.h>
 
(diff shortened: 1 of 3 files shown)


samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
```diff
@@ -30,29 +32,36 @@ constexpr uint16_t kAdvertisingIntervalMin = 400;
...
+static uint32_t sDfuSyncMutexId;
+
 namespace
 {
-enum mgmt_cb_return UploadConfirmHandler(uint32_t event,
-					 enum mgmt_cb_return prev_status,
-					 int32_t *rc, uint16_t *group,
-					 bool *abort_more, void *data,
-					 size_t data_size)
+enum mgmt_cb_return UploadConfirmHandler(uint32_t event, enum mgmt_cb_return prev_status, int32_t *rc, uint16_t *group,
```
ArekBalysNordic, line 40: @markaj-nordic  it's yet another place where we can have different clang-format configs.

ArekBalysNordic (approved):
This part looks good.

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
kkasperczyk-no: So is it correct now, or my configuration is wrong?

markaj-nordic: Not sure, maybe the formatting here was invalid in the first place.

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
```diff
@@ -63,6 +72,14 @@ enum mgmt_cb_return CommandHandler(uint32_t event,
 	return MGMT_CB_OK;
 }
 
+enum mgmt_cb_return DfuStoppedHandler(uint32_t event, enum mgmt_cb_return prev_status, int32_t *rc, uint16_t *group,
```
markaj-nordic, line 75: Although our compiler does not complain about it, maybe it's still better to tag all arguments are unused or skip their names.

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
```diff
@@ -130,4 +158,24 @@ void DFUOverSMP::StartServer()
...
 
+void DFUOverSMP::Disconnected(bt_conn *conn, uint8_t reason)
+{
+	if (!sDfuInProgress) {
+		return;
+	}
+
+	bt_conn_info btInfo;
+
+	bt_conn_get_info(conn, &btInfo);
```
markaj-nordic, line 169: This function can return an error on failure, it's worth to consider such a case.

markaj-nordic (commented):
@kkasperczyk-no do you think it requires the release notes update?

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
kkasperczyk-no: Ok, I remove the unused argument names, also for the already existing callbacks.

samples/matter/common/src/dfu/smp/dfu_over_smp.cpp
kkasperczyk-no: Done

kkasperczyk-no:
> @kkasperczyk-no do you think it requires the release notes update?

TBH I don't think so, how about you? It's not a feature, rather a fix for bug existing from the very beginning, but we also didn't have known issue for that.

markaj-nordic:
> > @kkasperczyk-no do you think it requires the release notes update?
> 
> TBH I don't think so, how about you? It's not a feature, rather a fix for bug existing from the very beginning, but we also didn't have known issue for this bug.

Yehh, that's fair. It'd make sense if someone used this "feature" of doing 2 DFUs in parallel which I don't think is much probable :)

markaj-nordic (approved):

PR #16367, opened 2024-07-09
sysbuild: Modify the netcore selection mechanism

dchat-nordic:
This PR introduces changes to selection of network/radio core mechanism.
This is DRAFT, if no big changes are requested I will move it to normal PR and modify remaining samples and applications.

This PR allows to select the netcore explicitly ie. SB_NETCORE_<image name> or implicitly, by just using SB_CONFIG_BT/IEEE802154. The implicit choice will default to ipc_radio.
This PR also improves configuration of ipc_radio.


nordicjm (changes requested):
This is really not an improvement and just makes things far more confusing for users. Something like this is in the plans https://github.com/zephyrproject-rtos/zephyr/issues/70099 and as part of https://github.com/zephyrproject-rtos/zephyr/issues/69546 ` Allow building for a board's basename using Sysbuild` whereby the goal is you specify what a project needs and the image selection is done automatically including on what board targets

dchat-nordic:
@nordicjm Actually this sounds exactly as this PR.
You specify ie. SB_CONFIG_BT and ipc_radio is selected automatically and configured for BLE. But this change would be relevant to sdk-nrf.
The change itself was discussed with @kapi-no and @alstrzebonski , because our current solution is confusing.

nordicjm:
> @nordicjm Actually this sounds exactly as this PR. You specify ie. SB_CONFIG_BT and ipc_radio is selected automatically and configured for BLE. But this change would be relevant to sdk-nrf. The change itself was discussed with @kapi-no and @alstrzebonski , because our current solution is confusing.

I disagree, what is here is completely confusing. The _DEFAULT\_* sysbuild Kconfig options are **for ncs samples only**, customers should not be using them, and really sample applications should not either. The selection here actually has no relation to the application configuration making even less sense, plus you are wrongly restricting what a user can select. It is completely valid to have an application that only supports bluetooth on the nrf5340 and use it with the multiprotocol network image, either because users plan to use it in the future, or because they have multiple variant builds with different features or for any number of reasons whereas in this PR that choice is dead. You are also completely abusing things by adding a debug and release build if people want to use a debug build, users need to do that with e.g. file suffix, not some sysbuild Kconfig. Sysbuild does not nor should it ever set the "build configuration file type" of an application

kapi-no:
> > @nordicjm Actually this sounds exactly as this PR. You specify ie. SB_CONFIG_BT and ipc_radio is selected automatically and configured for BLE. But this change would be relevant to sdk-nrf. The change itself was discussed with @kapi-no and @alstrzebonski , because our current solution is confusing.
> 
> I disagree, what is here is completely confusing. The _DEFAULT_* sysbuild Kconfig options are **for ncs samples only**, customers should not be using them, and really sample applications should not either. The selection here actually has no relation to the application configuration making even less sense, plus you are wrongly restricting what a user can select. It is completely valid to have an application that only supports bluetooth on the nrf5340 and use it with the multiprotocol network image, either because users plan to use it in the future, or because they have multiple variant builds with different features or for any number of reasons whereas in this PR that choice is dead. You are also completely abusing things by adding a debug and release build - if people want to use a debug build, users need to do that with e.g. file suffix, not some sysbuild Kconfig. Sysbuild does not nor should it ever set the "build configuration file type" of an application

How does this solution restrict the use case that you mentioned with the "multiprotocol" network image?

nordicjm:
> > > @nordicjm Actually this sounds exactly as this PR. You specify ie. SB_CONFIG_BT and ipc_radio is selected automatically and configured for BLE. But this change would be relevant to sdk-nrf. The change itself was discussed with @kapi-no and @alstrzebonski , because our current solution is confusing.
> > 
> > 
> > I disagree, what is here is completely confusing. The _DEFAULT_* sysbuild Kconfig options are **for ncs samples only**, customers should not be using them, and really sample applications should not either. The selection here actually has no relation to the application configuration making even less sense, plus you are wrongly restricting what a user can select. It is completely valid to have an application that only supports bluetooth on the nrf5340 and use it with the multiprotocol network image, either because users plan to use it in the future, or because they have multiple variant builds with different features or for any number of reasons whereas in this PR that choice is dead. You are also completely abusing things by adding a debug and release build - if people want to use a debug build, users need to do that with e.g. file suffix, not some sysbuild Kconfig. Sysbuild does not nor should it ever set the "build configuration file type" of an application
> 
> How does this solution restrict the use case that you mentioned with the multiprotocol network image?

Did you check the commits?
```
config SUPPORT_NETCORE_MULTIPROTOCOL_RPMSG
	bool
	default y if (BT && !BT_RPC && IEEE802154)
```

nordicjm:
I'm opposed to the whole concept, this does not help customers in any way

kapi-no:
> > > > @nordicjm Actually this sounds exactly as this PR. You specify ie. SB_CONFIG_BT and ipc_radio is selected automatically and configured for BLE. But this change would be relevant to sdk-nrf. The change itself was discussed with @kapi-no and @alstrzebonski , because our current solution is confusing.
> > > 
> > > 
> > > I disagree, what is here is completely confusing. The _DEFAULT_* sysbuild Kconfig options are **for ncs samples only**, customers should not be using them, and really sample applications should not either. The selection here actually has no relation to the application configuration making even less sense, plus you are wrongly restricting what a user can select. It is completely valid to have an application that only supports bluetooth on the nrf5340 and use it with the multiprotocol network image, either because users plan to use it in the future, or because they have multiple variant builds with different features or for any number of reasons whereas in this PR that choice is dead. You are also completely abusing things by adding a debug and release build - if people want to use a debug build, users need to do that with e.g. file suffix, not some sysbuild Kconfig. Sysbuild does not nor should it ever set the "build configuration file type" of an application
> > 
> > 
> > How does this solution restrict the use case that you mentioned with the multiprotocol network image?
> 
> Did you check the commits?
> 
> ```
> config SUPPORT_NETCORE_MULTIPROTOCOL_RPMSG
> 	bool
> 	default y if (BT && !BT_RPC && IEEE802154)
> ```

Can we simply remove the `SUPPORT_NETCORE_MULTIPROTOCOL_RPMSG` option to remove this restriction or relax it as @dchat-nordic suggested? I don't see how this configuration detail could disqualify the whole approach used in this PR.

The `SB_CONFIG_BT` option is intended to provide abstraction so that we do not have to provide two configuration variants for every Bluetooth sample in NCS: one for single-core devices and one for multi-core devices.

Most customers would probably use the default configuration of the IPC Radio that is taken from some sample project and is the closest to their use case. For more advanced use cases, customers can use the custom configuration of the IPC radio application or even implement their own application for Network Core.

Also, the Kconfig options for sysbuild are optional, customers can as well implement their custom sysbuild CMake policy in their repository.

kapi-no:
@nordicjm, I just want to remind you that currently (and in NCS 2.7.0) single-core devices (e.g. nRF52840 DK) have IPC Radio configuration enabled and confuse users that the IPC Radio image is used in the device runtime.

dchat-nordic:
I have replaced the configuration of ipc_radio with FILE_SUFFIX approach.

dchat-nordic:
@nordicjm I have replaced the ipc_radio variant selection with FILE_SUFFIX. Please take a look.

dchat-nordic:
@wbober could you take a look?

nordicjm:
> @nordicjm, I just want to remind you that currently (and in NCS 2.7.0) single-core devices (e.g. nRF52840 DK) have IPC Radio configuration enabled (`SB_CONFIG_NRF_DEFAULT_IPC_RADIO`). This Kconfig may confuse users and wrongly indicate to them that the IPC Radio image is used in their project. This is a consequence of the following configuration scheme used in NCS sample projects (e.g. in the Fast Pair Locator Tag sample and many other Bluetooth samples) and defined in the sample `Kconfig.sysbuild` file:
> 
> ```
> config NRF_DEFAULT_IPC_RADIO
> 	default y
> ```

It's a default, it doesn't mean that's selected at all, this isn't something specific to ncs, this is how it works in zephyr, how it works in linux, how in works in a lot of projects.

> @nordicjm I have replaced the ipc_radio variant selection with FILE_SUFFIX. Please take a look.

The whole concept here is a no from me, it's confusing, incredibly unhelpful and frustrating and makes no sense

dchat-nordic:
@nordicjm Similar concept was used in child/parent-image builds. The application core config had CONFIG_BT/IEEE/etc. and the child image was selected accordingly without the intricate knowledge of what sample to select and attach.
In my opinion the whole knowledge of samples and configuration is of no use for the customer. As a customer one should just select BT and have a working BT stack.

nordicjm:
> @nordicjm Similar concept was used in child/parent-image builds. The application core config had CONFIG_BT/IEEE/etc. and the child image was selected accordingly without the intricate knowledge of what sample to select and attach. In my opinion the whole knowledge of samples and configuration is of no use for the customer. As a customer one should just select BT and have a working BT stack.

And ask the customers that had to use that system (me years ago when I was one) of how annoying that was if you wanted to add your own image or replace an image with something else - impossible. There's a very good reason sysbuild works the way it does

PR #16378, opened 2024-07-10
applications: nrf_desktop: Add harness to sample.yaml

MarekPieta:
PR adds harness to sample.yaml to introduce a simple validation. The harness is used only by configurations that provide logs over UART. PR also moves part of test configuration under common to avoid duplication.

PR also adds USB next release configuration to sample.yaml to ensure that the configuration builds without problems.

Jira: NCSDK-26996

### DIFF
--- a/applications/nrf_desktop/sample.yaml
+++ b/applications/nrf_desktop/sample.yaml
@@ -201,3 +201,12 @@ tests:
     extra_args: FILE_SUFFIX=keyboard
     extra_configs:
       - CONFIG_DESKTOP_USB_STACK_NEXT=y
+  applications.nrf_desktop.zrelease.usb_next:
+    build_only: true
+    platform_allow:
+      - nrf52840dk/nrf52840
+      - nrf52840dongle/nrf52840
+    integration_platforms:
+      - nrf52840dk/nrf52840
+      - nrf52840dongle/nrf52840
+    extra_args: FILE_SUFFIX=release
--- a/applications/nrf_desktop/sample.yaml
+++ b/applications/nrf_desktop/sample.yaml
@@ -1,15 +1,34 @@
 sample:
   name: nRF Desktop
   description: HID device reference design
+common:
+  sysbuild: true
+  tags: bluetooth ci_build sysbuild
+  harness: console
+  harness_config:
+    type: multi_line
+    ordered: false
+    regex:
+      - "app_event_manager: e:module_state_event module:main state:READY"
+      - "ble_state: Bluetooth initialized"
+      - "settings_loader: Settings loaded"
+      - "ble_bond: Selected BLE peers"
+      - "(ble_adv: Advertising started)|(ble_scan: Scan started)"
+      - "dfu: Secondary image slot is clean"
 tests:
+  applications.nrf_desktop.zdebug.uart:
+    platform_allow:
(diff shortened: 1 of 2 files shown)


zycz (approved):

applications/nrf_desktop/sample.yaml
```diff
@@ -1,15 +1,34 @@
...
+    ordered: false
+    regex:
+      - "app_event_manager: e:module_state_event module:main state:READY"
+      - "ble_state: Bluetooth initialized"
+      - "settings_loader: Settings loaded"
+      - "ble_bond: Selected BLE peers"
+      - "(ble_adv: Advertising started)|(ble_scan: Scan started)"
+      - "dfu: Secondary image slot is clean"
 tests:
+  applications.nrf_desktop.zdebug.uart:
```
kapi-no, line 19: With an intention of adding other DKs to this test option (once we make DKs run by default with UART)

```suggestion
  applications.nrf_desktop.dk.zdebug.uart:
```

applications/nrf_desktop/sample.yaml
```diff
@@ -22,47 +41,35 @@ tests:
...
-    tags: bluetooth ci_build sysbuild
     extra_args: FILE_SUFFIX=mcuboot_qspi
   applications.nrf_desktop.zdebug_mcuboot_smp:
-    sysbuild: true
     build_only: true
     platform_allow: nrf52840dk/nrf52840 nrf52840gmouse/nrf52840
     integration_platforms:
       - nrf52840dk/nrf52840
       - nrf52840gmouse/nrf52840
-    tags: bluetooth ci_build sysbuild
```
kapi-no, line 62: is it okay to remove these tags?

applications/nrf_desktop/sample.yaml
```diff
@@ -22,47 +41,35 @@ tests:
...
   applications.nrf_desktop.zdebug_mcuboot_qspi:
-    sysbuild: true
     build_only: true
     platform_allow: nrf52840dk/nrf52840
     integration_platforms:
       - nrf52840dk/nrf52840
-    tags: bluetooth ci_build sysbuild
     extra_args: FILE_SUFFIX=mcuboot_qspi
   applications.nrf_desktop.zdebug_mcuboot_smp:
-    sysbuild: true
```
kapi-no, line 56: it would be great to apply this change for the whole NCS in one go or at least for our projects

applications/nrf_desktop/sample.yaml
MarekPieta: Tags were moved under common, not removed

applications/nrf_desktop/sample.yaml
MarekPieta: I prefer not to include `dk` infix in the test name. Actually any board that provides logs over UART could be added to this group.

applications/nrf_desktop/sample.yaml
MarekPieta: I could add it to our projects. I will do it under follow-up ticket: https://nordicsemi.atlassian.net/browse/NCSDK-28352. I think we could limit to our samples/apps (at least for now).

kapi-no (approved):

miha-nordic (approved):

PR #16384, opened 2024-07-10
toolchain: Python 3.12

jangalda-nsc:
test-sdk-nrf: python-3.12

### DIFF
--- a/.github/workflows/west-commands.yml
+++ b/.github/workflows/west-commands.yml
@@ -14,27 +14,27 @@ jobs:
     runs-on: ubuntu-latest
     name: Run Python checks for west commands on patch series (PR)
     steps:
-    - name: Checkout the code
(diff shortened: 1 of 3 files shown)


shanthanordic (approved):

bencefr:
Shouldn't the python version bumped on all platforms? Majority of the users are on Windows and macOS, how's the compatibility ensured? These files are input to the toolchain builder as far as I know. We also noticed that kconfiglib might not be compatible with anything that is newer than python 3.9.x.

thst-nordic:
@bencefr, 

@jangalda-nsc tested building the bundle with Mac/Windows before summer break but they failed in the toolchain-bundler.  He reported this to the TCM team and is waiting for it to be resolved.

thst-nordic (approved):

tejlmand (approved):

PR #16405, opened 2024-07-11
suit: Handle duplicate MPIs as icorrect MPI value

tomchy:
In case of duplicate MPIs, it is better to enter failed state than reboot the device.

Ref: NCSDK-28377

### DIFF
--- a/subsys/suit/orchestrator/src/suit_orchestrator_sdfw.c
+++ b/subsys/suit/orchestrator/src/suit_orchestrator_sdfw.c
@@ -384,6 +384,10 @@ int suit_orchestrator_init(void)
 			LOG_ERR("Failed to load MPI: invalid MPI format (i.e. version, values)");
 			plat_err = suit_execution_mode_set(EXECUTION_MODE_FAIL_MPI_INVALID);
 			break;
+		case SUIT_PLAT_ERR_EXISTS:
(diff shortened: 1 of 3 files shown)


robertstypa (approved):

subsys/suit/orchestrator/src/suit_orchestrator_sdfw.c
```diff
@@ -384,6 +384,10 @@ int suit_orchestrator_init(void)
 			LOG_ERR("Failed to load MPI: invalid MPI format (i.e. version, values)");
 			plat_err = suit_execution_mode_set(EXECUTION_MODE_FAIL_MPI_INVALID);
 			break;
+		case SUIT_PLAT_ERR_EXISTS:
```
SylwesterKonczyk, line 387: Probable documentation for suit_plat_err_t suit_storage_init(void) should be amended, especially range of returned error codes.

kszromek-nordic (approved):

subsys/suit/orchestrator/src/suit_orchestrator_sdfw.c
tomchy: Nice catch!
Documentation extended.

ahasztag (approved):

PR #16410, opened 2024-07-11
[WIP] lib: lte_link_control: add Kconfig for ext. features

MirkoCovizzi:
Let some features be enabled via a new Kconfig option to reduce binary size.

For example, this saves around 10KB of flash on the `https_client` sample.


eivindj-nordic:
@MirkoCovizzi I think this is a good improvement for applications that does not need anything in the now extended feature set. Though, what about applications that needs one of the features? With this solution they would need to add all features, increasing the binary size a lot. Would it be an alternative to make it more fine tuned in a way that you can enable some extra features without adding all? Have you checked if there are applications in NCS that would benefit from that?

doc/nrf/libraries/modem/lte_lc.rst
```diff
@@ -19,6 +19,8 @@ Configuration
 *************
 
 To enable the library, set the Kconfig option :kconfig:option:`CONFIG_LTE_LINK_CONTROL` to ``y`` in the project configuration file :file:`prj.conf`.
+By default the library supports only network registration and connection status functionalities.
```
trantanen, line 22: Based on the changes, the default functionality seem to include also PSM (`lte_lc_psm_*()`) and periodic search (`lte_lc_periodic_search_*()`).
However, I think we should think about configuration for these as well. Especially for rarely used periodic search which should be behind configurations.

lib/lte_link_control/Kconfig
```diff
@@ -365,6 +368,12 @@ config LTE_LC_TRACE
 	  what has happened and has no associated payload.
 	  It is intended to be used for debugging purposes.
 
+config LTE_EXTENDED_FEATURES
```
trantanen, line 371: Cannot add comment to modem shell code so adding here.
This Kconfig should be enabled/selected in `samples/cellular/modem_shell/Kconfig` when `CONFIG_MOSH_LINK` is enabled. `CONFIG_LTE_EXTENDED_FEATURES` is enabled in MOSH by default but if you compile it with `CONFIG_MOSH_LOCATION=n`, extended features are disabled.

The places where we might need this flag might be a little harder to spot easily or all lte_lc functions should be checked. E.g., I found `subsys/caf/modules/net_state.c` using `LTE_LC_EVT_EDRX_UPDATE` event.

I'm trying to think if there are other libs/samples that could need similar change. I'm sure these are not the only ones.

trantanen (changes requested):
Great improvement and let's discuss how to adjust this to get it to main!

lib/lte_link_control/Kconfig
```diff
@@ -365,6 +368,12 @@ config LTE_LC_TRACE
...
+	bool "Extended features"
+	help
+	  Enables support for extended features, notifications, and events: eDRX,
+	  Periodic TAU Pre-Warning, Neighboring Cell Measurements, Modem Sleep, Modem Domain Events.
+
```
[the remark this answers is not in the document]
MirkoCovizzi: This is definitely a valid point that I've been considering while implementing this solution and ultimately decided to simplify to not introduce too many new Kconfig options. However, since there is interest in higher granularity I will revisit it. This means that a bigger refactor is required inside the `lte_link_control` library in a way to separate the different functionality domains. I'll mull it over.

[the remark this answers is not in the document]
trantanen: > However, since there is interest in higher granularity I will revisit it. This means that a bigger refactor might be required inside the `lte_link_control` library in a way to separate the different functionality domains across multiple files. Just adding more `ifdefs` across the library will only make it harder to manage in my opinion. I'll mull it over.

I've also had this modularization in my mind especially during the spring time when doing some changes to the library. I wasn't too far from starting it. Time and also the test coverage for the library was a little low for my liking so I didn't do it back then. I was kind of hoping to raise the test coverage and modularize the library later but it could well be done immediately.

But we should consider whether modularization is not with a different PR than Kconfig options which are set to 'n' by default to isolate the breaking changes into a smaller PR.

lib/lte_link_control/Kconfig
tokangas: To avoid this kind of hard to spot problems, why not remove the code returning `-EOPNOTSUPP` when an unsupported function is called? This would cause a build time failure when trying to use "extended features" with the relevant option disabled.

We could even consider flagging affected event IDs in `lte_lc.h` to make sure the application isn't expecting an event which will never come, if events are affected.

PR #16413, opened 2024-07-11
net: lib: nrf_cloud_logging: REST binary logs

plskeggs:
Added encoding and sending of dictionary (binary) logs to nRF Cloud as base64 to the d2c/bin topic. The cloud will convert this to the original binary before processing as it usually does for MQTT binary logs.

Jira: IRIS-7713


subsys/net/lib/nrf_cloud/src/nrf_cloud_log_backend.c
```diff
@@ -361,6 +365,64 @@ static void logger_notify(const struct log_backend *const backend, enum log_back
 	}
 }
 
+static int convert_to_quoted_base64(const uint8_t *ptr, size_t len, uint8_t **optr, size_t *olen)
```
jayteemo, line 368:
```suggestion
static int convert_to_quoted_base64(const uint8_t *input, size_t input_sz, uint8_t **b64_output, size_t *b64_output_sz)
```
or similar

jayteemo (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_log_backend.c
```diff
@@ -394,31 +457,47 @@ static int send_ring_buffer(void)
...
-				}
 				if (err != -EBUSY) {
 					LOG_ERR("Data: %s, len: %zd",
 						(const char *)output.data.ptr, output.data.len);
 				}
 				k_sleep(K_MSEC(100));
 			}
 		} while (err == -EBUSY);
+		if (log_format_current == LOG_OUTPUT_DICT) {
+			nrf_cloud_free((void *)output.data.ptr);
```
glarsennordic, line 498: I think instead of using `output.data.ptr` over and over in these various steps, we should set up separate pointers for the different ways in which this data can get handled.

It's weird to have this conditional `nrf_cloud_free` on a pointer that absolutely should not be freed if we are in non-REST mode

glarsennordic: So, instead of 

```c
output.data.len = ring_buf_get_claim(&log_nrf_cloud_rb, (uint8_t **)&output.data.ptr, stored);
```

I suggest 
```c
uint8_t *log_rb_chunk;
int log_rb_chunk_len = ring_buf_get_claim(&log_nrf_cloud_rb, &log_rb_chunk, stored);
```

Then, inside the REST sections, have

```c
uint8_t *log_b64;
int log_b64_len;
err = convert_to_quoted_base64(log_rb_chunk, log_rb_chunk_len, &log_b64, &log_b64_len);
```

Then during cleanup you can just have

```c
if (log_b64) {
    nrf_cloud_free(log_b64);
}
```

Then, do a conditional population of the output struct, looking something like this

```c
if (REST dict logs) {
	output.data.len = log_b64_len;
	output.data.buf = log_b64;
} else {
	output.data.len = log_rb_chunk_len;
	output.data.buf = log_rb_chunk;
}
```

Exact names of these variables being up to your choice.

I think this would be more readable, and easier to understand/maintain

jayteemo (approved):

glarsennordic (approved):

subsys/net/lib/nrf_cloud/src/nrf_cloud_log_backend.c
```diff
@@ -381,69 +445,96 @@ static int send_ring_buffer(void)
...
-						(const char *)output.data.ptr, output.data.len);
+						(const char *)output_data.ptr, output_data.len);
 				}
 				k_sleep(K_MSEC(100));
 			}
 		} while (err == -EBUSY);
+		if (log_b64_ptr) {
+			nrf_cloud_free(log_b64_ptr);
+			log_b64_ptr = NULL;
+		}
```
glarsennordic, line 511: Possibly move this to after the `cleanup` label, since it can live there with no problem

b-gent (approved):

PR #16415, opened 2024-07-12
applications: nrf_desktop: Support hid_subscriber_event in hid_forward

MarekPieta:
Change adds support for hid_subscriber_event to HID forward module.

Jira: NCSDK-25200


applications/nrf_desktop/src/modules/hid_forward.c
```diff
@@ -1119,21 +1099,19 @@ static void init(void)
...
+			APP_EVENT_SUBMIT(item->report);
 
-		k_free(item);
-
-		sub->busy = true;
+			k_free(item);
+			sub->report_cnt++;
+		} else {
+			break;
+		}
```
kapi-no, line 1114: Consider the following improvement:

```
	while (sub->report_cnt != sub->report_max) {
		struct enqueued_report *item = get_next_enqueued_report(&sub->enqueued_reports);

		if (!item) {
			break;
		}

		APP_EVENT_SUBMIT(item->report);

		k_free(item);
		sub->report_cnt++;
	}
```

kapi-no (approved):

MarekPieta:
> If you think this change warrants a changelog entry, please add it

I will add release note together with doc update that will be introduced in another PR

applications/nrf_desktop/src/modules/hid_forward.c
```diff
@@ -1258,34 +1215,126 @@ static void forward_hid_out_report(const struct hid_report_event *event,
...
 	}
 
-	size_t sub_idx;
+	return sub;
+}
 
-	for (sub_idx = 0; sub_idx < ARRAY_SIZE(subscribers); sub_idx++) {
-		if (subscribers[sub_idx].id == event->source) {
-			break;
+static void send_empty_hid_out_reports(struct hids_peripheral *per, struct subscriber *sub)
```
zycz, line 1232: Overall looks good. Please don't move code: "send_empty_hid_out_reports", "handle_hid_report_event" etc. so that commit looks more clearly.

MarekPieta: I actually wanted to reorganize code a bit before integrating HID report queue utility (to improve readability). Scope of this PR was kept small to ensure we will be able to properly review it. I would prefer to keep functions in new places if possible.

zycz (approved):

PR #16419, opened 2024-07-12
samples: matter: Added support for Matter OTA with SUIT

kkasperczyk-no:
Modified configuration files to allow enabling Matter OTA even if DFU multi image is not used (SUIT case).

### DIFF
--- a/samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf
+++ b/samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf
@@ -5,7 +5,6 @@
 #
 
 # TODO: Workaround to be removed once DFU and external flash will be supported on nRF54H20.
-CONFIG_CHIP_OTA_REQUESTOR=n
 CONFIG_CHIP_QSPI_NOR=n
 
 CONFIG_MPU_STACK_GUARD=n
--- a/samples/matter/template/sample.yaml
+++ b/samples/matter/template/sample.yaml
@@ -73,6 +73,7 @@ tests:
     build_only: true
     extra_args: SB_CONFIG_WIFI_NRF700X=y template_SHIELD=nrf700x_nrf54h20dk
       CONFIG_CHIP_WIFI=y CONFIG_CHIP_BOOTLOADER_NONE=y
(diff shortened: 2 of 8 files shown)


nordicjm (approved):

samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf
```diff
@@ -5,7 +5,6 @@
 #
 
 # TODO: Workaround to be removed once DFU and external flash will be supported on nRF54H20.
-CONFIG_CHIP_OTA_REQUESTOR=n
 CONFIG_CHIP_QSPI_NOR=n
```
ArekBalysNordic, line 8: I'm not sure whether it is needed or not. Maybe we can remove it all because the external flash will be driven by EXMIF. We can do it here, and within the PR for External flash support as well.

samples/matter/template/sample.yaml
```diff
@@ -73,6 +73,7 @@ tests:
     build_only: true
     extra_args: SB_CONFIG_WIFI_NRF700X=y template_SHIELD=nrf700x_nrf54h20dk
       CONFIG_CHIP_WIFI=y CONFIG_CHIP_BOOTLOADER_NONE=y
+      CONFIG_CHIP_OTA_REQUESTOR=n SB_CONFIG_MATTER_OTA=n
```
ArekBalysNordic, line 76: Actually, it should work because it was repaired here: https://github.com/nrfconnect/sdk-nrf/pull/16398/commits/bea1debf0b01e837335f969a7b63c8f382cbfb7d
Could you verify it and possibly remove those configs?

samples/matter/lock/boards/nrf54h20dk_nrf54h20_cpuapp.conf
kkasperczyk-no: Yeah, I assumed it's related to external flash, so I will leave it until we claim support for external flash.

samples/matter/template/sample.yaml
kkasperczyk-no: Discussed f2f, I tried to enable DFU for this build, but it didn't fit flash.

markaj-nordic (approved):

PR #16422, opened 2024-07-12
manifest: sdk-zephyr: pull nrfs ipc send fix

lstnl:
Pull fix to nrfs ipc service send function,
where wrong buffer was used.

Issue
https://nordicsemi.atlassian.net/browse/NRFX-6104


anangl:
Superseded by #16247.

PR #16424, opened 2024-07-12
applications: asset_tracker_v2: add tag for asset_tracker

nordic-piks:
Extend list of paths used to check if asset_tracker application and test should be executed.


nordic-piks:
In this case, there is defined tag (e.g. asset_tracker) and files/folder with source code related to that area.

To gather this list, I analyzed all source files which are compiled when asset_tracker tests and application is build - added them all.

nordic-segl (approved):

maxd-nordic (approved):

nordic-piks:
Only rebased and changed tag asset_tracker_v2 -> applications_asset_tracker_v2

PR #16425, opened 2024-07-12
sysbuild: Add WIP non-partition manager merge functionality

nordicjm:
Adds functionality that allows for sysbuild to merge hex files from images into final output files

NCSDK-28220

- [ ] Support flashing hex file with `west flash`

### DIFF
--- a/sysbuild/CMakeLists.txt
+++ b/sysbuild/CMakeLists.txt
@@ -419,7 +419,7 @@ function(${SYSBUILD_CURRENT_MODULE_NAME}_post_cmake)
     include_provision_hex()
   endif()
 
-  if(SB_CONFIG_MATTER_FACTORY_DATA_GENERATE AND SB_CONFIG_PARTITION_MANAGER)
+  if(SB_CONFIG_MATTER_FACTORY_DATA_GENERATE)
     include(${ZEPHYR_CONNECTEDHOMEIP_MODULE_DIR}/config/nrfconnect/chip-module/generate_factory_data_sysbuild.cmake)
     nrfconnect_generate_factory_data()
   endif()
--- a/west.yml
+++ b/west.yml
@@ -160,7 +160,7 @@ manifest:
     - name: matter
       repo-path: sdk-connectedhomeip
       path: modules/lib/matter
-      revision: c0df90bc9f48360bdcd3d5d2aedde020648ef6b6
+      revision: pull/463/head
       west-commands: scripts/west/west-commands.yml
       submodules:
         - name: nlio


sysbuild/CMakeLists.txt
```diff
@@ -490,6 +490,87 @@ function(${SYSBUILD_CURRENT_MODULE_NAME}_post_cmake)
...
+          list(APPEND ${target_variant}_basic_hex_files ${hex_file})
+          list(APPEND ${target_variant}_basic_targets ${image} ${image}_extra_byproducts)
+        else()
+          list(APPEND basic_hex_files ${hex_file})
+          list(APPEND basic_targets ${image} ${image}_extra_byproducts)
+        endif()
+      endforeach()
+
+      foreach(target ${basic_boards})
+        if(SB_CONFIG_SOC_SERIES_NRF53X AND NOT "${target}" STREQUAL "${default_target_variant}")
```
kkasperczyk-no, line 545: I think this is correct, but just want to make sure. We don't consider other multi-core families here, like nRF54H20, because it performs merging in the custom suit module, right?

nordicjm: It will probably be needed for nrf54h20, but that would be tbd

kkasperczyk-no: Would you like to handle it, as a follow-up of this change or could you consider extending it in this one? TBH from my perspective, we need this functionality only for the nRF54H20 (other families use Partition Manager, so we can use its macros to merge files), so it will not help us much in the current shape.

nordicjm: Will need to wait for them to get back from holiday

kkasperczyk-no: I know that @tomchy is absent, but maybe @e-rk could help us on that?

nordicjm: Have updated and tried it with a sample matter project, it generates the merged file though have not tested on actual hardware (and without UICR so that would need converting)

e-rk:
I don't quite understand the `DOMAIN` and `DEPENDENCIES` parameters. Can `DOMAIN` be set to anything, such as `suit` or should it have some specific value?

nordicjm:
> I don't quite understand the `DOMAIN` and `DEPENDENCIES` parameters. Can `DOMAIN` be set to anything, such as `suit` or should it have some specific value?

`DOMAIN` would be the core, e.g. `cpuapp` or `cpurad`

sysbuild/CMakeLists.txt
kkasperczyk-no: Just to confirm I'm working on that. I've tried this branch and after enabling `CONFIG_CHIP_FACTORY_DATA=y` for the app core I've got bus fault after the application flash. I'm not sure why yet, as I need to figure out how to debug nRF54h20. I will get back with more info when I have something/

kkasperczyk-no: It looks like the root cause is error: Cannot access memory at address 0x174000, where this address is factory data partition, but maybe it's not a matter of generated hex, but partitions configuration.

nordicjm: Maybe secure domain SUIT address allocation to the correct core?

kkasperczyk-no: I have some new information on that. The error code is sort of misleading, because in fact the factory data are not flashed into the memory at all. The reason is workaround that was done btw of introducing sysbuild: https://github.com/nrfconnect/sdk-connectedhomeip/blob/master/config/nrfconnect/chip-module/CMakeLists.txt#L252 so in the end your cmake method is not even called.

Nevertheless, even if we remove this workaround, the nrf54h20 is automatically flashed with the uicr_merged.hex files generated by the SUIT component (logs below):
-- runners.nrfutil: Flashing file: /home/kamilkasperczyk/ncs/nrf/samples/matter/lock/build/lock/zephyr/uicr_merged.hex
-- runners.nrfutil: Flashing file: /home/kamilkasperczyk/ncs/nrf/samples/matter/lock/build/ipc_radio/zephyr/uicr_merged.hex

And the `uicr_merged.hex` is created based on `merged.hex` file from application target binary dir not based on the `merged.hex` from the main build dir that you've used to merge the `factory_data.hex`.

The files input/output for the SUIT is:
Namespace(input_files=['/home/kamilkasperczyk/ncs/nrf/samples/matter/lock/build/lock/zephyr/uicr.hex', '/home/kamilkasperczyk/ncs/nrf/samples/matter/lock/build/lock/zephyr/zephyr.hex'], output=<_io.TextIOWrapper name='/home/kamilkasperczyk/ncs/nrf/samples/matter/lock/build/lock/zephyr/uicr_merged.hex' mode='w' encoding='UTF-8'>, overlap='error')

And for the factory data it's:
Namespace(output=<_io.TextIOWrapper name='/home/kamilkasperczyk/ncs/nrf/samples/matter/lock/build/merged.hex' mode='w' encoding='UTF-8'>, overlap='error', input_files=['/home/kamilkasperczyk/ncs/nrf/samples/matter/lock/build/lock/zephyr/zephyr.hex', '/home/kamilkasperczyk/ncs/nrf/samples/matter/lock/build/lock/zephyr/factory_data.hex'])

kkasperczyk-no: To sum up I think that to make it work for us, the SUIT should use your method to perform merging on the same files, or we could try to merge factory_data.hex with merged.hex in the application target dir. I'm not sure if any of these ideas seems to be correct to you.

kkasperczyk-no: Well it's maybe even more complex, I was wrong and SUIT seems to use `zephyr.hex` not `merged.hex`...

nordicjm: Yes suit needs to be updated, am awaiting them to be back from holiday

PR #16430, opened 2024-07-12
scripts: ci: tags: add tags for sidewalk

nordic-piks:
Sidewalk.


nordic-piks:
In this case, there is defined tag (e.g. Sidewalk) and files/folder with source code related to that area.

nordic-piks:
To create this list, I merged test-spec from sdk-nrf, sdk-zephyr and nrfxlib + sidewalk itself.

nordic-segl (approved):

totyz (approved):
lgtm

miha-nordic (approved):

PR #16432, opened 2024-07-12
samples: matter: Fix SUIT envelope template paths

ArekBalysNordic:
We had the wrong paths set to the common SUIT envelope templates. We need to use Sysbuild to setup the default paths to envelopes for application and radio cores instead of setting them within prj.conf files.

- Added the SB_CONFIG_MATTER_USE_COMMON_SUIT_ENVELOPE_TEMPLATES kconfig file to automatically set envelope template paths to the samples/matter/common/suit_templates directory.
- Removed paths from boards' prj.conf files.
- Enabled common SUIT envelope templates for Matter Template and Lock samples.


ArekBalysNordic:
@nordicjm As followup to https://github.com/nrfconnect/sdk-nrf/pull/16214, I found that we cannot use `${ZEPHYR_NRF_MODULE_DIR}` variable in the `prj.conf` files because it points to the sample directory instead of `ncs/nrf` directory. We can do it in Kconfig, but it doesn't allow us to define that path for the ipc_radio image, due to that the only way to set the default paths is to add the additional sysbuild kconfig and set paths in the sysbuild's CMakeLists.txt file as I did in this PR. If users want to use their own envelopes, they must set `SB_CONFIG_MATTER_USE_COMMON_SUIT_ENVELOPE_TEMPLATES` to `n` and provide their own paths.

What do you think? Can it be done this way?

nordicjm (approved):

nordicjm:
> @nordicjm As followup to #16214, I found that we cannot use `${ZEPHYR_NRF_MODULE_DIR}` variable in the `prj.conf` files because it points to the sample directory instead of `ncs/nrf` directory. We can do it in Kconfig, but it doesn't allow us to define that path for the ipc_radio image, due to that the only way to set the default paths is to add the additional sysbuild kconfig and set paths in the sysbuild's CMakeLists.txt file as I did in this PR. If users want to use their own envelopes, they must set `SB_CONFIG_MATTER_USE_COMMON_SUIT_ENVELOPE_TEMPLATES` to `n` and provide their own paths.
> 
> What do you think? Can it be done this way?

Have you got a minimal code sample? It should not point there

ArekBalysNordic:
> > @nordicjm As followup to #16214, I found that we cannot use `${ZEPHYR_NRF_MODULE_DIR}` variable in the `prj.conf` files because it points to the sample directory instead of `ncs/nrf` directory. We can do it in Kconfig, but it doesn't allow us to define that path for the ipc_radio image, due to that the only way to set the default paths is to add the additional sysbuild kconfig and set paths in the sysbuild's CMakeLists.txt file as I did in this PR. If users want to use their own envelopes, they must set `SB_CONFIG_MATTER_USE_COMMON_SUIT_ENVELOPE_TEMPLATES` to `n` and provide their own paths.
> > What do you think? Can it be done this way?
> 
> Have you got a minimal code sample? It should not point there

You can see it while building the `smp_transfer` https://github.com/nrfconnect/sdk-nrf/tree/main/samples/suit/smp_transfer and set for example: `CONFIG_SUIT_ENVELOPE_TEMPLATE="${ZEPHYR_NRF_MODULE_DIR}/../common/suit_templates/app_envelope.yaml.jinja2"` in the `prj.conf` file. Compilation will fail on generating SUIT envelope (build for nrf54h20), but there is a path visible:

image

`west build -b nrf54h20dk/nrf54h20/cpuapp`

As you can see it points to the sample's directory, and you also can see the .config file:
`CONFIG_SUIT_ENVELOPE_TEMPLATE="${ZEPHYR_NRF_MODULE_DIR}/../common/suit_templates/app_envelope.yaml.jinja2"`.

Moreover, if you build a sample with this config as a CMake argument, for example: `west build -b nrf54h20dk/nrf54h20/cpuapp -- -DCONFIG_SUIT_ENVELOPE_TEMPLATE=\"${ZEPHYR_NRF_MODULE_DIR}/../common/suit_templates/app_envelope.yaml.jinja2\" it will be empty:

image

I'm not sure whether it works properly or not, maybe we shouldn't use CMake variables in prj.conf file, so your input here will be much appreciated.

kkasperczyk-no (approved):

nordicjm:
It's not working because you never do anything in suit.cmake to replace the tokens and so it uses them whole e.g.
```
diff --git a/cmake/sysbuild/suit.cmake b/cmake/sysbuild/suit.cmake
index 97bf28136..6996fe72c 100644
--- a/cmake/sysbuild/suit.cmake
+++ b/cmake/sysbuild/suit.cmake
@@ -154,7 +154,9 @@ function(suit_create_package)
     endif()
 
     sysbuild_get(INPUT_ENVELOPE_JINJA_FILE IMAGE ${image} VAR CONFIG_SUIT_ENVELOPE_TEMPLATE KCONFIG)
+     string(CONFIGURE "${INPUT_ENVELOPE_JINJA_FILE}" INPUT_ENVELOPE_JINJA_FILE)
     suit_set_absolute_or_relative_path(${INPUT_ENVELOPE_JINJA_FILE} ${app_config_dir} INPUT_ENVELOPE_JINJA_FILE)
     sysbuild_get(target IMAGE ${image} VAR CONFIG_SUIT_ENVELOPE_TARGET KCONFIG)
     sysbuild_get(BINARY_DIR IMAGE ${image} VAR APPLICATION_BINARY_DIR CACHE)
     sysbuild_get(BINARY_FILE IMAGE ${image} VAR CONFIG_KERNEL_BIN_NAME KCONFIG)

```

ArekBalysNordic:
> It's not working because you never do anything in suit.cmake to replace the tokens and so it uses them whole e.g.
> 
> ```
> diff --git a/cmake/sysbuild/suit.cmake b/cmake/sysbuild/suit.cmake
> index 97bf28136..6996fe72c 100644
> --- a/cmake/sysbuild/suit.cmake
> +++ b/cmake/sysbuild/suit.cmake
> @@ -154,7 +154,9 @@ function(suit_create_package)
>      endif()
>  
>      sysbuild_get(INPUT_ENVELOPE_JINJA_FILE IMAGE ${image} VAR CONFIG_SUIT_ENVELOPE_TEMPLATE KCONFIG)
> +     string(CONFIGURE "${INPUT_ENVELOPE_JINJA_FILE}" INPUT_ENVELOPE_JINJA_FILE)
>      suit_set_absolute_or_relative_path(${INPUT_ENVELOPE_JINJA_FILE} ${app_config_dir} INPUT_ENVELOPE_JINJA_FILE)
>      sysbuild_get(target IMAGE ${image} VAR CONFIG_SUIT_ENVELOPE_TARGET KCONFIG)
>      sysbuild_get(BINARY_DIR IMAGE ${image} VAR APPLICATION_BINARY_DIR CACHE)
>      sysbuild_get(BINARY_FILE IMAGE ${image} VAR CONFIG_KERNEL_BIN_NAME KCONFIG)
> ```

I see, thanks for investigating it. I talked with a person from the SUIT team and we decided that I would add this line here. It should be useful, and then I can avoid adding the new `SB_CONFIG_MATTER_USE_COMMON_SUIT_ENVELOPE_TEMPLATES` config.

ArekBalysNordic:
@e-rk @ahasztag I've modified suit.cmake file, so please review.

@nordicjm I provided a fix/workaround that you suggested and I can see that we can use the CMake variables from prj.conf file. Please verify whether they are now ok.

e-rk:
Could you add the same change to line 185? 
```
  if(DEFINED INPUT_ROOT_ENVELOPE_JINJA_FILE AND NOT INPUT_ROOT_ENVELOPE_JINJA_FILE STREQUAL "")
    set(ROOT_NAME ${SB_CONFIG_SUIT_ENVELOPE_ROOT_ARTIFACT_NAME})
+   string(CONFIGURE "${INPUT_ROOT_ENVELOPE_JINJA_FILE}" INPUT_ROOT_ENVELOPE_JINJA_FILE)
    suit_set_absolute_or_relative_path(${INPUT_ROOT_ENVELOPE_JINJA_FILE} ${app_config_dir} INPUT_ROOT_ENVELOPE_JINJA_FILE)
```

e-rk (approved):

nordicjm (approved):

PR #16440, opened 2024-07-15
bluetooth: Always build hci_vs_sdc if CONFIG_BT is selected

alexstanoev-nordic:
When using a remote HCI image, the LL KConfig is not available. Match the approach used by host_extensions and always build the VS command wrappers if CONFIG_BT is set.
The remote LL can return unsupported BTLE command if it's not SDC.


rugeGerritsen (approved):

ppryga-nordic (approved):
LGTM

KAGA164 (approved):

PR #16441, opened 2024-07-15
v2.5: wifi: Add support for configuring buffered data retrieval mode when in PS

srinivas1201:
By default PS-poll is used to retrieve buffered packets while staying in power save, but alternatively we can exit power save by sending QoS Null frame with PM=0 and once retrieval is finished, then we can send QoS Null frame with PM=1 to enter power save.

This adds a Kconfig option to choose between the two modes.

Implements: SHEL-2947, SHEL-2918 and SHEL-2919


drivers/wifi/nrf700x/osal/fw_if/umac_if/inc/fmac_cmd.h
```diff
@@ -38,7 +38,11 @@ enum nrf_wifi_status umac_cmd_init(struct nrf_wifi_fmac_dev_ctx *fmac_dev_ctx,
 				   unsigned int phy_calib,
 				   enum op_band op_band,
 				   bool beamforming,
-				   struct nrf_wifi_tx_pwr_ctrl_params *tx_pwr_ctrl_params);
+				   struct nrf_wifi_tx_pwr_ctrl_params *tx_pwr_ctrl_params,
+#ifndef CONFIG_NRF700X_RADIO_TEST
+				   bool qos_nullframe_ps
+#endif /* !CONFIG_NRF700X_RADIO_TEST */
+				   );
```
krish2718, line 45: Test, will delete later.

drivers/wifi/nrf700x/Kconfig
```diff
@@ -403,3 +403,6 @@ config NRF_WIFI_AP_DEAD_DETECT_TIMEOUT
 	help
 		The number of seconds after which AP is declared dead if no beacons
 		are received from the AP. Used to detect AP silently going down e.g., power off.
+
+config NRF_WIFI_QOS_NULLFRAME_PS
+	bool "Enable Wi-Fi QoS null frame based PS. RPU uses QoS null frames to retrieve buffered frames from the AP."
```
krish2718, line 408: may be also add, what's the approach that is used when this is disabled?

drivers/wifi/nrf700x/osal/fw_if/umac_if/src/cmd.c
```diff
@@ -193,6 +193,10 @@ enum nrf_wifi_status umac_cmd_init(struct nrf_wifi_fmac_dev_ctx *fmac_dev_ctx,
 		umac_cmd_data->disable_beamforming = 1;
 	}
 
+#ifdef CONFIG_NRF_WIFI_QOS_NULLFRAME_PS
+	umac_cmd_data->nullframe_pwrsave = 1;
+#endif /* CONFIG_NRF_WIFI_QOS_NULLFRAME_PS */
+
```
krish2718, line 199: as this needs to be sent always, it might be better to do:
```suggestion
	umac_cmd_data->nullframe_pwrsave = IS_ENABLED(CONFIG_NRF_WIFI_QOS_NULLFRAME_PS);

```

drivers/wifi/nrf700x/Kconfig
krish2718: Also, better to wrap even if compliance is passing.

krish2718 (changes requested):

drivers/wifi/nrf700x/Kconfig
```diff
@@ -403,3 +403,9 @@ config NRF_WIFI_AP_DEAD_DETECT_TIMEOUT
 	help
 		The number of seconds after which AP is declared dead if no beacons
 		are received from the AP. Used to detect AP silently going down e.g., power off.
+
+config NRF_WIFI_PS_DATA_RETRIEVAL_MODE
+	bool "Use only QoS null frame to retrieve buffered data from AP."
```
krish2718, line 408:
```suggestion
	bool "Use either PS-poll or  QoS null mechanism to retrieve buffered data from AP"
```

drivers/wifi/nrf700x/Kconfig
```diff
@@ -403,3 +403,9 @@ config NRF_WIFI_AP_DEAD_DETECT_TIMEOUT
...
+	help
+	    By default RPU uses both QoS null frame and PS-Poll to retrieve buffered frames from the AP.
+	    Enable it to use only QoS null frames for retrieving buffered frames from the AP.
```
krish2718, line 411:
```suggestion
	    By default nRF70 uses PS-Poll mechanism to retrieve buffered frames from the AP.
	    Enable this to use only QoS null frame based mechanism for retrieving buffered frames from the AP.
```

drivers/wifi/nrf700x/Kconfig
```diff
@@ -403,3 +403,12 @@ config NRF_WIFI_AP_DEAD_DETECT_TIMEOUT
 	help
 		The number of seconds after which AP is declared dead if no beacons
 		are received from the AP. Used to detect AP silently going down e.g., power off.
+
+config NRF_WIFI_PS_DATA_RETRIEVAL_MODE
```
krish2718, line 407: As discussed please make it a choice as they are independent options.

udaynordic (approved):

PR #16445, opened 2024-07-15
mqtt_helper: Prevent mqtt_helper_thread hang on mqtt_live fail

ball-hayden:
If mqtt_live fails with an error other than -EAGAIN (for example -ENOTCONN), the poll loop exits.

connection_poll_sem is still held, and no attempt is made to abort the connection, so it is not possible for the loop to recover.

Instead, we call mqtt_abort on any occasion that we break out of the loop.

This means that a disconnect event occurs, allowing us to gracefully recover.

### DIFF
--- a/subsys/net/lib/mqtt_helper/mqtt_helper.c
+++ b/subsys/net/lib/mqtt_helper/mqtt_helper.c
@@ -754,7 +754,6 @@ MQTT_HELPER_STATIC void mqtt_helper_poll_loop(void)
 			ret = mqtt_input(&mqtt_client);
 			if (ret) {
 				LOG_ERR("Cloud MQTT input error: %d", ret);
-				(void)mqtt_abort(&mqtt_client);
 				break;
 			}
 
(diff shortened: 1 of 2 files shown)


sigvartmh (approved):

ball-hayden:
@jhn-nordic / @simensrostad any chance you might be able to have a look over this, please?

simensrostad (approved):
LGTM, updating the unit test would be appreciated

ball-hayden:
Thanks @simensrostad.

I tried following the guide at https://docs.nordicsemi.com/bundle/ncs-latest/page/zephyr/contribute/guidelines.html#continuous-integration, but hit issues building with twister (probably because `zephyr-env.sh` doesn't exist at the top level of the `sdk-nrf` repo).

Are you able to point me at some documentation for how to run the twister tests locally?
(or could you enable CI runs, so I can see them run there?)

simensrostad:
> Thanks @simensrostad.
> 
> I tried following the guide at https://docs.nordicsemi.com/bundle/ncs-latest/page/zephyr/contribute/guidelines.html#continuous-integration, but hit issues building with twister (probably because `zephyr-env.sh` doesn't exist at the top level of the `sdk-nrf` repo).
> 
> Are you able to point me at some documentation for how to run the twister tests locally? (or could you permit CI runs, so I can see them run there?)

@rlubos

rlubos:
I've added a lablel to the PR so that the CI tests run for you here.

There's not much philosophy in running twister, a simple `west twister -T tests/subsys/net/lib/mqtt_helper` from the sdk-nrf root should work, if it doesn't it's likely some environmental setup issue on your side. The test suites are regular apps after all, so if you're able to build sample you should also be able to build `tests/subsys/net/lib/mqtt_helper` test suite, for `native_sim` for instance.

ball-hayden:
Urgh. Sorry. Bad push.
I had to shift the commit around a bit to get to a working test environment.

Should all be good again now.

(Aside, I don't get to see Jenkins output unfortunately)

rlubos (approved):

ball-hayden:
Thanks for your help @rlubos.

PR #16454, opened 2024-07-16
applications: protocols_serialization: fix "ot state" command

Damian-Nordic:
The "ot state <role>" command cannot be procesed by the RPC client because APIs like otThreadBecomeRouter() are not serializeable yet. Send such commands to be processed by the RPC server to keep the existing functionality.

### DIFF
--- a/applications/protocols_serialization/client/src/ot_shell.c
+++ b/applications/protocols_serialization/client/src/ot_shell.c
@@ -38,16 +38,18 @@ static void ot_cli_lazy_init(const struct shell *sh)
 typedef otError(*ot_cli_command_handler_t)(const struct shell *, size_t argc, char *argv[]);
 
 /*
- * Adapter function for processing OpenThread shell commands, which is convenient to call from
- * a Zephyr shell command handler. It is meant to duplicate the behavior of the internal OT CLI
- * component with regards to printing the command's result.
+ * Invokes the OT shell command handler with the provided arguments. Then prints either
+ * "Done", if the handler returns no error, or an OpenThread error code that the handler
+ * returns.
+ *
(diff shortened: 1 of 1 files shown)


applications/protocols_serialization/client/src/ot_shell.c
```diff
@@ -59,6 +61,46 @@ static int ot_cli_command_invoke(ot_cli_command_handler_t handler, const struct
...
+/*
+ * Sends the command to the RCP server as a line of text. The command is then processed by
+ * the RPC server and its result is returned asynchronously using OT CLI output callback.
+ *
+ * This is a fallback function for commands that cannot be processed by the RPC client
+ * because serialization of the underlying OpenThread APIs is not yet implemented.
+ */
+static int ot_cli_command_send(const struct shell *sh, size_t argc, char *argv[])
+{
+	char buffer[256];
```
canisLupus1313, line 73: the buffer size in general is fine for now however for dataset related commands it won't be sufficient. 

https://github.com/openthread/openthread/blob/main/examples/apps/cli/cli_uart.cpp#L69

canisLupus1313 (approved):

applications/protocols_serialization/client/src/ot_shell.c
Damian-Nordic: The maximum dataset size is 254, so most of them will fit :). And bigger than that won't be possible to input because of `CONFIG_SHELL_CMD_BUFF_SIZE` that is by default 256.

Damian-Nordic: Replaced 256 with CONFIG_SHELL_CMD_BUFF_SIZE to make sure this is never too small.

PR #16455, opened 2024-07-16
Bluetooth: Mesh: add VERSION files to mesh samples

KyraLengfeld:
This commit adds VERSION files to the samples that were missing it, as well as setting the prj config CONFIG_NCS_APPLICATION_BOOT_BANNER_STRING accordingly.

Currently the boot banner is not printed when using CONFIG_BT_MESH_SHELL or CONFIG_SHELL, which affects the chat and sensor_client samples. This will be fixed at a later point.


alxelax (commented):
I see the version file is the same for all samples. I'm afraid having such approach causes updating all files every release preparation and keep in sync all of them.
Furthermore, I suggest having the single file in common place and copy it to the sample folder during compilation. It is possible to use already implemented Zephyr's functionality for this: https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L2865-L2899

nordicjm (approved):

KyraLengfeld:
> I see the version file is the same for all samples. I'm afraid having such approach causes updating all files every release preparation and keep in sync all of them. Furthermore, I suggest having the single file in common place and copy it to the sample folder during compilation. It is possible to use already implemented Zephyr's functionality for this: https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L2865-L2899

(Please correct me if any of the following is wrong but) this proves to be difficult, as the configs from prj.conf, including `CONFIG_NCS_APPLICATION_BOOT_BANNER_STRING` are already merged with the other configs and read during `find_package(Zephyr)` -step of the CMakeList file, yet without the zephyr package, I cannot use zephyr_file_copy. (zephyr_file_copy itself works, and copies the file correctly btw. just, too late...)
Not sure how I can solve this dilemma? 
The only other places I've found uses zephyr_file_copy is for cache, so a different situation there.

nordicjm:
> I see the version file is the same for all samples. I'm afraid having such approach causes updating all files every release preparation and keep in sync all of them. Furthermore, I suggest having the single file in common place and copy it to the sample folder during compilation. It is possible to use already implemented Zephyr's functionality for this: https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L2865-L2899

This is incorrect

alxelax:
> > I see the version file is the same for all samples. I'm afraid having such approach causes updating all files every release preparation and keep in sync all of them. Furthermore, I suggest having the single file in common place and copy it to the sample folder during compilation. It is possible to use already implemented Zephyr's functionality for this: https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L2865-L2899
> 
> This is incorrect

Hi @nordicjm, could you clarify what you mean?

nordicjm:
> > > I see the version file is the same for all samples. I'm afraid having such approach causes updating all files every release preparation and keep in sync all of them. Furthermore, I suggest having the single file in common place and copy it to the sample folder during compilation. It is possible to use already implemented Zephyr's functionality for this: https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L2865-L2899
> > 
> > 
> > This is incorrect
> 
> Hi @nordicjm, could you clarify what you mean?

That function will not work, the VERSION file needs to exist in the application config dir, it cannot be copied during a build and even if it could (which would require running cmake code after some of the base zephyr code has ran but not other parts), would not be used anyway as the build directory is not the application config directory

alxelax (approved):

alxelax:
> > > > I see the version file is the same for all samples. I'm afraid having such approach causes updating all files every release preparation and keep in sync all of them. Furthermore, I suggest having the single file in common place and copy it to the sample folder during compilation. It is possible to use already implemented Zephyr's functionality for this: https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L2865-L2899
> > > 
> > > 
> > > This is incorrect
> > 
> > 
> > Hi @nordicjm, could you clarify what you mean?
> 
> That function will not work, the VERSION file needs to exist in the application config dir, it cannot be copied during a build and even if it could (which would require running cmake code after some of the base zephyr code has ran but not other parts), would not be used anyway as the build directory is not the application config directory

Seems the current versioning system just causes a lot of useless identical files. This problem of course is not of part of this PR.

nordicjm:
> > > > > I see the version file is the same for all samples. I'm afraid having such approach causes updating all files every release preparation and keep in sync all of them. Furthermore, I suggest having the single file in common place and copy it to the sample folder during compilation. It is possible to use already implemented Zephyr's functionality for this: https://github.com/zephyrproject-rtos/zephyr/blob/main/cmake/modules/extensions.cmake#L2865-L2899
> > > > 
> > > > 
> > > > This is incorrect
> > > 
> > > 
> > > Hi @nordicjm, could you clarify what you mean?
> > 
> > 
> > That function will not work, the VERSION file needs to exist in the application config dir, it cannot be copied during a build and even if it could (which would require running cmake code after some of the base zephyr code has ran but not other parts), would not be used anyway as the build directory is not the application config directory
> 
> Seems the current versioning system just causes a lot of useless identical files. This problem of course is not of part of this PR.

Agreed. My proposed versioning system was fully Kconfig driven, and I could see from a mile off the pitfalls this method had and asked people to review them which no-one did, so that system ended up being merged with a huge maintenance burden at release time.

HaavardRei (approved):

PR #16462, opened 2024-07-16
samples: keys: add tags for samples/keys

nordic-piks:
Extend tags.


nordic-segl (approved):

nordic-piks:

miha-nordic (approved):

nordic-piks:
Just rebased.

PR #16466, opened 2024-07-16
samples: edge_impulse: add tags for samples/edge_impulse

nordic-piks:
Extend tags.

### DIFF
--- a/samples/edge_impulse/wrapper/sample.yaml
+++ b/samples/edge_impulse/wrapper/sample.yaml
@@ -32,4 +32,4 @@ tests:
   sample.edge_impulse.wrapper:
     sysbuild: true
     build_only: false
-    tags: sysbuild
+    tags: sysbuild sample_edge_impulse
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -597,3 +597,11 @@ applications_asset_tracker_v2:
     - zephyr/subsys/tracing/
     - zephyr/subsys/random/
     - zephyr/subsys/fs/
+
+sample_edge_impulse:
+  files:
+    - nrf/drivers/sensor/sensor_sim/
+    - nrf/lib/edge_impulse/
+    - nrf/lib/wave_gen/
+    - nrf/samples/edge_impulse/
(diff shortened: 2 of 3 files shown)


nordic-segl (approved):

nordic-piks:

miha-nordic (approved):

samples/edge_impulse/wrapper/sample.yaml
```diff
@@ -32,4 +32,4 @@ tests:
   sample.edge_impulse.wrapper:
     sysbuild: true
     build_only: false
-    tags: sysbuild
+    tags: sysbuild edge_impulse
```
MarekPieta, line 35: Should we add this tag to machine learning application?

scripts/ci/tags.yaml
```diff
@@ -477,3 +477,20 @@ cellular:
     - modules/lib/cjson/
     - modules/lib/hostap/
     - modules/lib/zcbor/
+
+edge_impulse:
+  files:
+    - modules/crypto/mbedtls/
```
MarekPieta, line 483: Edge Impulse by itself does not depend on many of these files (similarly to Event Manager).

MarekPieta (changes requested):

samples/edge_impulse/wrapper/sample.yaml
nordic-piks: I want to take is separately - it is different folder.
I do not want to spread tag across folders tree.
Currently I add tags based on folders rather than looking at sample/tests which has similar dependencies.

scripts/ci/tags.yaml
nordic-piks: I did compilation of all samples and configuration from this sample and those crypto files are compiled.
However, list of decencies listed for tag can be shorter than list of compiled modules - I already did that by removing whole kernel from zephyr.
So this dependency exist but I can remove it if is not important, what do you think?

MarekPieta: Discussed offline. Dependencies related to boards were removed.

MarekPieta (approved):

PR #16467, opened 2024-07-16
samples: app_event_manager: add tags for samples/app_event_manager

nordic-piks:
Extend tags.

### DIFF
--- a/samples/app_event_manager/sample.yaml
+++ b/samples/app_event_manager/sample.yaml
@@ -34,7 +34,7 @@ tests:
       - nrf9160dk/nrf9160/ns
       - nrf21540dk/nrf52840
       - nrf54h20dk/nrf54h20/cpuapp
-    tags: sysbuild
+    tags: sysbuild ci_samples_app_event_manager
   sample.app_event_manager_shell:
     sysbuild: true
     build_only: false
@@ -59,7 +59,7 @@ tests:
     extra_configs:
       - CONFIG_SHELL=y
     platform_exclude: native_posix
-    tags: sysbuild
+    tags: sysbuild ci_samples_app_event_manager
   sample.app_event_manager_shell.native_posix:
     sysbuild: true
     build_only: false
@@ -67,4 +67,4 @@ tests:
       - CONFIG_SHELL=y
       - CONFIG_NATIVE_UART_0_ON_STDINOUT=y
     platform_allow: native_posix
-    tags: sysbuild
+    tags: sysbuild ci_samples_app_event_manager
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -632,3 +632,15 @@ ci_tests_subsys_bootloader:
     - nrfxlib/crypto/
     - nrfxlib/nrf_modem/
     - zephyr/soc/nordic/
+
+ci_samples_app_event_manager:
+  files:
+    - nrf/samples/app_event_manager/
+    - nrf/subsys/app_event_manager/
+    - nrf/include/app_event_manager.h
+
+ci_tests_subsys_app_event_manager:
+  files:
+    - nrf/tests/subsys/app_event_manager/
+    - nrf/subsys/app_event_manager/
+    - nrf/include/app_event_manager.h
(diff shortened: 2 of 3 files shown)


nordic-segl (approved):

nordic-piks:

miha-nordic (approved):

miha-nordic:
@nordic-piks  it is worth to check if scope covers also event_manager_ proxy sample and app_event_manager_profiler_tracer sample. Maybe it can be concatenate in one tag.

scripts/ci/tags.yaml
```diff
@@ -477,3 +477,21 @@ cellular:
     - modules/lib/cjson/
     - modules/lib/hostap/
     - modules/lib/zcbor/
+
+app_event_manager:
+  files:
+    - modules/crypto/mbedtls/
```
MarekPieta, line 483: Event Manager by itself does not use a lot of these dependencies:

- `modules/crypto/mbedtls/`
- `modules/crypto/oberon-psa-crypto/`
- `nrf/drivers/hw_cc3xx/`
- `nrf/lib/modem_antenna/`
- `nrf/lib/nrf_modem_lib/`
- `nrf/subsys/nrf_security/`
- `nrfxlib/crypto/`
- `nrfxlib/nrf_modem/`
- `zephyr/soc/ti/lm3s6965/`
- `zephyr/subsys/ipc/`

Are you trying to cover dependencies of apps that use Event Manager (they seem not tagged now)? Maybe we could cover them with another tag to avoid confusion.

samples/app_event_manager/sample.yaml
```diff
@@ -59,12 +59,12 @@ tests:
     extra_configs:
       - CONFIG_SHELL=y
     platform_exclude: native_posix
-    tags: sysbuild
+    tags: sysbuild app_event_manager
```
MarekPieta, line 62: Should we also tag `tests/subsys/app_event_manager`? Should we tag applications that use Event Manager?

scripts/ci/tags.yaml
MarekPieta: On the other hand we may want to add dependency on Nordic boards from both `nrf` and `zephyr`. We could also consider adding dependency on Zephyr kernel features

MarekPieta (changes requested):

nordic-piks:
> @nordic-piks it is worth to check if scope covers also event_manager_ proxy sample and app_event_manager_profiler_tracer sample. Maybe it can be concatenate in one tag.

I am checking samples/tests with order, starting from ones which has a lot of configurations/targets.
The app_event_manager_profiler_tracer and event_manager_ proxy has very little targets, I did not plan to add them for tagging in this attempt.

samples/app_event_manager/sample.yaml
nordic-piks: tests/subsys/app_event_manager - good point, I will add.

> Should we tag applications that use Event Manager?
No, the approach is different.
Tags are used to exclude testing not to find something to be tested.

scripts/ci/tags.yaml
MarekPieta: Discussed offline and aligned list of dependencies together

samples/app_event_manager/sample.yaml
MarekPieta: Discussed offline and aligned approach.

samples/app_event_manager/sample.yaml
```diff
@@ -34,7 +34,7 @@ tests:
       - nrf9160dk/nrf9160/ns
       - nrf21540dk/nrf52840
       - nrf54h20dk/nrf54h20/cpuapp
-    tags: sysbuild
+    tags: sysbuild sample_app_event_manager
```
MarekPieta, line 37: btw. Consider placing tags under `common` to simplify maintenance (also for test)

scripts/ci/tags.yaml
```diff
@@ -546,3 +546,13 @@ esb:
     - nrf/subsys/mpsl/
     - nrfxlib/crypto/
     - zephyr/soc/nordic/
+
+sample_app_event_manager:
+  files:
+    - nrf/samples/app_event_manager/
+    - nrf/subsys/app_event_manager/
```
MarekPieta, line 553: Consider adding public and private header of app event manager (also for test). Private header contains some implementation.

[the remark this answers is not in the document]
nordic-piks: added.

PR #16469, opened 2024-07-16
tests: subsys: bluetooth: fast_pair: add fast_pair tag

nordic-piks:
Extend tags.

### DIFF
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -64,6 +64,13 @@ bluetooth:
     - nrf/tests/subsys/bluetooth/
     - nrf/applications/nrf_desktop/
     - nrf/samples/nrf5340/multiprotocol_rpmsg/
+    - modules/crypto/mbedtls/
+    - modules/crypto/oberon-psa-crypto/
+    - modules/crypto/tinycrypt/
+    - nrf/subsys/nrf_security/
+    - nrfxlib/crypto/
+    - zephyr/soc/nordic/
+    - zephyr/subsys/settings/
 
 net:
   files:
--- a/tests/subsys/bluetooth/fast_pair/crypto/testcase.yaml
+++ b/tests/subsys/bluetooth/fast_pair/crypto/testcase.yaml
@@ -10,7 +10,7 @@ tests:
       - nrf52840dk/nrf52840
       - nrf5340dk/nrf5340/cpuapp
     extra_args: FILE_SUFFIX=mbedtls
-    tags: sysbuild
+    tags: sysbuild bluetooth
   fast_pair.crypto.oberon:
     sysbuild: true
     platform_allow:
@@ -25,7 +25,7 @@ tests:
       - nrf5340dk/nrf5340/cpuapp
       - nrf5340dk/nrf5340/cpuapp/ns
       - nrf54l15pdk/nrf54l15/cpuapp
-    tags: sysbuild
+    tags: sysbuild bluetooth
   fast_pair.crypto.psa:
     sysbuild: true
     platform_allow:
@@ -43,7 +43,7 @@ tests:
       - nrf54l15pdk/nrf54l15/cpuapp
       - nrf54l15pdk/nrf54l15/cpuapp/ns
     extra_args: FILE_SUFFIX=psa
-    tags: sysbuild
+    tags: sysbuild bluetooth
   fast_pair.crypto.tinycrypt:
     sysbuild: true
     platform_allow:
@@ -61,4 +61,4 @@ tests:
       - nrf54l15pdk/nrf54l15/cpuapp
       - qemu_cortex_m3
     extra_args: FILE_SUFFIX=tinycrypt
-    tags: sysbuild
+    tags: sysbuild bluetooth
--- a/tests/subsys/bluetooth/fast_pair/storage/account_key_storage/testcase.yaml
+++ b/tests/subsys/bluetooth/fast_pair/storage/account_key_storage/testcase.yaml
@@ -15,7 +15,7 @@ tests:
       - nrf5340dk/nrf5340/cpuapp/ns
       - nrf54l15pdk/nrf54l15/cpuapp
       - qemu_cortex_m3
-    tags: sysbuild
+    tags: sysbuild bluetooth
   fast_pair.storage.account_key_storage.6keys:
     sysbuild: true
     platform_allow:
@@ -23,7 +23,7 @@ tests:
     integration_platforms:
       - qemu_cortex_m3
     extra_args: CONFIG_BT_FAST_PAIR_STORAGE_ACCOUNT_KEY_MAX=6
-    tags: sysbuild
+    tags: sysbuild bluetooth
   fast_pair.storage.account_key_storage.7keys:
     sysbuild: true
     platform_allow:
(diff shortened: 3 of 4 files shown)


nordic-piks:

nordic-segl (approved):

miha-nordic (approved):

scripts/ci/tags.yaml
```diff
@@ -477,3 +477,14 @@ cellular:
     - modules/lib/cjson/
     - modules/lib/hostap/
     - modules/lib/zcbor/
+
+fast_pair:
```
MarekPieta, line 481: Is it possible to let this tag "select" `crypto` and `bluetooth` tags? That could simplify maintenance.
Alternatively we may consider enabling these tags manually for samples that enable `fast_pair` to simplify maintenance

kapi-no, line 481: Shouldn't we add `fast_pair` tag also to the Fast Pair samples? `nrf/samples/bluetooth/fast_pair`

nordic-piks: No, the approach is different. Tags are used to remove tests from scope not to added them as depending on something
(so new test/sample will be by default compiled and there needs to be added tag and dependencies to possible remove it from scope).

Similarly, it should be avoided to mix functional tags e.g. crypto bluetooth and fast_pair in single test/sample.
If only for one tag the decencies will no be fulfilled, the test will be removed from scope. Even though from different tag decencies those tests should be checked.

nordic-piks: Those nrf/samples/bluetooth/fast_pair already have bluetooth tag.
So which one is more important? bluetooth or fast_pair for this sample.

Beside that samples and tests can have different decencies (usually samples are more wide, tests should be isolated).
For this particular case, I propose to keep it as it is.

kapi-no: Well, samples can have multiple tags as far as I understand. The Fast Pair category is subcategory of Bluetooth

Sort of like Bluetooth Mesh

We have 3 top-level categories for Bluetooth samples in our documentation right now:

https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/samples.html

Bluetooth samples
Bluetooth Fast Pair samples
Bluetooth Mesh samples

nordic-piks: Ok, so to be more consistent, I can add bluetooth tag instead and update its decencies?
Indeed, there could be multiple tags for sample but for test_plan tagging mechanisms, I want to add decencies only for single one (to prevent unintentional removal of some tests from scope)

kapi-no: I see that Mesh samples do not have a distinct tag associated with them - they use the `bluetooth` tag (also associated tests use the `bluetooth` tag).

Perhaps, we should use the `bluetooth` tag for the Fast Pair tests so that they have the same tags as Fast Pair samples.

nordic-piks: I see that bluetooth mesh has bluetooth tag

nordic-piks: > I see that Mesh samples do not have a distinct tag associated with them - they use the `bluetooth` tag (also associated tests use the `bluetooth` tag).
> 
> Perhaps, we should use the `bluetooth` tag for the Fast Pair tests so that they have the same tags as Fast Pair samples.

I will then adjust my PR :)

nordic-piks:
Added bluetooth tag for fast_pair tests.
Extended depencies for bluetooth testing tag.

kapi-no (approved):

scripts/ci/tags.yaml
kapi-no: thanks

MarekPieta: Thanks for clarification

kapi-no: I think we should add zephyr/subsys/settings to the list as it is common dependency of the Bluetooth subsystem and it is active in most Bluetooth samples (and its also dependency of the Fast Pair library)

tests/subsys/bluetooth/fast_pair/crypto/testcase.yaml
```diff
@@ -25,7 +25,7 @@ tests:
       - nrf5340dk/nrf5340/cpuapp
       - nrf5340dk/nrf5340/cpuapp/ns
       - nrf54l15pdk/nrf54l15/cpuapp
-    tags: sysbuild
+    tags: sysbuild bluetooth
```
MarekPieta, line 28: I think we should use `crypto` for Fast Pair crypto tests (these tests actually do not use Bluetooth). @kapi-no what do you think?

MarekPieta: For storage tests we could stick to `bluetooth` tag (Aleksander Strzeboński will add BLE bond management soon, so we may take it as "good enough" approach)

kapi-no: ok

scripts/ci/tags.yaml
MarekPieta: It sounds like a good idea too (since we added crypto already)

tests/subsys/bluetooth/fast_pair/crypto/testcase.yaml
MarekPieta: Discussed offline with @nordic-piks. We will stick to `bluetooth` tag both Fast Pair crypto and storage tests. Moving to `crypto` tag would require to add our tests' path under the `crypto` tag (otherwise our Fast Pair crypto tests would not run if a PR modifies both the test and any other file).

MarekPieta (approved):

tests/subsys/bluetooth/fast_pair/crypto/testcase.yaml
alstrzebonski: But will it be run if someone modifies crypto backend?

nordic-piks: yes, deps are added:
     - modules/crypto/mbedtls/
    - modules/crypto/oberon-psa-crypto/
    - modules/crypto/tinycrypt/
    - nrf/subsys/nrf_security/
    - nrfxlib/crypto/
    
    https://github.com/nrfconnect/sdk-nrf/blob/0217530428c8552cff22a91cf4059e5cbf630764/scripts/ci/tags.yaml#L66-L70

alstrzebonski: Oh, I didn't understand it at first. So we add a tag to **not** run a test when the tag-related thing changes? E.g. if we have `bluetooth` tag in testcase and something related to Bluetooth changes then this testcase won't be run. @nordic-piks, am I right?

nordic-piks: Opposite, We do not run tests marked with tag when dependencies listed for that tag are not changed (files changed in PR)

alstrzebonski: Okay, so we have `bluetooth` tag and `bluetooth` depends on crypto, settings etc. Now that makes sense. Thanks!

PR #16479, opened 2024-07-17
tests: subsys: bootloader: add bootloader tag

nordic-piks:
Extend tags and depencies.

### DIFF
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -612,3 +612,23 @@ sample_edge_impulse:
     - nrf/lib/wave_gen/
     - nrf/samples/edge_impulse/
     - zephyr/soc/nordic/
+
+ci_tests_subsys_bootloader:
+  files:
+    - bootloader/mcuboot/
+    - modules/crypto/mbedtls/
+    - modules/crypto/oberon-psa-crypto/
+    - nrf/drivers/hw_cc3xx/
+    - nrf/lib/fprotect/
+    - nrf/lib/hw_unique_key/
+    - nrf/lib/identity_key/
+    - nrf/lib/modem_antenna/
+    - nrf/lib/nrf_modem_lib/
+    - nrf/samples/bootloader/
+    - nrf/subsys/bootloader/
+    - nrf/subsys/fw_info/
+    - nrf/subsys/nrf_security/
+    - nrf/tests/subsys/bootloader/
+    - nrfxlib/crypto/
+    - nrfxlib/nrf_modem/
+    - zephyr/soc/nordic/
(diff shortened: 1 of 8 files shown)


nordic-piks:

nordic-segl (approved):

scripts/ci/tags.yaml
```diff
@@ -516,3 +516,23 @@ peripheral_802154:
...
+    - nrf/lib/fprotect/
+    - nrf/lib/hw_unique_key/
+    - nrf/lib/identity_key/
+    - nrf/lib/modem_antenna/
+    - nrf/lib/nrf_modem_lib/
+    - nrf/samples/bootloader/
+    - nrf/subsys/bootloader/
+    - nrf/subsys/fw_info/
+    - nrf/subsys/nrf_security/
+    - nrf/tests/subsys/bootloader/
```
gchwier, line 535:
```suggestion
    - nrf/tests/modules/mcuboot/
    - nrf/tests/subsys/bootloader/
    - nrf/tests/subsys/dfu/
    - nrf/tests/drivers/fprotect/
    - nrf/tests/drivers/flash/
    - nrf/samples/nrf5340/extxip_smp_svr/
```

scripts/ci/tags.yaml
```diff
@@ -516,3 +516,23 @@ peripheral_802154:
...
+    - nrfxlib/crypto/
+    - nrfxlib/nrf_modem/
+    - zephyr/soc/nordic/
```
gchwier, line 538:
```suggestion
    - zephyr/soc/nordic/
    - zephyr/samples/sysbuild/with_mcuboot/
    - zephyr/tests/subsys/dfu/
    - zephyr/tests/boot/test_mcuboot/
```

gchwier: what you think about adding mcumgr also here?
```
    - zephyr/subsys/mgmt/mcumgr/
    - zephyr/tests/subsys/mgmt/mcumgr/
    - zephyr/samples/subsys/mgmt/mcumgr/
    - zephyr/include/zephyr/mgmt/mcumgr/
```
to run all bootloader CI if any change in mcumgr will be done?

scripts/ci/tags.yaml
nordic-piks: Not really.
Those depencies are related to tests/subsys/bootloader only.
So this is a list for this specific tests not for all bootloader tests and samples.
Thus adding other tests does not fit the idea.

scripts/ci/tags.yaml
nordic-piks: The same as above, those paths are depencies for tests/subsys/bootloader.
Maybe I should make this tag name more specific?
Intention is not to mark all bootloader/dfu etc. tests.

scripts/ci/tags.yaml
nordic-piks: For suggested, I will propose to add separated tag e.g dfu, fprotect flash etc.
But this is different PR.

scripts/ci/tags.yaml
nordic-piks: For those proposed I will suggest to create different tag mgmt or dfu but this is outside of this PR.

gchwier (approved):

PR #16480, opened 2024-07-17
applications: protocols_serialization: Change spippet organisation.

canisLupus1313:
Adding new global nrf snippets:
	- ci-shell - for confguring CI fielndly shell on nrf.
	- zperf - for enabling zperf in performance tests.


snippets/ci-shell/ci-shell.conf
```diff
@@ -0,0 +1,13 @@
+#
```
Damian-Nordic, line 1: This is specifically for Thread CI, right? If so, it's worth naming it more distinctively like "thread-ci-shell" or sth like that when put in the common snippet directory.

Damian-Nordic (approved):

snippets/ci-shell/ci-shell.conf
canisLupus1313: it's related to CI. not necessarly to thread one. @tomaszkob89 is there any thing specific for Thread in here?

tomaszkob89: No, I think it should be applied to all the CI samples

tomaszkob89 (approved):

PR #16481, opened 2024-07-17
samples: 802154_sniffer: reduce chances of missed frames

Damian-Nordic:
I observed that with the default number of RX buffers, 16, the sniffer starts dropping frames when sending 40 frames of size 115B, with the gap of 1ms (tested with "ltx 40 1" PHY test sample's command).

Increase the number of RX buffers to reduce the frame drops. With the new setting, I am able to run "ltx 250 1" without a drop.

### DIFF
--- a/samples/peripheral/802154_sniffer/prj.conf
+++ b/samples/peripheral/802154_sniffer/prj.conf
@@ -7,6 +7,11 @@ CONFIG_NET_PKT_TIMESTAMP=y
 # Allow sharing the RTC between IEEE 802.15.4 and Zephyr
 CONFIG_NRF_RTC_TIMER_USER_CHAN_COUNT=3
 
+# Increase the number of RX buffers to minimize the chances of frame drops.
+# The value must not be higher than 127 - X, where X is the number of notification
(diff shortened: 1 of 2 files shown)


e-rk (approved):

samples/peripheral/802154_sniffer/prj.conf
```diff
@@ -7,6 +7,11 @@ CONFIG_NET_PKT_TIMESTAMP=y
 # Allow sharing the RTC between IEEE 802.15.4 and Zephyr
 CONFIG_NRF_RTC_TIMER_USER_CHAN_COUNT=3
 
+# Increase the number of RX buffers to minimize the chances of frame drops.
+# It must not be larger than 127 - X, where X is the number of notification
+# slots reserved by the radio driver (see NTF_PRIMARY_POOL_SIZE macro def).
```
greg-fer, line 12:
```suggestion
# The value must not be higher than 127 - X, where X is the number of notification
# slots reserved by the radio driver (see NTF_PRIMARY_POOL_SIZE macro def).
```

greg-fer (approved):
Approving, but please apply my edit from comment.

PR #16488, opened 2024-07-17
applications: nrf_desktop: use nrf_deskop specific tag

nordic-piks:
bluetooth tag does not fits here.

### DIFF
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -62,7 +62,6 @@ bluetooth:
     - nrf/samples/bluetooth/
     - nrf/tests/bluetooth/
     - nrf/tests/subsys/bluetooth/
-    - nrf/applications/nrf_desktop/
     - nrf/samples/nrf5340/multiprotocol_rpmsg/
     - modules/crypto/mbedtls/
     - modules/crypto/oberon-psa-crypto/
@@ -710,4 +709,50 @@ ci_applications_machine_learning:
     - zephyr/subsys/ipc/
     - zephyr/subsys/mgmt/
     - zephyr/subsys/net/
+    - zephyr/subsys/usb/
+
+ci_applications_nrf_desktop:
+  files:
+    - bootloader/mcuboot/
+    - modules/crypto/
+    - modules/lib/zcbor/
+    - nrf/applications/ipc_radio/
+    - nrf/applications/nrf_desktop/
+    - nrf/boards/
+    - nrf/cmake/
+    - nrf/drivers/mpsl/
+    - nrf/drivers/sensor/paw3212/
+    - nrf/drivers/sensor/pmw3360/
+    - nrf/subsys/app_event_manager/
+    - nrf/subsys/bluetooth/
+    - nrf/subsys/bootloader/
+    - nrf/subsys/caf/
+    - nrf/subsys/fw_info/
+    - nrf/subsys/mgmt/
+    - nrf/subsys/mpsl/
+    - nrf/subsys/nrf_security/
+    - nrf/subsys/partition_manager/
+    - nrf/sysbuild/
+    - nrfxlib/crypto/
+    - nrfxlib/mpsl/
+    - nrfxlib/softdevice_controller/
+    - zephyr/boards/nordic/
+    - zephyr/cmake/
+    - zephyr/drivers/usb/
+    - zephyr/drivers/bluetooth/
+    - zephyr/drivers/entropy/
+    - zephyr/drivers/flash/
+    - zephyr/share/
+    - zephyr/soc/nordic/
+    - zephyr/subsys/bluetooth/
+    - zephyr/subsys/dfu/
+    - zephyr/subsys/fs/
+    - zephyr/subsys/ipc/
+    - zephyr/subsys/logging/
+    - zephyr/subsys/mgmt/
+    - zephyr/subsys/net/
+    - zephyr/subsys/pm/
+    - zephyr/subsys/random/
+    - zephyr/subsys/settings/
+    - zephyr/subsys/storage/
     - zephyr/subsys/usb/
--- a/applications/nrf_desktop/sample.yaml
+++ b/applications/nrf_desktop/sample.yaml
@@ -3,7 +3,7 @@ sample:
   description: HID device reference design
 common:
   sysbuild: true
-  tags: bluetooth ci_build sysbuild
+  tags: ci_build sysbuild ci_applications_nrf_desktop
   harness: console
   harness_config:
     type: multi_line


scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - bootloader/mcuboot/
+    - modules/crypto/mbedtls/
+    - modules/crypto/oberon-psa-crypto/
+    - modules/crypto/tinycrypt/
+    - modules/lib/open-amp/
+    - modules/lib/zcbor/
+    - nrf/applications/ipc_radio/
+    - nrf/applications/nrf_desktop/
+    - nrf/boards/nordic/
+    - nrf/drivers/mpsl/temp_nrf5/
```
MarekPieta, line 626: Can we make it: `nrf/drivers/mpsl` ? We also use e.g. `flash_sync` algorithm from here

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - zephyr/drivers/bluetooth/
+    - zephyr/drivers/entropy/
+    - zephyr/drivers/flash/
+    - zephyr/soc/nordic/
+    - zephyr/subsys/bluetooth/
+    - zephyr/subsys/dfu/
+    - zephyr/subsys/fs/
+    - zephyr/subsys/ipc/
+    - zephyr/subsys/mgmt
+    - zephyr/subsys/net
```
MarekPieta, line 654: Should each entry end with `/` ?

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
     - nrf/lib/wave_gen/
     - nrf/samples/edge_impulse/
     - zephyr/soc/nordic/
+
+ci_applications_nrf_desktop:
+  files:
+    - bootloader/mcuboot/
+    - modules/crypto/mbedtls/
+    - modules/crypto/oberon-psa-crypto/
+    - modules/crypto/tinycrypt/
```
MarekPieta, line 620: Consider simplifying to `modules/crypto` here (no need to explicitly list these 3 subdirectories)

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - modules/lib/open-amp/
```
MarekPieta, line 621: Should we also add `modules/hal/nordic`?

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - modules/lib/zcbor/
+    - nrf/applications/ipc_radio/
+    - nrf/applications/nrf_desktop/
+    - nrf/boards/nordic/
```
MarekPieta, line 625: `nrf/boards` - nRF Desktop also uses nrf21540ek shield

scripts/ci/tags.yaml
MarekPieta: `nrf/cmake` and `nrf/sysbuild` should be added too I think (build system changes may break something too)

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - nrf/subsys/caf/
+    - nrf/subsys/fw_info/
+    - nrf/subsys/mgmt/
+    - nrf/subsys/mpsl/
+    - nrf/subsys/nrf_rpc/
+    - nrf/subsys/nrf_security/
+    - nrf/subsys/partition_manager/
+    - nrf/subsys/sdfw_services/
+    - nrf/subsys/suit/
+    - nrfxlib/crypto/
```
MarekPieta, line 641: `nrfxlib/mpsl` and `nrfxlib/softdevice_controller` -> If the binaries are updated we need to rerun tests

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - nrfxlib/nrf_rpc/
+    - zephyr/drivers/usb/
+    - zephyr/boards/nordic/
```
MarekPieta, line 644: Please keep it ordered alphabetically

MarekPieta: Also `zephyr/cmake` and `zephyr/share`

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - zephyr/drivers/bluetooth/
+    - zephyr/drivers/entropy/
+    - zephyr/drivers/flash/
```
MarekPieta, line 647: Also `zephyr/drivers/adc`,  `zephyr/drivers/gpio`,  `zephyr/drivers/hwinfo`,  `zephyr/drivers/led`, `zephyr/drivers/pwm`, `zephyr/drivers/spi`, `zephyr/drivers/watchdog` (or maybe simply: `zephyr/drivers`)?

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - zephyr/soc/nordic/
+    - zephyr/subsys/bluetooth/
+    - zephyr/subsys/dfu/
+    - zephyr/subsys/fs/
+    - zephyr/subsys/ipc/
```
MarekPieta, line 652: Also `zephyr/subsys/logging`

scripts/ci/tags.yaml
```diff
@@ -612,3 +611,49 @@ sample_edge_impulse:
...
+    - zephyr/subsys/mgmt
+    - zephyr/subsys/net
+    - zephyr/subsys/pm/
```
MarekPieta, line 655: Also `zephyr/subsys/settings`

MarekPieta (changes requested):

scripts/ci/tags.yaml
nordic-piks: > Can we make it: `nrf/drivers/mpsl` ? We also use e.g. `flash_sync` algorithm from here

Will add.

nordic-piks: > `nrf/cmake` and `nrf/sysbuild` should be added too I think (build system changes may break something too)

Indeed it may break something, but in this way all tests/samples should have dependency to this.
I prefer to not use this application to check build system behavior - its should be checked by all tests and samples and that should be force by the author of build system changes - by removing all optimization (similar like for upmerge).
Thus I am against adding such dependencies here.

scripts/ci/tags.yaml
nordic-piks: Good catch, will fix.

scripts/ci/tags.yaml
nordic-piks: Will do.

scripts/ci/tags.yaml
nordic-piks: Hm, of course this app depends on it (like probably all in nordic), however, I see it like this: nrf_desktop should not verify correctness of nrfx drivers. That should be done by drivers tests. Thus prefer not to add this.

scripts/ci/tags.yaml
nordic-piks: Will do.

scripts/ci/tags.yaml
nordic-piks: Will add.

scripts/ci/tags.yaml
nordic-piks: The same as for nrf/sysbuild and nrf/cmake
build system should not be verified by specific app, it should be checked by all of them.

scripts/ci/tags.yaml
nordic-piks: Hm, I would rather want to remove all drivers except some crucial one.

scripts/ci/tags.yaml
nordic-piks: It is on the list.

scripts/ci/tags.yaml
nordic-piks: I prefer not, the same like other drivers.

scripts/ci/tags.yaml
MarekPieta: ok

scripts/ci/tags.yaml
MarekPieta: So maybe all apps should depend on it (to ensure we spot issues early)?

scripts/ci/tags.yaml
MarekPieta: ok, I see you want to aggressively reduce the scope here

scripts/ci/tags.yaml
MarekPieta: ok, I see you want to aggressively reduce the scope here

scripts/ci/tags.yaml
MarekPieta: I am not sure about this one. Even a small change in Cmake files could lead to build failures. Let's discuss it briefly.

scripts/ci/tags.yaml
nordic-piks: Will add.

MarekPieta: Discussed offline. We will add logging dependency here

MarekPieta (approved):

PR #16490, opened 2024-07-17
cracen: Add handling of power failures

vili-nordic:
Adds basic handling and roll back of KMU in case of power failures.

test_crypto: PR-645

### DIFF
--- a/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.c
+++ b/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.c
@@ -19,6 +19,14 @@
 #include "common.h"
 #include "kmu.h"
 
+/* Reserved slot, used to record whether provisioning is in progress for a set of slots.
+ * We only use the metadata field, formatted as follows:
+ *      Bits 31-16: Unused
+ *      Bits 15-8:  slot-id
+ *      Bits 7-0:   number of slots
+ */
+#define PROVISIONING_SLOT 250
+
 extern nrf_security_mutex_t cracen_mutex_symmetric;
 
 /* The section .nrf_kmu_reserved_push_area is placed at the top RAM address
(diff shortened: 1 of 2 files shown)


subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.c
```diff
@@ -240,6 +242,74 @@ static bool can_sign(const psa_key_attributes_t *key_attr)
...
+				}
+			}
+		}
+
+		st = lib_kmu_revoke_slot(PROVISIONING_SLOT);
+		if (st != LIB_KMU_SUCCESS) {
+			return PSA_ERROR_HARDWARE_FAILURE;
+		}
+	}
+	return PSA_SUCCESS;
```
Vge0rge, line 273: Here we will return success even if this functions returns LIB_KMU_ERROR, right? Maybe better to check for specific error codes?

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/kmu.c
```diff
@@ -19,6 +19,8 @@
 #include "common.h"
 #include "kmu.h"
 
+#define PROVISIONING_SLOT 250
```
Vge0rge, line 22: Short description here for the process would be nice. Simple things like:
"We reserve one KMU slot to verify that the provisioning process was not interrupted by a power failure or similar reasons. 
We do this by writing the following metadata to the provisioning slot before we store the provisioning our own keys. We only use the provisioning slot metadata for this. 
The provisioning slot metadata are 32 bit long and they are used as follows:
Reserved for future use (16 bits) | Slot id (8 bits) | Number of slots (8 bits)
"

Vge0rge (approved):

stephen-nordic (approved):

PR #16492, opened 2024-07-17
applications: nrf_desktop: Add DVFS module

zycz:
Add DVFS module for nrf_desktop. It makes sure
application core works on 64MHz by deafult.
It upscales cores frequency to 320MHz when USB is used. It upscales cores frequency to 128MHz when LLPM is used.

JIRA: NCSDK-27399

I will add documentation and changelog in documentation task

### DIFF
--- a/applications/nrf_desktop/src/modules/Kconfig.dvfs
+++ b/applications/nrf_desktop/src/modules/Kconfig.dvfs
@@ -1,0 +1,45 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menuconfig DESKTOP_DVFS
+	bool "DVFS module"
+	depends on SOC_NRF54H20_CPUAPP
+	depends on !CAF_BLE_COMMON_EVENTS || (BT_MAX_CONN = 1)
+	default y
+	help
+	  This option enable DVFS module which switches frequency and voltage according
+	  to application needs.
+
+if DESKTOP_DVFS
+
+config DESKTOP_DVFS_RETRY_BUSY_TIMEOUT_MS
+	int "Retry timeout"
+	default 1
+	range 1 10000
+	help
+	  Timeout in milliseconds specifying time after which DVFS module will retry DVFS frequency change.
+	  This timeout is applied in case another DVFS change request is still in progress which causes the current request to fail.
+
+config DESKTOP_DVFS_RETRY_INIT_TIMEOUT_MS
+	int "Retry timeout"
+	default 500
+	range 1 10000
+	help
+	  Timeout in milliseconds specifying time after which DVFS module will retry DVFS frequency change.
+	  This timeout is applied in case DVFS is not yet initialized which causes the current request to fail.
+
+config DESKTOP_DVFS_RETRY_COUNT
+	int "Number of retries"
+	default 5
+	range 1 255
+	help
+	  Number of retries of DVFS frequency change after which DVFS module will report MODULE_STATE_ERROR.
+
+module = DESKTOP_DVFS
+module-str = DVFS
+source "subsys/logging/Kconfig.template.log_config"
+
+endif # DESKTOP_DVFS
--- a/applications/nrf_desktop/src/modules/ble_latency.c
+++ b/applications/nrf_desktop/src/modules/ble_latency.c
@@ -23,6 +23,9 @@ LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_BLE_LATENCY_LOG_LEVEL);
 #define LOW_LATENCY_CHECK_PERIOD_MS	K_SECONDS(5)
 #define DEFAULT_LATENCY			CONFIG_BT_PERIPHERAL_PREF_LATENCY
 #define DEFAULT_TIMEOUT			CONFIG_BT_PERIPHERAL_PREF_TIMEOUT
+/* Connection intervals used by LLPM are out of Bluetooth LE specification.
+ * The intervals are encoded by OR operation with a magic number of 0x0d00.
+ */
 #define REG_CONN_INTERVAL_LLPM_MASK	0x0d00
 #define REG_CONN_INTERVAL_BLE_DEFAULT	0x0006
 
--- a/applications/nrf_desktop/src/modules/dvfs.c
+++ b/applications/nrf_desktop/src/modules/dvfs.c
@@ -1,0 +1,270 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <caf/events/ble_common_event.h>
+
+#include "usb_event.h"
+
+#define MODULE dvfs
+#include <caf/events/module_state_event.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_DVFS_LOG_LEVEL);
+
+#include <ld_dvfs_handler.h>
+
+#define DVFS_RETRY_INITIALIZATION_TIMEOUT	K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_INIT_TIMEOUT_MS)
+#define DVFS_RETRY_BUSY_TIMEOUT			K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_BUSY_TIMEOUT_MS)
+#define DVFS_NUMBER_OF_RETRIES			CONFIG_DESKTOP_DVFS_RETRY_COUNT
+/* Connection intervals used by LLPM are out of Bluetooth LE specification.
+ * The intervals are encoded by OR operation with a magic number of 0x0d00.
+ */
+#define REG_CONN_INTERVAL_LLPM_MASK	0x0d00
+
+enum state {
+	STATE_DISABLED,
+	STATE_READY,
+	STATE_ERROR,
+	STATE_COUNT
+};
+
+static enum state module_state = STATE_DISABLED;
+
+enum dvfs_state {
+	DVFS_STATE_INITIALIZING,
+	DVFS_STATE_LLPM_CONNECTED,
+	DVFS_STATE_USB_CONNECTED,
+	DVFS_STATE_COUNT
+};
+
+static const uint8_t dvfs_high_freq_bitmask = BIT(DVFS_STATE_USB_CONNECTED) |
+					      BIT(DVFS_STATE_INITIALIZING);
+static const uint8_t dvfs_medlow_freq_bitmask = BIT(DVFS_STATE_LLPM_CONNECTED);
+
+/* Binary mask tracking which states are requested.
+ * We start with DVFS_STATE_INITIALIZING on as it is active on start.
+ */
+static uint8_t dfvs_requests_state = BIT(DVFS_STATE_INITIALIZING);
+
+static enum dvfs_frequency_setting current_freq = DVFS_FREQ_HIGH;
+
+static struct dvfs_retry_work_data {
+	struct k_work_delayable dvfs_retry_work;
+	uint8_t dfvs_retries_cnt;
+} dvfs_retry_data;
+
+static const char *get_dvfs_frequency_setting_name(enum dvfs_frequency_setting setting)
+{
+	switch (setting) {
+	case DVFS_FREQ_HIGH: return "DVFS_FREQ_HIGH";
+	case DVFS_FREQ_MEDLOW: return "DVFS_FREQ_MEDLOW";
+	case DVFS_FREQ_LOW: return "DVFS_FREQ_LOW";
+	default: return "Unknown";
+	}
+}
+
+static const char *get_dvfs_state_name(enum dvfs_state state)
+{
+	switch (state) {
+	case DVFS_STATE_INITIALIZING: return "DVFS_STATE_INITIALIZING";
+	case DVFS_STATE_LLPM_CONNECTED: return "DVFS_STATE_LLPM_CONNECTED";
+	case DVFS_STATE_USB_CONNECTED: return "DVFS_STATE_USB_CONNECTED";
+	default: return "Unknown";
+	}
+}
+
+static void cancel_dvfs_retry_work(void)
+{
+	(void) k_work_cancel_delayable(&dvfs_retry_data.dvfs_retry_work);
+	dvfs_retry_data.dfvs_retries_cnt = 0;
+}
+
+static void handle_dvfs_error(int32_t err)
+{
+	if (dvfs_retry_data.dfvs_retries_cnt >= DVFS_NUMBER_OF_RETRIES) {
+		LOG_ERR("DVFS retry count exceeded.");
+		module_set_state(MODULE_STATE_ERROR);
+		module_state = STATE_ERROR;
+		cancel_dvfs_retry_work();
+		return;
+	}
+	dvfs_retry_data.dfvs_retries_cnt++;
+	k_timeout_t timeout;
+
+	if (err == -EBUSY) {
+		LOG_DBG("DVFS frequency change in progress");
+		timeout = DVFS_RETRY_BUSY_TIMEOUT;
+	} else if (err == -EAGAIN) {
+		LOG_DBG("DVFS not initialized, trying again.");
+		timeout = DVFS_RETRY_INITIALIZATION_TIMEOUT;
+	} else {
+		LOG_ERR("DVFS freq change returned with error: %d", err);
+		module_set_state(MODULE_STATE_ERROR);
+		module_state = STATE_ERROR;
+		cancel_dvfs_retry_work();
+		return;
+	}
+	(void) k_work_reschedule(&dvfs_retry_data.dvfs_retry_work, timeout);
+}
+
+static void set_dvfs_freq(enum dvfs_frequency_setting target_freq)
+{
+	int32_t ret = dvfs_service_handler_change_freq_setting(target_freq);
+
+	if (ret) {
+		handle_dvfs_error(ret);
+	} else {
+		current_freq = target_freq;
+		LOG_INF("Have requested %s frequency",
+			get_dvfs_frequency_setting_name(target_freq));
+		cancel_dvfs_retry_work();
+	}
+}
+
+static enum dvfs_frequency_setting check_required_frequency(void)
+{
+	if (dfvs_requests_state & dvfs_high_freq_bitmask) {
+		return DVFS_FREQ_HIGH;
+	} else if (dfvs_requests_state & dvfs_medlow_freq_bitmask) {
+		return DVFS_FREQ_MEDLOW;
+	} else {
+		return DVFS_FREQ_LOW;
+	}
+}
+
+static void process_dvfs_states(enum dvfs_state state, bool turn_on)
+{
+	if (module_state == STATE_ERROR) {
+		return;
+	}
+
+	if (turn_on) {
+		dfvs_requests_state |= BIT(state);
+		LOG_DBG("%s ACTIVE", get_dvfs_state_name(state));
+	} else {
+		dfvs_requests_state &= ~BIT(state);
+		LOG_DBG("%s NOT ACTIVE", get_dvfs_state_name(state));
+	}
+
+	enum dvfs_frequency_setting required_freq = check_required_frequency();
+
+	if ((required_freq != current_freq) &&
+	    (!k_work_delayable_is_pending(&dvfs_retry_data.dvfs_retry_work))) {
+		set_dvfs_freq(required_freq);
+	} else if ((required_freq == current_freq) &&
+	    (k_work_delayable_is_pending(&dvfs_retry_data.dvfs_retry_work))) {
+		cancel_dvfs_retry_work();
+	}
+}
+
+static bool handle_ble_peer_conn_params_event(const struct ble_peer_conn_params_event *event)
+{
+	if (!event->updated) {
+		/* Ignore the connection parameters update request. */
+		return false;
+	}
+
+	__ASSERT_NO_MSG(event->interval_min == event->interval_max);
+
+	process_dvfs_states(DVFS_STATE_LLPM_CONNECTED,
+			    event->interval_min & REG_CONN_INTERVAL_LLPM_MASK);
+
+	return true;
+}
+
+static void dvfs_retry_work_handler(struct k_work *work)
+{
+	LOG_DBG("Retrying to change DVFS frequency.");
+
+	enum dvfs_frequency_setting required_freq = check_required_frequency();
+
+	__ASSERT_NO_MSG(required_freq != current_freq);
+
+	set_dvfs_freq(required_freq);
+}
+
+static void get_req_modules(struct module_flags *mf)
+{
+	module_flags_set_bit(mf, MODULE_IDX(main));
+#if CONFIG_CAF_SETTINGS_LOADER
+	module_flags_set_bit(mf, MODULE_IDX(settings_loader));
+#endif
+#if CONFIG_DESKTOP_BLE_BOND_ENABLE
+	module_flags_set_bit(mf, MODULE_IDX(ble_bond));
+#endif
+}
+
+static bool app_event_handler(const struct app_event_header *aeh)
+{
+	static struct module_flags req_modules_bm;
+
+	if (is_module_state_event(aeh)) {
+		const struct module_state_event *event = cast_module_state_event(aeh);
+
+		if (check_state(event, MODULE_ID(main), MODULE_STATE_READY)) {
+			__ASSERT_NO_MSG((dvfs_high_freq_bitmask & dvfs_medlow_freq_bitmask) == 0);
+			k_work_init_delayable(&dvfs_retry_data.dvfs_retry_work,
+					      dvfs_retry_work_handler);
+			get_req_modules(&req_modules_bm);
+			module_state = STATE_READY;
+
+			/* In case user implemented empty get_req_modules function */
+			if (module_flags_check_zero(&req_modules_bm)) {
+				process_dvfs_states(DVFS_STATE_INITIALIZING, false);
+				return false;
+			}
+		}
+
+		if (module_flags_check_zero(&req_modules_bm)) {
+			/* Frequency already changed */
+			return false;
+		}
+
+		if (event->state == MODULE_STATE_READY) {
+			module_flags_clear_bit(&req_modules_bm, module_idx_get(event->module_id));
+
+			if (module_flags_check_zero(&req_modules_bm)) {
+				process_dvfs_states(DVFS_STATE_INITIALIZING, false);
+			}
+		}
+		return false;
+	}
+
+	if (IS_ENABLED(CONFIG_CAF_BLE_COMMON_EVENTS) && is_ble_peer_conn_params_event(aeh)) {
+		return handle_ble_peer_conn_params_event(cast_ble_peer_conn_params_event(aeh));
+	}
+
+	if (IS_ENABLED(CONFIG_CAF_BLE_COMMON_EVENTS) && is_ble_peer_event(aeh)) {
+		if (cast_ble_peer_event(aeh)->state == PEER_STATE_DISCONNECTED) {
+			process_dvfs_states(DVFS_STATE_LLPM_CONNECTED, false);
+		}
+
+		return false;
+	}
+
+	if (IS_ENABLED(CONFIG_DESKTOP_USB_ENABLE) && is_usb_state_event(aeh)) {
+		const struct usb_state_event *event = cast_usb_state_event(aeh);
+
+		process_dvfs_states(DVFS_STATE_USB_CONNECTED, event->state == USB_STATE_ACTIVE);
+
+		return false;
+	}
+
+	/* If event is unhandled, unsubscribe. */
+	__ASSERT_NO_MSG(false);
+
+	return false;
+}
+
+APP_EVENT_LISTENER(MODULE, app_event_handler);
+APP_EVENT_SUBSCRIBE(MODULE, module_state_event);
+#if CONFIG_CAF_BLE_COMMON_EVENTS
+APP_EVENT_SUBSCRIBE(MODULE, ble_peer_conn_params_event);
+APP_EVENT_SUBSCRIBE(MODULE, ble_peer_event);
+#endif
+#if CONFIG_DESKTOP_USB_ENABLE
+APP_EVENT_SUBSCRIBE(MODULE, usb_state_event);
+#endif
--- a/applications/nrf_desktop/src/modules/CMakeLists.txt
+++ b/applications/nrf_desktop/src/modules/CMakeLists.txt
@@ -96,3 +96,6 @@ target_sources_ifdef(CONFIG_DESKTOP_CPU_MEAS_ENABLE
 
 target_sources_ifdef(CONFIG_DESKTOP_NRF_PROFILER_SYNC_GPIO_ENABLE
 		     app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/nrf_profiler_sync.c)
+
+target_sources_ifdef(CONFIG_DESKTOP_DVFS
+		     app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/dvfs.c)
--- a/applications/nrf_desktop/src/modules/Kconfig
+++ b/applications/nrf_desktop/src/modules/Kconfig
@@ -38,5 +38,6 @@ rsource "Kconfig.hotfixes"
 rsource "Kconfig.failsafe"
 rsource "Kconfig.cpu_meas"
 rsource "Kconfig.nrf_profiler_sync"
+rsource "Kconfig.dvfs"
 
 endmenu


applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/types.h>
+#include <zephyr/bluetooth/conn.h>
```
MarekPieta, line 8: Do we actually need `conn` header?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+
+#include <caf/events/ble_common_event.h>
+#include <caf/events/ble_smp_event.h>
+#include "config_event.h"
+#include "usb_event.h"
+#include <caf/events/button_event.h>
```
MarekPieta, line 14: Is button event header needed?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_DVFS_LOG_LEVEL);
+
+#include "ld_dvfs_handler.h"
+
+#define DVFS_INIT_TIMEOUT		K_SECONDS(1)
+#define REG_CONN_INTERVAL_LLPM_MASK	0x0d00
+
+enum dvfs_states_list {
```
MarekPieta, line 28: Naming here is a bit misleading - these are not mutually exclusive states.

Maybe we can e.g. make it `app_state` call members like: `APP_STATE_USB_CONNECTED`, `APP_LLPM_CONNECTED`? We also do not need the `DEFAULT` element here (default is used when no other bit is set). You could add `APP_STATE_INITIALIZING` and set in on boot to ensure we use high frequency until app is fully initialized.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+					break;
+				}
+			}
+		}
+	}
+
+	return highest_requested_freq;
+}
+
+static void update_dfvs_requests_state(enum dvfs_states_list state, bool turn_on)
```
MarekPieta, line 96: `turn_on` -> `enable` (to improve consistency with naming in other modules)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+{
+	if (turn_on) {
+		dfvs_requests_state |= BIT(state);
+	} else {
+		dfvs_requests_state &= ~BIT(state);
+	}
+}
+
+static void handle_dfvs_request(enum dvfs_states_list state, bool turn_on)
```
MarekPieta, line 105: `handle_dfvs_request` -> `app_state_update`?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+}
+
+static void set_dvfs_freq(enum dvfs_frequency_setting target_freq)
+{
+	size_t ret = dvfs_service_handler_change_freq_setting(target_freq);
+
+	if (ret == -EAGAIN) {
+		LOG_INF("DVFS not initialized, try again.");
+	} else if (ret == -EBUSY) {
+		LOG_INF("DVFS frequency change in progress");
```
MarekPieta, line 67: We should retry on error after a timeout (otherwise we may get stuck with invalid frequency)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+	DVFS_STATE_COUNT
+};
+
+/* Table containing the frequency settings for each state.
+ *It is tailored for nrf54h20dk and might need to be adjusted for other platforms.
+ */
+enum dvfs_frequency_setting dvfs_freq_table[DVFS_STATE_COUNT] = {
+	DVFS_FREQ_LOW,		/* DVFS_STATE_DEFAULT */
+	DVFS_FREQ_MEDLOW,	/* DVFS_STATE_LLPM */
+	DVFS_FREQ_HIGH		/* DVFS_STATE_USB */
```
MarekPieta, line 41: Consider switching to bitmasks for `MEDLOW` and `HIGH` states requests to speed up setting proper DVFS state on app_state_changes. IIRC, we agreed on this approach during our design meeting and I even pointed you to CAF Power Manager for an example.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+	case DVFS_FREQ_LOW: return "DVFS_FREQ_LOW";
+	default: return "Unknown";
+	}
+}
+
+static void set_dvfs_freq(enum dvfs_frequency_setting target_freq)
+{
+	size_t ret = dvfs_service_handler_change_freq_setting(target_freq);
+
+	if (ret == -EAGAIN) {
```
MarekPieta, line 64: I would use debug log level for the "expected" error codes (if we retry anyway, user may not need to see them)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+	if (!event->updated) {
+		/* Ignore the connection parameters update request. */
+		return;
+	}
+
+	__ASSERT_NO_MSG(event->interval_min == event->interval_max);
+
+	if (event->interval_min & REG_CONN_INTERVAL_LLPM_MASK) {
+		handle_dfvs_request(DVFS_STATE_LLPM, true);
+		LOG_DBG("LLPM_STATE_ACTIVE");
```
MarekPieta, line 126: You could map enum to string in similar way as for `get_dvfs_frequency_setting_name` and output this log in generic way as part of `app_state_update` (currently `handle_dfvs_request`)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+		handle_dfvs_request(DVFS_STATE_LLPM, false);
+		LOG_DBG("LLPM_STATE_NOT_ACTIVE");
+	}
+}
+
+static void DVFS_init_timeout_handler(struct k_work *work)
+{
+	LOG_DBG("DVFS should be initialized by now.");
+
+	handle_dfvs_request(DVFS_STATE_DEFAULT, true);
```
MarekPieta, line 137: I think it may be more reasonable to rely on app initialization chain here instead (wait for `settings_loader`, `ble_bond`, `ble_adv`/`ble_scan`). After app is fully initialized you can clear the `APP_INITIALIZING` bit

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+	if (is_ble_peer_conn_params_event(aeh)) {
+		check_if_llpm_state(cast_ble_peer_conn_params_event(aeh));
+
+		return false;
+	}
+
+	if (is_usb_state_event(aeh)) {
+		const struct usb_state_event *event = cast_usb_state_event(aeh);
+
+		if (event->state == USB_STATE_ACTIVE) {
```
MarekPieta, line 164: This could be simplified to `handle_dfvs_request(DVFS_STATE_USB, event->state == USB_STATE_ACTIVE)`

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+static void check_if_llpm_state(const struct ble_peer_conn_params_event *event)
+{
+	if (!event->updated) {
+		/* Ignore the connection parameters update request. */
+		return;
+	}
+
+	__ASSERT_NO_MSG(event->interval_min == event->interval_max);
+
+	if (event->interval_min & REG_CONN_INTERVAL_LLPM_MASK) {
```
MarekPieta, line 124: `handle_dfvs_request(DVFS_STATE_LLPM, event->interval_min & REG_CONN_INTERVAL_LLPM_MASK)`

MarekPieta: You must  also clear the flag on BLE disconnection

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <zephyr/types.h>
+#include <zephyr/bluetooth/conn.h>
+
+#include <caf/events/ble_common_event.h>
+#include <caf/events/ble_smp_event.h>
```
MarekPieta, line 11: `ble_smp_event` is not used now (also `config_event`)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,186 @@
...
+#include "config_event.h"
+#include "usb_event.h"
+#include <caf/events/button_event.h>
+#include <caf/events/power_event.h>
```
MarekPieta, line 15: Consider enforcing `DVFS_FREQ_LOW` when application is suspended (on `power_down_event`). If `power_down_event` is not going to be used, remove the header

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: Alternatively, if we want to ensure that user will see them, we could use warning log (unexpected situation relevant for end user)

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: Discussed offline, we will use `dvfs_state` and call members `DVFS_STATE_INITIALIZING`, `DVFS_STATE_USB_CONNECTED`, `DVFS_STATE_USB_CONNECTED`, `DVFS_STATE_LLPM_CONNECTED`

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: Discussed offline. We will call it `dvfs_state_update`

applications/nrf_desktop/src/modules/dvfs.c
zycz: disused offline will go with dvfs_state

applications/nrf_desktop/src/modules/dvfs.c
zycz: Created follow up task: https://nordicsemi.atlassian.net/browse/NCSDK-28474

applications/nrf_desktop/src/modules/dvfs.c
zycz: disused offline. Will base on retrying mechanism.

applications/nrf_desktop/src/modules/dvfs.c
zycz: changed to process_dvfs_states as we also do other things here and there already is function update_dvfs_states

applications/nrf_desktop/src/modules/dvfs.c
zycz: I don't recall it being said there but never mind.
Is this worth doing, because we don't control DVFS_FREQ_ macros and we would have to re declare them and we loose some memory and gain some performance.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+{
+	if (is_module_state_event(aeh)) {
+		const struct module_state_event *event =
+			cast_module_state_event(aeh);
+
+		if (check_state(event, MODULE_ID(main), MODULE_STATE_READY)) {
+
+			k_work_init_delayable(&DVFS_retry_data.DVFS_retry_work,
+					      DVFS_retry_work_handler);
+			process_dvfs_states(DVFS_STATE_INITIALIZING, false);
```
MarekPieta, line 178: Shouldn't we wait for `settings_loader` or `ble_bond` to speed up initialization?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+	/* If event is unhandled, unsubscribe. */
+	__ASSERT_NO_MSG(false);
+
+	return false;
+}
+
+APP_EVENT_LISTENER(MODULE, app_event_handler);
+APP_EVENT_SUBSCRIBE(MODULE, module_state_event);
+APP_EVENT_SUBSCRIBE(MODULE, ble_peer_conn_params_event);
+APP_EVENT_SUBSCRIBE(MODULE, ble_peer_event);
```
MarekPieta, line 215: Could we wrap also other events with similar ifdefs? (e.g. to enable building the module without BLE)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+
+	if (is_ble_peer_event(aeh)) {
+		if (cast_ble_peer_event(aeh)->state == PEER_STATE_DISCONNECTED) {
+			process_dvfs_states(DVFS_STATE_LLPM_CONNECTED, false);
+		}
+
+		return false;
+	}
+
+	if (is_usb_state_event(aeh)) {
```
MarekPieta, line 198: `if (IS_ENABLED(CONFIG_DESKTOP_USB_ENABLE) && is_usb_state_event(aeh))`. Otherwise you would see linker errors related to undefined event when building with USB disabled.

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menuconfig DESKTOP_DVFS_ENABLE
```
mkapala-nordic, line 7: `_ENABLE` is not needed.
As far as I see newer Kconfig options in nRF Desktop do not use it.

applications/nrf_desktop/src/modules/ble_latency.c
```diff
@@ -23,6 +23,7 @@ LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_BLE_LATENCY_LOG_LEVEL);
 #define LOW_LATENCY_CHECK_PERIOD_MS	K_SECONDS(5)
 #define DEFAULT_LATENCY			CONFIG_BT_PERIPHERAL_PREF_LATENCY
 #define DEFAULT_TIMEOUT			CONFIG_BT_PERIPHERAL_PREF_TIMEOUT
+/* 0x0d00 is a magic number which codes LLPM connection */
```
mkapala-nordic, line 26: do we need this comment? I do not think that it adds any more information than the define itself

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+};
+
+/* Table containing the frequency settings for each state.
+ * It is tailored for nrf54h20dk and might need to be adjusted for other platforms.
+ */
+enum dvfs_frequency_setting dvfs_freq_table[DVFS_STATE_COUNT] = {
+	DVFS_FREQ_HIGH,		/* DVFS_STATE_INITIALIZING */
+	DVFS_FREQ_MEDLOW,	/* DVFS_STATE_LLPM_CONNECTED */
+	DVFS_FREQ_HIGH		/* DVFS_STATE_USB_CONNECTED */
+	};
```
mkapala-nordic, line 39: Instead of comments:
```suggestion
enum dvfs_frequency_setting dvfs_freq_table[DVFS_STATE_COUNT] = {
	[DVFS_STATE_INITIALIZING] = DVFS_FREQ_HIGH,	
	[DVFS_STATE_LLPM_CONNECTED] = DVFS_FREQ_MEDLOW,
	[DVFS_STATE_USB_CONNECTED] = DVFS_FREQ_HIGH,
};
```

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+/* Binary mask tracking which states are requested */
+static uint8_t dfvs_requests_state;
+
+static enum dvfs_frequency_setting current_freq = DVFS_FREQ_HIGH;
+
+struct DVFS_retry_work_data {
+	struct k_work_delayable DVFS_retry_work;
+	enum dvfs_frequency_setting target_freq;
+	uint8_t dfvs_retries_cnt;
+} static DVFS_retry_data;
```
mkapala-nordic, line 50: other variables are named with dvfs in lowercase, let's align it:
```suggestion
static struct dvfs_retry_work_data {
	struct k_work_delayable dvfs_retry_work;
	enum dvfs_frequency_setting target_freq;
	uint8_t dfvs_retries_cnt;
} dvfs_retry_data;
```

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+	case DVFS_STATE_INITIALIZING: return "DVFS_STATE_INITIALIZING";
+	case DVFS_STATE_LLPM_CONNECTED: return "DVFS_STATE_LLPM_CONNECTED";
+	case DVFS_STATE_USB_CONNECTED: return "DVFS_STATE_USB_CONNECTED";
+	default: return "Unknown";
+	}
+}
+
+static void reschedule_dfvs_retry_work(enum dvfs_frequency_setting target_freq)
+{
+	if (DVFS_retry_data.dfvs_retries_cnt > 3) {
```
mkapala-nordic, line 74: maybe add define or kconfig for the retries count?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+
+#define MODULE dvfs
+#include <caf/events/module_state_event.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_DVFS_LOG_LEVEL);
+
+#include "ld_dvfs_handler.h"
+
+#define DVFS_RETRY_TIMEOUT		K_MSEC(500)
```
MarekPieta, line 21: Wouldn't it be better to use longer timeout if DVFS is uninitialized and shorter timeout after it reports busy? (e.g. after USB is connected we may need good performance earlier even if DVFS was busy)

MarekPieta: After an error we should wait for the work to perform retry (we should not retry instantly on application event).

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+
+	} else if (ret != 0) {
+		LOG_ERR("DVFS freq change returned with error: %d", ret);
+		module_set_state(MODULE_STATE_ERROR);
+	} else {
+		LOG_INF("Have requested %s frequency",
+			get_dvfs_frequency_setting_name(target_freq));
+		k_work_cancel_delayable(&DVFS_retry_data.DVFS_retry_work);
+		current_freq = target_freq;
+		DVFS_retry_data.dfvs_retries_cnt = 0;
```
mkapala-nordic, line 104: Consider bundling the k_work_cancel_delayable + zeroing retries count into the separate helper function (similarly to the `reschedule_dfvs_retry_work`) which would be executed here.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+
+static void set_dvfs_freq(enum dvfs_frequency_setting target_freq)
+{
+	size_t ret = dvfs_service_handler_change_freq_setting(target_freq);
+
+	if (ret == -EAGAIN) {
+		LOG_DBG("DVFS not initialized, trying again.");
+		reschedule_dfvs_retry_work(target_freq);
+
+	} else if (ret == -EBUSY) {
```
mkapala-nordic, line 92: if it is -EBUSY then dvfs is already initialized? 
If yes then the log message at the line L75 would be incorrect i think.

mkapala-nordic (commented):
Logic looks good, some small suggestions with improvements

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: Also you forgot to initialize the bitmask (the `DVFS_STATE_INITIALIZING` bit should be initialized to `1`)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+			k_work_init_delayable(&DVFS_retry_data.DVFS_retry_work,
+					      DVFS_retry_work_handler);
+			process_dvfs_states(DVFS_STATE_INITIALIZING, false);
+		}
+
+		return false;
+	}
+
+	if (is_ble_peer_conn_params_event(aeh)) {
+		check_if_llpm_state(cast_ble_peer_conn_params_event(aeh));
```
MarekPieta, line 185: Can we call it `handle_ble_peer_conn_params_event` (for consistency with existing code)

applications/nrf_desktop/src/modules/Kconfig.dvfs
MarekPieta: Both approaches have some pros and cons (this one explicitly shows that option enables a module). I would not enforce selecting one of the approaches.

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/Kconfig.dvfs
mkapala-nordic: IMO adding `_ENABLE` seems repetitive as by writing `=y` we are already saying that we are enabling the option, that is why i suggested it.

applications/nrf_desktop/src/modules/ble_latency.c
zycz: So you suggest somehow passing info that this is a magic number inside name - please suggest better naming then  REG_CONN_INTERVAL_LLPM_MASK. 

maybe LLPM_REG_CONN_INTERVAL_MAGIC_NUM?

applications/nrf_desktop/src/modules/dvfs.c
zycz: This one is problematic as we cannot lower (retries * timeout time) as we need ~1s in the beginning for DVFS to initialize. I don't expect anyone to  have this timeout even larger.

applications/nrf_desktop/src/modules/ble_latency.c
zycz: Also check Mares suggestion below.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
+/*
+ * Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#include <caf/events/ble_common_event.h>
+
+#if CONFIG_DESKTOP_USB_ENABLE
```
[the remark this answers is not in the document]
zycz: You mean only here or in all places? If the second what in case we don't have USB? wouldn't enabling it use more memory?

applications/nrf_desktop/src/modules/Kconfig.dvfs
MarekPieta: The `_ENABLE` suffix explicitly means that the option is used to enable a new functionality and not e.g. to configure an existing functionality (so it's not repetitive). Still, I would not insist to keep it.

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: In all places where you include an application event header. Including an application event header should not affect memory consumption I think.

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
zycz: Don't understand it is just by it and DVFS_RETRY_TIMEOUT  is of type , k_timeout_t so why uint8_t

applications/nrf_desktop/src/modules/dvfs.c
zycz: And how do you plan to do it in other way than the solution that was proposed previously?

applications/nrf_desktop/src/modules/dvfs.c
zycz: Ok, but IMO check_if_llpm_state is much better

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,218 @@
...
+{
+	if (!event->updated) {
+		/* Ignore the connection parameters update request. */
+		return;
+	}
+
+	__ASSERT_NO_MSG(event->interval_min == event->interval_max);
+
+	process_dvfs_states(DVFS_STATE_LLPM_CONNECTED,
+			    event->interval_min & REG_CONN_INTERVAL_LLPM_MASK);
```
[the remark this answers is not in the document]
zycz: https://nordicsemi.atlassian.net/browse/NCSDK-28489

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: `dfvs_retries_cnt` is handled on `uint8_t`. If you set limit e.g. to `257` it would silently be never reached

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: We could e.g. use `k_work_schedule` or rely on `atomic` to prevent rescheduling when already scheduled

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,214 @@
...
+ * DVFS_STATE_INITIALIZING	DVFS_FREQ_HIGH
+ * DVFS_STATE_LLPM_CONNECTED	DVFS_FREQ_MEDLOW
+ * DVFS_STATE_USB_CONNECTED	DVFS_FREQ_HIGH
+ */
+static uint8_t dvfs_freq_bitmask = 5;
+
+/* Binary mask tracking which states are requested.
+ * We start with DVFS_STATE_INITIALIZING on as it is active on start.
+ */
+static uint8_t dfvs_requests_state = 1;
```
MarekPieta, line 45: `BIT(DVFS_STATE_INITIALIZING)`?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,214 @@
...
+
+/* Binary mask containing the frequency settings for each state.
+ * "1" means that the state requires the DVFS_FREQ_HIGH and "0" the DVFS_FREQ_MEDLOW.
+ * It is tailored for nrf54h20dk and might need to be adjusted for other platforms.
+ *
+ * DVFS_STATE_INITIALIZING	DVFS_FREQ_HIGH
+ * DVFS_STATE_LLPM_CONNECTED	DVFS_FREQ_MEDLOW
+ * DVFS_STATE_USB_CONNECTED	DVFS_FREQ_HIGH
+ */
+static uint8_t dvfs_freq_bitmask = 5;
```
MarekPieta, line 40: `BIT(DVFS_STATE_USB_CONNECTED) | BIT(DVFS_STATE_INITIALIZING)`

MarekPieta: Can we call it `dvfs_high_frequency_bitmask`? Also it should be `static const uint8_t` (it's configuration that is not expected to change)

applications/nrf_desktop/src/modules/dvfs.c
mkapala-nordic: I do not see a problem here.
Kconfigs allow to define the range, so the 3 could be set as the lower range.
If we do not want it user configurable then just define it at the top of the file `#define DVFS_MAX_RETRY_COUNT 3` with additional comment why such value has been selected.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,214 @@
...
+	(void) k_work_reschedule(&dvfs_retry_data.dvfs_retry_work, DVFS_RETRY_TIMEOUT);
+}
+
+static void set_current_dvfs_frequency_and_reset_retries(enum dvfs_frequency_setting target_freq)
+{
+	LOG_INF("Have requested %s frequency", get_dvfs_frequency_setting_name(target_freq));
+	k_work_cancel_delayable(&dvfs_retry_data.dvfs_retry_work);
+	current_freq = target_freq;
+	dvfs_retry_data.dfvs_retries_cnt = 0;
+}
```
mkapala-nordic, line 91: This is not what i suggested in the https://github.com/nrfconnect/sdk-nrf/pull/16492#discussion_r1686251590 (suggested only the k_work_cancel_delayable + zeroing retries count).

I meant single responsibility, something like:
```
void cancel_dvfs_retry_work(void)
{
	k_work_cancel_delayable(&dvfs_retry_data.dvfs_retry_work);
	dvfs_retry_data.dfvs_retries_cnt = 0;
}
```

the frequency related stuff could have been like it was.

applications/nrf_desktop/src/modules/dvfs.c
zycz: I wouldn't change name. Both names are not precise but current version is less misleading IMO

applications/nrf_desktop/src/modules/dvfs.c
zycz: Ok I added two different timeouts

applications/nrf_desktop/src/modules/dvfs.c
zycz: Ok, added config

mkapala-nordic (dismissed):

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: DVFS integration uses 3 frequencies (high, medium, low). Because of that I think bitmask that elevates needed DVFS frequency to high needs to indicate high frequency in name. If not `dvfs_high_freq_bitmask`, then maybe we can call it `dvfs_high_freq_requests_bitmask` ?

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,36 @@
...
+if DESKTOP_DVFS
+
+config DESKTOP_DVFS_RETRY_TIMEOUT_MS
+	int "Retry timeout"
+	default 50
+	range 1 1000
+	help
+	  Timeout in milliseconds for retrying DVFS operation.
+
+config DESKTOP_DVFS_NUMBER_OF_RETRIES
```
MarekPieta, line 25: `DESKTOP_DVFS_RETRY_COUNT`

MarekPieta: To keep shared `DESKTOP_DVFS_RETRY` prefix to Kconfigs related to the retry

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,36 @@
...
+	  to application needs.
+
+if DESKTOP_DVFS
+
+config DESKTOP_DVFS_RETRY_TIMEOUT_MS
+	int "Retry timeout"
+	default 50
+	range 1 1000
+	help
+	  Timeout in milliseconds for retrying DVFS operation.
```
MarekPieta, line 23: The module requests DVFS frequency change on application events. It's possible that subsequent request is performed while the frequency change triggered by the previous request is still in progress. In that case, module needs to retry DVFS frequency change after some time.

The option specifies timeout for retrying the DVFS frequency change.

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,36 @@
...
+
+config DESKTOP_DVFS_NUMBER_OF_RETRIES
+	int "Number of retries"
+	default 3
+	range 3 255
+	help
+	  Number of retries for DVFS operation.
```
MarekPieta, line 30: The module requests DVFS frequency change on application events. It's possible that subsequent request is performed while the frequency change triggered by the previous request is still in progress. In that case, module needs to retry DVFS frequency change after some time.

The option specifies number of retries after error for a DVFS frequency change request. If the specified number of retries is exceeded, module reports MODULE_STATE_ERROR.

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,36 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menuconfig DESKTOP_DVFS
+	bool "DVFS module"
+	depends on SOC_NRF54H20_CPUAPP
+	depends on BT_MAX_CONN=1
```
MarekPieta, line 10: `depends on CAF_BLE_COMMON_EVENTS || (BT_MAX_CONN = 1)`

This condition does not need to be met for configurations without Bluetooth LE, you added ifdefs to allow building in such configuration already

MarekPieta: Looking at this it seems to make sense to introduce module-specific options that would control handling given types of application events (`CONFIG_DESKTOP_DVFS_STATE_USB`, `CONFIG_DESKTOP_DVFS_STATE_LLPM`). The options could depend on required application events (`CONFIG_DESKTOP_DVFS_STATE_LLPM` would also depend on proper connection count) and they could be enabled by default.

Thanks to that Kconfig would clearly show dependencies, user will gain fine-grained control over the module and the module could be used also on dongles (only for USB part).

applications/nrf_desktop/src/modules/dvfs.c
zycz: dvfs_high_medlow_freq_bitmask ?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,223 @@
...
+#include <caf/events/module_state_event.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_DVFS_LOG_LEVEL);
+
+#include "ld_dvfs_handler.h"
+
+#define DVFS_RETRY_INITIALIZATION_TIMEOUT	K_MSEC(500)
+#define DVFS_RETRY_TIMEOUT			K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_TIMEOUT_MS)
+#define DVFS_NUMBER_OF_RETRIES			CONFIG_DESKTOP_DVFS_NUMBER_OF_RETRIES
```
MarekPieta, line 21: Should this parameter control also retries related to DVFS initialization? If so then we should clearly describe it in Kconfig help I think.

I would suggest to keep the initialization retry count separated (it's technically a different parameter).

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,223 @@
...
+}
+
+static void reschedule_dfvs_retry_work(enum dvfs_frequency_setting target_freq, bool is_initialized)
+{
+	if (dvfs_retry_data.dfvs_retries_cnt > DVFS_NUMBER_OF_RETRIES) {
+		LOG_ERR("DVFS retry count exceeded.");
+		module_set_state(MODULE_STATE_ERROR);
+		return;
+	}
+	dvfs_retry_data.dfvs_retries_cnt++;
```
MarekPieta, line 84: You still reschedule the work without waiting for previous timeout to expire. The timeout specified by user in Kconfig may not be properly taken into account if events appear in quick succession.

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: > Shouldn't we wait for `settings_loader` or `ble_bond` to speed up initialization?

@zycz ?

applications/nrf_desktop/src/modules/Kconfig.dvfs
MarekPieta: If we don't want to let user control subset of handled events, we could make the options promptless

applications/nrf_desktop/src/modules/Kconfig.dvfs
MarekPieta: I also wonder if a single retry with proper timeout range would not be sufficient in our use-case (that may simplify implementation and configuration significantly)

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/Kconfig.dvfs
zycz: will take a look under this task: https://nordicsemi.atlassian.net/browse/NCSDK-27381

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: Discussed offline. We will use 2 separate bitmasks - one for high and the other one for medlow frequency

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,237 @@
...
+
+static enum dvfs_frequency_setting current_freq = DVFS_FREQ_HIGH;
+
+static struct dvfs_retry_work_data {
+	struct k_work_delayable dvfs_retry_work;
+	uint8_t dfvs_retries_cnt;
+} dvfs_retry_data;
+
+enum app_initial_task {
+	APP_INITIAL_TASK_SETTINGS_LOADER,
```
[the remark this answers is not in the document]
MarekPieta: Actually reinventing existing solution leads to extra complexity (you need to define a new enum that is specific to the dvfs module and make explicit mapping from module_idx to enum member)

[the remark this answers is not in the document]
mkapala-nordic: +1 for the usage of already implemented `module_flags` and @MarekPieta arguments.
It seems that it is quite easy to integrate so I do not see the reason why we should reimplement it.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,237 @@
...
+{
+	switch (state) {
+	case DVFS_STATE_INITIALIZING: return "DVFS_STATE_INITIALIZING";
+	case DVFS_STATE_LLPM_CONNECTED: return "DVFS_STATE_LLPM_CONNECTED";
+	case DVFS_STATE_USB_CONNECTED: return "DVFS_STATE_USB_CONNECTED";
+	default: return "Unknown";
+	}
+}
+
+static void reschedule_dfvs_retry_work(enum dvfs_frequency_setting target_freq, bool is_initialized)
```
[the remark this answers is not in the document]
zycz: ? We are checking if initialization is over. The error is only a symptom and not a cause.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,237 @@
...
+ * "1" means that the state requires the DVFS_FREQ_HIGH and "0" the DVFS_FREQ_MEDLOW.
+ * It is tailored for nrf54h20dk and might need to be adjusted for other platforms.
+ *
+ * DVFS_STATE_INITIALIZING	DVFS_FREQ_HIGH
+ * DVFS_STATE_LLPM_CONNECTED	DVFS_FREQ_MEDLOW
+ * DVFS_STATE_USB_CONNECTED	DVFS_FREQ_HIGH
+ */
+static const uint8_t dvfs_high_freq_bitmask = BIT(DVFS_STATE_USB_CONNECTED) |
+					      BIT(DVFS_STATE_INITIALIZING);
+static const uint8_t dvfs_medlow_freq_bitmask = BIT(DVFS_STATE_LLPM_CONNECTED);
```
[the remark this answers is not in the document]
zycz: can't do that this simple due to error: expression in static assertion is not constant

I don't think it is so important so won't add it.

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: Or maybe we could pass an error code here and react on it properly? (then function could be named `handle_dvfs_error`)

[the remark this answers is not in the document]
zycz: I don't like it cause I do not treat AGAIN or BUSY as an errors yet - in sense the module state still does not go into error sate immediately.

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: So maybe we can assert in once during initialization (to ensure valid configuration)?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,260 @@
...
+
+#define DVFS_RETRY_INITIALIZATION_TIMEOUT	K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_INIT_TIMEOUT_MS)
+#define DVFS_RETRY_TIMEOUT			K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_BUSY_TIMEOUT_MS)
+#define DVFS_NUMBER_OF_RETRIES			CONFIG_DESKTOP_DVFS_RETRY_COUNT
+/* Connection intervals used by LLPM are out of Bluetooth LE specification.
+ * The intervals are encoded by OR operation with a magic number of 0x0d00.
+ */
+#define REG_CONN_INTERVAL_LLPM_MASK	0x0d00
+
+static bool in_error_state;
```
MarekPieta, line 27: Consider using
```
enum state {
STATE_DISABLED,
STATE_READY,
STATE_ERROR,
}
```
To improve consistency with other application modules

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,260 @@
...
+	struct k_work_delayable dvfs_retry_work;
+	uint8_t dfvs_retries_cnt;
+} dvfs_retry_data;
+
+enum app_initial_task {
+	APP_INITIAL_TASK_SETTINGS_LOADER,
+	APP_INITIAL_TASK_BLE_BOND
+};
+
+static struct module_flags req_modules_bm;
```
MarekPieta, line 57: We could move it to `module_state_event` handler function scope

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,45 @@
...
+	default y
+	help
+	  This option enable DVFS module which switches frequency and voltage according
+	  to application needs.
+
+if DESKTOP_DVFS
+
+config DESKTOP_DVFS_RETRY_BUSY_TIMEOUT_MS
+	int "Retry timeout"
+	default 50
```
MarekPieta, line 20: Have you checked if we can make this timeout shorter?

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
mkapala-nordic: I think that you should be able to workaround it using defines:
```
#define DVFS_HIGH_FREQ_BITMASK_INIT        (BIT(DVFS_STATE_USB_CONNECTED) | BIT(DVFS_STATE_INITIALIZING))
#define DVFS_MEDLOW_FREQ_BITMASK_INIT      (BIT(DVFS_STATE_LLPM_CONNECTED))

BUILD_ASSERT((DVFS_HIGH_FREQ_BITMASK_INIT & DVFS_MEDLOW_FREQ_BITMASK_INIT) == 0);

static const uint8_t dvfs_high_freq_bitmask = DVFS_HIGH_FREQ_BITMASK_INIT;
static const uint8_t dvfs_medlow_freq_bitmask = DVFS_MEDLOW_FREQ_BITMASK_INIT;
```

Stuff under the macros should be constant.

applications/nrf_desktop/src/modules/dvfs.c
mkapala-nordic: +1 for enums, if not then maybe at least rename it to for example `initialization_err` or something more specific

applications/nrf_desktop/src/modules/Kconfig.dvfs
zycz: according to lowlevel team it takes about 1ms for dvfs operation to be complete. However we can encounter situation where also other domain (secdom) requests dvfs change - in that situation lets set timeout for 2 ms

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,260 @@
...
+	if (dfvs_requests_state & dvfs_high_freq_bitmask) {
+		return DVFS_FREQ_HIGH;
+	} else if (dfvs_requests_state & dvfs_medlow_freq_bitmask) {
+		return DVFS_FREQ_MEDLOW;
+	} else {
+		return DVFS_FREQ_LOW;
+	}
+}
+
+static void process_dvfs_states(enum dvfs_state state, bool turn_on)
```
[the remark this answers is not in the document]
zycz: IMO opinion we should keep track of current DVFS preferred state. Lets imagine problem is caused by sysctrl and after a while we manage to recover from it - we might want to set DVFS in correct state and would need to know what frequency is needed.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,260 @@
...
+
+static void cancel_dvfs_retry_work(void)
+{
+	(void) k_work_cancel_delayable(&dvfs_retry_data.dvfs_retry_work);
+	dvfs_retry_data.dfvs_retries_cnt = 0;
+}
+
+static void set_dvfs_freq(enum dvfs_frequency_setting target_freq)
+{
+	if (in_error_state) {
```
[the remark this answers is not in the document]
zycz: But we are doing that - there is no much cpu computation difference in checking k_work_delayable_is_pending - checks state of work and dvfs_service_handler_change_freq_setting  - checks one register

applications/nrf_desktop/src/modules/dvfs.c
zycz: Shouldn't we use module_state states? But I cannot find in api a function which gets current module state

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: @zycz, can we apply comment from @mkapala-nordic ?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,264 @@
...
+static void set_dvfs_freq(enum dvfs_frequency_setting target_freq)
+{
+	if (module_state == STATE_ERROR) {
+		return;
+	}
+
+	size_t ret = dvfs_service_handler_change_freq_setting(target_freq);
+
+	if (ret) {
+		reschedule_dfvs_retry_work(target_freq, ret);
```
MarekPieta, line 123: Can we call it `handle_dvfs_error` ?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,264 @@
...
+	dvfs_retry_data.dfvs_retries_cnt = 0;
+}
+
+static void set_dvfs_freq(enum dvfs_frequency_setting target_freq)
+{
+	if (module_state == STATE_ERROR) {
+		return;
+	}
+
+	size_t ret = dvfs_service_handler_change_freq_setting(target_freq);
```
MarekPieta, line 120: It's an `int32_t`, not `size_t`

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: We do not have code to perform recovery after a fatal error so it may be reasonable to skip updates after we enter error state (we would not update DVFS frequency anyway)

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: Generally, modules use `module_state_event` only to broadcast the current state and use module-specific way to remember their own states. Current approach looks ok.

applications/nrf_desktop/src/modules/Kconfig.dvfs
MarekPieta: Looks ok, 50 ms was far too much

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: Actually I am not sure if we should depend on implementation details there. From API user perspective, it may be reasonable to avoid requesting DVFS frequency update until timeout expires. @mkapala-nordic what do you think?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,264 @@
...
+
+#include "usb_event.h"
+
+#define MODULE dvfs
+#include <caf/events/module_state_event.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(MODULE, CONFIG_DESKTOP_DVFS_LOG_LEVEL);
+
+#include "ld_dvfs_handler.h"
```
MarekPieta, line 17: `#include <ld_dvfs_handler.h>`?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,264 @@
...
+
+#define DVFS_RETRY_INITIALIZATION_TIMEOUT	K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_INIT_TIMEOUT_MS)
+#define DVFS_RETRY_TIMEOUT			K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_BUSY_TIMEOUT_MS)
```
MarekPieta, line 20: `DVFS_RETRY_BUSY_TIMEOUT` (For consistency with `DVFS_RETRY_INITIALIZATION_TIMEOUT`)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,264 @@
...
+#define DVFS_NUMBER_OF_RETRIES			CONFIG_DESKTOP_DVFS_RETRY_COUNT
+/* Connection intervals used by LLPM are out of Bluetooth LE specification.
+ * The intervals are encoded by OR operation with a magic number of 0x0d00.
+ */
+#define REG_CONN_INTERVAL_LLPM_MASK	0x0d00
+
+enum state {
+STATE_INITIALIZING,
```
MarekPieta, line 28: Please indent entries (for consistency with other modules)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,264 @@
...
+
+static void reschedule_dfvs_retry_work(enum dvfs_frequency_setting target_freq, size_t dvfs_ret_val)
+{
+	if (dvfs_retry_data.dfvs_retries_cnt > DVFS_NUMBER_OF_RETRIES) {
+		LOG_ERR("DVFS retry count exceeded.");
+		module_set_state(MODULE_STATE_ERROR);
+		module_state = STATE_ERROR;
+		return;
+	}
+	if (k_work_delayable_is_pending(&dvfs_retry_data.dvfs_retry_work)) {
```
MarekPieta, line 87: This check needs to go above `if (dvfs_retry_data.dfvs_retries_cnt > DVFS_NUMBER_OF_RETRIES) {`

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,264 @@
...
+{
+	switch (state) {
+	case DVFS_STATE_INITIALIZING: return "DVFS_STATE_INITIALIZING";
+	case DVFS_STATE_LLPM_CONNECTED: return "DVFS_STATE_LLPM_CONNECTED";
+	case DVFS_STATE_USB_CONNECTED: return "DVFS_STATE_USB_CONNECTED";
+	default: return "Unknown";
+	}
+}
+
+static void reschedule_dfvs_retry_work(enum dvfs_frequency_setting target_freq, size_t dvfs_ret_val)
```
MarekPieta, line 79: `dvfs_err`? It's actually an error code

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: @mkapala-nordic what do you think?

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dvfs.c
zycz: I prefer current name

applications/nrf_desktop/src/modules/dvfs.c
zycz: I don't like it. module_state_event  should support both broadcast  and module specific. In that way there won't be a situation where we have different values the one broadcasted and one owned by module

zycz:
rebased

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: @mkapala-nordic what's your opinion on this? Handling error codes that do not lead to work rescheduling in `reschedule_dfvs_retry_work` is a bit fishy

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: @mkapala-nordic, what do you think?

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
mkapala-nordic: If we can't recover from fatal error then i do not see any reason why we should allow any operations then.

applications/nrf_desktop/src/modules/dvfs.c
mkapala-nordic: Marek's suggestions looks good to me, it explains better what is our intention here.
Also we reschedule on only 2 error codes, for the other we do not do that so it is misleading.

> it suggest we encounter an error and it is sth critical

IMO an error is an error, no matter the severity.

applications/nrf_desktop/src/modules/dvfs.c
mkapala-nordic: maybe just `err` or `ret` (either for me is fine) if we would change the name of the function to `handle_dvfs_error` as suggested in other command?

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
mkapala-nordic: I am not sure if there was some other implementation when this comment was written but it looks like we are already waiting for the timeout work.

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,45 @@
...
+	  Timeout in milliseconds specifying time after which DVFS module will retry DVFS frequency change.
+	  This timeout is applied in case another DVFS change request is still in progress which causes the current request to fail.
+
+config DESKTOP_DVFS_RETRY_INIT_TIMEOUT_MS
+	int "Retry timeout"
+	default 500
+	range 1 10000
+	help
+	  Timeout in milliseconds specifying time after which DVFS module will retry DVFS frequency change.
+	  This timeout is applied in case DVFS is not yet inicizlized which causes the current request to fail.
```
mkapala-nordic, line 32:
```suggestion
	  This timeout is applied in case DVFS is not yet initialized which causes the current request to fail.
```

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
zycz: I don't understand why we might not want to keep track of dfvs freq. for now we don't do it but in case of this module we might.  We do not stop other modules but maybe we should then on error stop the whole SoC as if the error happens it is probably sth with systemctrl and we should avoid working in such conditions imminently

applications/nrf_desktop/src/modules/dvfs.c
zycz: but api is not user friendly for such use case - maybe we should create a ticket to fix this problem

applications/nrf_desktop/src/modules/dvfs.c
zycz: > @mkapala-nordic what's your opinion on this? Handling error codes that do not lead to work rescheduling in `reschedule_dfvs_retry_work` is a bit fishy

I can always change back to first implementation there you won't have handling error codes that do not lead to work rescheduling

applications/nrf_desktop/src/modules/dvfs.c
MarekPieta: I am not sure what you mean by "not user friendly" api here. You can simply use `enum module_state` instead of custom state enum and pass it as argument to function that broadcasts module state.

zycz: Like we lack a way to easy access the state of the module.

Right now we need to have access to a event to check it:

```
static inline bool check_state(const struct module_state_event *event,
		const void *module_id, enum module_state state)
```
		
IMO you shouldn't need to have event to check it. And the name is confusing because we are not checking the current state but state which was sand in the event. The event might been copied by us and is no longer valid

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: Agreed to `__ASSERT_NO_MSG` on init.

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: Discussed offline. Agreed to stop tracking DVFS state requests after fatal error (they would not be applied anyway).

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: We will take the decision during tomorrow's standup.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,268 @@
...
+	case DVFS_STATE_USB_CONNECTED: return "DVFS_STATE_USB_CONNECTED";
+	default: return "Unknown";
+	}
+}
+
+static void handle_dvfs_error(enum dvfs_frequency_setting target_freq, int32_t err)
+{
+	if (k_work_delayable_is_pending(&dvfs_retry_data.dvfs_retry_work)) {
+		return;
+	}
```
MarekPieta, line 83: Discussed during standup. We will check retry work state before submitting subsequent DVFS frequency change request.

MarekPieta (changes requested):

mkapala-nordic (approved):

applications/nrf_desktop/src/modules/dvfs.c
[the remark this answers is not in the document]
MarekPieta: The delayed work check was moved

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,45 @@
+#
+# Copyright (c) 2024 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+menuconfig DESKTOP_DVFS
+	bool "DVFS module"
+	depends on SOC_NRF54H20_CPUAPP
+	depends on CAF_BLE_COMMON_EVENTS || (BT_MAX_CONN = 1)
```
MarekPieta, line 10: `depends on !CAF_BLE_COMMON_EVENTS || (BT_MAX_CONN = 1)` (as suggested in https://github.com/nrfconnect/sdk-nrf/pull/16492#discussion_r1687660293)

applications/nrf_desktop/src/modules/Kconfig.dvfs
```diff
@@ -0,0 +1,45 @@
...
+	default y
+	help
+	  This option enable DVFS module which switches frequency and voltage according
+	  to application needs.
+
+if DESKTOP_DVFS
+
+config DESKTOP_DVFS_RETRY_BUSY_TIMEOUT_MS
+	int "Retry timeout"
+	default 2
```
MarekPieta, line 20: Comment after our last discussion during standup:
Do you want to eventually set `retry timeout` to `1` and `retry count` to `6` (to speed up applying DVFS frequency change on `-EBUSY` error)?

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,269 @@
...
+#define DVFS_RETRY_INITIALIZATION_TIMEOUT	K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_INIT_TIMEOUT_MS)
+#define DVFS_RETRY_BUSY_TIMEOUT			K_MSEC(CONFIG_DESKTOP_DVFS_RETRY_BUSY_TIMEOUT_MS)
+#define DVFS_NUMBER_OF_RETRIES			CONFIG_DESKTOP_DVFS_RETRY_COUNT
+/* Connection intervals used by LLPM are out of Bluetooth LE specification.
+ * The intervals are encoded by OR operation with a magic number of 0x0d00.
+ */
+#define REG_CONN_INTERVAL_LLPM_MASK	0x0d00
+
+enum state {
+	STATE_INITIALIZING,
```
MarekPieta, line 28: `STATE_DISABLED` ? (for consistency with naming in existing modules). Keep in mind that this refers to module state and not state of DVFS

MarekPieta: Also you actually never enter `STATE_READY` (module should enter it on `main` module readiness as soon as it's initialized)

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,269 @@
...
+
+	enum dvfs_frequency_setting required_freq = check_required_frequency();
+
+	if ((required_freq != current_freq) &&
+	    (!k_work_delayable_is_pending(&dvfs_retry_data.dvfs_retry_work))) {
+		set_dvfs_freq(required_freq);
+	} else if ((required_freq == current_freq) &&
+	    (k_work_delayable_is_pending(&dvfs_retry_data.dvfs_retry_work))) {
+		cancel_dvfs_retry_work();
+	}
```
MarekPieta, line 160: During the standup, you discussed some performance concerns related to the module. Consider removing the `else` part and handle this in the work handler (zero retry counter + early return) to avoid impacting performance (the `else` case may be entered quite often by the code after we introduce handling of events that appear often - e.g. config channel or SMP DFU).

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dvfs.c
zycz: I do not expect it to happen often. And if it doesn't happen often it actually boost performance because we won't enter work handler at all. But we might see how it behaves when we introduce new functionality.

MarekPieta: We can leave it as is for now.

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,270 @@
...
+
+static void cancel_dvfs_retry_work(void)
+{
+	(void) k_work_cancel_delayable(&dvfs_retry_data.dvfs_retry_work);
+	dvfs_retry_data.dfvs_retries_cnt = 0;
+}
+
+static void handle_dvfs_error(int32_t err)
+{
+	if (dvfs_retry_data.dfvs_retries_cnt > DVFS_NUMBER_OF_RETRIES) {
```
MarekPieta, line 87: `>=`?

MarekPieta: Otherwise you retry one time more than expected

applications/nrf_desktop/src/modules/dvfs.c
```diff
@@ -0,0 +1,270 @@
...
+		if (check_state(event, MODULE_ID(main), MODULE_STATE_READY)) {
+			__ASSERT_NO_MSG((dvfs_high_freq_bitmask & dvfs_medlow_freq_bitmask) == 0);
+			k_work_init_delayable(&dvfs_retry_data.dvfs_retry_work,
+					      dvfs_retry_work_handler);
+			get_req_modules(&req_modules_bm);
+
+			/* In case user implemented empty get_req_modules function */
+			if (module_flags_check_zero(&req_modules_bm)) {
+				process_dvfs_states(DVFS_STATE_INITIALIZING, false);
+				state = STATE_READY;
```
MarekPieta, line 216: The `state = STATE_READY`; assignment should be done regardless of the `if (module_flags_check_zero(&req_modules_bm)) {` condition (otherwise you still actually do not set `state = STATE_READY;` ...)

MarekPieta (changes requested):

applications/nrf_desktop/src/modules/dvfs.c
zycz: I thought we enter state ready when we finish waiting for other modules to initialize

MarekPieta (approved):

mkapala-nordic (approved):

PR #16493, opened 2024-07-17
doc: nrfxlib: Fix doxygen warnings for SDC api

weeTike:
__ALIGN was not defined in the doxyfile


gmarull (commented):
can you please move formatting changes into its own commit?

olivier-le-sage (approved):

PR #16500, opened 2024-07-18
applications: machine_learning: add tags for app machine_learning

nordic-piks:
Extend tags for CI test scope selection.

### DIFF
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -682,3 +682,32 @@ ci_tests_drivers_nrfx_integration_test:
     - modules/hal/nordic/nrfx/
     - nrf/tests/drivers/nrfx_integration_test/
     - zephyr/modules/hal_nordic/
+
+ci_applications_machine_learning:
+  files:
+    - nrf/applications/ipc_radio/
+    - nrf/applications/machine_learning/
+    - nrf/boards/shields/pca63565/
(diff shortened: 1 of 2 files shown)


nordic-piks:

scripts/ci/tags.yaml
```diff
@@ -644,3 +644,30 @@ ci_tests_subsys_app_event_manager:
...
+    - nrf/lib/wave_gen/
+    - nrf/samples/nrf5340/netboot/
+    - nrf/subsys/app_event_manager/
+    - nrf/subsys/bluetooth/
+    - nrf/subsys/bootloader/
+    - nrf/subsys/caf/
+    - nrf/subsys/event_manager_proxy/
+    - nrf/subsys/mpsl/
+    - nrf/subsys/pcd/
+    - zephyr/drivers/sensor/adi/adxl362
```
nordic-segl, line 664: This sample needs sensors. Sensors are "connected" with SHIELD pca63565 and pca63566.
What about adding to the list files at:
nrf/boards/shields/pca63565
nrf/boards/shields/pca63566
https://github.com/nrfconnect/sdk-nrf/tree/main/boards/shields/pca63565

nordic-piks: Thanks, added.

nordic-segl (approved):

miha-nordic (approved):

PR #16503, opened 2024-07-18
sysbuild: Fix some issues with MCUboot signing

nordicjm:


cmake/sysbuild/image_signing.cmake
```diff
@@ -76,7 +76,14 @@ function(zephyr_mcuboot_tasks)
...
     endif()
   endif()
-  set(imgtool_sign ${PYTHON_EXECUTABLE} ${imgtool_path} sign --version ${CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION} --align 4 --slot-size @PM_MCUBOOT_PRIMARY_SIZE@ --pad-header --header-size @PM_MCUBOOT_PAD_SIZE@ ${imgtool_rom_command} CACHE STRING "imgtool sign command")
+
+  # Split fields, imgtool_sign_sysbuild is stored in cache which will have fields updated by
+  # sysbuild, imgtool_sign must not be stored in cache because it would then prevent those fields
+  # from being updated without a pristine build
+  # TODO: NCSDK-28461 sysbuild PM fields cannot be updated without a pristine build, will become
+  # invalid if a static PM file is updated without pristine build
+  set(imgtool_sign_sysbuild --slot-size @PM_MCUBOOT_PRIMARY_SIZE@ --pad-header --header-size @PM_MCUBOOT_PAD_SIZE@ ${imgtool_rom_command} CACHE STRING "imgtool sign sysbuild replacement")
```
57300, line 85: I haven't tested this, but could a solution to NCSDK-28461 be to clear out the cache variable after @ replacement? Something like:
```cmake
if($CACHE{imgtool_sign_sysbuild} MATCHES @)
  unset(imgtool_sign_sysbuild CACHE)
endif()
```

57300 (approved):

de-nordic (approved):

cmake/sysbuild/image_signing.cmake
nordicjm: I think you would then have a ping pong scenario of continually rebuilding what is already up to date

PR #16514, opened 2024-07-18
aplications: protocols_serializations: Add config for nrf54l.

canisLupus1313:
Commit adds config nesessary to compile nrf54l with `coex` and `log_rpc` snippets. Also adds former to sample.yml for twister build.

### DIFF
--- a/applications/protocols_serialization/sample.yaml
+++ b/applications/protocols_serialization/sample.yaml
@@ -35,3 +35,14 @@ tests:
     integration_platforms:
       - nrf52840dk/nrf52840
       - nrf54l15pdk/nrf54l15/cpuapp
+  applications.protocols_serialization.all:
+    build_only: true
+    platform_allow: >
+      nrf52840dk/nrf52840
+      nrf54l15pdk/nrf54l15/cpuapp
+    tags: ci_build
+    extra_args: >
+      SNIPPET="ble;openthread;debug;coex;log_rpc"
+    integration_platforms:
+      - nrf52840dk/nrf52840
+      - nrf54l15pdk/nrf54l15/cpuapp
--- a/applications/protocols_serialization/snippets/log_rpc/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
+++ b/applications/protocols_serialization/snippets/log_rpc/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
@@ -1,0 +1,35 @@
+/* Copyright (c) 2024 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+ / {
+	sram@2003FC00 {
+		compatible = "zephyr,memory-region", "mmio-sram";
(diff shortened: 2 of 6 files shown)


applications/protocols_serialization/client/sample.yaml
```diff
@@ -35,3 +35,14 @@ tests:
     integration_platforms:
       - nrf52840dk/nrf52840
       - nrf54l15pdk/nrf54l15/cpuapp
+  applications.protocols_serialization.client.all:
```
Damian-Nordic, line 38: Should we just extend the previous `.rpc` test to reduce the number of builds?

applications/protocols_serialization/server/snippets/log_rpc/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
```diff
@@ -0,0 +1,35 @@
...
+				/* This is the prefix which must appear at the front of the data */
+				prefix = [ 01 00 00 00  ];
+			};
+		};
+	};
+};
+
+/* Reduce SRAM_CPUAPP usage by 1KB to account for non-init area */
+&cpuapp_sram {
+	reg = <0x20000000 DT_SIZE_K(187)>;
```
Damian-Nordic, line 34: Actually, I can see that e.g. Matter samples override this to use the entire 256kB (since the flipper core is not used). I think we could do this for this project as well, right?

```
&cpuapp_sram {
	reg = <0x20000000 DT_SIZE_K(256)>;
	ranges = <0x0 0x20000000  0x40000>;
};
```

Damian-Nordic (approved):

alxelax (approved):

applications/protocols_serialization/client/sample.yaml
canisLupus1313: I'm not sure about this, we can talk about how to tackle this on a daily.

applications/protocols_serialization/server/snippets/log_rpc/boards/nrf54l15pdk_nrf54l15_cpuapp.overlay
canisLupus1313: We are doing here an example of an application that still could be modified on the customer site. and it would be confusing why there are problems with booting. So I would leave it as is untill we decide that we need extra space.

Damian-Nordic: I can see other samples don't have any issues with increasing this region and I think the customer will be more curious why they don't have all RAM available, but it's OK, we can do it later :).

applications/protocols_serialization/client/sample.yaml
canisLupus1313: @tomaszkob89 any thoughts?

Damian-Nordic: Feel free to disregard this for now. I just thought it would be good to only build combinations used by our CI (others might be built in nightly only). But this optimization can wait.

PR #16517, opened 2024-07-18
tests: net: openthread: rpc: commissioning unit tests

alxelax:
PR has two commits:

- unit test for ot rpc commissioning (both client and server)
- fix for ot rpc commissioning server otDatasetGetActiveTlvs


Damian-Nordic (commented):
What's the reason for splitting the client/server unit tests into multiple projects? From development point of view it seemed easier to smoke test the serialization with only 2 projects than 10. :)

alxelax:
> What's the reason for splitting the client/server unit tests into multiple projects? From development point of view it seemed easier to smoke test the serialization with only 2 projects than 10. :)

Kind of disagree. For both points of view of development and maintenance, it is better to have logically atomic, small parts of functionalities split in several files. This is very bad practice to keep everything in single many thousands line spaghetti style file that even scary to touch when it requires changes.

From the point of maintainability it is easy to scale the project in both ways as well downsizing functionality as vice verse. Everything is possible to do over build system ability to not pollute source code by conditional compilation.

Damian-Nordic:
> Kind of disagree. For both points of view of development and maintenance, it is better to have logically atomic, small parts of functionalities split in several files. This is very bad practice to keep everything in single many thousands line spaghetti style file that even scary to touch when it requires changes.
> 
> From the point of maintainability it is easy to scale the project in both ways as well downsizing functionality as vice verse. Everything is possible to do over build system ability to not pollute source code by conditional compilation.

I don't know what's spaghetti in a file containing a list of independent unit tests (more like a large lasagna ;)) but I wasn't meaning putting everything in a single source file either. You can have a single unit test project with multiple test files and I thought that would be easier to extend than having to copy-paste and adjust the entire project when you add a new command group, but I'm fine with whatever approach. Just wanted to understand the rationale.

tests/subsys/net/openthread/rpc/client/commissioning/src/main.c
```diff
@@ -0,0 +1,191 @@
...
+	zassert_false(result);
+}
+
+/* Test serialization of otDatasetSetActiveTlvs() */
+ZTEST(ot_rpc_commissioning_cli, test_otDatasetSetActiveTlvs)
+{
+	otError error;
+
+	const otOperationalDatasetTlvs dataset = {
+		{LISTIFY(OT_OPERATIONAL_DATASET_MAX_LENGTH, LSFY_IDENTITY, (,))},
```
Damian-Nordic, line 77: You could add another macro, such as:
```
#define INT_SEQUENCE(length)  LISTIFY(length, LSFY_IDENTITY, (,))
```
to slightly simplify the tests' code.

tests/subsys/net/openthread/rpc/client/commissioning/src/main.c
```diff
@@ -0,0 +1,191 @@
...
+	otOperationalDatasetTlvs dataset;
+	const otOperationalDatasetTlvs expected_dataset = {
+		{LISTIFY(OT_OPERATIONAL_DATASET_MAX_LENGTH, LSFY_IDENTITY, (,))},
+		OT_OPERATIONAL_DATASET_MAX_LENGTH};
+
+	mock_nrf_rpc_tr_expect_add(RPC_CMD(OT_RPC_CMD_DATA_GET_ACTIVE_TLVS), RPC_RSP(TLVS));
+	error = otDatasetGetActiveTlvs(NULL, &dataset);
+	mock_nrf_rpc_tr_expect_done();
+	zassert_equal(error, OT_ERROR_NONE);
+	zassert_mem_equal(&dataset, &expected_dataset, sizeof(otOperationalDatasetTlvs));
```
Damian-Nordic, line 112: Validate length as well?

tests/subsys/net/openthread/rpc/client/commissioning/src/main.c
```diff
@@ -0,0 +1,191 @@
...
+
+	error = otDatasetGetActiveTlvs(NULL, NULL);
+	zassert_equal(error, OT_ERROR_NOT_FOUND);
+}
+
+static void discover_cb(otActiveScanResult *result, void *ctx)
+{
+	uint8_t ext_addr[] = {LISTIFY(OT_EXT_ADDRESS_SIZE, LSFY_IDENTITY, (,))};
+	uint8_t nwk_name[] = {LISTIFY(OT_NETWORK_NAME_MAX_SIZE, LSFY_IDENTITY, (,)), 0};
+	uint8_t ext_pan_id[] = {LISTIFY(OT_EXT_PAN_ID_SIZE, LSFY_IDENTITY, (,)), 0};
```
Damian-Nordic, line 134: ext_pan_id an steer_data don't have to be null-terminated

tests/subsys/net/openthread/rpc/client/commissioning/src/main.c
```diff
@@ -0,0 +1,191 @@
...
+	mock_nrf_rpc_tr_expect_done();
+	zassert_equal(error, OT_ERROR_NOT_FOUND);
+
+	error = otDatasetGetActiveTlvs(NULL, NULL);
+	zassert_equal(error, OT_ERROR_NOT_FOUND);
+}
+
+static void discover_cb(otActiveScanResult *result, void *ctx)
+{
+	uint8_t ext_addr[] = {LISTIFY(OT_EXT_ADDRESS_SIZE, LSFY_IDENTITY, (,))};
```
Damian-Nordic, line 132: Worth validating if this callback has actually been called.

tests/subsys/net/openthread/rpc/server/commissioning/src/main.c
```diff
@@ -0,0 +1,236 @@
...
+		OT_OPERATIONAL_DATASET_MAX_LENGTH};
+
+	otDatasetSetActiveTlvs_fake.return_val = OT_ERROR_NONE;
+
+	mock_nrf_rpc_tr_expect_add(RPC_RSP(OT_ERROR_NONE), NO_RSP);
+	mock_nrf_rpc_tr_receive(RPC_CMD(OT_RPC_CMD_DATA_SET_ACTIVE_TLVS, TLVS));
+	mock_nrf_rpc_tr_expect_done();
+
+	zassert_equal(otDatasetSetActiveTlvs_fake.call_count, 1);
+	zassert_mem_equal(otDatasetSetActiveTlvs_fake.arg1_val, &expected_dataset,
```
Damian-Nordic, line 136: Validate the length as well?

Damian-Nordic (approved):

tests/subsys/net/openthread/rpc/client/commissioning/src/main.c
alxelax: Length is the part of `otOperationalDatasetTlvs` structure. It is checked in this comparison too.

tests/subsys/net/openthread/rpc/client/commissioning/src/main.c
alxelax: Sure, leftover after debugging. I'll remove it.

tests/subsys/net/openthread/rpc/server/commissioning/src/main.c
alxelax: Length is part of dataset.

Damian-Nordic: Ah, right, when testing the maximum dataset, it will work indeed.

Damian-Nordic (approved):

canisLupus1313 (approved):

PR #16521, opened 2024-07-19
Align tags names

nordic-piks:


nordic-piks:
The main idea here is to rename previously added tags, so that its name does not confuse to be a feature tag.
Now, the name has clear indication that it is used for CI.

nordic-bami (approved):

nordic-segl (approved):

miha-nordic (approved):

PR #16529, opened 2024-07-19
samples: bluetooth: fast_pair: locator_tag: Rework fp_adv module

mkapala-nordic:
Reworked fp_adv module so the triggers request the state of the advertising instead of setting it directly (works as a mediator).

Jira: NCSDK-28077

Depends on: 
+ https://github.com/nrfconnect/sdk-nrf/pull/16626

### DIFF
--- a/samples/bluetooth/fast_pair/locator_tag/CMakeLists.txt
+++ b/samples/bluetooth/fast_pair/locator_tag/CMakeLists.txt
@@ -30,6 +30,8 @@ target_sources(app PRIVATE
 
 zephyr_linker_sources(SECTIONS src/factory_reset_callbacks.ld)
 
+zephyr_linker_sources(DATA_SECTIONS src/fp_adv_trigger.ld)
+
 if (NOT CONFIG_BT_FAST_PAIR_FMDN_RING_COMP_NONE)
   target_sources(app PRIVATE src/ring.c)
 endif()
--- a/samples/bluetooth/fast_pair/locator_tag/CMakeLists.txt
+++ b/samples/bluetooth/fast_pair/locator_tag/CMakeLists.txt
@@ -28,6 +28,8 @@ target_sources(app PRIVATE
   src/fp_adv.c
 )
 
+zephyr_linker_sources(SECTIONS src/factory_reset_callbacks.ld)
+
 if (NOT CONFIG_BT_FAST_PAIR_FMDN_RING_COMP_NONE)
   target_sources(app PRIVATE src/ring.c)
 endif()
--- a/samples/bluetooth/fast_pair/locator_tag/include/app_factory_reset.h
+++ b/samples/bluetooth/fast_pair/locator_tag/include/app_factory_reset.h
@@ -20,16 +20,32 @@
 extern "C" {
 #endif
 
-/** Callback type used to notify the user that the factory reset has been executed. */
-typedef void (*app_factory_reset_executed_cb)(void);
+/** Factory reset callback descriptor. */
+struct app_factory_reset_callbacks {
+	/** Callback used to allow the user to prepare for the factory reset. */
+	void (*prepare)(void);
+
+	/** Callback used to notify the user that the factory reset has been executed. */
+	void (*executed)(void);
+};
+
+/** Register a factory reset callback descriptor.
+ *
+ *  @param _name Factory reset callbacks descriptor name.
+ *  @param _prepare Callback to define actions before the factory reset.
+ *  @param _executed Callback to define actions after the factory reset.
+ */
+#define APP_FACTORY_RESET_CALLBACKS_REGISTER(_name, _prepare, _executed)		\
+	static const STRUCT_SECTION_ITERABLE(app_factory_reset_callbacks, _name) = {	\
+		.prepare = _prepare,							\
+		.executed = _executed,							\
+	}
 
 /** Schedule the factory reset action.
  *
  * @param delay Time to wait before the factory reset action is performed.
- * @param cb    Callback to define the prepare action before the factory reset.
  */
-void app_factory_reset_schedule(k_timeout_t delay,
-				app_factory_reset_executed_cb cb);
+void app_factory_reset_schedule(k_timeout_t delay);
 
 /** Cancel the scheduled factory reset action. */
 void app_factory_reset_cancel(void);
--- a/samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
+++ b/samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
@@ -32,11 +32,54 @@ enum app_fp_adv_mode {
 	APP_FP_ADV_MODE_NOT_DISCOVERABLE,
 };
 
-/** Set the Fast Pair advertising mode.
+/** Fast Pair advertising trigger. */
+struct app_fp_adv_trigger {
+	/* Identifier. */
+	const char *id;
+};
+
+/** Register a trigger for Fast Pair advertising.
  *
- * @param adv_mode Fast Pair advertising mode.
+ *  @param _name Trigger structure name.
+ *  @param _id Trigger identifier.
  */
-void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode);
+#define APP_FP_ADV_TRIGGER_REGISTER(_name, _id)				\
+	BUILD_ASSERT(_id != NULL);					\
+	static STRUCT_SECTION_ITERABLE(app_fp_adv_trigger, _name) = {	\
+		.id = _id,						\
+	}
+
+/** Request turning on the Fast Pair advertising.
+ *
+ *  The Fast Pair advertising will be turned off only when no
+ *  trigger has requested to keep the Fast Pair advertising on.
+ *
+ *  If the current Fast Pair advertising mode is the same as the requested mode,
+ *  the Fast Pair advertising payload will not be refreshed.
+ *  Use the @ref app_fp_adv_payload_refresh function to manually refresh
+ *  the Fast Pair advertising payload.
+ *
+ *  @param trigger Trigger identifier.
+ *  @param enable true to request to turn on the Fast Pair advertising.
+ *		  false to remove request to turn on the Fast Pair advertising.
+ */
+void app_fp_adv_request(struct app_fp_adv_trigger *trigger, bool enable);
+
+/** Refresh the Fast Pair advertising payload.
+ *
+ *  This function is used to manually trigger the refreshing of the Fast Pair
+ *  advertising data to ensure that the advertising data is up to date.
+ *
+ *  Use with caution, as it updates the Fast Pair advertising payload asynchronously
+ *  to the RPA address rotation and to the update of the FMDN advertising payload.
+ */
+void app_fp_adv_payload_refresh(void);
+
+/** Get the current Fast Pair advertising mode.
+ *
+ *  @return Current Fast Pair advertising mode.
+ */
+enum app_fp_adv_mode app_fp_adv_mode_get(void);
 
 /** Set the suspension mode for the RPA rotations of the Fast Pair advertising set.
  *
@@ -65,23 +108,39 @@ int app_fp_adv_id_set(uint8_t id);
  */
 uint8_t app_fp_adv_id_get(void);
 
-/** Uninitialize the Fast Pair advertising module.
- *
- * @return 0 if the operation was successful. Otherwise, a (negative) error code is returned.
- */
-int app_fp_adv_uninit(void);
-
 /** Initialize the Fast Pair advertising module.
  *
  * @return 0 if the operation was successful. Otherwise, a (negative) error code is returned.
  */
 int app_fp_adv_init(void);
 
-/** Check if the Fast Pair advertising module is initialized.
+/** Enable the Fast Pair advertising module.
  *
- *  @return true when the module is initialized, false otherwise.
+ *  Can be called after the Fast Pair advertising module has been initialized using the
+ *  @ref app_fp_adv_init function.
+ *  Must be called after enabling the Fast Pair subsystem using the @ref bt_fast_pair_enable
+ *  function.
+ *
+ *  @return 0 if the operation was successful. Otherwise, a (negative) error code is returned.
  */
-bool app_fp_adv_is_init(void);
+int app_fp_adv_enable(void);
+
+/** Disable the Fast Pair advertising module.
+ *
+ *  Can be called after the Fast Pair advertising module has been initialized using the
+ *  @ref app_fp_adv_init function.
+ *  Should be called before disabling the Fast Pair subsystem using the @ref bt_fast_pair_disable
+ *  function.
+ *
+ *  @return 0 if the operation was successful. Otherwise, a (negative) error code is returned.
+ */
+int app_fp_adv_disable(void);
+
+/** Check if the Fast Pair advertising module is ready.
+ *
+ *  @return true when the module is ready, false otherwise.
+ */
+bool app_fp_adv_is_ready(void);
 
 #ifdef __cplusplus
 }
--- a/samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
+++ b/samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
@@ -36,10 +36,8 @@ static bool factory_reset_ui_requested;
 
 int bt_fast_pair_factory_reset_user_action_perform(void)
 {
-	int err;
 	int id;
 	size_t id_count;
-	bool fp_adv_is_init = app_fp_adv_is_init();
 	uint8_t bt_id;
 
 	/* Please be extra careful while implementing this hook function, as it can be
@@ -49,15 +47,6 @@ int bt_fast_pair_factory_reset_user_action_perform(void)
 	 * will be propagated as a return value of the bt_fast_pair_enable API.
 	 */
 	LOG_INF("Factory Reset: resetting Bluetooth identity within the factory reset");
-
-	/* Uninitialize the Fast Pair advertising module if it is active. */
-	if (fp_adv_is_init) {
-		err = app_fp_adv_uninit();
-		if (err) {
-			LOG_ERR("Factory Reset: app_fp_adv_uninit failed (err %d)", err);
-			return err;
-		}
-	}
 
 	/* Check if FP identity exists. */
 	bt_id = app_fp_adv_id_get();
@@ -73,15 +62,6 @@ int bt_fast_pair_factory_reset_user_action_perform(void)
 		}
 	} else {
 		LOG_INF("Factory Reset: identity for factory reset does not exist");
-	}
-
-	/* Initialize the Fast Pair advertising module if it was active before the reset. */
-	if (fp_adv_is_init) {
-		err = app_fp_adv_init();
-		if (err) {
-			LOG_ERR("Factory Reset: app_fp_adv_init failed (err %d)", err);
-			return err;
-		}
 	}
 
 	return 0;
@@ -109,7 +89,7 @@ static void factory_reset_perform(void)
 {
 	int err;
 	bool fast_pair_is_ready = bt_fast_pair_is_ready();
-	bool fp_adv_is_init = app_fp_adv_is_init();
+	bool fp_adv_is_ready = app_fp_adv_is_ready();
 
 	LOG_INF("Performing reset to factory settings...");
 
@@ -122,20 +102,20 @@ static void factory_reset_perform(void)
 
 	factory_reset_prepare();
 
+	/* Disable the Fast Pair advertising module if it is active. */
+	if (fp_adv_is_ready) {
+		err = app_fp_adv_disable();
+		if (err) {
+			LOG_ERR("Factory Reset: app_fp_adv_disable failed (err %d)", err);
+			goto finish;
+		}
+	}
+
 	/* Disable the Fast Pair subsystem if it is active. */
 	if (fast_pair_is_ready) {
 		err = bt_fast_pair_disable();
 		if (err) {
 			LOG_ERR("Factory Reset: bt_fast_pair_disable failed: %d", err);
-			goto finish;
-		}
-	}
-
-	/* Uninitialize the Fast Pair advertising module if it is active. */
-	if (fp_adv_is_init) {
-		err = app_fp_adv_uninit();
-		if (err) {
-			LOG_ERR("Factory Reset: app_fp_adv_uninit failed (err %d)", err);
 			goto finish;
 		}
 	}
@@ -149,20 +129,20 @@ static void factory_reset_perform(void)
 		goto finish;
 	}
 
-	/* Initialize the Fast Pair advertising module if it was active before the reset. */
-	if (fp_adv_is_init) {
-		err = app_fp_adv_init();
-		if (err) {
-			LOG_ERR("Factory Reset: app_fp_adv_init failed (err %d)", err);
-			goto finish;
-		}
-	}
-
 	/* Reenable the Fast Pair subsystem. */
 	if (fast_pair_is_ready) {
 		err = bt_fast_pair_enable();
 		if (err) {
 			LOG_ERR("Factory Reset: bt_fast_pair_enable failed: %d", err);
+			goto finish;
+		}
+	}
+
+	/* Reenable the Fast Pair advertising module if it was active before the reset. */
+	if (fp_adv_is_ready) {
+		err = app_fp_adv_enable();
+		if (err) {
+			LOG_ERR("Factory Reset: app_fp_adv_enable failed (err %d)", err);
 			goto finish;
 		}
 	}
--- a/samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
+++ b/samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
@@ -32,7 +32,6 @@ static void factory_reset_work_handle(struct k_work *w);
 
 static K_WORK_DELAYABLE_DEFINE(factory_reset_work, factory_reset_work_handle);
 static enum app_factory_reset_state factory_reset_state;
-static app_factory_reset_executed_cb user_cb;
 static bool factory_reset_ui_requested;
 
 int bt_fast_pair_factory_reset_user_action_perform(void)
@@ -88,6 +87,24 @@ int bt_fast_pair_factory_reset_user_action_perform(void)
 	return 0;
 }
 
+static void factory_reset_prepare(void)
+{
+	STRUCT_SECTION_FOREACH(app_factory_reset_callbacks, cbs) {
+		if (cbs->prepare) {
+			cbs->prepare();
+		}
+	}
+}
+
+static void factory_reset_executed(void)
+{
+	STRUCT_SECTION_FOREACH(app_factory_reset_callbacks, cbs) {
+		if (cbs->executed) {
+			cbs->executed();
+		}
+	}
+}
+
 static void factory_reset_perform(void)
 {
 	int err;
@@ -102,6 +119,8 @@ static void factory_reset_perform(void)
 	}
 
 	factory_reset_state = APP_FACTORY_RESET_STATE_IN_PROGRESS;
+
+	factory_reset_prepare();
 
 	/* Disable the Fast Pair subsystem if it is active. */
 	if (fast_pair_is_ready) {
@@ -150,12 +169,9 @@ static void factory_reset_perform(void)
 
 	LOG_INF("Reset to factory settings has completed");
 
-	if (user_cb) {
-		user_cb();
-	}
+	factory_reset_executed();
 
 	factory_reset_state = APP_FACTORY_RESET_STATE_IDLE;
-	user_cb = NULL;
 
 finish:
 	if (err) {
@@ -176,8 +192,7 @@ static void factory_reset_request_handle(enum app_ui_request request)
 	}
 }
 
-void app_factory_reset_schedule(k_timeout_t delay,
-				app_factory_reset_executed_cb cb)
+void app_factory_reset_schedule(k_timeout_t delay)
 {
 	if (factory_reset_state != APP_FACTORY_RESET_STATE_IDLE) {
 		LOG_ERR("Factory Reset: rejecting scheduling operation");
@@ -186,7 +201,6 @@ void app_factory_reset_schedule(k_timeout_t delay,
 
 	(void) k_work_schedule(&factory_reset_work, delay);
 	factory_reset_state = APP_FACTORY_RESET_STATE_PENDING;
-	user_cb = cb;
 }
 
 void app_factory_reset_cancel(void)
--- a/samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
+++ b/samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
@@ -30,12 +30,18 @@ LOG_MODULE_DECLARE(fp_fmdn, LOG_LEVEL_INF);
 #define FP_RPA_TIMEOUT			(13 * FP_RPA_SEC_PER_MIN)
 #define FP_RPA_TIMEOUT_OFFSET_MAX	(2 * FP_RPA_SEC_PER_MIN)
 
+#define BITS_PER_VAR(_var)	(__CHAR_BIT__ * sizeof(_var))
+
+static bool is_initialized;
+static bool is_enabled;
+
 static bool fmdn_provisioned;
 
 static struct bt_conn *fp_conn;
 static struct bt_le_ext_adv *fp_adv_set;
 static bool fp_adv_rpa_rotation_suspended;
 static enum app_fp_adv_mode fp_adv_mode = APP_FP_ADV_MODE_OFF;
+static uint32_t fp_adv_request_bm;
 static const char * const fp_adv_mode_description[] = {
 	[APP_FP_ADV_MODE_OFF] = "disabled",
 	[APP_FP_ADV_MODE_DISCOVERABLE] = "discoverable",
@@ -170,7 +176,7 @@ static int fp_adv_payload_set(bool rpa_rotated, bool new_session)
 	return 0;
 }
 
-static int fp_adv_start_internal(void)
+static int bt_stack_advertising_update(void)
 {
 	int err;
 	struct bt_le_ext_adv_start_param ext_adv_start_param = {0};
@@ -200,7 +206,7 @@ static int fp_adv_start_internal(void)
 	return 0;
 }
 
-static void fp_advertising_start(void)
+static void fp_advertising_update(void)
 {
 	int err;
 
@@ -214,7 +220,7 @@ static void fp_advertising_start(void)
 		return;
 	}
 
-	err = fp_adv_start_internal();
+	err = bt_stack_advertising_update();
 	if (!err) {
 		app_ui_state_change_indicate(APP_UI_STATE_FP_ADV,
 					     (fp_adv_mode != APP_FP_ADV_MODE_OFF));
@@ -249,6 +255,8 @@ static bool fp_adv_rpa_expired(struct bt_le_ext_adv *adv)
 	 */
 	__ASSERT_NO_MSG(!k_is_preempt_thread());
 	__ASSERT_NO_MSG(!k_is_in_isr());
+
+	__ASSERT_NO_MSG(is_enabled);
 
 	LOG_INF("Fast Pair: RPA expired");
 
@@ -288,6 +296,7 @@ static bool fp_adv_rpa_expired(struct bt_le_ext_adv *adv)
 	}
 
 	if (fp_adv_mode != APP_FP_ADV_MODE_OFF) {
+		/* Update the advertising payload. */
 		err = fp_adv_payload_set(expire_rpa, false);
 		if (err) {
 			LOG_ERR("Fast Pair: cannot set advertising payload (err: %d)", err);
@@ -300,7 +309,6 @@ static bool fp_adv_rpa_expired(struct bt_le_ext_adv *adv)
 static int fp_adv_set_setup(void)
 {
 	int err;
-	struct bt_le_oob oob;
 	static const struct bt_le_ext_adv_cb fp_adv_set_cb = {
 		.connected = fp_adv_connected,
 		.rpa_expired = fp_adv_rpa_expired,
@@ -312,15 +320,6 @@ static int fp_adv_set_setup(void)
 	err = bt_le_ext_adv_create(&fp_adv_param, &fp_adv_set_cb, &fp_adv_set);
 	if (err) {
 		LOG_ERR("Fast Pair: bt_le_ext_adv_create returned error: %d", err);
-		return err;
-	}
-
-	/* Force the RPA rotation to synchronize the Fast Pair advertising
-	 * payload with its RPA address using rpa_expired callback.
-	 */
-	err = bt_le_oob_get_local(fp_adv_param.id, &oob);
-	if (err) {
-		LOG_ERR("Fast Pair: bt_le_oob_get_local failed: %d", err);
 		return err;
 	}
 
@@ -348,7 +347,7 @@ static int fp_adv_set_teardown(void)
 
 static void fp_adv_restart_work_handle(struct k_work *w)
 {
-	fp_advertising_start();
+	fp_advertising_update();
 }
 
 static void fp_adv_conn_clear(void)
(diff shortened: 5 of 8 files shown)


samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -58,6 +58,11 @@ static bool fmdn_provisioned;
 static bool fmdn_recovery_mode;
 static bool fmdn_id_mode;
 static bool fp_account_key_present;
+static bool fp_adv_ui_request;
+
+static const char *fp_adv_trigger_clock_sync = "clock_sync";
+static const char *fp_adv_trigger_ak_write = "ak_write";
```
kapi-no, line 64: I would suggest changing name for this trigger

```suggestion
static const char *fp_adv_trigger_fmdn_provisioning = "fmdn_provisioning";
```

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -430,52 +484,56 @@ void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
...
-				break;
-			default:
-				__ASSERT_NO_MSG(0);
-			}
-		}
+	idx = ret;
 
-		fp_advertising_start();
-	}
+	adv_triggers[idx].enabled = enable;
```
kapi-no, line 525: Please do nothing if the state is already correct and simply return

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -348,24 +353,24 @@ static void fmdn_provisioning_state_changed(bool provisioned)
...
-			"advertising mode after a reset to factory settings");
+		LOG_INF("The device was unprovisioned and reset to factory setting");
+
+		/* Disable advertising requests related to the FMDN. */
+		app_fp_adv_request(fp_adv_trigger_ak_write, false);
+		app_fp_adv_request(fp_adv_trigger_clock_sync, false);
+
+		/* Disable advertising request from the UI. */
+		fp_adv_ui_request = false;
+		app_fp_adv_request(fp_adv_trigger_ui, fp_adv_ui_request);
```
kapi-no, line 364: At this point, the fp_adv module is already active and advertising. We do not want to trigger a pop-up right after notification, we need to disable advertising when the factory reset is scheduled: `if (fp_account_key_present != provisioned) {`

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -348,24 +353,24 @@ static void fmdn_provisioning_state_changed(bool provisioned)
 
 	/* Triggered on the unprovisioning operation. */
 	if (factory_reset_executed) {
-		LOG_INF("Please press a button to put the device in the Fast Pair discoverable "
-			"advertising mode after a reset to factory settings");
+		LOG_INF("The device was unprovisioned and reset to factory setting");
```
kapi-no, line 356:
```suggestion
		LOG_INF("The device was unprovisioned and reset to factory settings");
```

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -348,24 +353,24 @@ static void fmdn_provisioning_state_changed(bool provisioned)
 
 	/* Triggered on the unprovisioning operation. */
 	if (factory_reset_executed) {
-		LOG_INF("Please press a button to put the device in the Fast Pair discoverable "
```
kapi-no, line 351: You can leave this log here. This is information on how to trigger another FMDN provisioning (you need explicit user action).

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -430,52 +484,56 @@ void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
...
-	__ASSERT_NO_MSG(!k_is_preempt_thread());
-	__ASSERT_NO_MSG(!k_is_in_isr());
+	int idx;
+	int ret = app_fp_adv_registered_trigger_find(trigger);
 
-	if (!app_fp_adv_is_init()) {
+	if (ret < 0) {
+		LOG_ERR("Fast Pair: advertising trigger not registered");
+		__ASSERT_NO_MSG(0);
 		return;
```
MarekPieta, line 520: Maybe we can simplify it to `__ASSERT_NO_MSG(ret);` to avoid branching in release configuration?

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,29 @@ enum app_fp_adv_mode {
...
 };
 
-/** Set the Fast Pair advertising mode.
+/** Register a trigger for Fast Pair advertising.
  *
- * @param adv_mode Fast Pair advertising mode.
+ *  @param trigger Trigger identifier.
+ *  @return 0 if the operation was successful. Otherwise, a (negative) error code is returned.
+ */
+int app_fp_adv_trigger_register(const char *trigger);
```
MarekPieta, line 40: I would suggest to inform in API doc that the same pointer must be used for all the time by a subscriber (we do not perform strcmp). Pointer validity must be ensured too

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -385,6 +390,24 @@ static int fast_pair_prepare(void)
 		return err;
 	}
 
+	err = app_fp_adv_trigger_register(fp_adv_trigger_ak_write);
+	if ((err < 0) && (err != -EALREADY)) {
```
MarekPieta, line 394: Maybe we could assert that there would be no error? Error should not happen here unless the configuration was misaligned, so error check could be simplified to assert

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -41,6 +46,10 @@ static const char * const fp_adv_mode_description[] = {
 	[APP_FP_ADV_MODE_DISCOVERABLE] = "discoverable",
 	[APP_FP_ADV_MODE_NOT_DISCOVERABLE] = "non-discoverable",
 };
+static struct {
+	const char *trigger;
```
MarekPieta, line 50: Maybe we could use array of `const char *` and bitmask (index in array would be mapped to bit in bitmask). That would simplify checking if advertising is required.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -30,6 +30,11 @@ LOG_MODULE_DECLARE(fp_fmdn, LOG_LEVEL_INF);
 #define FP_RPA_TIMEOUT			(13 * FP_RPA_SEC_PER_MIN)
 #define FP_RPA_TIMEOUT_OFFSET_MAX	(2 * FP_RPA_SEC_PER_MIN)
 
+/* Max number of triggers that can be registered with app_fp_adv_trigger_register()
+ * to later use the app_fp_adv_request().
+ */
+#define MAX_ADV_TRIGGERS (4)
```
MarekPieta, line 36: Using code section and macro to register an element there could allow us to avoid configuration misalignment possibility. I would also simplify implementation and speed up execution (no need to look through array element by element; can rely on pointer difference)

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,29 @@ enum app_fp_adv_mode {
...
+ *		  false to request to turn off the Fast Pair advertising.
+ */
+void app_fp_adv_request(const char *trigger, bool enable);
+
+/** Get the current Fast Pair advertising mode.
+ *
+ *  @return Current Fast Pair advertising mode.
  */
-void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode);
+enum app_fp_adv_mode app_fp_adv_mode_get(void);
```
MarekPieta, line 57: Seems unused now

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -348,24 +353,24 @@ static void fmdn_provisioning_state_changed(bool provisioned)
...
-		} else {
-			fp_adv_mode = APP_FP_ADV_MODE_OFF;
-		}
-	} else {
-		fp_adv_mode = APP_FP_ADV_MODE_DISCOVERABLE;
-	}
-	app_fp_adv_mode_set(fp_adv_mode);
+	/* Request the Fast Pair advertising according to the FMDN provisioning state. */
+	app_fp_adv_request(fp_adv_trigger_clock_sync, provisioned && is_first_state_changed_cb);
+	app_fp_adv_request(fp_adv_trigger_ak_write, false);
```
MarekPieta, line 373: Would we advertise right after boot? Maybe we could set the `fmdn_provisioning` flag right on boot when not provisioned?

MarekPieta (changes requested):

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
MarekPieta: If you opt for bitmask, you could speed up the check a bit and put whole section in ROM. Still, I see that current implementation is quite simple and it's not performance critical code, so I would not insist to change it. :)

samples/bluetooth/fast_pair/locator_tag/src/main.c
MarekPieta: I think it may be a good idea. I think that sample could automatically start advertising right after boot when not provisioned to improve UX.

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -353,24 +376,21 @@ static void fmdn_provisioning_state_changed(bool provisioned)
...
-		if (is_first_state_changed_cb) {
-			fp_adv_mode = APP_FP_ADV_MODE_NOT_DISCOVERABLE;
-		} else {
-			fp_adv_mode = APP_FP_ADV_MODE_OFF;
-		}
-	} else {
-		fp_adv_mode = APP_FP_ADV_MODE_DISCOVERABLE;
-	}
-	app_fp_adv_mode_set(fp_adv_mode);
+	/* Request the Fast Pair advertising according to the FMDN provisioning state. */
```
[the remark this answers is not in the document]
MarekPieta: Similar for `if (fp_account_key_present != provisioned) {`

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -430,52 +446,53 @@ void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
...
+				 APP_FP_ADV_MODE_DISCOVERABLE;
+	} else {
+		requested_mode = APP_FP_ADV_MODE_OFF;
+	}
+
+	if (!requested_adv_on && (fp_adv_mode == APP_FP_ADV_MODE_OFF)) {
+		LOG_INF("Fast Pair: advertising: already disabled");
+		return;
+	} else if (requested_adv_on && (fp_adv_mode != APP_FP_ADV_MODE_OFF)) {
+		LOG_INF("Fast Pair: advertising: already enabled, refreshing");
```
MarekPieta, line 467: With the current approach, it would be better to provide an explicit API call to trigger advertising refresh and avoid automatic refresh when a module updates its request bit (to avoid refreshing advertising accidentally).

samples/bluetooth/fast_pair/locator_tag/src/main.c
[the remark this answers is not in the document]
MarekPieta: That could also help with https://github.com/nrfconnect/sdk-nrf/pull/16529#discussion_r1686315679

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -353,24 +376,21 @@ static void fmdn_provisioning_state_changed(bool provisioned)
...
+	app_fp_adv_request(&fp_adv_trigger_clock_sync, provisioned && is_first_state_changed_cb);
+	app_fp_adv_request(&fp_adv_trigger_fmdn_provisioning, false);
+
+	fp_adv_ui_request = !provisioned;
+	app_fp_adv_request(&fp_adv_trigger_ui, fp_adv_ui_request);
 
 	is_first_state_changed_cb = false;
```
MarekPieta, line 395: Should be assigned before early returns

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -416,7 +416,23 @@ static struct bt_fast_pair_fmdn_info_cb fmdn_info_cb = {
...
+	STRUCT_SECTION_FOREACH(app_fp_adv_trigger, adv_trigger) {
+		if (adv_trigger->enabled) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+static void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
```
kapi-no, line 435: please change the function name as it is no longer an API function

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -151,7 +179,7 @@ static void fp_account_key_written(struct bt_conn *conn)
 {
 	LOG_INF("Fast Pair: Account Key write");
 
-	app_fp_adv_mode_set(APP_FP_ADV_MODE_NOT_DISCOVERABLE);
+	app_fp_adv_request(&fp_adv_trigger_fmdn_provisioning, true);
```
kapi-no, line 182: This part can now be placed in the `if` statement as the FMDN provisioning is only triggered for the first Account Key write.

If you can change the code underneath in the following way, it would also be an improvement:

```
	/* The first and only Account Key write starts the FMDN provisioning. */
	if (!fp_account_key_present) {
		/* Fast Pair Implementation Guidelines for the locator tag use case:
		* trigger the reset to factory settings if there is no FMDN
		* provisioning operation within 5 minutes.
		*/
		fmdn_factory_reset_schedule(
			FACTORY_RESET_TRIGGER_PROVISIONING_TIMEOUT,
			K_MINUTES(FMDN_PROVISIONING_TIMEOUT));

```

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -348,24 +376,21 @@ static void fmdn_provisioning_state_changed(bool provisioned)
 
 	/* Triggered on the unprovisioning operation. */
 	if (factory_reset_executed) {
+		LOG_INF("The device was unprovisioned and reset to factory settings");
 		LOG_INF("Please press a button to put the device in the Fast Pair discoverable "
 			"advertising mode after a reset to factory settings");
```
kapi-no, line 381: The provisioning state change is already indicate in the log:

```suggestion
		LOG_INF("The device is reset to factory settings");
		LOG_INF("Please press a button to put the device in the Fast Pair discoverable "
			"advertising mode");
```

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -531,12 +550,13 @@ int app_fp_adv_init(void)
 		return err;
 	}
 
+	/* Start the advertising if any module already requested it. */
+	app_fp_adv_request_start();
```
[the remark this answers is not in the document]
kapi-no: Hmm, let's decouple the `bt_fast_pair_fmdn_info_cb_register` registration from the module enabling code. One idea is to keep the `_init` function that is used for the one-time registration action and define the new `app_fp_adv_enable` and `app_fp_adv_disable` actions. (Also remove the `uninit` and transform `is_init` to `is_ready`).

Ideally, this module should not operate when `bt_fast_pair_is_ready` is false. You can add an assert for this here. However, we need to change the order of execution in the factory_reset.c. We should first disable this module and then the Fast Pair library. Also the `fast_pair_prepare` needs to be aligned and the `user_reset_action`.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
kapi-no: Is refresh needed just to update the advertising payload?

samples/bluetooth/fast_pair/locator_tag/src/main.c
kapi-no: This is false as we do not call the `provisioning_state_changed` callback in the `bt_fast_pair_enable` function context. I removed this comment before submitting my review, but somehow it still got submitted.

mkapala-nordic (commented):
Applied suggestions from @MarekPieta and @kapi-no.

mkapala-nordic:
Rebased

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,51 @@ enum app_fp_adv_mode {
 	APP_FP_ADV_MODE_NOT_DISCOVERABLE,
 };
 
-/** Set the Fast Pair advertising mode.
+/** Fast Pair advertising trigger. */
+struct app_fp_adv_trigger {
+	const char *id;
+	bool enabled;
```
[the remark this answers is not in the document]
kapi-no: Or even better: can we use `void *ctx` instead of this field and assign the memory segments from the static array defined in the fp_adv.c source code?

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -493,32 +522,31 @@ uint8_t app_fp_adv_id_get(void)
...
 {
 	int err;
 
-	app_fp_adv_mode_set(APP_FP_ADV_MODE_OFF);
-	app_fp_adv_rpa_rotation_suspend(false);
-
-	fp_adv_conn_clear();
-
-	err = fp_adv_set_teardown();
+	err = bt_fast_pair_fmdn_info_cb_register(&fmdn_info_cb);
```
kapi-no, line 529:
```suggestion
	size_t i = 0;
	static bool trigger_ctx[TRIGGER_NUM_MAX] /* Macro to define based end and start section pointers */

	STRUCT_SECTION_FOREACH(app_fp_adv_trigger, adv_trigger) {
		__ASSERT_NO_MSG(i < TRIGGER_NUM_MAX);
		adv_trigger->ctx = &trigger_ctx[i++];
	}

	err = bt_fast_pair_fmdn_info_cb_register(&fmdn_info_cb);
```

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -426,57 +454,58 @@ void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
...
+	/* Switch the advertising mode only on change. */
+	if (requested_mode != fp_adv_mode) {
+		LOG_INF("Fast Pair: advertising: %sabling", requested_adv_on ? "en" : "dis");
+		fp_adv_mode_set(requested_mode);
+	}
 }
 
-static void fp_adv_request_handle(enum app_ui_request request)
+void app_fp_adv_request(struct app_fp_adv_trigger *trigger, bool enable)
 {
```
kapi-no, line 479: do not allow to call this in the uninitialized state

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -259,6 +262,11 @@ static bool fp_adv_rpa_expired(struct bt_le_ext_adv *adv)
 			(k_uptime_delta(&uptime) / MSEC_PER_SEC));
 	}
 
+	if (!is_enabled) {
+		LOG_INF("Fast Pair: RPA rotation blocked as the fp_adv module is disabled");
+		return false;
+	}
```
kapi-no, line 268: This is not needed as we still create advertising sets dynamically. We allocate it in the _enable function and release it in the _disable function. The newly allocated set will get the new RPA with the valid status, so we do not need to do this.

This callback will also never get called in the `!is_enabled` (disabled) state as the advertising set is deallocated.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -426,57 +454,58 @@ void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
...
+{
+	fp_adv_mode_set(fp_adv_mode);
+}
+
+void app_fp_adv_rpa_rotation_suspend(bool suspended)
+{
+	fp_adv_rpa_rotation_suspended = suspended;
+}
+
 int app_fp_adv_id_set(uint8_t id)
```
kapi-no, line 506: Consider also setting the ID after the app_fp_adv_init(), we should require init to be called before any other function

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -343,31 +370,29 @@ static void fmdn_provisioning_state_changed(bool provisioned)
 			FACTORY_RESET_TRIGGER_KEY_STATE_MISMATCH,
 			K_SECONDS(FACTORY_RESET_DELAY));
 
+		first_state_change_after_bootup = false;
```
kapi-no, line 373: This is not the best way to do this as the user needs to keep in mind to add this line in case of another early return.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -426,57 +454,58 @@ void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
...
+		fp_adv_request_start();
+	} else {
+		LOG_WRN("Fast Pair: advertising module is disabled");
 	}
 }
 
+void app_fp_adv_refresh(void)
+{
+	fp_adv_mode_set(fp_adv_mode);
+}
```
mkapala-nordic, line 499: consider adding guard to execute it only if the adv mode is in disc or not disc adv

samples/bluetooth/fast_pair/locator_tag/src/main.c
```diff
@@ -301,15 +331,14 @@ static void fmdn_clock_synced(void)
...
 	}
 }
 
 static void fmdn_provisioning_state_changed(bool provisioned)
 {
-	static bool is_first_state_changed_cb = true;
-
-	enum app_fp_adv_mode fp_adv_mode;
+	/* Used to determine if the clock sync is required. */
+	static bool first_state_change_after_bootup = true;
```
kapi-no, line 341: consider defining a function

```
static bool fmdn_provisioning_state_changed_first_cb_after_bootup(void)
{
	static bool first_state_change_after_bootup = true;
	bool is_first_state_change_after_bootup = first_state_change_after_bootup;
	
	first_state_change_after_bootup = false;

	return is_first_state_change_after_bootup;
}
```

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,51 @@ enum app_fp_adv_mode {
...
+ *  @param trigger Trigger identifier.
+ *  @param enable true to request to turn on the Fast Pair advertising.
+ *		  false to remove request to turn on the Fast Pair advertising.
+ */
+void app_fp_adv_request(struct app_fp_adv_trigger *trigger, bool enable);
+
+/** Refresh the Fast Pair advertising.
+ *
+ *  This function is used to manually trigger the refreshing of the Fast Pair
+ *  advertising data to ensure that the advertising data is up to date.
```
kapi-no, line 71: Write a cautionary note about using this API with care as updating the advertising payload asynchronously to the RPA address and the FMDN extension.

samples/bluetooth/fast_pair/locator_tag/src/main.c
kapi-no: you missed the the update of the second log

samples/bluetooth/fast_pair/locator_tag/src/main.c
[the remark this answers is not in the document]
MarekPieta: Avoiding early returns could help to make code easier to follow. Still, since it's a coding style preference, I would not insist,

samples/bluetooth/fast_pair/locator_tag/include/app_factory_reset.h
```diff
@@ -20,16 +20,31 @@
...
+	}
 
 /** Schedule the factory reset action.
  *
  * @param delay Time to wait before the factory reset action is performed.
- * @param cb    Callback to define the prepare action before the factory reset.
  */
-void app_factory_reset_schedule(k_timeout_t delay,
-				app_factory_reset_executed_cb cb);
+void app_factory_reset_schedule(k_timeout_t delay);
```
MarekPieta, line 47: Consider returning an error when this one fails (it may be reasonable looking at newly introduced architecture changes)

MarekPieta: Similar for `cancel`

samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
```diff
@@ -103,6 +120,9 @@ static void factory_reset_perform(void)
 
 	factory_reset_state = APP_FACTORY_RESET_STATE_IN_PROGRESS;
 
+	factory_reset_prepare();
+
+
```
MarekPieta: Please keep a single empty line here

samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
```diff
@@ -88,6 +87,24 @@ int bt_fast_pair_factory_reset_user_action_perform(void)
...
+		if (cbs->prepare) {
+			cbs->prepare();
+		}
+	}
+}
+
+static void factory_reset_executed(void)
+{
+	STRUCT_SECTION_FOREACH(app_factory_reset_callbacks, cbs) {
+		if (cbs->executed) {
```
MarekPieta, line 82: Maybe we could pass information about error here as callback parameter. Thanks to that listeners would always be aware of factory reset state. Also, main app module could control the `k_panic` call on factory reset failure.

MarekPieta: Having multiple callback may also allow us to untangle factory reset dependency on `fp_adv` and `fast_pair` subsystem - it would be reasonable to move those API calls to callback now I think

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,51 @@ enum app_fp_adv_mode {
...
+		.id = _id,						\
+	}
+
+/** Request turning on or off the Fast Pair advertising.
+ *
+ *  The Fast Pair advertising will be turned off only when no
+ *  trigger has requested to keep the Fast Pair advertising on.
+ *
+ *  If the current Fast Pair advertising mode is the same as the requested mode,
+ *  the Fast Pair advertising will not be refreshed.
```
MarekPieta, line 58: `advertising` -> `advertising payload`?

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,51 @@ enum app_fp_adv_mode {
...
+ *  Use the @ref app_fp_adv_refresh function to manually refresh the Fast Pair
```
MarekPieta, line 59: `advertising` -> `advertising payload`?

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
MarekPieta: We could assert proper `is_enabled` state then I think

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -405,14 +404,43 @@ BT_CONN_CB_DEFINE(conn_callbacks) = {
...
 	.provisioning_state_changed = fp_adv_provisioning_state_changed,
 };
 
-void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
+enum app_fp_adv_mode app_fp_adv_mode_get(void)
+{
+	return fp_adv_mode;
+}
+
+static bool fp_adv_trigger_is_enabled_any(void)
```
MarekPieta, line 432: `fp_adv_trigger_is_any_enabled` ?

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -426,57 +454,58 @@ void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
...
+	trigger->enabled = enable;
 
-		fp_advertising_start();
+	LOG_INF("Fast Pair: advertising request from trigger \"%s\": %s",
+		trigger->id, trigger->enabled ? "enable" : "disable");
+
+	if (is_enabled) {
+		fp_adv_request_start();
+	} else {
+		LOG_WRN("Fast Pair: advertising module is disabled");
```
MarekPieta, line 492: Do we need this warning log? (the request will eventually be handled so I think we can omit the log)

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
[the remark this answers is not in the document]
mkapala-nordic: > Or even better: can we use `void *ctx` instead of this field and assign the memory segments from the static array defined in the fp_adv.c source code?

Please see the https://github.com/nrfconnect/sdk-nrf/pull/16529#discussion_r1694874959

We could probably do some calculation of the index based on the trigger struct address and the STRUCT_SECTION_START address and later use it in bitmask in source code to not expose it (then the trigger number would be capped at for example 32 if used uint32_t, but would still require at least to add the ASSERT instead of the define to validate sizes).

However, I think that hiding the implementation is an overkill for the code at the sample level.
We can consider changes in the future if needed (for example if we will decide to move this module to the usecase layer). 
I'll just add the comment as suggested.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
MarekPieta: If we want this mapping, then maybe we could use bitmask to reduce memory consumption? Otherwise, I would stick to @mkapala-nordic's approach to simplify code.

samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
MarekPieta: `app_fp_adv_disable` may not be needed as part of `fp_adv` API too

MarekPieta: `app_fp_adv_is_ready` could also be removed from the `fp_adv` API

samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
```diff
@@ -122,21 +102,20 @@ static void factory_reset_perform(void)
...
-
-	/* Disable the Fast Pair subsystem if it is active. */
-	if (fast_pair_is_ready) {
-		err = bt_fast_pair_disable();
+	/* Disable the Fast Pair advertising module if it is active. */
+	if (fp_adv_is_ready) {
+		err = app_fp_adv_disable();
 		if (err) {
-			LOG_ERR("Factory Reset: bt_fast_pair_disable failed: %d", err);
+			LOG_ERR("Factory Reset: app_fp_adv_uninit failed (err %d)", err);
```
MarekPieta, line 109: `app_fp_adv_disable failed`

samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
```diff
@@ -150,20 +129,20 @@ static void factory_reset_perform(void)
...
 		if (err) {
-			LOG_ERR("Factory Reset: app_fp_adv_init failed (err %d)", err);
+			LOG_ERR("Factory Reset: bt_fast_pair_enable failed: %d", err);
 			goto finish;
 		}
 	}
 
-	/* Reenable the Fast Pair subsystem. */
+	/* Reenable the Fast Pair advertising module if it was active before the reset. */
 	if (fast_pair_is_ready) {
```
MarekPieta, line 142: Looking at the comment, shouldn't the check rely on `fp_adv_is_ready` ?

samples/bluetooth/fast_pair/locator_tag/src/main.c
MarekPieta: I am worried that this may make things even more complicated here (looking only at the function name I have no idea what is the purpose of calling it). Variable used to be was far more intuitive.

samples/bluetooth/fast_pair/locator_tag/src/main.c
MarekPieta: Avoiding early returns and using all of the information explicitly to assign `fp_adv` requests at the end could make this code easier to understand.

samples/bluetooth/fast_pair/locator_tag/src/main.c
mkapala-nordic: It simplifies by hiding the assignment management in the function, it is fine solution.
will rename function to better indicate what it is used for

samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
mkapala-nordic: Discussed, will create a follow-up based on this suggestions, for now we will keep the implementation as is.

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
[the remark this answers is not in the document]
kapi-no: I think that using sections to implement this is an overkill and made the API more complex. Additionally, we need to expose the trigger state to the external user. Our originally discussed approach with the `int app_fp_adv_trigger_register` (dynamic registration) and the `app_fp_adv_request` API approach was better.

samples/bluetooth/fast_pair/locator_tag/include/app_factory_reset.h
mkapala-nordic: IMO currently there is no need for it, I won't add it.

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,60 @@ enum app_fp_adv_mode {
...
+ *
+ *  @param _name Trigger structure name.
+ *  @param _id Trigger identifier.
+ */
+#define APP_FP_ADV_TRIGGER_REGISTER(_name, _id)				\
+	BUILD_ASSERT(_id != NULL);					\
+	static STRUCT_SECTION_ITERABLE(app_fp_adv_trigger, _name) = {	\
+		.id = _id,						\
+	}
+
```
kapi-no, line 57:
```suggestion
/** Fast Pair advertising trigger. */
/** Opaque type representing an advertising trigger */
struct app_fp_adv_trigger;

/** Register a trigger for Fast Pair advertising.
 *
 *  @param name Trigger name.
 *  @param trigger Trigger instance.
 *
 *  @return 0 if the operation was successful. Otherwise, a (negative) error code is returned.
 */
int app_fp_adv_trigger_register(const char *trigger_name, struct app_fp_adv_trigger **trigger)

```

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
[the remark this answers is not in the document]
kapi-no: See the following simplification:

https://github.com/nrfconnect/sdk-nrf/pull/16529#discussion_r1695159208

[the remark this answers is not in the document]
MarekPieta: Why is using section making this code more complex? It seems to significantly simplify trigger registration. It also helps to avoid misconfigurations.

If you want to keep `enable` information private to `app_fp_adv` we could map index in section to bit in bitmask kept internally in `app_fp_adv` (should be straightforward).

samples/bluetooth/fast_pair/locator_tag/include/app_factory_reset.h
MarekPieta: Generally upper layer should be aware that something went wrong (logging error may not be enough here I think). Since it used to be like that I would not insist on modifying it in scope of this PR, but generally it may need to be updated at some point.

samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
MarekPieta: Please link the follow-up ticket so I can resolve my comment here.

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
MarekPieta: Why should we register subscribers dynamically? It seems not to simplify anything (see my comment above for details).

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -405,14 +401,43 @@ BT_CONN_CB_DEFINE(conn_callbacks) = {
...
+	STRUCT_SECTION_FOREACH(app_fp_adv_trigger, adv_trigger) {
+		if (adv_trigger->enabled) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+static void fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
```
MarekPieta, line 440: `fp_advertising_start` is quite a confusing function name here, because under the hood it can actually be used to stop advertising

MarekPieta: Maybe we could name it `ble_stack_update`/`ble_stack_notify` (or we could flatten function calls here)?

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -527,12 +556,48 @@ int app_fp_adv_init(void)
...
+		LOG_ERR("Fast Pair: fp_adv module already disabled");
+		return 0;
+	}
+
+	is_enabled = false;
+
+	/* Suspend the requested advertising until the fp_adv module reinitializes. */
+	fp_adv_mode_set(APP_FP_ADV_MODE_OFF);
+
+	app_fp_adv_rpa_rotation_suspend(false);
```
MarekPieta, line 585: I wonder if we should actually clear it (inconsistent with handling fp_adv requests which are not automatically cleared)

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
MarekPieta: Improving naming could significantly simplify understanding the code here

MarekPieta (changes requested):

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -426,57 +451,62 @@ void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
...
-			}
-		}
+	trigger->enabled = enable;
 
-		fp_advertising_start();
+	LOG_INF("Fast Pair: advertising request from trigger \"%s\": %s",
+		trigger->id, trigger->enabled ? "enable" : "disable");
+
+	if (is_enabled) {
+		fp_adv_request_start();
```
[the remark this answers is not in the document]
mkapala-nordic: if `is_enabled=false` advertising is disabled and the advertising set is not created so no -> the var refers to the module itself

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
mkapala-nordic: Let's discuss it.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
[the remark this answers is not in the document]
MarekPieta: You are right. I confused this with function argument. Please disregard this comment.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
MarekPieta: Discussed during standup. We will move it to factory reset prepare callback of main.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
mkapala-nordic: Agreed but I would rather not to refactor whole module if not needed as I am not touching directly those function.
Discussed to rename `fp_advertising_start` -> `fp_advertising_update` and `fp_adv_start_internal` -> `bt_stack_advertising_update`

samples/bluetooth/fast_pair/locator_tag/CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ target_sources(app PRIVATE
 
 zephyr_linker_sources(SECTIONS src/factory_reset_callbacks.ld)
 
+zephyr_linker_sources(DATA_SECTIONS src/fp_adv_trigger.ld)
```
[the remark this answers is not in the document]
mkapala-nordic: Let's leave as is

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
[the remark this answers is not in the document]
mkapala-nordic: Changed to bitmask as discussed

samples/bluetooth/fast_pair/locator_tag/include/app_factory_reset.h
mkapala-nordic: Added as a part of the follow-up: https://nordicsemi.atlassian.net/browse/NCSDK-28552

samples/bluetooth/fast_pair/locator_tag/src/factory_reset.c
mkapala-nordic: https://nordicsemi.atlassian.net/browse/NCSDK-28552

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -405,78 +402,131 @@ BT_CONN_CB_DEFINE(conn_callbacks) = {
...
 	.provisioning_state_changed = fp_adv_provisioning_state_changed,
 };
 
-void app_fp_adv_mode_set(enum app_fp_adv_mode adv_mode)
+enum app_fp_adv_mode app_fp_adv_mode_get(void)
+{
+	return fp_adv_mode;
+}
+
+static int fp_adv_trigger_idx_get(struct app_fp_adv_trigger *trigger)
```
kapi-no, line 430: let's use the assert version from before. After all, you simply assert the error code in the app_fp_adv_request

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,54 @@ enum app_fp_adv_mode {
...
+ *		  false to remove request to turn on the Fast Pair advertising.
+ */
+void app_fp_adv_request(struct app_fp_adv_trigger *trigger, bool enable);
+
+/** Refresh the Fast Pair advertising payload.
+ *
+ *  This function is used to manually trigger the refreshing of the Fast Pair
+ *  advertising data to ensure that the advertising data is up to date.
+ *
+ *  Use with caution, as it updates the advertising payload asynchronously
```
kapi-no, line 73:
```suggestion
 *  Use with caution, as it updates the Fast Pair advertising payload asynchronously
```

mkapala-nordic:
Applied @kapi-no suggestions

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,54 @@ enum app_fp_adv_mode {
...
+ *  to the RPA address and the FMDN extension.
```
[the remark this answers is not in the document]
kapi-no: ` *  to the RPA address rotation and to the update of the FMDN advertising payload.`

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
```diff
@@ -32,11 +32,54 @@ enum app_fp_adv_mode {
...
+ */
+void app_fp_adv_request(struct app_fp_adv_trigger *trigger, bool enable);
+
+/** Refresh the Fast Pair advertising payload.
+ *
+ *  This function is used to manually trigger the refreshing of the Fast Pair
+ *  advertising data to ensure that the advertising data is up to date.
+ *
+ *  Use with caution, as it updates the Fast Pair advertising payload asynchronously
+ *  to the RPA address and the FMDN advertising payload.
```
kapi-no, line 74: See the exact phrashing here: https://github.com/nrfconnect/sdk-nrf/pull/16529#discussion_r1696998708

kapi-no (approved):

samples/bluetooth/fast_pair/locator_tag/include/app_fp_adv.h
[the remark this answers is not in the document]
mkapala-nordic: fixed

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -493,32 +528,38 @@ uint8_t app_fp_adv_id_get(void)
...
-		LOG_ERR("Fast Pair: fp_adv_set_teardown failed (err %d)", err);
+		LOG_ERR("Fast Pair: bt_fast_pair_fmdn_info_cb_register returned error: %d", err);
 		return err;
 	}
 
+	STRUCT_SECTION_COUNT(app_fp_adv_trigger, &trigger_cnt);
+	if (trigger_cnt > BITS_PER_VAR(fp_adv_request_bm)) {
+		LOG_ERR("Fast Pair: invalid size of the advertising request bitmap");
+		return -ENOMEM;
+	}
```
MarekPieta, line 546: Consider making it an assert (it's actually static configuration issue + caller can do nothing about it).

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
```diff
@@ -493,32 +528,38 @@ uint8_t app_fp_adv_id_get(void)
 	return fp_adv_param.id;
 }
 
-int app_fp_adv_uninit(void)
+int app_fp_adv_init(void)
 {
 	int err;
+	int trigger_cnt;
```
MarekPieta, line 534: Shouldn't it be `size_t` ?

MarekPieta (approved):

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
mkapala-nordic: It doesn't really matter which type we will use here.
Doubt that the section size would be > `2^31-1`.
In the zephyr itself once is used `int`, once `size_t` and once `uint32_t`. 
Let's leave it as is.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
mkapala-nordic: Seems resonable, changed.

samples/bluetooth/fast_pair/locator_tag/src/fp_adv.c
MarekPieta: Ok, I would not insist

MarekPieta (approved):

PR #16539, opened 2024-07-19
lib: modem: lte_net_if: Fix handling of missing IPv4

glarsennordic:
Currently, if IPv4 is disabled or was not assigned, the cellular network is incorrectly treated as
offline, even if an IPv6 gets assigned.

This patch fixes this behavior.

### DIFF
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -558,6 +558,7 @@ Modem libraries
 * :ref:`nrf_modem_lib_lte_net_if` library:
 
   * Added a log warning suggesting a SIM card to be installed if a UICC error is detected by the modem.
+  * Fixed a bug causing the cell network to be treated as offline if IPv4 is not assigned.
 
 * :ref:`modem_info_readme` library:
 
--- a/lib/nrf_modem_lib/lte_net_if/lte_net_if.c
+++ b/lib/nrf_modem_lib/lte_net_if/lte_net_if.c
@@ -195,21 +195,22 @@ static void on_pdn_activated(void)
 #if CONFIG_NET_IPV4
 	int ret;
 
+	/* If IPv4 is enabled, check whether modem has been assigned an IPv4. */
 	ret = lte_ipv4_addr_add(iface_bound);
+
 	if (ret == -ENODATA) {
 		LOG_WRN("No IPv4 address given by the network");
-		return;
 	} else if (ret) {
 		LOG_ERR("ipv4_addr_add, error: %d", ret);
 		fatal_error_notify_and_disconnect();
 		return;
 	}
 
-	update_has_pdn(true);
-
 #endif /* CONFIG_NET_IPV4 */
 
 	/* IPv6 is updated on a aseparate event */
+
+	update_has_pdn(true);
 }
 
 static void on_pdn_deactivated(void)
@@ -248,8 +249,6 @@ static void on_pdn_ipv6_up(void)
 		fatal_error_notify_and_disconnect();
 		return;
 	}
-
-	update_has_pdn(true);
 }
 
 static void on_pdn_ipv6_down(void)


glarsennordic:
I need to do some additional testing of this, but I am pretty sure this will work since, once again, it is a simple change.

lib/nrf_modem_lib/lte_net_if/lte_net_if.c
```diff
@@ -195,21 +195,20 @@ static void on_pdn_activated(void)
 #if CONFIG_NET_IPV4
 	int ret;
 
+	/* If IPv4 is enabled, check whether modem has been assigned an IPv4. */
 	ret = lte_ipv4_addr_add(iface_bound);
+
```
plskeggs, line 200: Should there be a similar warning here `#if CONFIG_NET_IPV6` and `lte_ipv6_addr_add()` fails with `ENODATA` (assuming there is such a function and response)?

plskeggs (approved):

lib/nrf_modem_lib/lte_net_if/lte_net_if.c
glarsennordic: Interestingly, nope -- a separate event alerts to IPv6 assignment

eivindj-nordic: I'm wondering, with `PDN_EVENT_IPV6_UP` as a separate event that calls `lte_ipv6_addr_add()` and `update_has_pdn(true)`, why is the `update_has_pdn(true)` needed here for IPv6? Shouldn't it be set to online when we get the  `PDN_EVENT_IPV6_UP` event?

glarsennordic: Ah, you are right,  I do see now that `on_pdn_ipv6_up` calls `update_has_pdn(true)`

But I am surprised to see that. I thought `PDN_EVENT_IPV6_UP`/`PDN_EVENT_IPV6_DOWN` can only be relied upon to detect IPv6 availability

While IPv6 being UP does guarantee PDN is up, it is possible for IPv6 to go down without the PDN going down -- so PDN down must be detected with `PDN_EVENT_DEACTIVATED`. Therefore, it stands to reason (to my mind) that PDN up should be detected only based on `PDN_EVENT_ACTIVATED`.

In other words, I _**think**_ we should remove the `update_has_pdn(true)` from `on_pdn_ipv6_up`

The problem is, the exact conditions under which these PDN events are fired is not well documented, and I don't have a good way to properly test the various edge-cases (IPv6-only, IPv4-only, IPv6 gain -> IPv4 gain -> IPv6 loss, etc), so I don't have a way to verify that my suggestion here is definitely correct.

I see that you are the one who initially added the `update_has_pdn(true)` call to `on_pdn_ipv6_up` -- is my argument here convincing to you?

eivindj-nordic: Based on the description in the AT commands guide and what I can see from the PDN sample with some modifications I agree. We get the `PDN_EVENT_ACTIVATED` event with PDP type IPV6 upon activation, and `PDN_EVENT_DEACTIVATED` upon deactivation.

glarsennordic: Thanks for verifying it with some direct testing -- I'll go ahead and make that change, then!

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -451,6 +451,7 @@ Modem libraries
 * :ref:`nrf_modem_lib_lte_net_if` library:
 
   * Added a log warning suggesting a SIM card to be installed if a UICC error is detected by the modem.
+  * Fixed bug causing cell network to be treated as offline if IPv4 not assigned.
```
b-gent, line 454:
```suggestion
  * Fixed a bug causing the cell network to be treated as offline if IPv4 is not assigned.
```

eivindj-nordic (approved):

lemrey (approved):
Thanks

glarsennordic:
I finally figured out a way to get my device onto IPv6 and confirmed this works

divipillai (approved):

PR #16540, opened 2024-07-22
scripts: ci: tags: add common build system code for dep

nordic-piks:
- zephyr/cmake/
- zephyr/share/
- nrf/cmake/
- nrf/sysbuild/

### DIFF
--- a/scripts/ci/tags.yaml
+++ b/scripts/ci/tags.yaml
@@ -46,52 +46,51 @@
 
 bluetooth:
   files:
-    - zephyr/drivers/bluetooth/
-    - zephyr/subsys/bluetooth/
-    - zephyr/include/zephyr/bluetooth/
-    - zephyr/samples/bluetooth/hci_ipc/
-    - nrfxlib/softdevice_controller/include/
-    - nrfxlib/softdevice_controller/lib/
-    - nrfxlib/mpsl/include/
-    - nrfxlib/mpsl/lib/
-    - nrf/subsys/bluetooth/
-    - nrf/include/bluetooth/
-    - nrf/subsys/nrf_rpc/
-    - nrf/subsys/mpsl/
-    - nrf/drivers/mpsl/
-    - nrf/samples/bluetooth/
-    - nrf/tests/bluetooth/
-    - nrf/tests/subsys/bluetooth/
-    - nrf/samples/nrf5340/multiprotocol_rpmsg/
     - modules/crypto/mbedtls/
     - modules/crypto/oberon-psa-crypto/
     - modules/crypto/tinycrypt/
+    - nrf/applications/ipc_radio/
+    - nrf/cmake/
+    - nrf/drivers/mpsl/
+    - nrf/include/bluetooth/
+    - nrf/samples/bluetooth/
+    - nrf/samples/nrf5340/multiprotocol_rpmsg/
+    - nrf/subsys/bluetooth/
+    - nrf/subsys/mpsl/
+    - nrf/subsys/nrf_rpc/
     - nrf/subsys/nrf_security/
+    - nrf/sysbuild/
+    - nrf/tests/bluetooth/
+    - nrf/tests/subsys/bluetooth/
     - nrfxlib/crypto/
+    - nrfxlib/mpsl/include/
+    - nrfxlib/mpsl/lib/
+    - nrfxlib/softdevice_controller/include/
+    - nrfxlib/softdevice_controller/lib/
+    - zephyr/cmake/
+    - zephyr/drivers/bluetooth/
+    - zephyr/include/zephyr/bluetooth/
+    - zephyr/samples/bluetooth/hci_ipc/
+    - zephyr/share/
     - zephyr/soc/nordic/
+    - zephyr/subsys/bluetooth/
     - zephyr/subsys/settings/
 
 ci_samples_net: &ci_net
   files:
-    - nrf/samples/net/
-    - nrf/tests/subsys/net/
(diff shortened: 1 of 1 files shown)


nordic-segl (approved):

scripts/ci/tags.yaml
```diff
@@ -50,10 +50,16 @@ bluetooth:
     - zephyr/subsys/bluetooth/
     - zephyr/include/zephyr/bluetooth/
     - zephyr/samples/bluetooth/hci_ipc/
+    - zephyr/soc/nordic/
+    - zephyr/subsys/settings/
+    - zephyr/cmake/
+    - zephyr/share/
```
MarekPieta, line 56: Do we want to keep it ordered alphabetically?

scripts/ci/tags.yaml
```diff
@@ -138,6 +146,8 @@ ci_samples_wifi:
 
 ci_applications_matter: &ci_matter
   files:
```
MarekPieta, line 148: Shouldn't we update all of the tags (e.g.`ci_samples_wifi`, `ci_samples_openthread` and `ci_samples_nfc` are not updated) ?

MarekPieta (changes requested):

scripts/ci/tags.yaml
nordic-piks: Will fix - will sort all tag's dependencies.

scripts/ci/tags.yaml
nordic-piks: Intentionally not, I added them only for those big one (with a lot of number of tests).
There is another method to execute all tests (adding label for PR).

miha-nordic (approved):

scripts/ci/tags.yaml
MarekPieta: Thanks for clarification

scripts/ci/tags.yaml
```diff
@@ -46,52 +46,50 @@
...
+    - nrf/tests/subsys/bluetooth/
     - nrfxlib/crypto/
+    - nrfxlib/mpsl/include/
+    - nrfxlib/mpsl/lib/
+    - nrfxlib/softdevice_controller/include/
+    - nrfxlib/softdevice_controller/lib/
+    - zephyr/cmake/
+    - zephyr/drivers/bluetooth/
+    - zephyr/include/zephyr/bluetooth/
+    - zephyr/samples/bluetooth/hci_ipc/
```
MarekPieta, line 72: You may also add dependency onto `nrf/applications/ipc_radio` here I think (IIRC it's used by default for Bluetooth on multi-core platforms). The `zephyr/samples/bluetooth/hci_ipc/` is no longer used by default so that maybe we could remove HCI IPC sample dependency too. @dchat-nordic, could you confirm?

scripts/ci/tags.yaml
```diff
@@ -107,261 +105,288 @@ ci_samples_net: &ci_net
...
     - nrfxlib/nrf_802154/serialization/
     - nrfxlib/nrf_802154/sl/
     - nrfxlib/nrf_wifi/
-    - nrfxlib/crypto/
+    - nrfxlib/openthread/
+    - zephyr/cmake/
+    - zephyr/drivers/net/
     - zephyr/include/dfu/
     - zephyr/include/mgmt/mcumgr/
+    - zephyr/samples/bluetooth/hci_ipc/
```
MarekPieta, line 185: @kkasperczyk-no, do you use HCI IPC sample for Matter? NCS already uses IPC radio application by default, so this one seems outdated

scripts/ci/tags.yaml
```diff
@@ -107,261 +105,288 @@ ci_samples_net: &ci_net
...
-    - modules/crypto/oberon-psa-crypto/
-    - modules/crypto/tinycrypt/
-    - modules/lib/open-amp/open-amp/
-    - modules/lib/zcbor/
-    - bootloader/mcuboot/
-    - sidewalk/
 
 find_my:
   files:
     - bootloader/mcuboot/
```
MarekPieta, line 388: Can we also add the build system dependencies to `find-my` please?

MarekPieta (approved):

scripts/ci/tags.yaml
nordic-piks: Will add.

scripts/ci/tags.yaml
dchat-nordic: The ipc_radio is used by default in almost all NCS. I think it should be added to bluetooth and ieee802154 (zigbee, thread etc) tags.

The hci_ipc is still used by a couple samples AFAIK: 2 matter samples and SUIT samples. I don't really know how they are tagged in this context. In my opinion if that's possible to leave those samples checked for hci_ipc it's fine to remove the hci_ipc for everything else.

scripts/ci/tags.yaml
dchat-nordic: I believe they still use the hci_ipc for some wifi coex sample variant.

scripts/ci/tags.yaml
nordic-piks: Added     - nrf/applications/ipc_radio/ for mentioned tags.

scripts/ci/tags.yaml
nordic-piks: Will keep it.

dchat-nordic (approved):

MarekPieta (approved):

PR #16546, opened 2024-07-22
samples: npm1300_one_button: Fix irq allocation for nRF52840 DK.

aasinclair:
Pin allocation for PMIC interrupt output on nRF52840 DK has been fixed.  It was previously allocated to a pin that is not available on a pin header.


WalkingTalkingPotato (changes requested):
Needs line 20 to be changed in nrf52840dk_nrf52840.overlay:
`host-int-gpios = <&gpio1 12 0>;`

Then it works as documented (tested).

shanthanordic (approved):

greg-fer (approved):

PR #16548, opened 2024-07-22
Bluetooth: Get rid of HCI boilerplate code for vendor specific commands

rugeGerritsen:
The new header `hci_vs_sdc.h` provides APIs that wraps handling of HCI commands creating code that is simpler to understand and maintain. This PR updates some existing code to use these APIs.

### DIFF
--- a/subsys/caf/modules/ble_state.c
+++ b/subsys/caf/modules/ble_state.c
@@ -14,9 +14,8 @@
 #include <zephyr/bluetooth/hci.h>
 
 #include <caf/events/ble_common_event.h>
-
 #ifdef CONFIG_CAF_BLE_USE_LLPM
-#include "sdc_hci_vs.h"
+#include <bluetooth/hci_vs_sdc.h>
 #endif /* CONFIG_CAF_BLE_USE_LLPM */
 
 #define MODULE ble_state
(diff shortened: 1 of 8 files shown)


rick1082 (approved):
Great to have this PR! The ISO read TX timestamp function is simpler!

subsys/caf/modules/ble_state.c
```diff
@@ -16,7 +16,7 @@
 #include <caf/events/ble_common_event.h>
 
 #ifdef CONFIG_CAF_BLE_USE_LLPM
-#include "sdc_hci_vs.h"
+#include <bluetooth/hci_vs_sdc.h>
 #endif /* CONFIG_CAF_BLE_USE_LLPM */
```
MarekPieta, line 20: Please remove `#ifdef` if no longer needed

subsys/caf/modules/ble_state.c
```diff
@@ -313,15 +313,11 @@ static void bt_ready(int err)
 	LOG_INF("Bluetooth initialized");
 
 #ifdef CONFIG_CAF_BLE_USE_LLPM
```
MarekPieta, line 315: Please use `if (IS_ENABLED(CONFIG_CAF_BLE_USE_LLPM) {` instead of `#ifdef` if possible after your change is applied. In general we try to avoid `#ifdef` if possible to keep compiler checks (unused could would be opted out anyway)

MarekPieta: Similarly for `ble_conn_params.c`

alwa-nordic (approved):

MarekPieta (changes requested):

subsys/caf/modules/ble_state.c
rugeGerritsen: It is still needed. This header includes sdc_hci_vs.h

PR #16549, opened 2024-07-22
Bluetooth: Expose API to obtain connection event anchor point

rugeGerritsen:
This is a proprietary SDC feature which allows the controller to produce reports for all ACL connections whenever	  a connection anchor point is updated. The reporting is enabled through a HCI command.

This information can be utilized for time synchronization between a central and peripheral.

### DIFF
--- a/include/bluetooth/hci_vs_sdc.h
+++ b/include/bluetooth/hci_vs_sdc.h
@@ -392,6 +392,16 @@ int hci_vs_sdc_scan_accept_ext_adv_packets_set(
  */
 int hci_vs_sdc_set_role_priority(const sdc_hci_cmd_vs_set_role_priority_t *params);
 
+/** @brief Set Event Start Task.
+ *
+ * For the complete API description, see sdc_hci_cmd_vs_set_event_start_task().
+ *
+ * @param[in]  params Input parameters.
+ *
+ * @return 0 on success or negative error value on failure.
+ */
+int hci_vs_sdc_set_event_start_task(const sdc_hci_cmd_vs_set_event_start_task_t *params);
+
 /** @brief Connection Anchor_Point Update Event Reports enable.
  *
  * For the complete API description,
--- a/include/bluetooth/hci_vs_sdc.h
+++ b/include/bluetooth/hci_vs_sdc.h
@@ -392,6 +392,18 @@ int hci_vs_sdc_scan_accept_ext_adv_packets_set(
  */
 int hci_vs_sdc_set_role_priority(const sdc_hci_cmd_vs_set_role_priority_t *params);
 
+/** @brief Connection Anchor_Point Update Event Reports enable.
+ *
+ * For the complete API description,
+ * see sdc_hci_cmd_vs_conn_anchor_point_update_event_report_enable().
+ *
+ * @param[in]  params Input parameters.
+ *
+ * @return 0 on success or negative error value on failure.
+ */
+int hci_vs_sdc_conn_anchor_point_update_event_report_enable(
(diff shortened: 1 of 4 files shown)


include/bluetooth/hci_vs_sdc.h
```diff
@@ -392,6 +392,28 @@ int hci_vs_sdc_scan_accept_ext_adv_packets_set(
  */
 int hci_vs_sdc_set_role_priority(const sdc_hci_cmd_vs_set_role_priority_t *params);
 
+/** @brief Set Event Start Task.
+ *
+ * For the complete API description, see sdc_hci_cmd_vs_set_event_start_task().
```
seko-nordic, line 397:
```suggestion
 * For the complete API description, see @ref sdc_hci_cmd_vs_set_event_start_task().
```
Will that work? would be nicer to have clickable link.
Actually I was not able to find generated docs for it here https://developer.nordicsemi.com/nRF_Connect_SDK_dev/doc/PR-16549/nrf/index.html. Do you know where is it placed?

jostn (approved):

include/bluetooth/hci_vs_sdc.h
rugeGerritsen: Refs from sdk-nrf to nrfxlib do not work AFAIK. I've asked a question about it here: https://github.com/nrfconnect/sdk-nrf/pull/16550#discussion_r1687766121.
@greg probably knows this better

subsys/bluetooth/controller/Kconfig
```diff
@@ -394,6 +394,17 @@ config BT_CTLR_SDC_EVENT_TRIGGER
...
+config BT_CTLR_SDC_CONN_ANCHOR_POINT_REPORT
+	bool "Connection event anchor point reports"
+	depends on BT_CONN
+	help
+	  Enable support for obtaining connection anchor point reports.
+	  This is a proprietary SDC feature which allows the controller
+	  to produce timestamps for ACL connection anchor points.
+	  A new timestamp is generated whenever a connection's anchor point changes.
+	  The reporting is enabled through an HCI command.
+	  See Core_v5.4, Vol 6, Part B, Section 4.5.1.
```
seko-nordic, line 406:
```suggestion
	  See Bluetooth Core Specification, Vol 6, Part B, Section 4.5.1.
```
I think it's guaranteed now that section numbers will not change in the future spec versions.
This way we don't need to update it when new spec comes and avoid questions/complains from customers that we are not on the latest spec.

rugeGerritsen: Are you ok with changing this in a separate PR to unblock the sample application PR? Then we can address all occurrences of similar patterns.

include/bluetooth/hci_vs_sdc.h
```diff
@@ -392,6 +392,28 @@ int hci_vs_sdc_scan_accept_ext_adv_packets_set(
...
+ *
+ * @param[in]  params Input parameters.
+ *
+ * @return 0 on success or negative error value on failure.
+ */
+int hci_vs_sdc_set_event_start_task(const sdc_hci_cmd_vs_set_event_start_task_t *params);
+
+/** @brief Connection Anchor_Point Update Event Reports enable.
```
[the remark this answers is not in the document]
rugeGerritsen: It is intentional, it is taken from https://github.com/nrfconnect/sdk-nrfxlib/blob/main/softdevice_controller/include/sdc_hci_vs.h#L1726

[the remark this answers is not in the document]
olivier-le-sage: ok. That should probably be fixed, it looks a bit unprofessional

olivier-le-sage (approved):

PR #15146, opened 2024-04-30
bluetooth: fast_pair: Add bond management

alstrzebonski:
Implements Fast Pair bond management. Associates bonds with Account Keys
and removes bonds when associated Account Keys are removed.

Short solution description: https://nordicsemi.atlassian.net/wiki/spaces/SPG/pages/269811994/New+Fast+Pair+storage+-+chosen+design#1.-Backwards-compatibility---standard-implementation%3A (see point 1.3. Bond information)

Detailed bond management flowchart: https://nordicsemi.atlassian.net/wiki/spaces/SPG/pages/672793094
bond_flow drawio (1) (1)

Jira: NCSDK-18829

TODO in scope of this PR:
- [x] Align Fast Pair storage tests (currently there are build failures)
- [x] Introduce Kconfig option that will enable/disable bond tracking functionality (should also fix tests build failures)
- [x] Perform detailed tests of all edge cases
- [x] Improve function documentation in header
TODO in separate PR:
- [ ] Add documentation.

PR #16554, opened 2024-07-22
Fast Pair improvements

alstrzebonski:
Adds Fast Pair improvements needed for https://github.com/nrfconnect/sdk-nrf/pull/15146.


zycz (approved):

alstrzebonski:
Rebased on main

subsys/bluetooth/services/fast_pair/fp_storage/fp_storage_ak.c
```diff
@@ -474,6 +474,13 @@ static int fp_storage_ak_reset(void)
 	int err;
 	bool was_enabled = is_enabled;
 
+	if (was_enabled) {
+		err = fp_storage_ak_uninit();
```
MarekPieta, line 478: Is it done to ensure that module is in uninitialized state after `fp_storage_ak_reset` returns with error?

If we want to apply it, then we should consider to also add it for other `fp_storage` modules (to ensure consistency).

MarekPieta (changes requested):

subsys/bluetooth/services/fast_pair/fp_storage/fp_storage_ak.c
alstrzebonski: Yes. It's also cleaner to call uninit if we want to call init after the reset.

alstrzebonski:
Adressed review comment.

zycz (approved):

MarekPieta (approved):

alstrzebonski:
Rebased on main.

PR #16555, opened 2024-07-22
Add more tags

nordic-piks:


zycz (approved):
CAF part looks good

nordic-segl (approved):

scripts/ci/tags.yaml
```diff
@@ -804,3 +804,49 @@ ci_applications_nrf_desktop:
...
+    - nrf/subsys/nrf_security/
+    - nrf/tests/subsys/emds/
+    - zephyr/subsys/bluetooth/
+
+ci_tests_lib_nrf_fuel_gauge:
+  files:
+    - nrf/tests/lib/nrf_fuel_gauge/
+    - nrfxlib/nrf_fuel_gauge/
+
+ci_tests_lib_at_cmd_parser:
```
trantanen, line 833: `at_cmd_parser` is going to be deprecated for 2.8.0 and eventually removed but I guess it should be here until then. Not sure anybody will remove it here although it doesn't break anything.

trantanen (approved):

scripts/ci/tags.yaml
nordic-piks: Let's keep it until test is in code.
Nothing will happen when code will be remove and tag will stay.

miha-nordic (approved):

PR #16556, opened 2024-07-22
cracen: Fix handling of key_bits=0

vili-nordic:
PSA specifies that key bits should be determined from the key buffer when bits=0 is specified in attributes at import.

test_crypto: PR-646

### DIFF
--- a/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
+++ b/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
@@ -38,139 +38,101 @@ extern const uint8_t cracen_N3072[384];
 
 extern nrf_security_mutex_t cracen_mutex_symmetric;
 
-static psa_status_t check_brainpool_alg_and_key_bits(psa_algorithm_t alg, size_t key_bits)
-{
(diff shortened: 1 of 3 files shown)


subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
```diff
@@ -38,79 +38,101 @@ extern const uint8_t cracen_N3072[384];
...
-	default:
-		return PSA_ERROR_INVALID_ARGUMENT;
+	size_t valid_bits[] = {192, 225, 256};
+
+	if (*key_bits == 0) {
+		GET_BITS(key_buffer_size, valid_bits);
 	}
+	CHECK_BITS(key_bits, valid_bits);
+
+	return PSA_ERROR_INVALID_ARGUMENT;
```
Vge0rge, line 87: Here we return invalid argument even for the 225 which used to be not-supported right? It's probably an edge case but did you intend to do this?

vili-nordic: tried simplifying things here a bit

Vge0rge (approved):

frkv (approved):
LGTM

PR #16559, opened 2024-07-22
samples: matter: Use the generic ConnectivityManager methods

markaj-nordic:
Limit the amount of ifdefs by utilizing the L2 agnostic network state checkers from Matter ConnectivityManager.


samples/matter/common/src/app/matter_init.cpp
```diff
@@ -147,6 +148,36 @@ CHIP_ERROR ConfigureThreadRole()
...
+	return error;
+}
+
+#elif defined(CONFIG_CHIP_WIFI)
+
+CHIP_ERROR InitNetworkingStack()
+{
+	CHIP_ERROR error{ CHIP_NO_ERROR };
+
+	if (!sLocalInitData.mNetworkingInstance) {
```
Damian-Nordic, line 171: Is this check necessary? If someone wants to postpone the network commissioning instance for whatever reason, that should be OK, right? Especially that this defaults to a non-null instance, so a user can't miss this setting unintentionally.

Damian-Nordic (approved):

samples/matter/common/src/app/matter_init.cpp
markaj-nordic: Would it be ok to keep the check before calling the Init() but do not report an error if the instance is NULL?

Damian-Nordic: Yes, this is what I meant :)

samples/matter/common/src/app/matter_init.cpp
```diff
@@ -179,23 +206,8 @@ void DoInitChipServer(intptr_t /* unused */)
...
-		sInitResult = CHIP_ERROR_INTERNAL;
-		VerifyInitResultOrReturn(sInitResult, "No valid commissioning instance");
-	}
-	sLocalInitData.mNetworkingInstance->Init();
-#else
-	sInitResult = CHIP_ERROR_INTERNAL;
-	VerifyInitResultOrReturn(sInitResult, "No valid L2 network backend selected");
-#endif /* CONFIG_NET_L2_OPENTHREAD */
+	/* Initialize L2 networking backend. */
+	sInitResult = InitNetworkingStack();
```
kkasperczyk-no, line 210: Shouldn't we verify the `sInitResult` value?

kkasperczyk-no (approved):
LGTM, left one comment to consider

ArekBalysNordic (approved):

PR #16576, opened 2024-07-23
sysbuild: Split variant image logic

57300:
A variant image copies several properties of the original image. This includes cache variables, which should be copied as late as possible, because more of these variables can be set during sysbuild processing - for example, through the original image's `sysbuild.cmake`, which may be sourced after `ExternalNcsVariantProject_Add()` was already called.

The latest possible point in sysbuild would be a PRE_IMAGE_CMAKE hook before the variant image is configured. Create this hook in NCS and move the relevant logic there.

Ref: NCSDK-28495


nordicjm (approved):
Looks like a nice fix! Thanks

MarekPieta (approved):

nordicjm (approved):

PR #16586, opened 2024-07-23
nrf_security: Configuration optimizations

Vge0rge:
Check the commit messages

### DIFF
--- a/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
+++ b/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
@@ -637,11 +637,15 @@ static psa_status_t export_ecc_public_key_from_keypair(const psa_key_attributes_
 		if (key_buffer_size != sizeof(ikg_opaque_key)) {
 			return PSA_ERROR_INVALID_ARGUMENT;
 		}
-		priv_key =
-			si_sig_fetch_ikprivkey(sx_curve, ((ikg_opaque_key *)key_buffer)->owner_id);
-		data[0] = SI_ECC_PUBKEY_UNCOMPRESSED;
-		pub_key.key.eckey.qx = &data[1];
-		pub_key.key.eckey.qy = &data[1 + sx_pk_curve_opsize(sx_curve)];
+
+		if (IS_ENABLED(PSA_NEED_CRACEN_ECDSA_SECP_R1_256)) {
+			priv_key = si_sig_fetch_ikprivkey(sx_curve, *key_buffer);
+			data[0] = SI_ECC_PUBKEY_UNCOMPRESSED;
+			pub_key.key.eckey.qx = &data[1];
+			pub_key.key.eckey.qy = &data[1 + sx_pk_curve_opsize(sx_curve)];
+		} else {
+			return PSA_ERROR_NOT_SUPPORTED;
+		}
 	} else {
 
 		switch (psa_curve) {
--- a/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/sign.c
+++ b/subsys/nrf_security/src/drivers/cracen/cracenpsa/src/sign.c
@@ -100,15 +100,19 @@ static int cracen_signature_prepare_ec_prvkey(struct si_sig_privkey *privkey, ch
 		return status;
 	}
 
+	/* IKG supports one SECP256_R1 key */
 	if (PSA_KEY_LIFETIME_GET_LOCATION(psa_get_key_lifetime(attributes)) ==
 	    PSA_KEY_LOCATION_CRACEN) {
 		if (key_buffer_size != sizeof(ikg_opaque_key)) {
 			return SX_ERR_INVALID_ARG;
 		}
-		*privkey =
-			si_sig_fetch_ikprivkey(*sicurve, ((ikg_opaque_key *)key_buffer)->owner_id);
 
-		return status;
+		if (IS_ENABLED(PSA_NEED_CRACEN_ECDSA_SECP_R1_256)) {
+			*privkey = si_sig_fetch_ikprivkey(*sicurve, *key_buffer);
+			return status;
+		} else {
+			return SX_ERR_INCOMPATIBLE_HW;
+		}
 	}
 
 	if (key_buffer_size != PSA_BITS_TO_BYTES(psa_get_key_bits(attributes))) {


vili-nordic (approved):

frkv (approved):
LGTM

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
```diff
@@ -1725,6 +1725,7 @@ config PSA_NEED_CRACEN_CTR_DRBG_DRIVER
 	default y
 	select PSA_ACCEL_GENERATE_RANDOM
 	# Is needed for AES counter measurements
+	depends on PSA_WANT_GENERATE_RANDOM
```
tomi-font, line 1729: Move the comment above below this line?

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
```diff
@@ -1843,11 +1843,16 @@ config PSA_NEED_CRACEN_KEY_DERIVATION_DRIVER
 		   PSA_NEED_CRACEN_SP800_108_COUNTER_CMAC
 
 config PSA_NEED_CRACEN_KMU_DRIVER
+	bool
+	default y
+	depends on CRACEN_LIB_KMU
```
tomi-font, line 1849: Why loosening of the required Kconfig options here?

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
```diff
@@ -1085,7 +1085,8 @@ config PSA_NEED_CRACEN_KEY_TYPE_ECC_TWISTED_EDWARDS_255
 		   PSA_NEED_CRACEN_KEY_TYPE_ECC_KEY_PAIR_IMPORT_TWISTED_EDWARDS_255 || \
 		   PSA_NEED_CRACEN_KEY_TYPE_ECC_KEY_PAIR_EXPORT_TWISTED_EDWARDS_255 || \
 		   PSA_NEED_CRACEN_KEY_TYPE_ECC_KEY_PAIR_GENERATE_TWISTED_EDWARDS_255 || \
-		   PSA_NEED_CRACEN_KEY_TYPE_ECC_KEY_PAIR_DERIVE_TWISTED_EDWARDS_255
+		   PSA_NEED_CRACEN_KEY_TYPE_ECC_KEY_PAIR_DERIVE_TWISTED_EDWARDS_255 || \
+		   (PSA_WANT_ECC_TWISTED_EDWARDS_255 && PSA_NEED_CRACEN_KMU_DRIVER)
```
tomi-font, line 1089: The commit message for this is wrong and incomplete. Needs fixing.

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
```diff
@@ -637,11 +637,15 @@ static psa_status_t export_ecc_public_key_from_keypair(const psa_key_attributes_
 		if (key_buffer_size != sizeof(ikg_opaque_key)) {
 			return PSA_ERROR_INVALID_ARGUMENT;
 		}
-		priv_key =
-			si_sig_fetch_ikprivkey(sx_curve, ((ikg_opaque_key *)key_buffer)->owner_id);
-		data[0] = SI_ECC_PUBKEY_UNCOMPRESSED;
-		pub_key.key.eckey.qx = &data[1];
-		pub_key.key.eckey.qy = &data[1 + sx_pk_curve_opsize(sx_curve)];
+
+		if (IS_ENABLED(PSA_NEED_CRACEN_ECDSA_SECP_R1_256)) {
```
tomi-font, line 641: Minor: How about using an `#ifdef` instead? (Seems more clear to me when editor syntax highlighting works according to the configuration.)

subsys/nrf_security/configs/nrf-config.h.template
```diff
@@ -10,6 +10,10 @@
  */
 #ifndef MBEDTLS_CONFIG_FILE_H
 #define MBEDTLS_CONFIG_FILE_H
+/****************************************************************/
```
tomi-font, line 13: Why this line? Plus there's a typo in this commit's message.

tomi-font (dismissed):

subsys/nrf_security/configs/nrf-config.h.template
Vge0rge: This commit will now get removed since with the updates in nrf_security it is not needed.

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
Vge0rge: I don't get it sorry, the line that you are pointing is exactly after the comment.

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
Vge0rge: Because we want to allow using non-encrypted KMU keys and this has less requirements.

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
Vge0rge: Will not do that, we decided to use the IS_ENABLED so time before all around the cracen driver. So first of all I want to be consistent. But also it is more easily readable than ifdef in my opinion and it allows you to catch build failures at build time even if the condition is not enabled.

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
tomi-font: My wording might've been a bit confusing. I meant, `move the comment that is right now above this line below it?` because the comment is about AES and counter measures and it feels like it was addressing the line below unless I am missing something?

subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
tomi-font: It seems that it's actually a bit of a mix (there are some `#ifdef`s)... But fine. :slightly_frowning_face:

tomi-font (commented):
Just one remark left open.

subsys/nrf_security/src/drivers/cracen/psa_driver.Kconfig
Vge0rge: Hm, the aes countermeasure requires the PSA_WANT_GENERATE_RANDOM. So isn't this logical?

tomi-font: Ah okay. Fine then. It's just that this comment used to be addressed to the `depends on` statement.

tomi-font (approved):

Vge0rge:
@nrfconnect/ncs-co-build-system please review.

nordicjm (approved):

PR #16607, opened 2024-07-24
samples: Bluetooth: mesh: change shell backend

akredalen:
Fixed issue where boot banner was not printed due to the Zephyr kernel and Shell module both using UART as a backend. Mesh samples using Shell are now using RTT as a backend instead.

Removed unnecessary printk statement from Chat
sample.


omkar3141 (commented):
Ohh ... is this the solution? Then I'm afraid we will probably have to live with this issue. The RTT backend is not at all user friendly if you want to type many commands. That was the reason why we use UART as shell backend. If UART used as shell backend cannot print NCS boot banner then it is NCS wide issue that should perhaps be fixed in other ways.

PR #16627, opened 2024-07-25
applications: nrf_desktop: Define biggest supported HID report sizes

MarekPieta:
PR adds static definitions of the biggest supported HID input and output report sizes. PR also updates HID state and HID forward to use newly added definition of the biggest supported HID output report size

Jira: NCSDK-28375


applications/nrf_desktop/configuration/common/hid_report_desc.h
```diff
@@ -84,6 +84,34 @@ static const uint8_t output_reports[] = {
...
+	(IS_ENABLED(CONFIG_DESKTOP_HID_REPORT_CONSUMER_CTRL_SUPPORT) ?				\
+	 (REPORT_SIZE_CONSUMER_CTRL) : (0))
+#define _REPORT_BUFFER_SIZE_BOOT_MOUSE \
+	(IS_ENABLED(CONFIG_DESKTOP_HID_BOOT_INTERFACE_MOUSE) ? (REPORT_SIZE_MOUSE_BOOT) : (0))
+#define _REPORT_BUFFER_SIZE_BOOT_KEYBOARD \
+	(IS_ENABLED(CONFIG_DESKTOP_HID_BOOT_INTERFACE_KEYBOARD) ? (REPORT_SIZE_KEYBOARD_BOOT) : (0))
+
+/* Size of the biggest supported HID input report that is part of input reports map. */
+#define REPORT_BUFFER_SIZE_INPUT_REPORT								\
+	MAX(MAX(_REPORT_BUFFER_SIZE_MOUSE, _REPORT_BUFFER_SIZE_KEYBOARD_KEYS),			\
```
zycz, line 105: consider creating MAX6 macro:
#define MAX6(a, b, c, d, e, f) (MAX(MAX(MAX((a), (b)), MAX((c), (d))), MAX((e), (f))))

and using it for greater clarity

applications/nrf_desktop/configuration/common/hid_report_keyboard.h
```diff
@@ -23,6 +23,9 @@ extern "C" {
 #define REPORT_SIZE_KEYBOARD_KEYS	8 /* bytes */
 #define REPORT_SIZE_KEYBOARD_LEDS	1 /* bytes */
 
+/* Boot keyboard report uses the same format as keyboard keys report. */
+#define REPORT_SIZE_KEYBOARD_BOOT	REPORT_SIZE_KEYBOARD_KEYS
```
zycz, line 27: If both use the same size why do we need to add extra define? Cannot we use one for both cases?

MarekPieta: I wanted to make it more explicit. HID boot mouse report has its own definition for size too.
If the HID keyboard keys report format used by nRF Desktop changes at some point, the sizes would no longer be equal too.

applications/nrf_desktop/configuration/common/hid_report_desc.h
MarekPieta: The macro would be nice to have, but I think it should be part of Zephyr (not nRF Desktop application). We would also need to define `MAX_` variants for other number of arguments (or even make it `VA_ARGS` macro). Implementing it in generic way may take some work and I am not sure if it's worth to work on it on our end.

applications/nrf_desktop/configuration/common/hid_report_keyboard.h
zycz: maybe add this comment here?

applications/nrf_desktop/configuration/common/hid_report_desc.h
zycz: We can add just this one here and remove it once there is such macro in zephyr

MarekPieta: The header content is visible for application code. I am a bit against adding only `MAX6` here because of that (macro name would suggest that it's something more generic)

applications/nrf_desktop/configuration/common/hid_report_keyboard.h
MarekPieta: I already have a comment that the same format is used (it means sizes are equal too)

zycz: If the HID keyboard keys report format used by nRF Desktop changes at some point, the sizes would no longer be equal too.

zycz: After reading comment as it is now I wonder why to add this in the first place. Maybe if you don't want to add extra comment then remove existing one

applications/nrf_desktop/configuration/common/hid_report_desc.h
MarekPieta: On the other hand, providing `MAX6` suggests that other counts could be supported in a similar way. @kapi-no, @mkapala-nordic, @alstrzebonski, what do you think ?

applications/nrf_desktop/configuration/common/hid_report_keyboard.h
MarekPieta: > If the HID keyboard keys report format used by nRF Desktop changes at some point, the sizes would no longer be equal too.

It's defined as part of the same header file, so ensuring that these values are updated together should not be a problem. The comment is added because it's specific to nRF Desktop.

MarekPieta: @kapi-no, @mkapala-nordic, @alstrzebonski, what do you think ?

applications/nrf_desktop/configuration/common/hid_report_desc.h
zycz: don't know why having MAX6 suggest that others macros are supported

mkapala-nordic: If it is defined at the file level then why would it suggest that others are supported?
Personally it is fine for me either way, but prefer leaving it as is.

alstrzebonski: I like the MAX6 idea. It would be defined locally just in this file as a helper.

applications/nrf_desktop/configuration/common/hid_report_keyboard.h
mkapala-nordic: current version seems fine to me.

alstrzebonski: "Boot keyboard report **currently** uses the same format as keyboard keys report."?

mkapala-nordic (approved):

applications/nrf_desktop/configuration/common/hid_report_desc.h
MarekPieta: Keep in mind that C sources including the header would "see" the definition too. Also, the comment explicitly mentions how the macro works, so it should be quite understandable

applications/nrf_desktop/configuration/common/hid_report_keyboard.h
MarekPieta: `/* nRF Desktop's keyboard keys report currently uses the same format as boot keyboard reports. */`

applications/nrf_desktop/configuration/common/hid_report_desc.h
MarekPieta: > I like the MAX6 idea. It would be defined locally just in this file as a helper.

I tried to define `MAX_VA_ARGS` macro that would allow to select maximum from any number of arguments, but it seems tricky to do it in generic way (solution still does not work for me). I prefer not to define only `MAX6` macro here and stick with the existing approach if possible then.

zycz (approved):

mkapala-nordic (approved):

applications/nrf_desktop/configuration/common/hid_report_desc.h
```diff
@@ -84,6 +84,35 @@ static const uint8_t output_reports[] = {
...
+#define _REPORT_BUFFER_SIZE_CONSUMER_CTRL							\
+	(IS_ENABLED(CONFIG_DESKTOP_HID_REPORT_CONSUMER_CTRL_SUPPORT) ?				\
+	 (REPORT_SIZE_CONSUMER_CTRL) : (0))
+#define _REPORT_BUFFER_SIZE_BOOT_MOUSE \
+	(IS_ENABLED(CONFIG_DESKTOP_HID_BOOT_INTERFACE_MOUSE) ? (REPORT_SIZE_MOUSE_BOOT) : (0))
+#define _REPORT_BUFFER_SIZE_BOOT_KEYBOARD \
+	(IS_ENABLED(CONFIG_DESKTOP_HID_BOOT_INTERFACE_KEYBOARD) ? (REPORT_SIZE_KEYBOARD_BOOT) : (0))
+#define _MAX6(a, b, c, d, e, f)	MAX(MAX(a, b), MAX(MAX(c, d), MAX(e, f)))
+
+/* Size of the biggest supported HID input report that is part of input reports map. */
```
zycz, line 104: You can now drop this comment

zycz (approved):

applications/nrf_desktop/configuration/common/hid_report_desc.h
MarekPieta: Can we keep it (as it's part of the API and for consistency with `REPORT_BUFFER_SIZE_OUTPUT_REPORT`)?

MarekPieta: I would like to also ensure that user is aware exactly what is the meaning of the defined value

zycz: I gave you approve (second time) so I do not see this change as crucial. Do as you want. Just pointing out that IMO currently this comment is redundant. But do as you wish :)

alstrzebonski (approved):

PR #16633, opened 2024-07-26
modules: hostap: Fix null termination of ssid and psk

kapbh:
[SHEL-2900] Add null termination of ssid and psk. If user pass ssid/psk without null terminator it will add by default.

[SHEL-2900]: https://nordicsemi.atlassian.net/browse/SHEL-2900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/hostap/src/supp_api.c
```diff
@@ -309,6 +311,14 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
 		_wpa_cli_cmd_v("set_network %d mode 2", resp.network_id);
 	}
 
+	strncpy(ssid_null_terminated, params->ssid, WIFI_SSID_MAX_LEN);
+	ssid_null_terminated[params->ssid_length] = '\0';
+	params->ssid = ssid_null_terminated;
```
krish2718, line 316: is this necessary? can't we just use it locally? Ideally the `params` should have been `const`.
Also, if you are updating `params->ssid` then you also need to update `params->ssid_length`. Same for `psk_length` too.

modules/hostap/src/supp_api.c
```diff
@@ -309,6 +311,14 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
 		_wpa_cli_cmd_v("set_network %d mode 2", resp.network_id);
 	}
 
+	strncpy(ssid_null_terminated, params->ssid, WIFI_SSID_MAX_LEN);
+	ssid_null_terminated[params->ssid_length] = '\0';
```
sachinthegreen, line 315: Is a check necessary to make sure that `params->ssid_length <= WIFI_SSID_MAX_LEN` ? (Similarly for psk_len)

sachinthegreen (changes requested):

modules/hostap/src/supp_api.c
kapbh: We are passing only null terminated ssid/psk from shell. I think ssid_length/psk_length should be ssid/psk without '\0' only. If someone use hostap and pass ssid/psk without null terminator this will add by default.

modules/hostap/src/supp_api.c
kapbh: Yes need to add check length check for all layers. We also have separate ticket for that.

sachinthegreen: I believe you are referring to the ticket which checks for `minimum` length. I was referring to the upper limit check here.

modules/hostap/src/supp_api.c
```diff
@@ -309,7 +311,24 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
 		_wpa_cli_cmd_v("set_network %d mode 2", resp.network_id);
 	}
 
-	_wpa_cli_cmd_v("set_network %d ssid \"%s\"", resp.network_id, params->ssid);
+	if (params->ssid_length > WIFI_SSID_MAX_LEN) {
+		wpa_printf(MSG_ERROR, "SSID too long (max 32 characters)");
```
krish2718, line 315:
```suggestion
		wpa_printf(MSG_ERROR, "SSID too long (max %d characters)", WIFI_SSID_MAX_LEN);
```

modules/hostap/src/supp_api.c
```diff
@@ -309,7 +311,24 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
...
+		goto out;
+	}
+
+	if ((params->psk_length < WIFI_PSK_MIN_LEN) ||
+	    (params->psk_length > WIFI_PSK_MAX_LEN)) {
+		wpa_printf(MSG_ERROR, "Password should be min 8 and max 64 characters");
```
krish2718, line 321:
```suggestion
		wpa_printf(MSG_ERROR, "Passphrase should be in range (%d-%d) characters", WIFI_PSK_MIN_LEN, WIFI_PSK_MAX_LEN);
```

krish2718: check for `sae_password` too?

modules/hostap/src/supp_api.c
```diff
@@ -326,6 +336,28 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
 	}
 
 	if (params->security != WIFI_SECURITY_TYPE_NONE) {
+
```
sachinthegreen, line 339: Redundant blank line.

modules/hostap/src/supp_api.c
```diff
@@ -326,6 +336,28 @@ static int wpas_add_and_config_network(struct wpa_supplicant *wpa_s,
...
+			    (params->psk_length > WIFI_PSK_MAX_LEN)) {
+				wpa_printf(MSG_ERROR,
+					   "Passphrase should be in range (%d-%d) characters",
+					   WIFI_PSK_MIN_LEN, WIFI_PSK_MAX_LEN);
+				goto out;
+			}
+		}
+
+		strncpy(psk_null_terminated, params->psk, WIFI_PSK_MAX_LEN);
+		psk_null_terminated[params->psk_length] = '\0';
```
sachinthegreen, line 359: Shouldn't these use `params->sae_password` and `params->sae_password_length` for SAE?

sachinthegreen (changes requested):

modules/hostap/src/supp_api.c
kapbh: Yes it should be. Currently there is one crash because of this `strncpy`. Debugging the issue.

sachinthegreen (approved):

krga2022 (approved):

rado17 (approved):

PR #16638, opened 2024-07-26
Samples: Bluetooth: Fix use of local obtained conn refs

rugeGerritsen:
The returned reference is unused.

Found using the semantic patch file:
```
@@
local idexpression conn;
identifier e;
statement S1;
@@
* e = bt_conn_le_create(..., &conn);
... when != bt_conn_unref(conn)
    when != if (e == 0) { <+... bt_conn_unref(conn) ...+> } else S1
```


alwa-nordic (changes requested):
This is a step in the right direction. It correctly disposes of the reference when we drop the pointer. But the code leaves the local optimum of "it's weird, but it works" and probably does not work anymore.

We must remove all other `bt_conn_unref(conn);` in these files at the same time, since I suspect those are all disposing of the same reference (the one from `bt_conn_le_create`).

alwa-nordic (approved):

KAGA164 (approved):

PR #16639, opened 2024-07-26
doc: building: mention -v for verbose builds

greg-fer:
Added mention of the `-vvv` option to the optional build params list. Also propagated as a tip to the output build files page. NCSDK-27695.


hellesvik-nordic:
From west --help:
`  -v, --verbose     Display verbose output. May be given multiple times to increase verbosity.`

If we want to be more specific, we could also specify this.

hellesvik-nordic (approved):
Very nice!

FrancescoSer (approved):

PR #16659, opened 2024-07-29
partition_manager: TrustZone HW model 1 workaround

Vge0rge:
The configurations: NRF_TRUSTZONE_FLASH_REGION_SIZE and NRF_TRUSTZONE_RAM_REGION_SIZE are added after the release and they are not included in the hardware
model 1. This is a workaround for this use case,
since we don't expect to support the nRF54L15 devices with this model this workaround should work for the older generation of devices.

Ref: NCSDK-28479

### DIFF
--- a/subsys/partition_manager/Kconfig
+++ b/subsys/partition_manager/Kconfig
@@ -32,6 +32,23 @@ config SRAM_BASE_ADDRESS
 	hex
 	depends on !PARTITION_MANAGER_ENABLED
 
+# Workaround for the hardware model v1 which is not aware of the new
+# TrustZone definitions. Since nRF54L15 is not supported with the
+# hardware model v1, the values are set to the SPU values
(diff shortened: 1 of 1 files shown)


subsys/partition_manager/Kconfig
```diff
@@ -32,6 +32,23 @@ config SRAM_BASE_ADDRESS
 	hex
 	depends on !PARTITION_MANAGER_ENABLED
 
+# Workaround for the hardware model v1 which is not aware of the new
+# TrustZone definitions. Since nRF54L15 is not supported with the
+# hardware model v1, the alues are set to the SPU values
```
nordicjm, line 37: *values

nordicjm (approved):

vili-nordic (approved):

PerMac (approved):
fixes the issue

PR #16671, opened 2024-07-30
Cracen: Fix import/export/generate key implementations

vili-nordic:
Fixes some double read inconsitiencies and key validation in Cracen key management implementation.

### DIFF
--- a/subsys/nrf_security/src/drivers/cracen/common/include/cracen/mem_helpers.h
+++ b/subsys/nrf_security/src/drivers/cracen/common/include/cracen/mem_helpers.h
@@ -43,10 +43,22 @@ bool constant_memcmp_is_zero(const void *s1, size_t n);
 void safe_memset(void *dest, const size_t dest_size, const uint8_t ch, const size_t n);
 
 /*!
- * \brief A meset to zero function that is not optimized out by the compiler.
+ * \brief A memset to zero function that is not optimized out by the compiler.
  *
  * \param[in] dest		Pointer to the memory to set to zero.
  * \param[in] dest_size Size of the memory in bytes.
  *
  */
 void safe_memzero(void *dest, const size_t dest_size);
+
+/*!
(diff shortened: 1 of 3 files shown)


subsys/nrf_security/src/drivers/cracen/cracenpsa/src/key_management.c
```diff
@@ -939,6 +962,7 @@ static psa_status_t generate_rsa_private_key(const psa_key_attributes_t *attribu
 	size_t bits = psa_get_key_bits(attributes);
 	size_t key_size_bytes = PSA_BITS_TO_BYTES(bits);
 	size_t key_size_half = key_size_bytes / 2;
+	uint8_t key_workbuf[PSA_KEY_EXPORT_RSA_KEY_PAIR_MAX_SIZE(PSA_MAX_RSA_KEY_BITS)];
```
Vge0rge, line 965: Hm, why is this needed for the key generation?

Vge0rge (approved):

subsys/nrf_security/src/drivers/cracen/common/include/cracen/mem_helpers.h
```diff
@@ -43,10 +43,22 @@ bool constant_memcmp_is_zero(const void *s1, size_t n);
...
+ * \brief A memset to zero function that is not optimized out by the compiler.
  *
  * \param[in] dest		Pointer to the memory to set to zero.
  * \param[in] dest_size Size of the memory in bytes.
  *
  */
 void safe_memzero(void *dest, const size_t dest_size);
+
+/*!
+ * \brief Will copy memory, verifying that the buffer is non-zero.
```
tomi-font, line 55:
```suggestion
 * \brief Will copy memory, verifying that the source buffer is non-zero.
```

tomi-font (changes requested):

tomi-font (approved):

PR #16678, opened 2024-07-30
Issue28108 caf pm

zycz:
subsys: caf: Add support for nrf54h20 for power_manager

Add support for nrf54h20 for power_manager.
Support for nrf54h20 is limited and do not support state_off.


MarekPieta (changes requested):

subsys/caf/modules/power_manager.c
```diff
@@ -7,8 +7,10 @@
 #include <zephyr/types.h>
 
 #include <zephyr/device.h>
+#if !defined(CONFIG_SOC_SERIES_NRF54HX)
 #include <hal/nrf_power.h>
```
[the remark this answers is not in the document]
zycz: no

subsys/caf/modules/power_manager.c
```diff
@@ -148,19 +150,32 @@ static void system_off_on_error(void)
 	LOG_WRN("System turned off because of unrecoverable error");
 	LOG_PANIC();
 
+#if !defined(CONFIG_SOC_SERIES_NRF54HX)
```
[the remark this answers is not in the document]
zycz: 1) Yes of course I should have add CONFIG_PM thanks for finding it
2) NRF54L seams to work fine with old power management and also lets concentrate on one task at the time so I wouldn't make changes there in this task. 
3) Completly don't understand your idea with pm_state_force and extra state and extra policy - why should we do it? For now we can't enter state soft_off so lets not assume how it will be done in the future will we enter it using pm api or some other function like sys_poweroff(). Now we don't enter it and let's just leave it as it is.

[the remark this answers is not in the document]
MarekPieta: > NRF54L seams to work fine with old power management and also lets concentrate on one task at the time so I wouldn't make changes there in this task.

Actually we have not confirmed lower power consumption for nRF54L yet (unless you did it already?). Please just double check with people involved in PM development if the `CONFIG_PM` will be beneficial for nRF54L or there is no need to enable it there and we can stick to `sys_poweroff` API. It would be good to make an informed decision here already.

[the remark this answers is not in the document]
MarekPieta: > Completly don't understand your idea with pm_state_force and extra state and extra policy - why should we do it? For now we can't enter state soft_off so lets not assume how it will be done in the future will we enter it using pm api or some other function like sys_poweroff(). Now we don't enter it and let's just leave it as it is.

From what I heard so far, I would expect we will use PM APIs here. Still, maybe you already got newer information that would state otherwise. I would not insist here.

subsys/caf/modules/power_manager.c
```diff
@@ -148,19 +150,32 @@ static void system_off_on_error(void)
...
+	LOG_WRN("System suspended, the poweroff is not supported yet on nrf54h20");
+#endif
 }
 
 static void system_off_handler(struct k_work *work)
 {
 	LOG_WRN("System turned off");
 	LOG_PANIC();
+
+#if !defined(CONFIG_SOC_SERIES_NRF54HX)
```
MarekPieta, line 166: Can we introduce Kconfig for clearing the reset reason (Kconfig that depends on `!SOC_SERIES_NRF54HX`)?

MarekPieta: Similarly for `sys_poweroff` (we can also select either `CONFIG_POWEROFF` or `CONFIG_PM` depending on SoC) ?

MarekPieta (changes requested):

MarekPieta (changes requested):

subsys/caf/modules/power_manager.c
```diff
@@ -148,19 +151,34 @@ static void system_off_on_error(void)
 	LOG_WRN("System turned off because of unrecoverable error");
 	LOG_PANIC();
 
-	sys_poweroff();
+	if (IS_ENABLED(CONFIG_POWEROFF)) {
```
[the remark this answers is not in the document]
MarekPieta: You can also add a build assert to ensure that `CONFIG_POWEROFF` and `CONFIG_PM` are not enabled simultaneously

subsys/caf/modules/Kconfig.power_manager
```diff
@@ -11,7 +11,7 @@ config CAF_POWER_MANAGER
 	select PM_DEVICE
 	select CAF_PM_EVENTS
 	select CAF_POWER_MANAGER_EVENTS
-	select POWEROFF
+	select POWEROFF if !SOC_SERIES_NRF54HX
```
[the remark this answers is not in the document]
MarekPieta: `imply PM if SOC_SERIES_NRF54HX`

MarekPieta, line 14: We could also use `imply POWEROF if !SOC_SERIES_NRF54H`

applications/nrf_desktop/configuration/nrf54h20dk_nrf54h20_cpuapp/app.overlay
```diff
@@ -68,3 +68,13 @@
 &cpusec_bellboard {
 	status = "okay";
 };
+
+/* Remove ppr partition as ppr core is not used and extend app core code partition. */
+/delete-node/ &cpuppr_code_partition;
```
MarekPieta, line 73: Consider setting partition size to zero instead of deleting node (if any node refers to `cpuppr_code_partition`, it will need to be manually aligned too). As an alternative, we could reduce PPR partition size (e.g. cut in half).

MarekPieta (changes requested):

applications/nrf_desktop/configuration/nrf54h20dk_nrf54h20_cpuapp/prj.conf
```diff
@@ -141,6 +138,9 @@ CONFIG_MCUMGR_TRANSPORT_WORKQUEUE_STACK_SIZE=4096
 CONFIG_MCUMGR_GRP_OS=y
 CONFIG_MCUMGR_GRP_OS_MCUMGR_PARAMS=y
 
+# Enable Zephyr System Power Management
```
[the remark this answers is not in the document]
zycz: We already select CAF_POWER_MANAGER which implies CONFIG_PM. Implying PM here is redundant. Maybe we should then remove imply PM in  CAF_POWER_MANAGER  and move it here?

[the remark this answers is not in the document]
MarekPieta: It's all fine now. The comment simply no longer applies (it was related to previous version of your PR).

subsys/caf/modules/power_manager.c
```diff
@@ -7,10 +7,13 @@
 #include <zephyr/types.h>
 
 #include <zephyr/device.h>
+#if defined(CAF_POWER_CLEAR_RESET_REASON)
```
MarekPieta, line 10: `CONFIG_CAF_POWER_MANAGER_CLEAR_RESET_REASON`

subsys/caf/modules/power_manager.c
```diff
@@ -142,25 +145,47 @@ static void system_off(void)
 	}
 }
 
+static void system_off_post_action(void) {
+
+/* You shouldn't enable CONFIG_POWEROFF and CONFIG_PM simultaneously */
+#if defined CONFIG_POWEROFF && defined CONFIG_PM
+		BUILD_ASSERT(false, "Redundant power configs turned on");
+#endif
```
MarekPieta, line 153: `BUILD_ASSERT(!(IS_ENABLED(CONFIG_POWEROFF) && IS_ENABLED(CONFIG_PM)), "CAF Power Manager uses either power off or power management, but never both")`

subsys/caf/modules/power_manager.c
```diff
@@ -142,25 +145,47 @@ static void system_off(void)
...
+
+		if (IS_ENABLED(CONFIG_PM)) {
```
MarekPieta, line 155: Indent is not proper

subsys/caf/modules/power_manager.c
```diff
@@ -142,25 +145,47 @@ static void system_off(void)
...
+		/* On nRF54 series, the poweroff is not supported yet. */
+		LOG_WRN("System suspended, the poweroff is not supported yet on nrf54h20");
```
MarekPieta, line 157: Can we keep the information specific to nRF54H only in the comment (alternatively, you could make an additional conditional log - displaying this information for nRF52 SoCs would be confusing)

applications/nrf_desktop/configuration/nrf54h20dk_nrf54h20_cpuapp/prj.conf
MarekPieta, line 141: For other configurations we do not enable `CONFIG_POWEROFF`. Can we rely on imply also here, please? (to simplify configuration)

MarekPieta (changes requested):

subsys/caf/modules/power_manager.c
```diff
@@ -7,10 +7,13 @@
 #include <zephyr/types.h>
 
 #include <zephyr/device.h>
+#if defined(CONFIG_CAF_POWER_CLEAR_RESET_REASON)
```
MarekPieta, line 10: `CONFIG_CAF_POWER_MANAGER_CLEAR_RESET_REASON`

MarekPieta (changes requested):

MarekPieta (approved):

subsys/caf/modules/power_manager.c
```diff
@@ -142,25 +145,44 @@ static void system_off(void)
...
+	/* You shouldn't enable CONFIG_POWEROFF and CONFIG_PM simultaneously */
+	BUILD_ASSERT(!(IS_ENABLED(CONFIG_POWEROFF) && IS_ENABLED(CONFIG_PM)),
+		"CAF Power Manager uses either power off or power management, but never both");
+
+#if defined(CONFIG_PM)
+		(void)pm_state_force(0u, &(struct pm_state_info){.state = PM_STATE_SOFT_OFF,
+									.substate_id =  0,
+									.min_residency_us = 0,
+									.exit_latency_us = 0 });
+#elif defined(CONFIG_POWEROFF)
```
alstrzebonski, line 159: Maybe just `#if CONFIG_PM`? In theory it can be defined but set to 0.

subsys/caf/modules/power_manager.c
```diff
@@ -142,25 +145,44 @@ static void system_off(void)
...
-	sys_poweroff();
+	system_off_post_action();
 }
 
 static void system_off_handler(struct k_work *work)
 {
 	LOG_WRN("System turned off");
 	LOG_PANIC();
+
+#if defined(CONFIG_CAF_POWER_MANAGER_CLEAR_RESET_REASON)
```
alstrzebonski, line 178: Same here

alstrzebonski (approved):

PR #16682, opened 2024-07-31
logging: add RPC command to configure log level

Damian-Nordic:
Add RPC command to configure the maximum level of messages that are expected to be streamed to the RPC client.

### DIFF
--- a/include/logging/log_rpc.h
+++ b/include/logging/log_rpc.h
@@ -21,6 +21,30 @@ extern "C" {
 #endif
 
 /**
+ * nRF RPC logging level.
+ */
+enum log_rpc_level {
+	LOG_RPC_LEVEL_NONE = 0,
+	LOG_RPC_LEVEL_ERR,
+	LOG_RPC_LEVEL_WRN,
+	LOG_RPC_LEVEL_INF,
+	LOG_RPC_LEVEL_DBG,
+};
+
+/**
+ * Sets the logging stream verbosity level.
+ *
+ * Configures the remote device to only stream log messages whose level is less
+ * than or equal to the specified level.
+ *
+ * @param level		Logging level, see @c log_rpc_level.
+ *
+ * @retval 0		On success.
+ * @retval -errno	On failure.
+ */
+int log_rpc_set_stream_level(enum log_rpc_level level);
+
+/**
  * @brief Retrieves the crash log retained on the remote device.
  *
  * The function issues an nRF RPC command to obtain the last crash log retained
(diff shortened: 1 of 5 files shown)


include/logging/log_rpc.h
```diff
@@ -20,6 +20,25 @@
...
+	LOG_RPC_LEVEL_DBG,
+};
+
+/**
+ * Sets the logging stream verbosity level.
+ *
+ * Configures the remote device to only stream log messages whose level is less
+ * than or equal to the specified level.
+ */
+int log_rpc_set_stream_level(enum log_rpc_level level);
```
alxelax, line 40: Public API should have sufficient description of incoming parameters and return value with Doxygen tags.

samples/nrf_rpc/ps_client/src/log_rpc_shell.c
```diff
@@ -11,7 +11,29 @@
...
+	if (rc) {
+		shell_error(sh, "Invalid argument: %d", rc);
+		return rc;
+	}
+
+	rc = log_rpc_set_stream_level(level);
+
+	if (rc) {
+		shell_error(sh, "Error: %d", rc);
+		return rc;
```
alxelax, line 30: I'm not sure that it is correct to return any value in case of error. Public API says:
```
/**
 * @brief Shell command handler prototype.
 *
 * @param sh Shell instance.
 * @param argc  Arguments count.
 * @param argv  Arguments.
 *
 * @retval 0 Successful command execution.
 * @retval 1 Help printed and command not executed.
 * @retval -EINVAL Argument validation failed.
 * @retval -ENOEXEC Command not executed.
 */
```
https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/shell/shell.h#L162-L175
I guess it should be `-ENOEXEC`.
Here and below.

alxelax (changes requested):

include/logging/log_rpc.h
Damian-Nordic: Right, done.

samples/nrf_rpc/ps_client/src/log_rpc_shell.c
Damian-Nordic: Done.

alxelax: thanks, but, I guess, `cmd_log_rpc_crash` requires the same change, doesn't?

Damian-Nordic: right, sorry, done.

alxelax (approved):

canisLupus1313 (approved):

PR #16689, opened 2024-07-31
applications: matter-ble bridge over thread

kg-nordicsemi:
- modyfiy application configuration files, and  board files, to create support for nrf5340, in matter-bridge application.
- create support for nrf54h20.
- create .conf files to maximize BLE connections
- update sample.yaml
- update documentation

### DIFF
--- a/applications/matter_bridge/Kconfig.sysbuild
+++ b/applications/matter_bridge/Kconfig.sysbuild
@@ -17,6 +17,11 @@ choice BOOTLOADER
 	default BOOTLOADER_MCUBOOT if !SOC_SERIES_NRF54HX
 endchoice
 
+# Enable IEEE802.15.4 serialization to network core
+# This config automatically pulls in the `overlay-802154.conf` in the ipc_radio image
+config NETCORE_IPC_RADIO_IEEE802154
+	default y if (SOC_SERIES_NRF53X || SOC_SERIES_NRF54HX) && !WIFI_NRF700X
+
 if BOOTLOADER_MCUBOOT
 
 #### DFU multi-image support
@@ -33,7 +38,7 @@ config PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY
 if SOC_SERIES_NRF53X
 
 config MCUBOOT_UPDATEABLE_IMAGES
-	default 3 if BOARD_NRF5340DK_NRF5340_CPUAPP #### Support for WiFi patch on external flash memory.
+	default 3 if (BOARD_NRF5340DK_NRF5340_CPUAPP && WIFI_NRF700X) #### Support for WiFi patch on external flash memory.
 	default 2
 
 choice MCUBOOT_MODE
--- a/applications/matter_bridge/doc/matter_bridge_description.rst
+++ b/applications/matter_bridge/doc/matter_bridge_description.rst
@@ -532,18 +532,79 @@ Since the Matter stack uses one Bluetooth LE connection for commissioning, the m
 
 Increasing the number of Bluetooth LE connections affects the RAM usage on both the application and network cores.
 The current maximum number of Bluetooth LE connections that can be selected using the default configuration is ``10``.
-You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
-Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
 
-.. parsed-literal::
-   :class: highlight
+.. tabs::
 
-   west build -b nrf7002dk/nrf5340/cpuapp -- -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
+   You can configure the number of Bluetooth LE bridged devices depending on your device and configuration:
 
-.. parsed-literal::
-   :class: highlight
+   .. group-tab:: nRF53 DKs
 
-   west build -b nrf54h20dk/nrf54h20/cpuapp -- -DSB_CONFIG_WIFI_NRF700X=y -DCONFIG_CHIP_WIFI=y -Dmatter_bridge_SHIELD=nrf700x_nrf54h20dk -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
+      The nRF53 Series supports the Matter bridge over Wi-Fi and Matter bridge over Thread configurations.
+
+      .. tabs::
+
+         .. group-tab:: Matter bridge over Wi-Fi
+
+            You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
+
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the size of Bluetooth LE TX/RX buffers:
+
+            .. parsed-literal::
+               :class: highlight
+
+               west build -b nrf5340dk/nrf5340/cpuapp -p -- -Dmatter_bridge_SHIELD=nrf7002ek -DSB_CONFIG_WIFI_PATCHES_EXT_FLASH_STORE=y -DSB_CONFIG_DFU_MULTI_IMAGE_PACKAGE_WIFI_FW_PATCH=y -DSB_CONFIG_WIFI_NRF700X=y -Dmcuboot_CONFIG_UPDATEABLE_IMAGE_NUMBER=3 -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf" -DFILE_SUFFIX=nrf70ek
+
+         .. group-tab:: Matter bridge over Thread
+
+            You can not increase the default number of Bluetooth LE connections in this configuration using overlays.
+            This is because the configuration uses both Thread and Bluetooth LE protocols, and limited RAM memory.
+            You can still increase the number of connections by modifying the board files and decreasing the buffer sizes.
+            The default number of connections is nine.
+
+   .. group-tab:: nRF54 DKs
+
+      The nRF54 Series supports the Matter bridge over Wi-Fi and Matter bridge over Thread configurations.
+
+      .. tabs::
+
+         .. group-tab:: Matter bridge over Wi-Fi
+
+            You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
+
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+
+             .. parsed-literal::
+               :class: highlight
+
+               west build -b nrf54h20dk/nrf54h20/cpuapp -p -- -DSB_CONFIG_WIFI_NRF700X=y -DCONFIG_CHIP_WIFI=y -Dmatter_bridge_SHIELD=nrf700x_nrf54h20dk -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
+
+         .. group-tab:: Matter bridge over Thread
+
+            You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
+
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+
+             .. parsed-literal::
+               :class: highlight
+
+               west build -b nrf54h20dk/nrf54h20/cpuapp -p -- -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
+
+   .. group-tab:: nRF70 DKs
+
+      The nRF70 Series supports the Matter bridge over Wi-Fi configuration.
+
+      .. tabs::
+
+         .. group-tab:: Matter-Bridge over Wi-Fi
+
+            You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
+
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+
+             .. parsed-literal::
+               :class: highlight
+
+               west build -b nrf7002dk/nrf5340/cpuapp -- -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
 
 .. _matter_bridge_app_bt_security:
 
@@ -611,6 +672,13 @@ The application supports the following configurations:
      - Release version of the application.
 
        Enables only the necessary application functionalities to optimize its performance.
+   * - Matter bridge over Wi-Fi with nRF5340 DK and nRF7002 EK
+     - :file:`---`
+     - ``nrf70ek``
+     - nRF5340 DK with the nRF7002 EK shield attached
+     - Debug version of the application with Matter over Wi-Fi enabled.
+
+       Enables the Matter Bridge to work with Wi-Fi on nRF5340 DK.
 
 Building and running
 ********************
@@ -627,7 +695,7 @@ For example:
 
    .. code-block:: console
 
-      west build -b nrf5340dk/nrf5340/cpuapp -p -- -Dmatter_bridge_SHIELD=nrf7002ek -DSB_CONFIG_WIFI_PATCHES_EXT_FLASH_STORE=y -DSB_CONFIG_DFU_MULTI_IMAGE_PACKAGE_WIFI_FW_PATCH=y -DSB_CONFIG_WIFI_NRF700X=y -Dmcuboot_CONFIG_UPDATEABLE_IMAGE_NUMBER=3
+      west build -b nrf5340dk/nrf5340/cpuapp -p -- -Dmatter_bridge_SHIELD=nrf7002ek -DSB_CONFIG_WIFI_PATCHES_EXT_FLASH_STORE=y -DSB_CONFIG_DFU_MULTI_IMAGE_PACKAGE_WIFI_FW_PATCH=y -DSB_CONFIG_WIFI_NRF700X=y -Dmcuboot_CONFIG_UPDATEABLE_IMAGE_NUMBER=3 -DFILE_SUFFIX=nrf70ek
 
 
 To use the nRF54H20 DK with the ``nrf7002ek`` shield (2.4 GHz or 5 GHz), follow the :ref:`ug_nrf7002eb_nrf54h20dk_gs` user guide to connect all required pins.
--- a/applications/matter_bridge/sample.yaml
+++ b/applications/matter_bridge/sample.yaml
@@ -58,6 +58,7 @@ tests:
       mcuboot_CONFIG_UPDATEABLE_IMAGE_NUMBER=3 SB_CONFIG_WIFI_NRF700X=y
       SB_CONFIG_DFU_MULTI_IMAGE_PACKAGE_WIFI_FW_PATCH=y
       CONFIG_CHIP_DFU_OVER_BT_SMP=y
+      FILE_SUFFIX=nrf70ek
     integration_platforms:
       - nrf5340dk/nrf5340/cpuapp
     platform_allow: nrf5340dk/nrf5340/cpuapp
@@ -72,67 +73,12 @@ tests:
       - nrf7002dk/nrf5340/cpuapp
     platform_allow: nrf7002dk/nrf5340/cpuapp
     tags: sysbuild ci_applications_matter
-  applications.matter_bridge.nrf54h20.br_ble:
-    sysbuild: true
-    build_only: true
-    extra_args: SB_CONFIG_WIFI_NRF700X=y
-      CONFIG_CHIP_WIFI=y
-      matter_bridge_SHIELD=nrf700x_nrf54h20dk
-      CONFIG_BRIDGED_DEVICE_BT=y
-    integration_platforms:
-      - nrf54h20dk/nrf54h20/cpuapp
-    platform_allow: nrf54h20dk/nrf54h20/cpuapp
-    tags: sysbuild ci_applications_matter
-  applications.matter_bridge.nrf54h20:
-    sysbuild: true
-    build_only: true
-    extra_args: SB_CONFIG_WIFI_NRF700X=y
-      CONFIG_CHIP_WIFI=y
-      matter_bridge_SHIELD=nrf700x_nrf54h20dk
-      CONFIG_BRIDGED_DEVICE_SIMULATED=y
-    integration_platforms:
-      - nrf54h20dk/nrf54h20/cpuapp
-    platform_allow: nrf54h20dk/nrf54h20/cpuapp
-    tags: sysbuild ci_applications_matter
-  applications.matter_bridge.nrf54h20.br_ble.smart_plug:
-    sysbuild: true
-    build_only: true
-    extra_args: SB_CONFIG_WIFI_NRF700X=y
-      CONFIG_CHIP_WIFI=y
-      matter_bridge_SHIELD=nrf700x_nrf54h20dk
-      CONFIG_BRIDGED_DEVICE_BT=y
-      matter_bridge_SNIPPET=onoff_plug
-    integration_platforms:
-      - nrf54h20dk/nrf54h20/cpuapp
-    platform_allow: nrf54h20dk/nrf54h20/cpuapp
-    tags: sysbuild ci_applications_matter
-  applications.matter_bridge.nrf54h20.smart_plug:
-    sysbuild: true
-    build_only: true
-    extra_args: SB_CONFIG_WIFI_NRF700X=y
-      CONFIG_CHIP_WIFI=y
-      matter_bridge_SHIELD=nrf700x_nrf54h20dk
-      CONFIG_BRIDGED_DEVICE_SIMULATED=y
-      matter_bridge_SNIPPET=onoff_plug
-    integration_platforms:
-      - nrf54h20dk/nrf54h20/cpuapp
-    platform_allow: nrf54h20dk/nrf54h20/cpuapp
-    tags: sysbuild ci_applications_matter
-  applications.matter_bridge.br_ble.smart_plug:
+  applications.matter_bridge.nrf5340.br_ble.smart_plug.thread:
     sysbuild: true
     build_only: true
     extra_args: CONFIG_BRIDGED_DEVICE_BT=y
       matter_bridge_SNIPPET=onoff_plug
     integration_platforms:
-      - nrf7002dk/nrf5340/cpuapp
-    platform_allow: nrf7002dk/nrf5340/cpuapp
-    tags: sysbuild ci_applications_matter
-  applications.matter_bridge.smart_plug:
-    sysbuild: true
-    build_only: true
-    extra_args: CONFIG_BRIDGED_DEVICE_SIMULATED=y
-      matter_bridge_SNIPPET=onoff_plug
-    integration_platforms:
-      - nrf7002dk/nrf5340/cpuapp
-    platform_allow: nrf7002dk/nrf5340/cpuapp
+      - nrf5340dk/nrf5340/cpuapp
+    platform_allow: nrf5340dk/nrf5340/cpuapp
     tags: sysbuild ci_applications_matter
--- a/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
+++ b/doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
@@ -219,6 +219,7 @@ Matter Bridge
     By default, the option points to the :file:`src/default_zap` directory and can be changed to any path relative to application's location that contains the ZAP file and :file:`zap-generated` directory.
   * Support for the :ref:`zephyr:nrf54h20dk_nrf54h20`.
   * Optional smart plug device functionality.
+  * Support for the Thread protocol.
 
 nRF5340 Audio
 -------------
--- a/samples/matter/common/src/bridge/Kconfig
+++ b/samples/matter/common/src/bridge/Kconfig
@@ -48,4 +48,32 @@ config BRIDGE_BT_MINIMUM_SECURITY_LEVEL
 	   allowed to connect to the bridge. The level cannot be set to a value smaller than 2, as it would mean that
 	   the communication with the connected devices will not be encrypted, what is not considered to be secure.
 
+config BRIDGE_BT_SCAN_WINDOW
+	int "Time duration of a central actively scanning for devices within scan interval, it is equal BRIDGE_BT_SCAN_WINDOW*0.625 ms"
+	default 40  if NET_L2_OPENTHREAD # It is necessary to decrease Bluetooth LE activity to leave some radio time to Thread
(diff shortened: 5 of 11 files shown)


applications/matter_bridge/Kconfig.sysbuild
```diff
@@ -33,7 +38,7 @@ config PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY
 if SOC_SERIES_NRF53X
 
 config MCUBOOT_UPDATEABLE_IMAGES
-	default 3 if BOARD_NRF5340DK_NRF5340_CPUAPP #### Support for WiFi patch on external flash memory.
+	#default 3 if BOARD_NRF5340DK_NRF5340_CPUAPP #### Support for WiFi patch on external flash memory.
```
kkasperczyk-no, line 41: We should not remove it :) I think we should do something like 
```suggestion
	default 3 if (BOARD_NRF5340DK_NRF5340_CPUAPP && WIFI_NRF700X) #### Support for WiFi patch on external flash memory.
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,103 @@ Use the following configuration options to customize the number of supported bri
...
 
-Increasing the number of Bluetooth LE connections affects the RAM usage on both the application and network cores.
-The current maximum number of Bluetooth LE connections that can be selected using the default configuration is ``10``.
-You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
-Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+   .. group-tab:: nRF53 DKs
 
-.. parsed-literal::
-   :class: highlight
+      You have to consider several factors to decide how many Bluetooth LE bridged devices to support.
```
kkasperczyk-no, line 533: Can we put the text that is repeated for both tabs before the tabs section?

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,103 @@ Use the following configuration options to customize the number of supported bri
...
+      Every Bluetooth LE bridged device uses a separate Bluetooth LE connection, so you need to consider the number of supported Bluetooth LE connections (selected using the :kconfig:option:`CONFIG_BT_MAX_CONN` Kconfig option) when you configure the number of devices.
+      Since the Matter stack uses one Bluetooth LE connection for commissioning, the maximum number of connections you can use for bridged devices is one less than is configured using the :kconfig:option:`CONFIG_BT_MAX_CONN` Kconfig option.
 
-   west build -b nrf7002dk/nrf5340/cpuapp -- -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
+      .. tabs::
 
-.. parsed-literal::
-   :class: highlight
+         .. group-tab:: Matter-Bridge over wifi configuration
```
kkasperczyk-no, line 539:
```suggestion
         .. group-tab:: Matter-Bridge over Wi-Fi configuration
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,103 @@ Use the following configuration options to customize the number of supported bri
...
+            You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+
+            .. parsed-literal::
+               :class: highlight
+
+               west build -b nrf5340dk/nrf5340/cpuapp -p -- -Dmatter_bridge_SHIELD=nrf7002ek -DSB_CONFIG_WIFI_PATCHES_EXT_FLASH_STORE=y -DSB_CONFIG_DFU_MULTI_IMAGE_PACKAGE_WIFI_FW_PATCH=y -DSB_CONFIG_WIFI_NRF700X=y -Dmcuboot_CONFIG_UPDATEABLE_IMAGE_NUMBER=3 -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
+
+
+         .. group-tab:: Matter-Bridge over thread configuration
```
kkasperczyk-no, line 552:
```suggestion
         .. group-tab:: Matter-Bridge over Thread configuration
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,103 @@ Use the following configuration options to customize the number of supported bri
...
+
+            Increasing the number of Bluetooth LE connections affects the RAM usage on both the application and network cores.
+            The current maximum number of Bluetooth LE connections that can be selected using the default configuration is ``10``.
+            Due to using thread and BLE protocols in this configuration You can't increase the number of Bluetooth LE connections and additionally you have to decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, this will decrease the communication throughput.
+            Build the target using the following command in the project directory to enable a configuration that allow to set the number of Bluetooth LE connections to ``10`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
```
kkasperczyk-no, line 557: Isn't it that we say here - the default configuration is 10, but you can configure it to use max number which is 10? :smile:

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,103 @@ Use the following configuration options to customize the number of supported bri
...
+
+      You have to consider several factors to decide how many Bluetooth LE bridged devices to support.
+      Every Bluetooth LE bridged device uses a separate Bluetooth LE connection, so you need to consider the number of supported Bluetooth LE connections (selected using the :kconfig:option:`CONFIG_BT_MAX_CONN` Kconfig option) when you configure the number of devices.
+      Since the Matter stack uses one Bluetooth LE connection for commissioning, the maximum number of connections you can use for bridged devices is one less than is configured using the :kconfig:option:`CONFIG_BT_MAX_CONN` Kconfig option.
+
+      .. tabs::
+
+         .. group-tab:: Matter-Bridge over wifi configuration
+
+            Increasing the number of Bluetooth LE connections affects the RAM usage on both the application and network cores.
```
kkasperczyk-no, line 574: I can see a lot of duplications in this text between the tabs. I would suggest contacting @wiba-nordic to discuss how to avoid it, as the only things that changes are build command and the connections number.

applications/matter_bridge/overlay-bt_max_connections_app_thread_nrf5340.conf
```diff
@@ -0,0 +1,23 @@
+#
+# Copyright (c) 2023 Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+#
+
+# Set 10 BLE connections, as it is an upper limit supported caused by using thread and BLE.
```
kkasperczyk-no, line 7: Please fix thread->Thread and wifi-> Wi-Fi in all places. The thread using lowercase reminds the thread in the operating system, while we mean Thread protocol name.
```suggestion
# Set 10 BLE connections, as it is an upper limit supported caused by using Thread and BLE.
```

applications/matter_bridge/overlay-bt_max_connections_app_thread_nrf5340.conf
```diff
@@ -0,0 +1,23 @@
...
+CONFIG_BT_MAX_CONN=10
+
+# Set buffer sizes in a consistent way with the ones used by the network core.
+#id if oncomment it there is an error that rx buffer is out of range (minimal value is 69)
+#CONFIG_BT_BUF_ACL_RX_SIZE=42
```
kkasperczyk-no, line 12: Commented out Kconfigs.

applications/matter_bridge/overlay-bt_max_connections_app_thread_nrf5340.conf
```diff
@@ -0,0 +1,23 @@
...
+#CONFIG_BT_BUF_ACL_TX_SIZE=42
+
+# Set MTU size to fit in the single buffer and avoid fragmentation (BUF_SIZE = MTU_SIZE + 4 B of L2CAP header).
+CONFIG_BT_L2CAP_TX_MTU=80
```
kkasperczyk-no, line 16: Looking at the equation in the comment, should it be 80, considering that buffer size is 42? it means the MTU will be bigger than a buffer and will not fit in it.

applications/matter_bridge/pm_static_nrf5340dk_nrf5340_cpuapp nrf70ek.yml
```diff
@@ -0,0 +1,80 @@
+mcuboot:
```
kkasperczyk-no, line 1: Two things:
1. You have changed the pm_static name, by giving it a suffix which means it will be selected only if prj.conf with such suffix will be selected. And such prj_nrf70ek.conf does not exist. You have to create it, and please make sure that all related ipc configuration files are pulled correctly.
2. You have to describe the new prj.conf and a way of building nrf70ek+nrf5340dk in the readme.
3. You have a blank space in the pm_static name, which will certainly lead into problems :)

applications/matter_bridge/doc/matter_bridge_description.rst
wiba-nordic: I'll take a look once the doc build finishes. Hopefully there are parts we can move to be outside of the tabs, and maybe some parts we can have as includes.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -603,7 +662,14 @@ The application supports the following configurations:
      - All from `Requirements`_
      - Debug version of the application.
 
-       Enables additional features for verifying the application behavior, such as logs.
+      Enables additional features for verifying the application behavior, such as logs.
```
wiba-nordic, line 665:
```suggestion
       Enables additional features for verifying the application behavior, such as logs.
```
One missing space manages to break the doc build.

applications/matter_bridge/overlay-bt_max_connections_app_thread_nrf5340.conf
```diff
@@ -0,0 +1,25 @@
...
+CONFIG_BT_L2CAP_TX_MTU=65
+
+# Set max number of bridged BLE devices, which is CONFIG_BT_MAX_CONN-1, as 1 connection is reserved for Matter.
+CONFIG_BRIDGE_MAX_BRIDGED_DEVICES_NUMBER=9
+CONFIG_BT_MAX_PAIRED=9
+
+# Assume that every bridged device uses only 1 endpoint, however it can be increased if specific use case requires it.
+CONFIG_BRIDGE_MAX_DYNAMIC_ENDPOINTS_NUMBER=9
+
+CONFIG_NRF_802154_RX_BUFFERS=14
```
ArekBalysNordic, line 25: Can we add a small description here, why do we reduce RX buffers for radio? I suspect that it's due to a lack of free RAM, but we'll forget about it shortly :)

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,83 @@ Use the following configuration options to customize the number of supported bri
...
+            The current maximum number of Bluetooth LE connections that can be selected using the default configuration is ``10``.
+            You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+
+            .. parsed-literal::
+               :class: highlight
+
+               west build -b nrf5340dk/nrf5340/cpuapp -p -- -Dmatter_bridge_SHIELD=nrf7002ek -DSB_CONFIG_WIFI_PATCHES_EXT_FLASH_STORE=y -DSB_CONFIG_DFU_MULTI_IMAGE_PACKAGE_WIFI_FW_PATCH=y -DSB_CONFIG_WIFI_NRF700X=y -Dmcuboot_CONFIG_UPDATEABLE_IMAGE_NUMBER=3 -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf" -DFILE_SUFFIX=nrf70ek
+
+         .. group-tab:: Matter-Bridge over Thread configuration
```
ArekBalysNordic, line 551: You mentioned Matter Bridge over Thread only here and in the table below, but it is just a tab, so it is not visible enough. Shouldn't we inform about that somewhere at the bottom of this page, or somewhere else in our docs as well? I have a feeling that this is a bit of hidden information. @kkasperczyk-no what do you think?

ArekBalysNordic: For example here: 
image
there is mention that you can use only Wi-Fi as a transport for Matter in Matter Bridge.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,83 @@ Use the following configuration options to customize the number of supported bri
...
-The current maximum number of Bluetooth LE connections that can be selected using the default configuration is ``10``.
-You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
-Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+   You have to consider several factors to decide how many Bluetooth LE bridged devices to support.
+   Every Bluetooth LE bridged device uses a separate Bluetooth LE connection, so you need to consider the number of supported Bluetooth LE connections (selected using the :kconfig:option:`CONFIG_BT_MAX_CONN` Kconfig option) when you configure the number of devices.
+   Since the Matter stack uses one Bluetooth LE connection for commissioning, the maximum number of connections you can use for bridged devices is one less than is configured using the :kconfig:option:`CONFIG_BT_MAX_CONN` Kconfig option.
```
wiba-nordic, line 533: This part needs to be outside of the tabs. It disappears from the documentation once you select one of the DK tabs since it's not inside any of them.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,83 @@ Use the following configuration options to customize the number of supported bri
...
 
-   west build -b nrf7002dk/nrf5340/cpuapp -- -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
+      .. tabs::
 
-.. parsed-literal::
-   :class: highlight
+         .. group-tab:: Matter-Bridge over Wi-Fi configuration
+
+            Increasing the number of Bluetooth LE connections affects the RAM usage on both the application and network cores.
+            The current maximum number of Bluetooth LE connections that can be selected using the default configuration is ``10``.
```
wiba-nordic, line 542: These two lines also stay constant for all tabs, so they can be moved completely outside of the tabs.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -526,24 +526,83 @@ Use the following configuration options to customize the number of supported bri
...
-You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
-Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+   You have to consider several factors to decide how many Bluetooth LE bridged devices to support.
+   Every Bluetooth LE bridged device uses a separate Bluetooth LE connection, so you need to consider the number of supported Bluetooth LE connections (selected using the :kconfig:option:`CONFIG_BT_MAX_CONN` Kconfig option) when you configure the number of devices.
+   Since the Matter stack uses one Bluetooth LE connection for commissioning, the maximum number of connections you can use for bridged devices is one less than is configured using the :kconfig:option:`CONFIG_BT_MAX_CONN` Kconfig option.
 
-.. parsed-literal::
-   :class: highlight
+   .. group-tab:: nRF53 DKs
```
wiba-nordic, line 535: For this tab level, something before the inner tabs needs to be added to each tab. Since the only difference is going to be which configurations are supported, I think something like this works:
"The way you configure the number of Bluetooth LE bridged devices depends on the configuration you are using."

The only downside is that it doesn't really work for the 70 series, and I'm not sure what to add there instead.

wiba-nordic (changes requested):
Some initial thoughts, I'll provide more next week.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -532,18 +532,70 @@ Since the Matter stack uses one Bluetooth LE connection for commissioning, the m
 
 Increasing the number of Bluetooth LE connections affects the RAM usage on both the application and network cores.
 The current maximum number of Bluetooth LE connections that can be selected using the default configuration is ``10``.
-You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
-Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
 
-.. parsed-literal::
-   :class: highlight
+.. tabs::
```
wiba-nordic, line 537:
```suggestion
You can configure the number of Bluetooth LE bridged devices depending on your device and configuration:
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -532,18 +532,70 @@ Since the Matter stack uses one Bluetooth LE connection for commissioning, the m
...
+
+            You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
+
+             .. parsed-literal::
+               :class: highlight
+
+               west build -b nrf54h20dk/nrf54h20/cpuapp -p -- -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf"
+
+   .. group-tab:: nRF70 DKs
```
wiba-nordic, line 586:
```suggestion
   .. group-tab:: nRF70 DKs
   
      The nRF70 Series supports the Matter bridge over Wi-Fi configuration.
```
Bit of an odd

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -532,18 +532,70 @@ Since the Matter stack uses one Bluetooth LE connection for commissioning, the m
...
+
+      .. tabs::
+
+         .. group-tab:: Matter-Bridge over Wi-Fi configuration
```
wiba-nordic, line 590:
```suggestion
         .. group-tab:: Matter-Bridge over Wi-Fi
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -532,18 +532,70 @@ Since the Matter stack uses one Bluetooth LE connection for commissioning, the m
...
+
+            You can increase the number of Bluetooth LE connections if you decrease the size of the Bluetooth LE TX/RX buffers used by the Bluetooth controller, but this will decrease the communication throughput.
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
```
wiba-nordic, line 593:
```suggestion

            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the sizes of Bluetooth LE TX/RX buffers:
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -604,6 +656,13 @@ The application supports the following configurations:
      - Debug version of the application.
 
        Enables additional features for verifying the application behavior, such as logs.
+   * - Debug
+     - :file:`prj_nrf70ek.conf`
+     - ``nrf70ek``
+     - ``nrf5340dk``
+     - Debug version of the application.
+
+       Enables Matter-Bridge to work with Wi-Fi.
```
wiba-nordic, line 665: Aligning more with the Thread and Wi-Fi switching entry for the [Matter door lock configurations](https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/samples/matter/lock/README.html#matter_door_lock_custom_configurations)
image
```suggestion
   * - Matter bridge over Wi-Fi
     - :file:`prj_nrf70ek.conf`
     - ``nrf70ek``
     - nRF5340 DK with the nRF7002 EK shield attached
     - Debug version of the application with Matter over Wi-Fi enabled.
```
Should also be moved to be the last entry in the list, I think. We don't really have much precedent beyond the Door lock sample, which places the the Debug and Release configurations first.

Is this configuration only used when working with the nRF5340 DK and the shield? If so, that might need to be added to the description.

wiba-nordic (changes requested):

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -611,6 +670,13 @@ The application supports the following configurations:
      - Release version of the application.
 
        Enables only the necessary application functionalities to optimize its performance.
+   * - Matter bridge over Wi-Fi
+     - :file:`prj_nrf70ek.conf`
+     - ``nrf70ek``
+     - nRF5340 DK with the nRF7002 EK shield attached
+     - Debug version of the application with Matter over Wi-Fi enabled.
+
+       Enables Matter-Bridge to work with Wi-Fi.
```
kkasperczyk-no, line 679:
```suggestion
       Enables the Matter Bridge to work with Wi-Fi on nRF5340 DK.
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -611,6 +670,13 @@ The application supports the following configurations:
      - Release version of the application.
 
        Enables only the necessary application functionalities to optimize its performance.
+   * - Matter bridge over Wi-Fi
```
kkasperczyk-no, line 673:
```suggestion
   * - Matter bridge over Wi-Fi with nRF5340 DK and nRF7002 EK
```

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -611,6 +670,13 @@ The application supports the following configurations:
...
+     - :file:`prj_nrf70ek.conf`
```
kkasperczyk-no, line 674: This probably could be removed.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -656,7 +722,7 @@ To enable the Matter smart plug functionality, run the following command:
 
       .. code-block:: console
 
-         west build -b nrf54h20dk/nrf54h20/cpuapp -p -- -DSB_CONFIG_WIFI_NRF700X=y -DCONFIG_CHIP_WIFI=y -Dmatter_bridge_SHIELD=nrf700x_nrf54h20dk -DCONFIG_BRIDGED_DEVICE_BT=y -Dmatter_bridge_SNIPPET=onoff_plug
+         west build -b nrf54h20dk/nrf54h20/cpuapp -p -- -DSB_CONFIG_Wi-Fi_NRF700X=y -DCONFIG_CHIP_Wi-Fi=y -Dmatter_bridge_SHIELD=nrf700x_nrf54h20dk -DCONFIG_BRIDGED_DEVICE_BT=y -Dmatter_bridge_SNIPPET=onoff_plug
```
kkasperczyk-no, line 725: Please verify all configs using WIFI in this readme, as I think that you replaced all WIFI -> Wi-Fi occurrences what broke multiple build commands.

applications/matter_bridge/doc/matter_bridge_description.rst
```diff
@@ -532,18 +532,77 @@ Since the Matter stack uses one Bluetooth LE connection for commissioning, the m
...
+            Build the target using the following command in the project directory to enable a configuration that increases the number of Bluetooth LE connections to ``20`` by decreasing the size of Bluetooth LE TX/RX buffers:
+
+            .. parsed-literal::
+               :class: highlight
+
+               west build -b nrf5340dk/nrf5340/cpuapp -p -- -Dmatter_bridge_SHIELD=nrf7002ek -DSB_CONFIG_WIFI_PATCHES_EXT_FLASH_STORE=y -DSB_CONFIG_DFU_MULTI_IMAGE_PACKAGE_WIFI_FW_PATCH=y -DSB_CONFIG_WIFI_NRF700X=y -Dmcuboot_CONFIG_UPDATEABLE_IMAGE_NUMBER=3 -DCONFIG_BRIDGED_DEVICE_BT=y -DEXTRA_CONF_FILE="overlay-bt_max_connections_app.conf" -Dipc_radio_EXTRA_CONF_FILE="overlay-bt_max_connections_net.conf" -DFILE_SUFFIX=nrf70ek
+
+         .. group-tab:: Matter bridge over Thread
+
+            You can not increase the default number of Bluetooth LE connections in this configuration.
```
kkasperczyk-no, line 559: This paragraph is not quite true. You can't increase the number of supported Bluetooth LE connections using an overlay. But you can always do that by modifying the board files and decreasing the buffer sizes. I would also add a statement here that the default number of supported devices here is 9.

applications/matter_bridge/doc/matter_bridge_description.rst
wiba-nordic: For an empty entry in a table, you can use `---` as the value.

applications/matter_bridge/doc/matter_bridge_description.rst
wiba-nordic: Wording out Kamil's suggestion:
```suggestion
            You can not increase the default number of Bluetooth LE connections in this configuration using overlays.
```
And the following after Line 560:
```
            You can still increase the number of connections by modifying the board files and decreasing the buffer sizes.
            The default number of connections is nine.
```

applications/matter_bridge/doc/matter_bridge_description.rst
wiba-nordic: I don't think the devices need to be in the configuration name, in case we add more supported devices later.

kkasperczyk-no: But the thing is that there are already more devices that support Matter bridge over Wi-Fi and this particular configuration is related only to nrf5340dk+nrf70ek. Other configurations like nrf54h20+nrf70ek or nrf7002dk are built using other way.

applications/matter_bridge/doc/matter_bridge_description.rst
wiba-nordic:
```suggestion
       Enables Matter bridge to work over Wi-Fi.
```
We don't need to specify the device here either, since there's a specific entry for supported devices in the table. We also don't mention the devices for the similar network switching in the Door lock sample.

applications/matter_bridge/doc/matter_bridge_description.rst
wiba-nordic: Ah, I missed that part. In that case, it makes since to keep the device mentions in both places. 

```suggestion
   * - Matter bridge over Wi-Fi with the nRF5340 DK and nRF7002 EK
```
Note the definitive article.

wiba-nordic (approved):
Looks good from my side.

applications/matter_bridge/doc/matter_bridge_description.rst
kkasperczyk-no: @kg-nordicsemi could you address this one?

applications/matter_bridge/sample.yaml
```diff
@@ -118,6 +119,95 @@ tests:
       - nrf54h20dk/nrf54h20/cpuapp
     platform_allow: nrf54h20dk/nrf54h20/cpuapp
     tags: sysbuild ci_applications_matter
+  applications.matter_bridge.nrf54h20.br_ble.thread:
```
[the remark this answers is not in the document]
kkasperczyk-no: Discussed on a dedicated call. It will be optimized in the follow-up task.

samples/matter/common/src/bridge/Kconfig
```diff
@@ -48,4 +48,32 @@ config BRIDGE_BT_MINIMUM_SECURITY_LEVEL
 	   allowed to connect to the bridge. The level cannot be set to a value smaller than 2, as it would mean that
 	   the communication with the connected devices will not be encrypted, what is not considered to be secure.
 
+config BRIDGE_BT_SCAN_WINDOW
+	int "Time duration of a central actively scanning for devices withing scan interval, it is equal BRIDGE_BT_SCAN_WINDOW*0.625 ms"
```
kkasperczyk-no, line 52:
```suggestion
	int "Time duration of a central actively scanning for devices within scan interval, it is equal BRIDGE_BT_SCAN_WINDOW*0.625 ms"
```

samples/matter/common/src/bridge/Kconfig
```diff
@@ -48,4 +48,32 @@ config BRIDGE_BT_MINIMUM_SECURITY_LEVEL
...
+	default 40  if NET_L2_OPENTHREAD
+	default 48
+
+config BRIDGE_BT_SCAN_INTERVAL
+	int "Time between consecutive Bluetooth scan windows, it is equal BRIDGE_BT_SCAN_INTERVAL*0.625 ms"
+	default 160 if NET_L2_OPENTHREAD
```
kkasperczyk-no, line 58: Probably you could add similar comment in case of settings different defaults for Thread.
```suggestion
	default 160 if NET_L2_OPENTHREAD # It is necessary to decrease Bluetooth LE activity to leave some radio time to Thread
```

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -433,15 +433,31 @@ CHIP_ERROR BLEConnectivityManager::Init(const bt_uuid **serviceUuids, uint8_t se
...
 #endif /* CONFIG_BT_SMP */
 
+	struct bt_le_scan_param scan_param = {
+		.type = BT_LE_SCAN_TYPE_PASSIVE,
+		.options = BT_LE_SCAN_OPT_FILTER_DUPLICATE,
+		.interval = CONFIG_BRIDGE_BT_SCAN_INTERVAL,
+		.window = CONFIG_BRIDGE_BT_SCAN_WINDOW,
+	};
+
+	struct bt_le_conn_param conn_param_ = {
```
kkasperczyk-no, line 443: Additional underscore?
```suggestion
	struct bt_le_conn_param conn_param = {
```

samples/matter/common/src/bridge/ble_connectivity_manager.cpp
```diff
@@ -433,15 +433,31 @@ CHIP_ERROR BLEConnectivityManager::Init(const bt_uuid **serviceUuids, uint8_t se
...
+		.interval_min = CONFIG_BRIDGE_BT_CONNECTION_INTERVAL_MIN,
+		.interval_max = CONFIG_BRIDGE_BT_CONNECTION_INTERVAL_MAX,
+		.latency = CONFIG_BRIDGE_BT_CONNECTION_LATENCY,
+		.timeout = CONFIG_BRIDGE_BT_CONNECTION_TIMEOUT,
+	};
+
 	bt_scan_init_param scan_init = {
+		.scan_param = &scan_param,
 		.connect_if_match = 0,
+		.conn_param = &conn_param_,
```
kkasperczyk-no, line 453: Ditto
```suggestion
		.conn_param = &conn_param,
```

kkasperczyk-no (commented):
Please add release note entry to inform that Bridge is now supported with Thread

applications/matter_bridge/doc/matter_bridge_description.rst
kg-nordicsemi: It turned out that the prj_nrf70ek.conf file is unnecessary when it comes to building, so I removed it, but I forgot to make changes to the documentation, I'll change it now

doc/nrf/releases_and_maturity/releases/release-notes-changelog.rst
```diff
@@ -219,6 +219,7 @@ Matter Bridge
     By default, the option points to the :file:`src/default_zap` directory and can be changed to any path relative to application's location that contains the ZAP file and :file:`zap-generated` directory.
   * Support for the :ref:`zephyr:nrf54h20dk_nrf54h20`.
   * Optional smart plug device functionality.
+  * Support for the Thread network.
```
kkasperczyk-no, line 222:
```suggestion
  * Support for the Thread protocol.
```

kkasperczyk-no (approved):
One more minor thing.

ArekBalysNordic (approved):

PR #16701, opened 2024-08-01
doc: samples: zigbee: extend light switch sample tx power info

edmont:
Clarify that tx power is now set to 0 dBm for light switch.


milewr (approved):
Ok for me, Wille should confirm too.

wiba-nordic (approved):

PR #16702, opened 2024-08-01
samples: net: https_client: Fix client cert match

eivindj-nordic:
Allow compare to not be strict about ending newline when comparing
the given certificate with the one in the modem.

### DIFF
--- a/samples/net/https_client/src/main.c
+++ b/samples/net/https_client/src/main.c
@@ -95,11 +95,11 @@ int cert_provision(void)
 		}
 	}
 
-	printk("Provisioning certificate\n");
+	printk("Provisioning certificate to the modem\n");
 
 	/*  Provision certificate to the modem */
 	err = modem_key_mgmt_write(TLS_SEC_TAG, MODEM_KEY_MGMT_CRED_TYPE_CA_CHAIN, cert,
-				   sizeof(cert) - 1);
+				   sizeof(cert));
 	if (err) {
 		printk("Failed to provision certificate, err %d\n", err);
 		return err;
--- a/lib/modem_key_mgmt/modem_key_mgmt.c
+++ b/lib/modem_key_mgmt/modem_key_mgmt.c
@@ -225,6 +225,14 @@ int modem_key_mgmt_cmp(nrf_sec_tag_t sec_tag,
 		goto out;
 	}
 
+	if (*(end - 1) == '\n') {
+		end--;
+	}
+
+	if (((char *)buf)[len - 1] == '\n') {
+		len--;
+	}
+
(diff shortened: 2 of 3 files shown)


lemrey:
We should probably go through the documentation and usage of `modem_key_mgmt_write()` and `modem_key_mgmt_cmp()` and ensure that the intended usage is consistent and properly documented.

What does the documentation say and does it match what the modem expects?
Is the usage aligned across all samples and libraries in NCS?

eivindj-nordic:
> We should probably go through the documentation and usage of `modem_key_mgmt_write()` and `modem_key_mgmt_cmp()` and ensure that the intended usage is consistent and properly documented.
> 
> What does the documentation say and does it match with what the modem expects? Is the usage aligned across all samples and libraries in NCS? 

Usage is aligned in samples and libraries in NCS after this change, this is the only case where one character was subtracted. Modem accepts the certificate both with and without the newline. I've updated the `modem_key_mgmt_cmp` to accept comparing certificates with and without the newline at the end, so it should be a bit more robust now. Was thinking of using `-----END CERTIFICATE-----` string instead, but I don't think that is needed.

divipillai (approved):

lemrey (approved):
Could we add a snippet to the library documentation to show how to use the API, so that we can use that as a reference?

eivindj-nordic:
> Could we add a snippet to the library documentation to show how to use the API, so that we can use that as a reference?

Ok, I'll make a separate PR for that.

samples/net/https_client/src/main.c
```diff
@@ -95,11 +95,11 @@ int cert_provision(void)
 		}
 	}
 
-	printk("Provisioning certificate\n");
+	printk("Provisioning certificate to the modem\n");
 
 	/*  Provision certificate to the modem */
 	err = modem_key_mgmt_write(TLS_SEC_TAG, MODEM_KEY_MGMT_CRED_TYPE_CA_CHAIN, cert,
-				   sizeof(cert) - 1);
```
gregersrygg, line 102: Can you please explain why this change is needed? Has newline handling changed in `modem_key_mgmt_write`, or was our use incorrect?

eivindj-nordic: Before this PR, the certificate was written without the last character (`\n` at the end), while in `modem_key_mgmt_cmp()` it is included. So the certificates did not match and it was rewritten. The PR (not removing the last character) aligns with use elsewhere in NCS, while we also add logic to `modem_key_mgmt_cmp()` to make it accept both.

gregersrygg (approved):

PR #16703, opened 2024-08-01
tests: lib: lte_lc_api: Extending test coverage

trantanen:
Adding more test coverage to have 100% function coverage and increasing branch coverage of the library from ~69% to 82% by testing ~100 more branches out of the ~870.

Few small bug fixes found during test development.


tests/lib/lte_lc_api/src/lte_lc_api_test.c
```diff
@@ -1574,6 +2409,13 @@ void test_lte_lc_modem_sleep_event(void)
 	test_event_data[4].modem_sleep.type = LTE_LC_MODEM_SLEEP_FLIGHT_MODE;
 	test_event_data[4].modem_sleep.time = 0;
 	at_monitor_dispatch(at_notif);
+
+	/* No event for unknown type */
+	strcpy(at_notif, "%XMODEMSLEEP: 99,0\r\n");
+	test_event_data[4].type = LTE_LC_EVT_MODEM_SLEEP_EXIT;
+	test_event_data[4].modem_sleep.type = LTE_LC_MODEM_SLEEP_FLIGHT_MODE;
+	test_event_data[4].modem_sleep.time = 0;
```
tokangas, line 2417: This should be removed, right?

tokangas (approved):
Great, some minor comments.

lib/lte_link_control/lte_lc.c
```diff
@@ -706,6 +707,8 @@ static int connect_lte(bool blocking)
 	if (err || !blocking) {
 		goto exit;
 	}
+	/* We do not really need to restore original functional mode before setting it to normal */
+	original_func_mode_valid = true;
```
tokangas, line 711: Just a suggestion, if the name of the variable would be for example `func_mode_changed`, the code would be self-explanatory, I think.

tests/lib/lte_lc_api/src/lte_lc_api_test.c
trantanen: Yep

lib/lte_link_control/lte_lc.c
trantanen: Ok, I'll change it and remove the comment.

lib/lte_link_control/lte_lc.c
```diff
@@ -704,6 +705,7 @@ static int connect_lte(bool blocking)
 	if (err || !blocking) {
 		goto exit;
 	}
+	func_mode_changed = true;
```
MirkoCovizzi, line 708: Nit: empty line above this one

tests/lib/lte_lc_api/src/lte_lc_api_test.c
```diff
@@ -254,7 +259,6 @@ void test_lte_lc_deregister_handler_null(void)
 	TEST_ASSERT_EQUAL(-EINVAL, ret);
 }
 
-
```
MirkoCovizzi, line 257: There is another one of this extra newlines right under `sys_init_helper`.

MirkoCovizzi (commented):
First pass, I noticed that many tests have `fail` inside the name rather than at the end. I think it would be better to keep `success` and `fail` at the end of the name.

tests/lib/lte_lc_api/src/lte_lc_api_test.c
```diff
@@ -1962,6 +3051,109 @@ void test_lte_lc_periodic_search_get_success(void)
...
+	TEST_ASSERT_EQUAL(21, cfg.patterns[3].table.val_2);
+	TEST_ASSERT_EQUAL(31, cfg.patterns[3].table.val_3);
+	TEST_ASSERT_EQUAL(EXIT_SUCCESS, ret);
+}
+
+void test_lte_lc_reduced_mobility_set_success(void)
+{
+	int ret;
+
+	__mock_nrf_modem_at_printf_ExpectAndReturn("AT%REDMOB=0", 0);
```
MirkoCovizzi, line 3102: Rather than hardcoding the integer value to be sent, perhaps it would be better to sprintf the string using the actual mode value. This way if they get changed or reordered, the tests don't need to be updated.
Same for similar cases.

tests/lib/lte_lc_api/src/lte_lc_api_test.c
trantanen: This is a removal of an extra empty line to make 2 empty lines between functions to 1 empty line. So I guess this is fine, right?
I've removed an extra empty line after `sys_init_helper`.

trantanen:
> First pass, I noticed that many tests have `fail` in the middle of the name rather than at the end. I think it would be better to keep `success` and `fail` at the end of the name.

My logic has been as follows:
tests_<function_to_test_[success|fail]_<more_description_of_test_ie_what_fails>

I don't have strong opinions on the format as long as it's consistent. Unfortunately we can only be consistent within a test set as there is no repo-wide guidance.

PS. I'd use (and I've used elsewhere) `ok` instead of `success`.

tests/lib/lte_lc_api/src/lte_lc_api_test.c
trantanen: I specifically want to hardcode the integer value. This makes the test more readable in my opinion showing clearly the AT command that we are expecting lte_lc to send. We also avoid putting the same sprintf code into test case and real code.
More importantly, this integer is something that is specified in AT specs so we must not change it. Having said that, if a change would still be needed, we can modify the test case.
So, I'm trying to say that I want this test case to fail if somebody modifies the values of `lte_lc_reduced_mobility_mode` and same applies for all other AT commands and integers in there. I've used sprintf few times in some loops and Kconfig values.

trantanen:
> > First pass, I noticed that many tests have `fail` in the middle of the name rather than at the end. I think it would be better to keep `success` and `fail` at the end of the name.
> 
> My logic has been as follows: tests_<function_to_test_[success|fail]_<more_description_of_test_ie_what_fails>
> 
> I don't have strong opinions on the format as long as it's consistent. Unfortunately we can only be consistent within a test set as there is no repo-wide guidance.

I've now modified all old and new `fail` and `success` words to the end of test function name.

> PS. I'd use (and I've used elsewhere) `ok` instead of `success`.

Seems `success` is used in many test sets so keeping that.

MirkoCovizzi (approved):
Looks good!

tests/lib/lte_lc_api/src/lte_lc_api_test.c
MirkoCovizzi: That makes sense, thank you.

PR #16709, opened 2024-08-01
bluetooth: fast_pair: use_case: fix URL rendering in Kconfig help

kapi-no:
Fixed the URL rendering in the Kconfig help description for Fast Pair use case options.


kapi-no:
It seems that it is not possible to my links clickable. Closing my PR

PR #16714, opened 2024-08-01
wifi: Fix QSPI clock divider usage

krish2718:
cherry-pick fixes from QSPI-NOR driver.

@anangl FYI.

### DIFF
--- a/drivers/wifi/nrf700x/src/qspi/src/qspi_if.c
+++ b/drivers/wifi/nrf700x/src/qspi/src/qspi_if.c
@@ -93,6 +93,14 @@ BUILD_ASSERT(INST_0_SCK_FREQUENCY >= (NRF_QSPI_BASE_CLOCK_FREQ / 16),
 #error "Unsupported base clock divider for wake-up frequency."
 #endif
 
+/* After the base clock divider is changed, some time is needed for the new
+ * setting to take effect. This value specifies the delay (in microseconds)
+ * to be applied to ensure that the clock is ready when the QSPI operation
+ * starts. It was measured with a logic analyzer (unfortunately, the nRF5340
+ * specification does not provide any numbers in this regard).
+ */
+#define BASE_CLOCK_SWITCH_DELAY_US 7
+
 #else
 /*
  * On nRF52 Series SoCs, the base clock divider is not configurable,
@@ -357,6 +365,7 @@ static inline void qspi_lock(const struct device *dev)
 	 */
 #if defined(CONFIG_SOC_SERIES_NRF53X)
 	nrf_clock_hfclk192m_div_set(NRF_CLOCK, BASE_CLOCK_DIV);
+	k_busy_wait(BASE_CLOCK_SWITCH_DELAY_US);
 #endif
 }
 
@@ -366,6 +375,7 @@ static inline void qspi_unlock(const struct device *dev)
 	/* Restore the default base clock divider to reduce power consumption.
 	 */
 	nrf_clock_hfclk192m_div_set(NRF_CLOCK, NRF_CLOCK_HFCLK_DIV_4);
+	k_busy_wait(BASE_CLOCK_SWITCH_DELAY_US);
 #endif
 
 #ifdef CONFIG_MULTITHREADING
@@ -658,6 +668,7 @@ static inline void qspi_fill_init_struct(nrfx_qspi_config_t *initstruct)
 
 	/* Configure physical interface */
 	initstruct->phy_if.sck_freq = INST_0_SCK_CFG;
+
 	/* Using MHZ fails checkpatch constant check */
 	if (INST_0_SCK_FREQUENCY >= 16000000) {
 		qspi_config->qspi_slave_latency = 1;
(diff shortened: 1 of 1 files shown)


busineni (approved):

drivers/wifi/nrf700x/src/qspi/src/qspi_if.c
```diff
@@ -59,18 +59,43 @@ BUILD_ASSERT(INST_0_SCK_FREQUENCY >= (NRF_QSPI_BASE_CLOCK_FREQ / 16),
...
+ * starts. It was measured with a logic analyzer (unfortunately, the nRF5340
+ * specification does not provide any numbers in this regard).
+ */
+#define BASE_CLOCK_SWITCH_DELAY_US 7
+#if NRF53_ERRATA_159_ENABLE_WORKAROUND
+/* For 8 MHz, use HFCLK192M / 1 / (2*12) */
+#define INST_0_SCK_CFG_WAKE NRF_QSPI_FREQ_DIV12
+#else
 /* For 8 MHz, use HFCLK192M / 2 / (2*6) */
 #define INST_0_SCK_CFG_WAKE NRF_QSPI_FREQ_DIV6
```
leewkb4567, line 97: The setting of DIV6 assumes the base clock divider is /2.

Nevertheless, the lines above do not limit the QSPI clock frequency to 24MHz. When anomaly 159 workaround is not applied, the base clock divider may be /1 or /2. When the user specifies a 32MHz QSPI clock, the base clock divider is /1. The DIV6 assumption is broken then.

drivers/wifi/nrf700x/src/qspi/src/qspi_if.c
```diff
@@ -59,18 +59,43 @@ BUILD_ASSERT(INST_0_SCK_FREQUENCY >= (NRF_QSPI_BASE_CLOCK_FREQ / 16),
...
 #define INST_0_SCK_CFG (DIV_ROUND_UP(NRF_QSPI_BASE_CLOCK_FREQ / 2, \
 				     INST_0_SCK_FREQUENCY) - 1)
 #endif
+/* After the base clock divider is changed, some time is needed for the new
+ * setting to take effect. This value specifies the delay (in microseconds)
+ * to be applied to ensure that the clock is ready when the QSPI operation
+ * starts. It was measured with a logic analyzer (unfortunately, the nRF5340
+ * specification does not provide any numbers in this regard).
+ */
+#define BASE_CLOCK_SWITCH_DELAY_US 7
```
leewkb4567, line 91: The BASE_CLOCK_SWITCH_DELAY_US changes is spread across both commits. It is cleaner to confine the changes in a commit.

drivers/wifi/nrf700x/src/qspi/src/qspi_if.c
krish2718: > The setting of DIV6 assumes the base clock divider is /2.
> 
> Nevertheless, the lines above do not limit the QSPI clock frequency to 24MHz. When anomaly 159 workaround is not applied, the base clock divider may be /1 or /2. When the user specifies a 32MHz QSPI clock, the base clock divider is /1. The DIV6 assumption is broken then.

`INST_0_SCK_CFG_WAKE` is only 8MHz, hence we can assume either DIV6 or DIV1 based on anomaly159.

krish2718: another option is to explicitly set both `INST_0_SCK_CFG` and `INST_0_SCK_CFG_WAKE` for each combo like

```
#if BASE_CLOCK_DIV == NRF_CLOCK_HFCLK_DIV_1
/* For 8 MHz, use HFCLK192M / 1 / (2*12) */
#define INST_0_SCK_CFG_WAKE NRF_QSPI_FREQ_DIV12
#elif BASE_CLOCK_DIV == NRF_CLOCK_HFCLK_DIV_2
/* For 8 MHz, use HFCLK192M / 2 / (2*6) */
#define INST_0_SCK_CFG_WAKE NRF_QSPI_FREQ_DIV6
#else
#error "Unsupported base clock divider."
#endif
```

leewkb4567: Certainly, taking into account the BASE_CLOCK_DIV will get the INST_0_SCK_CFG_WAKE correctly defined.

When the 32MHz to 96MHz lines are removed, the driver supports only the following definitions:
- BASE_CLOCK_DIV --> NRF_CLOCK_HFCLK_DIV_2
- INST_0_SCK_CFG_WAKE --> NRF_QSPI_FREQ_DIV6

The driver code will be more concise. It should be easier to support going forward.

krish2718: Sure, but if I don't change the code from QSPI-NoR it makes it easier to maintain i.e., port changes in the future.

leewkb4567: It is not formally documented that nRF5340 is not recommended to drive nRF7002 by 32MHz QSPI clock. When the driver code has lines handling QSPI clock 32MHz to 96MHz, it is tempting for some curious users to try it out. Especially when nRF7002 PS states the QSPI interface supports up to 32MHz clock frequency.

A tone change is noticed. The previous message was the QSPI clock should be capped to 24MHz. Anyway, just my 2 cents.

krish2718: 32MHz is supported but only with high voltage, so, but > 32MHz isn't supported, anyways, IMHO, if a customer is reading internal code comments it's assumed that they know this stuff :).

